PR #3705, opened 2023-11-06
Double the web cpu request, reduce deploy cpu, and increase DJ by 50%.

yakloinsteak:
## Description

This lets us configure CPU requests (shares) for ECS services. I'll change the secrets yaml after this is merged to lock in the values for QA, but we need this merged and deployed everywhere before we can start setting the values for all the other environments.

## Testing and Docs

With this PR, we'll be able to add sections like this to the secrets yaml. I'll test QA with and without this and make sure it works as expected.
```
  :cpu_shares:
    :workoff: 256
    :cron: 256
    :web: 1024
    :deploy: 128
    :dj: 384
```

## Type of change
- [x] reliability improvement

## checklist before requesting review

### DIFF
--- a/config/deploy/docker/lib/deployer.rb
+++ b/config/deploy/docker/lib/deployer.rb
@@ -39,11 +39,6 @@ class Deployer
 
   attr_accessor :web_options
 
-  # github actions or other CI builds the image, so when true, don't try to
-  # build locally and push (which can be slow).
-  attr_accessor :assume_ci_build
-  attr_accessor :push_allowed
-
   # The fully-qualified domain name of the application
   # We use this so we can get data from the app about the deployment state
   attr_accessor :fqdn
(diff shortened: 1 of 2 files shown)


schwartzkylej (approved):
Looks good. Lets review once deployed to an environment. Thanks.

eanders:
@yakloinsteak stupid question, we have similar functionality for memory usage that uses dynamodb (I think) should we port that into this setup, or maybe use that setup for this?  It feels like we're wiggling 2 very similar nobs but choosing two very dissimilar approaches.

eanders (approved):
Looks good, but I'm not 100% sure on the yaml storage for the source of truth.

yakloinsteak:
I see the cpu as not being dynamically changing, while the RAM we try to adjust automatically. We can talk tomorrow about it.

yakloinsteak:
I'm going to merge this and proceed in the interests of saving time. dynamodb won't get these values encoded in git (in terraform) and it's going to be very hard to nail down the right values dynamically anyway given how ECS works and how this parameter serves multiple purposes (reserving minimum CPU _and_ specifying how to divide up capacity when there's contention). We want the same values for all installations initially anyway to test our theory about this being the problem with occasional timeouts.

PR #3713, opened 2023-11-08
Additional ProjectType 0 fixes for FY2024 changes

eanders:
## Description

This updates a few more places we seem to have missed the new project type 0.  I wish we had a more central way to handle these magic numbers. Open to suggestions.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/app/models/concerns/hud_reports/destinations.rb
+++ b/app/models/concerns/hud_reports/destinations.rb
@@ -66,10 +66,10 @@ module HudReports::Destinations
         ['TOTAL', leavers_clause],
         [
           'Total persons exiting to positive housing destinations',
-          a_t[:project_type].in([1, 2]).
+          a_t[:project_type].in([0, 1, 2]).
             and(a_t[:destination].in(positive_destinations(1))).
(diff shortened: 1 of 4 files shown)


app/models/concerns/hud_reports/destinations.rb
```diff
@@ -66,10 +66,10 @@ module HudReports::Destinations
         ['TOTAL', leavers_clause],
         [
           'Total persons exiting to positive housing destinations',
-          a_t[:project_type].in([1, 2]).
+          a_t[:project_type].in([0, 1, 2]).
```
eanders, line 69: @ttoomey tagging you on these because I believe this is right (I looked a bit at the spec to confirm) but wanted to confirm I'm not missing some reason ES Entry/Exit wasn't included here.

ttoomey: @eanders thanks, I think you're right. It looks like I just copied the logic from the previous report without adding ES-EE. Sorry I missed that!

awisema (approved):

PR #3723, opened 2023-11-13
Patch unit creation user_id

gigxz:
## Description

Patch bug in unit creation where we incorrectly stored HUD User UserID instead of the application user id.

PT: https://www.pivotaltracker.com/story/show/186428559

## Type of change
- [x] Bug fix


eanders (approved):

ttoomey (approved):
Thanks @gigxz!

One question I had was how we want to deal with the current unit user ids, which are likely all incorrect. Should we include a migration to reset them to system_user? Or would we do that manually after deploy?

gigxz:
Good question, I will do a manual change to set them to system user.

PR #3724, opened 2023-11-13
CsvTransformer default to NBN when missing TrackingMethod

gigxz:
## Description

**Old behavior**: If a row in a 2022 file has ProjectType 1 and TrackingMethod is empty, change the ProjectType to 0 (Entry/Exit)
**New behavior**: If a row in a 2022 file has ProjectType 1 and TrackingMethod is empty, keep the ProjectType as 1 (NBN)

Support request: https://www.pivotaltracker.com/story/show/186466712

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hud_twenty_twenty_two_to_twenty_twenty_four/app/models/hud_twenty_twenty_two_to_twenty_twenty_four/project/update_project_type.rb
+++ b/drivers/hud_twenty_twenty_two_to_twenty_twenty_four/app/models/hud_twenty_twenty_two_to_twenty_twenty_four/project/update_project_type.rb
@@ -10,6 +10,8 @@ module HudTwentyTwentyTwoToTwentyTwentyFour::Project
       project_type = if row['ProjectType'].to_i == 1
         if row['TrackingMethod'].to_i == 3
           1
+        elsif row['TrackingMethod'].nil?
+          1 # Default to NBN if TrackingMethod is missing
         else
           0
         end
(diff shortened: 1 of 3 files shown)


drivers/hud_twenty_twenty_two_to_twenty_twenty_four/app/models/hud_twenty_twenty_two_to_twenty_twenty_four/project/update_project_type.rb
```diff
@@ -10,6 +10,8 @@ def process(row)
       project_type = if row['ProjectType'].to_i == 1
         if row['TrackingMethod'].to_i == 3
           1
+        elsif row['TrackingMethod'].nil?
+          1 # Default to NBN if TrackingMethod is missing
```
eanders, line 14: This seems like a bad idea to me.  What's the reasoning behind it?  Historically, Project Type 1 with a non-3 TrackingMethod would be Entry/Exit.

gigxz: Hmm okay, I didn't know that. The purpose is to handle importing this hybrid-22-24 project file which has 0 and 1 project types but no Tracking Methods. I can try to fix it in [the custom importer](https://github.com/greenriver/hmis-warehouse/blob/release-93/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb) instead, that just seemed more difficult so figured I would propose this first.

eanders: Maybe an importer extension hook?  These can be enabled/disabled per installation.  Something like: `drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/force_valid_enrollment_coc.rb`

gigxz: https://github.com/greenriver/hmis-warehouse/pull/3727

PR #3733, opened 2023-11-14
186437207 - Update TalentLMS form to utilize edit & new paths

dtgreiner:
## Description

Update TalentLMS config tool to use conventional rails 'new' and 'edit' pages. 

## Type of change
- [x] Bug fix
- [ ] Dependency update


app/controllers/admin/talentlms_controller.rb
```diff
@@ -7,28 +7,57 @@
...
-      if config.valid?
-        @config.save
-        redirect_to action: :index
+      @config = config_scope.new
+      if @config.update(config_params)
+        respond_with(@config, location: edit_admin_talentlm_path(@config))
       else
-        render :index
+        render 'new'
       end
```
eanders: Pretty sure `respond_with` will handle the error situation, so this can probably just be:
```
@config = config_scope.create(config_params)
respond_with(@config, location: edit_admin_talentlm_path(@config))
```

app/controllers/admin/talentlms_controller.rb
```diff
@@ -7,28 +7,57 @@
...
+
+    private def set_config
+      @config = config_scope.first_or_initialize
+
+      # Checks to make sure we are working with the expected record in the database.
+      # If a record already exists, we want to be editing that record. If one does
+      # not exist, we want to be creating it.
+      # If we ever have the need for multiple records at once, this will need to be adjusted.
+      redirect_to new_admin_talentlm_path if @config.new_record?
+      redirect_to edit_admin_talentlm_path(@config) if @config.id && @config.id != params[:id].to_i
```
eanders: This seems overly complicated, since `set_config` is really simple, and basically already integrated in `new` maybe we should just throw it away and add `@config = config_scope.first_or_initialize` to the top of each `index`, `edit` and `update`

eanders: Looks like maybe we just went a redirect in `index` too to send you to new or edit for now.

eanders (changes requested):
@dtgreiner this looks great.  Thanks!  Let's see if we can cleanup the controller a bit more, I'm not particularly fond of the `set_config` method in this case, though we use similar methods elsewhere.

app/controllers/admin/talentlms_controller.rb
dtgreiner: That all makes sense and I can definitely run through this again. I wasn't sure how complicated we wanted the logic to get with the expectation that there's only one record in the database. For example if the new action is called when a record already exists, should that be recognized and redirected to an edit? Similarly with the edit if the wrong id is passed.

dtgreiner:
@eanders updates to address the raised concerns have been pushed.

eanders (approved):
Looks much cleaner, thank you!

PR #3746, opened 2023-11-17
Homepage without database hits

yakloinsteak:
## Description

We think that slow probe times are due to database load, and the probes hit the home page. 

I'm not sure how frequently the links table changes, so we could use redis or some other preferred way of memoizing in this application.

## Type of change

### DIFF
--- a/app/controllers/admin/links_controller.rb
+++ b/app/controllers/admin/links_controller.rb
@@ -20,6 +20,7 @@ module Admin
 
     def create
       @link = link_scope.create(link_params)
+      link_scope.invalidate_cache
       respond_with(@link, location: admin_links_path)
     end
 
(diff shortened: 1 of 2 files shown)


yakloinsteak:
@schwartzkylej fyi.

app/models/link.rb
```diff
@@ -5,8 +5,10 @@
 ###
 
 class Link < ApplicationRecord
-  scope :for, ->(location) do
-    where(location: location)
+  def self.for(location)
+    @cache ||= {}
+
+    @cache[location] ||= where(location: location)
```
eanders, line 11: These do change (there's an interface for adding menu items)  Not 100% sure it's the same items, I'm not remembering the `:for` scope, butt we should make sure to invalidate this cache if the items are added/removed.

app/controllers/admin/links_controller.rb
```diff
@@ -28,11 +29,13 @@ def edit
 
     def update
       @link.update(link_params)
+      link_scope.invalidate_cache
       respond_with(@link, location: admin_links_path)
     end
 
     def destroy
       @link.destroy
+      link_scope.invalidate_cache
```
eanders, line 38: @yakloinsteak I _think_ this handles the invalidation, look good to you?

yakloinsteak: That looks great. I'm glad we could do it without redis in this particular case.

PR #3748, opened 2023-11-17
Assessment hook to force prioritized placement

dtgreiner:
## Description

Add hook to import that allows forcing prioritized placement to 1
186248549

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_validation/non_blank.rb
```diff
@@ -9,6 +9,14 @@ def self.check_validity!(item, column)
     value = item[column]
     return if value.present?
 
+    if item['source_type'] == 'HmisCsvTwentyTwentyFour::Loader::Assessment' && column.to_sym == :PrioritizationStatus
```
eanders, line 13:
```suggestion
    # The ForcePrioritizedPlacementStatus import hook will fix blank PrioritizationStatus, but because this field is null false, we need to skip that check if the hook is enabled.
    if item['source_type'] == 'HmisCsvTwentyTwentyFour::Loader::Assessment' && column.to_sym == :PrioritizationStatus
```

drivers/hmis_csv_importer/spec/factories/clean_up_data_sources.rb
```diff
@@ -49,6 +49,17 @@
     end
   end
 
+  factory :importer_force_prioritized_placement_status, class: 'GrdaWarehouse::DataSource' do
```
eanders, line 52: Were there some tests we could easily add to confirm this?  If you take the source files you created and drop them into the fixtures directory you should be able to check the when this is enabled all `PrioritizationStatus` cells are `1` after cleanup and not before.

eanders (changes requested):
Looks good!  I think a test as noted would be really helpful.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/force_prioritized_placement_status.rb
```diff
@@ -0,0 +1,46 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisCsvImporter::HmisCsvCleanup
+  class ForcePrioritizedPlacementStatus < Base
+    def cleanup!
+      assessment_batch = []
```
eanders: You don't need this, line 12 handles updating everything.
```suggestion
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/force_prioritized_placement_status.rb
```diff
@@ -0,0 +1,46 @@
...
+
+      assessment_scope.update(PrioritizationStatus: 1)
+
+      assessment_scope.import(
+        assessment_batch,
+        on_duplicate_key_update: {
+          conflict_target: conflict_target(assessment_scope),
+          columns: [:PrioritizationStatus, :source_hash],
+        },
+      )
```
eanders: Pretty sure you don't need this either.
```suggestion
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/force_prioritized_placement_status.rb
```diff
@@ -0,0 +1,46 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisCsvImporter::HmisCsvCleanup
+  class ForcePrioritizedPlacementStatus < Base
+    def cleanup!
+      assessment_batch = []
+
+      assessment_scope.update(PrioritizationStatus: 1)
```
eanders: This should probably use `update_all`  I think newer rails can use `update` but generally that's for single record updates.
```suggestion
      assessment_scope.update_all(PrioritizationStatus: 1)
```

drivers/hmis_csv_importer/spec/models/importer/twenty_twenty_four/force_prioritization_status_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+        version: 'AutoMigrate',
+        deidentified: true,
+        run_jobs: false,
+      )
+    end
+
+    it 'all assessments will have prioritization status set to 1' do
+      GrdaWarehouse::Hud::Assessment.all.each do |assessment|
+        expect(assessment.PrioritizationStatus).to eq(1)
+      end
```
eanders: This is fine, but could be written like this to only do one query.  Also, we should probably confirm we have the same number of 1s as we expect since there was an issue with the null values for a while.  Something like:
```suggestion
      stati = GrdaWarehouse::Hud::Assessment.all.pluck(:PrioritizationStatus)
      expect(stati.all?(1)).to be true
      expect(stati.count).to eq(5)
```

drivers/hmis_csv_importer/spec/models/importer/twenty_twenty_four/force_prioritization_status_spec.rb
```diff
@@ -0,0 +1,30 @@
...
+    end
```
eanders, line 43: Can you add one more test that confirms the values in the CSV file are not `all?(1)`  I'd like to have the tests prove that it actually changed things.

eanders (changes requested):
Thanks @dtgreiner !  Sorry I missed the bits in the actual cleanup before, I believe that can be really simple.  
Also two more tests please!

dtgreiner:
> Thanks @dtgreiner ! Sorry I missed the bits in the actual cleanup before, I believe that can be really simple. Also two more tests please!

dtgreiner:
> Thanks @dtgreiner ! Sorry I missed the bits in the actual cleanup before, I believe that can be really simple. Also two more tests please!

@eanders the import updates and additional tests have been added to the PR

eanders (approved):
Thanks @dtgreiner !

PR #3749, opened 2023-11-17
Bugfixes

eanders:
## Description

Fixes (but needs thorough QA):
* On live reports, if you choose only homeless project types, it lost ES after the first attempt (regression from FY2024 changes)
* Bugfix for Core Demographics where a client might be missing (usually just dev data)
* LSA kick-off page was checking SO for inventory related issues.  LSA doesn't work on SO, so limit checks to project types that are relevant.
* Some LSA checks were not checking Project Type 0
* Standardize checks for LSA to project types the filter knows about
* State-level reporting was trying to use the new race codes, but the census is unaware of them, which made it unhappy.
* Demographic Summary Excel download included the incorrect title. 

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/app/controllers/concerns/base_filters.rb
+++ b/app/controllers/concerns/base_filters.rb
@@ -22,7 +22,7 @@ module BaseFilters
 
     private def set_filter
       @filter = filter_class.new(user_id: current_user.id)
-      @filter.set_from_params(filter_params[:filters]) if filter_params[:filters].present?
+      @filter.update(filter_params[:filters]) if filter_params[:filters].present?
       @comparison_filter = @filter.to_comparison
     end
 
--- a/app/models/grda_warehouse/document_export.rb
+++ b/app/models/grda_warehouse/document_export.rb
@@ -23,7 +23,7 @@ class GrdaWarehouse::DocumentExport < GrdaWarehouseBase
     @filter ||= begin
       f = ::Filters::FilterBase.new(user_id: user.id)
       filter_params = params['filters'].presence&.deep_symbolize_keys
-      f.set_from_params(filter_params) if filter_params
+      f.update(filter_params) if filter_params
 
       f
     end
--- a/app/views/filters/filter_controls/project_types/_input.haml
+++ b/app/views/filters/filter_controls/project_types/_input.haml
@@ -8,6 +8,7 @@
   ctl_name = 'display_project_type_ctl'
   homeless_types = HudUtility2024.homeless_project_types
     .map { |s| HudUtility2024.project_type_brief(s) }
+  homless_codes = HudUtility2024.homeless_project_type_codes
 
 .form-group.form-group-valid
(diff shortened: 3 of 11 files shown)


app/controllers/admin/talentlms_controller.rb
```diff
@@ -9,18 +9,32 @@ class TalentlmsController < ApplicationController
...
 
     def index
-      @config = Talentlms::Config.first_or_initialize
+      @config = config_scope.first_or_initialize
+      redirect_to new_admin_talentlm_path if @config.new_record?
+      redirect_to edit_admin_talentlm_path(@config) unless @config.new_record?
+    end
+
+    def new
+      @config = config_scope.first_or_initialize
```
awisema: I think the use of `first_or_initialize` in `new` and then `create` in `create`, could be confusing... If `new` gets triggered when there is already a config, it will add another unreachable config, and then appear to lose the configuration as the first config is displayed...

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -46,11 +46,16 @@ def record_name
     end
 
     def user
+      return unless object.whodunnit
       # 'unauthenticated' matches user_for_paper_trail in ApplicationController.
       # This happens when a Job updates records, which we should display as System changes.
       return User.system_user if object.whodunnit == 'unauthenticated'
 
-      Hmis::User.find_by(id: object.whodunnit)
+      # If user was impersonating, return the true user only. If someone performed the action while impersonating, whodunnit stores as '1 as 2' where 1 is the true user.
```
awisema: I can't parse this comment, maybe rephrase?

drivers/hmis/app/models/hmis/filter/application_user_filter.rb
```diff
@@ -0,0 +1,26 @@
...
+###
+
+class Hmis::Filter::ApplicationUserFilter < Hmis::Filter::BaseFilter
+  def filter_scope(scope)
+    scope = ensure_scope(scope)
+    scope = with_search_term(scope)
+    scope
+  end
+
+  protected
```
awisema: Move modifiers to methods instead of the section format?

drivers/hmis/app/models/hmis/hud/custom_client_address.rb
awisema, line 1: Philosophical question: Should the HUD module contain non-HUD defined models?

awisema (approved):
I'm pretty sure all of the comments are from other PRs that have been pulled into this branch, but I include them anyway :-)

eanders:
> I'm pretty sure all of the comments are from other PRs that have been pulled into this branch, but I include them anyway :-)

Oh, looks like I had the wrong base branch!  Sorry, adjusted.

gigxz (approved):

PR #3750, opened 2023-11-19
186395205 hmis changelog audit trail

ttoomey:
# squash merge please

## Reference

[PT Design/implement reporting feature for HMIS record change audit
](https://www.pivotaltracker.com/story/show/186395205)

[PT Enrollment Audit History UI](https://www.pivotaltracker.com/story/show/186382841)

[PR 186382841-enrollment-audit-history 
](https://github.com/greenriver/hmis-frontend/pull/565)
## Description

* Add metadata cols to warehouse versions table
* Track HMIS model metadata; paper trail is included on each relevant class rather than the base classes, double inclusion of paper trail causes bugs
* track user / true_user in paper trail controller info
* HMIS request specs now have paper trail enabled by default to detect regressions
* Paper trail gem version upgrade (might as well since we need to QA versions)
* Include simplecov gem so we can track which mutations are missing coverage
* Adjust `enrollment.invalidate_processing!` to work around stale version error in lifecycle hook

### test coverage
Tests can now generate coverage with `COVERAGE=1 rspec ...`. This generates html, total_results, and a rollups report into /app/coverage.

### Audit for paper trail version inclusion

Moving paper trail out of the base class raises the danger that it might be missed in child classes. This can be detected with the commands below. However this only works if using `eager_load` rails config - we have this turned off in our test env so it doesn't work as an automated test.
```ruby
puts Hmis::Hud::Base.descendants.reject { |d| d.respond_to?(:versions) }.map(&:name).inspect
puts Hmis::HmisBase.descendants.reject { |d| d.new.respond_to?(:versions) }.map(&:name).inspect
```

## Testing
Changes made to enrollment and enrollment-related records should be visible in the new Enrollment Audit History (#186382841)

## Type of change
- [x] Bug fix


drivers/hmis/app/graphql/concerns/graphql_mutation_helper.rb
```diff
@@ -6,30 +6,16 @@ def self.date_string_argument(name, description, **kwargs)
     argument name, String, description, validates: { format: { with: /\d{4}-\d{2}-\d{2}/ } }, **kwargs
   end
 
-  # Default CRUD Update functionality
-  # If confirm is not specified, treat as confirmed (aka ignore warnings)
-  def default_update_record(record:, field_name:, input:, confirmed: true, **auth_args)
```
ttoomey, line 11: this method is not used afaict

app/models/concerns/hud_concerns/enrollment.rb
```diff
@@ -40,7 +40,8 @@ def self.invalidate_processing!
     end
 
     def invalidate_processing!
-      update(processed_as: nil, processed_hash: nil)
+      # use update_columns to avoid stale object errors when this method is called from Service after_commit hooks
+      update_columns(processed_as: nil, processed_hash: nil)
```
ttoomey, line 44: This was causing test failures. Flagging this for review

drivers/hmis/app/models/hmis/hud/organization.rb
```diff
@@ -9,7 +9,6 @@ class Hmis::Hud::Organization < Hmis::Hud::Base
   self.sequence_name = "public.\"#{table_name}_id_seq\""
   include ::HmisStructure::Organization
   include ::Hmis::Hud::Concerns::Shared
-  include ::Hmis::Hud::Concerns::ProjectRelated
```
ttoomey, line 12: I don't think this makes sense for an organization

drivers/hmis/app/models/hmis/hud/hmis_service.rb
```diff
@@ -4,11 +4,14 @@
...
 
+# backed by database view
 class Hmis::Hud::HmisService < Hmis::Hud::Base
   self.table_name = :hmis_services
   self.primary_key = :id
 
-  include ::Hmis::Hud::Concerns::EnrollmentRelated
+  replace_scope :viewable_by, ->(user) do
+    joins(:enrollment).merge(Hmis::Hud::Enrollment.viewable_by(user))
+  end
```
ttoomey, line 14: this is the one model that wants EnrollmentRelated but not paper trail that comes with it

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -9,6 +9,7 @@
 #   If the assessment is non-WIP: The HUD data is stored in records (IncomeBenefit, HealthAndDv, etc) that are referenced by this form_processor directly. (health_and_dv_id etc)
 class Hmis::Form::FormProcessor < ::GrdaWarehouseBase
   self.table_name = :hmis_form_processors
+  has_paper_trail
```
ttoomey, line 12: should we track meta data related to enrollment here via custom_assessment?
```ruby
    meta: { 
     client_id: ->(r) { r.custom_assessment&.paper_trail_meta_value(:client_id) },
      enrollment_id: ->(r) { r.custom_assessment&.paper_trail_meta_value(:enrollment_id) },
      project_id: ->(r) { r.custom_assessment&.paper_trail_meta_value(:project_id) },
    }
```

app/controllers/concerns/base_application_controller_behavior.rb
```diff
@@ -55,14 +55,6 @@ def append_info_to_payload(payload)
       payload[:request_start] = request.headers['HTTP_X_REQUEST_START'].try(:gsub, /\At=/, '')
     end
 
-    def info_for_paper_trail
```
ttoomey, line 58: this is set in the hmis base controller class

db/warehouse/migrate/20231203221840_idx_warehouse_versions.rb
```diff
@@ -0,0 +1,10 @@
+class IdxWarehouseVersions < ActiveRecord::Migration[6.1]
+  disable_ddl_transaction!
+
+  def change
+    # indexes for audit reports
+    add_index :versions, :client_id, algorithm: :concurrently
+    add_index :versions, :enrollment_id, algorithm: :concurrently
+    add_index :versions, :project_id, algorithm: :concurrently
```
eanders, line 8: We should make sure and watch this on staging/QA deploys.  I've seen weird behavior with concurrent index adding (as in sometimes it hangs forever, which is what it's trying to avoid.)

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -27,10 +27,7 @@ class Application::User < Types::BaseObject
     end
 
     def name
-      names = [object.first_name, object.last_name].compact_blank
-      return "User #{object.id}" unless names.any?
-
-      names.join(' ')
+      object.full_name || "User #{object.id}"
```
eanders, line 30: Not sure where `full_name` is coming from, but I assume a `User` class.  Does that return `nil` or a blank string if `full_name` is empty.  Maybe:
```suggestion
      object&.full_name.presence || "User #{object.id}"
```

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -36,13 +36,47 @@ def self.build(node_class, field_permissions: nil, transform_changes: nil)
...
+
     def record_name
-      object.item_type.demodulize.gsub(/^CustomClient/, '').underscore.humanize.titleize
+      case object.item_type
+      when 'Hmis::Hud::CustomAssessment'
+        'Assessment'
+      when 'Hmis::Hud::Assessment'
+        'CE Assessment'
+      when 'Hmis::Hud::Event'
+        'CE Event'
```
eanders, line 57: Are these names defined on the associated classes?  I feel like these are brittle (and down on line 74.)  Can we ask `Hmis::Hud::Assessment` for some sort of title, so if we change them in the future they'll align?

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -106,9 +106,44 @@ def self.static_options_for_type(pick_list_type)
...
+        [Hmis::Hud::CurrentLivingSituation],
+        [Hmis::Hud::Service],
+        [Hmis::Hud::IncomeBenefit],
+        [Hmis::Hud::HealthAndDv],
+        [Hmis::Hud::EmploymentEducation],
+        [Hmis::Hud::YouthEducationStatus],
+        [Hmis::Hud::Disability],
+        [Hmis::Hud::Exit],
+        [Hmis::Hud::Event, 'CE Event'],
+        [Hmis::Hud::Assessment, 'CE Assessment'],
```
eanders, line 139: This pops up again here.

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
```diff
@@ -189,6 +191,23 @@ def self.summary_field(name, type = nil, **kwargs)
 
     field :move_in_addresses, [HmisSchema::ClientAddress], null: false
 
+    EXCLUDED_HISTORY_FIELDS = ['id', 'DateCreated', 'DateUpdated', 'DateDeleted'].to_set.freeze
```
eanders, line 195: Should we note that other excluded fields would need to be mountain case (or match DB casing), or if that's not a requirement, maybe we want to snake case these?

drivers/hmis/app/graphql/types/application/user.rb
eanders: disregard, found implementation further down, it does a compact & presence check.

eanders (approved):

app/models/concerns/hud_concerns/enrollment.rb
gigxz, line 44: Good find, thanks!

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -36,13 +36,47 @@ def self.build(node_class, field_permissions: nil, transform_changes: nil)
...
     field :record_name, String, null: false
+    field :graphql_type, String, null: false
     field :event, HmisSchema::Enums::AuditEventType, null: false
     field :created_at, GraphQL::Types::ISO8601DateTime, null: false
     field :user, Application::User, null: true
     field :object_changes, Types::JsonObject, null: true, description: 'Format is { field: { fieldName: "GQL field name", displayName: "Human readable name", values: [old, new] } }'
 
+    available_filter_options do
+      arg :audit_event_record_type, [ID]
+      arg :user_id, [ID]
```
gigxz: For consistency, can we call this `arg :user, [ID]`? That's what we do for other ID selection, and we want it to show up in the UI as "User" not "User ID"

drivers/hmis/app/graphql/types/base_audit_event.rb
ttoomey: Yeah my hope is that this is good enough for mvp and we can take another pass later. I'm not sure that putting that responsibility on the model is exactly right (assessment.display_name) but maybe? Some considerations:
* we may want to differentiate the name based on attributes (address has logical record types of "Mailing Address", "Street Address", etc). This would have implications for filtering too
* different installations may want custom labels

gigxz: Yeah, I wonder if we could do something like this @ttoomey:
* Add a `display_name` attribute to all the classes. `Hmis::Hud::Base` could have a default that returns the demodulized class name, so we only need to override it where necessary
* Add an enum `EnrollmentRecordTypes`. The enrollment audit filter option could look like `arg :audit_event_record_type, [HmisSchema::Enums::EnrollmentRecordType]`. There would be no need for a custom PickListOption, since the frontend filter component will just use the existing Enum.
  *  The Enum values correspond to graphql type names and the descriptions have the `display_name`s from the classes? That way we can maybe also solve the `graphql_type` thing below, and we just need to maintain a mapping between class name -> graphql type name, which could also be done on the model classes.

That's a rough idea, let me know if you want to chat about options.!

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
ttoomey: the fields should match our DB casing, so they will be inconsistent. I'll add a comment

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: I don't think thats necessary.. we are tracking changes to the underlying records, we don't really need to track (or display) change to the form processor

drivers/hmis/app/graphql/types/forms/enums/pick_list_type.rb
```diff
@@ -35,5 +35,7 @@ class Forms::Enums::PickListType < Types::BaseEnum
     value 'ENROLLMENTS_FOR_CLIENT', 'Enrollments for the client, including WIP and Exited.'
     value 'OPEN_HOH_ENROLLMENTS_FOR_PROJECT', 'Open HoH enrollments at the project.'
     value 'CE_EVENTS', 'Grouped HUD CE Event types'
+    value 'AUDIT_EVENT_RECORD_TYPES'
```
gigxz: I think this should be called `ENROLLMENT_AUDIT_EVENT_RECORD_TYPES` since it's specifically records that are related to enrollments

drivers/hmis/app/graphql/types/base_audit_event.rb
gigxz: Discussed with @ttoomey, I agree we can leave it as-is until we know what pattern we need. It likely makes sense to fetch the record type list from the backend, if we want to incorporate specific custom data elements into the list for filtering.

spec/support/simple_cov_custom_formatter.rb
```diff
@@ -0,0 +1,64 @@
+# emits csv coverage report and rollups for areas of concern
+require 'csv'
+
+class SimpleCovCustomFormatter
```
gigxz, line 4: great! thanks for adding

gigxz (approved):

app/models/concerns/hud_concerns/enrollment.rb
gigxz: Thanks Theron, I opened this change separately https://github.com/greenriver/hmis-warehouse/pull/3808

PR #3752, opened 2023-11-20
More FY2024 adjustments for the HMIS DQ tool

eanders:
## Description

We seem to have missed the living situation changes and one set of EnrollmentCoC adjustments needed for FY2024 in the HMIS DQ tool

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
+++ b/drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
@@ -14,9 +14,9 @@ module HmisDataQualityTool
     include HudReports::Clients
     acts_as_paranoid
 
-    HOMELESS_LIVING_SITUATIONS = HudUtility2024::SITUATION_HOMELESS_RANGE.to_a
-    INSTITUTIONAL_LIVING_SITUATIONS = HudUtility2024::SITUATION_INSTITUTIONAL_RANGE.to_a
-    HOUSED_LIVING_SITUATIONS = HudUtility2024::SITUATION_TEMPORARY_RANGE.to_a + HudUtility2024::SITUATION_PERMANENT_RANGE.to_a
+    HOMELESS_LIVING_SITUATIONS = HudUtility2024.homeless_situations(as: :prior)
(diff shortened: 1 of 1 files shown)


drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -14,11 +14,9 @@ class Enrollment < ::HudReports::ReportClientBase
     include HudReports::Clients
     acts_as_paranoid
 
-    # These should probably live somewhere else, but for now, they live here
-    # they come from the HUD DQ report Q5
-    HOMELESS_LIVING_SITUATIONS = [16, 1, 18].freeze
-    INSTITUTIONAL_LIVING_SITUATIONS = [15, 6, 7, 25, 4, 5].freeze
-    HOUSED_LIVING_SITUATIONS = [29, 14, 2, 32, 36, 35, 28, 19, 3, 31, 33, 34, 10, 20, 21, 11, 8, 9, 99].freeze
+    HOMELESS_LIVING_SITUATIONS = HudUtility2024::SITUATION_HOMELESS_RANGE.to_a
```
gigxz: It looks like this is only used for Prior Living Situations, so it would probably be better to use `HudUtility2024.homeless_situations(as: :prior)`. That gives us the actual valid homeless situations, instead of the whole range of 100 numbers, most of which are invalid.

Though it would depend if you want to include situations that are potentially invalid but that we can still categorize -- for example if a Destination situation is stored as the answer to PLS. It's invalid technically, but we still can figure out which category it's in.

gigxz (approved):

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
eanders: Good call, I'll make that change.

eanders:
@gig, any reason you changed the target branch to `stable`?

gigxz:
You opened it against `stable`. I changed it to `release-93` but then I changed it back, in case you had opened against stable on purpose

eanders:
Ah, thanks.  I think Github must have been having issues (or my brain was having issues) I just had that happen to two PRs I would swear I opened against `release-93`.  Thanks!

PR #3698, opened 2023-11-03
185847014 annuals due filter

ZachCMP:
## Description

Include a summary of the changes and a link to the related issue. List any dependencies that are required for this change.

## Type of change
- [ ] Bug fix
- [x] New feature (adds functionality)
- [ ] Code clean-up / housekeeping
- [ ] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #3755, opened 2023-11-20
Bug fixes in annual filter

gigxz:
## Description

Build on https://github.com/greenriver/hmis-warehouse/pull/3698 to add the following logic, plus tests for each:
* Base the anniversary date off of the `earliest_entry` in the household (NOT the Entry Date of the given enrollment)
* Don't include enrollment in `Annual Due` filter results if the client entered the household before the anniversary date
* Do include enrollment in `Annual Due` filter results if they don't have an annual assessment during the "due period", EVEN IF they have one later. For example if the annual was due in Jan-Feb, but they have an annual dated for March, then they should be included in the filter results. I think this is the desired behavior because this tool will be used to help identify enrollments where the annual is missing for data quality purposes.
* Use a full composite join (EnrollmentID + AssessmentID + data source id) when joining to assessments

## Type of change
- [x] Bug fix


drivers/hmis/app/models/hmis/filter/enrollment_filter.rb
```diff
@@ -62,49 +62,49 @@ def with_household_tasks(scope)
...
-          # This enrollment has not had any annual assessments
-          cas_t[:assessment_date].eq(nil).
-          # OR an annual assessment is now due for this year and hasn't been done yet
-          or(start_date.lteq(Date.today).and(cas_t[:assessment_date].lt(start_date))).
-          # OR an annual assessment is not due yet for this year but one was not done for last year
-          or(start_date.gteq(Date.today).and(cas_t[:assessment_date].lt(last_start_date))),
-        ).
-        distinct
+        # Client entered household before this years anniversary
+        where(e_t[:entry_date].lt(Arel.sql(anniversary_date))).
```
gigxz: this needs a conditional and another test case - if anniversary is in the future of the current year, then we actually want to check if the client entered before _last years_ anniversary

drivers/hmis/app/models/hmis/filter/enrollment_filter.rb
```diff
@@ -55,56 +55,63 @@ def with_household_tasks(scope)
...
       # The SQL statement "interval"((extract(day from "EntryDate")::integer - 1) || ' days') is present to handle leap
       # year dates. That is, if the entry date is 2/29/2020, then just using make_date directly will produce an
       # anniversary date of 2/29/2023, which is invalid. To prevent this, make_date is used to create a date of
       # 2/1/2023, then we add 28 days (29th - 1) to that date to shift it to the proper anniversary date. For any date
       # except for 2/29, this shift should result in the same month and day as the entry date, and for 2/29 it should
       # shift it to 3/1, which is a valid date and avoids the leap year issue.
 
-      # Start of the 60-day window during which this year's Annual should be performed. (Anniversary - 30 days)
-      start_date = Arel.sql <<~SQL
+      anniversary_date = "
```
eanders: @gigxz any reason we switched away from a heredoc?  They tend to be more resilient for SQL since they allow both kinds of quotes without escaping.

eanders (approved):
I think this looks good.  Thanks!

drivers/hmis/app/models/hmis/filter/enrollment_filter.rb
gigxz: I couldn't get string interpolation to work with that, is it possible? I did switch back `anniversary_date` var to use this style though

eanders: Hmm, not sure about interpolation, I haven't noticed an issue before. Looks like at some point in history it was possible. https://medium.com/rubycademy/avoid-interpolation-in-heredoc-59a5d907133d

gigxz:
This is going into 93 so we can push out this filter with the bug fixes. I'll update 94 when merged too

drivers/hmis/app/models/hmis/filter/enrollment_filter.rb
eanders: This has some more details. https://blog.saeloun.com/2020/04/08/heredoc-in-ruby-and-rails/

gigxz: oh I got it working now, thanks!

eanders:
@gigxz feel free to merge once tests pass.

PR #3758, opened 2023-11-21
186522571-colocated-warehouse-versions

ttoomey:
# Please squash merge

## Description

[Colocated versions table for warehouse records
](https://www.pivotaltracker.com/story/show/186522571)

Adds new warehouse-specific paper trail version table modeld with `GrdaWarehouse::Version`. It uses the paper trail mix-in and inherits from `GrdaWarehouseBase` to get the appropriate db. A difference from the old versions table is that it stores objects as jsonb.

Note, this is a different approach from `Health::HealthVersion` which uses `establish_connection` within the class. I'm not sure that approach is supported by Rails and may break in the future.

### Data migration
Changes made to hmis client records can be imported via:
```
rails driver:hmis:migrate_versions_20231121
```

## Type of change
- [x] Bug fix


app/models/grda_warehouse_base.rb
```diff
@@ -31,4 +31,11 @@ def self.needs_migration?
   def self.partitioned?(table_name)
     Dba::PartitionMaker.new(table_name: table_name).done?
   end
+
+  # default colocated versions table for warehouse records
+  def self.has_paper_trail(options = {}) # rubocop:disable Naming/PredicateName
+    versions = options.fetch(:versions, {}).merge(class_name: 'GrdaWarehouse::Version')
+    skip = options.fetch(:skip, [:lock_version])
```
ttoomey, line 38: To reduce bloat, we could skip tracking DateUpdated here too. But maybe it's safer to track it

db/warehouse/migrate/20231120221840_colocated_warehouse_versions.rb
```diff
@@ -0,0 +1,17 @@
...
+      t.string   :item_type, null: false
+      t.bigint   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.jsonb    :object
+      t.jsonb    :object_changes
+      t.datetime :created_at
+      t.string :session_id
+      t.string :request_id
+      t.references :user, index: false
```
ttoomey, line 13: We could add an index but this table will probably be huge. Not sure if it's worth the cost.

gigxz: I don't think it's worth it if it makes insertion slower. its OK if the audit history views take a bit longer to load.

app/models/grda_warehouse_base.rb
```diff
@@ -31,4 +31,11 @@ def self.needs_migration?
   def self.partitioned?(table_name)
     Dba::PartitionMaker.new(table_name: table_name).done?
   end
+
+  # default colocated versions table for warehouse records
+  def self.has_paper_trail(options = {}) # rubocop:disable Naming/PredicateName
+    versions = options.fetch(:versions, {}).merge(class_name: 'GrdaWarehouse::Version')
```
gigxz, line 37: Does this let you override which versions table is used? We may want to do that for [GrdaWarehouse::UserViewableEntity](https://github.com/greenriver/hmis-warehouse/blob/0d11a228fcb9926a63f8e2e1987d1a6d9ccc67a1/app/models/grda_warehouse/user_viewable_entity.rb#L9-L14) to keep its versions in the app db, since it needs those metadata columns and references both dbs. But @eanders can confirm..

app/models/grda_warehouse/version.rb
```diff
@@ -0,0 +1,11 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module GrdaWarehouse
+  class Version < GrdaWarehouseBase
+    include PaperTrail::VersionConcern
```
gigxz, line 9: This makes sense to me, just noting that it uses a different pattern than the health db https://github.com/greenriver/hmis-warehouse/blob/0d11a228fcb9926a63f8e2e1987d1a6d9ccc67a1/app/models/health/health_version.rb#L9-L10 which inherits from a shared `GrPaperTrail::Version` class to similar effect

db/warehouse/migrate/20231120221840_colocated_warehouse_versions.rb
```diff
@@ -0,0 +1,17 @@
+class ColocatedWarehouseVersions < ActiveRecord::Migration[6.1]
+  def change
+    create_table :versions do |t|
+      t.string   :item_type, null: false
+      t.bigint   :item_id,   null: false
+      t.string   :event,     null: false
+      t.string   :whodunnit
+      t.jsonb    :object
+      t.jsonb    :object_changes
```
gigxz, line 9: Yay for JSON!

The client audit history feature currently calls [client.versions](https://github.com/greenriver/hmis-warehouse/blob/0d11a228fcb9926a63f8e2e1987d1a6d9ccc67a1/drivers/hmis/app/graphql/types/hmis_schema/has_audit_history.rb#L33) to get the version history, and then [expects object_changes to be YAML](https://github.com/greenriver/hmis-warehouse/blob/0d11a228fcb9926a63f8e2e1987d1a6d9ccc67a1/drivers/hmis/app/graphql/types/base_audit_event.rb#L64). So I think this would break that feature.

If that's right, we need to update the code to expect the changes to be JSON. Hopefully that is simple.

How difficult do you think it would be to migrate over the version records just for this scope `GrPaperTrail::Version.where(item_type: 'Hmis::Hud::Client')`? Client audit history is the only thing that's currently visible in the HMIS application, and I think we want to keep that history from the past couple months - but if it's a huge lift, we can discuss options, I'd just want to talk to customers first. It's also something we could migrate later on -- including maybe for other record types - if we need to.

app/models/grda_warehouse_base.rb
ttoomey: Good catch, I think that since user_viewable_entities lives in the warehouse database, we should use the warehouse versions table to track it. I can add referenced_user_id and referenced_entity_name to the warehouse versions table as well

app/models/grda_warehouse/version.rb
ttoomey: Yeah I suspect calling establish_connection like that isn't supported or at least it's not well documented. So probably better to avoid it if possible

db/warehouse/migrate/20231120221840_colocated_warehouse_versions.rb
ttoomey: > If that's right, we need to update the code to expect the changes to be JSON. Hopefully that is simple.

Good catch! Yeah it doesn't look hard, I can do that here.

> How difficult do you think it would be to migrate over the version records just for this scope GrPaperTrail::Version.where(item_type: 'Hmis::Hud::Client')?

It shouldn't be a big lift, I think I can add a rake task for that in this PR

ttoomey: Added a one-time rake task for this:
```sh
rails driver:hmis:migrate_versions_20231121
```

app/models/grda_warehouse.rb
```diff
@@ -8,4 +8,8 @@ module GrdaWarehouse
   def self.table_name_prefix
     ''
   end
+
+  def self.paper_trail_versions
+    GrdaWarehouse::Version.all
+  end
```
ttoomey, line 14: I think it is a little nicer to use a module method rather than scattering the class name around.

app/models/grda_warehouse_base.rb
eanders: Hmm, yeah, I agree with @ttoomey we should keep things in the DB the table exists in.  This change will require updating the UI for auditing users, and maybe clients.

app/models/grda_warehouse_base.rb
eanders: I say we leave it.

db/warehouse/migrate/20231120221840_colocated_warehouse_versions.rb
eanders: Does it make sense to migrate the existing app table from yaml to json?  I'm guessing that'd be _hours_ of processing, so maybe not, but having them in two different formats seems like a future recipe for confusion.

app/models/grda_warehouse_base.rb
gigxz: @eanders I was looking for potential places this would break the warehouse views. As far as I can tell, there are 2 places we show "audit history":
* Client dashboard "audit" tab (Access History) uses [Admin::AuditsController](https://github.com/greenriver/hmis-warehouse/blob/release-94/app/controllers/admin/audits_controller.rb) which is based on ActivityLog, so it should not be affected. The client audit just shows who has viewed the client, it does not show changes to the client record. Can you confirm?
* User "Edit History" uses [Admin::EditHistoriesController](https://github.com/greenriver/hmis-warehouse/blob/release-94/app/controllers/admin/edit_histories_controller.rb), which I think should correctly be based off of the App DB since that's where Users and access control-related things live. But if UserViewableEntity versions are in the other DB, sounds like it would complicate that.
* User "Audit Trail" is similarly based on ActivityLog only

gigxz: For reference @ttoomey 
https://qa-hmis-warehouse.openpath.host/clients/==0g5mRw2yHRr7rkV7q1mKCw==/audits
https://qa-hmis-warehouse.openpath.host/admin/users/==kM3z30Nc2+STWyq9o+9f0g==/edit_history

eanders: @gigxz , that matches my understanding, I'd forgotten about the client side using activity log (I think that's what the user view history uses too.  This should be fine for the warehouse, the number of places someone can change a client record are very small.) Looks like the [EditHistoriesController](https://github.com/greenriver/hmis-warehouse/blob/release-94/app/controllers/admin/edit_histories_controller.rb) would need an update to somehow union the changes from both versions tables if we move the user viewable entities.  That said, at least in one installation, it looks like we changed something so that these haven't seen a change since 2019.

gigxz: Ah yeah, I can't find any meaningful usages of `GrdaWarehouse::UserViewableEntity` in the code base.. just a rake task that was written 4 years ago. The user model doesn't have a relationship to it at all. So, that seems safe to ignore, and maybe we should remove it entirely.

eanders: Hmm, maybe, pretty sure those transitioned to `GroupViewableEntity` which may have a similar issue.

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
```diff
@@ -0,0 +1,39 @@
...
+
+  def transform_record(version)
+    result = version.attributes
+    # track the previous version
+    result['migrated_app_version_id'] = result.delete('id')
+    ['object', 'object_changes'].each do |field|
+      doc = result.delete(field)
+      result[field] = parse_yaml(doc) if doc.present?
+    end
+    result.compact_blank
```
gigxz: I needed the following change to get this working locally. I was getting this error on a `Hmis::Hud::Client` record:
```
ArgumentError: Hash key mismatch.

When importing an array of hashes, all hashes must have the same keys.
If you have records that are missing some values, we recommend you either set default values
for the missing keys or group these records into batches by key set before importing.

Required keys: ["item_type", "item_id", "event", "whodunnit", "created_at", "request_id", "migrated_app_version_id", "object_changes"]
Extra keys: ["object"]
Missing keys: []
```

```suggestion
    result['migrated_app_version_id'] = result.delete('id')
    ['object', 'object_changes'].each do |field|
      doc = result.delete(field)
      result[field] = doc.present? ? parse_yaml(doc) : nil
    end
    result
```

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
```diff
@@ -0,0 +1,39 @@
+# frozen_string_literal: true
+
+desc 'One time data migration from app.versions to warehouse.versions'
+# rails driver:hmis:migrate_versions_20231121
+task migrate_versions_20231121: [:environment] do
+  import_scope = GrPaperTrail::Version.where(item_type: 'Hmis::Hud::Client')
```
gigxz:
```suggestion
  hmis_client_record_types = [
    'Hmis::Hud::Client',
    'Hmis::Hud::CustomClientName',
    'Hmis::Hud::CustomClientAddress',
    'Hmis::Hud::CustomClientContactPoint',
  ]
  import_scope = GrPaperTrail::Version.where(item_type: hmis_client_record_types)
```

I forgot about these, we want to bring them over too. I made this change and was able to run it locally (with the below modification)

gigxz (commented):
Looks good to me! A couple notes on the migration, otherwise this looks good to me. Thanks for adding that in.

Confirmed locally:
- Able to run migration task and view client audit histories in HMIS (with one modification, see comment)
- Able to view user edit history in Warehouse
- Able to view client access history in Warehouse

app/models/grda_warehouse_base.rb
gigxz: We don't have PaperTrail enabled on GroupViewableEntity 🧐 
Maybe a consideration for [this ticket](https://www.pivotaltracker.com/story/show/186497193).

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
ttoomey: Good catch!

db/warehouse/migrate/20231120221840_colocated_warehouse_versions.rb
ttoomey: Yeah that seems like a good idea. I assume there's millions of rows in the existing tables so we probably need a migration strategy that's a bit more involved, recommend we make another issue for that. 

We may want to benchmark conversion so we can ball-park estimate how long the migration will take for a given db (hours, days, weeks?)

The paper trail gem suggests something like the following. Does this seem reasonable?
* rename existing object cols out of the way (object -> object_yml) and add new json-typed cols for object and object_changes
* paper trail should start using the new json cols going forward but will temporarily lose access to old object data
* Add a resumable task that pulls a batch of versions and migrates them to yaml. Scope would be something like `where object_yml IS NOT NULL AND object IS NULL`. This could be parallelized if needed
* Could also take the opportunity to fix other fields (whodunnit)
* Finally remove the object_yml and object_changes_yml once all rows are migrated

gigxz (approved):

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+desc 'One time data migration from app.versions to warehouse.versions'
+# rails driver:hmis:migrate_versions_20231121
+task migrate_versions_20231121: [:environment] do
+  hmis_client_record_types = [
+    'Hmis::Hud::Client',
+    'Hmis::Hud::CustomClientName',
+    'Hmis::Hud::CustomClientAddress',
+    'Hmis::Hud::CustomClientContactPoint',
```
ttoomey, line 10: should be hmis::hud*

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
```diff
@@ -0,0 +1,45 @@
...
+  ]
```
eanders: This should be `Hmis::Hud::*`

ttoomey:
Just double-checking, this is okay to merge to release-94?

gigxz:
I just tried running the updated task locally, just to see. I get the error `Psych::DisallowedClass: Tried to load unspecified class: BigDecimal` - any thoughts on what that might be from? I can do some spelunking to see which item type might be triggering it

```
36d119329aed:/app# rails driver:hmis:migrate_versions_20231121
FileMagic v0.7.3: compiled magic version [5.43] does not match with shared library magic version [5.44]
  GrdaWarehouse::Version Destroy (7.5ms)  DELETE FROM "versions" WHERE "versions"."migrated_app_version_id" IS NOT NULL /*application:BostonHmis*/
Query Trace:
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:160:in `invoke_task'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `block (2 levels) in top_level'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `each'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `block in top_level'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:125:in `run_with_threads'
  ↳ /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:160:in `invoke_task'
  GrPaperTrail::Version Load (9.7ms)  SELECT "versions".* FROM "versions" WHERE (item_type like 'Hmis::Hud::%') ORDER BY "versions"."id" ASC LIMIT $1 /*application:BostonHmis*/  [["LIMIT", 1000]]
  ↳ /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:160:in `invoke_task'
Query Trace:
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:160:in `invoke_task'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `block (2 levels) in top_level'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `each'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:116:in `block in top_level'
      /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:125:in `run_with_threads'
  ↳ /bundle/ruby/3.1.0/gems/rake-13.0.6/lib/rake/application.rb:160:in `invoke_task'
rails aborted!
Psych::DisallowedClass: Tried to load unspecified class: BigDecimal

Tasks: TOP => driver:hmis:migrate_versions_20231121
(See full trace by running task with --trace)
```

gigxz:
I think it's when transforming an IncomeBenefit record

eanders:
@gigxz there's a line in there that allows Date, DateTime, and something else, sounds like we need BigDecimal as well.

drivers/hmis/lib/tasks/one_time_migration_20231121.rake
```diff
@@ -0,0 +1,39 @@
...
+      doc = result.delete(field)
+      result[field] = doc.present? ? parse_yaml(doc) : nil
+    end
+    result
+  end
+
+  def parse_yaml(doc)
+    YAML.load(
+      doc,
+      permitted_classes: [Time, Date, Symbol],
```
eanders: @gigxz 
```suggestion
      permitted_classes: [Time, Date, Symbol, BigDecimal],
```
There will probably be others as well.

gigxz: I got successfully through my local versions with just this change. thanks

eanders: Maybe we should just use `unsafe_load`.  This should be a pretty safe location to load from.

ttoomey: Thanks, sorry I missed that! I added BigDecimal. Using `unsafe_load` is probably fine but my inclination would to be extra careful and stick with the allow-list. If we discover new types of entities that need to be included, we could always adjust the task on the fly and re-run.

eanders: works for me.

eanders:
@ttoomey I'm good with you merging into `release-94` whenever @gigxz is.

gigxz:
yeah go for it!

PR #3760, opened 2023-11-21
Add posting export to DW Upload

gigxz:
## Description

Add a new requested export for linking Posting IDs to Enrollment IDs. https://www.pivotaltracker.com/n/projects/2591838/stories/186506330

## Type of change

- [] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
@@ -1,0 +1,67 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisExternalApis::AcHmis::Exporters
+  class PostingExport
+    attr_accessor :output
+
+    def initialize(output = StringIO.new)
+      require 'csv'
+      self.output = output
+    end
+
+    def run!
+      Rails.logger.info 'Generating content of posting export'
+
+      write_row(columns)
+      total = postings.count
+
+      Rails.logger.error "There are #{total} postings to export. That doesn't look right" if total < 10
+
+      postings.find_each.with_index do |posting, i|
+        Rails.logger.info "Processed #{i} of #{total}" if (i % 1000).zero?
+
+        enrollment = posting.hoh_enrollment
+
+        if enrollment.nil?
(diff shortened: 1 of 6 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
```diff
@@ -0,0 +1,67 @@
...
+          warehouse_id, # PersonalID matching HMIS CSV export
+          enrollment.id, # EnrollmentID matching HMIS CSV export
+          posting.identifier, # EntityPostingID
+          posting.created_at.strftime('%Y-%m-%d %H:%M:%S'), # AssignedDate
+        ]
+        write_row(values)
+      end
+    end
+
+    private
```
awisema, line 47: I have some recollection that we decided to put a `private` on each `def` rather than having a `private` section. For me, this is because the structure of Ruby files make is too easy to accidentally set the visibility of a method, and then not know about it until invocation.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
```diff
@@ -0,0 +1,67 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisExternalApis::AcHmis::Exporters
+  class PostingExport
+    attr_accessor :output
+
+    def initialize(output = StringIO.new)
+      require 'csv'
```
awisema, line 12: What is the advantage of putting the `require` here vs at the top of the file?

awisema (approved):

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
gigxz: I'm not sure, this is a copy of a job that Todd created originally. I'd guess it lets you run the job in the rails console without doing an additional import

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/posting_export.rb
gigxz: Oh really, thats good to know. I like that better. This copies the pattern from all the other AC DW export jobs.

PR #3761, opened 2023-11-22
Release 93 -> 94

ttoomey:
# Please do _NOT_ squash-merge

Looks like a few PRs ended up getting merged in release-93 instead of release-94

* Fix incorrect form logic for HousingType 
* hmis system e2e tests


eanders:
@ttoomey , I just re-targeted `release-94` and updated.  I can't do that initially since they are identical, but now that somethings been merged in, the PR will stay open.

PR #3767, opened 2023-11-23
186395195-stale-object-error-when-submitting-household-assessments

ttoomey:
## Description
Depends on https://github.com/greenriver/hmis-frontend/pull/558

Intake assessment test to assist with [tracking down this issue](https://www.pivotaltracker.com/story/show/186395195). 

* Also support disabling animations to speed up e2e tests and make them more reliable (hopefully). 
* Changed SystemSpecHelper module to be a context which I think is more standard
* Note there are fixes in the HMIS branch (186395195-stale-object-error-when-submitting-household-assessments) that address timing issues that are needed to make this test pass

#

## Type of change


bin/run_hmis_system_tests.sh
```diff
@@ -31,7 +31,7 @@ fi
 cleanup() {
   echo "Removing temporary directory ${TEMP_DIR}"
   # comment this out to preserve the directory on exit
-  # rm -rf "$TEMP_DIR"
+  rm -rf "$TEMP_DIR"
```
ttoomey, line 34: I had accidentally commented this out

spec/support/e2e_setup.rb
```diff
@@ -26,10 +29,20 @@ def set_hidden_field_value(id, value)
...
+    end
+  end
+
   def mui_select(choice, from:)
     label = find('label', text: from)
     id = label['for']
-    find("##{id}").click
+    # we seem to have invalid ids such as "3.917A.1"
+    # find("##{id}").click
+    find("[id='#{id}']").click
```
gigxz, line 47: Oh I see, I didn't realize they were invalid. IDs must start with A-Za-z, that's the only issue right? The ID gets set to the unique `link_id` in the form definition. Maybe we should do something like `LINK_ID_${linkId}` instead. (Not requesting a change, just pointing out)

gigxz (approved):

spec/support/e2e_setup.rb
ttoomey: That's right, I think they just need to start with a letter (you can also escape them too). To clarify, the ids are valid in HTML, they just can't be used in CSS selectors

PR #3769, opened 2023-11-23
New homeless days calculation

eanders:
## Description

Include self-reported days prior to entry in homeless days calculation.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


app/models/grda_warehouse/cas_project_client_calculator/mdha.rb
```diff
@@ -109,5 +111,17 @@ def value_for_cas_project_client(client:, column:)
     private def assessor_email(client)
       client.source_eccovia_assessments.max_by(&:assessed_at)&.assessor_email
     end
+
+    private def days_homeless_plus_prior(client)
+      homeless_days = client.days_homeless
+      range = Date.yesterday .. Date.current
+      ongoing_enrollments = client.source_enrollments.select { |en| en.open_during_range?(range) }
```
awisema: `Enrollment` has a scope `ongoing` -- but it doesn't include `yesterday` in the range... I think this should either use it, or include a comment explaining the reason it is different.

awisema (approved):

app/models/grda_warehouse/cas_project_client_calculator/mdha.rb
eanders: Ah, yeah, I'll note we're not using it because we preloaded `source_enrollments` and adding a scope would add an N+1.

PR #3770, opened 2023-11-24
LSA preflight updates

eanders:
## Description

This:
* Removes the null restriction on Enrollment.EnrollmentCoC
* Adds additional pre-kick-off checks for HMISParticipation.HMISParticipationType and Inventory.HouseholdType
* Adds a pre-flight check when running the LSA to fail quickly if there are known structural issues with any projects in the LSA
* Adds an error_details column to HUD report instances
* Populates the error_details and exposes them on the report history page if the pre-flight check fails

## Type of change

### DIFF
--- a/db/warehouse_structure.sql
+++ b/db/warehouse_structure.sql
@@ -19861,7 +19861,8 @@ CREATE TABLE public.hud_report_instances (
     failed_at timestamp without time zone,
     percent_complete double precision,
     export_id bigint,
-    type character varying DEFAULT 'HudReports::ReportInstance'::character varying NOT NULL
+    type character varying DEFAULT 'HudReports::ReportInstance'::character varying NOT NULL,
+    error_details text
 );
 
 
@@ -57760,7 +57761,7 @@ CREATE UNIQUE INDEX tx_id_ds_id_ft_idx ON public.financial_transactions USING bt
 -- Name: uidx_external_id_ns_value; Type: INDEX; Schema: public; Owner: -
 --
 
-CREATE UNIQUE INDEX uidx_external_id_ns_value ON public.external_ids USING btree (source_type, namespace, value) WHERE ((namespace)::text <> ALL ((ARRAY['ac_hmis_mci'::character varying, 'ac_hmis_mci_unique_id'::character varying])::text[]));
+CREATE UNIQUE INDEX uidx_external_id_ns_value ON public.external_ids USING btree (source_type, namespace, value) WHERE ((namespace)::text <> ALL (ARRAY[('ac_hmis_mci'::character varying)::text, ('ac_hmis_mci_unique_id'::character varying)::text]));
 
 
 --
@@ -59870,6 +59871,7 @@ INSERT INTO "schema_migrations" (version) VALUES
 ('20231107190301'),
 ('20231110134113'),
 ('20231114235529'),
-('20231115170459');
+('20231115170459'),
+('20231124171521');
 
 
--- a/drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
+++ b/drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
@@ -123,7 +123,9 @@ module HudLsa::Generators::Fy2023
 
       # Prevent report.complete_report from hiding the error
       @failed = true
-      project_names = issues.values.flatten.select { |r| r[:id].in?(issue_project_ids) }.map { |r| r[:project] }.uniq
+      project_names = issues.values.flatten.select { |r| r[:id].in?(issue_project_ids) }.map do |r|
+        "#{r[:project]} (id: #{r[:id]})"
+      end.uniq
       failure = "The following projects are missing data that will cause the LSA to fail: #{project_names.join(', ')}"
       fail_report(failure)
       false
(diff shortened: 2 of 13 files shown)


db/warehouse_structure.sql
```diff
@@ -57760,7 +57761,7 @@ CREATE UNIQUE INDEX tx_id_ds_id_ft_idx ON public.financial_transactions USING bt
 -- Name: uidx_external_id_ns_value; Type: INDEX; Schema: public; Owner: -
 --
 
-CREATE UNIQUE INDEX uidx_external_id_ns_value ON public.external_ids USING btree (source_type, namespace, value) WHERE ((namespace)::text <> ALL ((ARRAY['ac_hmis_mci'::character varying, 'ac_hmis_mci_unique_id'::character varying])::text[]));
+CREATE UNIQUE INDEX uidx_external_id_ns_value ON public.external_ids USING btree (source_type, namespace, value) WHERE ((namespace)::text <> ALL (ARRAY[('ac_hmis_mci'::character varying)::text, ('ac_hmis_mci_unique_id'::character varying)::text]));
```
eanders, line 57764: @ttoomey or @gigxz do I need to re-run a migration?  This seems subtly different (but maybe just a difference in PG versions?)

ttoomey: My guess is this is due to a difference in pg versions. It looks like the first line casts the whole array to text[] where as the second line casts each member to text individually. That seems equivalent, right?

Managing our schemas across multiple pg versions is a minor friction point

eanders: Thanks.

drivers/hud_lsa/app/views/hud_lsa/lsas/_missing_data.haml
```diff
@@ -16,7 +16,10 @@
             %th Zipcode
             %th Inventory Start Date
             %th Inventory CoC Code
+            %th Inventory Houshold Type
```
gigxz:
```suggestion
            %th Inventory Household Type
```

gigxz (approved):
Looks good!

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/status_progression_concern.rb
```diff
@@ -28,4 +28,14 @@ def finish_report
       remaining_questions: [],
     )
   end
+
+  def fail_report(reason = nil)
+    self.update( # rubocop:disable Style/RedundantSelf:
```
ttoomey, line 33: safer to use `update!` instead?

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
```diff
@@ -45,6 +45,8 @@ def lookback_stop_date
 
     def run!
       setup_notifier('LSA')
+      return unless preflight_passes?
```
ttoomey: This is fine, suggest setting `@failed` within the `preflight_passes?` method so it isn't a side-effect
```
@failed = !preflight_passes?
```

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/missing_data_concern.rb
```diff
@@ -143,6 +146,17 @@ def missing_data(user)
     )
   end
 
+  private def missing_inventory_household_types(user)
+    missing_data_rows(
+      GrdaWarehouse::Hud::Inventory.joins(project: :organization).
+      includes(project: :funders).
```
ttoomey: For clarity, maybe better to use `joins()` or `preload()` explicitly rather than `includes()`?

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
```diff
@@ -112,6 +114,23 @@ def calculate
       finish_report
     end
 
+    # Confirm the chosen projects are not missing critical data.
+    # Any structural failures that will cause the run to fail should be caught here
+    private def preflight_passes?
+      issues = missing_data(user).except(:show_missing_data)
+      issue_project_ids = issues.values.flatten.map { |r| r[:id] }.uniq & filter.effective_project_ids
+      return true unless issue_project_ids.any?
```
ttoomey: maybe clearer
```suggestion
      return true if issue_project_ids.empty?
```

ttoomey (approved):
Code looks good! Left a few suggestions

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
eanders: Thanks!

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/lsa.rb
eanders: Much

drivers/hud_lsa/app/models/hud_lsa/generators/fy2023/missing_data_concern.rb
eanders: Hmm, I can't remember the reasoning behind not just throwing the `:funders` into the joins above.  Maybe left-outer joins?  It doesn't look like we use it explicitly, but I think we're using it to limit something. That said, it could totally be a copy/paste issue.  I'll investigate further, thanks for noticing.

PR #3772, opened 2023-11-26
More careplan status fixes

awisema:
## Description

Patient dashboards and careplan displays were not considering all careplans when showing status.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/models/health/pctp_careplan.rb
+++ b/app/models/health/pctp_careplan.rb
@@ -97,5 +97,30 @@ module Health
       where(instrument_id: v1_ids, instrument_type: 'Health::Careplan').
         or(where(instrument_id: v2_ids, instrument_type: 'HealthPctp::Careplan'))
     end
+
+    def expires_on
+      return nil unless instrument.careplan_sent_on.present?
+
+      instrument.careplan_sent_on + 12.months
+    end
+
+    def active?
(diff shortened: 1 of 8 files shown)


app/models/health/pctp_careplan.rb
```diff
@@ -97,5 +97,30 @@ class PctpCareplan < HealthBase
...
+      return false unless instrument.careplan_sent_on.present?
+      return false if expired?
+
+      expires_on >= Date.current
+    end
+
+    def expiring?
+      return false unless expires_on.present?
+
+      expires_on > Date.current
```
eanders: Should there be any date range on this? If I'm reading this correctly it is expiring 12 months after it was sent, but expired... Actually, can you add some comments (and maybe tests for these?)  The logic isn't super clear.

awisema: I reworked these and added some documentation -- clearer now?

eanders: Much thank you.  I still think some tests are in order.

awisema:
Simple tests added

eanders (approved):
Thank you!  Those tests, while simple will really help us keep track of the expected logic (and when we cause accidental regressions.)

eanders (approved):

PR #3775, opened 2023-11-27
185234956 auto exit job

ZachCMP:
## Description

## Type of change
- [ ] Bug fix
- [ ] Dependency update


ZachCMP:
@gigxz This should be ready now. The only outstanding question I have is whether there's some kind of translation utility I should be using to get the integer values for the HUD integer enums (i.e. project type, type provided, data collection stage, etc.). Happy to add that if there's something I should be using there.

db/warehouse/migrate/20231122185546_add_hmis_auto_exit_config.rb
```diff
@@ -0,0 +1,13 @@
+class AddHmisAutoExitConfig < ActiveRecord::Migration[6.1]
+  def change
+    add_column :Exit, :auto_exited, :datetime, null: true
+
+    create_table(:hmis_auto_exit_configs) do |t|
+      t.integer :length_of_absence_days, null: false, default: 30
+      t.integer :project_type
+      t.integer :organization_id
+      t.integer :project_id
```
gigxz: It looks like we're pretty inconsistent about this, but I think references should be bigint columns. Or `t.references` which also adds a foreign key constraint

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Hmis
+  class AutoExitJob < BaseJob
+    def perform
+      Hmis::Hud::Project.all.each do |project|
```
gigxz:
```suggestion
      Hmis::Hud::Project.hmis.each do |project|
```
This will limit scope to only projects in HMIS data sources

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+        config = Hmis::AutoExitConfig.config_for_project(project)
+        next unless config.present?
+
+        project.enrollments.each do |enrollment|
```
gigxz: We can skip exited enrollments and WIP enrolments
```suggestion
        project.enrollments.open_excluding_wip.each do |enrollment|
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+
+    def configs
+      @configs ||= Hmis::AutoExitConfig.all
+    end
+
+    private
+
+    def auto_exit(enrollment, most_recent_contact)
+      exit_date = contact_date_for_entity(most_recent_contact)
+      exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
```
gigxz, line 41:
```suggestion
      # If most recent contact was a Bed Night service, the Exit Date should be the day after they received service
      exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+
+module Hmis
+  class AutoExitJob < BaseJob
+    def perform
+      Hmis::Hud::Project.all.each do |project|
+        config = Hmis::AutoExitConfig.config_for_project(project)
+        next unless config.present?
+
+        project.enrollments.each do |enrollment|
+          most_recent_contact = if HudUtility2024.residential_project_type_ids.include?(project.project_type)
```
gigxz: Bed Night should only be used as the most recent contact if this is a Night-by-night shelter. Other project types don't record bed-nights.
```suggestion
          most_recent_contact = if project.project_type == 1 # Night-by-night Emergency Shelter
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+            enrollment.services.where(record_type: 200).order(:date_provided).last
```
gigxz:
```suggestion
            enrollment.services.bed_nights.order(:date_provided).last
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+          else
+            [
+              enrollment.services.order(:date_provided).last,
+              enrollment.custom_services.order(:date_provided).last,
+              enrollment.current_living_situations.order(:information_date).last,
+              enrollment.custom_assessments.order(:assessment_date).last,
+            ].compact.max_by { |entity| contact_date_for_entity(entity) }
+          end
+
+          next unless most_recent_contact.present?
```
gigxz: This shouldn't really happen, because every enrollment should at least have an intake assessment. If it does, though, let's use the `enrollment.entry_date` as the `most_recent_contact_date` instead. That way, if someone enters and doesn't have _anything_ recorded for 3 months (or whatever), they would still get auto-exited after 3 months.

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+
+          most_recent_contact_date = contact_date_for_entity(most_recent_contact)
+          next unless (Date.today - most_recent_contact_date).to_i >= config.length_of_absence_days
```
gigxz: Date.current is better for comparisons https://thoughtbot.com/blog/its-about-time-zones#a-summary-of-do39s-and-don39ts-with-time-zones
```suggestion
          next unless (Date.current - most_recent_contact_date).to_i >= config.length_of_absence_days
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Hmis
+  class AutoExitJob < BaseJob
+    def perform
+      Hmis::Hud::Project.all.each do |project|
+        config = Hmis::AutoExitConfig.config_for_project(project)
+        next unless config.present?
+
```
gigxz, line 14: Can we add in a safeguard here, like `raise 'Auto-exit config unusually low: {days}' if config. length_of_absence_days < 30`?  I worry about some config getting created with something like 0 days, which would cause problems and shouldn't happen.

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+          most_recent_contact_date = contact_date_for_entity(most_recent_contact)
+          next unless (Date.today - most_recent_contact_date).to_i >= config.length_of_absence_days
+
+          auto_exit(enrollment, most_recent_contact)
+        end
+      end
+    end
+
+    def configs
+      @configs ||= Hmis::AutoExitConfig.all
```
gigxz: This isn't referenced anywhere, is it?

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+      assessment = Hmis::Hud::CustomAssessment.new_with_defaults(
+        enrollment: enrollment,
+        user: Hmis::Hud::User.system_user(data_source_id: enrollment.data_source_id),
+        form_definition: Hmis::Form::Definition.find_definition_for_role(:EXIT, project: enrollment.project),
+      )
+      assessment.data_collection_stage = 3
+      assessment.assessment_date = exit_date
+      assessment.form_processor.assign_attributes(
+        values: {},
+        hud_values: {},
```
gigxz, line 65: This also needs to set `exit_id: exit.id` so that the assessment is properly linked to the Exit record

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+    def auto_exit(enrollment, most_recent_contact)
+      exit_date = contact_date_for_entity(most_recent_contact)
+      exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
+      Hmis::Hud::Exit.create!(
+        personal_id: enrollment.personal_id,
+        enrollment_id: enrollment.enrollment_id,
+        data_source_id: enrollment.data_source_id,
+        user_id: Hmis::Hud::User.system_user(data_source_id: enrollment.data_source_id).user_id,
+        exit_date: exit_date,
+        destination: 30,
```
gigxz: Can you add a function to `HudUtility2024`  so we can do something like this?

```
        destination: HudUtility2024.destination_no_exit_interview_completed,
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+        auto_exited: DateTime.now,
```
gigxz:
```suggestion
        auto_exited: DateTime.current,
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+      )
+      assessment = Hmis::Hud::CustomAssessment.new_with_defaults(
+        enrollment: enrollment,
+        user: Hmis::Hud::User.system_user(data_source_id: enrollment.data_source_id),
```
gigxz: This lookup occurs on line 49 too. Can pull into a variable or an instance variable

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+        form_definition: Hmis::Form::Definition.find_definition_for_role(:EXIT, project: enrollment.project),
```
gigxz: This isn't necessary, we can leave it empty. No form definition was used to perform the assessment, so there is no need to store the reference to the one that was used. When we migrate in assessments, we also leave it blank. Maybe just don't use the `new_with_defaults` function
```suggestion
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,80 @@
...
+      )
+      assessment.data_collection_stage = 3
+      assessment.assessment_date = exit_date
+      assessment.form_processor.assign_attributes(
+        values: {},
+        hud_values: {},
+      )
+      assessment.form_processor.run!(owner: assessment, user: Hmis::User.system_user)
```
gigxz:
```suggestion
```

This is not necessary

gigxz (commented):
Partial review - will take a look at the config piece first thing tomorrow!

drivers/hmis/app/models/hmis/auto_exit_config.rb
```diff
@@ -0,0 +1,48 @@
...
+    scope = Hmis::Hud::Project.all
+
+    scope = scope.where(project_type: project_type) if project_type.present?
+    scope = scope.where(organization_id: Hmis::Hud::Organization.find_by(id: organization_id).organization_id) if organization_id.present?
+    scope = scope.where(project_id: Hmis::Hud::Project.find_by(id: project_id).project_id) if project_id.present?
+
+    scope
+  end
+
+  def self.all_projects_config
```
gigxz: For clarity, maybe call this `default_config`? It doesn't apply to all projects, since some projects may have more specific configs

drivers/hmis/app/models/hmis/auto_exit_config.rb
```diff
@@ -0,0 +1,48 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::AutoExitConfig < Hmis::HmisBase
+  self.table_name = 'hmis_auto_exit_configs'
+
+  def matching_projects
```
gigxz: This isnt used anywhere?

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+      e2 = create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, user: u1, entry_date: Date.today - 2.months
+      create :hmis_hud_service, data_source: ds1, client: c1, enrollment: e1, user: u1, record_type: 200
+      s2 = create :hmis_hud_service, data_source: ds1, client: c1, enrollment: e2, user: u1, record_type: 200, date_provided: Date.today - 31.days
+
+      Hmis::AutoExitJob.perform_now
+
+      expect(Hmis::Hud::Enrollment.exited).to include(e2)
+      expect(Hmis::Hud::Enrollment.exited).not_to include(e1)
+      expect(e2.exit).to have_attributes(auto_exited: be_present, exit_date: s2.date_provided + 1.day, destination: 30)
+      expect(e2.custom_assessments).to contain_exactly(have_attributes(assessment_date: s2.date_provided + 1.day, data_collection_stage: 3))
```
gigxz: Maybe more readable:
`expect(e2.exit_assessment&.assessment_date).to eq(s2.date_provided + 1.day)`

gigxz (commented):
Looks great! Mostly small comments. I think the only behavior change is to handle cases where the client doesn't have any "contact" records, in which case we should consider the most recent contact to be the entry date. Thanks!

ZachCMP:
@gigxz Addressed your feedback

db/warehouse/migrate/20231122185546_add_hmis_auto_exit_config.rb
```diff
@@ -0,0 +1,13 @@
+class AddHmisAutoExitConfig < ActiveRecord::Migration[6.1]
+  def change
+    add_column :Exit, :auto_exited, :datetime, null: true
+
+    create_table(:hmis_auto_exit_configs) do |t|
+      t.integer :length_of_absence_days, null: false, default: 30
+      t.integer :project_type
+      t.references :organization, foreign_key: { to_table: 'Organization' }
+      t.integer :project_id, foreign_key: { to_table: 'Project' }
```
gigxz: Did you mean to make this one reference too?

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: This one should be added @ZachCMP

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: This too, just to be safe

db/warehouse/migrate/20231122185546_add_hmis_auto_exit_config.rb
ZachCMP: Whoops, yes I did. Fixed!

gigxz:
@ZachCMP are you not seeing some of my unresolved comments? Do you see the button "6 hidden conversations" 👀

ZachCMP:
@gigxz I'm sorry Gig, I forgot that GitHub hides conversations like that so I didn't notice the missing comments. Should be all set now.

gigxz:
No problem! Thanks

ZachCMP:
I'll fix the test failures and conflict real quick

app/models/grda_warehouse/utility.rb
```diff
@@ -117,6 +117,8 @@ def self.clear!
       tables << CasCeData::Synthetic::Assessment
     end
 
+    tables << Hmis::AutoExitConfig if RailsDrivers.loaded.include?(:hmis)
```
ZachCMP: I think we need this to prevent tests from breaking due to foreign key constraints on the auto exit configs

ZachCMP:
@gigxz Conflicts should be fixed and most of the test failures should be set. However, there is one that's regarding an issue during test cleanup caused by the foreign keys on the auto exit configs. I added a line to the file throwing the error thinking that should fix it, but it didn't seem to resolve. I don't want to mess around in that utility too much and break something, so I wanted to ask how I should proceed before trying anything else. What do you think?

gigxz:
Hm, that's fixed by adding `GrdaWarehouse::Hud::Organization` to this [list](https://github.com/greenriver/hmis-warehouse/blob/9971068325039ea1efe7755355469bf1681ae1f9/app/models/grda_warehouse/utility.rb#L163-L170). But, I think we can just remove the foreign key altogether. 


I was running locally and also had to fix an issue with the migration to get the test passing. The migration created a column `project_id_id`. I'll just push up the code since I made those changes locally anyway!

gigxz (approved):
Looks good to me, thanks @ZachCMP!!

I tested locally with creating a config for a particular project with lots of old enrollments, and running the job. Confirmed that people auto-exited and the Exit Assessments appear correctly in the UI. Future enhancement will probably be to indicate in the interface that the enrollment was auto-exited.

I'll tag @eanders on here too in case he wants to take a look. One thing of note is adding a column to the Exit table so that we can easily tell which records were auto exited.

ZachCMP:
Yeah, I added a timestamp to the Exit table called `auto_exited` that should indicate when the enrollment was auto-exited if it's been auto-exited

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,84 @@
...
+module Hmis
+  class AutoExitJob < BaseJob
+    def perform
+      Hmis::Hud::Project.hmis.each do |project|
+        config = Hmis::AutoExitConfig.config_for_project(project)
+        next unless config.present?
+        raise "Auto-exit config unusually low: #{config.length_of_absence_days}" if config.length_of_absence_days < 30
+
+        project.enrollments.open_excluding_wip.each do |enrollment|
+          most_recent_contact = if project.project_type == 1 # Night-by-night Emergency Shelter
```
eanders: maybe (or do we not have that on the HMIS side?
```suggestion
          most_recent_contact = if project.es_nbn?
```

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,84 @@
...
+            [
+              enrollment.services.order(:date_provided).last,
+              enrollment.custom_services.order(:date_provided).last,
+              enrollment.current_living_situations.order(:information_date).last,
+              enrollment.custom_assessments.order(:assessment_date).last,
+              enrollment,
+            ].compact.max_by { |entity| contact_date_for_entity(entity) }
+          end
+
+          most_recent_contact_date = contact_date_for_entity(most_recent_contact)
```
eanders, line 28: this seems overly complex, why not just return `maximum(:date_provided)` (or whatever relevant date it is) and then call `compact.max` on the array?  That would give you the date without having to then check types and pick the right date.
Perhaps a slightly better alternative is to do what we do in the LSA `lib/rds_sql_server/lsa/fy2023/hmis_sql_server.rb` and setup methods in each of the models that returns the relevant date.
```
private def useful_date_column
  'DateProvided'
end
```
That way the above could be something like:

```
[
  enrollment.services.maximum(Hmis::Hud::Service.relevant_date_column),
  enrollment.custom_services.maximum(Hmis::Hud::Service.relevant_date_column),
  enrollment.current_living_situations.maximum(Hmis::Hud::CurrentLivingSituation.relevant_date_column),
  enrollment.custom_assessments.maximum(Hmis::Hud::CustomAssessment.relevant_date_column),
  enrollment,
].compact.max
```
Might not be a big gain here, but I bet we'd use those in the future.

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -0,0 +1,84 @@
...
+        end
+      end
+    end
+
+    private
+
+    def auto_exit(enrollment, most_recent_contact)
+      exit_date = contact_date_for_entity(most_recent_contact)
+      # If most recent contact was a Bed Night service, the Exit Date should be the day after they received service
+      exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
```
eanders, line 41: so this would argue for returning the full record (from above) if we need to know which kind of record it was.

eanders: I suppose the extra day could be added when fetched, since we know the source at that time.

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+  before(:all) { cleanup_test_environment }
+
+  let!(:ds1) { create(:hmis_data_source) }
+  let!(:u1) { create :hmis_hud_user, data_source: ds1, user_email: 'test@example.com' }
+  let!(:o1) { create :hmis_hud_organization, data_source: ds1, user: u1 }
+
+  describe 'with residential projects' do
+    let!(:p1) { create :hmis_hud_project, data_source: ds1, organization: o1, user: u1, project_type: 0 }
+    let!(:c1) { create :hmis_hud_client, data_source: ds1, user: u1 }
+    let!(:aec) { create :hmis_auto_exit_config, length_of_absence_days: 30 }
```
eanders, line 20: Might be worth throwing a test in here for `length_of_absence_days: 29` and confirming it throws an error.

eanders (approved):
Some yak shaving suggestions that might clean things up a bit, but looks good to me!

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
ZachCMP: Added this

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
ZachCMP: So initially this was simpler and just pulled the dates directly from the entities; But as you mention below it turned out that I needed the full most recent contact record below so I could adjust the most recent contact date if the most recent contact was a bednight service. So I adjusted this to return the whole record, not just the date, which is what made it a bit more complicated.

Since we need to know that the record was a bednight service, not just any service in order to adjust the date, I'm not sure we can just do the date adjustment on fetching as you mention below. However, I'm happy to adjust it if you think it could be improved. Let me know what you'd like me to do.

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
eanders: Thanks

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
eanders: Yep, I think we should leave it as is for now.  I'm not a fan of the case statement in `contact_date_for_entity` but I don't see a better way.  Maybe `contact_date_for_entity` should raise if the entity type is unknown/else?

ZachCMP: Added a raise to the case statement

eanders: Thanks!

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
ZachCMP: We don't seem to have this on the HNMI side. I tried `Hmis::Hud::Project.last.es_nbn?` and it gave me a no method error IIRC

eanders: Huh, @gigxz any objection to putting something like the following into `Hmis::Hud::Project`?  (We can simplify it because you just use the `project_type` column on the HMIS side, no overrides.
```
HudUtility2024.residential_project_type_numbers_by_code.each do |k, v|
      scope k, -> { where(project_type_column => v) }
      define_method "#{k}?" do
        v.include? project_type_to_use
      end
    end
```

gigxz: Yeah lets do it

ZachCMP: @gigxz This is done

gigxz:
@eanders is it OK if I merge this in with https://github.com/greenriver/hmis-warehouse/pull/3775/commits/094d36e96da531fe13c7d8816c845530b6fc3d6d ? I assume that was a recent change to 94 that didn't make it in to 95 yet

eanders:
@gigxz, should be fine, I just updated 95 from 94, so maybe you just want to update the base branch here?  Also totally happy for you to just merge this into 95 if you are ready.

PR #3778, opened 2023-11-27
186302867 Resolve last updated user from PaperTrail whodunnit

ttoomey:
## Description

[Resolve last updated user from PaperTrail whodunnit](https://www.pivotaltracker.com/story/show/186302867)

[FE PR](https://github.com/greenriver/hmis-frontend/pull/559)

* added supporting data loader
* removed old User type

## Type of change
- [x] Housekeeping / Bugfix


drivers/hmis/app/graphql/types/base_object.rb
```diff
@@ -49,6 +49,18 @@ def load_ar_association(object, association, scope: nil)
...
+    def load_ar_scope(scope:, id:)
+      dataloader.with(Sources::ActiveRecordScope, scope).load(id)
+    end
+
+    def load_last_user_from_versions(object)
+      versions = load_ar_association(object, :versions)
+      last_user_id = versions.max_by(&:created_at)&.whodunnit
+      return unless last_user_id && last_user_id =~ /\A[0-9]+\z/
+
+      load_ar_scope(scope: Hmis::User.all, id: last_user_id)
```
ttoomey: Perhaps this should be `Hmis::User.viewable_by(current_user)` however that requires that the user has `can_impersonate_users` which is maybe not right. Perhaps we should adjust Hmis;:User.viewable_by ?

drivers/hmis/app/graphql/types/hmis_schema/file.rb
```diff
@@ -71,13 +71,9 @@ def date_updated
 
     # HUD User that most recently touched the record, to match convention on HUD-like types
```
gigxz:
```suggestion
```

drivers/hmis/app/graphql/types/hmis_schema/has_hud_metadata.rb
```diff
@@ -13,7 +13,7 @@ module HasHudMetadata
 
       included do
         # HUD User that most recently touched the record
```
gigxz:
```suggestion
        # User that most recently touched the record
```

drivers/hmis/app/graphql/types/base_object.rb
gigxz: In this case, the scope should be `Hmis::User.all` because there are no restrictions on seeing who last touched a record that you can see.

The `viewable_by` scope will need to be adjusted to remove the "can impersonate" limit, I think. That scope is used to determine which users show up on the User list on the HMIS Admin dashboard. "viewable by" probably isn't the right term for that scope, since it won't be user-dependent. It should probably just be called `hmis_users` or something. The change feels unrelated to this PR though.

gigxz: Nice. Loading all the versions into memory feels a bit scary, but I think the alternative is to not use the data loader, which would probably be worse.

I think elsewhere we use `load_ar_association` in a way it probably can't handle -- loading relations that have a scope condition on them. So for example if we added a "relation" `has_one :latest_version, -> { order(created_at: :desc) }` and then tried to do `load_ar_association(object, :latest_version)`. I did that elsewhere, like [here](https://github.com/greenriver/hmis-warehouse/blob/4f8c89aae6d80e403ec646ae5e291934f0a691ad/drivers/hmis/app/graphql/types/hmis_schema/merge_audit_event.rb#L31)/[here](https://github.com/greenriver/hmis-warehouse/blob/4f8c89aae6d80e403ec646ae5e291934f0a691ad/drivers/hmis/app/models/hmis/client_merge_audit.rb#L13-L14), but I have a feeling that it's not actually using the data loader effectively.  I'm curious if you have any immediate thoughts about that approach 👀

gigxz (approved):

drivers/hmis/app/graphql/types/base_object.rb
eanders: Do we need to include_deleted?  Hmis::User is acts_as_paranoid, so if a user was deleted, we should probably still know they changed something.

gigxz: Good call, yes we should include deleted

ttoomey: Thanks, yeah I was thinking there might be less than 100 version records here but maybe that's not a good assumption. I narrowed the selection on the versions to avoid loading the version object/changes and moved the order to the db as you suggest, that should be more performant.

I'll review the usage of `load_ar_association()` you mentioned. It feels like you'd need to group by versions.item to make that work in a batch but I'm probably missing something

ttoomey: Good catch, I'm including deleted users now

gigxz: Hmmm I am confused. I wasn't suggesting a particular change, sorry for my lack of clarity. I didn't think that passing the `scope` to `load_ar_association` worked in the way you added it. I specifically added a comment a few months ago to `load_ar_association` that says
```
    # Note: 'scope' is intended for ordering or to modify the default
    # association in a way that is constant with respect to the resolver,
    # for example `scope: FooBar.order(:name)`. It is NOT used to filter down results.
```

I think, that was based on a conversation we had 😃  my understanding is vaguely that you basically _cant_ filter down the scope if you want to use the data loader. I would love to discuss further, maybe we can jump on a call!

gigxz: Ookay, per discussion: this works and is exactly how the `scope` param is intended to be used. The refinements are "constant" in that they do not depend on the object at all. In my comment `It is NOT used to filter down results.` is not accurate. It can be used to "filter" (`where.not(whodunnit: nil)`), as long as that filter is constant

PR #3781, opened 2023-11-28
Adjustments for medicaid exchange calculations

eanders:
## Description

This adds in a requirement that an enrollment be "with service" when picking which clients to include, and when picking which days to include.  Additionally, a client must have at least one "entry" into any project in the past year to be included (basically looking for some HMIS activity.)

## Type of change


app/models/grda_warehouse/service_history_enrollment.rb
```diff
@@ -294,6 +294,17 @@ def self.service_types
     #   exists)
   end
 
+  scope :with_service, -> do
+    where(
+      GrdaWarehouse::ServiceHistoryService.
+      where(
+        shs_t[:service_history_enrollment_id].eq(she_t[:id]).
+        and(shs_t[:client_id].eq(she_t[:client_id])),
```
awisema, line 302: Confirming: so, this will exclude a household member who does not actually have any service, even if other members do?

Are there any cases where this might exclude people who shouldn't be? Example, a service is provided to a parent, but not required for the child, should the child be excluded?

drivers/medicaid_hmis_interchange/app/models/medicaid_hmis_interchange/health/submission.rb
```diff
@@ -54,16 +54,25 @@ def response_filename
...
     private def generate_submission
       file_path = File.join(@file_path, submission_filename)
       count = 0
       return send_test_file(file_path) if test_file
 
       seen_medicaid_ids = Set.new
+      clients_with_entries_in_past_year = GrdaWarehouse::ServiceHistoryEnrollment.entry.
+        where(entry_date: 1.year.ago.to_date..)
+      # `homeless_on_date` encompasses ongoing homeless enrollment with with service and no overlapping moved-in PH
       GrdaWarehouse::Hud::Client.homeless_on_date.
```
awisema, line 72: Assign this to a variable? I had trouble following all this logic only to end up with an iteration

awisema (approved):

app/models/grda_warehouse/service_history_enrollment.rb
eanders: I could be wrong, but in this context, we're really only talking about bed-nights, and I believe everyone who stays at an NbN ES should receive a bed night, regardless of age.

PR #3786, opened 2023-11-29
Resolve basic user access history

gigxz:
## Description

* Add permission "can audit users". This will govern access to activity logs ("access history") and papertrail audit history ("edit history")
* Resolve very basic Activity Log. As discussed re:[the ticket](https://www.pivotaltracker.com/n/projects/2591838/stories/186538152), it does NOT resolve any untrusted information from headers. Per the ticket, we'll need to add post-processing to make the interface more helpful. For now, it's just showing the raw records IDs (eg "Client 4", "Service 6", ...)

## Type of change


drivers/hmis/app/graphql/types/application/activity_log.rb
```diff
@@ -0,0 +1,35 @@
...
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :resolved_records, [Types::Application::ActivityLogRecord], null: false
+
+    def user
+      load_ar_association(object, :user)
+    end
+
+    def resolved_records
+      object.resolved_fields.keys.map do |key|
+        record_type, record_id = key.split('/')
```
ttoomey: this is fine as as. You can also provide a second arg to split just in-case there's an extra '/' in key somehow
```suggestion
        record_type, record_id = key.split('/', 2)
```

drivers/hmis/app/graphql/types/hmis_schema/has_access_history.rb
```diff
@@ -0,0 +1,30 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Types
+  module HmisSchema
+    module HasAccessHistory
```
ttoomey: hmm... I might be missing something but is this module used?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -257,7 +257,18 @@ def merge_candidates
...
+      resolve_application_users(Hmis::User.hmis_users, **args)
+    end
+
+    field :user, Types::Application::User, 'User lookup', null: true do
+      argument :id, ID, required: true
+    end
+    def user(id:)
+      raise 'access denied' unless id == current_user.id.to_s || current_user.can_audit_users? || current_user.can_impersonate_users?
+
+      Hmis::User.find_by(id: id)
```
ttoomey: Should this be restricted to `Hmis::User.hmis_users` scope? Also might as well use the loader even for single-record query roots like this 🤷 

```suggestion
      load_ar_scope(scope: Hmis::User.hmis_users, id: id)
```

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -22,11 +22,17 @@ class Hmis::User < ApplicationRecord
   has_many :access_controls, through: :user_groups
   has_many :access_groups, through: :access_controls
   has_many :roles, through: :access_controls
+  has_many :activity_logs, class_name: 'Hmis::ActivityLog'
 
   has_recent :clients, Hmis::Hud::Client
   has_recent :projects, Hmis::Hud::Project
   attr_accessor :hmis_data_source_id # stores the data_source_id of the currently logged in HMIS
 
+  scope :hmis_users, -> do
```
ttoomey: If I call `Hmis::User.hmis_users` it's a little hard to tell what it does just by the name. Perhaps `Hmis::User.active_members` or "operational_users" would be more descriptive?

ttoomey (approved):
Looks good! Left a few comments

drivers/hmis/app/graphql/types/hmis_schema/has_access_history.rb
gigxz: oops, dont think so! will remove

drivers/hmis/app/models/hmis/user.rb
gigxz: Hmm yeah I agree. I want to capture that they are _hmis_ users -- warehouse-only users are not included. Maybe it would be better to split into two descriptive scopes like `Hmis::User.active.with_hmis_access`. We will eventually probably want to show inactive HMIS users at some point too, so I think that makes sense.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: I suppose it can be restricted down to hmis-only users. We should allow inactive because we're probably going to incorporate that into the UI at some point, so I'll leave those allowed

PR #3793, opened 2023-12-03
Ignore cp1 for initial assessments

awisema:
## Description

Change the logic so that an expiring CP1 careplan does not need to be "renewed", instead, they patient needs an intake into CP2.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/models/health/careplan.rb
+++ b/app/models/health/careplan.rb
@@ -373,7 +373,9 @@ module Health
       return false unless cp1?
 
       # Only the most recent PCTP needs review
-      patient.pctp_careplans.sorted.first&.instrument&.id == id
+      return false unless patient.recent_pctp_careplan&.instrument == self
+
+      careplan.careplan_sent_on.blank?
     end
 
     def compact_future_issues
--- a/app/models/health/careplan.rb
+++ b/app/models/health/careplan.rb
@@ -368,13 +368,16 @@ module Health
       ! cp1?
     end
 
+    # For the transition between CP1 and CP2, MH allowed careplans in development
+    # at the time of the transition to be carried forward after a review.
     def needs_rereview?
       # Only CP1 careplans need to be reviewed for CP2
       return false unless cp1?
 
-      # Only the most recent PCTP needs review
+      # Only the most recent PCTP can need review
       return false unless patient.recent_pctp_careplan&.instrument == self
 
+      # Once sent to the PCP, the careplan is completed.
       careplan.careplan_sent_on.blank?
     end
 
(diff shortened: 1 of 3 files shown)


app/models/health/careplan.rb
```diff
@@ -373,7 +373,9 @@ def needs_rereview?
       return false unless cp1?
```
eanders, line 375: Does this need to be removed?  Looks like cp2 won't ever get past this.

eanders: Can you add some comments to this method to indicate what the rules are for needing review?  I'll be better able to review this for functionality if I know what the intention is.

app/models/health/patient.rb
```diff
@@ -411,22 +415,22 @@ def self.cp_2_engagement(on) # rubocop:disable Naming/MethodParameterName
...
       joins(:recent_pctp_careplan).
-        merge(Health::PctpCareplan.recent.sent_within(.. on - 335.days)) # 1 year - 30 days
+        merge(Health::PctpCareplan.recent.cp2.sent_within(.. on - 335.days)) # 1 year - 30 days
     end
 
     scope :overdue_for_renewal, ->(on: Date.current.end_of_month) do
       joins(:recent_pctp_careplan).
-        merge(Health::PctpCareplan.recent.sent_within(.. on - 365.days)) # 1 year
+        merge(Health::PctpCareplan.recent.cp2.sent_within(.. on - 365.days)) # 1 year
     end
```
eanders, line 445: Maybe a comment above these scopes (and below?) to indicate they only apply to cp2?

app/models/health/pctp_careplan.rb
```diff
@@ -127,8 +135,10 @@ def expiring?
     # # @return [Boolean] True if the careplan is expired
```
eanders: Can you enumerate what constitutes an expired care plan?

app/models/health/pctp_careplan.rb
```diff
@@ -131,8 +131,10 @@ def expiring?
       active? && expires_on - 1.month < Date.current
     end
 
-    # A Careplan has expired if the expiration date is in the past
-    # # @return [Boolean] True if the careplan is expired
+    # A Careplan has expired if the expiration date is in the past except for careplans
+    # from prior CP programs, which need to be replaced instead of reviewed.
```
eanders:
```suggestion
    # from prior CP programs, which need to be replaced instead of renewed.
```

awisema:
After discussion yesterday, I think we don't need to do this, so closing.

PR #3799, opened 2023-12-05
186522571 bugfix audit history

ttoomey:

## Description
[Bugfix for 186522571](https://www.pivotaltracker.com/story/show/186522571)

Fix bug in audit reporting around client race/gender changes. I believe this was because version.reify changes version.item unexpectedly. Rather than reifing, I think we can use the changes directly since these fields are simple integer codes

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/client.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/client.rb
@@ -93,23 +93,16 @@ module Types
       transform_changes: ->(version, changes) do
         result = changes
         [
-          ['race', Hmis::Hud::Client.race_enum_map, :RaceNone],
-          ['gender', Hmis::Hud::Client.gender_enum_map, :GenderNone],
+          ['race', Hmis::Hud::Client.race_enum_map, 'RaceNone'],
+          ['gender', Hmis::Hud::Client.gender_enum_map, 'GenderNone'],
         ].each do |input_field, enum_map, none_field|
           relevant_fields = [*enum_map.base_members.map { |member| member[:key].to_s }, none_field.to_s, input_field]
           next unless changes.slice(*relevant_fields).present?
 
           result = result.except(*relevant_fields)
-          old_client = version.reify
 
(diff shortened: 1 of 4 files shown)


ZachCMP (approved):
LGTM

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -93,25 +93,30 @@ def self.configuration
...
+          ].map do |doc|
+            none_value = doc[none_field]
+            if none_value.nil? || none_value.zero?
+              enum_map.base_members.
+                filter { |item| doc[item[:key].to_s] == 1 }.
+                map { |item| item[:value] }
+            else
+              [none_value]
+            end
+          end
```
gigxz, line 117: I haven't tracked down the cause yet, but I'm not seeing the expected behavior here. If I set a client up with 2 genders, then remove one of them, this is resolving that gender got changed to `null`, even though it got changed to "culturally specific." I confirmed in the version table that the correct change was recorded, but it doesn't appear to be resolved accurately.

gigxz: `changes` is just the `version.object_changes`, it looks like (see base audit event). Why are we using `version.object` to compare with `version.object_changes.transform_values(&:last)` -- wouldn't they be the same? I think I'm confused..


As a general concern, it would be preferred to continue to show ALL gender (or race) fields in the change history. I can't tell if this code is trying to do that or not.

For example, if someone previously had `Man, Non-Binary` selected, and then removed the `Man` option, the change history should appear as `Gender:     Man, Non-Binary  =>  Non-Binary`. I don't think we can do that without reifying because we need to know the previous value for all genders (since the `Non-Binary` column didn't change, it probably wouldn't be included in object_changes, right?).

If that presents issues, we could display it differently in the interface, like having one "row" per gender. (eg `Man:    True => False`, but I think its less clear.

gigxz (approved):

PR #3800, opened 2023-12-05
Limit export lengths

eanders:

## Description

This enforces HMIS structure max-lengths when exporting HMIS data.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/assessment.rb
+++ b/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/assessment.rb
@@ -12,14 +12,6 @@ module HmisCsvTwentyTwentyFour::Exporter
       @options = options
     end
 
-    def process(row)
-      row = assign_export_id(row)
-      row = self.class.adjust_keys(row, @options[:export])
-      row = enforce_lengths(row)
-
-      row
-    end
-
     def self.adjust_keys(row, export)
       row.UserID = row.user&.id || 'op-system'
       # Pre-calculate and assign. After assignment the relations will be broken
--- a/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/assessment.rb
+++ b/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/assessment.rb
@@ -15,6 +15,7 @@ module HmisCsvTwentyTwentyFour::Exporter
     def process(row)
       row = assign_export_id(row)
       row = self.class.adjust_keys(row, @options[:export])
+      row = enforce_lengths(row)
 
       row
     end
--- a/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/enrollment/overrides.rb
+++ b/drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/enrollment/overrides.rb
@@ -30,8 +30,6 @@ module HmisCsvTwentyTwentyFour::Exporter
       row = assign_move_in_date(row)
       row = assign_missing_coc(row)
 
-      row.PreferredLanguageDifferent = row.PreferredLanguageDifferent[0...100] if row.PreferredLanguageDifferent
-
       row
     end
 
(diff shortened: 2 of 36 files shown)


drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/export_concern.rb
```diff
@@ -110,12 +110,28 @@ def self.csv_header_override(keys)
...
       row = self.class.adjust_keys(row)
 
       row
     end
 
+    def enforce_lengths(row)
+      length_limited_columns.each do |k, opts|
+        next if row[k].blank? || row[k].length <= opts[:limit]
+
+        row[k] = row[k][0..opts[:limit]]
```
awisema: `(0..32).length == 33`, so I think this should include one too many chars?
```suggestion
        row[k] = row[k][0...opts[:limit]]
```
The 0 is optional, but maybe clearer than `...opts[:limit]`?

drivers/hmis_csv_twenty_twenty_four/spec/models/exporter/single_enrollment_tests.rb
```diff
@@ -25,6 +25,12 @@
       csv = CSV.read(csv_file_path(@enrollment_class), headers: true)
       expect(csv.first['PersonalID']).to_not be_empty
     end
+
+    it 'Correctly limits the length of PreferredLanguageDifferent to 100 characters' do
+      expect(@exporter.enrollment_scope.first.PreferredLanguageDifferent.length).to eq(200)
+      csv = CSV.read(csv_file_path(@enrollment_class), headers: true)
+      expect(csv.first['PreferredLanguageDifferent'].length).to eq(100)
```
awisema, line 32: Something else is masking this, I commented out `enforce_lengths`, and the test still passes.

awisema (changes requested):

drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/export_concern.rb
eanders: Thanks!

drivers/hmis_csv_twenty_twenty_four/spec/models/exporter/single_enrollment_tests.rb
eanders: Found and fixed.  Thank you

drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/enrollment/overrides.rb
```diff
@@ -30,8 +30,6 @@ def self.apply_overrides(row)
       row = assign_move_in_date(row)
       row = assign_missing_coc(row)
 
-      row.PreferredLanguageDifferent = row.PreferredLanguageDifferent[0...100] if row.PreferredLanguageDifferent
```
awisema, line 33: I think we should remove the other places this is done by hand (e.g., `HmisCsvTwentyTwentyFour::Exporter::Client::Overrides#apply_length_limits`) so that changes in the structure files are correctly reflected in the exports.

drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/exporter/assessment.rb
```diff
@@ -15,6 +15,7 @@ def initialize(options)
     def process(row)
```
awisema: These look the same as in the concern, can we remove them?

Also, I find myself wondering about making sure this all gets done everywhere... Either break it into two parts, and have `process` call a `common_processing` method, or use inheritance. Thenin the cases where we need to extend `process` we can either invoke `common_processing` or `super`?

awisema (changes requested):
I'm sorry --some proposed architectural changes...

eanders:
@awisema , I DRYed things up as best I could.  There were some subtle differences on the `adjust_keys` methods, but I think this takes care of that.  I didn't want to make this many changes, but it's probably for the best.  Let me know if you see anything else we should fix.

awisema (approved):
Thank you! Sorry to make this a bigger clean up, but I think this will help with future spec changes.

PR #3801, opened 2023-12-06
Update confidential files

eanders:

## Description

This: 
1. Sets the default when uploading confidential files to "confidential" if the user can see confidential files
2. Allows uploads of confidential files not connected to an enrollment.  Anyone with the "can see confidential files" will then be able to see those uploads.

## Type of change

### DIFF
--- a/app/models/grda_warehouse/client_file.rb
+++ b/app/models/grda_warehouse/client_file.rb
@@ -26,7 +26,7 @@ module GrdaWarehouse
     validates_presence_of :effective_date, on: :requires_expiration_and_effective_dates, message: 'Effective date is required'
     validates_presence_of :expiration_date, on: :requires_expiration_and_effective_dates, message: 'Expiration date is required'
 
-    validates_presence_of :enrollment_id, if: :confidential?
+    # validates_presence_of :enrollment_id, if: :confidential?
 
     scope :confidential, -> do
       where(confidential: true)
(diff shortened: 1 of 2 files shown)


app/models/grda_warehouse/client_file.rb
```diff
@@ -26,7 +26,7 @@ class ClientFile < GrdaWarehouse::File
     validates_presence_of :effective_date, on: :requires_expiration_and_effective_dates, message: 'Effective date is required'
     validates_presence_of :expiration_date, on: :requires_expiration_and_effective_dates, message: 'Expiration date is required'
 
-    validates_presence_of :enrollment_id, if: :confidential?
+    # validates_presence_of :enrollment_id, if: :confidential?
```
awisema: Delete instead of comment out?

app/models/grda_warehouse/client_file.rb
```diff
@@ -40,16 +40,22 @@ class ClientFile < GrdaWarehouse::File
...
+    # OR if the file does not have an associated enrollment and the user can view some confidential files
     scope :confidential_visible_by, ->(user) do
       permission = :can_see_confidential_files
       return none unless user&.send("#{permission}?")
 
+      confidential_no_enrollment_scope = confidential.where(enrollment_id: nil)
       ids = user.viewable_project_ids(permission)
       # If have a set (not a nil) and it's empty, this user can't access any projects
-      return none if ids.is_a?(Set) && ids.empty?
+      return confidential_no_enrollment_scope if ids.is_a?(Set) && ids.empty?
```
awisema: I'm concerned what happens here, if a change in `viewable_project_ids` accidentally, or deliberately changes the return type. Maybe

```suggestion
      return confidential_no_enrollment_scope if !ids.nil? && ids.empty?
```

awisema (changes requested):

app/models/grda_warehouse/client_file.rb
eanders: Hmm, I think I'd rather check for a set.  If the return changes to an array, what you have would work, but not everything response to `.empty?`.   Maybe we should raise if it's not a `Set`.

app/models/grda_warehouse/client_file.rb
```diff
@@ -40,16 +38,24 @@ class ClientFile < GrdaWarehouse::File
...
     scope :confidential_visible_by, ->(user) do
       permission = :can_see_confidential_files
       return none unless user&.send("#{permission}?")
 
+      confidential_no_enrollment_scope = confidential.where(enrollment_id: nil)
       ids = user.viewable_project_ids(permission)
+
       # If have a set (not a nil) and it's empty, this user can't access any projects
-      return none if ids.is_a?(Set) && ids.empty?
+      raise 'Unexpected response from user.viewable_project_ids' unless ids.is_a?(Set)
```
awisema, line 51:
```suggestion
      raise 'Unexpected response from user.viewable_project_ids' unless ids.nil? || ids.is_a?(Set)
```
?

awisema (changes requested):

app/models/grda_warehouse/client_file.rb
awisema: Actually, nevermind... `nil` responds to `is_a?`.

awisema (approved):

PR #3806, opened 2023-12-07
186632025 annuals due bug

ZachCMP:

## Description

Updates the logic of the reminder generator to use the closest anniversary date rather than `date.today`

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/models/hmis/reminders/reminder_generator.rb
+++ b/drivers/hmis/app/models/hmis/reminders/reminder_generator.rb
@@ -89,7 +89,9 @@ module Hmis
         # not due for an assessment yet
         return if today < (hoh_entered_on + (1.year - window))
 
-        hoh_anniversary = hoh_entered_on.change(year: today.year)
+        # Find the closest HOH entry anniversary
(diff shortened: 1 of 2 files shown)


ZachCMP:
@gigxz FYI I merged your regression tests branch into this one so I could use the tests you wrote.

There are a couple ways that this could be addressed, and this seemed like the easiest. Essentially it attempts to find the closest anniversary date to `today` rather than just changing the anniversary date to `today.year`. This makes the tests pass and seems generally good, but there is a potential downside I could see. If there is an enrollment that has annuals due past the due date, and still has not completed them by the time it passes the threshold to use the next anniversary date, then the annuals will no longer show as overdue. Instead it will show as not due yet, since the closest anniversary is in the future.

If we wanted to address this, we'd probably have to change the code to check against both anniversaries, probably returning the most recent reminder. That should eliminate the downside above, but would increase the complexity and calculation needed. Let me know if you'd like me to go this route instead.

gigxz (approved):
Nice!

PR #3820, opened 2023-12-12
Static form definitions

gigxz:

## Description

* Add a new category of Form Definition "roles" to represent static forms.
* Add a query to fetch a specific static form definition
* Static forms are not customizable. They are not managed using form instances.
* Static forms do not use a shared mutation (e.g. `SubmitForm`). Rather, each static form type likely uses its own custom mutation(s) for creation and updating.

## Type of change


drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -200,6 +200,14 @@ def get_service_form_definition(service_type_id:, project_id:)
       Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
     end
 
+    field :static_form_definition, Types::Forms::FormDefinition, null: false do
+      argument :role, Types::Forms::Enums::StaticFormRole, required: true
+    end
+    def static_form_definition(role:)
+      # Direct lookup for static form by role. Static forms don't require instances to enable them, since they are always present and non-configurable.
+      Hmis::Form::Definition.with_role(role).first!
```
ttoomey: It doesn't look like role is unique for form definition. If there were two definitions with the same role, it looks like you'd get one at random. Maybe we just assume that's okay? If so, let's add an order so at least we'll get the same definition consistently:
```suggestion
      Hmis::Form::Definition.order(:id).with_role(role).first!
```

ttoomey (approved):
looks good!

ZachCMP (approved):

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Yeah, making an assumption here that static forms will have unique roles. I'll add in the ordering, thanks!

PR #3824, opened 2023-12-15
Tt/spm 185769360

ttoomey:
[#185769360
](https://www.pivotaltracker.com/story/show/185769360)SPM test fixes for measure 4, 5, and 7. Tests are close to expected results, discrepancies are noted in all_projects_spec

**NOTE**: I believe current failures in CI are because we need to update the testkit fixpoints. Previously we were correcting enrollments with invalid coc codes, testkit seems to expect us to exclude those. The importer now allows skipping the location correction. I have fix point files already generated I can share.

* adjust previous income for leavers to fix measure 4.4, 4.5, 4.6
* use hud total income calc for measure 4.6
* fix missing records from prior counts in 5.1/5.2
* fix conditions in in 7a.1 and 7b.1 universe
* adjust 7b.2 universe calculation to more closely match spec
* exclude move_in date that occurs before entry date, fixes 7b.2
* include stayers in 7b.2 c3 count

* db migration to match enrollment income field data type to IncomeBenefits, remove unused indexes
* uncommented relevant tests

- [x] I have reviewed my code
- [x] I have run it and it doesn't crash


drivers/hud_spm_report/spec/models/datalab_testkit/all_projects_spec.rb
```diff
@@ -50,112 +51,93 @@
...
-  #       skip: [
-  #         'C4',
-  #       ],
-  #     )
-  #   end
+    it 'Measure 7b.2' do
+      compare_results(
+        file_path: result_file_prefix + 'spm',
+        question: '7b.2',
+        # off by 30
```
ttoomey: Not sure about the source of these extra records. Something I notice is that 23 of the extra records have entry_date > move_in_date; all of the expected records have entry_date <= move_in_date. I don't see direction in this in the spec except as it relates to "General Report Instructions: household member joins the household after they are already housed" which should already be accounted for in `SpmEnrollment.move_in_date`

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_seven.rb
```diff
@@ -165,30 +170,30 @@ def run_question!
...
       stayers = so_enrollments.where(spm_e_t[:exit_date].eq(nil).or(spm_e_t[:exit_date].gt(filter.end)))
-      leavers = so_enrollments.where.not(id: stayers.select(:id))
+      leavers = so_enrollments.where.not(client_id: stayers.select(:client_id))
       enrollments = HudSpmReport::Fy2023::SpmEnrollment.one_for_column(:exit_date, source_arel_table: spm_e_t, group_on: :client_id, scope: leavers)
       enrollments = enrollments.where.not(spm_e_t[:destination].in(M7A_REJECTED))
 
       members = enrollments.map do |enrollment|
         [enrollment.client, enrollment]
       end.to_h
-      @universe.add_universe_members(members)
```
ttoomey, line 180: It seems like universe doesn't need to be an instance var, right?

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/spm_enrollment.rb
```diff
@@ -198,7 +198,9 @@ def self.detail_headers
       end
     end
 
-    private_class_method def self.previous_income_benefits(enrollment, annual_date)
+    private_class_method def self.previous_income_benefits(enrollment, annual_date, end_date)
+      return enrollment.income_benefits_at_entry if enrollment.exit.present? && enrollment.exit.exit_date <= end_date
```
ttoomey, line 223: mirror behavior of current_income_benefits() above

drivers/hud_spm_report/spec/models/datalab_testkit/all_projects_spec.rb
eanders: Entry dates must be <= move-in.  If they are after move-in, the move-in is considered invalid and should not be used.  Maybe that's what's going on?  If a household member joins a household after the move-in of the HoH, their move-in date is their entry date.

ttoomey: Good to know, that seems to have resolved the issue and 7b.2 now matches testkit. Thanks!

eanders: Excellent!

awisema (approved):

PR #3825, opened 2023-12-15
Dg/demographic-summary-report-186468324

dtgreiner:

## Description

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
@@ -121,20 +121,42 @@ module
       end
     end
 
-    def age_count(type)
-      mask_small_population(clients_in_age_range(type)&.count&.presence || 0)
+    def age_coc_count(age_category, coc_code)
+      age_category_clause = "#{age_category}_clause"
+      mask_small_population(
+        report_scope.
+          in_coc(coc_code: coc_code).
+          joins(:client).
+          where(send(age_category_clause)).
+          select(:client_id).distinct.count,
+      )
     end
 
-    def clients_in_age_range(type)
-      client_ages.select { |_, age| age.in?(type) }
+    def age_gender_coc_count(age_category, gender_col, coc_code)
+      age_category_clause = "#{age_category}_clause"
+      mask_small_population(
+        report_scope.
+          in_coc(coc_code: coc_code).
+          joins(:client).
+          where(send(age_category_clause).and(gender_clause(gender_col))).
+          select(:client_id).distinct.count,
+      )
+    end
+
+    def age_count(type, coc = base_count_sym)
+      mask_small_population(clients_in_age_range(type, coc)&.count&.presence || 0)
+    end
+
+    def clients_in_age_range(type, coc = base_count_sym)
+      client_ages[coc].select { |_, age| age.in?(type) }
     end
 
     def client_ids_in_age_range(type)
       clients_in_age_range(type).keys
     end
 
-    def age_percentage(type)
-      total_count = mask_small_population(client_ages.count)
+    def age_percentage(type, coc = base_count_sym)
+      total_count = mask_small_population(client_ages[coc].count)
       return 0 if total_count.zero?
 
       of_type = age_count(type)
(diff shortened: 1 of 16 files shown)


drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
```diff
@@ -152,25 +174,47 @@ def age_data_for_export(rows)
         rows["_#{age_category_title} Break"] ||= []
         rows["*#{age_category_title}"] ||= []
         rows["*#{age_category_title}"] += ['Gender', nil, 'Count', 'Average Age', nil]
+        available_coc_codes.each do |coc_code|
+          rows["*#{age_category_title}"] += [coc_code]
+        end
```
eanders, line 179: Isn't this the same as 
```suggestion
        rows["*#{age_category_title}"] += available_coc_codes
```

drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
```diff
@@ -152,25 +174,47 @@ def age_data_for_export(rows)
...
+        rows["*#{age_category_title}"] += [nil]
         rows["_#{age_category_title} - All"] ||= []
         rows["_#{age_category_title} - All"] += ['All', nil, send("#{age_category}_count"), send("average_#{age_category}_age"), nil]
+        available_coc_codes.each do |coc_code|
+          rows["_#{age_category_title} - All"] += [age_coc_count(age_category, coc_code.to_sym)]
+        end
```
eanders, line 185: Similarly, I think this could be written like the following, which might be easier to follow, though maybe not.  I'm not as convinced about this one since it still requires iterating through the array.
```suggestion
        rows["_#{age_category_title} - All"] += available_coc_codes.map { |coc_code| [age_coc_count(age_category, coc_code.to_sym)] }

```

drivers/core_demographics_report/app/models/core_demographics_report/high_acuity_calculations.rb
```diff
@@ -39,60 +39,134 @@ def high_acuity_data_for_export(rows)
...
+        'Adult and Child Households' => :with_children,
+        'Child only Households' => :only_children,
+        'Youth Only' => :unaccompanied_youth,
+        '1 Disability' => :one_disability,
       }
     end
 
+    private def without_children
+      @without_children ||= enrollment_ids_in_household_type(:without_children)
+    end
```
eanders: I think having these in multiple concerns is unnecessary, and potentially dangerous if they have different definitions.  I't probably worth gathering them up in the `report_concern.rb` or a `household_type_definitions` concern and then including that in the report before these concerns.  If there are differences in the implementation, then we should make sure they have different names.  
This is one of the pitfalls of these reports with lots of concerns, it can be hard to track down method replacement and keep track of the API.

drivers/core_demographics_report/app/models/core_demographics_report/household_type_calculations.rb
```diff
@@ -60,38 +60,78 @@ def household_type_client_percentage(type)
...
 
     def available_household_types
       @filter.available_household_types.merge('Youth Only' => :unaccompanied_youth, 'Unknown' => :unknown)
     end
 
     private def calculate_households
+      @hoh_enrollments ||= Rails.cache.fetch([self.class.name, cache_slug, "#{__method__}_hoh_enrollments"], expires_in: expiration_length)
+      @households ||= Rails.cache.fetch([self.class.name, cache_slug, "#{__method__}_households"], expires_in: expiration_length)
+
+      return unless @hoh_enrollments.nil? && @households.nil?
```
eanders, line 107: Hmm, this isn't how we usually use Rails.cache.fetch.  It' might work just fine, and maybe this is the right way since you're setting two instance variables that use related calculations.  Usually we pass a block to cache.fetch, which is only executed if the cached variable doesn't exist.

Another, perhaps bigger issue, is that we don't want to unintentionally change the Core Demographics report, which uses these same concerns.  I think we need to move `available_coc_codes` into an attribute accessor and set it from the calling controller.  That way we can ensure we never set it to anything but an empty array for Core Demographics, but can pass the appropriate CoC Codes for the Demographic Summary report.

eanders: I had a thought on how to prevent running the CoC Calculations when unnecessary, I'll push some code shortly.

dtgreiner: I added the caching in for the hoh_enrollments and households variables because the query setting them was a long running one for this section and it helped with development. I'm putting together a new commit with the other suggestions and will plan on pulling it back out to keep the caching consistant with the rest of the application.

PR #3827, opened 2023-12-16
Mailhog removal and documentation updates

eanders:

## Description

* This removes mailhog and makes it an external dependency with documentation for setup in the app. 
* Also better documentation for traefik including necessary files.  
* AND, this makes ngrok into a manually started service since we only need it 0.01% of the time
* We will want a corollary change on CAS.

## Type of change

- [x] Dependency update

### DIFF
--- a/docs/traefik/docker-compose.yml
+++ b/docs/traefik/docker-compose.yml
@@ -1,0 +1,50 @@
+version: '3.9'
+
+services:
+  reverse-proxy:
+    # The official v2 Traefik docker image
+    image: traefik:v2.8
+    container_name: traefik
+    # Enables the web UI and tells Traefik to listen to docker
+    command:
+      # - --log.level=DEBUG
+      - --api.insecure=true
+      - --entrypoints.web.address=:80
+      - --entrypoints.web-secure.address=:443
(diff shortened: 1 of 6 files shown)


docs/traefik/docker-compose.yml
```diff
@@ -0,0 +1,50 @@
+version: '3.9'
```
yakloinsteak, line 1: I wonder if we can put the traefik and network related files into the virtual-machines repo and just point to that repo. Is it here because this is open source? We can probably add alt names for all the projects and make it work for everybody.

yakloinsteak (approved):
My vote is definitely for putting the non-open-path stuff into the virtual machines repo, but I don't feel strongly enough to block anything.

docs/traefik/docker-compose.yml
eanders: Yeah, only leaving it in the repo because some users don't have access to the virtual machines repo (which is closed source.)

dtgreiner (approved):
Everything looks good to me. I was able to have it all up and running without issue. The only other thing I can think of is reminding people to clean up any override files they've been using relating to mailhog.

PR #3829, opened 2023-12-17
186685876-mci-timeout

ttoomey:

## Description
[186685876](https://www.pivotaltracker.com/story/show/186685876)

Bug:
While the extended timeout works for the initial token request, the default 5 sec timeout is applied to the subsequent resource request for some reason. This seems to be an issue in farday, rather than the oauth2 lib that calls it.

Solution:
It's possible to use a monkey patch to address this but I think it is better to upgrade faraday to fix the bug. This is a bit complicated because our custom fork of the hellosign-ruby-sdk gem has a dependency on faraday 2.2.x and prevents upgrading to 2.7 (current). To upgrade we need to patch our fork.

The HelloSign dependency is no longer needed but can be left as-is until we can remove the dead code. It does not need QA. 

## Type of change
- [x] Bug fix
- [x] Dependency update


gigxz:
Thanks for tracking this down @ttoomey!! Nice work

eanders (approved):
Thanks for the changes!

gigxz (approved):
💯

gigxz:
The test failure is a flaky one that I have seen occasionally; unrelated

PR #3831, opened 2023-12-18
Support fixing hh with multiple HoH or no HoH

gigxz:

## Description

* Fix issue where if a household has 2 HoHs, you couldn't make 1 of them the HoH
* Add tests for both data-quality-fix scenarios (no HoH or multiple HoH)
* Switch to raising errors where appropriate

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/mutations/create_units.rb
+++ b/drivers/hmis/app/graphql/mutations/create_units.rb
@@ -13,14 +13,16 @@ module Mutations
     def resolve(input:)
       project = Hmis::Hud::Project.viewable_by(current_user).find_by(id: input.project_id)
 
+      raise 'Not found' unless project.present?
+      raise 'Access denied' if project.present? && !current_user.permissions_for?(project, :can_manage_inventory)
+
+      unit_type = Hmis::UnitType.find_by(id: input.unit_type_id)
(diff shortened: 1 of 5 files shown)


drivers/hmis/app/graphql/mutations/create_units.rb
```diff
@@ -13,14 +13,16 @@ class CreateUnits < BaseMutation
     def resolve(input:)
       project = Hmis::Hud::Project.viewable_by(current_user).find_by(id: input.project_id)
 
+      raise 'Not found' unless project.present?
+      raise 'Access denied' if project.present? && !current_user.permissions_for?(project, :can_manage_inventory)
```
eanders, line 17: For a future iteration.  I wonder if these should be raising some sort of HmisError so that if we wanted to catch them we could, and they could be decorated with additional bits to allow for conditional catching...

eanders (approved):

drivers/hmis/app/graphql/mutations/create_units.rb
gigxz: we raise `HmisErrors::ApiError` when there is additional context that we need to capture conditionally..but yeah maybe we should do that all the time

PR #3835, opened 2023-12-19
186692527 update spm related non hud reports

ttoomey:

## Description
[186692527
](https://www.pivotaltracker.com/story/show/186692527)

Also:
* Remove twenty_twenty_four_warning from views
* update exit destination codes for 2024
* adjust 1a and 1b cols/rows for off-by one from 2021 implementation (C2 => B1)

Notes:
* I adapted  Performance Metrics, CoC Performance Measurement Dashboard, and System Performance Measures by Sub-Population reports. These reports made use of the SpmClient records in 2020; those are not present in 2024 due to refactoring. While the reports run, **I have low confidence that I translated them correctly to use the new structures** (SpmClient/Return/Episode). There are no existing tests or any written specs for these reports.
* After discussion with Elliot, we decided to use the HUD data directly rather than denormalizing more values onto the Spm report records. While the HUD records are not stable, they are unlikely to change while the report is running
* The project scorecard interface seems broken, the controller expects params that are not in the form. I hacked it to get it to run.
 

## Type of change


drivers/performance_metrics/app/models/performance_metrics/report.rb
```diff
@@ -607,13 +607,18 @@ def report_labels
         spm_returners = answer_clients(spm_report, '2', 'I7')
         # M2 I7 is Total Number of Persons who Exited to a Permanent Housing Destination (2 Years Prior)
         spm_leavers = answer_clients(spm_report, '2', 'B7')
+
+        # FIXME, the leavers/returners (universe_members) are SpmEnrollments in 2024; they used to be SpmClients.
+        # We probably need to to unpack enrollment.episodes to get equiv of m1a_es_sh_th_days and m2_reentry_days.
```
ttoomey: I might need some advice on how to get the equivalent fields from SPM Enrollment

drivers/performance_metrics/app/models/performance_metrics/report.rb
```diff
@@ -676,20 +682,15 @@ def report_labels
 
     private def run_spm
       # puts 'Running SPM'
-      # Looking for SPM Measure 1A E3
-      # Looking for SPM Measure 2 J7 (total returns to homelessness within 2 years)
-      questions = [
-        'Measure 1',
-        'Measure 2',
-      ]
+      questions = ['Measure 2']
```
ttoomey: doesn't seem like measure 1 is used, comment seems out of date

eanders: Totally possible.

app/views/warehouse_reports/_twenty_twenty_four_warning.haml
eanders, line 1: We may need to leave this in place if we are not updating the performance_measurement report yet.

ttoomey: Yeah, if we want to do this in phases we'd still need to banner on the metrics/measures reports until we fix them. Or we could try and update them all in one pass, either way is fine with me

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -259,6 +259,19 @@ def can_see_client_details?(user)
...
     end
 
+    private def spm_enrollments_from_answer_member(member)
+      case member
+      when HudSpmReport::Fy2023::SpmEnrollment
+        [member]
+      when HudSpmReport::Fy2023::Episode
+        member.enrollments
+      when HudSpmReport::Fy2023::Return
+        [member.return_enrollment, member.exit_enrollment]
```
ttoomey, line 269: maybe this should just be the exit enrollment?

eanders (approved):
@ttoomey thank you so much for this!  It looks great.

eanders:
Oh, looks like there are actually a few tests for this.  I'm holding off on merging until we circle back to see why those are failing.

ttoomey:
> Oh, looks like there are actually a few tests for this. I'm holding off on merging until we circle back to see why those are failing.

I'll check it out, thanks!

ttoomey:
> > Oh, looks like there are actually a few tests for this. I'm holding off on merging until we circle back to see why those are failing.
> 
> I'll check it out, thanks!

@eanders seems like we were didn't connect the client filters (race, age, etc). That should be corrected now in c13e736953b545783be453c61ca4592e1a808f32. However there are still test failures, it looks like we uncommented tests for `hispanic_latinaeo` filters but our sample Client.csv does not include matching clients. Not sure how that sample data was generated; should we change the tests to match the data or vice versa?

app/models/grda_warehouse/service_history_enrollment.rb
```diff
@@ -4,6 +4,10 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# = GrdaWarehouse::ServiceHistoryEnrollment
+#
+# ServiceHistoryEnrollments flatten Hud Enrollments and related records to serve reporting needs. These records are
+# generated automatically. There is a 1:1 correspondence with Hud Enrollment records
```
eanders, line 10: This is only partially true.  There should be a ServiceHistoryEnrollment with a `record_type=entry` that corresponds to the Enrollment record, one with `record_type=exit` that corresponds to the Exit (if the client has an Exit record) and potentially one with `record_type=first` if the enrollment is the client's first homeless enrollment.  So there is a 1:1 with Enrollment for `GrdaWarehouse::ServiceHistoryEnrollment.entry`

drivers/hud_spm_report/app/models/hud_spm_report/adapters/service_history_enrollment_filter.rb
```diff
@@ -4,6 +4,10 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# = HudSpmReport::Adapters::ServiceHistoryEnrollmentFilter
+#
+# Allow us to use the existing report form to filter SPM reports. Needed since the existing form filters
+# ServiceHistoryEnrollments but the SPM report operates on raw HUD data
```
eanders, line 10: Ah!  Good find.

eanders:
> Referenc

Thanks @ttoomey , I uncommented those since they said they _should_ be functional once we get to the FY2023 SPM.  I'll take a look and see if there's an easy fix.  Probably just need to update the source data.  If all the other tests are passing, that's excellent!

PR #3839, opened 2023-12-19
Load static forms on deploy

gigxz:

## Description

`SeedMaker` function needed to be updated to load the static forms. 

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/db/seed_maker.rb
+++ b/db/seed_maker.rb
@@ -382,6 +382,8 @@ class SeedMaker
       builder.seed_record_form_definitions
       # Load ALL the latest assessment definition froms JSON files.
       builder.seed_assessment_form_definitions
+      # Load admin forms (not configurable)
+      builder.seed_static_forms
       # In development, create the initial instances for occurrence-point collection.
       builder.create_default_occurrence_point_instances! if Rails.env.development?
     end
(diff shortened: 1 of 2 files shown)


db/seed_maker.rb
```diff
@@ -382,6 +382,8 @@ def load_hmis_data
       builder.seed_record_form_definitions
       # Load ALL the latest assessment definition froms JSON files.
       builder.seed_assessment_form_definitions
+      # Load admin forms (not configurable)
+      builder.seed_static_forms
       # In development, create the initial instances for occurrence-point collection.
       builder.create_default_occurrence_point_instances! if Rails.env.development?
     end
```
gigxz, line 389: Alternatively we could replace this block with 
```
Rails.application.load_tasks
Rake::Task['driver:hmis:seed_definitions'].invoke
```

That seems preferable to me, only 1 thing to maintain, but the load_tasks seems expensive and I'm not sure if it's ok to do it here

eanders: I think I prefer this.  Not sure why we'd invoke a rake task here if we have the single method that accomplishes what we need.

eanders (approved):

db/seed_maker.rb
gigxz: The logic is duplicated between this file and the rake task (lines 379-389)

PR #3841, opened 2023-12-20
Use cleaned user_id for last user from versions

gigxz:
## Description

`load_last_user_from_versions` is used to resolve the user that most recently touched the given record. use the new method that provides a "cleaned" version of the numeric user id for the true user.
## Type of change

### DIFF
--- a/drivers/hmis/app/graphql/types/base_object.rb
+++ b/drivers/hmis/app/graphql/types/base_object.rb
@@ -62,7 +62,7 @@ module Types
       latest_version = versions.last # db-ordered so we choose the last record
       return unless latest_version
 
-      last_user_id = latest_version.true_user_id || latest_version.clean_user_id
+      last_user_id = latest_version.clean_true_user_id || latest_version.clean_user_id
       return unless last_user_id
 
       load_ar_scope(scope: Hmis::User.with_deleted, id: last_user_id)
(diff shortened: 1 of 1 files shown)


drivers/hmis/app/graphql/types/base_object.rb
```diff
@@ -57,10 +57,13 @@ def load_last_user_from_versions(object)
...
         order(:created_at, :id).
-        select(:id, :whodunnit, :item_id, :item_type) # select only fields we need for performance
+        select(:id, :whodunnit, :item_id, :item_type, :user_id, :true_user_id) # select only fields we need for performance
       versions = load_ar_association(object, :versions, scope: refinement)
-      last_user_id = versions.last&.whodunnit # db-ordered so we choose the last record
-      return unless last_user_id && last_user_id =~ /\A[0-9]+\z/
+      latest_version = versions.last # db-ordered so we choose the last record
+      return unless latest_version
+
+      last_user_id = latest_version.true_user_id || latest_version.clean_user_id
```
eanders: Where is `clean_user_id` coming from?  I don't think that's a column on the versions table (I could be wrong) and I'm not seeing that method anywhere in the codebase (I am on a different branch.)

gigxz: It's a more recent method added to GrdaWarehouse::Version to deal with coalescing the various ids (`whodunnit` as id string, `whodunnit`  as `x as y` string, and `user_id` column)

gigxz: I meant to write `clean_true_user_id`, whoops

gigxz: I'd be happy to incorporate feedback about those into this PR! The methods are here https://github.com/greenriver/hmis-warehouse/blob/fae92c18aee5efd4290f7158f5fb19dee98162ab/app/models/grda_warehouse/version.rb#L19-L36

eanders (approved):

PR #3853, opened 2023-12-21
186557580 json form editor

ZachCMP:

## Description

- CRUD mutations for FormDefinition
- Query for getting FormDefinitionJson for arbitrary JSON
- Updates to form JSON validation to support use outside of rake tasks
- New form definition and role for FormDefinitions

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/593

## Type of change


drivers/hmis/app/graphql/mutations/delete_form_definition.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    field :form_definition, Types::Forms::FormDefinition, null: true
+
+    def resolve(id:)
+      raise 'not allowed' unless current_user.can_configure_data_collection?
+
+      definition = Hmis::Form::Definition.find_by(id: id)
+      raise 'not found' unless definition
+
+      definition.destroy!
```
gigxz, line 20: Can we add a check here to disallow deleting definitions that are referenced by any rules? This should just help ensure that things arent deleted by accident

```
raise 'can't delete definition with active rules' if definition.instances.active.exists?
```

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -0,0 +1,40 @@
...
+      definition.assign_attributes(**input.to_attributes)
+
+      errors = HmisErrors::Errors.new
+      ::HmisUtil::JsonForms.new.tap do |builder|
+        builder.validate_definition(definition.definition, on_error: ->(err) { errors.add(:definition, message: err) })
+      end
+
+      return { errors: errors } if errors.present?
+
+      # definition.assign_attributes(**attrs)
```
gigxz: can remove?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -209,6 +209,21 @@ def static_form_definition(role:)
...
+      json = JSON.parse(input)
+      errors = []
+      ::HmisUtil::JsonForms.new.tap do |builder|
+        builder.validate_definition(json, on_error: ->(err) { errors << err })
+      end
+
+      return { errors: errors, definition: nil } if errors.present?
+
+      return { errors: [], definition: json }
+    end
```
gigxz: This *feels* to me like it should be a Mutation (`validateFormDefinitionJson` or `parseFormDefinitionJson`), since it's not actually querying any data. (It's not mutating any either, though, so I could see an argument either way).

Was there a particular reason you decided to go with a query over a mutation?

The name `form_definition_for_json` feels a bit off to me, I think `parse` or `validate` would be clearer about what it's actually doing.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -66,6 +67,7 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   ].freeze
 
   validates :role, inclusion: { in: FORM_ROLES.map(&:to_s) }
+  validates :identifier, uniqueness: true
```
gigxz: Can we switch this to validate uniqueness of identifier+version? As discussed in the future there will be ability to create new versions of the "same" definition

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -203,19 +205,19 @@ def self.find_definition_for_service_type(service_type, project:)
   end
 
   # Validate JSON definition when loading, to ensure no duplicate link IDs
-  def self.validate_json(json, valid_pick_lists: [])
+  def self.validate_json(json, valid_pick_lists: [], on_error: ->(err) { raise err })
```
gigxz: I haven't see this kind of pattern used in ruby much, it seems very JS-ish. Why not do a `rescue` to track the failures instead? We could define a specific error type on the Definition.

gigxz (commented):
Looks good!

drivers/hmis/app/models/hmis/form/definition.rb
ZachCMP: I think rescue will only be hit for the first error, I want to collect ALL of the errors, not just the first one. Maybe a more ruby-ish pattern would be to `yield` errors, which would let the caller of the validate method decide how to handle them. That's basically what the this is doing, but`yield`ing would be more ruby-ish I suppose.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
ZachCMP: Yeah, it's a bit borderline. I decided to go with a query because it feels like we're asking the backend for information –"Is this JSON a valid form definition? If so, please process it and give me the full definition schema"–rather than telling it to change anything. I think it feels mutation-y because we're asking it to transform something, but that something is just the `input` we passed in, which feels like a valid thing to have a query do as well. Also, since the input is JSON, it feels a bit like a POST request, which typically aligns with mutations, so I think that makes it feel mutation-y as well.

I do think it makes more sense as a query given that it's really just asking for information. Regardless I can make it a mutation instead if you think that's better, there's no mechanical reason it couldn't be a mutation.

I'd be fine with changing the name, I agree it's awkward. Maybe we could go with `parsed_form_definition` if we keep it a query? If we switch to mutation, I think `parse_form_definition` would be fine.

drivers/hmis/app/models/hmis/form/definition.rb
ZachCMP: Adjusted this so that it will `yield` errors if a block is given, and will `raise` when it encounters an error if no block is given

ZachCMP:
@gigxz Addressed your feedback. I left the form parsing as a query for now, let me know if you want me to change it to a mutation

gigxz (approved):

PR #3854, opened 2023-12-22
Update MA EOHHS report

gigxz:

## Description

* Update MA YYA report to pull from HMIS data for zip code, flex fund services, and language.
* Update `direct_assistance` calculation to include any client who received a 16 Event or any Flex Funds during *any* Enrollment in the scope
* Update Flex-funds-related measures to include any Flex Funds received during report range, even if there is no CE Event 16
* Fix bug in LGBTQ definition for `D4b`
* Use HudUtility for situation integers (no change in logic)
* Fix bug in incorrect `NativeHIOtherPacific` naming

[185796953](https://www.pivotaltracker.com/n/projects/2591838/stories/185796953)

## Type of change
- [x] Bug fix


drivers/ma_yya_report/app/models/ma_yya_report/universe_calculator.rb
```diff
@@ -147,15 +147,8 @@ def calculate(&post_processor)
...
-        'AmIndAKNative' => 1,
-        'Asian' => 2,
-        'BlackAfAmerican' => 3,
-        'NativeHIOtherPacific' => 4,
-        'White' => 5,
-        'HispanicLatinaeo' => 6,
-        'MidEastNAfrican' => 7,
-      }.freeze
+      # NativeHIOtherPacific => NativeHIPacific -- ok?
+      HudUtility2024.race_id_to_field_name.excluding(8, 9, 99).invert
```
gigxz: Is this correct? one value changed as noted in the comment

gigxz:
Still in draft mode because I haven't tried running yet, but would appreciate a look at the logic. I didn't think I needed to make any changes to YouthEducationStatus because it looks like it's already pulling it from the normal `YouthEducationStatus` table.

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
```diff
@@ -119,7 +119,7 @@ def describe_filter_as_html
               )),
         A5d: a_t[:direct_assistance].eq(true).
           and(Arel.sql(
-                json_contains_text(:flex_funds, 'Rent arrears'),
+                json_contains_text(:flex_funds, 'Rent arrears'), # Change casing? Collected as 'Rent Arrears'
```
eanders: Maybe better yet, we should update
```
private def json_contains_text(field, text)
   "#{field} ? '#{text}'"
end
```
to something that is case insensitive.  (I haven't found it yet though...)

drivers/ma_yya_report/app/models/ma_yya_report/universe_calculator.rb
eanders: That look like a bug fix to me.  This gets called via the `race` method and is looking at a destination client, so should be using the new columns.

drivers/ma_yya_report/app/models/ma_yya_report/universe_calculator.rb
```diff
@@ -196,19 +189,56 @@ def calculate(&post_processor)
...
+
+      hmis_personal_ids = hmis_source_clients_for(client).pluck(:personal_id)
+      # IDs for all Flex Fund services rendered to the client(s)
+      flex_funds_service_ids = Hmis::Hud::CustomService.hmis.where(personal_id: hmis_personal_ids, service_type: flex_funds_service_type).pluck(:id)
+      # CustomDataElements tied to these CustomServices indicate which fund types were provided
+      flex_funds_types_cded.values.
+        where(owner_id: flex_funds_service_ids).
+        pluck(:value_string).uniq.
+        # Remove parenthesized text from "Move-in (Including first, last and security)"
+        map { |ff_type| ff_type.gsub(/\([^()]*\)/, '').strip }.uniq
```
eanders: This is missing the date range limitation `within_range(@filter.start_date .. @filter.end_date).` (we should only be looking at flex funds within the report dates, and I _think_ we can save at least one query, by making that a `select` instead of `pluck` so it becomes a subquery.  They are in the same database so that _should_ be safe.

drivers/ma_yya_report/app/models/ma_yya_report/universe_calculator.rb
```diff
@@ -196,19 +189,56 @@ def calculate(&post_processor)
...
     end
 
+    # Valid options are [nil, "English", "Spanish", "Other"]
     private def language(client)
-      client.source_hmis_clients.map(&:primary_language).detect(&:present?)
+      client.source_clients.map do |source_client|
+        # Choose the most recent enrollment that has language information. Sort by entry date since language is collected at entry.
+        enrollment = source_client.enrollments.where.not(translation_needed: nil).order(:entry_date).last
```
eanders: Should this be limited to enrollments visible to the user?  Feels slightly odd to pull from all of a client's enrollments, but given the data returned, maybe that's ok?

drivers/ma_yya_report/app/models/ma_yya_report/synthetic/youth_education_status.rb
```diff
@@ -5,6 +5,7 @@
 ###
 
 module MaYyaReport::Synthetic
+  # Remove? Likely not used since HMIS transition
```
eanders: I think that's true, removing it may be somewhat involved... not sure.

eanders (changes requested):
See inline for some questions/possible adjustments.

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
gigxz: Oh. I didn't actually look at `json_contains_text` -- another bug is that it looks like currently "Rent arrears" would be scooped up by the "Rent" category, even though they are two separate categories of flex funds.

drivers/ma_yya_report/app/models/ma_yya_report/universe_calculator.rb
gigxz: I don't know.. it doesn't look like the Flex Funds were ever limited by enrollment visibility, so I don't think this should be either, but I can make a note to clarify next time.

eanders: It probably doesn't matter, given the scope of the report, but if you could drop a note that it's not limited, that seems like a good middle ground.  Thanks!

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
```diff
@@ -197,7 +197,8 @@ def describe_filter_as_html
 
         D4a: a_t[:pregnant].eq(1).and(a_t[:due_date].gt(report_start_date)).
           or(Arel.sql(custodial_parent_query)).and(a_t[:head_of_household].eq(true)),
-        D4b: a_t[:sexual_orientation].in([2, 3, 4, 5]).or(a_t[:gender].in([5, 6])).and(a_t[:head_of_household].eq(true)),
+        # Why did D4b define LGBTQ differently from G3a? It was including 8/9 gender responses as LGBTQ
+        D4b: lgbtq_query.and(a_t[:head_of_household].eq(true)),
```
gigxz: @awisema or @eanders , any idea whether it was purposeful that D4b (_Number of YYA served who were LGBTQ+_) and G3a (_Number of YYA served who were LGBTQ+_ for youth who transitioned into stabilized housing) had different definitions of LGBTQ? 

Looks like a bug to me. D4b previously included gender "6" which which usually means Questioning, but we overload it in the universe_calculator to include "Doesn't know" and "prefers not to answer" as well.

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
```diff
@@ -214,9 +215,10 @@ def describe_filter_as_html
         F1a: a_t[:subsequent_current_living_situations].not_eq([]).and(a_t[:followup_previous_period].eq(false)),
         F1b: a_t[:followup_previous_period].eq(false).
           and(Arel.sql(
-                json_contains(:subsequent_current_living_situations,
-                              [215, 206, 207, 225, 204, 205, 329, 314, 332, 336, 335, 435, 410, 421, 411]),
-              )),
+            # Add note explaining why 302 is excluded? (Transitional housing)
+            json_contains(:subsequent_current_living_situations,
+                          HudUtility2024.permanent_situations(as: :current) + HudUtility2024.temporary_situations(as: :current) + HudUtility2024.institutional_situations(as: :current) - [302]),
+          )),
```
gigxz: Updated this to use the HudUtility helper methods rather than raw numbers. Was `302` excluded purposefully or was that a mistake?

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
```diff
@@ -259,6 +261,14 @@ def label(key)
       end
     end
 
+    private def lgbtq_query
+      # Report defines LGBTQ as:
+      #   SexualOrientiation = gay(2), lesbian(3), bisexual(4), questioning(5), OR
+      #   Gender = transgender(5)
+      # Is it intentional that "other" SOs, and NB gender, are not considered LGBTQ for the report?
```
gigxz: Wondering if there is background on this, or we can clarify. I didn't make any change to the logic as it was defined for `G3a`.

drivers/ma_yya_report/app/models/ma_yya_report/report.rb
```diff
@@ -214,8 +214,7 @@ def describe_filter_as_html
         F1a: a_t[:subsequent_current_living_situations].not_eq([]).and(a_t[:followup_previous_period].eq(false)),
         F1b: a_t[:followup_previous_period].eq(false).
           and(Arel.sql(
-                json_contains(:subsequent_current_living_situations,
-                              [215, 206, 207, 225, 204, 205, 329, 314, 332, 336, 335, 435, 410, 421, 411]),
+                json_contains(:subsequent_current_living_situations, HudUtility2024.permanent_situations(as: :current) + HudUtility2024.temporary_situations(as: :current) + HudUtility2024.institutional_situations(as: :current) - [302]),
```
awisema:
```suggestion
                json_contains(:subsequent_current_living_situations, HudUtility2024.permanent_situations(as: :current) + HudUtility2024.temporary_situations(as: :current) + HudUtility2024.institutional_situations(as: :current) - [302]), # Excludes 302: Transitional housing for homeless persons (including homeless youth) 
```

awisema (approved):

eanders (approved):

config/initializers/inflections.rb
```diff
@@ -17,6 +17,7 @@
 ActiveSupport::Inflector.inflections(:en) do |inflect|
   inflect.irregular 'census', 'censuses'
   inflect.irregular 'supplemental_enrollment_data', 'supplemental_enrollment_data'
+  inflect.irregular 'hmis', 'hmis'
```
gigxz: It looks like this is causing some weird test failures

gigxz: or maybe its the extensions https://github.com/greenriver/hmis-warehouse/actions/runs/7389041046/job/20101087304

PR #3863, opened 2023-12-28
186589424 user access log v2

ttoomey:

## Description
[User access log v2 #186589424](https://www.pivotaltracker.com/story/show/186589424)

[Front-end PR](https://github.com/greenriver/hmis-frontend/pull/599)

Build associations to clients, enrollments, and projects from resolved entities in the access logs
* cron job runs every hour to process new access logs
* job populates join tables for each entity type
* rollup "summary" views are built to report on activity in aggregate

Resolvers for per-user activity logs for enrollments and clients, including search by user id and date.

## Type of change


config/schedule.rb
```diff
@@ -173,6 +174,12 @@
     at: '11:00am',
     interruptable: true,
   },
+  {
+    task: 'driver:hmis:process_activity_logs',
+    frequency: 60.minutes,
+    trigger: hmis_trigger,
+    interruptable: true,
+  },
```
eanders: Instead of adding a new task, can this be added to `task: 'grda_warehouse:hourly'`?  Each of these tasks spins up a new container which requires additional capacity and potentially adds pressure to the existing DB that is outside of what's expected.  If it doesn't need to run exactly each hour, you can add it to the hourly task and have it queue up a delayed job that will be processed when the worker gets there.  Sometimes those also have to spin up an extra container, but they are limited in the number of containers they'll spin up, so the DB use should be somewhat limited.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
```diff
@@ -0,0 +1,153 @@
...
+        joins(
+          e_t.create_join(
+            c_t,
+            e_t.create_on(
+              # joins to clients; intentionally includes deleted records
+              [e_t[:PersonalID].eq(c_t[:PersonalID]), e_t[:data_source_id].eq(c_t[:data_source_id])].inject(&:and),
+            ),
+          ),
+        ).
+        pluck(e_t[:id], c_t[:id]).to_h
```
eanders: Why all this complexity instead of `Hmis::Hud::Enrollment.with_deleted.joins(:client).merge(Hmis::Hud::Client.with_deleted).pluck(e_t[:id], c_t[:id])`?  On the warehouse side we also have built out `GrdaWarehouse::Hud::WithDeleted::Enrollment` etc. models, might be worth exploring here.  I'm somewhat opposed to these manual joins since we already encapsulate the logic in the relations, and it's not a single column join.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
```diff
@@ -0,0 +1,153 @@
...
+          joins(
+            cas_t.create_join(
+              e_t,
+              cas_t.create_on(
+                # joins to enrollments; intentionally includes deleted records
+                [
+                  e_t[:EnrollmentID].eq(cas_t[:EnrollmentID]),
+                  e_t[:PersonalID].eq(cas_t[:PersonalID]), # this seems redundant
+                  e_t[:data_source_id].eq(cas_t[:data_source_id]),
+                ].inject(&:and),
```
eanders: same here, I'm pretty sure a merge with the `with_deleted` scopes will give you the same result.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
```diff
@@ -0,0 +1,153 @@
...
+
+      table = Arel::Table.new(table_name)
+      manager = Arel::InsertManager.new
+      manager.into(table)
+      manager.columns << table[:activity_log_id]
+      manager.columns << table[:"#{references}_id"]
+      manager.values = manager.create_values_list(rows)
+
+      connection.execute manager.to_sql
+    end
```
eanders: This seems really complicated for a bulk insert, can we not use ActiveRecordImport? If this is significantly better performance, and we believe that's an expected issue, then fine, but I'd rather build up the array/hash of values and just call `klass.import!(hashes)`

drivers/hmis/app/models/hmis/activity_log.rb
```diff
@@ -4,7 +4,25 @@
...
+    jt = Arel::Table.new(:hmis_activity_logs_clients)
+    join_clause = arel_table.create_join(jt, arel_table.create_on(jt[:activity_log_id].eq(arel_table[:id])))
+    joins(join_clause).select(jt[:client_id])
+  end
+
+  def self.select_enrollment_ids
+    jt = Arel::Table.new(:hmis_activity_logs_enrollments)
+    join_clause = arel_table.create_join(jt, arel_table.create_on(jt[:activity_log_id].eq(arel_table[:id])))
+    joins(join_clause).select(jt[:enrollment_id])
+  end
```
eanders, line 32: This also feels like we're rebuilding some active record.  Is there a reason we're avoiding using relations?  If so, let'd document why.

drivers/hmis/app/models/hmis/enrollment_access_summary.rb
```diff
@@ -0,0 +1,47 @@
...
+        scope = scope.where(enrollment_id: log_scope.select_enrollment_ids)
+      end
+      if search_term.present?
+        enrollment_ids = Hmis::Hud::Client.with_deleted.
+          matching_search_term(search_term).
+          joins(
+            c_t.create_join(
+              e_t,
+              c_t.create_on(
+                # joins to enrollments; intentionally includes deleted records
```
eanders: Same, here, I think `merge(...with_deleted)` should do the same thing.

drivers/hmis/lib/tasks/activity_logs.rake
```diff
@@ -0,0 +1,4 @@
+desc 'expose log processor job for cron-job'
+task process_activity_logs: [:environment] do
+  Hmis::ActivityLogProcessorJob.perform_now
```
eanders, line 3: Oh, excellent that this is a DJ already, should be easy enough to move to `grda_warhouse:hourly`

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
ttoomey: That's fair. I'll try using merge() as you suggest. I'm not sure about adding models or associations just support access to deleted rows.

It's worth mentioning that I did figure out a way to globally disable the paranoia gem via an env variable. I reverted it since it felt kind of heavy just for this use-case.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
ttoomey: There's no active record class for these join tables; I assume ActiveRecordImport won't work for that. I guess I could add classes and a primary key to those tables but that feels like a lot just to support a different insert syntax. 

It seems like we use Arel pretty extensively in this app but perhaps there are parts of its API we want to avoid?

drivers/hmis/app/models/hmis/activity_log.rb
ttoomey: I'll add some documentation. These join tables span the app and warehouse database boundary, that's why I avoided using active record associations here. Active record associations seem like they would be hard to use without accidentally creating a cross-db join.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
eanders: Ah, I'd forgotten there were no classes to back the tables.  I guess if we don't have those or need them, this is fine.  We do make heavy use of Arel, but this is a new part of the API we haven't used, so it felt like something that we must be handling with ActiveRecord.  Overall our approach has been, use ActiveRecord until you can't, then use Arel until you can't or it becomes unwieldy, then use SQL.

drivers/hmis/app/models/hmis/activity_log.rb
eanders: Are you getting the cross-db join to actually work?!?  That'd be great.  Active record hasn't shown an ability to do that I'm aware of, but we have a handful of places it would be really useful.  We do have relations that cross DB boundaries, but you can't do things like `User.joins(:clients)` because of the cross-db nature of that.  Our approach, historically, has been to pluck IDs and force two queries.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
eanders: I'll take what you have over globally disabling paranoia, but I'm pretty sure the merge will work.

drivers/hmis/app/models/hmis/activity_log.rb
ttoomey: No, sorry I didn't mean to imply I got cross-db joins working. I didn't try it, although supposedly it is possible. I fetch the ids down as you describe to bridge the boundary.

I did see Rails 7 supports `disable_joins` on associations which seems to take care of the fetch/stitch operation for you automatically

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
ttoomey: Looks like merging with_deleted doesn't work as we might expect. Note the below sql is still filtering clients by DateDeleted which isn't what we want. (I think I'm invoking it correctly but I might be missing something)
```ruby
puts  Hmis::Hud::Enrollment.with_deleted.joins(:client).merge(Hmis::Hud::Client.with_deleted).select(e_t[:id]).to_sql
# => SELECT "Enrollment"."id" FROM "Enrollment" INNER JOIN "Client" ON "Client"."DateDeleted" IS NULL AND "Client"."data_source_id" = "Enrollment"."data_source_id" AND "Client"."PersonalID" = "Enrollment"."PersonalID"
```

Interestingly, it looks like the merged condition is applied to enrollment instead of client, so the following statements result in equivalent sql
```ruby
Hmis::Hud::Enrollment.joins(:client).merge(Hmis::Hud::Client.with_deleted).select(e_t[:id])
Hmis::Hud::Enrollment.with_deleted.joins(:client).select(e_t[:id])
# => true
```

I suspect this is because the paranoia gem is playing games with default_scope under the hood which seems gnarly.

eanders: This seems to work:
```
Hmis::Hud::Client.unscoped { Hmis::Hud::Enrollment.joins(:client).merge(Hmis::Hud::Client.with_deleted).to_sql }
# SELECT "Enrollment".* FROM "Enrollment" INNER JOIN "Client" ON "Client"."data_source_id" = "Enrollment"."data_source_id" AND "Client"."PersonalID" = "Enrollment"."PersonalID"
```

It's not the most obvious solution.  Looks like:
```
Hmis::Hud::Client.unscoped { Hmis::Hud::Enrollment.with_deleted.joins(:client).to_sql }
# SELECT "Enrollment".* FROM "Enrollment" INNER JOIN "Client" ON "Client"."data_source_id" = "Enrollment"."data_source_id" AND "Client"."PersonalID" = "Enrollment"."PersonalID"
```
also works, so maybe that's a better/simpler implementation.  Basically you can pass `unscoped` a block for a given model and it'll remove the paranoia restriction, but only for **that** model.

eanders: I don't like this as much, but it looks like you can nest the unscoped blocks.
```
Hmis::Hud::Client.unscoped { Hmis::Hud::Enrollment.unscoped { puts Hmis::Hud::Enrollment.joins(:client).to_sql }}
# SELECT "Enrollment".* FROM "Enrollment" INNER JOIN "Client" ON "Client"."data_source_id" = "Enrollment"."data_source_id" AND "Client"."PersonalID" = "Enrollment"."PersonalID"
```

ttoomey: I didn't realize unscoped takes a block, good to know! I'd say that seems reasonable as long as we are confident there's never going to be anything other than paranoia managing the default scope of these models; then we can treat Model.unscoped as equivalent to with_deleted. I'll make that change

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -38,9 +46,23 @@ def recent_items
...
 
       object.activity_logs
     end
+
+    def client_access_summaries(filters: nil)
+      access_denied! unless current_user.can_audit_users?
+
+      Hmis::ClientAccessSummary.order(last_accessed_at: :desc, client_id: :desc).
+        apply_filter(user: object, starts_on: filters&.on_or_after, search_term: filters&.search_term)
+    end
```
gigxz, line 59: I like this filter application mechanism, it's nice and simple

drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    protected
+
+    def enrollment
+      load_ar_association(object, :enrollment, scope: Hmis::Hud::Enrollment.with_deleted)
+    end
+
+    def project
+      load_ar_association(enrollment, :project, scope: Hmis::Hud::Project.with_deleted) if enrollment
```
gigxz: This will be nil for WIP Enrollments. We need to do something similar to how we load projects on the Enrollment type:

```
    def project
      if object.in_progress?
        wip = load_ar_association(object, :wip)
        load_ar_association(wip, :project)
      else
        load_ar_association(object, :project)
      end
    end
```

drivers/hmis/app/models/hmis/enrollment_access_summary.rb
```diff
@@ -0,0 +1,43 @@
...
+    self.primary_key = 'id'
+
+    belongs_to :user, class_name: 'Hmis::User'
+    belongs_to :enrollment, class_name: 'Hmis::Hud::Enrollment'
+
+    def readonly?
+      true
+    end
+
+    def self.apply_filter(user:, starts_on: nil, search_term: nil)
```
gigxz: Can you add a filter for Project IDs? This would be particularly helpful in trying to figure out whether a user has accessed any enrollments at a project that they shouldn't have (if they were granted access in error, for example).

drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
ttoomey: Good catch, thanks! I'll fix taht.

Something to note is that it looks like Hmis::Wip does not use soft-deletion despite having a deleted_at column. Perhaps it should though?

```
> Hmis::Wip.paranoid?
=> false
> Hmis::Wip.column_names.grep /delete/
=> ["deleted_at"]
```

drivers/hmis/app/models/hmis/wip.rb
```diff
@@ -5,6 +5,7 @@
 ###
 
 class Hmis::Wip < Hmis::HmisBase
+  acts_as_paranoid
```
ttoomey, line 8: this was missing, looks like we intended it to be there since the table has a deleted_at column.

lib/tasks/grda_warehouse.rake
```diff
@@ -294,6 +294,13 @@ namespace :grda_warehouse do
     rescue StandardError => e
       puts e.message
     end
+    begin
+      Rake::Task['driver:hmis:process_activity_logs'].invoke if ENV['ENABLE_HMIS_API'] == 'true'
+    rescue StandardError => e
+      puts e.message
+      Sentry.capture_exception(e)
```
ttoomey, line 303: The other tasks just rescue exceptions; I guess there must be another mechanism for reporting these exceptions?

drivers/hmis/app/models/hmis/enrollment_access_summary.rb
ttoomey: Yes I've added it. It was bit complicated though; to avoid having to fetch down all enrollment ids for a filtered project, I added the project_id on to the join table, which is now populated in the activity job

drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
ttoomey: I decided to add soft-deletion the wip model. There's the possibility this causes regressions if we are using the wip model without the default scope somewhere

gigxz: Ok, thanks. We shouldn't be.

drivers/hmis/app/models/hmis/enrollment_access_summary.rb
gigxz: Ok, thanks!

lib/tasks/grda_warehouse.rake
gigxz: I don't know, I had the same question. @eanders ?

lib/tasks/grda_warehouse.rake
```diff
@@ -294,6 +294,13 @@ namespace :grda_warehouse do
     rescue StandardError => e
       puts e.message
     end
+    begin
+      Rake::Task['driver:hmis:process_activity_logs'].invoke if ENV['ENABLE_HMIS_API'] == 'true'
```
gigxz: checking `HmisEnforcement.hmis_enabled?` will check the env variable and that the hmis driver is loaded. So you could use that for consistency.

gigxz (approved):

lib/tasks/grda_warehouse.rake
eanders: I don't have an answer to that.  I don't think they handle anything magically....

ttoomey: I'm going to add calls to sentry for the other cases. If it ends up being too noisy we can always remove it

gigxz (approved):
Looks good, and I tested out the new project filtering!

gigxz:
Lets wait for @eanders before merging

eanders (approved):

drivers/hmis/app/models/hmis/wip.rb
gigxz: @martha identified that this introduced a bug, since the `hmis_client_projects` view doesn't check for deletion status.

@martha it looks like `hmis_households` view also needs the same change, to only look at non-deleted hmis_wips. Can you make a PT story for that one?

martha: sure! Created https://www.pivotaltracker.com/story/show/187289783

PR #3869, opened 2023-12-30
Tt/staggered db pks in tests

ttoomey:

## Description

[Use staggered primary keys for tests](https://www.pivotaltracker.com/story/show/186757111). After this change, it should be unlikely that user.id == group.id when creating new records. The idea would be to catch bugs related to mis-use of ids, for example
```diff
- User.where(id: group.select(:id)
+ User.where(id: group.select(:user_id)
```
The above would pass if user/group ids are in lock-step on a fresh database. By stagger the ids, this type of bug should be easier to find in our tests

I also took the opportunity to roll the existing table truncation statement into utility model as a bit of code cleanup

## Type of change


app/models/grda_warehouse/utility.rb
```diff
@@ -145,6 +150,7 @@ def self.clear!
     nil
   end
 
+  # FIXME: document what this is for
```
eanders:
```suggestion
  # Tool to reset postgres sequences, which occasionally get out of sync when restoring a database, dependent on restore method.
```

eanders (approved):

ttoomey:
It seems like the tests on this branch take a really long time to run (like 3 hours). Some investigation needed

ttoomey:
It looks like the slowness was may be something else. Tests for release-98 are clocking in at 3+ hours

https://github.com/greenriver/hmis-warehouse/actions/runs/7402894996

PR #3872, opened 2024-01-01
185674113 warehouse session expiry

ttoomey:

## Description
[Improve warehouse session expiration handling](https://www.pivotaltracker.com/story/show/185674113)

* replace auto-session-timeout and polling with new implementation that relies on local state and inferring session state from xhr responses

### test case 1
* login the the warehouse, wait until five minutes before your session will expire. You should see a modal with a countdown timer "You will be logged out automatically in X minutes"
* allow the timer to expire. The screen should be cleared and you will see "Your session has expired message"

### test case 2
* login the the warehouse. Open the warehouse in a second tab in the same browser
* sign out in one of the tabs
* the first tab should show "Your session is invalid. You may have signed out in another window."

## Type of change
- [x] Dependency update


eanders:
@ttoomey I can't seem to get this so that calling some paths doesn't bump the `last_request_at`.  Specifically, I'd like to exclude `'/warehouse_reports/hmis_exports/running'` which comes form here: https://hmis-warehouse.dev.test/warehouse_reports/hmis_exports and a bunch of paths from HUD reports like `hud_reports/paths/running` from https://hmis-warehouse.dev.test/hud_reports/paths/history.  I tried adding them to `config/initializers/rack_attack.rb` and restarting, but they still seem to bump the timeout.  Thoughts?  We might even want to just blanket ignore any path that ends in `running` (especially if it's an XHR request.)  

Otherwise this looks great!  Thanks so much.

eanders:
Oh, looks like what I actually want to do is call 
```
def skip_timeout
  request.env['devise.skip_trackable'] = true if request.xhr?
end
```

eanders:
@ttoomey looks like those are already XHR requests, so I _think_ they should prevent the timeout from getting pushed forward, but that doesn't seem to be the case.  

Unrelatedly, I'm getting the following screen fairly often.  I think when an XHR request throws an error.   Usually those manifest as 500s, but only within the section of the page where they are replacing content, so maybe this is an improvement, except it comes up on pages that are open when there's an error on a different tab.

Screenshot 2024-01-02 at 7 51 47 PM

eanders:
@ttoomey I _think_ [this accomplishes what I was looking to do](https://github.com/greenriver/hmis-warehouse/pull/3872/commits/11239ef224bad31270e8461bd757f17ec719838a) does that look like a reasonable solution, or is there a better way to prevent some background refreshes from triggering a session expiration update?

ttoomey:
Adding discussion from slack:
This looks good! The only concern I have is that the rack attack change would allow any user to trivially circumvent the rate limiting we have in place.

Perhaps we can add a catch-all that rate limits everyone to 300 req/min and doesn’t check if you’re logged in. Or we can figure out how to get the current user from warden without triggering the trackable side-effect.

From yesterdays slack convo: Using a parameter to skip trackable seems fine but probably better to use a header like X-skip-trackable

eanders:
@ttoomey pretty sure this handles the case where someone isn't logged in and hits any page that isn't one of the special pages.

```
send(tracker, 'requests per unauthenticated user per ip', limit: 10, period: 1.seconds) do |request|
    if tracking_enabled?(request)
      if !warden_user_present?(request) && !(sign_in_path(request) || hmis_sign_in_path(request) || history_pdf_path(request) || asset_paths(request))
        request.ip
      end
    end
  end
```

That will limit access to 10 requests per second.  
My reading is that `!warden_user_present?(request)` will be `true` (note the `!`).
If one of the special paths is hit, then that limit will apply, otherwise it will default to the 10/second.

ttoomey:
> @ttoomey pretty sure this handles the case where someone isn't logged in and hits any page that isn't one of the special pages.

Good find, so looks like we do have a catch-all already so perhaps this is fine as is

PR #3873, opened 2024-01-02
184440788-indicate-account-locked

ttoomey:

## Description

[Show appropriate message when user account is locked
](https://www.pivotaltracker.com/story/show/184440788)

https://github.com/greenriver/hmis-frontend/pull/600

* for non-okta sign in, display a message to the user if their account is locked
* Use a custom header method to differentiate between errors encountered within the API vs okta sign-in errors

### Testing okta:
* Sign-in. Lock the currently signed in account. Hit any page that makes an api request. Should see an informative error message
* Sign out. Lock account. Attempt to sign in via okta. Should see "account locked message"

### Testing non-okta (email/pass)
* Sign out. Lock account. Attempt to sign with email/password form with valid or invalid password (account must not have okta configured). Should see "account locked message"

## Type of change
- [x] Bug fix


eanders (approved):

gigxz (commented):
What will happen if there is a failure `GET`ting `app_settings` for example? Would this revert that from a JSON response to a HTML response? Seems less than ideal but maybe it's fine because we should be logging it to sentry as an error anyway (probably like an "unable to parse response body").

ttoomey:
@eanders  and @gigxz I'm unsure about the change to the sessions controller. What do you guys think? I couldn't find a better solution within a few hours of digging

drivers/hmis/app/controllers/hmis/sessions_controller.rb
```diff
@@ -22,6 +22,16 @@ def new
 
   # POST /hmis/login
   def create
+    # If the account has been locked, show an appropriate message. Ideally this would happen after the password is
+    # validated however at that point there doesn't seem to be a way to differentiate failures for bad credentials vs
+    # those due to locked accounts without performing a second validation (which could trigger unintended lockouts)
+    #
+    # There is a potential security issue in that this allows account emails to be enumerated
+    if sign_in_params['email'] && sign_in_params['password']
```
eanders: Is this only relevant to Okta?  I feel like the existing warehouse returns the locked message correctly, but maybe that's not true on the HMIS side.  
I guess I don't have a big problem with this, you can already enumerate email addresses by guessing with incorrect passwords.  
We should confirm the `hmis/sessions*` paths are all in rack_attack similarly to how the main login forms are.

ttoomey:
> What will happen if there is a failure `GET`ting `app_settings` for example? Would this revert that from a JSON response to a HTML response? Seems less than ideal but maybe it's fine because we should be logging it to sentry as an error anyway (probably like an "unable to parse response body").

Sorry I missed this. I don't believe you would encounter an auth error for the app_settings endpoint since it's not authenticated. However if we had some other JSON GET end point that required authentication, and authentication failed, the request would be redirected which is probably not ideal for an API request.

Considering it again, maybe it's more robust to use the presence of one of our custom headers rather than the HTTP verb. Looks like the `X-Csrf-Token` can be used as a hint to the server about how best to respond. I adjusted the logic to use that instead. I think the only time we want to redirect is when there's a failure during okta sign-in

drivers/hmis/app/controllers/hmis/sessions_controller.rb
ttoomey: This change is only relevant for email/password logins. Okta authentication doesn't use the sessions controller. Looks like the `hmis_sign_in_path` endpoint is protected by rack attack, just as we do for the warehouse side.

gigxz (approved):

drivers/hmis/app/controllers/hmis/sessions_controller.rb
```diff
@@ -22,6 +22,16 @@ def new
 
   # POST /hmis/login
   def create
+    # If the account has been locked, show an appropriate message. Ideally this would happen after the password is
+    # validated however at that point there doesn't seem to be a way to differentiate failures for bad credentials vs
+    # those due to locked accounts without performing a second validation (which could trigger unintended lockouts)
```
gigxz: Just an FYI of something tangentially related: there is a bug with the devise-two-factor gem that causes the lockout count to be incremented twice from 1 attempt. as a result we always set the # failed attempts as double what it should be. 

https://github.com/greenriver/hmis-warehouse/blob/a99f2c8030b403ed3e2778a5f9918cb477b09762/drivers/hmis/spec/requests/hmis/sessions_controller_spec.rb#L55-L60

ttoomey:
@eanders @gigxz 
Sorry to bring this back up but, on second thought, I think we can address the email enumeration issue with a transaction/rollback. It's kind of hacky though. What do you guys think?

I also updated the session tests to verify authentication success/failure by attempting API requests

ttoomey:
Adjustments made after discussion today

eanders (approved):
Thanks!

PR #3882, opened 2024-01-04
Add Expiration Feature to TalentLMS Trainings - 185215135

dtgreiner:

## Description

Add Expirations to the TalentLMS training course.
PT-185215135

## Type of change
- [ ] Bug fix
- [ ] Dependency update


app/controllers/user_training_controller.rb
```diff
@@ -7,19 +7,26 @@
 class UserTrainingController < ApplicationController
   def index
     config = Talentlms::Config.first
-    if config.nil? || !current_user.training_required? || current_user.training_completed?
+    lms = Talentlms::Facade.new
+
+    # Verifying with local data before hitting the API. This prevents unneeded API calls
+    # and ensures local data is updated when new trainings have been completed.
+    if config.nil? || !current_user.training_required? || (current_user.training_completed? && !lms.training_expired?(current_user, false))
```
eanders: I wonder if we should encapsulate 
```
!current_user.training_required? || (current_user.training_completed? && !lms.training_expired?(current_user, false)
```
in `lms.training_required?(current_user)`  or maybe `current_user.training_required?(lms)`.  I think we're always going to check all three items together, so this (and similar checks could just be):

```suggestion
    if config.nil? || !current_user.training_required?(lms)
```

app/models/talentlms/facade.rb
```diff
@@ -61,19 +61,58 @@ def enroll(user, course_id)
...
+    #
+    # @param user [User] the user
+    # @param completion_date [Date] the date which the training was completed
+    # @param course_id [Integer] the id of the course
+    # @return [CompletedTraining] the completed training data
+    def log_course_completion(user, completion_date, course_id = @api.courseid)
+      login = Login.find_by(user: user)
+      return nil if login.nil?
+
+      return CompletedTraining.where(login: login, config: @api, course_id: course_id, completion_date: completion_date).first_or_create
```
eanders: Can't believe rubocop didn't complain about this, but `return` is unnecessary.
```suggestion
      CompletedTraining.where(login: login, config: @api, course_id: course_id, completion_date: completion_date).first_or_create
```

app/models/talentlms/facade.rb
```diff
@@ -61,19 +61,58 @@ def enroll(user, course_id)
...
+    # @param verify_with_api [Boolean] call the API for the last completed date or use local data
+    # @return [Boolean] true if the user's training has expired
+    def training_expired?(user, verify_with_api = true)
+      # Recertification is only required when a value has been set for months_to_expiration
+      return false unless @api.months_to_expiration.present?
+
+      completed_on = verify_with_api ? complete?(user, @api.courseid) : user.last_training_completed
+      return false if completed_on.nil?
+
+      return (completed_on.to_date + @api.months_to_expiration.months).past?
```
eanders:
```suggestion
     (completed_on.to_date + @api.months_to_expiration.months).past?
```

app/models/talentlms/facade.rb
```diff
@@ -61,19 +61,58 @@ def enroll(user, course_id)
...
+    #
+    # @param user [User] the user
+    # @param verify_with_api [Boolean] call the API for the last completed date or use local data
+    # @return [Boolean] true if the user's training has expired
+    def training_expired?(user, verify_with_api = true)
+      # Recertification is only required when a value has been set for months_to_expiration
+      return false unless @api.months_to_expiration.present?
+
+      completed_on = verify_with_api ? complete?(user, @api.courseid) : user.last_training_completed
+      return false if completed_on.nil?
```
eanders: This is probably fine, but generally, it's better to check `.blank?` over `.nil?` unless you are trying to identify the difference between `''` and `nil` since `''` will return false for a `.nil?` check.

`.blank?` will return `true` for `''`, ` `, `{}`, `[]`, and `nil` (probably others too.)

db/reporting_structure.sql
eanders, line 1: Pretty sure this file shouldn't be changed here.

db/structure.sql
eanders, line 1: This file shouldn't have changes either given the migrations.

eanders (commented):
Looks great, see inline comments.  Thank you!

app/controllers/user_training_controller.rb
dtgreiner: I've added this as a method in the lms Facade class. I've kept the ability to specify if we want to hit the API as the original logic is using these checks to see if the local data needs to be updated.

app/models/talentlms/facade.rb
dtgreiner: the return here has been removed

app/models/talentlms/facade.rb
dtgreiner: the return here has been removed

app/models/talentlms/facade.rb
dtgreiner: I've updated this to .blank?

db/reporting_structure.sql
dtgreiner: This file has been reverted

db/structure.sql
dtgreiner: This file has been reverted

app/models/talentlms/facade.rb
```diff
@@ -61,19 +61,70 @@ def enroll(user, course_id)
...
+    #
+    # @param user [User] the user
+    # @param verify_with_api [Boolean] call the API for the last completed date or use local data
+    # @return true if the user requires training
+    def training_required?(user, verify_with_api = true)
+      return unless user.training_required?
+      return true unless user.training_completed?
+
+      training_expired?(user, verify_with_api)
+    end
```
eanders: Is this ever called with verify_with_apy = true?  or maybe more to the point, is there any reason we would check `training_expired?` on line 125 without asking it to go out and check?

dtgreiner:
@eanders The requested changes should be good to go in this PR

eanders (approved):
Thanks @dtgreiner !  Will merge after tests complete.

PR #3883, opened 2024-01-04
Transformation to display audit history for disability correctly

gigxz:
## Description

Fixes a bug where Disability Response "Yes" was always showing up in the Audit UI as "Alcohol Use Disorder", even for non-Substance Use disability types.

This approach uses a new enum that is a union of 1.8 and 4.10.2.
Screenshot 2024-01-04 at 10 24 29 AM

It also includes a change in the Audit Event resolver for Enrollment-related records, in order to override the "1" so that it will be displayed using the correct enum label. This could have been done on the frontend instead, but this felt cleaner because the all of the logic is contained to the backend (As opposed to spread across both).

Screenshot 2024-01-04 at 10 25 53 AM

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/enums/complete_disability_response.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/enums/complete_disability_response.rb
@@ -1,0 +1,27 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::Enums::CompleteDisabilityResponse < Types::BaseEnum
+    graphql_name 'CompleteDisabilityResponse'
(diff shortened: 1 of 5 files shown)


drivers/hmis/app/graphql/types/hmis_schema/enums/complete_disability_response.rb
```diff
@@ -0,0 +1,27 @@
...
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::Enums::CompleteDisabilityResponse < Types::BaseEnum
+    graphql_name 'CompleteDisabilityResponse'
+    # Used for Disability.DisabiltiyResponse field. It is a combination of DisabilityResponse+NoYesMissing enums,
+    # with a special fake value for ALCOHOL_USE_DISORDERS which is the only one that overlaps with a different label.
+
+    SUBSTANCE_USE_1_OVERRIDE_VALUE = 10
```
ttoomey: maybe safer to use something like 1000 that isn't contiguous with the HUD values?

ttoomey (approved):
Looks good!

eanders (approved):

PR #3891, opened 2024-01-08
Add root enabledFeatures query

gigxz:
## Description

Simple query that resolves the list of "features" that are enabled for the client dashboard.

This is going to need some tweaking to support the more complex configurability that we'll need. But for now this is a simple approach to deciding whether to show the read-only Case Notes on the Client Dashboard.

For installations that don't use Case Notes anywhere, this will hide the Case Notes tab from the Client dash.

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/603

## Type of change

### DIFF
--- a/drivers/hmis/app/graphql/types/forms/enums/client_dashboard_feature.rb
+++ b/drivers/hmis/app/graphql/types/forms/enums/client_dashboard_feature.rb
@@ -1,0 +1,16 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class Forms::Enums::ClientDashboardFeature < Types::BaseEnum
+    graphql_name 'ClientDashboardFeature'
+
+    value 'CASE_NOTE'
+    value 'FILE'
+  end
+end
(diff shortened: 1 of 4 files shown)


drivers/hmis/app/graphql/types/forms/enums/client_dashboard_feature.rb
```diff
@@ -0,0 +1,16 @@
+###
+# Copyright 2016 - 2023 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
```
ttoomey: I think this magic comment doesn't do anything unless it's the very first line in the file

ttoomey (approved):
Looks good! This seems fine for a small number of feature flags.

If the features map cleanly into components, a feature-list approach might work well too:

```graphql
client {
  features {
    ...FeatureFields
  }
}
```

```jsx
<ClientDash>
{client.features.map(({__typename, ...otherProps}) => {
  switch(__typename) {
    case 'CaseNoteFeature': 
      return <CaseNoteList client={client} {...otherProps}/>
  }
})}
</ClientDash>
```

drivers/hmis/app/graphql/types/forms/enums/client_dashboard_feature.rb
gigxz: oh ok, that was maybe copy pasta, I don't think we need it here

PR #3901, opened 2024-01-11
186658870 post exit assessment r20

ttoomey:

## Description
* add post-exit assessment
* migration to create synthetic post-exit assessments
* move logic to determine eligible assessments to API
* add misc documentation

[Support Post-Exit assessment for R20 Aftercare
](https://www.pivotaltracker.com/story/show/186658870)

[Front-end PR](https://github.com/greenriver/hmis-frontend/pull/607)

## Type of change


drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -36,6 +36,9 @@ def self.configuration
     description 'HUD Client'
     field :id, ID, null: false
     field :lock_version, Integer, null: false
+    field :assessment_eligibilities, [HmisSchema::AssessmentEligibility], null: false do
+      argument :enrollment_id, ID, required: true
+    end
```
ttoomey: This could also go on enrollment. Having this on client leaves open the possibility that we might not have an enrollment but want to query assessments are applicable for a given client/project combination (perhaps multiple intakes?).

drivers/hmis/app/models/hmis/client_assessment_eligibility_list.rb
```diff
@@ -0,0 +1,61 @@
...
+    POST_EXIT_ROLE => 6,
+  }.freeze
+
+  def each
+    roles = []
+    # Show "intake" item even if the client is entered but does not have an intake
+    roles << INTAKE_ROLE unless assessed?(INTAKE_ROLE)
+
+    # Exit/Update/Annual can only be added to open enrollment
+    roles += [EXIT_ROLE, UPDATE_ROLE, ANNUAL_ROLE] unless assessed?(EXIT_ROLE)
```
ttoomey: is it okay to use the presence of an exit assessment rather than the exit date? If the enrollment is exited but no assessment was performed, we might want to allow assessment to back filled?

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: Interesting. I'm having a hard time finding a scenario where we would want to use this without an enrollment present. I guess the only one I can think of is when you're viewing all the Assessments that a client has had across all enrollments, with the ability to filter that list by assessment type, having the filter only show "relevant" assessments types. But even in that scenario if we wanted to do it, we might as well set the filter options based on what assessment types the client already has.

Did you have something specific in mind? And, can you explain what you mean by multiple intakes using this?

Just want to think it through it because this feels like a new API pattern. (Passing the `enrollment_id` to a field on Client rather than resolving the thing on the `Enrollment`).

drivers/hmis/app/models/hmis/client_assessment_eligibility_list.rb
gigxz: this looks correct.

edit: although the condition should check that there is a _completed_ (non-WIP) Exit assessment. Staff may start working on an exit assessment, but not submit it yet, in which case the enrollment remains open and they should still be allowed to backfill assessments

drivers/hmis/app/models/hmis/client_assessment_eligibility_list.rb
```diff
@@ -0,0 +1,61 @@
...
+    roles << POST_EXIT_ROLE if assessed?(EXIT_ROLE) && !assessed?(POST_EXIT_ROLE) && enrollment&.head_of_household?
+
+    # FIXME: this could be one query rather than ~25 queries :(
```
gigxz: yeah :<
we do soo many queries to the hmis_form_instances table.. which really is a small table so I feel like we should just use in memory structures to determine which ones to use

drivers/hmis/lib/form_data/default/assessments/base_post_exit.json
```diff
@@ -0,0 +1,10 @@
+{
+  "item": [
+    {
+      "fragment": "#r20_post_ext_information_date"
```
gigxz:
```suggestion
      "fragment": "#r20_post_exit_information_date"
```

and renaming the file

drivers/hmis/app/models/hmis/client_assessment_eligibility_list.rb
```diff
@@ -0,0 +1,61 @@
...
+  }.freeze
+
+  def each
+    roles = []
+    # Show "intake" item even if the client is entered but does not have an intake
+    roles << INTAKE_ROLE unless assessed?(INTAKE_ROLE)
+
+    # Exit/Update/Annual can only be added to open enrollment
+    roles += [EXIT_ROLE, UPDATE_ROLE, ANNUAL_ROLE] unless assessed?(EXIT_ROLE)
+    roles << POST_EXIT_ROLE if assessed?(EXIT_ROLE) && !assessed?(POST_EXIT_ROLE) && enrollment&.head_of_household?
```
gigxz: Same here let's only show the Post-Exit option if the Exit Assessment has been submitted.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
ttoomey: Thanks for the feedback! Sounds like this may be too much of a departure from the existing pattern, particularly if there's no reasonable use-case we can anticipate were we would need this flexibility.

I can move this field to enrollment assuming there will never be more than one type of intake for a given project. For example when enrolling a client, there will never be a need for the user to choose between different customized intakes. A use case I could imagine would be a project wanting to pilot or demo new intake forms.

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -318,7 +319,7 @@ def load_definition(form_definition:, identifier:, role:, title: nil)
 
     # Load form definitions for HUD assessments
     public def seed_assessment_form_definitions
-      roles = [:INTAKE, :EXIT, :UPDATE, :ANNUAL]
+      roles = [:INTAKE, :EXIT, :UPDATE, :ANNUAL, :POST_EXIT]
```
gigxz, line 322: This will create a "default instance" for Post-Exit assessment, meaning it sill show up for all projects regardless of funder. Unlike other HUD assessments, we don't want that to happen for post-exit assessments. Post-exit rules are going to be configured on an installation-by-installation basis most likely. So, I don't think we should run lines 345-348 if `role == :POST_EXIT`

drivers/hmis/lib/form_data/default/fragments/r20_post_ext_information_date.json
```diff
@@ -0,0 +1,20 @@
...
+      "type": "DATE",
+      "link_id": "R20.1",
+      "required": false,
+      "warn_if_empty": true,
+      "text": "Aftercare Information Date",
+      "assessment_date": true,
+      "mapping": {
+        "record_type": "EXIT",
+        "field_name": "aftercareDate"
+      }
```
gigxz: [data dictionary says](https://files.hudexchange.info/resources/documents/HMIS-Data-Dictionary-2024.pdf)
> Information may be entered post exit for a period of up to 180 days after which
point no data should be entered. Any data entered prior to the date of exit or after
the 180th day will not be considered in reporting or exports.


Can we add a value bound constraint with `$exitDate` (min) `$exitDate + 180 days` (max)? There is a way to specify the number of offset days in the JSON, look for `offset`. The exitDate should probably already be present as a "local constant" when the form is rendered. let me know if you have any questions/issues!

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: I think it's fair to assume there will be 1 intake assessment.

If a community wanted to pilot a modified intake assessment, they would do something like:
* create a new FormDefinition called "modified intake assessment", with role INTAKE
* create a form instance specifying that modified intake assessment should be used for heads of households in PSH programs (for example)
* that would lead to certain clients getting different intake assessments than other clients. I don't think we need to support letting the user choose which intake. something like that could be supported with custom assessments if we really need it.

gigxz: So yeah, lets switch it over to enrollment then. Thanks!

gigxz (approved):
Looks great @ttoomey!

I pushed some slight changes to the form.

PR #3905, opened 2024-01-12
186821553 rds metrics

ttoomey:

## Description

[Postgres monitoring ](https://www.pivotaltracker.com/n/projects/1603489/stories/186821553)

TODO:
- [ ] Coordinate around remote credential for s3
- [ ] Decide on if/when this should be added to cron

## Type of change
- [x] Infra feature 

### DIFF
--- a/app/jobs/importing/run_daily_imports_job.rb
+++ b/app/jobs/importing/run_daily_imports_job.rb
@@ -168,6 +168,8 @@ module Importing
         create_statistical_matches
         generate_logging_info
 
+        # collect_postgres_stats
+
         finish_processing
       end
     end
@@ -250,6 +252,32 @@ module Importing
 
       GrdaWarehouse::Tasks::PushClientsToCas.new.sync!
       @notifier.ping('Pushed Clients to CAS')
+    end
+
+    # UNTESTED partial implementation. This should probably go in a utility class somewhere else
+    def collect_postgres_stats
+      Dir.mktmpdir do |dir|
+        GrdaWarehouse::PostgresInspector.inspect_each do |inspector|
+          [
+            :database_stats,
+            :table_stats,
+            :index_stats,
+          ].each do |stats_method|
+            results = inspector.send(stats_method)
+            filename = Pathname.new(dir).join("#{inspector.name}.#{stats_method}.csv")
+            CSV.open(filename.to_s, 'w') do |csv|
+              headers = results.first.keys
+              csv << headers
+
+              results.each do |result|
+                csv << result.values_at(**headers)
+              end
+            end
+          end
+        end
+        raise 'implementation tbd'
+        # send_directory_to_s3(dir)
+      end
     end
   end
 end
--- a/app/jobs/importing/run_daily_imports_job.rb
+++ b/app/jobs/importing/run_daily_imports_job.rb
@@ -253,31 +253,5 @@ module Importing
       GrdaWarehouse::Tasks::PushClientsToCas.new.sync!
       @notifier.ping('Pushed Clients to CAS')
     end
-
-    # UNTESTED partial implementation. This should probably go in a utility class somewhere else
-    def collect_postgres_stats
-      Dir.mktmpdir do |dir|
-        GrdaWarehouse::PostgresInspector.inspect_each do |inspector|
-          [
-            :database_stats,
-            :table_stats,
-            :index_stats,
-          ].each do |stats_method|
-            results = inspector.send(stats_method)
-            filename = Pathname.new(dir).join("#{inspector.name}.#{stats_method}.csv")
-            CSV.open(filename.to_s, 'w') do |csv|
-              headers = results.first.keys
-              csv << headers
-
-              results.each do |result|
-                csv << result.values_at(**headers)
-              end
-            end
-          end
-        end
-        raise 'implementation tbd'
-        # send_directory_to_s3(dir)
-      end
-    end
   end
 end
--- a/app/jobs/importing/run_daily_imports_job.rb
+++ b/app/jobs/importing/run_daily_imports_job.rb
@@ -168,8 +168,6 @@ module Importing
         create_statistical_matches
         generate_logging_info
 
-        # collect_postgres_stats
-
         finish_processing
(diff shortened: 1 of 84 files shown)


ttoomey:
If this looks okay, I could probably complete this feature. I'd just need guidance on how we'd like to sync to s3. Also not sure how we'd like to orchestrate the job, I dropped it into the daily import which was just a guess.

app/models/grda_warehouse/postgres_inspector.rb
```diff
@@ -0,0 +1,100 @@
...
+  def initialize(name:, connection:)
+    self.name = name
+    self.connection = connection
+  end
+
+  def self.inspect_each
+    configs = Rails.configuration.database_configuration[Rails.env]
+
+    configs.each_pair do |name, config|
+      next unless config['adapter'] == 'postgresql'
```
eanders: not sure it matters, but we have postgis dbs, does that get reflected in the adapter?

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -251,5 +253,31 @@ def sync_with_cas
...
+              results.each do |result|
+                csv << result.values_at(**headers)
+              end
+            end
+          end
+        end
+        raise 'implementation tbd'
+        # send_directory_to_s3(dir)
+      end
+    end
```
eanders: 1. I'd like to see the logic moved into a method on `PostgresInspector` so you could call something like `PostgresInspector.collect_statistics!`.
2. I think we should stick it in `lib/tasks/grda_warehouse.rake#hourly` and implement it as we have some others, where it runs in a default hour, but can be overridden with an ENV or site config (I'm guessing ENV is fine.)

Depending on how long it takes to run, we may want to setup it's own scheduled task or throw it into a delayed job so we don't block the `hourly` task.

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -251,5 +253,31 @@ def sync_with_cas
...
+              csv << headers
+
+              results.each do |result|
+                csv << result.values_at(**headers)
+              end
+            end
+          end
+        end
+        raise 'implementation tbd'
+        # send_directory_to_s3(dir)
```
eanders: These should be dropped into a file on S3 in the metrics bucket (issue opened with devops) with a path something like:
```
CLIENT-ENV-APP/YYYY-MM-DD--CONNECTION-METHOD.csv
```

Example: `qa-staging-warehouse/2024-01-12--staging_qa_warehouse-database_stats.csv`

We'll probably want to setup an RemoteCredential that knows how to find the location.

app/jobs/importing/run_daily_imports_job.rb
ttoomey: Sounds good. I'll make a new module for this, probably `AppResourceInspector::PostgresInspector` and maybe move it under lib/util

Fwiw, none of these queries seem to be very expensive to run at least locally. Perhaps this is different in production

Maybe in the future this could be an API rather than writing to s3. Or perhaps we could use something like [this prometheus exporter](https://github.com/prometheus-community/postgres_exporter)

app/models/grda_warehouse/postgres_inspector.rb
ttoomey: adjusted for postgis

app/models/app_resource_monitor/app_inspector.rb
```diff
@@ -0,0 +1,35 @@
...
+# == AppResourceMonitor::AppInspector
+#
+# Utility to collect and report app usage
+#
+class AppResourceMonitor::AppInspector
+  def self.utilization_stats
+    [
+      {
+        active_users: User.where(active: true).count,
+        hud_clients: GrdaWarehouse::Hud::Client.count,
```
eanders:
```suggestion
        hud_clients: GrdaWarehouse::Hud::Client.source.count,
```

app/models/app_resource_monitor/app_inspector.rb
```diff
@@ -0,0 +1,35 @@
...
+        hud_enrollments: GrdaWarehouse::Hud::Enrollment.count,
+        hud_projects: GrdaWarehouse::Hud::Project.count,
+        hud_reports: HudReports::ReportInstance.count,
```
eanders, line 19:
```suggestion
        hud_reports: HudReports::ReportInstance.count,
        simple_reports: SimpleReports::ReportInstance.count,
```

app/jobs/importing/run_daily_imports_job.rb
ttoomey: * I put the s3 export logic into AppResourceMonitor::Report
* job will sync to s3 if there's a remote credential with slug for app_stats
* added timestamp, client, and RAILS_ENV to path/file: `/qa_hmis-staging/postgres_database_stats-20240113180926.csv

eanders (commented):
That looks like a solid start to me!

app/jobs/importing/run_daily_imports_job.rb
ttoomey: > Depending on how long it takes to run, we may want to setup it's own scheduled task or throw it into a delayed job so we don't block the hourly task.

Do we really want an **hourly** task for this? That would generate an awful lot of reports. The PT issue specified weekly. 

For our purposes of initial discovery of low hanging fruit, I suspect we could just run this manually on a few installations. I've left this out of the cron  config for now

eanders: Sorry, I didn't mean to imply that would run every hour, only that if we add it to the hourly rake task, then we can set a `run_hour` on the remote credential, similar to how we do it here: `GrdaWarehouse::CustomImports::Config`.  It checks each hour if it should be run, and does nothing if the hour isn't correct, but queues the delayed job if the hour matches.  That way we can stagger them across installations and adjust timing easily, by updating the DB (or an ENV) with the appropriate hour.  

It looks like `remote_credentials` doesn't have an hour column, so that may be worth adding.  Running manually is probably a good idea anyway.

PR #3910, opened 2024-01-12
SPM Performance Tweaks

awisema:

## Description

Refactor and use preloads to improve SPM performance.

## Type of change

### DIFF
--- a/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
@@ -16,6 +16,7 @@ module HudSpmReport::Fy2023
     has_many :hud_reports_universe_members, inverse_of: :universe_membership, class_name: 'HudReports::UniverseMember', foreign_key: :universe_membership_id
 
     attr_accessor :report # FIXME?
+    attr_writer :filter
 
     def compute_episode(enrollments, included_project_types:, excluded_project_types:, include_self_reported_and_ph:)
       raise 'Client undefined' unless client.present?
@@ -80,11 +81,17 @@ module HudSpmReport::Fy2023
         if enrollment.project_type.in?(HudUtility2024.project_type_number_from_code(:es_nbn))
           # NbN only gets service nights in the report range
           bed_nights.merge!(
-            enrollment.enrollment.
-              services.merge(GrdaWarehouse::Hud::Service.bed_night.between(start_date: report_start_date, end_date: report_end_date)).
-              pluck(s_t[:id], s_t[:date_provided]).map do |service_id, date|
+            enrollment.enrollment.services. # Preloaded
+              # merge(GrdaWarehouse::Hud::Service.bed_night.between(start_date: report_start_date, end_date: report_end_date)).
+              # pluck(s_t[:id], s_t[:record_type], s_t[:date_provided]).
+              map do |service|
+                service_id = service.id
+                record_type = service.record_type
+                date = service.date_provided
+                next unless record_type == HudUtility2024.record_type('Bed Night', true) && date.between?(report_start_date, report_end_date)
+
                 [enrollment, service_id, date]
-              end.group_by(&:last).
+              end.compact.group_by(&:last).
               transform_values { |v| Array.wrap(v).last }, # Unique by date
           )
         else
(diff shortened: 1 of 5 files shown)


eanders:
@awisema , looks like this broke some SPM tests.

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
```diff
@@ -102,11 +81,17 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
...
+            enrollment.enrollment.services. # Preloaded
+              # merge(GrdaWarehouse::Hud::Service.bed_night.between(start_date: report_start_date, end_date: report_end_date)).
+              # pluck(s_t[:id], s_t[:record_type], s_t[:date_provided]).
+              map do |service|
+                service_id = service.id
+                record_type = service.record_type
+                date = service.date_provided
+                next unless record_type == HudUtility2024.record_type('Bed Night', true) && date.between?(report_start_date, report_end_date)
+
                 [enrollment, service_id, date]
```
eanders: maybe: 
```suggestion
              map do |service|
                bed_night = service.record_type == == HudUtility2024.record_type('Bed Night', true)
                date = service.date_provided
                next unless bed_night && date.between?(report_start_date, report_end_date)

                [enrollment, service.id, date]
```

Since we aren't using service_id more than once, and the only reason to assign the record type is to determine if it is a bedight, seems like it's a bit cleaner to only assign the boolean, and then just pull the id from the record later.

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/spm_enrollment.rb
```diff
@@ -256,25 +261,42 @@ def self.detail_headers
...
+        #   income_benefits_annual_update.
+        #   where(information_date: ..end_date).
+        #   where(annual_update_window_sql(enrollment)).
+        #   order(information_date: :desc).
+        #   first
+        enrollment.income_benefits.select do |ib|
+          ib.data_collection_stage == 5 && ## Annual update
+            ib.information_date <= end_date &&
+            date_in_annual_update_window(ib.information_date, enrollment)
+        end.sort_by(&:information_date).reverse.first
```
eanders: Any reason not to:
```suggestion
        end.sort_by(&:information_date).last
```

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/spm_enrollment.rb
```diff
@@ -256,25 +261,42 @@ def self.detail_headers
...
+      # enrollment.
+      #   income_benefits_annual_update.
+      #   where(information_date: ...annual_date).
+      #   where(annual_update_window_sql(enrollment)).
+      #   order(information_date: :desc).
+      enrollment.income_benefits.select do |ib|
+        ib.data_collection_stage == 5 && ## Annual update
+          ib.information_date < annual_date &&
+          date_in_annual_update_window(ib.information_date, enrollment)
+      end.sort_by(&:information_date).reverse.first || enrollment.income_benefits_at_entry # Default to entry assessment if less than 2 years
```
eanders: same here:
```suggestion
      end.sort_by(&:information_date).last || enrollment.income_benefits_at_entry # Default to entry assessment if less than 2 years
```

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/spm_enrollment.rb
```diff
@@ -256,25 +261,42 @@ def self.detail_headers
...
+    end
+
+    private_class_method def self.date_in_annual_update_window(date, enrollment)
+      entry_date = enrollment.entry_date
+      interval = 30.days
+      window_date = Date.new(date.year, entry_date.month, entry_date.day)
+
+      window_date - interval .. window_date + interval
```
eanders: This appears to be returning the range rather than the boolean.  Seems like it should be returning a true or false "in the window".  Also, maybe we should add a `?` to the method name to indicate that?

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
awisema: The local variables were left to keep the code as close to the `pluck` based forms as possible in case we wanted it back. But, there does seem to be a significant performance speed up from these changes -- so I cleaned this up

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/spm_enrollment.rb
awisema: This was a bug -- fixed and cleaned up.

PR #3912, opened 2024-01-15
186836824-object-level-authorization

ttoomey:

## Description

[Object-level authorization for Enrollment and Client
](https://www.pivotaltracker.com/story/show/186836824)

Object-level authorization provides redundancy with authorization in the resolver and fields.
* add object-level authorization to "hub" nodes: enrollment, client, and project
* throw an exception with object authorization fails

This change should not cause any difference in behavior, it should not have any noticeable performance impact

## Type of change
- [x] Security improvement 


drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -26,6 +26,12 @@ def self.configuration
       Hmis::Hud::Project.hmis_configuration(version: '2024')
     end
 
+    # check for the most minimal permission needed to resolve this object
+    def self.authorized?(object, ctx)
+      permission = :can_view_project
+      super && GraphqlPermissionChecker.current_permission_for_context?(ctx, permission: permission, entity: object)
+    end
```
gigxz, line 33: This might break some things, we should do a close QA on it. It may mean we need to change some places that resolve Project to resolve PublicProjectInfo .

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
```diff
@@ -28,6 +28,14 @@ def self.configuration
       Hmis::Hud::Enrollment.hmis_configuration(version: '2024')
     end
 
+    # check for the most minimal permission needed to resolve this object
+    def self.authorized?(object, ctx)
+      permissions = [:can_view_enrollment_details, :can_view_limited_enrollment_details]
+      super && permissions.any? do |permission|
+        GraphqlPermissionChecker.current_permission_for_context?(ctx, permission: permission, entity: object)
+      end
```
gigxz: Technically you need `(can_view_project AND can_view_enrollment_details) OR can_view_limited_enrollment_details` so we should check all that here if we can :\

ttoomey: Okay, I think I've represented that auth logic

drivers/hmis/app/graphql/types/hmis_schema/project.rb
ttoomey: We could skip this if you think it's too risky; I figured it was worth adding a check since Project has so many edges.

drivers/hmis/app/graphql/mutations/delete_client.rb
```diff
@@ -20,6 +20,10 @@ def resolve(id:, client_lock_version: nil, confirmed: false)
       client = Hmis::Hud::Client.viewable_by(current_user).find_by(id: id)
       client.lock_version = client_lock_version if client_lock_version
 
+      # While this is redundant with the viewable_by() scope above, this check caches the authorization result so that
+      # the client object-level authorization check will succeed even after the client has been deleted
+      access_denied! unless current_permission?(permission: :can_view_clients, entity: client)
```
ttoomey, line 25: Is `:can_view_clients` enough to allow deletion?

gigxz: No, `can_delete_clients` is required. But that is checked below on line 35.

drivers/hmis/app/graphql/types/hmis_schema/project.rb
gigxz: no this is good you can ignore my comment. I looked and it is not risky

gigxz (approved):
looks good

PR #3915, opened 2024-01-17
Add gql concern for "flat" project and organization names

gigxz:

## Description

**BREAKING API CHANGES**

* Make it easier to resolve "flat" fields describing a Project and Organization. In some places this is preferable because we don't NEED to expose access to the entire `Project` type, which is an entry point for lots of PII. (Even though it does have object-level authorization https://github.com/greenriver/hmis-warehouse/pull/3912 so this is not an attempt to use the schema for authorization)
* Replace some instances where we resolve `Project` to resolve flat project fields instead
* Fix some places where `ProjectName` or `OrganizationName` were non-nullable in the schema. Its not the case in the DB so we shouldn't expect it.

## Type of change


ttoomey (approved):
So this seems fine but I'm wondering if we could consider using a [graphql interface](https://graphql-ruby.org/type_definitions/interfaces) for this instead

An interface is more rigid and would mean we wouldn't get to dynamically adjust the schema (nullable and skip_project_type wouldn't work). I believe all types have to implement the interface in the same way. Maybe that's a deal-breaker though.

It's appealing to me to use a built-in feature of graphql rather than create another HasWidgets module.

Here's what an interface might looks like:

```ruby
module Types
  module HmisSchema
    module ProjectFieldsInterface
      include Types::BaseInterface

      field(:project_id, GraphQL::Types::ID, null: true)
      field(:project_name, GraphQL::Types::String, null: true)
      field(:project_type, Types::HmisSchema::Enums::ProjectType, null: true)
      field(:organization_id, GraphQL::Types::ID, null: true)
      field(:organization_name, GraphQL::Types::String, null: true)

      delegate :id, :name, :type, to: :project, prefix: true, allow_nil: true
      delegate :id, :name, to: :object, prefix: true, allow_nil: true

      def project
        load_ar_association(object, :project)
      end

      def organization
        load_ar_association(object, :organization)
      end
    end
  end
end
```

```diff
 module Types
   class HmisSchema::ReferralPosting < Types::BaseObject
+    implements HmisSchema::ProjectFieldsInterface
     description 'A referral for a household of one or more clients'
 
     field :id, ID, null: false
```

PR #3916, opened 2024-01-17
Chronic individuals system cohort

awisema:

## Description

Add a new system cohort that contains only clients with enrollments that are both chronic at entry, and present as an individual.

Also, fix a bug introduced when we updated cohort tabs that prevented display of newly created system cohorts.

## Type of change
- [X] Bug fix


app/models/grda_warehouse/system_cohorts/chronic_individual.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    def cohort_name
+      'Chronic Individual'
+    end
+
+    private def enrollment_source
+      GrdaWarehouse::ServiceHistoryEnrollment.
+        entry.ongoing(on_date: @processing_date).
+        where(
+          presented_as_individual: true,
```
eanders: Hmm, I'm not sure I want to use this criteria.  Can we adjust to "Adult Only Household".  I don't think the intention is that there is only one person in the household, but rather to avoid including children.   That'll probably make this more complicated (find open chronic enrollments, collect up household members, throw out enrollments with children...)

app/views/admin/configs/_system_cohorts.haml
```diff
@@ -13,5 +13,6 @@
     = f.input :chronic_cohort, label: 'Chronic', hint: 'The chronic cohort includes all clients who are currently chronically homeless with open enrollment in ES, SO, TH, SH and no move-in date in a current PH project.'
     = f.input :adult_and_child_cohort, label: 'Adult and Child', hint: 'The adult and child cohort includes all clients in an adult and child household with open enrollment in ES, SO, TH, SH and no move-in date in a current PH project.'
     = f.input :adult_only_cohort, label: 'Adult Only', hint: 'The adult only cohort includes all clients in a household with only adults with open enrollment in ES, SO, TH, SH and no move-in date in a current PH project.'
+    = f.input :chronic_individual_cohort, label: 'Chronic Individual', hint: 'The chronic individidual cohort includes all unaccompanied clients who are currently chronically homeless with open enrollment in ES, SO, TH, SH and no move-in date in a current PH project.'
```
eanders: This should be all "Adult only households" or maybe we want to go with Adult only or child only.  Not sure on that.

eanders:
@awisema can you fix the structure file conflict and pull in the current version of `release-100` seems like some other stuff is getting in the way of your diff.

db/structure.sql
eanders, line 1: Can you pull this out?  No migrations against the app db.

eanders (approved):
@awisema looks good except the `db/structure.sql` file shouldn't be in there.

PR #3923, opened 2024-01-18
Add table for scan card codes; mutations for managing them; searching

gigxz:
## Description

* New table `hmis_scan_card_codes` stores scan card values
* Mutations for create/delete/restore
* Search by scan card code for alphanumeric text inputs
* **How to test**: enable permission, go to client dashboard, click "generate code", then search by the code in any client text search

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/613

## Type of change


db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
```diff
@@ -0,0 +1,12 @@
+class AddScanCardCodesTable < ActiveRecord::Migration[6.1]
+  def change
+    create_table(:hmis_scan_card_codes) do |t|
+      t.references :client, null: false
+      t.string :code, index: true, null: false # code to embed in scan card
+      t.references :created_by # user that generated code
+      t.references :deleted_by # user that deleted code
```
eanders:
```suggestion
      t.string :code, index: true, null: false, comment: 'code to embed in scan card'
      t.references :created_by, comment: 'user that generated code'
      t.references :deleted_by, comment: 'user that deleted code'
```

Why not document this in the DB as well.

drivers/hmis/app/graphql/mutations/create_scan_card_code.rb
```diff
@@ -0,0 +1,24 @@
...
+
+    field :scan_card_code, Types::HmisSchema::ScanCardCode, null: true
+
+    def resolve(client_id:)
+      client = Hmis::Hud::Client.viewable_by(current_user).find(client_id)
+      raise 'unauthorized' unless current_permission?(permission: :can_manage_scan_cards, entity: client)
+
+      scan_card_code = Hmis::ScanCardCode.new(client: client, created_by: current_user)
+      scan_card_code.assign_code
+      scan_card_code.save!
```
eanders: This is totally fine as is, but you could probably just do:
```suggestion
      scan_card_code = Hmis::ScanCardCode.create!(client: client, created_by: current_user)
```
if you move `assign_code` to an after initialize hook.

drivers/hmis/app/models/hmis/scan_card_code.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  def self.generate_code
+    'S' + SecureRandom.hex(5).upcase
+  end
+
+  def assign_code
+    return if code.present?
+
+    assign_attributes(code: self.class.generate_code)
+  end
```
eanders: I think we might need some duplicate detection, like maybe generate a code, and confirm that it doesn't exist in the table before assigning it?  5 characters is a pretty big namespace, but I wouldn't be surprised if we had occasional collisions.

drivers/hmis/app/models/hmis/scan_card_code.rb
```diff
@@ -0,0 +1,25 @@
...
+  belongs_to :deleted_by, class_name: 'Hmis::User', optional: true
+
+  def self.generate_code
+    'S' + SecureRandom.hex(5).upcase
+  end
+
+  def assign_code
+    return if code.present?
+
+    assign_attributes(code: self.class.generate_code)
```
ttoomey: this is fine, conditional assignment is more idiomatic:
```ruby
self.code ||= self.class.generate_code
```

app/models/concerns/client_search.rb
```diff
@@ -64,6 +66,7 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
       # dummy condition to start the OR chain. This method needs refactoring
       where ||= never_cond
       where = search_by_external_id(where, text) if alpha_numeric && respond_to?(:search_by_external_id) && RailsDrivers.loaded.include?(:hmis_external_apis) && HmisExternalApis::AcHmis::Mci.enabled?
+      where = scan_card_search(where, text) if alpha_numeric
```
ttoomey: We decided not to go with a dedicated search?

Maybe we can restrict this new behavior to just the HMIS? Maybe something like:
```ruby
      where = scan_card_search(where, text) if alpha_numeric && hmis_enabled?
```

db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
```diff
@@ -0,0 +1,12 @@
+class AddScanCardCodesTable < ActiveRecord::Migration[6.1]
+  def change
+    create_table(:hmis_scan_card_codes) do |t|
+      t.references :client, null: false
+      t.string :code, index: true, null: false # code to embed in scan card
```
ttoomey: Should there be an expiry time on these codes too? (They shouldn't be valid forever presumably)

Probably should give code a unique index. Maybe deleted codes are included in the index to ensure they aren't re-used:
```ruby
  t.string :code, index: { unique: true }, null: false
  t.timestamp :expires_at, null: false
```

ttoomey: this is fine but wanted to mention that it may be better to name an attribute in a way that differentiates it from the likely variable name the the model. For example since this attr is named `code`, in the rails you'd probably end up with `code.code` which is less ergnomic/descriptive than something like `code.value` (imo)

drivers/hmis/app/graphql/mutations/delete_scan_card_code.rb
```diff
@@ -0,0 +1,26 @@
...
+  class DeleteScanCardCode < CleanBaseMutation
+    argument :id, ID, required: true
+
+    field :scan_card_code, Types::HmisSchema::ScanCardCode, null: true
+
+    def resolve(id:)
+      code = Hmis::ScanCardCode.find(id)
+      raise 'unauthorized' unless current_permission?(permission: :can_manage_scan_cards, entity: code.client)
+
+      return code if code.deleted_at.present?
```
ttoomey: This should never happen because `Hmis::ScanCardCode.find(id)` will never return deleted records

drivers/hmis/app/graphql/mutations/restore_scan_card_code.rb
```diff
@@ -0,0 +1,26 @@
...
+  class RestoreScanCardCode < CleanBaseMutation
+    argument :id, ID, required: true
+
+    field :scan_card_code, Types::HmisSchema::ScanCardCode, null: true
+
+    def resolve(id:)
+      code = Hmis::ScanCardCode.with_deleted.find(id)
+      raise 'unauthorized' unless current_permission?(permission: :can_manage_scan_cards, entity: code.client)
+
+      return code unless code.deleted_at.present?
```
ttoomey: I don't think returning code is what we want, it's not the right shape.

Suggest we raise an error instead:
```suggestion
      raise unless code.deleted?
      # or
      return { scan_card_code: code } unless code.deleted?
```
Or even better, require the code to be deleted
```ruby
      code = Hmis::ScanCardCode.only_deleted.find(id)
```

ttoomey (commented):
Looking good! I left a few notes but I think the only consequential change is that code needs a unique constraint of some kind

db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
gigxz: Oh good point, `value` is better.
I can add an expiration column but I'm not sure what that duration should be, would need to get some feedback on that

drivers/hmis/app/graphql/mutations/restore_scan_card_code.rb
gigxz: for some reason I was intending the mutations to be lenient, not error if trying to delete something thats already deleted. In case there is wonkiness in the UI where that happens. But it doesn't match patterns in other mutations, which are stricter on input. so I can change it back to that

drivers/hmis/app/models/hmis/scan_card_code.rb
gigxz: If I add a uniqueness constraint like @ttoomey suggested, an exception will be thrown and the user will just have to try again creating one. does that seem ok?

db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
gigxz: @eanders what do you think about an expiration timestamp and duration?

app/models/concerns/client_search.rb
gigxz: Yes we can restrict to HMIS.

What do you mean by dedicated search? The logic flow of this function is so strange.

We don't know anything for certain about the shape of the scan card code, aside from the fact that it's alphanumeric.  I suppose we could check it first before anything else, and short-circuit the search if we get a result. This kind of does the opposite..skips the name-short-circuit if it _is_ a match. I'll try adding it earlier and see if it makes more sense.

gigxz: Ok, changed: https://github.com/greenriver/hmis-warehouse/pull/3923/files#diff-e0530ee1d7a90940cb0b96a903efe907a875c5495aa71f1088a00488cd20606c 
This makes more sense to me

drivers/hmis/app/models/hmis/scan_card_code.rb
eanders: Hmm, I guess it's ok, but we don't need to fail, we can just try again until we get one that isn't in use.  Seems odd that you'd get an error on the front end for something you don't control.  If you picked the code manually, I'd feel fine with that, but since we're generating them, we should probably do our best to avoid duplication.

gigxz: https://github.com/greenriver/hmis-warehouse/pull/3923/commits/a4a84b878c62963523b7ad570ec5f345602e78fc

db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
eanders: I think an optional expiration is a good idea, so we the card can be expired without having to be deleted.  I _think_ that would manifest as something that was set at creation time (and optionally updated later.)  I can also see having an indefinite card, depending on who is issuing it or what it is for.  I don't know why you'd expire a card that is just used to attach a meal to a person, but maybe there's a reason you'd want to check the PII of a person every year or two, so having the expiration could trigger a workflow in HMIS when scanned.

drivers/hmis/app/models/hmis/scan_card_code.rb
eanders: Thank you, I think that looks good.

eanders (approved):

app/models/concerns/client_search.rb
ttoomey: Re dedicated search: sorry what I meant was that we could have a much simpler method that only searched by scan code and would return just the one matching client record. That would mean we'd use a dedicated search input. With this approach it looks like we plan to use the existing client search for scan cards rather than a dedicated input

app/models/concerns/client_search.rb
```diff
@@ -26,6 +26,12 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
       max_pk = 2_147_483_648 # PK is a 4 byte signed INT (2 ** ((4 * 8) - 1))
       term_is_possibly_pk = numeric ? text.to_i < max_pk : false
 
+      # If alphanumeric search term matches a Scan Card code, return immediately
+      if alpha_numeric && HmisEnforcement.hmis_enabled?
+        matching_scan_card = Hmis::ScanCardCode.find_by(value: text)
+        return where(sa[:id].eq(matching_scan_card.client_id)) if matching_scan_card
+      end
```
ttoomey: If a scan card can be 'S' + hex chars, it could be the same as someone's name right? For example "Sade" is a human name and a hex card code. These collisions are a tiny edge case but could cause very unexpected search results.

db/warehouse/migrate/20240118203430_add_scan_card_codes_table.rb
gigxz: Scan card expiration seems like a very cumbersome way to trigger that type of workflow.. at least with the current level of difficulty for re-creating and re-printing a scan card. 

This does feels like scope creep since it will change the interface t(o require user input for an expiration period), which I'm not sure we want. I'll add the column for now but hold off on UI changes until we can validate the utility of this feature.

gigxz: The expiration thing also seems like something that would site or installation-wide, not decided by the user every time they make a card. In that case maybe it would just be a job that deactivates (deletes) scan cards that are a certain number of years old. 

I added a commit with the optional expiration column though.

app/models/concerns/client_search.rb
gigxz: Yes it is possible though very unlikely. We could throw another random number on there to reduce the likelihood, since names generally should not have numbers in there. sound good ?

gigxz: I changed it so the prefix is `S1` instead of `S`

app/models/concerns/client_search.rb
gigxz: Oh I see. Yeah we landed on that for now, to reduce the feature scope (no new frontend work needed for enrollment and bed night workflows). It should be fine because the migrated-in codes have format `P123124` (P + number)

ttoomey (approved):
Looks great!

PR #3926, opened 2024-01-19
Enable enrollment deletion if no intake exists

martha:

## Description

Enables deleting enrollments if the intake assessment does not exist: something that should never happen now, but could happen due to bad data, or in the future if we support projects that don't do intake assessments.

https://www.pivotaltracker.com/n/projects/2591838/stories/186726767/

## Type of change
- [ ] Bug fix
- [ ] Dependency update

- [?] I have updated the documentation (or not applicable)
- [?] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

Question for you @gigxz , are there any unit tests that I should update along with this change?


gigxz (commented):
Logic looks good!

Yes, lets update the `delete_enrollment_spec` to cover the new case.

One of the cases failed so that will need updating too:  https://github.com/greenriver/hmis-warehouse/actions/runs/7588439136/job/20670952396

Running the test locally will be something like:
```
dcr shell

rails db:setup_test RAILS_ENV=test # if you haven't set up your test db yet
rails db:migrate RAILS_ENV=test # if you have set it up, but there are newer migrations to run

rspec ./drivers/hmis/spec/requests/hmis/delete_enrollment_spec.rb
```


(also, you can ignore the rubocop failure in the github action.)

martha:
Thanks @gigxz , please let me know what you think about the changes - I'm definitely open to further feedback including nitpicks about ruby style!

drivers/hmis/app/graphql/mutations/delete_enrollment.rb
```diff
@@ -13,14 +13,24 @@ class DeleteEnrollment < BaseMutation
...
+
         enrollment.destroy!
       else
-        # Deleting non-WIP Enrollments requires can_delete_enrollments, and can only occur via DeleteAssessment mutation (deleting intake)
+        # WIP Enrollments can be deleted if user has "can_edit_enrollments" access for this project
+        raise HmisErrors::ApiError, 'Access denied' unless current_user.permissions_for?(enrollment, :can_edit_enrollments)
+
+        # Deleting non-WIP Enrollements requires can_delete_enrollments, and can only occur via DeleteAssessment mutation (deleting intake)
         errors << HmisErrors::Error.new(:base, full_message: 'Completed enrollments can not be deleted. Please exit the client instead.')
       end
```
gigxz, line 32: This logic is sound, but I think it could be cleaned up a little. The user really should have `can_edit_enrollments` in all 3 cases. (There is no scenario where someone should be able to delete enrollments but not edit them.) So I wonder, rather than having the can_edit_enrollments check on both lines 20 and 30, if we should just keep it on line 17 where it was. That way we would _always_ ensure the user has `can_edit_enrollments` before proceeding.

The comment, `WIP Enrollments can be deleted if user has "can_edit_enrollments" access for this project`, only really makes sense to stay on line 19 I think -- since that is where we are actually handling the WIP Enrollment case.

martha: Thanks Gig, perfect!

gigxz (approved):
💯

drivers/hmis/spec/requests/hmis/delete_enrollment_spec.rb
```diff
@@ -58,6 +64,16 @@ def perform_mutation(enrollment)
...
 
+  it 'should allow deleting non-WIP enrollments if no intake assessment exists' do
+    add_permissions(access_control, :can_edit_enrollments, :can_delete_enrollments)
+    enrollment, errors = perform_mutation(e3)
+    expect(enrollment).to be_present
+    expect(errors).to be_empty
+    expect(e1.reload).not_to be_deleted
+    expect(e2.reload).not_to be_deleted
+    expect(e3.reload).to be_deleted
+  end
```
ttoomey, line 75: Thanks for including tests 🙌  

I might be missing something but do we need to test e1 and e2? It seems unnecessary to me - @gigxz what do you think? If we really want to, maybe we could just have one dedicated test for this: "Enrollment deletion only deletes the selected enrollment".

Also I think we are pretty inconsistent with how we structure tests. This is subjective but I'd suggest using something like the following pattern:
* use an enclosing context to to help clarify what variables like e3 are for
* use an `expect` block to concisely check for expected changes

```suggestion
  context 'a completed enrollment without an intake assessment' do
    let(:c3) { create :hmis_hud_client, data_source: ds1, user: u1 }
    let!(:e3) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c3, relationship_to_ho_h: 1 }
    it 'should be deleted successfully if authorized' do
      add_permissions(access_control, :can_edit_enrollments, :can_delete_enrollments)
      expect do
        enrollment, errors = perform_mutation(e3)
        expect(enrollment).to be_present
        expect(errors).to be_empty
      end.to change(Hmis::Hud::Enrollment.where(id: e3.id), :count).by(-1)
    end
    it 'should throw error if unauthorized' do
      expect_gql_error post_graphql(input: { id: e3.id }) { mutation }
    end
  end
```

ttoomey (approved):

drivers/hmis/spec/requests/hmis/delete_enrollment_spec.rb
martha: Thanks @ttoomey for your feedback!
> do we need to test e1 and e2?

No probably not! I think I just did this because I was following the example from line 63-64, but I agree that it would be clearer if the tests all have a single purpose and separate contexts. I'll open a new PR since this one is already merged.

martha: https://github.com/greenriver/hmis-warehouse/pull/3945

PR #3928, opened 2024-01-21
186801323 generate ce assessments from custom assessments

ttoomey:

## Description

[Support fully custom (non-HUD) assessments
](https://www.pivotaltracker.com/story/show/186494490)
[Support generating CE Assessments from Custom Assessments
](https://www.pivotaltracker.com/story/show/186801323)

https://github.com/greenriver/hmis-frontend/pull/616

* improve support for custom assessments. 
* Form definitions for custom assessments have the custom assessment role. 
* More than one definition with this role may be applicable to an enrollment
* support filtering enrollments by `instance.data_collected_about`

details:
* includes demo custom assessments as POC
* add ce_assessment association to form_processor
* extracted matching logic for form instance into project and enrollment matches. Matching is done in memory to allow for greater flexibility and reduce the overall number of sql queries
* improve documentation for submit_assessment mutation
* add missing unique index to hmis_form_definitions table

json form adjustments:
* support loading custom_assessments
* adjust how we populate pick lists to use rails descendants rather than traverse graphql schema. This was needed cause we seemed to be missing enums otherwise
* dry up builder usage into `builder.seed_all` method

## Type of change


drivers/hmis/app/models/hmis/form/instance_project_match.rb
```diff
@@ -0,0 +1,62 @@
...
+    MATCH_RANKS[match]
+  end
+
+  def valid?
+    !!match
+  end
+
+  protected
+
+  # match to project. Order is significant, should return the best ranked match
```
ttoomey, line 40: this could use another set of eyes. I've tried to capture the logic previously in `instance. detect_best_instance_scope_for_project`

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
```diff
@@ -75,6 +77,14 @@ def in_progress
       object.in_progress?
     end
 
+    protected def form_processor
```
gigxz: should we update all the below methods to use this as well (eg income_Benefit)?

drivers/hmis/app/models/hmis/form/instance_enrollment_match.rb
```diff
@@ -0,0 +1,49 @@
...
+    when HOH_AND_ADULTS_MATCH
+      enrollment.head_of_household? && enrollment.adult?
+    when HOH_MATCH
+      enrollment.head_of_household?
+    when ALL_VETERANS_MATCH
+      enrollment.client.veteran?
+    when VETERAN_HOH_MATCH
+      enrollment.head_of_household? && enrollment.client.veteran?
+    else
+      raise "unknown data_collected about on instance##{instance.id}: #{data_collected_about}"
```
gigxz, line 40: the `data_collected_about` field is optional, null should be treated as ALL_MATCH

drivers/hmis/app/models/hmis/form/instance_project_match.rb
```diff
@@ -0,0 +1,65 @@
...
+###
+
+# determine if project is a valid match for a form instance. The match is ranked so that when there are multiple form
+# definitions that match a project, we can take the definition with the best rank
+#
+class Hmis::Form::InstanceProjectMatch
+  include Memery
+  attr_accessor :project, :instance
+
+  # match types ordered by rank. Lower rank is better
```
gigxz:
```suggestion
  # Match types ordered by rank, from most specific to least specific. lower rank number if more specific.
```
maybe? "Better" was unclear to me

drivers/hmis/app/models/hmis/form/instance_project_match.rb
```diff
@@ -0,0 +1,65 @@
...
+      case instance.entity_type
+      when Hmis::Hud::Project.sti_name
+        return PROJECT_MATCH if instance.entity_id == project.id
+      when Hmis::Hud::Organization.sti_name
+        return ORGANIZATION_MATCH if instance.entity_id == project.organization.id
+      else
+        # entity type is specified but doesn't match
+        return
+      end
+    end
```
gigxz: won't this "fall through"  down to line 54 if there is an entity type specified but the current one doesn't match? if there is an entity specified and it doesn't match, we want to return immediately.

```suggestion
    if instance.entity_type
      case instance.entity_type
      when Hmis::Hud::Project.sti_name
        return PROJECT_MATCH if instance.entity_id == project.id
      when Hmis::Hud::Organization.sti_name
        return ORGANIZATION_MATCH if instance.entity_id == project.organization.id
      end
      
      return  # entity type is specified but doesn't match
    end
```

drivers/hmis/app/models/hmis/form/instance_project_match.rb
```diff
@@ -0,0 +1,65 @@
...
+
+    if instance.project_type
+      if project.project_type && instance.project_type == project.project_type
+        return PROJECT_TYPE_AND_FUNDER_MATCH if instance.funder.in?(project.funders.map(&:funder))
```
gigxz: This doesn't check for `other_funder`, right? It should, as should the `PROJECT_FUNDER_MATCH` check below.

```
matches_project_type = instance.project_type == project.project_type
matches_funder = instance.funder&.in?(project.funders.map(&:funder).compact) || instance.other_funder&.in?(project.funders.map(&:other_funder).compact_blank)

return PROJECT_TYPE_AND_FUNDER_MATCH if matches_project_type && matches_funder
```

drivers/hmis/app/models/hmis/form/instance_project_match.rb
gigxz: thanks for setting this up and parsing through all that logic! I think it is much more readable. Left a few comments but otherwise looks good.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -311,6 +311,10 @@ def age(date = Date.current)
     GrdaWarehouse::Hud::Client.age(date: date, dob: self.DOB)
   end
 
+  def veteran?
+    self.VeteranStatus == 1
+  end
+
```
gigxz: Not a big deal, but this could go [into the shared concern](https://github.com/greenriver/hmis-warehouse/blob/release-101/app/models/concerns/hud_concerns/client.rb) and be removed from the grda warehouse [class](https://github.com/greenriver/hmis-warehouse/blob/eca466263a61e6d08703d1aefe50dae2f501d68c/app/models/grda_warehouse/hud/client.rb#L1985-L1987). The intention, at least originally, was to move shared logic to the concern over time.

db/warehouse/migrate/20240113025936_add_ce_assessment_to_form_processor.rb
```diff
@@ -0,0 +1,7 @@
+class AddCeAssessmentToFormProcessor < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      add_reference :hmis_form_processors, :ce_assessment
```
gigxz, line 4: It looks like the warehouse structure file does not reflect this change

drivers/hmis/app/models/hmis/form/instance_project_match.rb
gigxz: I haven't tried running it yet on your branch but https://github.com/greenriver/hmis-warehouse/blob/release-101/drivers/hmis/spec/models/hmis/form/definition_spec.rb should have decent coverage.

ttoomey: Thanks! Yes that spec passes for me

drivers/hmis/app/models/hmis/form/instance_project_match.rb
ttoomey: Good catch! I tightened up the conditions here

db/warehouse/migrate/20240113025936_add_ce_assessment_to_form_processor.rb
ttoomey: Thanks! I missed it as I was extracting the diff. Should be fixed now

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -8,6 +8,21 @@
...
+#     in combination with identifier, uniquely identify this form
+#   role
+#     the significance of this form within the system (INTAKE, EXIT, etc)
+#   status
+#     NOT IMPLEMENTED: aspirational support for draft status
+#   definition
+#     JSON field defines the inputs, labels, validation, and mapping to HMIS fields. A JSON-schema exists to validate
+#     the format of the definition
+#   title
+#     User-facing title of the form definition
```
gigxz, line 25: Thank you for adding this!

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -204,13 +231,14 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   end
 
   def self.find_definition_for_role(role, project: nil, version: nil)
```
gigxz, line 233: I don't think this function needs to accept `version` at all honestly. I don't know why I added that. The point of version is that the most recent one is the active one. If you wanted to look up a specific older definition  you would do that by the definition ID, not this.

What you have here is fine, but I think deleting this `version` arg and removing that test case would be just as good. Sorry about that

drivers/hmis/lib/form_data/qa_hmis/custom_assessments/HAT.json
```diff
@@ -0,0 +1,65 @@
+{
+  "item": [
```
gigxz, line 2: Thanks for adding these for QA!
I'm not seeing anything saved when I submit though. Am I missing something? It also looks like the frontend branch has a different graphql schema, so maybe something is not updated yet.
Screenshot 2024-01-23 at 12 16 45 PM

ttoomey: Ah oops! I must have forgotten to QA custom assessment/Ce assessment after refactoring. Sorry about that!

fixed in b4e6f6757c623a6271a5283c2c984b66105ad56a
* Added a missing container name for CeAssessment to target the right processor. I believe this is because of the new mapping in RelatedRecordType
* Raise on invalid container name in development or alert to sentry in production. Does this seem okay? I'd like to get an exception if this behavior occurs in the future

gigxz: Yes definitely thanks!

gigxz (approved):
looks great! 👏

PR #3937, opened 2024-01-23
Add Client Alert table and model

martha:

## Description

This PR adds a database and Rails model for HMIS Client Alerts, a new feature where HMIS users can attach alerts to clients which can then be viewed on the client dashboard and the enrollment dashboard.

At least one more PR to come, with graphql mutations to support the frontend queries.

https://www.pivotaltracker.com/n/projects/2591838/stories/184132729

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/spec/models/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,26 @@
+###
```
martha, line 1: @gigxz What spec tests should I include at this stage? I was thinking to just write a quick test for the connections to `client` and `created_by`, the things you recommended manually testing in the rails console, but I can't figure out around the
```
can't write unknown attribute `created_by_id`
```
when trying to run these tests, and it seems most of the spec files that I'm comparing against as examples have GraphQL mutations already. When you have a chance, maybe it would be good to pair a bit about GraphQL.
Also, should I create a FactoryBot Factory for the ClientAlerts at this point?

drivers/hmis/app/models/hmis/client_alert.rb
```diff
@@ -0,0 +1,10 @@
+###
```
martha, line 1: Is it correct to copy this license info into the top of every file? Should I set my ide to do so automatically?

drivers/hmis/app/models/hmis/client_alert.rb
```diff
@@ -0,0 +1,10 @@
...
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::ClientAlert < Hmis::HmisBase
+  belongs_to :created_by, class_name: 'Hmis::User'
```
martha, line 9: The ActiveRecord docs state that "`belongs_to` does not ensure reference consistency if optional is set to true, so depending on the use case, you might also need to add a database-level foreign key constraint on the reference column." Is this important here? (https://guides.rubyonrails.org/association_basics.html)

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -154,6 +154,8 @@ def self.authorized?(object, ctx)
       composite_perm :can_upload_client_files, permissions: [:manage_any_client_files, :manage_own_client_files], mode: :any
       composite_perm :can_view_any_files, permissions: [:manage_own_client_files, :view_any_nonconfidential_client_files, :view_any_confidential_client_files], mode: :any
       can :audit_clients
+      can :view_client_alerts
```
martha, line 161: There must be a command I need to run to get the schema.graphql file to update, right?

drivers/hmis/app/models/hmis/client_alert.rb
gigxz: Yes, thanks! We have a task that inserts them so we do that periodically. i don't have anything automatic with my editor. This just came up recently https://www.pivotaltracker.com/story/show/186889763/comments/239959866

drivers/hmis/app/models/hmis/client_alert.rb
gigxz: It does apply, but I think it's OK. The database column is non-nullable, but there is no foreign key constraint, so its possible that the user could have been deleted. In that case, the `alert.created_by` would be null.

That's fine, generally, we tend not to add foreign keys for lots of things.. I think the main reasoning is that we don't want to have to cascade deletion. it's ok to keep these records around if the user gets deleted for whatever reason (which is not expected btw).

The only gotcha is that on the ClientAlert GraphQL type, the createdBy user field should be nullable. That way, the frontend can prepare for that scenario, rather than crash and burn (which is what GraphQL does if it tries to resolve a null value for a non-nullable schema field).

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -384,6 +384,22 @@ def self.permissions_with_descriptions
           'Administrative',
         ],
       },
+      can_view_client_alerts: {
+        description: 'Grants the ability to view Client Alerts',
+        administrative: true,
+        access: [:editable],
```
gigxz: I don't know if this category is actually used anywhere anymore, but lets keep it consistent anyway. This perm doesn't grant any edit access so its considered viewable
```suggestion
        access: [:viewable],
```

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -384,6 +384,22 @@ def self.permissions_with_descriptions
...
+        categories: [
+          'Administrative',
```
gigxz, line 392: Lets put both of these in `Client Access` category instead. I wouldn't say they are administrative

gigxz: Or `Client` if that's a category

drivers/hmis/spec/models/hmis/client_alert_spec.rb
gigxz: That error looks to me like the test db needs to be migrated: `rails db:migrate RAILS_ENV=test`

We don't typically add "model" tests for simple classes like this.  Most likely instead we will have request tests for the Query / Mutation(s) that you'll add next (resolving alerts on a client; creating an alert; & deleting an alert).

It wouldn't hurt to add a factory in this PR if you want- You can also look at my PR for an example, we are doing similar things! https://github.com/greenriver/hmis-warehouse/pull/3923/files#diff-0c48dca452b35364c6e8e56c0a1d2b1553b586b4db59abe9c4de097ae80331a6

martha: Thank you! That was indeed the issue. Sounds good, I will add a factory here.

drivers/hmis/app/models/hmis/role.rb
martha: In this file I only see `Client Access` and `Client Details` -- I'll change it to `Client Details unless you suggest otherwise!

drivers/hmis/app/models/hmis/client_alert.rb
```diff
@@ -0,0 +1,10 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::ClientAlert < Hmis::HmisBase
+  belongs_to :created_by, class_name: 'Hmis::Hud::User'
```
martha: What is the difference between `Hmis::Hud::User` and `Hmis::User`?

gigxz: Ah, good question. This should actually be a reference to `Hmis::User`, which will be set to the `current_user` that is performing the mutation to create the alert.

* `Hmis::User` and `User` models are both backed by the `users` table in the app database. These are the "application users" aka the user accounts that people log in with.
* `Hmis::Hud::User` is backed by the `User` table in the warehouse database, which matches [User.csv](https://files.hudexchange.info/resources/documents/HMIS-CSV-Format-Specifications-2024.pdf) from HUD. It's really not used for anything except imported data. Its HUD's way of representing users. HUD records, like Client and Enrollment, both store references to that table as specified in the CSV spec.

Unfortunate I know 😞  but the answer is 90% of the time we are working with `Hmis::User`s a.k.a Application Users.

gigxz (approved):
Looks good!!

Can you update the PR description so we have a sentence of context when looking back at changes? That is usually helpful.

drivers/hmis/spec/models/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,31 @@
...
+  include_context 'hmis base setup'
+
+  describe 'basic client alert tests' do
+    let!(:c1) { create :hmis_hud_client, data_source: ds1 }
+    let!(:a1) { create :hmis_client_alert, created_by: hmis_user }
+
+    it 'should be able to create a client alert' do
+      expect(a1.created_by.class).to eq(Hmis::User)
+    end
+  end
```
eanders: I don't think this is testing the permission logic, just that the relation `created_by` is working.  To test the permissions you'll need to go through the graphQL endpoint with a submission that creates the alert.

gigxz: she is doing the graphql endpoints in a separate pr, just the setup here

eanders: That's fine, but checking the class of the created_by relation isn't actually checking `it 'should be able to create a client alert'`  Only that the save worked successfully and the relation is correctly setup.

martha: From talking with Gig earlier my understanding is that the Open Path norm would be not to add any tests testing the functionality I added in this PR, is that right? Should I just delete this test file in this PR?

eanders: No, don't delete it!  We may not be great at adding tests for everything, but we always appreciate when someone takes the time to add them.  I was more commenting that the 'it' line doesn't reflect what the test is doing.  perhaps the following would be a more accurate description:
```suggestion
  describe 'basic client alert tests' do
    let!(:c1) { create :hmis_hud_client, data_source: ds1 }
    let!(:a1) { create :hmis_client_alert, created_by: hmis_user }

    it 'client alert is saved successfully and connected to the saving user' do
      expect(a1.created_by.class).to eq(Hmis::User)
    end
  end
```

martha: Oh perfect, thank you for clarifying!!

martha:
When committing the `release-101` merge, I used `--no-verify` since the precommit hooks were failing on some lint errors related to some things I didn't change. Let me know if I did the right thing there or should revert and make any changes?

eanders:
> When committing the `release-101` merge, I used `--no-verify` since the precommit hooks were failing on some lint errors related to some things I didn't change. Let me know if I did the right thing there or should revert and make any changes?

I should have mentioned, you can `SKIP=RuboCop git commit -am "Your message"` to get around a particular hook (rubocop in this case).

eanders (approved):

drivers/hmis/app/models/hmis/client_alert.rb
```diff
@@ -0,0 +1,10 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::ClientAlert < Hmis::HmisBase
```
ttoomey, line 7: looks like this model has a deleted_at attribute so it seems like we want to use soft delete. If so, I think we'd need to include the behavior from the `paranoia` gem
```suggestion
class Hmis::ClientAlert < Hmis::HmisBase
  acts_as_paranoid
```

drivers/hmis/app/models/hmis/client_alert.rb
```diff
@@ -0,0 +1,10 @@
...
+  belongs_to :created_by, class_name: 'Hmis::User'
+  belongs_to :client, class_name: 'Hmis::Hud::Client'
```
ttoomey, line 10: suggest adding the other side of the relationship to the client model. That way if a client record is destroyed, the alerts will be destroyed too
```ruby
class Hmis::Hud::Client
  # ...
  has_many :client_alerts, class_name: '::Hmis::ClientAlert', dependent: :destroy, inverse_of: :client
```

ttoomey (approved):
looks good!

drivers/hmis/app/models/hmis/client_alert.rb
martha: Thanks, Theron!

gigxz:
Go ahead and merge into release-102 when you're ready @martha  👍

PR #3946, opened 2024-01-24
Add ClientAlerts to Client graphql query

martha:

## Description

In this PR is the graphql query needed to enable showing Client Alerts on the frontend. ~I've opened the PR against `martha/client-alerts` for now, but will update the base branch when that PR is merged in.~

~Depending on the merge timing, I'll either add the graphql mutation to this PR or merge this and make a separate future PR.~ [I'll do this in a separate PR]

In the meantime, please let me know if it looks like I'm on the right track!

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,56 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
martha, line 1: I added this spec file to the `requests` directory, which means there are now two `client_alert_spec.rb` files -- this one and `drivers/hmis/spec/models/hmis/client_alert_spec.rb`. Is that incorrect? Should I name them differently?

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -0,0 +1,20 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::ClientAlert < Types::BaseObject
+    description 'Alert'
+    field :id, ID, null: false
+    field :client, HmisSchema::Client, null: false
```
gigxz: Does the frontend actually need this in the schema?  Since the alert is resolved on the client, we already know who the client is.  If its not used I would recommend removing it.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -275,6 +277,10 @@ def addresses
       load_ar_association(object, :addresses)
     end
 
+    def alerts
+      load_ar_association(object, :alerts)
```
gigxz: We need a permission check here:
`return unless current_permission?(permission: :can_view_client_alerts, entity: object)`

This `current_permission?` function is a special function that use the GraphQL data loader mechanism to check permissions for the current user, optionally against an `entity`. Passing the `entity: object` (which is a client) allows us to enforce the permission on a client-by-client basis. For example, a provider might have permission to see Alerts for clients at "their" projects, but not clients at other projects.

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -0,0 +1,20 @@
...
+    field :note, String, null: false
+
```
gigxz: We'll need additional fields here, since the mocks include:
* user that created the alert
* datetime when the alert was created
* expiration date

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -23,6 +23,7 @@ class HmisSchema::Client < Types::BaseObject
     include Types::HmisSchema::HasGender
     include Types::HmisSchema::HasCustomDataElements
     include Types::HmisSchema::HasHudMetadata
+    include Types::HmisSchema::HasClientAlerts
```
gigxz: This is nice and simple how you have it below.. I think we can remove `include Types::HmisSchema::HasClientAlerts`, I don't think it's necessary to have a concern there since this is a very simple association loading. Sorry for the confusion there!

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: We should make sure the Alerts always show up in a deterministic order. We can do that by adding a sort here with `sort_by`. I think it would make sense to just use the creation date, newer->older.

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
gigxz: I think that's fine

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+        }
+      GRAPHQL
+    end
+
+    let!(:c1) { create :hmis_hud_client, data_source: ds1 }
+    let!(:a1) { create :hmis_client_alert, created_by: hmis_user, client: c1 }
+
+    it 'should return the client with an alert' do
+      response, result = post_graphql(id: c1.id) { query }
+      expect(response.status).to eq 200
```
gigxz: looks good!
I've started doing something like `expect(response.status).to eq(200), result.inspect` here. That way if the request throws an exception, the result (which contains the error) will be logged in the test output. (We should probably have some tooling around this for tests, but we dont right now)

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      alerts = result.dig('data', 'client', 'alerts')
+      expect(alerts.size).to eq 1
+      expect(alerts[0]['note']).to eq 'Something wacky has happened'
```
gigxz: I don't think it's good practice to expect specific values from the factory. If the factory changed, this test would break. For testing specific values like this, I would add `note: 'the value i expect for this specific case'` when creating this alert on line 42, and then test for that.

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+            }
+          }
+        }
+      GRAPHQL
+    end
+
+    let!(:c1) { create :hmis_hud_client, data_source: ds1 }
+    let!(:a1) { create :hmis_client_alert, created_by: hmis_user, client: c1 }
+
+    it 'should return the client with an alert' do
```
gigxz, line 45: lets add a test case for a user that does not have permission to view alerts. it should return no alerts even if they exist

gigxz (commented):
looking good!

drivers/hmis/app/graphql/types/hmis_schema/client.rb
martha: Thanks for the detailed explanation!

Is it correct to return `[]` here if the user doesn't have permission? I did this because I remember in a conversation earlier this week about this field, I remember you saying that since it's a list it should always be non-null, since it could just be an empty list. However, maybe that was specific to the Rails model and not the graphql type?

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
martha: Good plan, let me know what you think about the test case that I added. I don't know if it would ever really be the case that a user has permission to view a client but not that client's alerts -- maybe more likely they would either be able to view a client and their alerts, or not view the client at all? -- I just wrote it this way so the test is testing the actual permission `:can_view_client_alerts` instead of something less granular. But let me know if I should take a different approach with this!

gigxz: Test looks good to me!

I do think it's likely that a user may have access to view a Client but not their Alerts. I can think of at least two scenarios:
1. The installation doesn't want to use Client Alerts at all, so we don't give the alert-viewing permission to anyone
2. A setup where Providers can see all Clients, but they can only see Client Alerts for the clients that are enrolled at their programs

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: That's fine to return. I _think_ returning nil will also lead to an empty array being resolved, because ruby-graphql just does that

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -0,0 +1,19 @@
...
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::ClientAlert < Types::BaseObject
+    description 'Alert'
+    field :id, ID, null: false
+    field :note, String, null: false
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :created_by, Types::Application::User, null: true
+    field :expiration_date, GraphQL::Types::ISO8601DateTime, null: true
```
gigxz: I think the expiration date was stored as a date not a datetime, right? If so it should resolve a `ISO8601Date`

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -0,0 +1,19 @@
...
+    field :severity, String, null: true
```
gigxz: I think we landed on `priority` for the user-facing language, so lets use that in the api schema too.

Also - can you create a GraphQL enum for this?

`Types::HmisSchema::Enums::ReminderTopic` is one example to look at. It reads the possible values from a different class. You could do the same thing, reading the priority levels from the ClientAlert model: `Hmis::ClientAlert::PRIORITY_LEVELS`. Where the values are`['low','medium','high']`. You can also add a validation rule on ClientAlert, which will ensure that any value written to the severity field is one of the approved strings. (See rails `validates` docs with  `inclusion`)

gigxz: And lets make it `null: false`. You can add `default_value: 'low'`

drivers/hmis/app/graphql/types/hmis_schema/client.rb
martha: I'm seeing `RuntimeError: Cannot return null for non-nullable field Client.alerts` when I change it from `return [] unless` to just `return unless`

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
martha: that makes sense, thank you!

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
martha: That's right! Thank you :)

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
martha: @gigxz this all sounds good to me, but I'm not quite sure how to create a module `Hmis::ClientAlert` with `PRIORITY_LEVELS` inside it when `Hmis::ClientAlert` already exists as a class? Is that what you're indicating here, or should it be named something else in order to not conflict?

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -0,0 +1,19 @@
...
+
+module Types
+  class HmisSchema::ClientAlert < Types::BaseObject
+    description 'Alert'
+    field :id, ID, null: false
+    field :note, String, null: false
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :created_by, Types::Application::User, null: true
+    field :expiration_date, GraphQL::Types::ISO8601Date, null: true
+    field :priority, HmisSchema::Enums::ClientAlertPriorityLevel, null: false, default_value: 'low'
```
martha: Here, instead of `default_value: 'low'`, do you know if there is any way to refer to a value in the enum (the way you can do in java/python)? Something like `Hmis::AlertPriority::PRIORITY_LEVELS.LOW`?

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
gigxz: `PRIORITY_LEVELS` would be a constant thats defined inside the existing class

martha: of course haha 🤦‍♀️

gigxz (approved):
Looks great!

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
gigxz: You could set it up like
```
class Hmis::ClientAlert < Hmis::HmisBase
  acts_as_paranoid

  HIGH = 'high'.freeze
  MEDIUM = 'medium'.freeze
  LOW = 'low'.freeze
  PRIORITY_LEVELS = [LOW, MEDIUM, HIGH].freeze

```

Then you would be able to access `Hmis::ClientAlert::LOW` directly.

martha: I like it, thanks for the suggestion!

ttoomey (approved):
Looks great!!

PR #3947, opened 2024-01-25
Download CSV for Upload of 2023 SPMs to HDX 2.0

awisema:

## Description

HUD has defined an upload format for SPMs to HDX 2.0. It is optional this year, but expected to be required for the next FY.
This adds the ability to download this file from a generated SPM.

The upload CSV has 3 parts, metadata, data from the SPM, and data from HMIS DQs for different project types. The DQ rows are included in the CSV, but are currently blank.

Also, there is a pending AAQ https://www.hudexchange.info/program-support/my-question/?askaquestionaction=public%3Amain.answer&key=C4650996-9192-45C5-AFB05D17757A7458 that may affect portions of the generated file.

## Type of change


drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_two.rb
```diff
@@ -20,13 +20,13 @@ def self.client_class
     def self.table_descriptions
       {
         'Measure 2' => 'The Extent to which Persons Who Exit Homelessness to Permanent Housing Destinations Return to Homelessness within 6, 12, and 24 months',
-        '2a and 2b' => 'The Extent to which Persons Who Exit Homelessness to Permanent Housing Destinations Return to Homelessness within 6, 12, and 24 months.',
+        # '2a and 2b' => 'The Extent to which Persons Who Exit Homelessness to Permanent Housing Destinations Return to Homelessness within 6, 12, and 24 months.',
```
eanders, line 23: Didn't you just update the exiting data to use `2a and 2b`?  I'm not following the logic in determining which it should or should not be and where.

eanders (approved):
I think this looks good.  Thanks @awisema !  I'd like to do a bit of local testing before merging.

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_two.rb
awisema: Table descriptions is used by the view to add titles to the view. Because M2 only has 1 table, we don't need both, and since the table had been incorrectly named, this second line didn't appear -- but having fixed the table name, the result was confusing.

PR #3952, opened 2024-01-26
Add mutation for creating a ClientAlert

martha:

## Description

This PR adds the mutation and test allowing creation of a ClientAlert through graphQL

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -35,6 +35,23 @@
     GRAPHQL
   end
 
+  let(:create_alert) do
+    <<~GRAPHQL
+      mutation CreateClientAlert($id: ID!, $note: String!, $expirationDate: ISO8601Date, $priority: ClientAlertPriorityLevel) {
```
martha: @gigxz the way `expirationDate` needed to be in camel case here really baffled me. When I had it in snake case, graphql didn't fail loudly, but the query would simply silently fail to have an expiration date by the time it got to the mutation ruby code (at which point it's in snake case anyway). How can I better understand where and why graphql does this weird magic with changing cases around?

gigxz: On this line, `$expirationDate` is just a variable name, it could be `$foo`. As long as you pass it to the `expirationDate` argument on the line below. The argument casing is `expirationDate` because graphql uses camel case. Ruby-graphql handles the camelizing of the Ruby-defined arguments and field names.

More on variables: https://graphql.org/learn/queries/#:~:text=.-,Variables

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
```diff
@@ -50,10 +67,26 @@
...
+      ) { create_alert }
+      expect(response.status).to eq(200), result.inspect
+      alert = result.dig('data', 'createClientAlert', 'clientAlert')
+      expect(alert).not_to be_nil
+      expect(alert['note']).to eq('raspberries')
+      expect(alert['priority']).to eq('high')
+      expect(Date.parse(alert['expirationDate'])).to eq((Date.current + 2.months))
+      expect(alert.dig('createdBy', 'id')).to eq(hmis_user.id.to_s)
+    end
   end
```
gigxz: This is alright, but I think it makes a bit more sense to structure the assertions around the AR model, rather than the resolved object.

I think a rule of thumb could be:
* _Query_ tests should ensure that fields are resolved correctly. (The test case above does this, on line 59). 
* _Mutation_ tests should ensure that the record is created and saved correctly. To do that, instead of asserting on the returned `data.createClientAlert.clientAlert`, you would assert on the active record model itself. We often do this in really simple way but just [ensuring that a record was created](https://github.com/greenriver/hmis-warehouse/blob/3b1d1318cf17e2cff1f55bacf75d727712a4eb09/drivers/hmis/spec/requests/hmis/update_bed_nights_spec.rb#L52-L61). But to make it more in-depth, you could assert on the created record: `(alert.priority).to eq('high')` etc.  It doesn't hurt to also check the returned object (at least the `id`) but I don't think it's the right place to assert each resolve field like this.


Does that make sense?

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
martha: thank you!

drivers/hmis/spec/requests/hmis/client_alert_spec.rb
martha: Thanks for this! I like this rule of thumb.

By this logic I'll keep lines 67-68 -- they are testing that the graphql results are returned in the correct order, which is happening in `drivers/hmis/app/graphql/types/hmis_schema/client.rb` -- but replace the other asserts with checks on the AR, if you agree w/that approach

drivers/hmis/app/graphql/mutations/create_client_alert.rb
```diff
@@ -0,0 +1,20 @@
+module Mutations
+  class CreateClientAlert < CleanBaseMutation
+    argument :input, Types::HmisSchema::ClientAlertInput, required: true
+
+    field :client_alert, Types::HmisSchema::ClientAlert, null: true
+    field :errors, [Types::HmisSchema::ValidationError], null: false, resolver: Resolvers::ValidationErrors
+
+    def resolve(input:)
+      raise 'not allowed' unless current_user.can_manage_client_alerts
```
martha: I borrowed this line from other examples in the codebase, but I was wondering, on the backend do we need to do more granular permission management -- where we make sure the user has permission to manage client alerts on this particular client? And if so, is there an example elsewhere in the codebase where we're doing that?

drivers/hmis/app/graphql/types/hmis_schema/client_alert_input.rb
```diff
@@ -0,0 +1,21 @@
...
+    description 'Client Alert Input'
+
+    argument :client_id, ID, required: true
+    argument :note, String, required: true
+    argument :expiration_date, GraphQL::Types::ISO8601Date, required: false
+    argument :priority, Types::HmisSchema::Enums::ClientAlertPriorityLevel, required: false
+
+    def to_params
+      h = to_h
+      h['created_by'] = current_user
```
martha: Now that we are using this `BaseInputObject`, is this the right place for adding the `created_by` attribute? (rather than in `mutations/create_client_alert.rb`, which is where I moved it out of)

drivers/hmis/lib/form_data/static/client_alert.json
```diff
@@ -0,0 +1,38 @@
...
+    {
+      "type": "TEXT",
+      "link_id": "note",
+      "text": "Alert Note",
+      "mapping": {
+        "field_name": "note"
+      }
+    },
+    {
+      "type": "DATE",
```
martha, line 25: Just confirming, the more elaborate date/duration picker component in the figma file is a non mvp requirement, correct?

And just out of curiosity, once we do take that on, how will we incorporate it since the form is generated from these json files? Will we need to create a new form type?

gigxz: Yes, non mvp. Since it ultimately collects a date value, I think it would be a new "component" we would need to add. Defined something like `"type": "DATE", "component": "DURATION_PICKER"`. The "type" should more or less define the data type thats collected, while the "component" is used to specify which presentational element to use (if different from the default). I think it would fall into this category but it might need some more thought.

drivers/hmis/app/graphql/types/hmis_schema/client_alert_input.rb
gigxz: I think adding it in the mutation itself is cleaner. it's not an input parameter after all.

drivers/hmis/app/graphql/mutations/create_client_alert.rb
gigxz: This is an unexpected state, so that's why we raise an exception. The frontend should never let the user attempt to create an alert if the user lacks permission to do so. It would be nice if we had some kind of error wrapper around authorization but we just haven't added that.

Look for current_permission there are examples for checking permissions against an "entity":
`current_permission?(permission: : can_manage_client_alerts, entity: client)`

drivers/hmis/app/graphql/types/hmis_schema/enums/client_alert_priority_level.rb
```diff
@@ -7,8 +7,8 @@ module Types
   class HmisSchema::Enums::ClientAlertPriorityLevel < Types::BaseEnum
     graphql_name 'ClientAlertPriorityLevel'
 
-    Hmis::ClientAlert::PRIORITY_LEVELS.each do |topic|
-      value topic
+    Hmis::ClientAlert::PRIORITY_LEVELS.each_with_index do |level, index|
+      value level, description: "#{index + 1} - #{level.titleize}"
```
gigxz, line 11: Nice 👍

gigxz (approved):

PR #3964, opened 2024-01-31
Coc Performance Measurement updates

eanders:

## Description

This adds household type calculations to the summary metrics and fixes a bug where the comparison "year" was the same as the report year.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/performance_measurement/app/models/performance_measurement/report.rb
+++ b/drivers/performance_measurement/app/models/performance_measurement/report.rb
@@ -383,7 +383,7 @@ module PerformanceMeasurement
               project_id: nil,
               for_question: parts[:key], # allows limiting for a specific response
               period: variant_name,
-              household_type: project_client[:household_type],
+              household_type: project_client.try(:[], :household_type),
             }
           end
         end
--- a/drivers/performance_measurement/app/models/performance_measurement/report.rb
+++ b/drivers/performance_measurement/app/models/performance_measurement/report.rb
@@ -278,7 +278,7 @@ module PerformanceMeasurement
     private def add_clients(report_clients)
       # Run CoC-wide SPMs for year prior to selected date and period 2 years prior
       # add records for each client to indicate which projects they were enrolled in within the report window
-      project_clients = Set.new
+      project_clients = {}
       involved_projects = Set.new
       run_spm.each do |variant_name, spec|
         spm_fields.each do |parts|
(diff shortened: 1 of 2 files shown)


drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -303,7 +303,8 @@ def can_see_client_details?(user)
               parts[:questions].each do |question|
                 report_client["#{variant_name}_#{question[:name]}"] = question[:value_calculation].call(member)
                 hud_project_ids.each do |project_id|
-                  project_clients << {
+                  project_clients[hud_client.id] ||= Set.new
+                  project_clients[hud_client.id] << {
```
awisema: Maybe move this into a method so it doesn't have to get repeated every time? Not worth a class, but `add_client(project_clients, { ... })` maybe?

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -366,14 +369,21 @@ def can_see_client_details?(user)
             client["#{variant_name}_#{parts[:key]}"] = value
             next unless value
 
+            # Use the previously calculated household_type, for now, just get the first for the client
+            # that matches one of the prior calculations
+            project_client = project_clients[client_id].detect do |pc|
```
awisema: Is this safe? Should it be
```suggestion
            project_client = project_clients[client_id]&.detect do |pc|
```

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -366,14 +369,21 @@ def can_see_client_details?(user)
...
-            project_clients << {
+            project_clients[client_id] ||= Set.new
+            project_clients[client_id] << {
               report_id: id,
               client_id: client_id,
               project_id: nil,
               for_question: parts[:key], # allows limiting for a specific response
               period: variant_name,
-              household_type: nil,
+              household_type: project_client[:household_type],
```
awisema: Same question --  either 379 will never fire, or I think this needs to be
```suggestion
              household_type: project_client.try(:[], :household_type),
```
?

awisema (approved):

drivers/performance_measurement/app/models/performance_measurement/report.rb
eanders: Well, it _should_ be safe, but probably worth catching.

drivers/performance_measurement/app/models/performance_measurement/report.rb
eanders: Yeah, I had similar thoughts...  will do.

awisema (approved):
Looks good thanks!

PR #3971, opened 2024-02-01
Form processor field validation

gigxz:

## Description

* Ensure that form submissions can only change fields that are mapped in the corresponding form definition.
* Update the Form Processor tests accordingly

[pt story](https://www.pivotaltracker.com/n/projects/2591838/stories/186955153)

## Type of change


gigxz:
Welp I have some more tests to fix.

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -427,4 +437,37 @@ def collect_record_validations(user: nil, household_members: nil)
 
     "#{container}.#{field}"
   end
+
+  # Ensure that a given field is valid for this FormDefinition
+  private def ensure_submittable_field!(container, field)
+    @mapped_form_fields ||= definition.link_id_item_hash.values.map(&:mapping).compact.
```
ttoomey: Suggestions:
* mapped_form_fields could be a lookup table rather rather than an array. A lookup table (hash/set) should scale better with the number of fields in the form. Probably doesn't matter much until we have 100s of fields but might as well build it to support that.
* Maybe break the method up into two parts? Personally I like to wrap memoized variables in their own methods with the same name, that makes it harder to accidentally create an instance variable naming collision

Maybe something like this?

```ruby
  # @return <Hash{container_name=> Set<fields> }>
  private def mapped_form_fields
    @mapped_form_fields ||= {}.tap do |result|
      definition.link_id_item_hash.each_value do |item|
        mapping = item.mapping
        next unless mapping

        # convert the record_type to a "container name" that matches the form processor names
        container_name = if mapping.record_type
          enum = Types::Forms::Enums::RelatedRecordType.values[mapping.record_type]
          raise "Invalid record type '#{mapping.record_type}'" unless enum

          enum.description
        else
          owner_container_name
        end

        result[container_name] ||= Set.new
        # only one of these can be present, right?
        result[container_name].add(mapping.field_name || mapping.custom_field_key)
      end
    end
  end

  # Ensure that a given field is valid for this FormDefinition
  private def ensure_submittable_field!(container, field)
    # Find the FormItem Mapping that matches this field.
    # If it's not found, then this is not a valid submission.
    found_mapping = mapped_form_fields[container]&.include?(field)

    raise "Not a submittable field for Form Definition id #{definition.id} (#{container}.#{field})" unless found_mapping
  end
```

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -11,7 +11,9 @@
 RSpec.describe Hmis::Form::FormProcessor, type: :model do
   include_context 'hmis base setup'
 
-  let!(:fd) { create :hmis_form_definition }
+  let(:fd) { Hmis::Form::Definition.find_by!(role: :INTAKE) }
+  let(:fd_exit) { Hmis::Form::Definition.find_by!(role: :EXIT) }
+  let(:fd_post_exit) { Hmis::Form::Definition.find_by!(role: :POST_EXIT) }
```
ttoomey, line 16: Thanks for updating the spec! Sorry there was so much churn. 

Shouldn't there be a test to check that hud_values which are not present in the form definition mappings result in an error or rejection? I don't see that test case here (but I might be missing it in this monster spec). I hate to recommend adding more code to this but I do think it would be a worthwhile to add a minimal test to ensure the restriction is applied.

ttoomey (commented):
Looking great, thanks for addressing this!!

I left a few suggestions and I think it might need an additional test. See comments

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
gigxz: yeah I will add for sure

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: thanks thats much cleaner

ttoomey (approved):
Looks great! Thanks for all the spec improvements! 🙌

drivers/hmis/spec/requests/hmis/assessments/submit_household_assessments_spec.rb
```diff
@@ -22,10 +22,10 @@
...
-  let!(:e2) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c2, user: u1, entry_date: '2022-01-01', household_id: e1.household_id, relationship_to_ho_h: 99 }
-  let!(:e3) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c3, user: u1, entry_date: '2022-01-01', household_id: e1.household_id, relationship_to_ho_h: 99 }
-  let!(:e4) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c4, user: u1, entry_date: '2022-01-01', relationship_to_ho_h: 99 }
+  let!(:e1) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, user: u1, entry_date: 2.weeks.ago }
+  let!(:e2) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c2, user: u1, entry_date: 2.weeks.ago, household_id: e1.household_id, relationship_to_ho_h: 99 }
+  let!(:e3) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c3, user: u1, entry_date: 2.weeks.ago, household_id: e1.household_id, relationship_to_ho_h: 99 }
+  let!(:e4) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c4, user: u1, entry_date: 2.weeks.ago, relationship_to_ho_h: 99 }
```
eanders, line 28: `2.weeks.ago` is going to give you a timestamp.  I _think_ you want `2.weeks.ago.to_date`

PR #3975, opened 2024-02-02
Validate dates in MPER importer

gigxz:

## Description

Cancel import if we get a bad date. Sometimes when they do an upgrade they start sending `30-JUN-24` it has happened twice before. Our HUD HMIS importer imports that as something like `0024-06-30`. Maybe it should be fixed downstream.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb
@@ -394,7 +394,7 @@ module HmisExternalApis::AcHmis::Importers
 
             # Abort import if we find a malformatted date. Dates like '30-JUN-24' would be incorrectly
             # parsed and lead to unexpected behavior in the HMIS.
-            raise(AbortImportException, "Incorrectly formatted date in #{file}.#{col}: #{r[col]}")
+            raise(AbortImportException, "Incorrectly formatted date in #{file} #{col}: #{r[col]}")
           end
         end
       end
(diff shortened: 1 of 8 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb
```diff
@@ -454,5 +470,19 @@ def infer_hud_version_from_project_cols
         '2022'
       end
     end
+
+    # Validate date format 'YYYY-MM-DD'
+    def valid_date?(str)
+      format_ok = str.match(/\d{4}-\d{2}-\d{2}/)
```
eanders, line 476: Might be worth writing a test.  Depending on how you fetch data from the CSV it may already have been converted to a ruby date object, which this probably wouldn't catch.  Also, we can probably take some other formats if you want to be more lax. There are some decent notes in `drivers/hmis_csv_twenty_twenty_four/app/models/hmis_csv_twenty_twenty_four/importer/import_concern.rb`

gigxz: Thanks for that link.

I tested out the function `fix_time_format` in the file you linked, it actually does parse `"01-JUL-20"` into the correct date. Maybe it cant handle other dates in that format, I don't know. But I would rather just keep it strict, it is supposed to be formatted as HUD data after all.

I did add a test case that uses the problematic date format.

eanders: Great, the test case seems more important, I'm all for enforcing the format.

eanders (approved):

PR #3976, opened 2024-02-02
Add user audit capability

martha:

## Description

We already can audit users' views of clients and enrollments; this adds the ability to view users' edit events too.
https://www.pivotaltracker.com/n/projects/2591838/stories/186570782

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/schema.graphql
```diff
@@ -192,6 +192,7 @@ User account for a user of the system
 """
 type ApplicationUser {
   activityLogs(limit: Int, offset: Int): ActivityLogsPaginated!
```
martha, line 194: I'm curious about this "activity logs," is it related to the user edit audit functionality I'm adding? I can see later in the class that the activity logs are only returned if the user has `can_audit_users` permission, so maybe yes?
I think it would be good for me to understand better what the relationship is, then maybe I can add some comments clarifying

gigxz: Yeah good question! The tldr difference is one is "access" history (Activity Logs) and "edit" history (PaperTrail versions).  We use the generic "audit" term to encompass both types of history (view and edit). But many places in the code ,"audit" just refers to  edit history, so it can get confusing.

1) Access History
The `hmis_activity_logs` table in the app DB stores logs about every GraphQL request that was made. That is pretty raw data so it's not helpful to show directly to the user, so, we do some additional processing in the `Hmis::ActivityLogProcessorJob` that populates `hmis_activity_logs_clients` and `hmis_activity_logs_enrollments`. Those 3 tables are ultimately what back the "Access History" page for a given user, which lets you see which clients and enrollments that a given user has **viewed**.


2) Edit History
The `versions` table in the warehouse DB stores [PaperTrail](https://github.com/paper-trail-gem/paper_trail) change history for any record that has `has_paper_trail` specified on the model. Each time a Client record is changed (for example), a new "version" record is added that has the before-and-after changed fields, along with who made the change.

     This table is what populates the "Client Audit History" and "Enrollment Audit History", as well as the view youre adding here.

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -30,6 +32,19 @@ class Application::User < Types::BaseObject
     field :date_created, GraphQL::Types::ISO8601DateTime, null: false
     field :date_deleted, GraphQL::Types::ISO8601DateTime, null: true
 
+    audit_history_field(
+      filter_args: { type_name: 'UserAuditEvent' },
```
gigxz:
```suggestion
      filter_args: { type_name: 'UserAuditEvent', omit: [:user] },
```

This will omit the `user` option from the filter options. It doesn't make much sense to filter by User since everything is related to the same user.

drivers/hmis/app/graphql/schema.graphql
martha: Great - thank you. I'll add some comments summarizing this to `user.rb` unless you think that it belongs elsewhere?

drivers/hmis/app/graphql/types/application/user.rb
martha: after implementing this change I see an error:
```
undefined method `user' for #<#<Class:0x0000ffff6b4cfe80>:0x0000ffff7c651d60 @context=#<Query::Context ...>, @ruby_style_hash={}, @arguments=#<GraphQL::Execution::Interpreter::Arguments @keyword_arguments={}>> scope = scope.where(v_t[:user_id].in(filters.user).or(v_t[:whodunnit].in(filters.user)).or(v_t[:true_user_id].in(filters.user))) if filters&.user&.present? ^^^^^^
```
I'll troubleshoot but if I haven't figured it out tomorrow maybe we can pair on this as well

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -30,6 +37,22 @@ class Application::User < Types::BaseObject
     field :date_created, GraphQL::Types::ISO8601DateTime, null: false
     field :date_deleted, GraphQL::Types::ISO8601DateTime, null: true
 
+    # audit_history returns the changes this user has made (as opposed to activity_logs which is just views, not edits).
+    # We use the generic term 'audit' to encompass both types of history (view and edit), but many places in the code,
+    # 'audit' just refers to edit history.
+    audit_history_field(
+      filter_args: { type_name: 'UserAuditEvent', omit: [:user] },
```
gigxz, line 46: ah I see the paper trail filter is not happy with this because `user` attribute is not defined on `input` anymore.

replace [this line](https://github.com/greenriver/hmis-warehouse/blob/c5f34ec587c96af1b99f7581242d3371f75fae1f/drivers/hmis/app/models/hmis/filter/paper_trail_version_filter.rb#L12) with the below fixes it, because the `with_filter` helper checks `respond_to?`

```
    with_filter(scope, :user) do
      scope = scope.where(v_t[:user_id].in(filters.user).or(v_t[:whodunnit].in(filters.user)).or(v_t[:true_user_id].in(filters.user)))
    end
```

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -8,6 +8,10 @@
 
 module Types
   class Application::User < Types::BaseObject
+    EXCLUDED_KEYS = ['owner_type', 'enrollment_address_type', 'wip'].freeze
```
martha: Is there a better place for this?

martha:
I want to confirm something I think we talked about this morning: update events should not be shown if the fields changed is going to be blank, but troubleshooting that is out of scope of this ticket for now, correct? Should I make a new ticket?

drivers/hmis/app/graphql/types/application/user.rb
martha: Also: Right now filtering out `wip` types appears to not be working, but I wasn't able to track down why; maybe we can pair more on this?

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -30,6 +39,26 @@ class Application::User < Types::BaseObject
...
+    audit_history_field(
+      excluded_keys: EXCLUDED_KEYS,
+      # filter_args: { type_name: 'UserAuditEvent' },
+      filter_args: { type_name: 'UserAuditEvent', omit: [:user] },
+    )
+
+    def audit_history(filters: nil)
+      v_t = GrdaWarehouse.paper_trail_versions.arel_table
+      scope = GrdaWarehouse.paper_trail_versions.
+        where(user_id: object.id).
```
gigxz: Whoops I think we actually want this to check `true_user_id`.
Actions that a user takes while "impersonating" a different user should show up in the REAL user's user audit history. For example if I impersonate you and create a client, the client creation should show up on my audit history (And not yours).

Does that make sense? To test it out, Admin>Users>Impersonate User. If its not there you need to enable the permission. Thanks!

drivers/hmis/app/graphql/types/application/user.rb
gigxz: Yes probably...these are enrollment-related, so I would think more likely `Types::HmisSchema::Enrollment::EXCLUDED_KEYS_FOR_AUDIT`. The User type here would reference that constant to exclude them (And to exclude the `Types::HmisSchema::Client::EXCLUDED_KEYS_FOR_AUDIT` keys as well, and so on). It feels messy but I think its ok for now.

gigxz: Ah I see. The Enrollment audit page uses that to filter out changes to the `CustomAssessment.wip` column, which is used to indicate whether an assessment is wip.

What you're seeing in the UI are probably changes to the `Hmis::Wip` _table_ which is used to store information about WIP ("incomplete") _Enrollments_. In order to get rid of that you need to add a `where.not(item_type: ['Hmis::Wip']).` (or `where.not(item_type: EXCLUDED_RECORD_TYPES).`) to the `audit_history` scope on this file. I can't think of any other record types to exclude right now, but they may come up.

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -54,6 +59,16 @@ def self.build(node_class, field_permissions: nil, excluded_keys: nil, transform
...
     end
 
+    def client_name
+      client = load_ar_association(object, :hmis_client)
+      client&.full_name
+    end
+
+    def project_name
+      project = load_ar_association(object, :hmis_project)
+      project&.ProjectName
```
gigxz: We have all the hud column names aliased to snake case, so we usually use those in the code
```suggestion
      load_ar_association(object, :hmis_project)&.project_name
```

gigxz:
> I want to confirm something I think we talked about this morning: update events should not be shown if the fields changed is going to be blank, but troubleshooting that is out of scope of this ticket for now, correct? Should I make a new ticket?

Correct. Actually it would be great if you can add a comment about [this ticket](https://www.pivotaltracker.com/story/show/186677366). Thanks!

drivers/hmis/app/graphql/types/application/user.rb
martha: Yes it makes sense, I made the change you suggested and tested it out locally, and it seems to be working correctly.

A new question: While I was testing in local dev, I noticed that the user audit screen isn't recording when someone creates a client alert. Is that expected? I can't see it in the `activity_logs` table either, but I'm not sure if that is the right table to check?

gigxz (approved):
💯

gigxz:
@martha looks like there is a test failure on the last build
```
Failures:

  1) User Audit History Query user who created a client should return audit history of user
     Failure/Error: expect(records.size).to eq(1)

       expected: 1
            got: 0

       (compared using ==)
     # ./drivers/hmis/spec/requests/hmis/user_audit_history_spec.rb:57:in `block (3 levels) in <main>'
     # /usr/local/bundle/gems/webmock-3.14.0/lib/webmock/rspec.rb:37:in `block (2 levels) in <main>'

Finished in 20 minutes 32 seconds (files took 8.41 seconds to load)
2004 examples, 1 failure, 14 pending

Failed examples:

rspec ./drivers/hmis/spec/requests/hmis/user_audit_history_spec.rb:55 # User Audit History Query user who created a client should return audit history of user
```

drivers/hmis/app/graphql/types/application/user.rb
gigxz: Paper Trail is not enabled on the `Hmis::ClientAlert` model, but it would probably be a good idea to add that. I created a ticket https://www.pivotaltracker.com/story/show/187012163 feel free to pick it up if you are looking for work 🙂

PR #3979, opened 2024-02-03
Improve root page caching concerns

ttoomey:

## Description

We've encountered weird latency issues recently (504s). I did some investigation and tried to make improvements to how we cache links and translations, which are both relevant for the warehouse home page.

**This work is speculative**, I do not have evidence that these changes address the issue we are experiencing. That being said, the translations do expire after 15 minutes which is sort of close to the observed 20 min frequency of the latency issues.

Details:
* improve Link class memoization, thread safety
* extract translation cache builder out of request lifecycle into a job. The idea is the job would be called periodically rather than expiring the cache keys. Include advisory lock to avoid simultaneous job runs
* use bulk-write operation to update cache translation cache
* use a digest for translation cache keys (maybe not necessary but seems safer than keying on user-defined strings)

## Type of change
- [x] performance fix?


app/jobs/build_translation_cache_job.rb
```diff
@@ -0,0 +1,25 @@
...
+class BuildTranslationCacheJob < BaseJob
+  LOCK_NAME = 'build_translation_cache'.freeze
+
+  def perform
+    Translation.with_advisory_lock(LOCK_NAME, shared: false, timeout_seconds: 0) do
+      translations = Translation.order(:id).pluck(:key, :text)
+
+      translations.in_groups_of(500, false) do |batch|
+        values = batch.to_h.transform_keys { |key| Translation.cache_key(key) }
+        Rails.cache.write_multi(values)
```
ttoomey, line 17: write_multi should be more efficient than previous implementation

app/models/link.rb
```diff
@@ -5,14 +5,19 @@
 ###
 
 class Link < ApplicationRecord
-  def self.for(location)
-    @cache ||= {}
+  @cache_mutex = Mutex.new
 
-    @cache[location] ||= where(location: location)
```
ttoomey, line 11: I don't think caching a scope/relation (as opposed to records) is a good idea

app/models/translation.rb
```diff
@@ -8,40 +8,32 @@ class Translation < ApplicationRecord
   include NotifierConfig
 
   def self.translate(text)
-    Rails.cache.fetch("translations/#{text}", expires_in: 15.minutes) do
```
ttoomey, line 11: This old implementation seems like it could cause a performance problem. My local db has ~1000 translation records so that's 1000 write operations to redis. Probably gets slower if multiple nodes miss at the same time and are all writing simultaneously

lib/tasks/grda_warehouse.rake
```diff
@@ -322,6 +322,8 @@ namespace :grda_warehouse do
 
     stats_collector = AppResourceMonitor::CollectStatsJob.new
     AppResourceMonitor::CollectStatsJob.perform_later if stats_collector.should_enqueue?
+
+    BuildTranslationCacheJob.perform_later
```
ttoomey, line 326: This would occur hourly instead of every 15min. Is that okay? Could also call this job from an after_commit hook on Translation

app/models/link.rb
```diff
@@ -5,14 +5,19 @@
 ###
 
 class Link < ApplicationRecord
-  def self.for(location)
-    @cache ||= {}
+  @cache_mutex = Mutex.new
```
ttoomey, line 8: A mutex seems safer when dealing with class ivars

app/models/translation.rb
```diff
@@ -8,40 +8,32 @@ class Translation < ApplicationRecord
...
-
-        # Add it to the database if we don't have it
-        Translation.where(key: text).first_or_create
-        Rails.logger.info(msg)
-      end
-
-      translated.presence || text
+      translation.text
     end
+    translated || text
```
eanders: Pretty sure we want 
```suggestion
    translated.presence || text
```

Otherwise if we delete the value from the UI (to restore the original) you'll get an empty string instead.
```
irb(main):001:0> "" || 'first'
=> ""
```

lib/tasks/grda_warehouse.rake
eanders: Yeah, this feels completely fine.  It should still update the particular translation after it has been saved, the every 15 minutes was really just a failsafe for if we missed updating the cache when changing a translation somehow.

app/models/translation.rb
ttoomey: Good catch, thanks!

app/jobs/build_translation_cache_job.rb
eanders: Thanks!

PR #3984, opened 2024-02-05
Add bool to specify whether a given CDED should be shown in summary

gigxz:

## Description

Flag to specify whether a custom field should show up in "summary " view. In most cases it's a question of whether is should be shown in the table row (as opposed to only visible when viewing the form).

Currently we are showing ALL custom data elements in the Service table. That s becoming untenable because we will have services collecting a lot of details, most of which do not need to show up in the table.

Here are 2 screen shot examples. For CurrentLivingSituations (and CustomCaseNote pending new PR) we show custom fields as additional columns. 

Screenshot 2024-02-05 at 6 13 56 PM
Screenshot 2024-02-05 at 6 13 30 PM

I'm not sure if this column will be useful for CDEDs collected on Custom Assessments. It could also be used to show a particular value in the Assessment table if desired. 

## Type of change
- [ ] Bug fix
- [ ] Dependency update


db/warehouse/migrate/20240205230723_add_display_hook_to_custom_data_element_def.rb
```diff
@@ -0,0 +1,5 @@
+class AddDisplayHookToCustomDataElementDef < ActiveRecord::Migration[6.1]
+  def change
+    add_column :CustomDataElementDefinitions, :show_in_summary, :boolean, default: false, null: false, comment: 'Whether to show this custom field in summary views such as in a table row when viewing a Service/CLS/Note'
```
ttoomey, line 3: Maybe the comment here could go on the graphql field description? I'm unsure of the value of putting documentation in sql comments but maybe I just don't have a good workflow for discovering them.

ttoomey (approved):
Code looks good. The only question I have is if we should try and design the field to meet additional use-cases. This might be scope creep though, feel free to ignore

I'm thinking of something like a 'visibility' field? Enums could be:
```
 summary: summary or high-level information
 detail: visible only on forms
 internal: should not be shown in the interface (legacy fields, ids)
 restricted: confidential values that should have additional restrictions (ssn, etc)
```

db/warehouse/migrate/20240205230723_add_display_hook_to_custom_data_element_def.rb
gigxz: Yeah I'll put the comment in gql. The col descriptions show up in TablePlus which I sometimes find helpful, but they aren't maintained, so it's not the best thing.

gigxz:
@ttoomey thanks for the suggestion. I think of it more as a "display hook" than a 'visibility' rule. 

There are other places we might want display "hooks" for custom fields:
* Show a custom client field prominently on the Client Dashboard (or even specify if it should go in the "demographics" accordion vs another section)
* Show a custom client field prominently on the Enrollment Dashboard
* Show a custom enrollment field prominently on the Enrollment Dashboard, or on the left nav card
* Show a custom field on project in the left nav card

These aren't immediate use-cases, but could come up... I agree we need something general, my concern with your suggestion is that it only allows for 1 "visibility" level to be specified. I think we may want to support multiple "hooks" which is why I took a column approach where we could add a col for each hook. Does that makes sense?

ttoomey:
> These aren't immediate use-cases, but could come up... I agree we need something general, my concern with your suggestion is that it only allows for 1 "visibility" level to be specified. I think we may want to support multiple "hooks" which is why I took a column approach where we could add a col for each hook. Does that makes sense?

Yes, that makes sense. Thanks for the explanation!

This is just a suggestion, I'm fine with leaving it as-is and we can revisit it later if needed. So I think using bool cols on the db table is fine. But for the API design, I wonder about using an array of enums, maybe something like this? 
```js
object.display_hooks: [
  "client_summary", // (or client_highlight)  Show a custom client field prominently on the Client Dashboard 
  "client_demographic_details",  // it should go in the "demographics" accordion vs another section
  "enrollment_summary" // Show a custom client field prominently on the Enrollment Dashboard
]
// then the query is just
common_data_element {
  display_hooks
}
```

I guess I like the idea of encapsulating these related flags in to one field to keep the object, schema, and queries a little cleaner, particularly if we end up with a lot of hooks. It is more complex than a 1:1 mapping though so maybe not worth the bother

gigxz:
Yeah I like that idea.  I'm a little unsure how to call this though, whether it should be split like `service_table_summary` `note_table_summary` `cls_table_summary`... that feels unnecessary. I think I will just call it one thing like `table_summary`.

ttoomey:
> Yeah I like that idea. I'm a little unsure how to call this though, whether it should be split like `service_table_summary` `note_table_summary` `cls_table_summary`... that feels unnecessary. I think I will just call it one thing like `table_summary`.

Yeah, the naming is hard. I was thinking it might be good to use names that are a bit fuzzy with respect to the specifics of the UI implementation. But I guess the UI and API are so tightly coupled it probably doesn't matter

ttoomey (approved):

PR #3990, opened 2024-02-07
Fix bug when processing boolean Custom Data Elements

gigxz:
## Description

Fix bug where changing a boolean custom data element value would sometimes remove the custom data element value.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/processors/base.rb
+++ b/drivers/hmis/app/models/hmis/hud/processors/base.rb
@@ -128,7 +128,7 @@ class Hmis::Hud::Processors::Base
     # If this custom field only allows 1 value and there already is one, update it.
     if !cded.repeats && existing_values.exists?
       cde_attributes = { id: existing_values.first.id }
-      if value.present?
+      if value.present? || value.instance_of?(FalseClass)
         cde_attributes[value_field_name] = value
         cde_attributes.merge!(attrs)
       else
--- a/drivers/hmis/app/models/hmis/hud/processors/base.rb
+++ b/drivers/hmis/app/models/hmis/hud/processors/base.rb
@@ -120,19 +120,21 @@ class Hmis::Hud::Processors::Base
       data_source_id: @processor.hud_user.data_source_id,
       data_element_definition: cded,
     }
+    # Normalize the input value
+    value = normalize_custom_field_value(cded, value)
+    # Infer the field name on the CustomDataElement
     value_field_name = "value_#{cded.field_type}"
-    value = attribute_value_for_enum(nil, value) # converts HIDDEN => nil
 
     existing_values = record.custom_data_elements.where(data_element_definition: cded, owner: record)
 
     # If this custom field only allows 1 value and there already is one, update it.
     if !cded.repeats && existing_values.exists?
       cde_attributes = { id: existing_values.first.id }
-      if value.present? || value.instance_of?(FalseClass)
+      if value.nil?
+        cde_attributes[:_destroy] = 1
+      else
         cde_attributes[value_field_name] = value
         cde_attributes.merge!(attrs)
-      else
-        cde_attributes[:_destroy] = 1
       end
     # If value(s) haven't changed, just update the User and timestamps
     elsif existing_values.map(&:value) == Array.wrap(value)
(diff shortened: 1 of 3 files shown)


drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1524,56 +1524,74 @@ def process_record(record:, hud_values:, user:, save: true)
...
+        [:string, :value_string, 'same', 'same'],
+        [:integer, :value_integer, 10, 0],
+        [:integer, :value_integer, 0, 10],
+        [:integer, :value_integer, 0, 0],
+        [:boolean, :value_boolean, true, false],
+        [:boolean, :value_boolean, true, true],
+        [:boolean, :value_boolean, false, false],
+        [:boolean, :value_boolean, false, true],
+      ].each do |field_type, field_name, old_value, new_value|
+        it "updates a CustomDataElement on an existing record (#{field_type}, #{old_value}=>#{new_value}) (repeats: false)" do
```
gigxz, line 1588: I added a bunch but the specific regression tests are:
```
      updates a CustomDataElement on an existing record (boolean, true=>false) (repeats: false) (FAILED - 1)
      updates a CustomDataElement on an existing record (boolean, false=>false) (repeats: false) (FAILED - 2)
```

These would previously destroy the CDE.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -128,7 +128,7 @@ def process_custom_field(field, value)
     # If this custom field only allows 1 value and there already is one, update it.
     if !cded.repeats && existing_values.exists?
       cde_attributes = { id: existing_values.first.id }
-      if value.present?
+      if value.present? || value.instance_of?(FalseClass)
```
eanders: can you just?
```suggestion
      unless value.nil?
```

gigxz: I don't really want to save `""` strings, thats the main reason why I left the presence check

gigxz: Is there a better railsy way to do that?

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -128,7 +128,7 @@ def process_custom_field(field, value)
     # If this custom field only allows 1 value and there already is one, update it.
     if !cded.repeats && existing_values.exists?
```
ttoomey: This is scope creep but it looks like this condition will result in a lot of extra queries. Maybe we can add this to the backlog to improve. 

This method is run for each field which will execute one or two queries against record.custom_data_elements (`exists` and `first` will each be a query).

Suggest something like the following (untested code)
```ruby
  def existing_values_by_definition_and_owner
    @existing_values_by_definition_and_owner ||= record.custom_data_elements.order(:id).group_by do |element|
      [element.data_element_definition_id, element.owner_name, element.owner_id]
    end
  end

  def process_custom_field
    # ...
    existing_values = existing_values_by_definition_and_owner[cded.id, record.sti_name, record.id]
```

drivers/hmis/app/models/hmis/hud/processors/base.rb
ttoomey: Gig I think what you have is fine although maybe it could use a comment.

However I'm a little unsure as to what value might be at this point. Sounds like it could be a string or a boolean? The code suggests that it might also be an array sometimes?

If the value is not well normalized already, maybe we should do that? Please ignore this if it's already the case. Thinking about something like the following case statement tree that will error if we see something unexpected (untested)

```ruby
  def process_field 
    #...
    value = normalize_input_value(cded, value)
    #...
    if value.nil?
      cde_attributes[:_destroy] = 1
    else
      cde_attributes[value_field_name] = value
      cde_attributes.merge!(attrs)
    end
  end

  def normalize_input_value(cded, value)
    ## do we get arrays here?
    return value.map { |v| v.normalize_input_value(cded, v)}.compact.presence if value.is_a?(Array)

    case cded.field_type
    when 'string', 'text'
      case value
      when String
        value&.presence&.strip
      else
        raise "unexpected value \"#{value}\" for #{field}"
      end
    # when 'json', 'integer', 'float', 'date'
    when 'boolean'
      case value
      when nil, ""
        nil
      when true, false
        value
      ## do we get strings here? ("true", "false")
      # when String
      #   ActiveModel::Type::Boolean.new.cast(value)
      else
        raise "unexpected value \"#{value}\" for #{field}"
      end
    else
      raise "unknown field_type #{field_type}"
    end
  end
```

ttoomey (approved):
Looks good!! Left a few suggestions but nothing that should block this bug fix

drivers/hmis/app/models/hmis/hud/processors/base.rb
eanders: @gigxz sorry for the delay, I'm with @ttoomey, this is fine as is, but checking `instance_of?` always rubs me the wrong way, unless there's some specific set of types we expect, and then, I'm with @ttoomey we should handle the different types as a case statement, or at least intentionally.   Happy to just keep it as is, but maybe a comment about why it's checking for a boolean or blank?

eanders (approved):

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: Thanks, I can normalize it like that, its clearer to read. There is a validation to make sure the definition type and the value type match, so it does error if they dont match.

gigxz: Is `when String` not a code smell but `instance_of?` is? Just wanna get my railsing right here.. to me these seem the same

gigxz: I think feedback is addressed @ttoomey can you take a look?

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: https://github.com/greenriver/hmis-warehouse/pull/3990/commits/048d2a6c723cc1d797baf005da924247b681aaad
thanks!

drivers/hmis/app/models/hmis/hud/processors/base.rb
eanders: For me it was more of separating the logic from the type determination.  So normalizing the thing you are checking is one operation, then determining validity of the normalized data becomes more of a business operation. It's a very weak smell if you ask me, but definitely something that caught my eye.

PR #3991, opened 2024-02-07
TC CAS Readiness

awisema:

## Description

- CAS Calculator for TC HAT that references the HUD Assessment table instead of HMIS Forms
- Background processor to create HUD AssessmentQuestions after the FormProcessor runs.

## Type of change


drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
```diff
@@ -0,0 +1,99 @@
...
+
+    def form_definition
+      @form_definition ||= {}.tap do |h|
+        @order = 1
+        definition = form_processor.definition.definition['item']
+        h.merge!(parse(definition))
+      end
+    end
+
+    private def parse(definition, group: nil)
```
eanders, line 80: @awisema can you add a comment here about what this is doing?

app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
```diff
@@ -0,0 +1,315 @@
...
+      ]
+    end
+
+    def most_recent_assessment_for_destination(client)
+      return unless client.present?
+
+      cas_assessment_collected_at(client)&.to_date&.to_s
+    end
+
+    private def cas_assessment(client)
```
eanders, line 171: Should we memoize this? 
```
@cas_assessment ||= ...
```

app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
```diff
@@ -0,0 +1,315 @@
...
+        max_by(&:assessment_date)
+    end
+
+    private def for_string(client, key)
+      cas_assessment(client).assessment_questions.find_by(assessment_question: assessment_keys[key])&.assessment_answer
+    end
+
+    private def for_boolean(client, key)
+      for_string(client, key)&.downcase == 'yes'
+    end
```
eanders: Does it makes sense to move these `for_boolean` and `for_string` methods up to `AssessmentQuestion`?

eanders (approved):
Looks good to me.  Thanks @awisema !

app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
awisema: The CAS calculator gets reused in the sync process, so not without including the client -- I added `memoize` -- but if this uses too much memory, we may need to rethink.

eanders: oh, sorry, yeah, that makes sense.

drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
```diff
@@ -0,0 +1,103 @@
...
+          form_processor.hud_values.each do |key, value|
+            next if key.include?('.') # Field is stored in another HUD object
+            next if value == HIDDEN_FIELD_VALUE
+
+            value = case question_type(key)
+            when 'BOOLEAN'
+              if value
+                'Yes'
+              else
+                'No'
```
gigxz, line 31: This would represent a `nil` value as `No`, I dont think that is desirable? There are lots of non-required questions

drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
```diff
@@ -0,0 +1,103 @@
+###
```
gigxz, line 1: So you asked about testing.. I think an rspec that tests `Hmis::AssessmentQuestionsJob` is enough? 

For setup:
* to build the assessment with the form processor and values: `build(:hmis_wip_custom_assessment, hud_values: ...)`
* to build the form definition: we don't have a good way of doing this right now. I just proposed a new approach in https://github.com/greenriver/hmis-warehouse/pull/3995 where we can add test definitions in the `form_data/test` environment (it gets loaded because of env var `CLIENT=test`). You could add another one that has just the specific HAT questions that are used for CAS readiness. Does that work? (you can pull out the `item`s from https://qa-hmis.openpath.host/admin/forms/70/edit..)

gigxz (approved):

drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
gigxz: On second thought, we should match whatever the previous behavior was, which I'm sure is what you are doing here :)

PR #3997, opened 2024-02-08
Add Signature dynamic field component

martha:

## Description

This begins to add support for e-signatures in forms: https://www.pivotaltracker.com/n/projects/2591838/stories/186786179

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/lib/form_data/tarrant_county/services/service.json
```diff
@@ -0,0 +1,228 @@
+{
```
martha: Note- This file should not be merged as is

drivers/hmis/lib/form_data/default/fragments/signature.json
```diff
@@ -0,0 +1,100 @@
...
+  "text": "eSignature",
+  "subtitle": "The following signatures are required on this document.",
+  "item": [
+    {
+      "type": "GROUP",
+      "text": "Head of Household",
+      "link_id": "head_of_household_signature_group",
+      "item": [
+        {
+          "type": "STRING",
```
martha, line 15: One result of having these be separate inputs is that I don't think there's any existing example in `DynamicField` of having 2 inputs next to each other on a line when they are separated like this in the json. The comparable examples I could find, such as the Name input, use `"type": "OBJECT"`. Maybe we should take that approach here, as well? (this approach could also help us out with the question of how to display the checkbox if it doesn't correspond to any field on the backend.)

drivers/hmis/lib/form_data/default/fragments/signature.json
```diff
@@ -0,0 +1,100 @@
...
+          "mapping": {
+            "custom_field_key": "head_of_household_signature"
+          }
+        },
+        {
+          "type": "DATE",
+          "link_id": "head_of_household_signature_date",
+          "text": "Today's Date",
+          "required": true,
+          "enable_when": [
```
martha, line 30: @ttoomey when we paired on Friday we had discussed this method of requirement validation, where some signatures are required and others are optional, and the date is always required if the signature is non-null.

I think I set it up the way we had discussed, and I'm wondering now if this is really the desired behavior. One thing I noticed about it is that this way the "Required" label gets rendered next to both the signature and the date -- whereas in Ash's designs, the label is rendered next to the heading ("Head of Household" or "Case Manager") which makes a little more sense from a ux perspective, I think.

drivers/hmis/app/graphql/types/forms/enums/variant.rb
```diff
@@ -0,0 +1,14 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
```
ttoomey: can remove this magic comment, it only works if it is on the first line of the file. So it get clobbered by whatever script adds copyright info 🙄

drivers/hmis/app/graphql/types/forms/enums/variant.rb
```diff
@@ -0,0 +1,14 @@
...
+
+module Types
+  class Forms::Enums::Variant < Types::BaseEnum
+    graphql_name 'Variant'
```
ttoomey: I think calling this "variant" is too generic. Suggest a more specific name, maybe 'FormGroupVariant'?

drivers/hmis/lib/form_data/default/fragments/signature.json
```diff
@@ -0,0 +1,100 @@
+{
+  "type": "GROUP",
+  "variant": "SIGNATURE",
+  "link_id": "signature_group",
+  "text": "eSignature",
+  "subtitle": "The following signatures are required on this document.",
```
ttoomey: This content seems a little more like helper text than a subtitle. Maybe this should be named "helper_text" instead? (We already have that property I think)

drivers/hmis/lib/form_data/default/fragments/signature.json
```diff
@@ -0,0 +1,100 @@
...
+          "required": true,
+          "mapping": {
+            "custom_field_key": "head_of_household_signature"
+          }
+        },
+        {
+          "type": "DATE",
+          "link_id": "head_of_household_signature_date",
+          "text": "Today's Date",
+          "required": true,
```
ttoomey, line 27: Maybe we should use:
```json
 "disabled_display": "PROTECTED"
```
This always shows the date field but it is disabled if there's no signature

drivers/hmis/lib/form_data/default/fragments/signature.json
ttoomey: It's a good point. I could see an argument either way although TBH I would lean more towards keeping the required label next to the field (as you have it). Repeating the text is seems more clear and it is consistent with the treatment of other required fields in the app

drivers/hmis/lib/form_data/default/fragments/signature.json
martha: This does have the effect of showing a greyed-out "Required" label, which may be confusing? But I think I still like it better than having the form dynamically render the date input.

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -4022,6 +4022,11 @@ type FormItem {
   text: String
   type: ItemType!
 
+  """
+  Component variant to use for display
+  """
+  variant: Variant
```
gigxz: @martha @ttoomey I'm wondering about the difference between `variant` and `component` on the Form Item?

ttoomey: Good point, maybe they are redundant. My own understanding of the`component` prop is probably incorrect; I imagined it was to describe behavioral/functional differences (checkbox vs radio) rather then describing cosmetics. However I totally buy that distinction doesn't matter and consolidating around a single property is a better way to go.

gigxz: I see. The `component` prop should be 100% cosmetic. I can see the confusion with radio buttons, but it doesn't actually work that way. The `repeats: true` is the deciding factor on whether something ends up being a single-select or multi-select input.

martha: just to confirm, it sounds like I should change this back to using `component`, and remove `variant`?

ttoomey: Sounds like it! Sorry for the confusion and thanks for clarifying @gigxz

ttoomey (approved):
Looks good! Probably remove the `db/warehouse_structure.sql` thought since there are no migrations in this branch

drivers/hmis/app/graphql/schema.graphql
gigxz: Yes! I am sorry for the delay! This looks great, I am going to re-target to release-104 and merge. Thanks both 🖊️ 👍

PR #4008, opened 2024-02-10
Attempt to get importer tests working on new path

eanders:

## Description

DataLab test kit adjustments
* The Data Lab released a new and fixed full source data set 2/8, this is included
* Tests weren't actually running in CI because the path on S3 was incorrect, fixed
* Re-checked to confirm tests are only off by a few where xit'ed
* Added first-half of CE-APR logic to exclude projects that do not have CE-Participation statuses
 
## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hud_apr/spec/models/fy2024/datalab_2_0_spec.rb
+++ b/drivers/hud_apr/spec/models/fy2024/datalab_2_0_spec.rb
@@ -43,7 +43,7 @@ RSpec.describe 'Datalab 2024', type: :model do
   if File.exist?('drivers/datalab_testkit/spec/fixtures/inputs/merged/source/Export.csv')
     include_context 'datalab multiple projects apr'
     include_context 'datalab organization a rrh apr'
-    # include_context 'datalab organization o sso apr'
+    include_context 'datalab organization o sso apr'
     # include_context 'datalab organization o th apr'
     # include_context 'datalab organization s rrh apr'
     # include_context 'datalab organization t psh apr'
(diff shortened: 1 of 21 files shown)


dtgreiner (approved):

drivers/hud_apr/spec/models/fy2024/datalab_2_0_spec.rb
```diff
@@ -43,7 +43,7 @@ def run(generator, filter)
   if File.exist?('drivers/datalab_testkit/spec/fixtures/inputs/merged/source/Export.csv')
     include_context 'datalab multiple projects apr'
     include_context 'datalab organization a rrh apr'
-    # include_context 'datalab organization o sso apr'
+    include_context 'datalab organization o sso apr'
     # include_context 'datalab organization o th apr'
```
awisema, line 47: If the goal is to avoid accidentally missed tests, maybe we should push this down into `xit`s or something instead of commenting them out? Not ideal, but at least we get a message...

drivers/hud_spm_report/app/views/hud_spm_report/measures/result.haml
```diff
@@ -5,7 +5,8 @@
   = "#{report_short_name} #{@question}: #{@generator.describe_table(@question)}"
 
 %h1= content_for :title
-
+.well= debug @report
+.well= debug @question
```
awisema:
```suggestion
```

awisema (changes requested):

drivers/hud_spm_report/app/views/hud_spm_report/measures/result.haml
eanders: Thank you!

drivers/hud_apr/spec/models/fy2024/datalab_2_0_spec.rb
eanders: Yes, but these haven't been QA'd at all yet.  Once we've completed the first pass, we should uncomment them.

PR #4011, opened 2024-02-12
Support conditionally collected related HMIS records on CustomAssessment

gigxz:

## Description

**Desired/Implemented behavior**: Support assessments that _conditionally_ collect related records. For example, a Custom Assessment that optionally collects a CE Event or CurrentLivingSituation.
  * If all the fields for the CE Event are _hidden_, then the CE Event record should NOT be created.
  * If all the fields for the CE Event are _hidden_, and there was a previously associated CE Event record from this assessment, then that CE Event should be **destroyed**.
  * If any of the fields for the CE Event are _NOT hidden_, then the CE Event should be created. (If any required fields for CE Event are left empty, it should validate).

**Previous behavior**: Attempts to create CE Event record even when all fields are hidden. This results in an error submitting the assessment because Event.EventDate is required.

**Implementation Approach**: After processing all fields, _remove_ records for which all fields are hidden.

## Type of change
- [x] Bug fix


drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -69,16 +69,46 @@ def run!(owner:, user:)
...
-    relevant_container_names.each do |container|
-      container_processor(container)&.assign_metadata
-      container_processor(container)&.information_date(custom_assessment.assessment_date) if custom_assessment.present?
+    # Iterate through each container (record type) to apply metadata and information dates
+    hud_values_by_container.keys.each do |container|
+      processor = container_processor(container)
+
+      # If this is an assessment and all fields pertaining to this record type were hidden,
+      # the related record should be destroyed. (For example, a Custom Assessment that conditionally creates a CE Event).
+      if custom_assessment.present? && containers_with_all_fields_hidden.include?(container) && !['Assessment', 'Enrollment'].include?(container)
```
eanders: Is there a better way to write this? 
```
!['Assessment', 'Enrollment'].include?(container)
```

Can you ask the container if it's custom or something?  Just trying to think about the future and what other classes may have required dependent items, seems somewhat easier to follow if the container can answer that question.

gigxz: Yes. I put this in at the last minute because I got afraid of unintentional deletions. I think it would be better if we mark on the processor whether its ALLOWED to be a conditional record. Then we can just turn it on for CE Event, Ce Assmt, and CLS.

eanders (approved):

drivers/hmis/app/models/hmis/form/form_processor.rb
eanders: could this be implemented as:

```
container.dependent_destroyable?
``` 

or similar?

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1899,13 +1899,146 @@ def add_cde_item_to_definition
...
+          # link 1 of the cls to an assessment
+          assessment.save!
+          assessment.form_processor.update(current_living_situation: cls1)
+
+          # check setup
+          expect(assessment.current_living_situation).to eq(cls1)
+          expect(assessment.form_processor.current_living_situation).to eq(cls1)
+
+          # "re-submit" the assessment, this time with all the event fields hidden
+          expect { process_assessment }.to change(e1.current_living_situations.reload, :count).by(-1)
```
ttoomey, line 2034: I don't think you need to reload since relation.count always performs a query afaik (different than relation.size which will use the cache)

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -26,10 +26,14 @@ def seed_all
     protected
 
     def env_key
-      return ENV['CLIENT'] if ENV['CLIENT'].present?
-
-      # default to QA environment in development to get forms with all possible questions enabled
-      'qa_hmis' if Rails.env.development?
+      @env_key ||= if Rails.env.test?
```
ttoomey, line 29: This seems like a very inexpensive method. Is memoization worth the bother?

ttoomey (approved):
Looks great!

drivers/hmis/app/models/hmis/form/form_processor.rb
ttoomey: If this is just an extra safety check, I suggest it should be an exception rather than part of the condition. something like:
```ruby
if custom_assessment.present? && containers_with_all_fields_hidden.include?(container) 
 raise "unexpected container #{container}" if ['Assessment', 'Enrollment'].include?(container)
 processor.destroy_record
```

drivers/hmis/lib/hmis_util/json_forms.rb
gigxz: not really but I had my print statements in there and I was bothered by the repetition.. I was having trouble getting `CLIENT=test` to be consistently set in the test environment, which is why I switched it to check `Rails.env.test?` instead

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
gigxz: ah ok thanks

PR #4014, opened 2024-02-12
186994227-log-graphql-api-response-time

ttoomey:

## Description

[186994227](https://www.pivotaltracker.com/story/show/186994227)
* capture response time in activity log
* improve instrumentation approach for logging so there's less indirection (use our own tracing impl rather than using active support notification)
* remove unused Yabeda graphql gem (we installed the gem but never enabled it for some reason?). I reviewed this gem and had concerns with it; best to remove it imo
* disable rubocop numeric predicate style cop (the value of this cop seems questionable to me and it broke a simple rspec test) 

## Type of change


.rubocop.yml
```diff
@@ -248,6 +248,8 @@ Style/KeywordParametersOrder:
   Enabled: false
 Style/MethodDefParentheses:
   Enabled: false
+Style/NumericPredicate:
+  Enabled: false
```
ttoomey, line 252: @eanders is this okay with you? It means both `x > 0` and `x.positive?` will pass rubocop

gigxz (approved):

.rubocop.yml
eanders: Yeah, that's fine.  There are definitely times `.positive?` is more readable, but depending on the situation, it can make a set of tests less clear.

Gemfile
```diff
@@ -113,7 +113,6 @@ gem 'logstop'
 gem 'yabeda-rails'
 gem 'yabeda-prometheus'
 gem 'yabeda-puma-plugin'
-gem 'yabeda-graphql'
```
eanders, line 116: Is this gem no longer helpful?

eanders (approved):

Gemfile
ttoomey: @eanders My understanding is that this gem doesn't do anything unless it is enabled so I think it is currently a no-op. (according to the docs, you have to add `use Yabeda::GraphQL` to your schema).

We could enable it but I'm not sure this gem is something we should be using in production (it's not widely used, pre-1.0, written targeting an older version or ruby graphql).

PR #4015, opened 2024-02-12
Add hide_when_x fields

gigxz:

## Description

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/schema.graphql
```diff
@@ -3950,6 +3950,16 @@ type FormItem {
   """
   hidden: Boolean!
 
+  """
+  Whether to hide the item when form is in view-mode
```
gigxz, line 3954: fix comment

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -3950,6 +3950,16 @@ type FormItem {
...
+  """
+  hideWhenEditing: Boolean!
+
+  """
+  Whether to hide the item when form is in view-mode
+  """
+  hideWhenViewing: Boolean!
+
```
gigxz, line 3962: What do you think of this @ttoomey ? I wanted to go with something simpler, rather than incorporating it as an "enable_when" condition, since it's conditional upon the context (as opposed to conditional on other questions or variables).

It feels messy though somehow. Another approach might be
`visibility: 'VIEWING' | 'EDITING' | 'BOTH' (default=BOTH)`
or
`hidden: VIEW_MODE | EDIT_MODE | ALWAYS (default=null=never)` to replace current `hidden: boolean` field

gigxz: I also dont really have a use case for a field that should be hidden when Editing. But I can imagine maaaybe fields that are only relevant for printing, like showing some client or program details, for example..

PR #4017, opened 2024-02-13
Add pathways upload feed

gigxz:

## Description

Upload feed for "pathways" custom elements

https://www.pivotaltracker.com/story/show/186955457

## Type of change


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/pathways_export.rb
```diff
@@ -0,0 +1,132 @@
...
+          find_pathway(pathways, 'client_pathway_3_narrative'),
+          find_pathway(pathways, 'client_pathway_3', date_updated: true),
+        ]
+        write_row(values)
+      end
+    end
+
+    private
+
+    def find_pathway(pathways, key, date_updated: false)
```
eanders: Generally speaking we tend to write these:
```suggestion
    private def find_pathway(pathways, key, date_updated: false)
```
That way the method can be moved around as necessary and private-ness isn't controlled by location in the file.

drivers/hmis_external_apis/lib/tasks/export.rake
```diff
@@ -11,5 +11,6 @@ namespace :export do
     HmisExternalApis::AcHmis::DataWarehouseUploadJob.perform_later('project_crosswalk')
     HmisExternalApis::AcHmis::DataWarehouseUploadJob.perform_later('move_in_addresses')
     HmisExternalApis::AcHmis::DataWarehouseUploadJob.perform_later('postings')
+    HmisExternalApis::AcHmis::DataWarehouseUploadJob.perform_later('pathways')
```
eanders: This is fine, but it's starting to feel like:
```
[
  'clients_with_mci_ids_and_address',
  'hmis_csv_export',
  'project_crosswalk',
  'move_in_addresses',
  'postings',
  'pathways',
].each do |export|
  HmisExternalApis::AcHmis::DataWarehouseUploadJob.perform_later(export)
end
```
might be worth doing.

eanders (approved):

drivers/hmis_external_apis/lib/tasks/export.rake
gigxz: It was a nice idea but we are doing them all at the same time. I'm just going to make the argument optional so if its not passed, it does all of them. sound ok?

gigxz: or maybe there was some reason Todd did it this way so that they would run in separate Delayed Jobs, and be queued separately? It seems fine to me to do them all at once. The HMIS HUD Export and the Client.csv export are the bigger ones.

eanders: calling `.perform_later` just queues them up in delayed job, so as written it will generate a job for each one. Looping over the array will not change the functionality, only cleanup the code duplication.

eanders:
Looks good to me!

eanders:
Ugh, hit the wrong button.

eanders:
Merge whenever you are ready.

drivers/hmis_external_apis/lib/tasks/export.rake
gigxz: I meant if I combined them so **all** the exports would run with only one call to `perform_later`. I'll just leave it as you suggested though.

PR #4025, opened 2024-02-14
Auto exit bug fix: release unit and notify LINK

gigxz:

## Description

* Release assigned unit when enrollment is auto-exited
* Notify LINK when enrollment is auto-exited
* Add a relation to ReferralRequest to make cleanup in training env easier (unrelated to auto exit)

## Type of change
- [x] Bug fix


drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -83,6 +83,11 @@ def auto_exit(enrollment, most_recent_contact)
       )
       assessment.build_form_processor(exit: exit_record)
       assessment.save!
+
+      # Release the unit that was assigned to this Enrollment (if applicable)
+      enrollment.release_unit!(exit_date, user: system_user)
```
gigxz, line 88: This queues up a bunch of jobs, one per exited enrollment (if AC HMIS Extension is enabled). That could become problematic. 

There shouldn't be very many enrollments exited per day (est 10-20) *Except* for when auto-exit is turned on for a project for the first time, in which case there could be hundreds)

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -83,6 +83,11 @@ def auto_exit(enrollment, most_recent_contact)
...
+      # Close referral in External LINK system (if applicable)
+      enrollment.close_referral!(current_user: system_user)
```
gigxz, line 90: This could also be problematic since it makes an API request to external system once per enrollment

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: I think this will help https://github.com/greenriver/hmis-warehouse/pull/4025/commits/c38b206b8c8dd39a5456d2c4598477b86c95bf57

ttoomey: What if we modified enrollment.release_unit and track_availability so that it didn't always call UpdateUnitAvailabilityJob? Then we could defer the job until later. Something like:

```ruby
def auto_exit
  #...
  enrollment.release_unit!(exit_date, user: system_user, skip_update_availability_job: true)
end

def perform
  # ...
  Hmis::Hud::Project.hmis.each do |project|
    #...
  end
  HmisExternalApis::AcHmis::UpdateUnitAvailabilityJob.perform_now
  
```

drivers/hmis_external_apis/extensions/hmis/unit_type_extension.rb
```diff
@@ -32,9 +32,13 @@ def track_availability(project_id:, user_id:)
           user_id: user_id,
           unit_type_id: id,
         )
-        HmisExternalApis::AcHmis::UpdateUnitAvailabilityJob
-          .set(wait: 1.minute) # short wait to accumulate batch of changes before update
-          .perform_later
+
+        # Don't re-queue if job is already queued
+        return if Delayed::Job.queued?('HmisExternalApis::AcHmis::UpdateUnitAvailabilityJob')
```
ttoomey, line 37: Looks like there's a chance for race conditions here. Maybe it doesn't matter if this is called frequently and so dropped jobs are just delayed for a short time.

ttoomey (approved):
I think this is should work most of time time, I had a suggestion about a different solution but not sure if it's worth the bother (see comment)

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
eanders: Does `HmisExternalApis::AcHmis::UpdateUnitAvailabilityJob` update all units?  I think c38b206 should be good enough if it does.

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
eanders: As this is a background job, unless there's a batch API, I'm guessing that's what we got, and it's probably fine.

drivers/hmis_external_apis/extensions/hmis/unit_type_extension.rb
eanders: I can't remember, but I think `queued?` looks for one that hasn't started.  If it has started, it should queue up the next one, so I can see a situation where there may be more than one in the queue, I guess we should make sure they are using an advisory lock so they don't try to process the same data at the same time.  

trying to determine if there's a race condition where the job gets picked up while it's checking, so it appears queued, but is actually running and for whatever reason misses the newly dirtied records.....

eanders (approved):

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: Yup we didnt get a batch api. I think its ok too.

drivers/hmis_external_apis/extensions/hmis/unit_type_extension.rb
gigxz: Yes, `queued?` only looks for jobs that have started.

If there is a potential race condition here, then it would also affect the service history processing that queues in an after_save hook using the same mechanism.


I'm going to go ahead and merge this, as I don't think it's a concern if there is an occasional delay with unit changes.

PR #4027, opened 2024-02-15
Update Core Demographic report to include HIV counts without Indefinite and Impairs. Prevent details from being viewed when user does not have access.

dtgreiner:

## Description

Update the Core Demographics report to pull in HIV data without Indefinite and Impairs responses. Also set the disability details to be viewable only by users with the required permission.

## Type of change
- [X] Bug fix

### DIFF
--- a/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
@@ -35,7 +35,7 @@ module
     def detail_scope_from_key(key)
       detail = detail_hash[key]
       return report_scope.none unless detail
-      return report_scope.none if detail[:cannot_view_details].present?
+      return report_scope.none if ! detail[:can_view_details].nil? && ! detail[:can_view_details]
 
       detail[:scope].call.distinct
     end
--- a/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
@@ -35,6 +35,7 @@ module
     def detail_scope_from_key(key)
       detail = detail_hash[key]
       return report_scope.none unless detail
+      return report_scope.none if detail[:cannot_view_details].present?
 
       detail[:scope].call.distinct
     end
--- a/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/details.rb
@@ -35,7 +35,7 @@ module
     def detail_scope_from_key(key)
       detail = detail_hash[key]
       return report_scope.none unless detail
-      return report_scope.none if ! detail[:can_view_details].nil? && ! detail[:can_view_details]
+      return report_scope.none if detail[:can_view_details] == false # nil and true should be allowed.
 
       detail[:scope].call.distinct
     end
--- a/drivers/core_demographics_report/app/models/core_demographics_report/disability_calculations.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/disability_calculations.rb
@@ -13,6 +13,7 @@ module
         HudUtility2024.disability_types.each do |key, title|
           hashes["disability_#{key}"] = {
             title: "Disability #{title}",
+            cannot_view_details: ! can_view_client_disability?(@filter.user, key),
             headers: client_headers,
             columns: client_columns,
             scope: -> { report_scope.joins(:client, :enrollment).where(client_id: client_ids_in_disability(key)).distinct },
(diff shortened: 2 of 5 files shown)


drivers/core_demographics_report/app/models/core_demographics_report/report_concern.rb
```diff
@@ -78,6 +78,13 @@ def can_see_client_details?(user)
       user.can_access_some_version_of_clients? && ! filter.mask_small_populations
     end
 
+    def can_view_client_disability?(user, disability_id)
+      return user.can_view_hiv_status? if disability_id == 8 # 'HIV/AIDS'
+      return user.can_view_dmh_status? if disability_id == 9 # 'Mental health disorder'
+
+      true
+    end
```
eanders, line 86: This is nice!  I wonder if we should move it somewhere more centralized, but here is good for now.  Thanks!

drivers/core_demographics_report/app/models/core_demographics_report/disability_calculations.rb
```diff
@@ -13,6 +13,7 @@ def disability_detail_hash
         HudUtility2024.disability_types.each do |key, title|
           hashes["disability_#{key}"] = {
             title: "Disability #{title}",
+            cannot_view_details: ! can_view_client_disability?(@filter.user, key),
```
eanders: might be easier to read if this was:
```suggestion
            can_view_details:  can_view_client_disability?(@filter.user, key),
```
You'd have to invert the usage of it as well, but I find reasoning about the positive generally easier.

drivers/core_demographics_report/app/views/core_demographics_report/warehouse_reports/shared/_disabilities.haml
```diff
@@ -23,7 +23,7 @@
             - HudUtility2024.disability_types.each do |id, title|
               %tr
                 %th= title
-                %td= link_to_if @report.can_see_client_details?(current_user), number_with_delimiter(report.disability_count(id)), polymorphic_path(@report.detail_path_array, report.filter.for_params.deep_merge(key: "disability_#{id}", filters: { start: report.filter.start_date, end: report.filter.end_date })), target: :_blank
+                %td= link_to_if @report.can_see_client_details?(current_user) && @report.can_view_client_disability?(current_user, id), number_with_delimiter(report.disability_count(id)), polymorphic_path(@report.detail_path_array, report.filter.for_params.deep_merge(key: "disability_#{id}", filters: { start: report.filter.start_date, end: report.filter.end_date })), target: :_blank
```
eanders, line 26: Do we need to enforce this in the controller as well, or does enforcing it the details.rb sufficiently block it such that if you navigate to the page you won't get any data?

eanders (approved):
Thanks @dtgreiner, this looks great.  I left some yak shaving comments.  Probably worth doing, but none should be functional changes.

drivers/core_demographics_report/app/views/core_demographics_report/warehouse_reports/shared/_disabilities.haml
dtgreiner: This section is preventing the link from being active if the user doesn't have the correct permission. The cannot_view_details (soon to be can_view_details) on disability_detail_hash being set to false is making it so the details page will load but show no data if user doesnt have permission to view the disability data.

eanders: As long as the detail view doesn't include any data (which was my read, and you confirmed,) we're good.  Thanks!

drivers/core_demographics_report/app/models/core_demographics_report/disability_calculations.rb
dtgreiner: This is updated in the latest commit

drivers/core_demographics_report/app/models/core_demographics_report/details.rb
```diff
@@ -35,6 +35,7 @@ def detail_hash
     def detail_scope_from_key(key)
       detail = detail_hash[key]
       return report_scope.none unless detail
+      return report_scope.none if ! detail[:can_view_details].nil? && ! detail[:can_view_details]
```
eanders: This _should_ handle both cases a bit more cleanly.  
```suggestion
      return report_scope.none unless detail&.try(:[], :can_view_details)
```

If `detail` isn't set, it will return nil which will evaluate to false, if it's false, it'll be false, if it's true, it'll be true.

dtgreiner: The issue is the can_view_details is currently only set on the disability calculations so the other calculations' details won't populate because the value is nil. But this is a lot cleaner and maybe the better route is adding this and a can_view_details that defaults to true for all of the other details sections in this report

dtgreiner: Looking at it more, it may be easier to run this in the controller and redirect back to the report if the user tries to load details they don't have access too.

eanders (approved):
Thanks @dtgreiner

PR #4032, opened 2024-02-18
Import overrides

eanders:

## Description

This provides a more generic mechanism to override data on import.  This will eventually replace all PDDE overrides.

## Type of change


drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_overrides/new.haml
```diff
@@ -0,0 +1,18 @@
...
+
+%h1 Add Override for #{@data_source.name}
+= simple_form_for :override, url: hmis_csv_importer_data_source_import_overrides_path(@data_source) do |f|
+  = f.error_notification
+  .row
+    .col-md-6
+      .well
+        .form-inputs
+          = f.input :file_name, collection: HmisCsvImporter::ImportOverride.available_classes, label: 'File', as: :select_two, include_blank: false
+          = f.input :matched_hud_key, label: 'Matched HUD Key', hint: 'Given the chosen table, should we only replace data in rows with this matching key? If blank, all rows will be replaced.  For example if you would like all enrollments to have the same EnrollmentCoC, leave this blank.  If you would like to change the type of a single project, enter the value of the ProjectID for that project here.', required: false
```
gigxz: I haven't seen us refer to the HUD IDs as "Keys" before. In the HMIS we always refer to them as IDs. I feel like it should be consistent with terminology in the warehouse, which I think also uses "ID" language?

eanders: `hud_key` is how we've historically referenced the ID field in the HMIS CSV tables.  I guess I'd call the data `hud_id` maybe, but the name of the key has been `hud_key`.  Maybe you just mean in the UI though, yeah, maybe we should update that to `Matched HMIS ID`

gigxz: Yeah I meant primarily user-facing language. I think the change you made is good.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/import_override.rb
```diff
@@ -0,0 +1,86 @@
...
+
+    row
+  end
+
+  def applies?(row)
+    row = normalize_row(row)
+    # Double check we actually have the column we're looking for (protects against typos or future spec changes)
+    return false unless row.key?(replaces_column)
+    # We were expecting a specific HUD key, and this is not it
+    return false if matched_hud_key.present? && row[hud_key] != matched_hud_key
```
ttoomey: safe navigation maybe more concise?
```suggestion
    return false unless matched_hud_key.presence&.== row[hud_key]
```

db/warehouse/migrate/20240217192729_create_import_overrides.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :import_overrides do |t|
+      t.string :file_name, null: false
+      t.string :matched_hud_key
+      t.string :replaces_column, null: false
+      t.string :replaces_value
+      t.string :replacement_value, null: false
+      t.references :data_source, null: false, foreign_key: true
+
+      t.timestamps
+      t.datetime :deleted_at
```
ttoomey, line 12: should this include uniqueness constraint? 
```ruby
    reversible do |dir|
      dir.up do
        execute <<-SQL
          CREATE UNIQUE INDEX uidx_import_overrides_rules
          ON import_overrides (data_source_id, file_name, replaces_column, COALESCE(matched_hud_key, 'ALL'), COALESCE(replaces_value, 'ALL'))
          WHERE deleted_at IS NOT NULL;
        SQL
      end
    end
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -197,6 +199,10 @@ def pre_process_class!(file_name, klass)
       end
     end
 
+    private def import_overrides_for(file_name)
+      HmisCsvImporter::ImportOverride.where(file_name: file_name)
```
ttoomey: I might be missing something but it seems like these wouldn't be sorted so there could be non deterministic results if multiple overrides apply to the same record/col. I think there should be a sort order with id tie-breaker to avoid that:
```suggestion
      HmisCsvImporter::ImportOverride.where(file_name: file_name).
        to_a.sort_by { |r| [r.specificity, r.id]}
```
```ruby
  class ImportOverride
    def specificity
      if  matched_hud_key && replaces_value
        3
      elsif replaces_value
        2
      elsif matched_hud_key
        1
      else
        0
      end
    end
  end
  
```

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_overrides/new.haml
```diff
@@ -0,0 +1,18 @@
...
+          = f.input :matched_hud_key, label: 'Matched HMIS ID', hint: 'Given the chosen table, should we only replace data in rows with this matching ID? If blank, all rows will be replaced.  For example if you would like all enrollments to have the same EnrollmentCoC, leave this blank.  If you would like to change the type of a single project, enter the value of the ProjectID for that project here.', required: false
+          = f.input :replaces_column, label: 'Column to replace', hint: 'EnrollmentCoC or ProjectID in the above examples.'
+          = f.input :replaces_value, label: 'Value to replace', hint: 'Leave this blank to replace the column on all matching rows.  If you only want to replace one value, for example to change all 99s in ContinuumProject to 0, since 99 is not allowed, enter 99 here.', required: false
+          = f.input :replacement_value, label: 'Replacement value', hint: 'Enter the value you would like instead of the exsiting value here.  In the above example for ContinuumProject, enter 0.'
+        .form-actions
+          = f.button :submit, class: 'btn btn-primary', value: 'Add Override'
```
ttoomey, line 18: In other projects, I try and add `data-disable-with="Loading…"` from UJS on form buttons to avoid duplicate submissions from double clicks. Looks like we don't use that pattern in the warehouse though?

ttoomey (approved):
Looks good! I left a few suggestions

db/warehouse/migrate/20240217192729_create_import_overrides.rb
eanders: Seems reasonable.  There's definitely no advantage to having duplicates.  Thanks.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/import_override.rb
eanders: Oh, that's not a syntax I'm familiar with.  Not sure what I think about `&.== `.  We'll give it a try!  Though `&.!=` since that's the comparison we're actually making.  Thanks!

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: Yes, thank you!  That will be somewhat critical if we ever have less specific uses.

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_overrides/new.haml
eanders: We've removed that in the warehouse.  It was a long time ago, but if memory serves there were times where the button would be disabled and never re-enable.

PR #4034, opened 2024-02-19
Remove PDDE overrides in favor of import overrides

eanders:

## Description

This is massive cleanup of previous override implementation

## Type of change


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb
```diff
@@ -297,6 +297,7 @@ def cleanup_dangling_funders
       dangling_funders.each(&:destroy!)
     end
 
+    # FIXME: Are these still needed?  If they are, can we move to the new mechanism?
```
eanders: @gigxz any thoughts on this?

drivers/hmis_external_apis/spec/models/hmis_external_apis/ac_hmis/importers/projects_importer_spec.rb
```diff
@@ -42,6 +42,7 @@
     expect(Hmis::Unit.where(unit_type: active_unit_type).count).to eq(10)
   end
 
+  # FIXME: these should be removed, or replaced with import_overrides
```
eanders: @gigxz any thoughts on this?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/importers/projects_importer.rb
eanders: @gigxz confirmed, no longer necessary

drivers/hmis_external_apis/spec/models/hmis_external_apis/ac_hmis/importers/projects_importer_spec.rb
eanders: @gigxz confirmed, no longer necessary

eanders (commented):
Notes inline

gigxz (approved):

app/models/grda_warehouse/hud/project.rb
```diff
@@ -634,26 +617,18 @@ def self.related_item_keys
...
-    def compute_project_type
-      act_as_project_type.presence || self.ProjectType
-    end
-
     # Originally wasn't PH, but is overridden to PH
     def project_type_overridden_as_ph?
       @psh_types ||= HudUtility2024.residential_project_type_numbers_by_code[:ph]
       ! @psh_types.include?(self.ProjectType) &&
-        @psh_types.include?(compute_project_type)
+        @psh_types.include?(loaded_items_2024.last&.ProjectType)
```
awisema: I don't understand this line?

awisema (approved):

app/models/grda_warehouse/hud/project.rb
eanders: So, with the move to overriding the imported data the difference in value moved from between `ProjectType` and `computed_project_type` to `ProjectType` on the Imported (or post-processed) tables and the equivalent `ProjectType` in the loaded table.  So this now checks the last loaded data set for this record had the same ProjectType.
  
That said, this could definitely use some further thought.  
* The loaded_items for the record should probably be version agnostic
* Instances where the data comes from our own HMIS, there may not be a loaded item, in which case we should short-circuit since there are no overrides there
* loaded_items might be a string instead of an integer, need to double check. 

Thanks for flagging this!
At the moment this is only used during import to determine if a move-in date should be assigned, so maybe the logic should be completely moved to `def self.assign_move_in_date(row)` and pulled out of project.

db/warehouse/migrate/20240218201801_move_override_data.rb
```diff
@@ -0,0 +1,50 @@
...
+      act_as_project_type: :ProjectType,
+      # hud_continuum_funded: :ContinuumProject,
+      housing_type_override: :HousingType,
+      operating_start_date_override: :OperatingStartDate,
+      operating_end_date_override: :OperatingEndDate,
+      hmis_participating_project_override: :HMISParticipatingProject,
+      target_population_override: :TargetPopulation,
+    }.each do |override, source|
+      GrdaWarehouse::Hud::Project.where.not(override => nil).
+        update_all(source => p_t[override])
```
ttoomey, line 14: should this include soft-deleted records too? (with_deleted)

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
```diff
@@ -0,0 +1,78 @@
...
+      target_population_override: :TargetPopulation,
+    }.each do |override, source|
+      GrdaWarehouse::Hud::Project.where.not(override => nil).each do |project|
+        HmisCsvImporter::ImportOverride.where(
+          file_name: 'Project.csv',
+          matched_hud_key: project.ProjectID,
+          replaces_column: source,
+          replacement_value: project[override],
+          data_source_id: project.data_source_id,
+        ).first_or_create
```
ttoomey, line 20: how many overrides will need to be created? Would it be worth it to collect them and bulk insert to reduce the runtime for this migration?

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
```diff
@@ -0,0 +1,78 @@
...
+      end
+    end
+    # Special case Continuum Project
+    GrdaWarehouse::Hud::Project.where(hud_continuum_funded: true).each do |project|
+      HmisCsvImporter::ImportOverride.where(
+        file_name: 'Project.csv',
+        matched_hud_key: project.ProjectID,
+        replaces_column: ContinuumProject,
```
ttoomey: ContinuumProject should be symbol?

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
```diff
@@ -0,0 +1,78 @@
...
+        replacement_value: 1,
+        data_source_id: project.data_source_id,
+      ).first_or_create
+    end
+    GrdaWarehouse::Hud::Project.where(hud_continuum_funded: false).each do |project|
+      HmisCsvImporter::ImportOverride.where(
+        file_name: 'Project.csv',
+        matched_hud_key: project.ProjectID,
+        replaces_column: ContinuumProject,
```
ttoomey: ContinuumProject should be symbol?

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
```diff
@@ -0,0 +1,78 @@
...
+      GrdaWarehouse::Hud::ProjectCoc.where.not(override => nil).each do |project_coc|
+        HmisCsvImporter::ImportOverride.where(
+          file_name: 'ProjectCoC.csv',
+          matched_hud_key: project_coc.ProjectCoCID,
+          replaces_column: source,
+          replacement_value: project_coc[override],
+          data_source_id: project_coc.data_source_id,
+        ).first_or_create
+      end
+    end
```
ttoomey, line 76: should there be another migration to rename these fields out of the way? Maybe todo_or_die to drop the cols in the future?
```
rename_column :Project, :act_as_project_type, :old___act_as_project_type
# etc...
```

app/models/grda_warehouse/hud/project.rb
```diff
@@ -866,19 +841,17 @@ def project_type_to_use
...
     def self.project_type_column
-      if GrdaWarehouse::Config.get(:project_type_override)
-        :computed_project_type
-      else
-        :ProjectType
-      end
+      :ProjectType
     end
 
+    # TODO: remove this and just use operating start date
```
ttoomey, line 841: todo_or_die maybe?

ttoomey (commented):
I don't have the context or historical knowledge to understand these change but it seems like a great simplification! Left a few comments around the migrations

db/warehouse/migrate/20240218201801_move_override_data.rb
eanders: Good question, but no.  If the Project was deleted we don't care about the overrides.  If it ever comes back we can re-add.

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
eanders: We're talking a couple dozen at most, so this should be fine un-batched.

db/warehouse/migrate/20240218222444_move_existing_overrides.rb
eanders: Yes, thank you!

PR #4037, opened 2024-02-20
186704906 public pages

ttoomey:

## Description

https://www.pivotaltracker.com/n/projects/2591838
https://www.pivotaltracker.com/story/show/187040557

[Front-end PR](https://github.com/greenriver/hmis-frontend/pull/679)

* Low dependency stand alone static form page generator for public forms. Forms should work on old / underpowered devices
* Forms are driven off existing form definitions using a subset of the form definition language
* Uses CDED/CDE's to allow us to re-use existing code to to display/update external form submissions.
* publishes static forms to public bucket
* api to expose and update submissions for review (project.external_form_submissions)
* adds relationship between form_definition and custom_data_element_definition
* hourly task to consume the submissions from s3

## Testing
```
rails db:migrate:up:warehouse
# creates the appropriate definition and runs the publish job
bundle exec rake driver:hmis_external_apis:tc_hmis:seed_external_forms
```

The test harness controller should work in development/test mode to exercise the forms
```
https://hmis-warehouse.dev.test/hmis_external_api/external_forms/tchc/prevention_screening
````

Jobs:
If s3 is configured, running PublishExternalFormsJob should upload the rendered form html to the path (s3 object key) defined in form_definition.external_form_object_key (in development/test this is skipped) in the public bucket. Note this s3 upload has not yet been tested.

The ConsumeExternalFormSubmissionsJob should pull all submissions from the submissions bucket and attempt to save them to the local db as external form (HmisExternalApis::ExternalForms::FormSubmission). If saved successfully, the submission object is deleted from the s3 bucket.

## Deployment
* We should set EXTERNAL_FORM_CAPTCHA_KEY in the env to the public "site" recaptcha key. This needs to match the recaptcha key in the AWS lambda env.
* need to create RemoteCredential to access the appropriate external forms bucket. Production & staging should both have access to this bucket already
* for CDEDS in the external form that should show up in the submissions table in the HMIS, set show_in_summary=true
* 
will need to create a new AES secret key as a remote crdential:
```ruby
    GrdaWarehouse::RemoteCredentials::SymmetricEncryptionKey.where(slug: 'external_forms_shared_key').first_or_create! do |record|
      record.algorithm = 'aes-256-cbc'
      record.key_hex = 'shared secret with the aws lamdba'
    end
```

## Outstanding issues:
* currently the lambda is not in version control. We should find a home for it
* I have a google recaptcha token for this form but that should probably be managed by a central account
* date fields is not currently validated. [See PT issue](https://www.pivotaltracker.com/story/show/187250024)
* assuming IE 11 is a concern here, we should make a pass at compatibility. We are mostly there already but may need to find a polyfill for the Blob API
* we may want to upload the form logo manually to the bucket
* The CSP is currently disabled. We should probably adjust it and re-enable it

## Type of change


drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -0,0 +1,33 @@
...
+    include Types::HmisSchema::HasCustomDataElements
+
+    available_filter_options do
+      arg :status, [String]
+    end
+
+    field :id, ID, null: false
+    field :submitted_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :spam_score, Float, null: true
+    field :status, String, null: false # probably should be an enum to new | resolved
```
gigxz: 👍 to making this an enum

drivers/hmis_external_apis/app/helpers/hmis_external_apis/external_forms_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+      presign_url ||= presign_hmis_external_apis_external_form_path
+      # for development on hmis-warehouse.dev.test
+      recaptcha_key ||= '6Ldbm4UpAAAAAAK9h9ujlVDig1nC4DghpOP6WFfQ'
+    end
+
+    # FIXME move away from this hard-coded configuration
+    @page_config ||= HmisExternalApis::ExternalForms::Config.new(
+      site_title: 'Tarrant County Homeless Coalition',
+      # FIXME we should host this logo
+      site_logo_url: 'https://ahomewithhope.org/wp-content/themes/tchc/assets/images/logo.png',
```
gigxz: We have this logo, it is set in the env with `LOGO`

The code we use to get the path in hmis is:
```
    logo = ENV['LOGO']
    if logo.present?
      logo_path = SerializedAsset.exists?(logo) ? SerializedAsset.get_src(logo) : "theme/logo/#{logo}"
    end
```

I think the second case is for local development

gigxz: I'd say just use that for now (i.e. assume that the `LOGO` should go on public forms). If we need another var later we can add

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
```diff
@@ -0,0 +1,112 @@
...
+  def update_cdeds(definition)
+    owner_type = HmisExternalApis::ExternalForms::FormSubmission.sti_name
+    cded_by_key = definition.custom_data_element_definitions.index_by(&:key)
+    missing = []
+    definition.walk_definition_nodes do |node|
+      cded_key = node['link_id']
+      attrs = nil
+      case node['type']
+      when 'STRING', 'CHOICE', 'BOOLEAN'
+        # treat everything as a string for now
```
gigxz: seems ok for now

gigxz: but, we might need to migrate the values if we need to change this later to allow for editing the contents from hmis.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
```diff
@@ -0,0 +1,112 @@
...
+  # construct custom data element definitions from the nodes in the form definition
+  def update_cdeds(definition)
+    owner_type = HmisExternalApis::ExternalForms::FormSubmission.sti_name
+    cded_by_key = definition.custom_data_element_definitions.index_by(&:key)
+    missing = []
+    definition.walk_definition_nodes do |node|
+      cded_key = node['link_id']
+      attrs = nil
+      case node['type']
+      when 'STRING', 'CHOICE', 'BOOLEAN'
```
gigxz: I see we collect DOB too, won't that be a DATE type?

gigxz:
~deleted comment~ nvm seems like I can get it working despite lacking s3 setup

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/tchc/_helpline_shelters.haml
```diff
@@ -0,0 +1,16 @@
+= render_dependent_block(input_name: 'household_type', input_value: 'single_adult') do
+  %ul.mb-0.pl-3
```
gigxz: I don't think its a good idea to keep these around in the codebase. If only for the sake of like exposing these phone numbers to bots crawling github. Maybe you can upload them onto the helpline PT story instead so we can reference them for future helpline work?

gigxz (commented):
overall this approach looks good to me @ttoomey . thanks for all this work 🙀  I think/hope this will make implementations of the future public forms go smoothly!

I left some comments inline. I didn't test locally due to S3 issues, but I can try to get that sorted out this week. I think @martha might need it too in order to get the full end-to-end working locally. (We both had minio issue last time, I know there are some other approaches that are not documented).

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/tchc/_helpline_shelters.haml
ttoomey: Sure thing! All this info was taken from the public facing forms so I it's not disclosing anything. It's cleaner to remove it and archive in PT though

drivers/hmis_external_apis/app/helpers/hmis_external_apis/external_forms_helper.rb
ttoomey: Thanks! This is much better but it looks like it will break if the warehouse isn't publicly available. I was thinking it should probably be uploaded to the public bucket along with the HTML.

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -234,6 +234,15 @@ def data_collection_features
     end.compact
   end
 
+  def external_form_submissions
```
martha: I'm not sure if my understanding is all correct here, but I think perhaps using `detect_best_instance_for_project` here isn't right, because it returns only one instance's worth of form submissions, but we want this to return all the form submissions for the project -- and so we can then filter them from the graphql query depending on which form type the user selects in the dropdown @gigxz  mentioned [here](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL?node-id=1879:12521&mode=design#724436105).

this is a snippet of code that I got working locally, though as usual, I'm not sure it's airtight from an arel perspective:
```
  def external_form_submissions
    definition_identifiers = Hmis::Form::Instance.
      for_project(self).
      with_role(:EXTERNAL_FORM).
      active.order(updated_at: :desc).
      pluck(:definition_identifier).
      uniq

    return Hmis::HmisExternalApis::ExternalForms::FormSubmission.none if definition_identifiers.empty?

    HmisExternalApis::ExternalForms::FormSubmission.
      joins(:definition).
      where(definition: { identifier: definition_identifiers })
  end
```

gigxz: That looks correct to me. I think based on the design, we are able to _require_ the definition when querying for the Form Submissions, which would make this a bit simpler. Like:

```
  def external_form_submissions(definition_identifier:)
    # do some validation to make sure the definition identifier is valid for the project
    raise "not valid for the project" unless Hmis::Form::Instance.for_project(self).with_role(:EXTERNAL_FORM).with_identifier(definition_identifier).exists?

    HmisExternalApis::ExternalForms::FormSubmission.joins(:definition).where(fd_t[:identifier].eq(definition_identifier))
  end
```

If we go that route, the frontend would already have the `definition_identifier` because they selected it from the dropdown. Assuming that the `PickListOption` returns something like `{ label: "Prevention Screening Form", code: <definition identifier> }`.

ttoomey: Thanks for the improvement and simplification, I'll make that update

ttoomey: Updated the resolver and added a required formDefinitionIdentifier argument

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -0,0 +1,32 @@
...
+    field :id, ID, null: false
+    field :submitted_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :spam_score, Float, null: true
+    field :status, HmisSchema::Enums::ExternalFormSubmissionStatus, null: false
+    field :notes, String, null: true
+    field :definition, Forms::FormDefinition, null: false
+    custom_data_elements_field
+
+    def custom_data_elements
+      resolve_custom_data_elements(object, definition_scope: definition.custom_data_element_definitions)
```
martha: Based on playing around with the UI changes, I think we should order these somehow so they come back in a consistent and sensible order (e.g. first name then last name). `.order(id: :asc)` maybe?

ttoomey: Sounds good, it should sort by key now. Also adjusted the resolver to use a data loader.

Probably relying on id/key order isn't really right and the cdeds should have an optional sort order. But we can tackle that down the road

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -0,0 +1,35 @@
...
+module Mutations
+  class UpdateExternalFormSubmission < CleanBaseMutation
+    argument :input, Types::HmisSchema::ExternalFormSubmissionInput, required: true
+    argument :id, ID, required: true
+
+    field :external_form_submission, Types::HmisSchema::ExternalFormSubmission, null: false
+    field :errors, [Types::HmisSchema::ValidationError], null: false, resolver: Resolvers::ValidationErrors
+
+    def resolve(id:, input:)
+      record = Hmis::HmisSchema::ExternalFormSubmission.find(id)
```
martha: I think we need to resolve the model here and not the hmis schema object,
```suggestion
      record = HmisExternalApis::ExternalForms::FormSubmission.find(id)
```

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -0,0 +1,33 @@
...
+    include Types::HmisSchema::HasCustomDataElements
+
+    available_filter_options do
+      arg :status, HmisSchema::Enums::ExternalFormSubmissionStatus
+      arg :submitted_date, GraphQL::Types::ISO8601Date
+    end
+
+    field :id, ID, null: false
+    field :submitted_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :spam_score, Float, null: true
```
martha: I noticed that sometimes spam is a score (float), and other times it's a boolean, but I think (unless I've missed it) that we still need to add the translation between the two. This raises a set of ui-related questions,
- do we want to expose the float spam score to the user in the UI -- I assume yes?
- do we want that spam score float field to be settable by the user, or should they separately be able to check a box yes/no for whether they think it's spam, which I guess would then set spam score to 0 or 100?
- relatedly, but on a broader topic, the [mock](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-Phase-2-(2024)?type=design&node-id=1883-14891&mode=design&t=s8now9TLznGWadbU-0) shows a trash button that would enable to delete a submission entirely. What is the difference in use case between marking as spam vs. deleting entirely? I understand these will be different functionally but I'm wondering how we hope users will use them. Maybe in the future we want to build a 'delete all spam' button or similar?

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
ttoomey: Yes! I didn't get around to actually exercising this mutation. Thanks for the fix!

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
ttoomey: @martha 
The recaptcha spam score is a float between 0 (likely spam) and 1.0 (likely real). The threshold to automatically flag a submission as spam is up to us but we'll probably start with 0.5. Something to note is that it's quite possible for a real submission to get marked as spam.

I would lean towards not exposing the numeric spam score in the API. I imagine users can flag the message as spam with a simple toggle backed by a boolean field (maybe "mark as spam" under the "mark as resolved checkbox"). The resolver and mutation would then map from the boolean true/false to 0/1.0 as you describe.

> What is the difference in use case between marking as spam vs. deleting entirely? I understand these will be different functionally but I'm wondering how we hope users will use them.

Here are some possible use-cases for the delete button:
* review a message that was marked as spam and delete it
* delete a duplicate submission
* delete a test submission

> Maybe in the future we want to build a 'delete all spam' button or similar?

Yes, in the future we might want a 'delete all spam' button particularly if the table mixes spam and non-spam submissions together.

Another solution would be to hide spam messages by default and delete them automatically after 30 days (or some time period). Users would be able to review the spam submissions and mark them as 'not-spam'. This mirrors how many email clients handle spam so it should be familiar to the user.

martha: Ok. All sounds good! I opened a PR [here](https://github.com/greenriver/hmis-warehouse/pull/4099) against this branch that incorporates some of this.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
martha: We also collect several INTEGER types and a TEXT

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
martha: I'm not sure, but I think this shortcut also causes the need for some special-casing from the frontend, specifically with choice (pick list) and boolean components, in the existing `DynamicView` framework.

Pick list special-casing: https://github.com/greenriver/hmis-frontend/pull/691/files#diff-905e9c45cedd2c5cbc677a3ad50f299481520a061440b9c9d134cf5ae338a656R38-R40
and see also my comment here about what's still not working: https://github.com/greenriver/hmis-frontend/pull/691#issue-2176768038

drivers/hmis_external_apis/app/helpers/hmis_external_apis/external_forms_helper.rb
```diff
@@ -0,0 +1,110 @@
...
+      radios = render(partial_path('form/radio_group_options'), options: options, name: name, required: required, html_id: next_html_id)
+      extra = capture(&block) if block
+      safe_join([radios, extra].compact_blank, "\n")
+    end
+  end
+
+  def render_section(title:, &block)
+    tag.section do
+      safe_join(
+        [tag.h2(title, class: 'h4'), capture(&block)],
```
eanders, line 46: Stupid question, but why are we using an H2 tag with an H4 class?

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
```diff
@@ -0,0 +1,112 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+
+require 'nokogiri'
```
eanders, line 6: I could be wrong, but the gem file includes nokogiri always, do we need to require it here as well, or is this just an indication that we're going to be using it?

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
```diff
@@ -0,0 +1,167 @@
...
+
+    def parent_node
+      @stack[-2]
+    end
+
+    def render_input_node(node)
+      render_form_group(node: node) do
+        case node['component']
+        when 'PHONE'
+          render_form_input(label: node['text'], name: node['link_id'], required: node['required'], input_type: 'tel')
```
eanders: I'm sure I'm missing something, and I don't think it's worth changing at this point, but it sure feels like we should have been able to lean a bit more on the rails form generation logic.  It's hard to see us re-writing something that has so many edge cases that have been handled by other tools.

drivers/hmis_external_apis/app/helpers/hmis_external_apis/external_forms_helper.rb
ttoomey: Just for aesthetics. Semantically it should be h2 but h4 looks more appropriate on the page (imo)

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
ttoomey: Good point, it's probably safe to remove. I think it's harmless though.

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
ttoomey: This method could definitely be DRY-ed up, thanks for surfacing that. I'll try and give it another pass

> It's hard to see us re-writing something that has so many edge cases that have been handled by other tools.

Sorry I don't quite follow what you mean by edge cases here. My understanding is that regardless of how the inputs are rendered, you'd still need the logic in this generator to parse our custom form definition and invoke the appropriate renderer.

I take your point that generally we want to use rails builders when possible and perhaps you could replace some of the code in `ExternalFormsHelper` with the low-level rails helpers (text_field_tag etc). However simple_form and the higher-level rails form builders assume that you have a model that persists the data and that rails itself will be validating and processing the form submission. That's not true here so the benefit you'd get seems limited. 

I think that this implementation is reasonable given our requirements to host this in aws/lamba and generate the form from our definition config. Happy to discuss further though

eanders: Yeah, sorry, I should have dropped this as an overall comment, not specific to that line.  I don't think we should refactor at this time, but it started to feel like `form_for` or some of the slightly lower level `input_field` and similar methods might allow us to render fields without writing custom methods and partials.  That said, this looks great, and will probably serve us for years to come.  

The edge cases I was eluding to are things like having to handle selects with multiple options turned on, or specifics around input types for integers, dates, etc.  Forms are less trouble than date handling, but a completely generic solution is probably not worth trying to implement since there are so many other form tools out there.

I agree, this implementation is reasonable and totally fine for now.
Thanks!

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -473,53 +496,47 @@ def introspect_custom_data_element_definitions(set_definition_identifier: false)
     cdeds_by_key = cded_scope.index_by(&:key)
 
     cded_records = []
-    recur_traverse = lambda do |items|
```
ttoomey, line 476: This change isn't strictly needed but seemed better to make use of walk_definition_nodes rather than a lambda.

db/warehouse/migrate/20240223002628_add_external_forms.rb
```diff
@@ -0,0 +1,30 @@
+class AddExternalForms < ActiveRecord::Migration[6.1]
+  def change
+    add_column :hmis_form_definitions, :external_form_object_key, :string
+    add_column 'CustomDataElementDefinitions', :form_definition_identifier, :string
```
ttoomey, line 4: per team discussion, we decided to keep this relationship

eanders (approved):
Looks good to me.  Thanks @ttoomey !

gigxz:
Lets retarget to 109

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -394,6 +396,15 @@ def self.open_hoh_enrollments_for_project(project)
...
     end
 
+    def self.external_form_types_for_project(project)
+      return [] unless project.present?
+
+      Hmis::Form::Instance.for_project(project).
+        preload(:definition).
+        order(:id).
+        map(&:to_pick_list_option)
+    end
```
gigxz: Whoops, missed this the first time around. I'm seeing 2 issues:
1) this is returning non-external-forms, if there are other forms that have rules for this project (fixed by `with_role`)
2) if for some reason you had two duplicate rules for the project, it would return duplicates(fixed by `uniq`)

```suggestion
      Hmis::Form::Instance.for_project(project).
        with_role(:EXTERNAL_FORM).
        preload(:definition).
        order(:id).
        map(&:to_pick_list_option).uniq
    end
```

martha: thank you for the good catch @gigxz !

gigxz (approved):
Looks awesome. 1 suggested change for the pick list resolution, lgtm after that.

Created a couple follow-up tickets
* Enable CSP https://www.pivotaltracker.com/story/show/187295145
* IE 11 pass https://www.pivotaltracker.com/story/show/187295135

ttoomey:
@martha @gigxz 
 Is this PR and [the FE PR](https://github.com/greenriver/hmis-frontend/pull/679) ready to merge into release-109? If so I can take care of that

Thanks!

gigxz:
Yep go for it @ttoomey! Thanks!

PR #4039, opened 2024-02-20
set cohort download as a background job

dtgreiner:

## Description

Update the cohort download to be performed as a background job

## Type of change

### DIFF
--- a/app/controllers/cohorts_controller.rb
+++ b/app/controllers/cohorts_controller.rb
@@ -64,6 +64,7 @@ class CohortsController < ApplicationController
     @population = params[:population]
     respond_to do |format|
       format.html do
+        @excel_export = GrdaWarehouse::Cohorts::DocumentExports::CohortExcelExport.new
         @visible_columns = [CohortColumns::Meta.new]
         @visible_columns += @cohort.visible_columns(user: current_user)
         delete_column = if params[:population] == 'deleted'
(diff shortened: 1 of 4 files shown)


eanders (changes requested):
@dtgreiner this looks great overall.  One small change though, we need to adjust the "query_string" saved in the `document_exports` table such that if you were to change the visible columns it should re-generate the download the next time you request it.  

Because of the long expiration in `DocumentExportBehavior` you'll get the same file for 30 days regardless of cohort configuration or who is on the cohort.

Bonus points for adding a client_count=xxx to the `query_string` (which is really a cache key) so that if someone is added or removed, the download would change.

dtgreiner:
@eanders I just pushed an update to include the visible columns and client count in the export query string

app/controllers/cohorts_controller.rb
```diff
@@ -110,6 +111,11 @@ def show
           end
           options
         end
+
+        # included so the excel download can be regenerated when the visible columns change
+        params['visible_columns'] ||= @visible_columns.map { |m| m.column.to_s }.join(',')
+        # included so the excel download can be regenerated if the total number of clients changes
+        params['client_count'] ||= @cohort.search_clients(population: params[:population], user: current_user).count
```
eanders: Maybe a simple MD5 to keep these small, what do you think?  The idea is simply to force a rebuild if anything has changed.
I'll test:
```suggestion
        # included so the excel download can be regenerated if the total number of clients changes
        params['cache_key'] ||= Digest::MD5.hexdigest(@visible_columns.map { |m| m.column.to_s }.join(',') << @cohort.search_clients(population: params[:population], user: current_user).count)
```

dtgreiner: This makes sense to me and looks much nicer in the database. It's also working in my tests.

PR #4044, opened 2024-02-21
Add Auto-Entry functionality for non-HUD projects

martha:

## Description

This PR sets up single-table inheritance for ProjectConfigs, from which AutoEnter and AutoExit configs will inherit, as described in https://www.pivotaltracker.com/n/projects/2591838/stories/186800449.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


db/structure.sql
```diff
@@ -317,6 +317,7 @@ CREATE TABLE public.activity_logs (
 --
 
 CREATE SEQUENCE public.activity_logs_id_seq
+    AS integer
```
martha: I'd like to ask about these changes to `structure.sql` and `warehouse_structure.sql` that come up when I run migrate but don't seem related to my changes -- are these ok to include in the PR?

drivers/hmis/app/graphql/mutations/submit_form.rb
```diff
@@ -96,7 +96,26 @@ def _resolve(input:, record_lock_version: nil)
...
+              assessment_date: record.entry_date,
+              user: Hmis::Hud::User.from_user(current_user),
+              wip: false,
+              data_collection_stage: 1,
+              client: record.client,
+              enrollment: record,
+            )
+
+            # Save the FormProcessor tied to the form definition id that would be used for intakes in this project
+            form_processor.custom_assessment = assessment
```
martha: I added this because of the instruction from the PT issue:
> and (maybe) a Form Processor tied to the form definition id that would be used for intakes in the project. 

However, I'll admit that I don't really understand why exactly this FormProcessor is needed or what it's doing?

drivers/hmis/app/models/hmis/project_auto_enter_config.rb
```diff
@@ -0,0 +1,8 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::ProjectAutoEnterConfig < Hmis::ProjectConfig
```
martha, line 7: In the PT issue it states that changes to the auto-exit config UI are out-of-scope for this ticket, but possibly in-scope for the follow-up ticket. Is that right and does it mean that no UI changes are required for this ticket at all? (there's currently no way to create an Auto Enter config from the UI)

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
```diff
@@ -555,6 +555,36 @@ def submit_enrollment_form(input)
...
+    end
+
+    it 'should save new enrollment without WIP status' do
+      response, result = post_graphql(input: { input: test_input }) { mutation }
+      errors = result.dig('data', 'submitForm', 'errors')
+      expect(response.status).to eq(200), result.inspect
+      expect(errors).to be_empty
+      record = result.dig('data', 'submitForm', 'record')
+      expect(record['inProgress']).to eq(false)
+      e_id = Hmis::Hud::Enrollment.find(record['id']).enrollment_id
```
martha: I noticed while debugging these tests that there are 2 different concepts of an enrollment_id -- the one that's returned on the record isn't the same one as `.enrollment_id` on the `Hmis::Hud::Enrollment` instance, and that one is the one needed to find the created `CustomAssessment`. Does what I've done here look correct? and what should I be understanding better about these different enrollment IDs?

drivers/hmis/app/graphql/mutations/submit_form.rb
```diff
@@ -96,7 +96,26 @@ def _resolve(input:, record_lock_version: nil)
           HmisExternalApis::AcHmis::CreateReferralRequestJob.perform_now(record)
         when Hmis::Hud::Enrollment
           record.client.save! if record.client.changed? # Enrollment form may create or update client
-          if record.new_record? || record.in_progress?
+
+          config = Hmis::ProjectAutoEnterConfig.config_for_project(record.project)
+          if config
```
gigxz: We only want to run this for _new_ enrollments.

```suggestion
          # If this is a brand new Enrollment record, and the project has auto-enter configured, save the enrollment as non-WIP and generate an empty intake assessment
          if record.new_record? && Hmis::ProjectAutoEnterConfig.config_for_project(record.project)
```


This mutation is tricky because it's used for lots of different form submissions. If you go to the Enrollment Dashboard and find "Move In Date" or "Date of Engagement" on the overview card, hit edit and it will bring up a modal. That mini-form does a SubmitForm for the Enrollment, in order to change those values. That would be a good thing to QA–– on both WIP and non-WIP enrollments–– to make sure this doesn't break that behavior.

drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def self.configs_for_project(project)
+    pc_t = Hmis::ProjectConfig.arel_table
+
+    where(
+      pc_t[:project_id].eq(project.id).
+        or(pc_t[:organization_id].eq(project.organization.id)).
+        or(pc_t[:project_type].eq(project.project_type)),
+    )
+  end
```
gigxz: For consistency we usually prefer to use `scope` over a class method for this kind of thing

```suggestion
  scope :for_project, ->(project) do
    pc_t = Hmis::ProjectConfig.arel_table

    where(
      pc_t[:project_id].eq(project.id).
        or(pc_t[:organization_id].eq(project.organization.id)).
        or(pc_t[:project_type].eq(project.project_type)),
    )
  end
```


I think we need another scope for `active` since the rules can be enabled or not.

drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -0,0 +1,55 @@
...
+
+  def self.config_for_project(project)
+    configs = configs_for_project(project)
```
gigxz: The idea of configs being "enabled" or not is new, so we should add it here. Maybe:

```suggestion
    configs = configs_for_project(project).active
```

Where `active` is a scope that filters down to enabled configs only

drivers/hmis/app/models/hmis/project_auto_enter_config.rb
gigxz: yep no UI changes!

drivers/hmis/app/graphql/mutations/submit_form.rb
gigxz: Ah yep good question.

Each `CustomAssessment` has exactly one `FormProcessor` that holds the raw data that was submitted, & the references to the records that were updated using the assessment. There are a few notes [here](https://docs.google.com/document/d/1gV_naB47tUd0xB57EZFyWPjybbOMj2TKm2T1UDC2FU4/edit#heading=h.din53rzf7gz) about those tables. It can also optionally reference the FormDefinition that was used to submit the form. In this case, we didn't use a FormDefinition to submit the form, since it's really just a stub. So we can leave it empty.  [When the assessment is resolved,](https://github.com/greenriver/hmis-warehouse/blob/ca8a5ac066f671acfec417a22102eadbbff4bd7b/drivers/hmis/app/graphql/types/hmis_schema/assessment.rb#L59-L75) (e.g. if a user re-opens this stub assessment to add some information to it), then the "most appropriate" form definition will be chosen for rendering it. That's the  `Hmis::Form::Definition.find_definition_for_role` bit in the linked code (role = INTAKE, in this case).

What you have here looks mostly correct, except that the `form_processor` that you're referencing is actually not the form_processor for the intake assessment. It's the processor that was used to "process" the submitted values in this `SubmitForm` operation -- thats the json blob of information about this _enrollment_. That `form_processor` does not need to be persisted. We only persist processors for Assessments, since those are the only forms that need to update lots of other record types.


I hope that made some sense! It's happy to talk through it.


So - my suggestion would be move this logic onto the Enrollment, since it might be useful in other circumstances.
```
# on enrollment model:

  def build_synthetic_intake_assessment
    assessment = Hmis::Hud::CustomAssessment.new(
      enrollment: self,
      client: client,
      data_source: data_source,
      assessment_date: entry_date,
      user: user, # same user that's on the enrollment
      data_collection_stage: 1, # Intake
      wip: false,
    )
    # Build a FormProcessor. It has no references since no records are created. It is not necessary
    # to link a Form Definition, because a Form Defintion will be selected when/if the form is opened for editing.
    assessment.build_form_processor
    assessment
  end
```

Then here we can do:
```
            # Save the enrollment as non-WIP
            record.save_not_in_progress
            # Save an empty intake assessment. "Auto-entered" enrollments do not require
            # an intake. We save an empty one so that the user doesn't get
            # reminders about performing the intake.
            intake_assessment = record.build_synthetic_intake_assessment
            intake_assessment.save!
```

drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -0,0 +1,55 @@
...
+  AUTO_EXIT_CONFIG = 'Hmis::ProjectAutoExitConfig'.freeze
+  AUTO_ENTER_CONFIG = 'Hmis::ProjectAutoEnterConfig'.freeze
+  TYPE_OPTIONS = [AUTO_EXIT_CONFIG, AUTO_ENTER_CONFIG].freeze
+  validates :type, inclusion: { in: TYPE_OPTIONS }
+
+  # TODO: The following 3 methods are repeated code in auto_exit_config.rb.
+  # In the future auto_exit_config.rb will be fully replaced by project_auto_exit_config.rb.
+  def self.default_config
+    find_by(project_type: nil, organization_id: nil, project_id: nil)
+  end
```
gigxz: Hmm, so this conflicts with the `project_org_type_xor` rule. It looks like the idea here was that you could set up a default site-wide rule for auto-exit.

It makes me a little nervous, because I could see someone accidentally setting a rule for the whole site. I think we should remove this `default_config` piece entirely. I don't see a use-case for it for autoexit or autoenter. If someone did want to implement a project-level config site-wide, it could be done by adding multiple project-type-level configs. I think I would prefer that approach.

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
gigxz: This is such an FAQ and I have been meaning to write it up. Finally just did here https://github.com/greenriver/hmis-warehouse/pull/4047 (and https://github.com/greenriver/hmis-warehouse/pull/4048!) let me know if you have further questions.

We [resolve](https://github.com/greenriver/hmis-warehouse/blob/1c194524832860549913fc8f9194115ae3d001c2/drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb#L78) the database id (Enrollment.id) as the record id on the graphql Enrollment type. We never show the "HUD Key" for Enrollment in the HMIS application, so we don't even resolve it. We do use the ProjectID and OrganizationID in the interface in a few places, because it's helpful for migrated-in data to have those IDs match the old system.

db/structure.sql
gigxz: General rule is that you should back out changes to sql files if your migration(s) don't touch them. So here you should `git checkout release-105 db/structure.sql` and re commit

gigxz (commented):
looks great! 🎈  comments below

app/controllers/cohorts_controller.rb
```diff
@@ -249,7 +249,7 @@ def cohort_id
   end
 
   def load_cohort_names
-    @cohort_names ||= cohort_source.pluck(:id, :name, :short_name).  # rubocop:disable  Naming/MemoizedInstanceVariableName
```
martha, line 252: This was a change that came up during merge to keep rubocop happy(er) -- but let me know if I should revert it!

db/structure.sql
martha: Ok! Sounds good. What does this mean on a general level, though? -- has something gotten out of sync on the release branch, or am I doing something wrong when I'm running migrations?

drivers/hmis/app/graphql/mutations/submit_form.rb
martha: This makes sense! And yep the change you suggested looks good, now the intake assessment can be edited after the enrollment if you go looking for it, though it's not required. (Before your suggested change the intake assessment didn't have all the expected fields, now they are present with 'Data not collected' which I think is the expected behavior.)

drivers/hmis/app/models/hmis/project_config.rb
martha: Completely removing it sounds fine to me! I wonder if we should rename `xor` since it's not really an xor, more like validating that 0 or 1 (but not more) of those 3 attributes exist. Also, now that we are not allowing a site-wide config, should that logic be changed to exactly 1?

drivers/hmis/app/models/hmis/project_config.rb
martha: Perfect- thanks for the reminder!

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
martha: Ok, I think that makes sense -- thank you for the write-up, it's really helpful!
Does it make sense then, that in order to resolve the `CustomAssessment` that's associated with this enrollment, I've had to 'translate' from the database's ID to the hmis's UUID?

drivers/hmis/app/graphql/mutations/submit_form.rb
martha: Sounds good and thanks for the catch! I tested this locally for both wip and non-wip cases, move in date and engagement date both still seem to behave normally. I also added a note to the PT ticket that we should check this behavior in the QA environment as well.

app/controllers/cohorts_controller.rb
eanders: Thanks!

db/structure.sql
gigxz: It's normal, all developers on the team have the same issue. We get out-of-sync from running different migrations locally, mostly likely  while developing migrations. When we all have slightly different files, they continually overwrite each other in git. We re-set them every so often by collapsing the migration history. `db/warehouse/migrate/20230127151606_warehouse_db_load.rb` was the last time we did that. This file only affects the tests database so its not a huge deal.

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
gigxz: Yep it does. I think it could be adjusted to be a bit more readable, though:

```
      record = result.dig('data', 'submitForm', 'record')
      expect(record['inProgress']).to eq(false)
      e_id = Hmis::Hud::Enrollment.find(record['id']).enrollment_id
      ca = Hmis::Hud::CustomAssessment.where(enrollment_id: e_id).last
      expect(ca).not_to be_nil
      expect(ca.wip).to eq(false)
      fp = Hmis::Form::FormProcessor.where(custom_assessment: ca).last
      expect(fp).not_to be_nil
```
=>
```
      enrollment_id = result.dig('data', 'submitForm', 'record', 'id')

      enrollment = Hmis::Hud::Enrollment.find_by(id: enrollment_id)
      expect(enrollment.in_progress?).to eq(false)
      expect(enrollment.intake_assessment).to be_present
      expect(enrollment.intake_assessment.assessment_date).to eq(enrollment.entry_date)
      expect(enrollment.intake_assessment.wip).to eq(false)
      expect(enrollment.intake_assessment.form_processor).to be_present
```

drivers/hmis/app/models/hmis/project_config.rb
gigxz: Yes it should probably be exactly one. Can you make that change? sorry for the delayed response(s)!

gigxz (approved):
Just some suggestion on test cleanup, otherwise LGTM!! 💯

db/structure.sql
martha: Gotcha. Thank you!

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
martha: Oh great, this is much better!

PR #4045, opened 2024-02-22
Update Auto-Exit code to consolidate with Project Configs

martha:

## Description

Following up on the added support for Auto-Enter configs under the new generic Project Config model, this PR updates the existing Auto-Exit configs and jobs to use the new generic model.

What still remains to be done:
- Remove the old file `drivers/hmis/app/models/hmis/auto_exit_config.rb` -- which has to be done in a separate PR because it must be done after the migration in this PR runs
- Change the UI for auto-exit configs so it can be used generically to modify project configs. Possibly this should be a separate ticket as it needs some design work first?

Depends on: https://www.pivotaltracker.com/story/show/186800449 / https://github.com/greenriver/hmis-warehouse/pull/4044
PT issue: https://www.pivotaltracker.com/story/show/187091393

## Type of change
- [ ] Bug fix
- [ ] Dependency update


db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
```diff
@@ -0,0 +1,36 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
martha, line 1: Confirming, do we want to use the db migration file to do this data migration even though [some consider it an anti pattern](https://www.ombulabs.com/blog/rails/data-migrations/three-useful-data-migrations-patterns-in-rails.html)?

Related question, is it ok for this migration to cause any downtime from the auto-exit config UI's perspective? I guess what we might do if it were critical to prevent any downtime, is to modify the graphql code and the auto-exit config job code so that they can use both the old and the new model for awhile, then run the migration, then push an update again that removes the old model. I haven't done that here because @gigxz you mentioned that there is only 1 auto-exit config currently configured anywhere, so maybe this isn't where our effort should go?

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
```diff
@@ -5,6 +5,37 @@
...
+
+  def length_of_absence_days=(value)
+    self.config_options = { 'length_of_absence_days': value }.to_json
+  end
+
+  def length_of_absence_days
+    JSON.parse(config_options)['length_of_absence_days']
+  end
+
+  def length_of_absence_days_ge_30
```
martha, line 21: Does this error-checking look sufficiently rails-idiomatic to you? I would welcome feedback and improvements on style!!

drivers/hmis/spec/models/hmis/auto_exit_config.rb
```diff
@@ -1,58 +0,0 @@
-###
```
martha, line 1: GH is marking this file as deleted, but it was moved here: `drivers/hmis/spec/models/hmis/project_auto_exit_config_spec.rb`
The first test case is unchanged, and then `it 'should throw errors if the auto-exit is not configured properly' do` was added 🐥

db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
gigxz: Yep this should be fine! Agree not worth the effort, we have one job that runs in the evening

db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
```diff
@@ -0,0 +1,36 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class MigrateAutoExitRules < ActiveRecord::Migration[6.1]
+  def up
+    Hmis::AutoExitConfig.all.each do |old_auto_exit_config|
+      Hmis::ProjectAutoExitConfig.create(
```
gigxz:
```suggestion
      Hmis::ProjectAutoExitConfig.create!(
```

Most often want to use the "bang" operator when creating/updating/deleting, so that rails raises an exception if it fails.

db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
```diff
@@ -0,0 +1,36 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class MigrateAutoExitRules < ActiveRecord::Migration[6.1]
+  def up
+    Hmis::AutoExitConfig.all.each do |old_auto_exit_config|
```
gigxz: Lets do this in a transaction, so that it rolls back if any of the operations fail:
```
  Hmis::AutoExitConfig.transaction do

   # 1) iterate and create
 
   # 2) destroy
   Hmis::AutoExitConfig.each(&:destroy!)
  end
```

gigxz: It would be better practice to `destroy` the old records in a separate migration, probably, but again I'm not that worried since we only have the 1

db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
```diff
@@ -0,0 +1,36 @@
...
+      Hmis::ProjectAutoExitConfig.create(
+        project_type: old_auto_exit_config.project_type,
+        organization_id: old_auto_exit_config.organization_id,
+        project_id: old_auto_exit_config.project_id,
+        config_options: { length_of_absence_days: old_auto_exit_config.length_of_absence_days }.to_json,
+        created_at: old_auto_exit_config.created_at,
+        updated_at: old_auto_exit_config.updated_at,
+      )
```
gigxz: Isn't there an "active" or "enabled" status that we need to set here? The old table didn't have that, so we should migrate all old records as active.

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
```diff
@@ -5,6 +5,37 @@
...
+    unless config_options
+      errors.add(:base, 'config_options must be present')
+      return
+    end
```
gigxz: I think this can be replaced with a simple `validates :config_options, presence: true` on the model. Then this custom validation function can just `return unless config_options`.

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
```diff
@@ -5,6 +5,37 @@
...
+
+    begin
+      json_blob = JSON.parse(config_options)
+    rescue JSON::ParserError
+      errors.add(:base, 'config_options must be JSON')
```
gigxz: I think this can be validated on the parent model, since we know it should always be JSON.

I like the approach todd took here, with some helpers for reading and setting the json field.
https://github.com/greenriver/hmis-warehouse/blob/release-105/app/models/grda_warehouse/remote_credentials/oauth.rb

So maybe on the Hmis::ProjectConfig model:
```suggestion
  validate :validate_config_options_json

  def validate_config_options_json
    return unless config_options

    begin
      JSON.parse(config_options)
    rescue JSON::ParserError
      errors.add(:base, 'config_options must be JSON')
    end
  end

  def options= hash
    self.config_options = hash.to_json
  end

  def options
    @options ||= JSON.parse(config_options)
  rescue JSON::ParserError, TypeError
    nil
  end
```

Then this validator here can simply read the value from `options` hash and validate it as you do:
```
length_of_absence_days = options['length_of_absence_days']
```

drivers/hmis/spec/factories/hmis/hud/project_configs.rb
```diff
@@ -9,4 +9,10 @@
     created_at { Date.parse('2019-01-01') }
     updated_at { Date.parse('2019-01-01') }
   end
+
+  factory :hmis_project_auto_exit_config, class: 'Hmis::ProjectAutoExitConfig' do
+    created_at { Date.parse('2019-01-01') }
+    updated_at { Date.parse('2019-01-01') }
```
gigxz: Can you switch all these to `Time.current`? The timestamp fields should be DateTimes not Dates

gigxz: And as an aside, we always make sure to use `.current` (NOT `.now` or `.today`) in tests (and everywhere, really) because they can cause timezone issues: https://discuss.rubyonrails.org/t/time-now-vs-time-current-vs-datetime-now/75183

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
gigxz: Added some comments below 😃

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
```diff
@@ -5,6 +5,37 @@
...
+
+    length_of_absence_days = json_blob['length_of_absence_days']
+    unless length_of_absence_days.is_a? Integer
+      errors.add(:base, 'config_options must contain an integer length_of_absence_days')
+      return
+    end
+
+    return unless length_of_absence_days < 30
+
+    errors.add(:base, 'length_of_absence_days must greater than or equal to 30')
```
gigxz: Rails errors should be human-readable:
```suggestion
    errors.add(:base, 'Length of Absence must greater than or equal to 30 days')
```

We do display them directly to the user:
Screenshot 2024-02-26 at 5 19 21 PM



It's not perfectly executed in this case, because the input field isn't tied to the error message. But I think that will come later with re-designing the AutoExit UI into a more general project configuration tool.

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
```diff
@@ -5,6 +5,37 @@
...
+    begin
+      json_blob = JSON.parse(config_options)
+    rescue JSON::ParserError
+      errors.add(:base, 'config_options must be JSON')
+      return
+    end
+
+    length_of_absence_days = json_blob['length_of_absence_days']
+    unless length_of_absence_days.is_a? Integer
+      errors.add(:base, 'config_options must contain an integer length_of_absence_days')
```
gigxz:
```suggestion
      errors.add(:base, 'Length of Absence is required')
```

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
martha: Okay sounds good!

This line was giving me trouble:
`@options ||= JSON.parse(config_options)`
Creating an `options` instance variable I think was causing `options` and `config_options` to get out of sync, so for instance (in the rails console),
```
irb(main):024:0> aec.options = {"length_of_absence_days": 40}
=> {:length_of_absence_days=>40}
irb(main):025:0> aec.options
=> {"length_of_absence_days"=>40}
irb(main):026:0> aec.config_options
=> "{\"length_of_absence_days\":40}"
irb(main):027:0> aec.config_options = '{"foo":"bar"}'
=> "{\"foo\":\"bar\"}"
irb(main):030:0> aec.config_options
=> "{\"foo\":\"bar\"}"
irb(main):029:0> aec.options
=> {"length_of_absence_days"=>40}
```

I've instead just made options return `JSON.parse(config_options)` so there is only 1 source of truth and no other info being stored. Let me know if this looks ok.

db/warehouse/migrate/20240221195839_migrate_auto_exit_rules.rb
martha: Yes good catch, thank you!

drivers/hmis/spec/factories/hmis/hud/project_configs.rb
martha: Cool, thank you!

drivers/hmis/app/models/hmis/project_auto_exit_config.rb
gigxz: Sounds great, thanks

gigxz:
sorry @martha looks like there were some sticky merge conflicts, I hope its not too bad to resolve! I think it's because of the squash-and-merge approach for PRs going into release branch.

gigxz (approved):
LGTM once conflicts are resolved. thank you!!

martha:
@gigxz retargeting at `release-106` fixes the merge conflict, is this OK to do?

gigxz:
That's just because release-106 was behind 105 @martha  but I just updated it. Good idea to target to 106 but the conflicts still need to be resolved unfortunately

PR #4055, opened 2024-02-23
Make AutoExit and AutoEnter generic into ProjectConfig

martha:

## Description

After the work on Auto Enter configs, we had discussed changing the existing auto-exit-config UI and graphql migrations to make them all generic for managing project configs.
PT issue: https://www.pivotaltracker.com/story/show/187091393

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/hmis_schema/enums/project_config_type.rb
```diff
@@ -0,0 +1,23 @@
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class HmisSchema::Enums::ProjectConfigType < Types::BaseEnum
+    graphql_name 'HmisProjectConfigType'
+
+    def self.get_project_config_type(cls_name)
```
martha: One thing I'd like feedback on -- do you think this code is being a bit too clever for its own good and perhaps sacrificing readability in pursuit of DRY-ness?

drivers/hmis/lib/form_data/static/project_config.json
```diff
@@ -1,10 +1,15 @@
 {
   "item": [
     {
-      "type": "INTEGER",
-      "link_id": "lengthOfAbsenceDays",
-      "text": "Length of Absence (Days)",
-      "mapping": { "field_name": "lengthOfAbsenceDays" }
+      "type": "CHOICE",
```
martha, line 4: I think the most sensible from both a UX and code perspective would be the Config Type is read-only after creation. So if the record doesn't already exist, this dropdown can be used to select what type of ProjectConfig it should be, but if the record does already exist, then it can't be changed.

Do we do anything similar to this already? If not, the approach I had been considering was to set up something similar to `enable_when`, like `read_only_when`, and I wanted to check for your thoughts on that before proceeding!

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -410,11 +410,11 @@ def form_rule(id:)
       Hmis::Form::Instance.find_by(id: id)
     end
 
-    field :auto_exit_configs, Types::HmisSchema::AutoExitConfig.page_type, null: false
```
gigxz, line 413: This is OK for this field, but for future reference:

If you need to remove a field or query that's actively being used by the interface in the previous release (release-105), we want to deprecate the field rather than removing it. (By adding arg `deprecation_reason: 'something'`) https://graphql-ruby.org/fields/introduction.html#field-documentation

This is because when a deploy occurs while someone is using the application, their "old" frontend starts talking to the new backend version, while they still have that frontend js cached in their browser. So the user would be running release-105 on the frontend, and release-106 on the backend. (Even though the deploys happen at the same time, this situation still comes up because of the browser caching). So - in that scenario, if the frontend tries to request a field or query that no longer exists, they will see an error. For that reason we are starting to try to make API changes backwards-compatible as much as possible. Deprecated fields can be removed in the following release (we don't really have a process for that yet though)

The reason I say don't worry about it here, is because GR Devs are the only users that currently have access to this interface. Other users won't be making this query, so they won't hit the above scenario. Just wanted to explain that! [this ticket](https://www.pivotaltracker.com/story/show/184908546/comments/240087679) tracks some of the thinking about this over time.

drivers/hmis/lib/form_data/static/project_config.json
gigxz: Let's leave this as-is and revisit with the larger design discussion around config tooling.

My immediate thought is that we should end up with separate sections per config type anyway. (E.g. One table for "Auto Exit Configurations" and one table for "Auto Enter Configurations" each with separate forms that wouldn't show the "Config Type" dropdown to the user at all. I like that approach, but lets hold off until we get some designs together.

drivers/hmis/app/graphql/types/hmis_schema/enums/project_config_type.rb
gigxz: Yeah I think so. We shouldn't have terribly many of these. I think it would be preferable to just list them out individually as `value`s.

drivers/hmis/app/graphql/types/hmis_schema/enums/project_config_type.rb
```diff
@@ -0,0 +1,23 @@
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class HmisSchema::Enums::ProjectConfigType < Types::BaseEnum
+    graphql_name 'HmisProjectConfigType'
```
gigxz:
```suggestion
    graphql_name 'ProjectConfigType'
```

No need to prefix with `Hmis` in the graphql schema

drivers/hmis/app/graphql/types/hmis_schema/enums/project_config_type.rb
gigxz: Oh, I see how the value is working now too, with the `input.config_type.constantize` in the mutation. That feels a bit overly clever to me too tbh. I think just for readability, my preference would be to change `CreateProjectConfig` to just use a `case` statement on the expected values for `input.config_type` (AUTO_EXIT, AUTO_ENTER), and raise an error if its an unrecognized type. I think it will make it easier for future-selves-and-others to understand how project configs work.

```
  class_name = case input.config_type
      when 'AUTO_ENTER"
        Hmis::ProjectAutoEnterConfig
      when 'AUTO_EXIT"
        Hmis::AutoExitConfig
      else
        raise "Unsupported type: #{input.config_type}"
      end
```

gigxz:
A few comments about complexity but otherwise looks great!

drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -21,7 +21,6 @@ def exactly_one_of_project_org_type
   AUTO_ENTER_CONFIG = 'Hmis::ProjectAutoEnterConfig'.freeze
   TYPE_OPTIONS = [AUTO_EXIT_CONFIG, AUTO_ENTER_CONFIG].freeze
   validates :type, inclusion: { in: TYPE_OPTIONS }
-
   validate :validate_config_options_json
 
   def validate_config_options_json
```
martha: @gigxz in this file, I was thinking `config_for_project` should be renamed to something more clear, now that the former class method `configs_for_project` has turned into the scope `for_project`. But, I was kind of stuck coming up with a good name because I realized I don't really understand what this code is doing or why it's necessary:
```
[
  :project_id,
  :organization_id,
  :project_type,
].each do |field|
  scope = configs.where.not(field => nil)
  return scope.first if scope.exists?
end
```

Why is this needed instead of just returning `scope.first` after line 63?

drivers/hmis/lib/form_data/static/project_config.json
```diff
@@ -1,10 +1,15 @@
...
+      "link_id": "configType",
+      "text": "Config Type",
+      "pick_list_reference": "ProjectConfigType",
+      "mapping": {
```
martha, line 8: I noticed that, if there is an existing auto_exit_config row in the `hmis_form_definitions` table, then the frontend /admins/form interface will fail with a message saying that the form role is not allowed. Should we write another migration to mitigate this?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -76,7 +76,7 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   # Non-configurable forms. These are submitted using custom mutations.
   STATIC_FORM_ROLES = [
     :FORM_RULE,
-    :AUTO_EXIT_CONFIG,
+    :PROJECT_CONFIG,
```
gigxz, line 80: @martha I think because you removed the `AUTO_EXIT_CONFIG` role here which drops it from the graphql API. Lets leave it and we can remove it later once the migrations have been run

drivers/hmis/app/models/hmis/project_config.rb
gigxz: @martha good question. This logic is for selecting the "most specific" rule that is applied. For auto-exit for example, you might have one config that says "auto exit after 90 days for all NBN shelters", and another config that says "auto exit after 30 days for This Specific Shelter". The specificity order is Project>Org>ProjectType, so in that scenario the Project-specific rule would be selected.


There is a similar function on Form Instance with signature `def self.detect_best_instance_for_project(project:)`. That is wordy but descriptive. Maybe we should keep the same pattern here and call it `detect_best_config_for_project`.

martha: Perfect!

drivers/hmis/app/models/hmis/form/definition.rb
martha: Good thinking. Thanks!

gigxz (approved):
👍  great!

gigxz:
The test failure looks like it might be related to the leap year, unrelated

PR #4057, opened 2024-02-26
Binstub regeneration

yakloinsteak:
## Description

`bundle binstubs --force --all --path /app/bin` run, but I have not tested.


eanders:
Closing, these seem to not quite be right, and the binstubs generated by `rails app:update:bin` are very similar to the existing stubs.  We can revisit as part of the rails 7 upgrade.

PR #4061, opened 2024-02-27
active_cohort column breaking on download when cohort names does not exist

dtgreiner:

## Description

Fixes issue when creating the cohorts export when the active cohorts column is selected and the cohort_names does not exist.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/models/cohort_columns/active_cohorts.rb
+++ b/app/models/cohort_columns/active_cohorts.rb
@@ -18,7 +18,7 @@ module CohortColumns
 
     def value(cohort_client) # OK
       cohort_ids = cohort_client.client.active_cohort_ids - [cohort.id]
-      cohort_names.values_at(*cohort_ids).join('; ')
+      cohort_names&.values_at(*cohort_ids)&.join('; ')
     end
   end
 end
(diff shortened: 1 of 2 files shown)


eanders (approved):
Thanks @dtgreiner !

eanders (changes requested):
@dtgreiner something really odd is going on with this.  It no longer throws a 500, but the results in the download are empty for this column, though they show in the cohort.  Can you test with a cohort with a client who has at least 2 cohorts and see if you can track down why that is empty?

dtgreiner:
@eanders It looks like it was a deeper issue than the missing &'s. I missed adding @cohort_names to the export file when backgrounding this process. When adding this, the previous update is working without the added &'s. Do we want to keep those in just in case, or remove them so it flags an issue if something like this pops up again

eanders:
@dtgreiner good question.  Let's keep them.  I'd rather have the file run than fail completely I think.

dtgreiner:
@eanders this should be good to go now

eanders (approved):
Works as advertised, thanks!

PR #4063, opened 2024-02-27
Label custom assessments in audit events

martha:

## Description

Admins would like to use the User Audit history page to track users performing custom assessments, but as it stands now, these are just labelled "Assessment" without any further context when it's a non-HUD assessment. This change adds the custom assessment's title if applicable.
PT issue: https://www.pivotaltracker.com/n/projects/2591838/stories/187048098

It works for me testing using the UI:
Screenshot 2024-02-27 at 2 25 37 PM

I added an rspec test for this, and would like to confirm this is in the right place -- this is a display-related change, but not actually a graphql change, so I wasn't sure.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


martha:
Hm also I remember that yesterday we discussed making this a left join for performance purposes, which I'm not sure exactly how to do (using `load_ar_association`?) I will convert the PR to draft mode and keep thinking on it!

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -110,6 +111,11 @@ def graphql_type
       end
     end
 
+    private def custom_assessment_title
+      ca = load_ar_scope(scope: Hmis::Hud::CustomAssessment.with_deleted, id: object.item_id)
```
martha, line 115: A quick interesting note about using `with_deleted` here -- I only happened to notice this because I've been working recently on the bug ticket about custom assessment deletion, but this doesn't work correctly (won't show the custom assessment's real name, just 'Assessment') when the assessment has been deleted, even though I added `with_deleted` here. The problem is that `FormProcessor` doesn't have `acts_as_paranoid`, so it is really deleted from the DB even when the associated Assessment is just soft-deleted.

gigxz: Ah good find!! That definitely seems problematic because we would not be able to restore an accidentally deleted assessment. I created a [story](https://www.pivotaltracker.com/story/show/187150766) for moving form processor to soft-deletion.

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -110,6 +111,11 @@ def graphql_type
...
+      ca&.form_processor&.definition&.title
```
gigxz: This will create an N+1 I believe, because each audit even in the table will be joining to form processor and definition separately. I'm not sure if there is a way to avoid it in this case though. It sounded like @ttoomey  had some ideas there?

drivers/hmis/app/graphql/types/base_audit_event.rb
martha: Thank you!

drivers/hmis/app/graphql/types/base_audit_event.rb
ttoomey: To avoid n+1 issues in graphql, you should be able to use the data loader pattern rather than the normal active record associations. Looks like CustomAssessment has a relationship to definition so we can do something like this:

```suggestion
      ca ? load_ar_association(object, :definition)&.title : nil
```

ttoomey (approved):
Looks good to me, thanks! Just needs a small fix for that n+1 issue

drivers/hmis/app/graphql/types/base_audit_event.rb
martha: Something is funny with this -- I think because `object` is a `GrdaWarehouse::Version`, and not a `CustomAssessment` ? 
```
Association named 'definition' was not found on GrdaWarehouse::Version; perhaps you misspelled it?
```

ttoomey: oh shoot sorry! I think that should be `ca` instead of `object`- something like:
```ruby
ca ? load_ar_association(ca, :definition)&.title : nil
```

martha: I've updated it to use the `changed_record` from line 120 (just below), let me know if you think that is correct

martha: Ah no you're totally right, it should be `ca` and not `changed_record`, because `changed_record` will return nil if the item has been soft deleted.

I was thinking since these are audit events it might be nice to change that, and end up with something like this:
```
private def custom_assessment_title
  return unless changed_record.is_a?(Hmis::Hud::CustomAssessment)
  
  load_ar_association(changed_record, :definition)&.title
end

private def changed_record
  load_ar_association(object, :item, scope: with_deleted)
end
```
but the with_deleted scope does not work here because not all of the objects we might be auditing act as paranoid.

(Anyway, this seems like it may be scope creep 🦂)

ttoomey: Hmm, you  might be able to achieve something like that by introspecting on the item.owner_type to see if the scope should use `with_deleted`. Something like this: 
`object.item_type.constantize.paranoid?`

I'm not sure it's really getting us much here though; it seems okay to leave as it is

martha: Yes, right now it wouldn't be getting us anything since the FormProcessor models don't act as paranoid anyway! But I added a comment on this [PT issue](https://www.pivotaltracker.com/n/projects/2591838/stories/187150766) so we remember to come back to it when we make that change

gigxz (approved):

PR #4064, opened 2024-02-27
Fix bug with CE Assessment and Event deletion

martha:

## Description

This fixes a bug where, when custom assessments that create CE Events and CE Assessments were deleted, the related records were not getting properly deleted.

PT issue: https://www.pivotaltracker.com/story/show/187118113

Questions
- Are there any other factories in this file that should be updated with similar logic? e.g. `current_living_situation_factory`, `service_factory`?
- What is the best way to test HAT and SPDAT locally? I dug around a tiny bit and found the HatLoader, should I use this? How should it be used? (I tested this PR using a local copy of [this custom assessment form](https://qa-hmis.openpath.host/admin/forms/72) linked from the ticket, but the HAT I haven't been able to test fully yet I think because of custom attributes that don't exist in the backend) 

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1929,6 +1929,15 @@ def process_assessment(new_hud_values: nil)
           and change { assessment.reload.ce_event.event }.to(1) # value updated
       end
 
+      it 'should find the CE event related record' do
+        process_assessment
+        expect(assessment.ce_event.event).to eq(2)
+        related_records = assessment.form_processor.related_records
+        expect(related_records.length).to eq(2)
+        expect(related_records[0].class.name).to eq('Hmis::Hud::Enrollment')
+        expect(related_records[1].class.name).to eq('Hmis::Hud::Event') # Coordinated Entry event
```
gigxz: Thank you for adding this!

I think these assertions could be replaced with `expect(related_records).to include(e1)` since that is really what we want to test here.

The indexing-into-an-array caught my eye here, because the order  of the related records really shouldn't matter.  Another way to check the class names would be `expect(related_records.map(&:class)).to contain_exactly(Hmis::Hud::Enrollment, Hmis::Hud::Event)` (again I don't think it's really necessary for this test, but also shouldn't hurt).

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -208,13 +208,17 @@ def exit_factory(create: true)
   def ce_assessment_factory(create: true)
     return owner if owner.is_a? Hmis::Hud::Assessment
 
-    self.ce_assessment ||= enrollment_factory.assessments.build(user_id: custom_assessment&.user_id) if create
+    return ce_assessment if ce_assessment.present? || !create
+
+    self.ce_assessment ||= enrollment_factory.assessments.build(user_id: custom_assessment&.user_id)
```
gigxz: Ah good find. The same issue is probably there with current living situation factory, too, since that is also one that is set up to be optionally collected on Custom Assessments. Can you fix that one as well?

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -208,13 +208,17 @@ def exit_factory(create: true)
   def ce_assessment_factory(create: true)
     return owner if owner.is_a? Hmis::Hud::Assessment
 
-    self.ce_assessment ||= enrollment_factory.assessments.build(user_id: custom_assessment&.user_id) if create
+    return ce_assessment if ce_assessment.present? || !create
```
gigxz, line 213:
```suggestion
    return ce_assessment if !create
```

I think this would be a little clearer.

* If we are _not_ creating a new record, than we just return what we have.
* If we _are_ creating a new record, then we return what we have OR we build a new one. Thats what the conditional assignment operator `||=` does below. (It only executes the right-side if the the left-side is not present.

gigxz:
@martha  re:testing, do you have a custom assessment FormDefinition called "qa ce event assessment"? It should load from `drivers/hmis/lib/form_data/qa_hmis/custom_assessments/qa_ce_event_assessment.json` with the seed_definition task, when run locally. You may just need to enable it by adding an active form rule. That one is set up to test a custom assessment that conditionally collects a CE Event.

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: Although, now I see you are following the pattern from a bunch of other factories. Maybe then it is better to leave what you have but replace the `||=` with a `=`. Lets go with that for consistency!

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: And, just to give some more context here... CurrentLivingSituation, CE Assessment, and CE Event were updated recently (at which point I introduced this bug) because they special in that they can be collected in 2 types of forms:
1. A form where the [Event] record is the form **owner**. In the interface, that would be collection under the CE 
Events tab on the enrollment dashboard.
1. A form where the [Event] record is a related record, and a Custom Assessment is the form **owner**. In the interface, that would be a Custom Assessment that collects a CE Event.

martha:
@gigxz I must be missing something obvious, but, I do see the `qa_ce_event_assessment.json` in my repo but I don't see it show up in the `/admin/forms` UI even after running `rails driver:hmis:seed_definitions` from the dcr shell and then `yarn graphql:codegen` from the frontend repo.

I was wondering if it was because these assessments are in the `qa_hmis` folder, so I tried switching my `CLIENT` env var in `.env.development.local` to `qa_hmis` (was <> before, since working on the TB dates ticket), and then re-running the above, but I didn't have any luck.

gigxz:
@martha qa_hmis forms will load by default in development, unless you have `CLIENT=` set which would override it. https://github.com/greenriver/hmis-warehouse/blob/860b84166a18bbe6993d1052e386bdf929ff847f/drivers/hmis/lib/hmis_util/json_forms.rb#L31-L36 try unsetting CLIENT, bringing containers up and down, exit re enter shell, and re-seed? happy to jump on a call if youre still stuck!

martha:
Thanks @gigxz this worked! (it was the step of bringing `dcr shell` back down and up that I had missed.)
What about testing the HAT form? I don't see this one's json file on branch release-106.
This could probably be merged and tested in QA, but I think it would be good for me to test it locally first!

gigxz:
@martha I think testing on QA is totally fine- since this is in release-106 we have plenty of time before it makes it to prod. But to test the CE Assessment behavior locally, I would just create a form similar to `drivers/hmis/lib/form_data/qa_hmis/custom_assessments/qa_ce_event_assessment.json` that collects ce assessment instead of ce event. For example (pulled from hat):

```
        {
          "text": "Assessment Date",
          "type": "DATE",
          "link_id": "hat-assessment-date",
          "required": true,
          "assessment_date": true,
          "mapping": {
            "record_type": "ASSESSMENT",
            "field_name": "assessmentDate"
          },
          "bounds": [
            {
              "id": "max",
              "_comment": "cannot be in the future",
              "type": "MAX",
              "value_local_constant": "$today"
            },
            {
              "id": "max-exit",
              "_comment": "cannot be after exit date",
              "type": "MAX",
              "value_local_constant": "$exitDate"
            },
            {
              "id": "min",
              "_comment": "cannot be before entry date",
              "type": "MIN",
              "value_local_constant": "$entryDate"
            }
          ]
        },
        {
          "text": "Assessment Location",
          "type": "STRING",
          "initial": [
            {
              "initial_behavior": "OVERWRITE",
              "value_local_constant": "$projectName"
            }
          ],
          "link_id": "hat-assessment-location",
          "required": true,
          "hidden": true,
          "mapping": {
            "record_type": "ASSESSMENT",
            "field_name": "assessmentLocation"
          }
        },
        {
          "text": "Assessment Type",
          "type": "CHOICE",
          "link_id": "hat-assessment-type",
          "required": true,
          "pick_list_reference": "AssessmentType",
          "mapping": {
            "record_type": "ASSESSMENT",
            "field_name": "assessmentType"
          }
        },
        {
          "text": "Assessment Level",
          "type": "CHOICE",
          "hidden": true,
          "initial": [
            {
              "value_code": "HOUSING_NEEDS_ASSESSMENT",
              "initial_behavior": "OVERWRITE"
            }
          ],
          "link_id": "4.19.4",
          "mapping": {
            "field_name": "assessmentLevel",
            "record_type": "ASSESSMENT"
          },
          "required": true,
          "pick_list_reference": "AssessmentLevel"
        },
        {
          "text": "Prioritization Status",
          "type": "CHOICE",
          "link_id": "hat-assessment-prioritization",
          "required": true,
          "pick_list_reference": "PrioritizationStatus",
          "mapping": {
            "record_type": "ASSESSMENT",
            "field_name": "prioritizationStatus"
          },
          "hidden": true,
          "initial": [
            {
              "value_code": "PLACED_ON_PRIORITIZATION_LIST",
              "initial_behavior": "OVERWRITE"
            }
          ]
        }
```

martha:
Oh wonderful. This works -- thank you!

gigxz (approved):

PR #4068, opened 2024-02-28
Make client alert expiry minimum tomorrow

martha:

## Description

This fixes a bug where creating an alert whose expiry date is equal to today causes the alert to immediately disappear since it's already expired. The desired behavior instead is that users shouldn't be able to create alerts that expire today. Alerts should expire tomorrow or later.
PT issue: [#187135490](https://www.pivotaltracker.com/story/show/187135490)
For once, I think, this warehouse PR depends on the frontend PR and not the other way around? 😮 so I'm not sure what that means with regards to the timing of releases, but let me know if there is anything I can do to help with keeping track of the timing relative to the release schedules of the two different repos.
https://github.com/greenriver/hmis-frontend/pull/670

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/lib/form_data/static/client_alert.json
+++ b/drivers/hmis/lib/form_data/static/client_alert.json
@@ -34,7 +34,8 @@
           "id": "min-expiration",
           "type": "MIN",
           "_comment": "entry date cannot be in the past",
-          "value_local_constant": "$tomorrow"
+          "value_local_constant": "$today",
+          "offset": 1
         }
       ]
     }
--- a/drivers/hmis/lib/form_data/static/client_alert.json
+++ b/drivers/hmis/lib/form_data/static/client_alert.json
@@ -34,7 +34,7 @@
           "id": "min-expiration",
           "type": "MIN",
           "_comment": "entry date cannot be in the past",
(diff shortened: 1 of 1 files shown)


drivers/hmis/lib/form_data/static/client_alert.json
```diff
@@ -34,7 +34,7 @@
           "id": "min-expiration",
           "type": "MIN",
           "_comment": "entry date cannot be in the past",
-          "value_local_constant": "$today"
+          "value_local_constant": "$tomorrow"
```
gigxz: You can actually do this with  `"value_local_constant": "$today", "offset": 1` already. We needed this to specify some different bounds that are needed around 7-14-30 days in the future or after entry/exit, and so on.
https://github.com/greenriver/hmis-warehouse/blob/860b84166a18bbe6993d1052e386bdf929ff847f/drivers/hmis/lib/form_data/allegheny/records/referral_request.json#L27

martha: Oh wonderful 😅  then I'll close the frontend PR and update this one. Thank you!

gigxz (approved):

PR #4071, opened 2024-02-29
Superset roles [devops 702]

yakloinsteak:
## Description

Removes superset config which will live in the superset-sync repo. Provide list of role names to superset so it can determine the corresponding role on that side. 

How to test:
- Needs a forthcoming update to superset to be testable. Ask @yakloinsteak to know if it's ready.

On login, if
- your email ends in greenriver.(org|com), you should become a superset **Admin**
- you set the superset_role value to a valid superset role, you become that role in superset
- If you set nothing for the superset roles and your email isn't a greenriver email, you become a report dashboard reader, which only exists for M***. populating that role has not been automated yet.

## Type of change


app/controllers/oauth_controller.rb
```diff
@@ -14,6 +14,14 @@ class OauthController < ApplicationController
   def user
     user = User.find(doorkeeper_token.resource_owner_id)
 
-    render(json: { id: user.id, first_name: user.first_name, last_name: user.last_name, email: user.email })
+    payload = {
+      id: user.id,
+      first_name: user.first_name,
+      last_name: user.last_name,
+      email: user.email,
+      roles: user.legacy_roles.map(&:name),
```
eanders: Oh, maybe this wasn't clear, but the warehouse roles won't align with the superset roles.  We only plan to send two roles to Superset (default and admin.)  Ideally, anyone with a `@greenriver.xxx` in their email would get admin, everyone else would get the default role.

yakloinsteak: I know they won't, but won't they be able to inform the role to use in superset? I think it will be more flexible than matching on just the email address. I was definitely following the code before, but why wouldn't you want to base the superset role on the warehouse role and email rather than just an email address?  Maybe we should talk. We can still make all greenriver email account admins, but we gain the ability to look at the warehouse roles to make a choice about the superset role if we have more of those in the future. We also can choose to make some greenriver emails non-admins for easier testing. Like tblackman+non-admin-testing@greenriver.com

eanders: So, the roles in the warehouse are 100% flexible, and in flux at the moment (as you noted with the ACL transition).  There are no consistent roles between installations and no default permissions.  At the moment we are only planning two roles (maybe plus a public role) in Superset, and some installations have 20+ roles in the warehouse.  I guess I'd rather base it off of something other than roles, I think that will add significant complexity later.  Perhaps we should add a field to the user that encapsulates their superset access.

yakloinsteak: It's up to you. I'll take this line out and just base it on the email on the superset side, but I just don't think that's what we want to be doing. Will anybody manually change anybody's role on the superset side? I need to know that to make sure I get this right over there.

eanders: Let's add a `superset_role` column to `users` and store a comma delimited string (or serialized array) of roles in there.  Then we can adjust the value in the warehouse DB and assume we won't need to adjust manually in Superset.  Does that work for you?

yakloinsteak: That works. I've made the superset side honor the green river email address as a fallback if there are no roles returned from the warehouse. That should let the two things get deployed independently I think. We can pull that out of the superset side (if you want) as a later step.

eanders: Thank you!

PR #4076, opened 2024-03-01
Add new option and functionality to MigrateAssessmentsJob

martha:

## Description

PT ticket: [186477869](https://www.pivotaltracker.com/n/projects/2591838/stories/186477869)

Makes two changes to the MigrateAssessmentsJob:
- add the `generate_empty_intakes` option to the job. When true, the job creates empty intake assessments (using the same `build_synthetic_intake_assessment` created for auto-entry for non-hud projects).
- change the clobber behavior so it only clobbers custom HUD assessments and not fully custom assessments (data processing stage 99).

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
```diff
@@ -217,6 +222,17 @@ def build_assessments(enrollment_scope:, data_collection_stages:, unique_by_info
...
       end
 
+      # If there weren't any related records, then no intake assessment would have been created by this point.
+      # This script provides the option generate_empty_intakes to create empty intake assessments,
+      # so the UI won't prompt the user to complete intake assessments on these enrollments.
+      if generate_empty_intakes && data_collection_stages.include?(1)
+        Hmis::Hud::CustomAssessment.transaction do
+          enrollment_scope.each do |e|
+            e.build_synthetic_intake_assessment.save! unless e.intake_assessment
+          end
```
gigxz: Hmm, I noticed that this will create an empty intake assessment in the event that we _tried_ to create an intake but it was invalid for some reason. I think we want to skip in that case. We usually investigate those cases individually if they come up.


Unelatedly: a minor optimization would be to filter down the `enrollment_scope` with one query, so you don't have to perform the `e.intake_assessment` lookup once per iteration.

```
      enrollments_missing_intakes = enrollment_scope.left_outer_joins(:intake_assessment).
        where(intake_assessment: { id: nil }).
        where.not(enrollment_id: skipped_intake_enrollment_ids) # enrollment_ids with intake assessments that were skipped because they were invalid

      enrollments_missing_intakes.each do |enrollment|
        enrollment.build_synthetic_intake_assessment.save!
      end
```

martha: Thanks, great points both!

Separately, when trying to write a spec for this change, I was having trouble forcing an assessment to get into an invalid state -- I ended up adding 
```ruby
 || Hmis::Hud::Validators::CustomAssessmentValidator.validate_assessment_date(assessment).any?
```
on line 214 because this isn't called automatically with `assessment.valid?`. Let me know if this is ok!

martha: update on the above - we instead decided to invalidate the assessment by deleting the client, because we actually _don't_ want to validate assessment dates on imported records. (See comments added in the assessment validator for a bit more context)

drivers/hmis/app/models/hmis/hud/validators/custom_assessment_validator.rb
```diff
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# TODO: This validator is no longer used as an ActiveRecord validator, but just holds the date validator function
+# that's called when the assessment form is submitted. It could be refactored to reduce confusion
```
gigxz, line 8: thanks for the comments!

gigxz (approved):
👍

gigxz:
I'm going to merge this into release-106 for the sake of getting it deployed sooner

PR #4082, opened 2024-03-04
Always make HUD User optional

gigxz:

## Description

HUD User should always be optional. We get imported data with `UserID`s that don't map to users in the HUD table, and that's ok.

We should have done this a while ago. Whenever data is entered into HMIS we set the UserID to the current user, so this wasn't coming up much as an error –– but I have seen it occasionally. We shouldn't be failing on this when checking `valid?` on the records.

## Type of change
- [x] Bug fix


eanders (approved):
Thank you!

martha (approved):
🎉 thanks for the detailed explanation on the PR for why the change!

PR #4085, opened 2024-03-04
Update HMIS link style

gigxz:

## Description

* update styling of hmis link
* only show link if user is a member of any HMIS groups

## Type of change

### DIFF
--- a/app/models/menu/menu.rb
+++ b/app/models/menu/menu.rb
@@ -21,15 +21,15 @@ class Menu::Menu
       menu << health_menu
       menu << data_menu
       menu << administration_menu
-      hmis_menu.each do |item|
-        menu << item
-      end
       menu << support_menu
       links_menu.each do |item|
         menu << item
       end
       menu << style_guide_menu unless Rails.env.production?
       menu << account_menu
+      hmis_menu.each do |item|
+        menu << item
+      end
     end
   end
 
@@ -285,10 +285,10 @@ class Menu::Menu
     GrdaWarehouse::DataSource.hmis.distinct.map do |hmis_ds|
       Menu::Item.new(
         user: user,
-        visible: ->(_user) { HmisEnforcement.hmis_enabled? },
+        visible: ->(user) { user.any_hmis_access? },
         path: "//#{hmis_ds.hmis}",
-        title: Translation.translate('HMIS'),
-        icon: 'icon-house',
+        title: Translation.translate('Open HMIS'),
+        icon: 'icon-link-ext',
         target: :_blank,
       )
     end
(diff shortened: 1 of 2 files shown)


app/models/menu/menu.rb
```diff
@@ -285,10 +285,10 @@ def hmis_menu
     GrdaWarehouse::DataSource.hmis.distinct.map do |hmis_ds|
       Menu::Item.new(
         user: user,
-        visible: ->(_user) { HmisEnforcement.hmis_enabled? },
+        visible: ->(user) { user.any_hmis_access? },
```
gigxz, line 288: is this ok performance-wise @eanders?

I don't like advertising this link to warehouse-only users, since they technically can go log in but won't see anything.

gigxz:
Screenshot 2024-03-04 at 5 08 56 PM

app/models/menu/menu.rb
eanders: Should be fine, it will add a single exists query.  maybe we should memoize the result.  I'll note that in the method.

app/models/user.rb
```diff
@@ -179,6 +179,12 @@ def related_hmis_user(data_source)
     Hmis::User.find(id)&.tap { |u| u.update(hmis_data_source_id: data_source.id) }
   end
 
+  def any_hmis_access?
+    return false unless HmisEnforcement.hmis_enabled?
+
+    Hmis::UserGroupMember.where(user_id: id).exists? # belongs to any HMIS user groups
```
eanders, line 185: maybe?
```suggestion
    return @any_hmis_access unless @any_hmis_access.nil?
    return false unless HmisEnforcement.hmis_enabled?

    @any_hmis_access  = Hmis::UserGroupMember.where(user_id: id).exists? # belongs to any HMIS user groups
    @@any_hmis_access
```

eanders: Not sure that gains us anything at the moment ^^^ given that we only call it once, but if we expect other HMIS links in the menu, it would mean only executing the query once.

eanders (approved):
Looks good.  Feel free to implement the memoization or not.  I don't think it gains anything at the moment.

app/models/user.rb
gigxz: Thanks, ok. I was worried that the nav template code would execute multiple times as you navigate around. I am gonna go ahead and merge. I think when we have a multi-HMIS install we will need to revisit this linking anyway .

eanders: Sounds good.  It will get executed on every page load at the moment, but hopefully that's a well optimized query.  If not, we should work on that, or cache for a few minutes.

gigxz: it should be well optimized. there is an index on the user_id column (same structure as warehouse user_group_members)

PR #4086, opened 2024-03-05
Only resolve CustomDataElements that exist

gigxz:

## Description

* **Old behavior**: When `customDataElements` are resolved on a graphql record, we resolve ALL CustomDataElementDefinitions that exist for that record type. For example when resolving an assessment, it would resolve all the CDEDs with owner_type CustomAssessment -- even ones that are totally irrelevant to the assessment.
* **Problem**: loads unnecessary data. we do not need to load CDEDs for which the record has no value.
* **New behavior**: only resolve `Types::HmisSchema::CustomDataElement`s that have 1 or more value(s)

This will require a minor frontend changes to how we choose which table columns to display: https://github.com/greenriver/hmis-frontend/pull/680 

## Type of change


drivers/hmis/app/graphql/types/hmis_schema/has_custom_data_elements.rb
```diff
@@ -31,13 +31,22 @@ def custom_data_elements_field(
...
+
+        # Load all CustomDataElementDefinitions that have values for this record
+        cde_definitions = load_ar_association(record, :custom_data_element_definitions)
+
+        # Array of relevant CustomDataElementDefinitions, with value(s)
+        cde_definitions.uniq.map do |cded|
+          OpenStruct.new(
+            **cded.slice(:id, :key, :field_type, :label, :repeats, :show_in_summary),
+            values: cde_values[cded.id],
+          )
```
gigxz: This is the main thing to look at.

I preserved the API schema, just changing how it's loaded. The OpenStruct resolves as a `Types::HmisSchema::CustomDataElement`.
It's slightly annoying that we'll need to keep adding any new attributes on CDEDs into the `slice` on line 47. I support another way this could be set up is just: `OpenStruct.new(definition: cded, values: cde_values[cded.id])`. Any preference there? It always feels a bit hacky to me to use an OpenStruct for resolving data, I don't think I have a good idea around best practices for this type of thing.

martha: If we do the second change you suggested, would that be a pretty big change to the api, or would we make a resolution elsewhere so that the type still looks the same from the frontend's perspective?

I wonder if it would be possible/desirable to move the indication of whether the field belongs in the OpenStruct closer to the source, into the CDED model, if that makes sense?-- this is not a very fleshed out idea, but some kind of annotation on the attributes of the CDED model and then here we could just introspect and list through the ones that have that annotation. We might still forget to make the annotation change when adding new fields on that model, but we might be more likely to remember if it's in the model code instead of a faraway concern. But maybe such changes will be infrequent enough to not matter.

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
```diff
@@ -21,28 +21,15 @@ class Hmis::Hud::CustomDataElementDefinition < Hmis::Hud::Base
     :json,
   ].freeze
 
-  SERVICE_OWNER_TYPES = ['Hmis::Hud::Service', 'Hmis::Hud::CustomService'].freeze
```
martha, line 24: Are these removals related to the main change in this PR or is it separate?

martha (approved):

drivers/hmis/app/graphql/types/hmis_schema/has_custom_data_elements.rb
martha: If we do the second option and it does mean a change to the api, maybe the frontend wants to consume a map instead of an array of cdes? (https://github.com/greenriver/hmis-frontend/pull/680#discussion_r1525062359)

ttoomey: I don't really mind this too much except I wonder if calling OpenStruct in a loop mints a whole new class for every item. In any case I'd prefer to peal it out into a proper named class. It's easier to debug named classes and it's more greppable. 

If we want to clean it up, I think passing the definition/values into an Adapter class like you suggest seems like less fuss to me. The Adapter class could even hold the responsibilities from `HmisSchema::CustomDataElement` type; then the type would just have field definitions without implementation. Maybe something like this:

```ruby
    GraphqlCdeValueAdapter = Struct.new(:definition, :values, keyword_init: true) do
      def id
        [definition.id, *values.map(&:id)].join(':')
      end

      def values
        definition.repeats ? values : nil
      end

      ## other methods from CustomDataElement, etc
    end

    module HasCustomDataElements
     def resolve_custom_data_elements(record)
       # ...
       cde_definitions.uniq.map do |cded|
         GraphqlCdeValueAdapter.new(definition: cded, values: cde_values[cded.id] || [])
       end
     end
    end
```

ttoomey (approved):
This is great, thank you for refactoring the CDE type! It's much easier to understand now!

drivers/hmis/app/graphql/types/hmis_schema/has_custom_data_elements.rb
gigxz: Thanks for this great suggestion!

gigxz: I wasn't quite sure where to put the new class. I added it to the same dir as the CustomDataElement models. https://github.com/greenriver/hmis-warehouse/pull/4086/commits/5c26ef44a31939c79f5217e2b80b0742cccbeefd

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
gigxz: It's unrelated, just removing unused code. Previously we thought we needed to tie the CDEDs to service types, but that didn't really make sense, since we are now using one CDED for multiple different service types.

gigxz:
unrelated test failure, fixed in https://github.com/greenriver/hmis-warehouse/pull/4125

PR #4087, opened 2024-03-05
Bulk services mutation and queries for bed nights v2

gigxz:

Frontend: https://github.com/greenriver/hmis-frontend/pull/676

## Description

* add new field to `Client` gql that lets you resolve THE active enrollment (if any) for a given date at a given project. If there are multiple (invalid) it chooses one to resolve. 🚨This feels like a new-ish pattern for gql, curious how it is received. It makes it easy to handle on the frontend.
```
client {
   id
   firstName
   // resolve zero or one enrollment
   activeEnrollment(projectId: $projectId, openOnDate: $serviceDate) {
       id
       entryDate        
   }
}
```
* new filter `serviceInRange` on Client, that lets you filter down to clients who received a specific type of service at a specific project in a specific date range. That obviously requires a whole input object just for the args for that one filter, which is also new 🚨
* new field on Enrollment to resolve the last service date of a certain typ
```
enrollment {
   lastServiceDate(serviceTypeId: $serviceTypeId)
}
```
* add `bulk` flag to indicate which service types support bulk. and `AVAILABLE_BULK_SERVICE_TYPES` pick list option.
* add `BulkAssignService` mutation to support the v2 interface. this mutation handles both creating services AND enrolling un-enrolled clients
* add `BulkRemoveService` mutation
* some refactoring for new enrollment creation

## 👋 outstanding items
- [x] Tests
- [x] Use Data Loader effectively
- [ ] ability to search for clients via `household:<id>` - need separate ticket

## Type of change


drivers/hmis/app/graphql/mutations/create_service.rb
```diff
@@ -1,38 +0,0 @@
-###
-# Copyright 2016 - 2024 Green River Data Analysis, LLC
-#
-# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
-###
-
-module Mutations
-  class CreateService < BaseMutation
```
gigxz, line 8: this mutation hasn't been used in since many releases ago so its safe to remove

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -234,6 +234,21 @@ def data_collection_features
     end.compact
   end
 
+  # Service types that are collected in this project. They are collected if they have an active form definition and instance.
+  def available_service_types
```
gigxz, line 238: refactored from `drivers/hmis/app/graphql/types/forms/pick_list_option.rb`

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -0,0 +1,104 @@
...
+      project_has_units = project.units.exists?
+      available_units = project.units.unoccupied_on(input.date_provided).order(updated_at: :desc).to_a
+
+      Hmis::Hud::Service.transaction do
+        clients.each do |client|
+          # Look for Enrollment at the project that is open on the service date
+          enrollment = client.enrollments.
+            open_on_date(input.date_provided).
+            with_project(project.id).
+            order(entry_date: :desc, date_created: :asc).last # Tie-break: newest entry date; oldest record
```
gigxz: This logic is duplicated right now with the `active_enrollment` field on Client. Need to either:
1) consolidate logic, 
2) change it so the frontend is passing enrollment id for some clients and client id for other clients. I didn't like this approach, imo cleaning for the backend to handle that, but could be swayed

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -287,7 +287,22 @@ def self.generate_enrollment_id
     generate_uuid
   end
 
-  def save_in_progress
+  def save_new_enrollment!
```
gigxz, line 292: refactored from SubmitForm

gigxz:
@ttoomey I still need to finish up the tests for this and do some optimization, but I requested your review early just to get started with feedback on the API changes (or anything else). A few things are called out in the description.

db/warehouse/migrate/20240228192937_add_bulk_flag_to_service_type.rb
```diff
@@ -0,0 +1,5 @@
+class AddBulkFlagToServiceType < ActiveRecord::Migration[6.1]
+  def change
+    add_column :CustomServiceTypes, :bulk, :boolean, default: false, null: false, comment: 'whether to support bulk service assignment for this type in the hmis application'
```
ttoomey: suggestion: maybe use a more descriptive column name?
```suggestion
    add_column :CustomServiceTypes, :supports_bulk_assignment, :boolean, default: false, null: false, comment: 'whether to support bulk service assignment for this type in the hmis application'
```

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -162,6 +164,39 @@ class << self
...
+      service_relation = :hmis_services
+      service_arel = Hmis::Hud::HmisService.arel_table
+      matches_type = Arel::Nodes::True
+    end
+
+    # Clients with services rendered
+    scope = joins(enrollments: service_relation)
+
+    # Filter down to only clients with services rendered at the specified project, if applicable
+    scope = scope.merge(Hmis::Hud::Enrollment.with_project(project_id)) if project_id
```
ttoomey, line 189: Looks like this includes wip enrollments. Those probably wouldn't have services though?

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
```diff
@@ -233,6 +236,16 @@ def self.summary_field(name, type = nil, **kwargs)
       end,
     )
 
+    # load ALL the services of this type, and find the max date. could this be optimized more?
```
ttoomey: If you're looking for optimization, would it work to use the hmis_services view so you only need one query? Not sure if it would be much faster but might be worth a try
```ruby
   load_ar_association(object, :hmis_services).
     filter {|s| s.custom_service_type_id&.to_s == service_type_id}.
     max_by(&:DateProvided)
```

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -185,6 +192,18 @@ def enrollments(**args)
...
+    def active_enrollment(project_id:, open_on_date:)
+      # TODO: data loader. not sure how to make it most efficient. Each client probably doesn't have THAT many enrollments,
+      # but this is fetched for a page of clients. The project_id (and date!) should be "constant with respect to the resolver" so maybe
+      # we can use it in the scope like below? doubt that will work, try it
+      # enrollment = load_ar_association(object, :enrollments, scope: Hmis::Hud::Enrollment.with_project(project_id).open_on_date(open_on_date).order(entry_date: :desc, date_created: :asc))&.last
+
+      object.enrollments.
+        open_on_date(open_on_date).
+        with_project(project_id).
+        order(entry_date: :desc, date_created: :asc).last
```
ttoomey: I think what you propose with the load_ar_association scope would mostly work as long as project and date really are constant. I think it would be possible to create a query where this is not true (or perhaps if queries with different params are batched together this might occur?) If it doesn't work, it might be hard to detect incorrect behavior which is a worry.

Assuming there's not too many enrollments for each client, could we just pull all the client enrollments into memory and filter them there?
```ruby
     project = load_ar_scope(Hmis::Hud::Project.viewable_by(current_user), id: project_id)
     return nil unless project

     enrollments = load_ar_association(
       object, 
       :enrollments,
       scope: Hmis::Hud::Enrollment.order(entry_date: :desc, id: :asc). # use id over date_created as tie breaker since it will be unique
         viewable_by(current_user).
         preload(:exit) # I think this works?
     )
     range = open_on_date..open_on_date
     enrollments.detect { |e| e.project_id == project.project_id && e.open_during_range?(range) }
       
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -301,14 +316,16 @@ def save_in_progress
     )
     save!(validate: false)
   end
+  alias save_in_progress save_in_progress!
 
-  def save_not_in_progress
+  def save_not_in_progress!
```
ttoomey, line 323: small scope creep, on line 324, maybe we should change destroy to raise on failure too? 
```diff
- wip&.destroy
+ wip&.destroy!
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -287,7 +287,22 @@ def self.generate_enrollment_id
...
   end
 
-  def save_in_progress
+  def save_new_enrollment!
+    raise 'Unexpected: save_new_enrollment called on a persisted enrollment' if persisted?
+
+    if Hmis::ProjectAutoEnterConfig.detect_best_config_for_project(project)
+      # If auto-enter is configured for this project, save as non-WIP and generate an empty intake.
+      save_not_in_progress!
+      build_synthetic_intake_assessment.save!
```
ttoomey, line 298: maybe these should be in a transaction together?
```suggestion
      transaction do 
        save_not_in_progress!
        build_synthetic_intake_assessment.save!
      end
```

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
ttoomey: One quick way to dry it up would be to pull it into a method on GraphqlApplicationHelper. It's not the best place for it but might be okay as a parking spot for business logic that doesn't have a natural home yet.

ttoomey (changes requested):
Code looks good overall, nice work! I haven't yet reviewed the FE PR to understand the full context for the new API pieces.

I left a few suggestions, happy to brainstorm about any performance concerns.

> add new field to Client gql that lets you resolve THE active enrollment (if any) for a given date at a given project. If there are multiple (invalid) it chooses one to resolve. 🚨This feels like a new-ish pattern for gql, curious how it is received. It makes it easy to handle on the frontend.

This seems fine to me, particularly if it makes things easier on the front-end. I guess there might be some extra steps needed on the client to keep the apollo cache consistent? 🤔 

>new filter serviceInRange on Client, that lets you filter down to clients who received a specific type of service at a specific project in a specific date range. That obviously requires a whole input object just for the args for that one filter, which is also new 🚨

This also seems fine to me. We use generated input types on other resolvers but I have no problem with explicit type declarations, IMO it's easier to understand.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: Thanks for the suggestion!
I think I like your proposal, except it will miss WIP Enrollments due to the `e.project_id` comparison. Maybe could be resolved with a preload on the wip table.

I think the project + date really are constant, at least they are in the way we are using them.. I updated and added a warning about that https://github.com/greenriver/hmis-warehouse/pull/4087/commits/e5e42e926e0bd47e10fef12419c2f4dc87eda7b8 which feels ok to me - what do you think?

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
gigxz: thanks for the suggestion!  implemented and added a test case for n+1, looks good. I also updated the `lastBedNight` field to behave in a similar way. https://github.com/greenriver/hmis-warehouse/pull/4087/commits/069cafef61e258abce59c9eb0fa49f19926fd4bd

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: They could definitely have service. Someone might be "wip enrolled" for a bed night, stay several nights, and then have the intake assessment completed a few days later. 

Added a comment to clarify intent

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: `save_new_enrollment!` is expected to be called within a transaction already. Also I just noticed `save_not_in_progress!` has a transaction in it, so we already have a problem with nesting.

From my understanding, we definitely DON'T want nested transactions within a mutation execution, because that would lead to rollbacks being ignored. Even adding `requires_new`, though, isn't what we want in these cases. I'm looking at https://api.rubyonrails.org/classes/ActiveRecord/Transactions/ClassMethods.html#module-ActiveRecord::Transactions::ClassMethods-label-Nested+transactions.

Actually this made me realize we have a few places with nested transactions that probably are at risk of committing data even if a rollback is raised. Specifically: [here](https://github.com/greenriver/hmis-warehouse/blob/d8ddb3a0fdc8fb027f10684210191e0619b71986/drivers/hmis/app/graphql/mutations/submit_form.rb#L17) we wrap the whole SubmitForm mutation in a transaction, but from within the mutation we call `enrollment.save_new_enrollment!` which, as mentioned, uses a nested transaction to delete the wip record and save the enrollment record.

I feel like the best way to deal with this is to remove transactions from the model helper methods. It is on the "caller" (eg the mutation code) to ensure that it is called within a transaction if it is updating multiple records. Does that make sense? Any other ideas?

LMK if you agree this is an issue, broadly. I will open a separate ticket.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
ttoomey: Thanks Gig, I think e5e42e9 is a reasonable solution to get this out the door.

While this case isn't likely to be a problem, I'm a uncomfortable with long-term use of resolvers that return incorrect results under some circumstances. Maybe we could plan to revisit this at some point? I'm thinking we could add an issue in the ice-box to track it or use todo-or-die

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
gigxz: Hm, actually I couldn't do that because now the shared logic differs in implementation. One uses normal AR and one uses data loaders, and the data loader functions are not available in the mutation code. Should they be? I can see this being something that comes up in the future as well -- wanting to share lookup code between query and mutations.

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: > From my understanding, we definitely DON'T want nested transactions within a mutation execution, because that would lead to rollbacks being ignored

Yes, it's also my understanding that raising `ActiveRecord::Rollback` within a nested transaction has unintuitive results. The rollback exception is special though and is used to fail the transaction without breaking execution flow. Looks like there's only 9 occurrences of this in our codebase currently. 

I believe other kinds of exceptions should bubble up to the outer transaction and prevent it from being committed. For example, if `save!` fails within a nested tx then the error should cause whole transaction to fail (unless the error is explicitly rescued within the tx).

So if we have cases where we are raising a rollback exception within a nested transaction, we should fix that. I don't see that we are doing that right now though although maybe I'm missing it? 

I agree that it's best to let the caller manage the transaction, so I'm in favor of pulling transactions up into the mutation. However I don't know that our current usage would cause problems.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: definitely. [PT story](https://www.pivotaltracker.com/story/show/187200498)

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Neeevermind..  I tested this, and it's not what I thought. TLDR no cause for concern, and I will indeed add a transaction here, even though it will lead to a nested transaction.

If there is an exception raised, it will still break out of the parent transaction and roll back everything in the parent transaction. The only [exception](https://github.com/adjust/streets/blob/677224ccebbbab210b0c852f2824b0c60045d1e8/activerecord/lib/active_record/transactions.rb#L101) is ActiveRecord::Rollback error, which will actually not roll back the changes when in a nested block:

```
def test_transactions
  Hmis::Hud::Base.transaction do
    # outer change
    Hmis::Hud::Client.find(58726).update(last_name: "changed last name")

    Hmis::Hud::Base.transaction do
      # inner change
      Hmis::Hud::Client.find(58726).update(first_name: "changed first name")

      # raise "something went wrong"    # <==  both changes are rolled back
      raise ActiveRecord::Rollback # <==   neither change is rolled back
    end
  end
end

Hmis::Hud::Client.find(58726).update(first_name: "pom", last_name: "jones"); test_transactions; Hmis::Hud::Client.find(58726).full_name
```

There are only a handful of places in the driver where we raise a `raise ActiveRecord::Rollback`, and none of them look problematic. But this is good for me to know, I didn't understand this distinction before.

cc @martha for learning purposes 🦀

gigxz: Whoops somehow I didn't see your comment before responding @ttoomey! THank you for the explanation, I got there 😅  should've done my own testing first

gigxz:
Up for re review, see frontend pr https://github.com/greenriver/hmis-frontend/pull/676  for instructions.

@ttoomey thank you for all the helpful comments already. We are doing simultaneous QA from Mary Ann on this one. So, not a huge rush to review the PR- I'm happy to keep making changes into next week, we'll just need to keep this branch deployed to a couple staging environments.

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
ttoomey: I don't think there's any technical reason not to include the data loader helpers in the base mutation. I've never tried calling a loader from within a mutation but I assume it would work. If it does, we could extract the helpers out of BaseMutation into a new module that could be included in BaseMutation and BaseObject.

Or we could try and unify all our helpers into GraphqlApplicationHelper but that might be a larger change

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -147,7 +147,7 @@ def save_submitted_assessment!(current_user:, as_wip: false)
         enrollment.save!
         enrollment.touch # Update even if no changes to Enrollment
         # Move Enrollment out of WIP if this is a submitted intake
-        enrollment.save_not_in_progress if intake?
+        enrollment.save_not_in_progress! if intake?
```
martha, line 150: To check my understanding -- above you aliased `save_not_in_progress` to `save_not_in_progress!` so this change is not necessary, but it's better to always call with the `!` to be more clear?

drivers/hmis/app/graphql/types/hmis_schema/bulk_assign_service_input.rb
```diff
@@ -0,0 +1,18 @@
...
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::BulkAssignServiceInput < Types::BaseInputObject
+    description 'Input for BulkAssignService mutation'
+    argument :project_id, ID, required: true
+    argument :client_ids, [ID], required: true, description: 'Clients that should receive service. Clients that are unenrolled in the project will be enrolled in the project.'
+    argument :service_type_id, ID, required: true
+    argument :coc_code, String, required: false, description: 'CoC code to store as EnrollmentCoC when enrolling a new client. Only needed if Project operaties in multiple CoCs.'
```
martha, line 15: What does it mean for a project to operate in multiple CoCs?

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -0,0 +1,105 @@
...
+              entry_date: input.date_provided,
+              user_id: hud_user_id,
+              household_id: Hmis::Hud::Base.generate_uuid,
+              enrollment_coc: coc_code,
+              relationship_to_hoh: 1, # Head of Household
+            )
+            raise 'unauthorized' unless can_enroll_clients
+            raise 'bulk service assignment generated invalid enrollment' unless enrollment.valid?
+
+            # Attempt to assign this enrollment to a unit if this project has units. This is AC-specific for now, and does
```
martha, line 56: What is AC-specific?

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -0,0 +1,105 @@
...
+            # not support specifying the unit type. Needs improvement if/when we expand unit capabilities.
+            if project_has_units
+              error_out('Failed to enroll client because there are no available units.') if available_units.empty?
+
+              enrollment.assign_unit(unit: available_units.pop, start_date: input.date_provided, user: current_user)
```
martha, line 61: is my understanding correct: because of `available_units.pop` here, there could be an error halfway through the list of clients being bulk enrolled, in which case this would error out, and then the whole transaction would be rolled back -- so the user would have to perform the bulk enrollment again but fix the mismatch?

drivers/hmis/app/graphql/types/hmis_schema/service_type.rb
```diff
@@ -14,6 +14,7 @@ class HmisSchema::ServiceType < Types::BaseObject
     field :id, ID, null: false
     field :name, String, null: false
     field :hud, Boolean, null: false, method: :hud_service?
+    field :supports_bulk_assignment, Boolean, null: false, default_value: false
```
martha, line 17: ~how many services support bulk assignment? Is it just bed nights for now or are there / will there be more?~ I understand there are a lot more like meals and so on

drivers/hmis/spec/requests/hmis/client_active_enrollment_spec.rb
```diff
@@ -0,0 +1,143 @@
...
+        create(:hmis_hud_enrollment, data_source: ds1, client: c, project: p1, entry_date: 1.week.ago)
+      end
+
+      expect do
+        _, result = post_graphql(dob: shared_client_dob, project_id: p1.id, date: 2.days.ago) { query }
+        clients = result.dig('data', 'clientSearch', 'nodes').map(&:deep_symbolize_keys)
+        expect(clients.size).to eq(20)
+        # make sure each of them resolved an activeEnrollment
+        expect(clients.map { |c| c[:activeEnrollment] }.compact.size).to eq(20)
+      end.to make_database_queries(count: 10..35)
```
martha: cool! ✨

martha (commented):
Thanks @gigxz for adding me to this PR, I learned a lot from reading it over! All my comments are about learning areas for me (rather than feedback/things to change) so feel free to resolve them right away, I just commented here rather than making a list and asking separately because then I could attach comments to line numbers.

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
gigxz: Cool. It does work. I moved shared code into `GraphqlApplicationHelper`. I think it is a lot cleaner! There is some code churn, but for the best imo. Commit: https://github.com/greenriver/hmis-warehouse/pull/4087/commits/6004847fe4e406caa7df66faa997b10c0aa64efd

I also removed `default_create_record` helper. We had been waiting to remove it until service creation was removed, looks like there was 1 more usage (`CreateProjectConfig`) but I refactored it.

gigxz: actually - rspec says otherwise, something might be off with the data loading .Looking.

gigxz: PHEW that was a doozy. When I started using the loader code in the mutation, I noticed some weird behavior with how the `scope:` filter worked. I switched to a simpler approach as you had originally suggested:

```
  def load_open_enrollment_for_client(client, project_id:, open_on_date:)
    # Load all visible enrollments for the client
    enrollments = load_ar_association(
      client,
      :enrollments,
      scope: Hmis::Hud::Enrollment.viewable_by(current_user).preload(:client_project, :exit),
    )

    # Filter down by project and date
    enrollments.filter do |en|
      en.open_on_date?(open_on_date) && en.client_project.project_id.to_s == project_id.to_s
    end.min_by { |e| [e.entry_date, e.id] }
  end
```

The preload does indeed work, which is great. Confirmed with the n+1 test case.

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
gigxz: I originally aliased it because I made the change to the function signature, and didn't want to break code in any places where we were calling it. However I am fairly confident that I updated the code everywhere, so I'm going to remove the alias. `!` is more clear here because it hints at the fact that this function does call `save!`

drivers/hmis/app/graphql/types/hmis_schema/bulk_assign_service_input.rb
gigxz: In data, it means that the Project has multiple ProjectCoC records. Each enrollment in the project is linked to only one CoC code, which indicates where the client was _actually_ served. In the real world, I think it is just a Project that serves clients across CoC boundaries.  CoC Code represents a geographic area.

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
gigxz: Yep exactly

drivers/hmis/app/graphql/types/hmis_schema/service_type.rb
gigxz: There will certainly be others! For example, meal service at a day shelter. It could also be used for recording "group" services like group counseling services or something like that.

drivers/hmis/spec/requests/hmis/client_active_enrollment_spec.rb
gigxz: right! it's also helpful testing tool imo to reduce the query count to something that you know will fail, because then the failure error will print a list of all the db queries that were made.

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -249,5 +264,9 @@ def occurrence_point_form_instances
     end.compact
   end
 
+  def uniq_coc_codes
+    @uniq_coc_codes ||= project_cocs.pluck(:CoCCode).uniq.compact_blank
```
ttoomey: could the order matter here?
```suggestion
    @uniq_coc_codes ||= project_cocs.pluck(:CoCCode).uniq.compact_blank.sort
```

drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -0,0 +1,106 @@
...
+            end
+
+            enrollment.save_new_enrollment!
+          end
+
+          # Initialize using the HmisService view. Based on the CustomServiceType, the class will initialize
+          # either a Hmis::Hud::Service or Hmis::Hud::CustomService as the `owner`
+          service = Hmis::Hud::HmisService.new(
```
ttoomey, line 69: Note: I know this has been around for a while but it still strikes me as odd to use HmisService as a factory. Maybe we could revisit this in future and find a better home for this responsibility

ttoomey (approved):
Looks great! 🚀

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -565,6 +565,32 @@ enum BulkActionType {
   REMOVE
 }
```
gigxz, line 567: Schema has no breaking changes, so I'm merging this despite the frontend not passing review yet. The old Bed Night interface still works with these changes.

drivers/hmis/app/graphql/types/hmis_schema/bulk_assign_service_input.rb
martha: Oh that makes sense. Thank you!

PR #4088, opened 2024-03-05
Update the HUD JSON files to pull from machine readable excel file provided by HUD

dtgreiner:

## Description

Adding a generation process to read the new machine readable excel files for the HUD Lists and generate the JSON file used in the rest of the process. 

## Type of change


dtgreiner:
@gigxz 
I've added a process for generating the HUD  list json files used in the HUD List generation process. I've got a couple things I wanted to run by you to make sure it won't break anything in HMIS. The post-processing json file is included in this PR.

1. Does capitalization matter (e.g. "Path Services" vs "Path services")? There are a good amount of descriptions with slightly varied names. in the excel file vs what we have in the current JSON file
2. There are some character differences in the list items (e.g. FundingSource (2.06.1) Item 21 has the encoded & "\u0026" instead of the &, or using ’ instead of ' ) If these are going to break things down the line, I can make adjustments in the JSON generator

Thanks!

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,108 @@ module Types::HmisSchema::Enums::Hud
...
+    json_data = JSON.parse(File.read(json_file_path))
+
+    # TODO: is there a way to better skip these?
+    ignore_codes = ['List', '*', '(see note)']
+    excel_file = Roo::Excelx.new(excel_file_path)
+
+    # Parse Excel file for list codes
+    excel_data = {}.tap do |codes|
+      # Sheet 0: CSV|DE#|Name|Type|List|Null|Notes|Validate
+      # Pull from this sheet to get the names for each list of codes
```
eanders, line 158: We should probably check the header here, to confirm it matches the expected list, and throw an error if it doesn't.

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,108 @@ module Types::HmisSchema::Enums::Hud
...
     [map_name, lookup_fn_name]
   end
+
+  def generate_hud_list_json(year, excel_file_path)
+    # Parse the existing JSON data. This will be the base from which we start.
+    json_file_path = "lib/data/#{year}_hud_lists.json"
+    json_data = JSON.parse(File.read(json_file_path))
+
+    # TODO: is there a way to better skip these?
+    ignore_codes = ['List', '*', '(see note)']
```
eanders: Seems reasonable to me.
```suggestion
    # Some codes are irrelevant to this transformation, we'll ignore them based on string comparison
    ignore_codes = ['List', '*', '(see note)']
```

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,108 @@ module Types::HmisSchema::Enums::Hud
...
+      # Pull from this sheet to get the names for each list of codes
+      excel_file.sheet(0).each(code: 'List', name: 'Name') do |row|
+        # Clean data coming in from excel
+        code = clean_json_text(row[:code].to_s)
+        list_name = clean_json_text(row[:name].to_s)
+
+        # Only need rows that have referenced codes - not all data on this sheet utilizes a code list
+        unless code.blank? || ignore_codes.include?(code)
+          codes[code] ||= {}
+          (codes[code][:list_name] ||= []) << list_name
```
eanders: That's some hairy syntax, maybe?
```suggestion
          codes[code] ||= { list_name: [] }
          codes[code][:list_name] << list_name
```

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,108 @@ module Types::HmisSchema::Enums::Hud
...
+        end
+      end
+      # Sheet 1: List|Value|Text
+      # This sheet does not have names for the lists, but it may include some that are note referenced in
```
eanders:
```suggestion
      # This sheet does not have names for the lists, but it may include some that are not referenced in
```

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,108 @@ module Types::HmisSchema::Enums::Hud
...
+        # Clean data coming in from excel
+        code = clean_json_text(row[:code].to_s)
+        value = clean_json_text(row[:value].to_s)
+        text = clean_json_text(row[:text].to_s)
+        value = value.to_i if Integer(value, exception: false)
+
+        # Skip header and blank rows if they exist
+        unless code.blank? || ignore_codes.include?(code)
+          codes[code] ||= { list_name: ['Unknown'] }
+          (codes[code][:values] ||= {})[value] = text
```
eanders: I'm probably missing something, but ?
```suggestion
          codes[code] ||= { list_name: ['Unknown'], values: {} }
          codes[code][:values][value] = text
```

lib/data/README.md
```diff
@@ -1,9 +1,18 @@
...
+```
+rails code:generate_hud_list_json\["2024","lib/data/CSV Specifications Machine-Readable_FY2024.xlsx"\]
+```
+
+This will generate the JSON files. To account for some inconsistencies with the excel data, some names will be generated as "Unknown" and will need to be verified manually with the lists specified in the HUD HMIS CSV specifications.
+
+After changes to the JSON files have been confirmed, please run the following tasks to re-generate dependent code:
 
 ```
 rails code:generate_hud_lists
```
gigxz, line 18: Cool!!

Running this (and the line below it) shows the impact of this change. We do use the labels in the HMIS UI (as does the warehouse) so the casing changes will show up in the interface. Can you run those scripts and re-push @dtgreiner so we can review the effects of this change?

I need to take time to go through it more closely. A few items of concern jump out to me, though: 
* label change `Bed Night` => `BedNight`
* label changes where the apostrophe becomes a different type of apostrophe. specifically `Client doesn’t know` I want to avoid churn on the 1.8 list (yes/no/dk/r/dnc) because we use that one for custom data elements which store the string label value). would prefer to avoid this churn
* label changes where they got more verbose (relationships_to_hoh) - I can/will add overrides to the frontend if we want to keep it this way in the backend
* labels where n-dash is replaced with m-dash -- feels like unnecessary churn but is probably fine
* `record_types` label casing became lowercase
* labels where instructional parenthesized text is ok -- maybe fine, but I want to go through them

gigxz: Also hitting this -- I had separated out all the different living situation categories in hud_lists.json so we need to keep that somehow. (probably as additional_lists)

```
82c816eeaedb:/app$ rails driver:hmis:dump_graphql_schema
rails aborted!
NoMethodError: undefined method `destinations' for HudUtility2024:Module

    hud_enum HudUtility2024.destinations
                           ^^^^^^^^^^^^^
Did you mean?  destination_type
```

dtgreiner:
@eanders @gigxz I've updated the hud list json generator to clean the apostrophes and dashes to match what is currently in the system and included the destinations list that was missing.

I've included the updated ruby generation in this PR but  I believe the goal is to get this json file close enough and then do a manual check to clean up any discrepancies before generating the ruby lists. With that said, if there are any changes needed, I think we will want to do a manual adjustment of the JSON files and then regenerate.

lib/util/hud_code_gen.rb
```diff
@@ -141,4 +142,121 @@ module Types::HmisSchema::Enums::Hud
...
+    not_numeric.sort_by { |x| [x['code'].split('.').map(&:ord), x['code']] } # Sort by numeric portions of each code
+    json_data_clean = numeric + not_numeric # Rejoin with alpha codes at the end
+
+    # Output to file
+    File.write(json_file_path, JSON.pretty_generate(json_data_clean))
+  end
+
+  private def clean_json_text(str)
+    str = str.strip.delete("\u00A0").
+      gsub(/\u2019/, "'"). # replace the character U+2019 "’" could be confused with the ASCII character U+0027 "'",
```
eanders, line 258: now we're having fun!

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -886,11 +925,11 @@ def rental_subsidy_type(id, reverse = false)
...
       {
-        1 => 'Self (head of household)',
-        2 => 'Child',
-        3 => 'Spouse or partner',
-        4 => 'Other relative',
-        5 => 'Unrelated household member',
+        1 => 'Self (Head of Household)',
+        2 => "Head of household's child",
+        3 => "Head of household's spouse or partner",
+        4 => "Head of household's other relation member",
```
eanders, line 931: Interesting that this is the language HUD provides.  @gigxz didn't we recently override these because they didn't really make sense in the app?

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1064,44 +1103,10 @@ def when_d_v_occurred(id, reverse = false)
...
-
-    def current_living_situation(id, reverse = false)
-      _translate current_living_situations, id, reverse
-    end
-
     # 4.14
     def bed_night_options
       {
-        200 => 'Bed Night',
+        200 => 'BedNight',
```
eanders, line 1109: We should add an explicit replacement for this `BedNight => Bed Night`

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1749,9 +1754,9 @@ def rhy_services(id, reverse = false)
     # R15.B
     def count_exchange_for_sexes
       {
-        1 => '1-3',
-        2 => '4-7',
-        3 => '8-11',
+        1 => '44929',
+        2 => '45023',
+        3 => '45149',
```
eanders, line 1759: This looks like a failure.  Is this wrong in the source file? We should report it if it is.
Screenshot 2024-03-11 at 4 45 54 PM

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1779,11 +1784,12 @@ def project_completion_status(id, reverse = false)
...
-        2 => 'Left for other opportunities - education',
-        3 => 'Left for other opportunities - military',
-        4 => 'Left for other opportunities - other',
-        5 => 'Needs could not be met by project',
+        1 => 'Criminal activity/destruction of property/violence',
+        2 => 'Non-compliance with project rules',
+        3 => 'Non-payment of rent/occupancy charge',
+        4 => 'Reached maximum time allowed by project',
+        5 => 'Project terminated',
+        6 => 'Unknown/disappeared',
```
eanders, line 1792: Oooh, this looks like we just had the wrong set.  And this is why we wanted to automate it.

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -2371,7 +2363,8 @@ def dependent_under_6_options
       {
         0 => 'No',
         1 => 'Youngest child is under 1 year old',
-        2 => 'Youngest child is 1 to 6 years old and/or one or more children (any age) require significant care',
+        2 => "Youngest child is 1 to 6 years old and/or one or more children (any age) require significant
+care",
```
eanders, line 2367: Hmm, is there a `\n` in there somewhere.  Maybe we should strip newlines from the content, I don't think any of the values should contain them.

drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
```diff
@@ -470,12 +470,6 @@ class VamcStationNumber < Types::BaseEnum
     hud_enum HudUtility2024.vamc_station_numbers
   end
 
-  class NoPointsYes < Types::BaseEnum
-    description 'V7.1'
-    graphql_name 'NoPointsYes'
-    hud_enum HudUtility2024.no_points_yes_options
-  end
```
eanders, line 477: This seems to be missing.

eanders (approved):
Thanks @dtgreiner dropped a couple notes, but generally I think this looks right.  @gigxz will probably have some detailed feedback around the overall json file.

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -709,10 +707,9 @@ def funding_sources
         43 => 'HUD: CoC - Youth Homeless Demonstration Program (YHDP)',
         44 => 'HUD: CoC - Joint Component TH/RRH',
         45 => 'VA: Grant Per Diem - Case Management/Housing Retention',
-        46 => 'Local or Other Funding Source',
+        46 => 'Local or Other Funding Source (Please Specify)',
         47 => 'HUD: ESG - CV',
         48 => 'HUD: HOPWA - CV',
-        49 => 'HUD: CoC - Joint Component RRH/PSH [Deprecated]',
```
gigxz, line 715: I'm pretty sure this is fine, i.e. won't have any impact in hmis. I just want to flag that we had a convo around intentionally keeping deprecated options around, so that they show up in the HMIS as their real value, rather than just "invalid value". @eanders  I think that was still the right decision. Will there be a way to do that going forward?

lib/util/concerns/hud_lists_2024.rb
gigxz: Yeah can we change these back to the old language? If you want to keep them like this in the warehouse, I can add a custom Enum for HMIS that has the labels we want.

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1194,7 +1199,7 @@ def data_collection_stages
         2 => 'Update',
         3 => 'Project exit',
         5 => 'Annual assessment',
-        6 => 'Post-exit',
+        6 => 'Post-exit (not used in CSV)',
```
gigxz, line 1202: can we remove this extra parenthesized text?

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1206,10 +1211,10 @@ def data_collection_stage(id, reverse = false)
     def moving_on_assistance_options
       {
         1 => 'Subsidized housing application assistance',
-        2 => 'Financial assistance for Moving On (e.g., security deposit, moving expenses)',
-        3 => 'Non-financial assistance for Moving On (e.g., housing navigation, transition support)',
+        2 => 'Financial assistance for Moving On',
+        3 => 'Non-financial assistance for Moving On',
```
gigxz, line 1215: Note to self @gigxz add this label back custom on the frontend. I think its helpful, but I can see why it might not be needed in the warehouse. @eanders

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -1675,6 +1680,7 @@ def reason_not_enrolleds
       {
         1 => 'Client was found ineligible for PATH',
         2 => 'Client was not enrolled for other reason(s)',
+        3 => 'Unable to locate client',
```
gigxz, line 1683: Whoops we were missing an option here. nice

lib/util/concerns/hud_lists_2024.rb
eanders: I agree, I like the old language better.  We should override.

lib/util/concerns/hud_lists_2024.rb
eanders: Good call, I think we should figure out a mechanism to preserve removed values (where we think it important, like this one.)  Maybe we can provide another set of optional lists that get merged into (reverse merged maybe) the provided lists.

drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
```diff
@@ -284,6 +266,24 @@ class DataCollectionStage < Types::BaseEnum
...
+
+  class SSVFFinancialAssistance < Types::BaseEnum
+    description 'V3.3'
+    graphql_name 'SSVFFinancialAssistance'
+    hud_enum HudUtility2024.ssvf_financial_assistance_options
+  end
+
+  class EarlyExitReason < Types::BaseEnum
+    description 'R17.A'
+    graphql_name 'EarlyExitReason'
```
gigxz, line 283: Welp, it looks like we are not using this enum in the place that we should. `hud_field :early_exit_reason, Types::HmisSchema::Enums::Hud::ExpelledReason` should be using this enum. I can make a note to myself to fix after.

lib/util/concerns/hud_lists_2024.rb
eanders: I don't think we need it in the warehouse, but I also don't think it hurts anything if we want to keep it.

drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
gigxz: thats ok it is not used in the graphql schema. IDK what this was but question V7.1 uses the standard 1.7 list.

drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
```diff
@@ -560,18 +554,6 @@ class SubsidyInformation < Types::BaseEnum
     hud_enum HudUtility2024.subsidy_informations
   end
 
-  class SubsidyInformationA < Types::BaseEnum
```
gigxz, line 563: This is removed but we need it in the schema

lib/data/README.md
gigxz: I still get errors when running `rails driver:hmis:dump_graphql_schema` due to some missing enums. Sorry I'm assuming this probably wasn't part of the original scope of this work @dtgreiner @eanders. I'm happy to take on the graphql parts of things if we want to defer that, but we probalby won't schedule work for that for several weeks, and I don't love the idea of the schema being out-of-sync for so long from the source files. Maybe we can check in about this today?

gigxz (changes requested):
has breaking changes to hmis graphql schema https://github.com/greenriver/hmis-warehouse/pull/4088/files#r1521410536

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -67,16 +67,16 @@ def disability_type(id, reverse = false)
...
-        142 => 'RHY Service Connections',
-        143 => 'HOPWA Service',
-        144 => 'SSVF Service',
-        151 => 'HOPWA Financial Assistance',
-        152 => 'SSVF Financial Assistance',
-        161 => 'PATH Referral',
-        200 => 'Bed Night',
-        210 => 'HUD-VASH OTH Voucher Tracking',
-        300 => 'Moving On Assistance',
+        141 => 'PATH service',
```
gigxz, line 70: I would prefer to keep the old casing here, but this is probably fine. I'm a little confused on which files are now manually editable vs strictly system generated.

dtgreiner:
There were some issue merging the latest data into this PR, so I've created a [new PR](https://github.com/greenriver/hmis-warehouse/pull/4278) based on the up to date release branch without the generated JSON or HudList ruby files. We are planning on manually adjusting the JSON after generation to account for any text formatting changes or other overrides from the CSV file.

PR #4093, opened 2024-03-07
Match field names and normalize location

gigxz:
Match export field names; normalize service location. I think we can skip normalizing contact type, that one looked relatively cosnsitent.


drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_services_loader.rb
```diff
@@ -354,11 +354,17 @@ def perform
           service_fields: {},
           id_prefix: 'dh-cm',
           elements: {
-            'Contact Location/Method' => {
+            # FIXME: the full label that comes through from ETO for 'Contact Location/Method' is below. Maybe we can split labels by newline and drop the rest before trying to match?
```
gigxz, line 386: this one still needs help @ttoomey

PR #4097, opened 2024-03-08
Update destinations report to include Personal ID in the client data web and excel report

dtgreiner:

## Description

Update the destinations report to include the Client Persona ID in the client details page table as well as the client details excel export.

## Type of change

### DIFF
--- a/drivers/destination_report/app/models/destination_report/details.rb
+++ b/drivers/destination_report/app/models/destination_report/details.rb
@@ -148,7 +148,7 @@ module
     def client_columns
       [
         c_t[:id],
-        c_t[:PersonalID],
+        ex_t[:PersonalID],
         c_t[:FirstName],
         c_t[:LastName],
         c_t[:DOB],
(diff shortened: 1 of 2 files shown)


drivers/destination_report/app/models/destination_report/details.rb
```diff
@@ -147,6 +148,7 @@ def client_headers
     def client_columns
       [
         c_t[:id],
+        c_t[:PersonalID],
```
eanders: One possible complicating factor.  Is this a `source_client` or `destination_client`  If these are destination clients, we'll need to fetch the PersonalIDs of their source clients...  The goal would be to get the PersonalID related to the exit, so, maybe better `ex_t[:PersonalID]` (that'd be fine as long as we have an exit record somewhere.

dtgreiner: The Exit table is within the scope where these are being referenced and the ex_t arel is loading data in both the html and the excel export. I'm updating this to use the exit arel instead of the client arel.

eanders (approved):
Works great, thanks @dtgreiner

PR #4098, opened 2024-03-08
SPM Performance and error correction

eanders:
…ged; catch missing client

## Description

This changes the SPM to only load the services needed for the report.  Additionally it attempts to catch more situations where source data has disappeared.

This also fixes the following bugs:
* The SPM was still including projects from all types if anything other than the project type was chosen in the filters when running the SPM
* The HDX upload was being generated with an incorrect `report_type` so you couldn't drill into the detail tables
* The HDX upload wasn't correctly limiting the projects included to those in the SPM unless the SPM was run only choosing project types 
* Service preloading was extremely slow when clients have long NbN enrollments, this moves those to a pluck that only brings the necessary columns, and only returns services within the report range
* Enrollments for the SPM do not need the entire width of the Enrollment table, to save memory and time, we now pull a subset of columns when loading Enrollments.
* In some situations, the HDX upload was attempting to run for data sets with no data multiple times.  If we don't find any projects in the project type, these are now skipped and not retried.
* The SPM enrollment filter was using project types codes where it needed numbers

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
+++ b/drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
@@ -7,6 +7,7 @@
 module HudApr::Dq
   class CellsController < HudApr::CellsController
     include DqConcern
+    before_action :generator
     before_action :set_report
     before_action :set_question
 
--- a/drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
+++ b/drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
@@ -7,6 +7,7 @@
 module HudApr::Dq
   class CellsController < HudApr::CellsController
     include DqConcern
+    # Make sure @generator is available.  TODO: maybe make `generator` a helper method with lazy load?
     before_action :generator
     before_action :set_report
     before_action :set_question
--- a/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
@@ -158,6 +158,8 @@ module HudSpmReport::Fy2023
               map.with_index do |date, i|
                 next unless date.between?(first_night, last_night)
 
+                # return a triple [enrollment, id, date], using the index in place of the id so we don't need to load
+                # it from the DB (middle item needs to be unique within the enrollment)
                 [enrollment, i, date]
               end.compact.group_by(&:last).
               transform_values { |v| Array.wrap(v).last }, # Unique by date
--- a/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
@@ -18,7 +18,7 @@ module HudSpmReport::Fy2023
     has_many :hud_reports_universe_members, inverse_of: :universe_membership, class_name: 'HudReports::UniverseMember', foreign_key: :universe_membership_id
 
     attr_accessor :report # FIXME?
-    attr_writer :filter
+    attr_writer :filter, :services
 
     def self.detail_headers
       client_columns = ['client_id', 'enrollment.first_name', 'enrollment.last_name', 'enrollment.personal_id']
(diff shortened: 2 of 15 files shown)


dtgreiner (approved):

drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
```diff
@@ -7,6 +7,7 @@
 module HudApr::Dq
   class CellsController < HudApr::CellsController
     include DqConcern
+    before_action :generator
```
ttoomey, line 11: Why do we call this as a before_action and not lazy load? Maybe add comment if there a side-effect we need?

drivers/hud_spm_report/spec/models/fy2023/active_record_preload_spec.rb
```diff
@@ -0,0 +1,61 @@
...
+
+  describe 'preloads work as expected' do
+    it '4 services are created' do
+      expect(GrdaWarehouse::Hud::Service.count).to eq(4)
+    end
+    it '2 services are between 2022-01-02 and 2022-01-15' do
+      expect(GrdaWarehouse::Hud::Service.where(DateProvided: '2022-01-02'.to_date .. '2022-01-15'.to_date).count).to eq(2)
+    end
+    it 'when preloading, one enrollment has 4 services' do
+      expect(GrdaWarehouse::Hud::Enrollment.preload(:services).first.services.count).to eq(4)
```
ttoomey: Is the intention of these assertions to test if it preloaded 4 services? If so should this use size instead of count? Size will use the preloaded records if available while count ignores preloaded records and makes a new query

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
```diff
@@ -148,15 +148,17 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
...
-              map do |service|
-                date = service.date_provided
-                next unless service.record_type == HudUtility2024.record_type('Bed Night', true) && date.between?(first_night, last_night)
-
-                [enrollment, service.id, date]
+            enrollment_services.
+              map.with_index do |date, i|
+                next unless date.between?(first_night, last_night)
+
+                [enrollment, i, date]
```
ttoomey, line 163: small scope creep but maybe candidate_bed_nights could use a little documentation about the return value. In particular, what's the intention of the index `i`

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/hdx_upload.rb
```diff
@@ -213,29 +213,49 @@ def spm(table_name, cell_name)
     def dq(section, table_name, cell_name)
       return unless RailsDrivers.loaded.include?(:hud_apr)
 
+      @attempted ||= Set.new
+      @reports ||= {}
+
+      # prevent retrying reports that don't have any projects
+      return '' if @attempted.include?(section) && @reports[section].nil?
```
ttoomey, line 221: do we use an empty string because this is written to a non-nullable col?

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode_batch.rb
```diff
@@ -49,16 +80,17 @@ def calculate_batch(client_ids)
     end
 
     # The associations seem to make imports run one at a time, so, they are passed separately in parallel arrays
-    private def save_episodes!(episodes, bed_nights, enrollment_links)
+    private def save_episodes!(episodes, _bed_nights, enrollment_links)
```
ttoomey, line 85: We're leaving the bed_nights var because we might want to enable this in the future?

ttoomey (approved):
Looks good, thanks for the performance improvement!

drivers/hud_apr/app/controllers/hud_apr/dq/cells_controller.rb
eanders: No particular reason, just following the existing pattern.  I _think_ `generator` isn't tagged as `helper_method` but I may have missed that.  We probably call `@generator` and this was failing prior to adding this.

Definitely worth revisiting in the future.

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode.rb
eanders: Definitely.  I don't like this pattern, returning a triple, but wanted to remain compatible with the current API.  The middle item was the services ID, and I believe just needs to be unique within the enrollment.

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/episode_batch.rb
eanders: Yes, we disabled it to prevent saving hundreds of thousands of rows that aren't currently displayed in the UI, but wanted to leave the pattern behind in case we want to re-enable that in the future.

PR #4099, opened 2024-03-08
Add static form for reviewing external submissions

martha:

## Description

This adds a static form role and json definition for the frontend form, so a user can review an external form submission, marking it as Reviewed or Spam and adding notes.

Following up on [this thread](https://github.com/greenriver/hmis-warehouse/pull/4037#discussion_r1517847147), this also makes some changes to the way spam score is passed around. It should be a boolean in the API, but a score (from recaptcha) in the backend. One thing I realized is that this means if the user manually marks something as spam, we will lose the recaptcha spam score from the database, since it gets overwritten with 1 or 0.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/lib/form_data/static/external_form_submission_review.json
+++ b/drivers/hmis/lib/form_data/static/external_form_submission_review.json
@@ -1,0 +1,22 @@
+{
+  "item": [
+    {
+      "type": "TEXT",
+      "link_id": "notes",
+      "text": "User Notes",
+      "mapping": {
+        "field_name": "notes"
+      },
+      "required": true
+    },
+    {
+      "link_id": "reviewed",
+      "type": "BOOLEAN",
+      "component": "CHECKBOX",
+      "text": "Mark reviewed",
+      "mapping": {
+        "field_name": "reviewed"
+      }
+    }
+  ]
+}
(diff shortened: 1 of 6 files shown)


drivers/hmis/lib/form_data/static/external_form_submission_review.json
```diff
@@ -0,0 +1,31 @@
...
+      "type": "TEXT",
+      "link_id": "notes",
+      "text": "User Notes",
+      "mapping": {
+        "field_name": "notes"
+      },
+      "required": true
+    },
+    {
+      "link_id": "reviewed",
```
martha, line 13: I wasn't able to find: is there an existing pattern in our form json definitions matching this image from the mocks,
Screenshot 2024-03-08 at 3 12 25 PM

? I thought of putting them in an item group with only one item in it, but it looks very different visually of course.

ttoomey: I don't know that there is a way to replicate this treatment right now. But it seems functional as-is so probably okay for now.

ttoomey (approved):
Looks good to me, thanks!

gigxz (approved):

drivers/hmis/lib/form_data/static/external_form_submission_review.json
martha: Ok, sounds good! I updated both checkboxes to use the word 'mark' for consistency.

PR #4100, opened 2024-03-09
Support searching by household id and enrollment id

gigxz:

## Description

Update `text_searcher` to support:
* `household:<household id>`
* `enrollment:<enrollment database id>`

🚨affects both Warehouse and HMIS search. tested in both.

### Why
* So that the HMIS can pre-fill a search query for bringing up all household members after a household is created
* To help with QA. It can be helpful to bring up a client quickly when you only know the enrollment id. Could be further expanded to support hud key `EnrollmentID` but I didn't see an immediate use for that personally.

## Type of change


app/models/concerns/client_search.rb
```diff
@@ -51,6 +51,14 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
           )
           where = conditions.reduce(:or)
         end
+      elsif /household:[[[:alnum:]]-]+/.match(text)
+        household_id = text.gsub('household:', '')
+        client_ids = GrdaWarehouse::Hud::Enrollment.where(HouseholdID: household_id).joins(:client).pluck(sa[:id])
```
gigxz, line 56: Another options would be to just search Personal IDs, e.g. `GrdaWarehouse::Hud::Enrollment.where(HouseholdID: household_id).pluck(:PersonalID); sa[:PersonalID].in(personal_ids)` . I chose not to do that because it could lead to errant clients showing up from other data sources when used in the warehouse.

drivers/hmis/spec/requests/hmis/client_search_spec.rb
```diff
@@ -120,60 +120,79 @@
...
+          ['text: scan card code for another client', 'P9999', false],
+          ['text: enrollment id', "enrollment:#{e1.id}", true],
+          ['text: WIP enrollment id', "enrollment:#{e2_wip.id}", true],
+          ['text: enrollment id without correct prefix', e1.id, false],
+          ['text: wrong enrollment id', 'enrollment:123', false],
+          ['text: wrong enrollment id (other hhm\'s id)', "enrollment:#{e3_other_hhm.id}", false],
+          ['text: household id', "household:#{e1.household_id}", true],
+          ['text: numeric household id', 'household:999', true],
+          ['text: WIP household id', "household:#{e2_wip.household_id}", true],
+          ['text: household id without correct prefix', e1.household_id, false],
```
gigxz, line 168: Moved these guys into a single `it` so that we could reference variables here, which feels very necessary. Is there a better way to handle that?

I added custom error messages so they still show you which case failed.

eanders: seems reasonable.

app/models/concerns/client_search.rb
eanders: I think this is great.  I like that we're starting to support faceted searching.  I wonder if the regex should allow whitespace between `household:` and the values, to support `household: 123`.  Even an optional space would probably be welcome.  That said, if the front-end is mashing these together from a custom search field specifically for household, then not having the space is totally good.

eanders (approved):
Thanks @gig!

app/models/concerns/client_search.rb
```diff
@@ -51,6 +51,14 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
           )
           where = conditions.reduce(:or)
         end
+      elsif /household:[[[:alnum:]]-]+/.match(text)
+        household_id = text.gsub('household:', '')
```
ttoomey: Nice! Some thoughts, feel free to ignore
* assuming we are storing normalized uuids, adjust the ID string so it will match by upcasing and removing hyphens. This lets someone search with a formatted uuid
* limit to some max len (uuids should be 32)
* match start/end of line to ensure there isn't other cruft in term that we don't support
```suggestion
      elsif text =~ /\Ahousehold:([-[:alnum:]]+)\z/
        household_id = $1.gsub('-', '').upcase.first(32)
```

app/models/concerns/client_search.rb
```diff
@@ -51,6 +51,14 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
...
+        client_ids = GrdaWarehouse::Hud::Enrollment.where(HouseholdID: household_id).joins(:client).pluck(sa[:id])
+        where = sa[:id].in(client_ids)
+      elsif /enrollment:[\d-]+/.match(text)
```
ttoomey: This is fine. It is an odd wrinkle that we use UUIDs in some places and DB ids in others

ttoomey (approved):
Looks good, thanks!

app/models/concerns/client_search.rb
gigxz: Yeah that's because Enrollment DB IDs are user-facing and Household IDs are user-facing. Household IDs are sometimes UUIDs but not always (if they are imported).

app/models/concerns/client_search.rb
gigxz: They are not necessarily UUIDs. Data imported from other system could have numeric Household IDs for example. I agree with removing dashes since we do that in the PersonalID search too, I'll add that change. & the start/end match. Thanks!

app/models/concerns/client_search.rb
gigxz: 👍  Added support for the space, why not

PR #4103, opened 2024-03-11
Rake task for generating CustomDataElementDefinitions

gigxz:

## Description

Add a rake task for generating/updating CustomDataElementDefinitions by introspecting on form definition items.

This should eventually be managed as part of the Form Editor tool, but for now we are running it manually.

## Type of change


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -443,4 +443,65 @@ def apply_conditionals(enrollment)
...
+        when 'BOOLEAN'
+          'boolean'
+        when 'DATE'
+          'date'
+        when 'INTEGER'
+          'integer'
+        when 'CURRENCY'
+          'float'
+        when 'DISPLAY', 'GROUP'
+          raise "unexpected: custom_field_key on non-question item. form: #{identifier}, link_id: #{item.link_id}, key: #{key}"
```
gigxz: I expect this to come up in a few places. I think we should fix those to use the appropriate type with `readonly: true` flag now that Martha has fixed it. 

Open to discussion about it, though, it feels wrong to me that data is stored/populated for a DISPLAY item.

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
```diff
@@ -28,7 +28,7 @@ class Hmis::Hud::CustomDataElementDefinition < Hmis::Hud::Base
   belongs_to :custom_service_type, class_name: 'Hmis::Hud::CustomServiceType', optional: true
   has_many :values, class_name: 'Hmis::Hud::CustomDataElement', inverse_of: :data_element_definition, foreign_key: :data_element_definition_id
 
-  validate :validate_service_type
+  # validate :validate_service_type
```
gigxz: removing this column entirely in a separate pr https://github.com/greenriver/hmis-warehouse/pull/4086
we don't use it, just skipping the validation for now

gigxz:
If reviewer(s) can try running it locally that would be much appreciated 🙂

drivers/hmis/lib/tasks/setup.rake
```diff
@@ -85,3 +85,39 @@ task seed_e2e: [:environment, 'log:info_to_stdout'] do
...
+task :generate_custom_data_elements, [:role] => [:environment, 'log:info_to_stdout'] do |_task, args|
+  role = args.role
+  custom_data_element_definitions = []
+  seen_keys = Set.new
+
+  definition_scope = Hmis::Form::Definition.non_static
+  definition_scope = definition_scope.with_role(role) if role.present?
+  puts "#{definition_scope.size} forms to process: #{definition_scope.map(&:identifier).join(', ')}"
+
+  definition_scope.order(:id).each do |definition|
```
gigxz, line 108: this should really only look at the latest version of each form identifier, but figured its not worth it now since this logic is going to move out of a rake task and into the form editor tooling

drivers/hmis/lib/tasks/setup.rake
```diff
@@ -85,3 +85,39 @@ task seed_e2e: [:environment, 'log:info_to_stdout'] do
...
+  existing_cdeds = custom_data_element_definitions.select(&:persisted?)
+
+  Hmis::Hud::CustomDataElementDefinition.transaction do
+    existing_cdeds.each(&:save!)
+    new_cdeds.each(&:save!)
+  end
+
+  puts "Saved #{existing_cdeds.size} EXISTING keys: #{existing_cdeds.size > 50 ? 'truncated' : existing_cdeds.map(&:key)}"
+  puts "Saved #{new_cdeds.size} NEW keys: #{new_cdeds.map(&:key)}"
+end
```
gigxz, line 140: Might be also nice to log "dangling" custom data elements that are not referenced by any forms

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -443,4 +443,65 @@ def apply_conditionals(enrollment)
...
+  # Form Editor admin tool. For now, it is called by a rake task manually.
+  def introspect_custom_data_elements
+    owner_type = if ASSESSMENT_FORM_ROLES.include?(role.to_sym)
+      Hmis::Hud::CustomAssessment
+    else
+      FORM_ROLE_CONFIG.dig(role.to_sym, :owner_class)
+    end
+    raise "unable to determine owner class for form role: #{role}" unless owner_type
+
+    data_source = GrdaWarehouse::DataSource.hmis.first
```
martha: Would this break if we had more than 1 data source?

martha:
This is great, thank you for making a rake task! I'm able to run it locally and get output that looks normal, after fixing a few fields that have a custom mapping despite not collecting data, as you mentioned in your comment. I'm then able to submit data on the custom assessments and it gets saved and everything looks good. 👍 

One thing I didn't expect was that when examining it in my local dev database, the table `development_openpath_warehouse.public.CustomDataElementDefinitions` is still empty. Is that expected? Where are these custom data element definitions actually getting saved?

gigxz:
That's odd @martha were you running the rake task or just the `introspect_custom_data_elements` fn? The rake task is the one that saves them.

edit: the table you mention is the one where they are saved

martha:
> That's odd @martha were you running the rake task or just the introspect_custom_data_elements fn? The rake task is the one that saves them.

Seems it's resolved now -- it must be that I was querying the db incorrectly before, sorry for my confusion! I was also running it with `rails` instead of `rake`, would that make any difference?

gigxz:
nope that should be the same!

martha (approved):

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -452,4 +452,67 @@ def apply_conditionals(enrollment)
...
+    hud_user_id = Hmis::Hud::User.system_user(data_source_id: data_source.id).UserID
+
+    cded_records = []
+    recur_traverse = lambda do |items|
+      items.each do |item|
```
ttoomey, line 477: This is fine. I think we did end up doing some redundant work, I have [similar but more limited functionality in the public touchpoints branch](https://github.com/greenriver/hmis-warehouse/pull/4037/files#diff-c313e61225428efbe27e27e1e616781a579b725d8f3fd9b78a2b43599292501aR45). We should probably reconcile these two implementations when these get merged

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -452,4 +452,67 @@ def apply_conditionals(enrollment)
...
+
+        # skip unless custom_field_key specified for this item
+        key = item.mapping&.custom_field_key
+        next unless key
+
+        # find CDED if it exists, or initialize a new one with defaults
+        cded = Hmis::Hud::CustomDataElementDefinition.where(key: key).first_or_initialize(
+          data_source_id: data_source.id,
+          UserID: hud_user_id,
+        )
```
ttoomey: This would be one query per form element. We could optimize that, probably worth it if it's going to be called in a request (on form save). I think something like this would work:
```ruby
    cded_scope = Hmis::Hud::CustomDataElementDefinition.where(owner_type: owner_type, data_source: data_source)
    cdeds_by_key = cded_scope.index_by(&:key)
    recur_traverse = lambda do |items|
      # ...
      cded = cdeds_by_key[key] || cded_scope.new(key: key,  UserID: hud_user_id)
```

drivers/hmis/lib/tasks/setup.rake
```diff
@@ -85,3 +85,56 @@ task seed_e2e: [:environment, 'log:info_to_stdout'] do
...
+  definition_scope = Hmis::Form::Definition.non_static
+  definition_scope = definition_scope.with_role(role) if role != 'ALL'
+  puts "#{definition_scope.size} forms to process: #{definition_scope.map(&:identifier).join(', ')}\n\n"
+
+  definition_scope.order(:id).each do |definition|
+    # Skip any CDEDs that have already been processed
+    new_cdeds = definition.introspect_custom_data_elements.reject { |cded| seen_keys.include?(cded.key) }
+
+    # Add to seen_keys
+    new_cdeds.map(&:key).each { |key| seen_keys.add(key) }
```
ttoomey: The keys by themselves are not necessarily unique right? I think the constraint is [owner_type, key]. So if two definitions had keys with the same name but different owner types, it seems like cdeds would be excluded.

Maybe something like this (not tested):
```ruby
  definition_scope.order(:id).each do |definition|
    definition.introspect_custom_data_elements.each do |cded|
      key = [cded.owner_type, cded.key]
      next if seen_keys.include?(key)
      seen.add(key)
      custom_data_element_definitions << cded
    end
  end
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -452,4 +452,67 @@ def apply_conditionals(enrollment)
   def project_and_enrollment_match(...)
     instances.map { |i| i.project_and_enrollment_match(...) }.compact.min_by(&:rank)
   end
+
+  # Find and/or initialize CustomDataElements that are collected by this form. Eventually this should be done as part of the
+  # Form Editor admin tool. For now, it is called by a rake task manually.
+  def introspect_custom_data_elements
```
ttoomey: We return cdeds, not cdes right? I think it would be better if the method name reflected that
```suggestion
  # Find and/or initialize CustomDataElementDefinitions that are collected by this form. Eventually this should be done as part of the
  # Form Editor admin tool. For now, it is called by a rake task manually.
  def introspect_custom_data_element_definitions
```

ttoomey (commented):
Thanks for working on this!

I think there's a small bug around unique keys with cdeds that should probably get addressed.

I also am uncertain what happens when two forms have conflicting definitions for the same data element. Right now it looks like the form that was created first (lower id) would win. I think we should probably detect this type of conflict and raise an error or something

The rake task ran for me locally (I made it skip the external_touch point definitions I have locally)

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: Yes, added comment

gigxz:
Alright I think this is ready to merge in and @martha can use it in the public forms work

PR #4110, opened 2024-03-12
Add permission caching that fixes bug with project deletion

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/187201007

**Description of bug**: When deleting a Project from the HMIS ui, the deletion modal wouldn't close on success. This was because the mutation wasn't returning the project's ID back with the successful response, which the `DeleteMutationButton` checks for, [here](https://github.com/greenriver/hmis-frontend/blob/5a89dcddcac5be5b9fabc9900994418acc66c3c2/src/modules/dataFetching/components/DeleteMutationButton.tsx#L61).

**What was the issue**: After the project was deleted, the project's `self.authorized?` [method](https://github.com/greenriver/hmis-warehouse/blob/a29e2f125380909658091f105babe4f80dcd6fd3/drivers/hmis/app/graphql/types/hmis_schema/project.rb#L29-L33) would return false.

Poking around, I thought maybe this issue would also affect the Enrollment and Client schema types, because both have similar `self.authorized?` checks on the schema object themselves. However, they aren't affected by the bug, it turns out for two different reasons:
- Enrollment: the `self.authorized` check also checks if the user has permission to view that enrollment's project ([here](https://github.com/greenriver/hmis-warehouse/blob/a29e2f125380909658091f105babe4f80dcd6fd3/drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb#L42-L43)). Since the project still exists, even after the enrollment is deleted, this returns `true`
- Client: Someone already discovered this bug on the client deletion mutation, and added [this code](https://github.com/greenriver/hmis-warehouse/blob/a29e2f125380909658091f105babe4f80dcd6fd3/drivers/hmis/app/graphql/mutations/delete_client.rb#L23-L25) to mitigate.

**What I've changed here**: Added similar code to the delete project mutation. It caches the result of a permission call before the project is deleted, so the project can still be returned to the frontend in the response to the deletion call.

**Open questions**:
- This caching fix seems a bit rickety to me, should we investigate/discuss other solutions? Even if we keep this approach, should it be moved somewhere more central so it can be used in common?
- The ticket mentions, "We could modify the onSuccess so it doesn't care if the object is resolved. It doesn't really matter, we just need to ensure there are no errors, which it already checks." Having discovered this bug, should I still make that frontend change? Is there any concern that it could end up swallowing issues that we actually want to know about?

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/mutations/delete_project.rb
+++ b/drivers/hmis/app/graphql/mutations/delete_project.rb
@@ -12,6 +12,11 @@ module Mutations
 
     def resolve(id:)
       record = Hmis::Hud::Project.viewable_by(current_user).find_by(id: id)
+
+      # While this is redundant with the viewable_by() scope above, this check caches the authorization result so that
+      # the project object-level authorization check will succeed even after the project has been deleted
+      access_denied! unless current_permission?(permission: :can_view_project, entity: record)
+
       default_delete_record(record: record, field_name: :project, permissions: :can_delete_project)
     end
   end
--- a/drivers/hmis/spec/requests/hmis/delete_project_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/delete_project_spec.rb
@@ -1,0 +1,59 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+require 'rails_helper'
+require_relative 'login_and_permissions'
+require_relative '../../support/hmis_base_setup'
+
+RSpec.describe Hmis::GraphqlController, type: :request do
+  before(:all) do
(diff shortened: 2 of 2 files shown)


gigxz (approved):
Good find @martha and thanks for the explanation.

I think lets merge and mark this bug as fixed. Would you mind creating a follow-up ticket in the icebox for reviewing how we resolve deleted records? I think it would be reasonable to change all these `DeleteX` mutations to just resolve `{success: true}` rather than resolving the record. I can't think of any instances where we actually need that record back. I agree we should discuss as a group at a later time. Thanks!

[edit: and no need for changes to the frontend at this time]

martha:
Yes sounds good, created https://www.pivotaltracker.com/story/show/187229708! Thanks for merging 🚀

PR #4114, opened 2024-03-12
Bugfix in MigrateAssessmentsJob for generating empty intakes; performance improvement

gigxz:

## Description

* Fix a bug in `build_synthetic_intake_assessment` that would cause it to fail when `UserID` was unresolvable (aka ` enrollment.user` is nil) or when `UserID` was nil.
* Use AR Import to import CustomAssessment records, instead of calling `save!` on each individually.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
+++ b/drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
@@ -198,7 +198,6 @@ module Hmis
 
       skipped_invalid_assessments = 0
       skipped_exit_assessments = 0
-      generate_empty_intake_count = 0
       skipped_intake_enrollment_ids = []
 
       # For each grouping of Enrollment+InformationDate+DataCollectionStage,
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -403,11 +403,11 @@ def unit_occupied_on(date = Date.current)
 
   def build_synthetic_intake_assessment
     assessment = Hmis::Hud::CustomAssessment.new(
-      enrollment: self,
-      client: client,
-      data_source: data_source,
+      enrollment_id: enrollment_id,
+      personal_id: personal_id,
       assessment_date: entry_date,
-      user: user, # same user that's on the enrollment
```
gigxz, line 391: what happened is: if `Enrollment.UserID` is not real, then `user: user` would set `data_source_id: nil`, because of the way the user is related via data source(see `hmis_enrollment_relation`). sneaky!

to avoid all that, I updated each of these to just set the ids. It also helps us avoid database lookups for the whole client and data source records if they're not already on hand.

drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
```diff
@@ -62,18 +64,22 @@ def perform(data_source_id:, project_ids: nil, clobber: false, delete_dangling_r
           each(&:really_destroy!)
       end
 
+      total = full_enrollment_scope.count
+      Rails.logger.info "#{total} Enrollments to process"
+
       # Note: joining with project drops WIP enrollments. That should be fine since they won't have assessment records yet.
```
martha: Maybe would be good to move this comment down to line 110 where the join is now happening?

drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
```diff
@@ -231,18 +243,26 @@ def build_assessments(enrollment_scope:, data_collection_stages:, unique_by_info
       # so the UI won't prompt the user to complete intake assessments on these enrollments.
       if generate_empty_intakes && data_collection_stages.include?(1)
         Hmis::Hud::CustomAssessment.transaction do
```
martha: Is this transaction still useful here, since the save call is moved out of this block?

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Wow, sneaky indeed. Thank you!

martha (approved):

drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
gigxz: nope good catch, I'll remove

gigxz: oh, that made me notice another bug. the `enrollments_missing_intakes` check assumes that all the generated intakes have been saved so far, which they haven't at this point. Looks like we didn't have a test case to catch that! I will fix and add one.

gigxz: alright, fixed in https://github.com/greenriver/hmis-warehouse/pull/4114/commits/c475fbe46512dcb41208daa4cd58b3dcc14b3ac5
I updated the test to check that we don't create double intake assessments

eanders (approved):

PR #4120, opened 2024-03-13
Use the real field type from CDE

martha:

## Description

Previously, for external form submissions we saved all values as strings. However, we want to use DynamicDisplay in the frontend to display these components. So, this PR makes changes to use the correct type (date, boolean, etc.) so that DynamicDisplay can display everybody correctly. 

Prior discussion that led to this change:
https://github.com/greenriver/hmis-frontend/pull/691#pullrequestreview-1926926937
https://github.com/greenriver/hmis-warehouse/pull/4037#discussion_r1513529474

[PT issue](https://www.pivotaltracker.com/story/show/187040557)

I'll add a few open questions inline...

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
```diff
@@ -48,25 +48,23 @@ def update_cdeds(definition)
...
-        }
-      end
-      if attrs
-        cded = cded_by_key[cded_key]
-        cded ? cded.update!(attrs) : missing.push(attrs)
-      end
+      next unless ['STRING', 'CHOICE', 'BOOLEAN', 'INTEGER', 'TEXT', 'DATE'].include?((node['type']))
+
+      field_type = node['type'].downcase || 'string'
+      field_type = 'string' if field_type == 'choice'
```
martha: This special-casing for choice types is messy. And, it's connected to [this](https://github.com/greenriver/hmis-frontend/pull/691/files#diff-736ca0dfcf22fbe66ce15749daa3f61b9b3dadfef7764d7fd3450f73009d4ed9R79-R88) messy special-casing in the frontend code.

Should we spend more time to find a neater way? For example, add a `value_choice` column to the CDED table?

Or should we just add some todos in the code for now?

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -38,24 +38,28 @@ def self.from_raw_data(raw_data, object_key:, last_modified:, form_definition:)
       submission
     end
 
+    def cde_helper(data_source_id:, system_user_id:)
+      @cde_helper ||= HmisExternalApis::TcHmis::Importers::Loaders::CustomDataElementHelper.new(
```
martha: Does this look okay? I think the effect of this is to just use the `data_source_id` and `system_user_id` from the first CDED in the iteration. Hypothetically it could be different for every CDED but in practice it will not be different, right?

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_data_element_helper.rb
martha, line 1: Is it okay to share this helper across `BaseLoader` and now `FormSubmission`? Should it be moved somewhere else?

martha (commented):
Also, while working on this I discovered a couple of interesting bugs:
- ~If the user submits but there's an error while saving the CDEs, they see an error message ('check your internet connection'), but the form submission is saved even though the CDEs failed to be saved.~ addressed [here](https://github.com/greenriver/hmis-warehouse/pull/4120/files#diff-34c13c8335974559751928b36cbca0a3363274e0103ed5062da19bf038a918a7R36-R39) in this PR
- Weird bug with DOB collection in the form. Steps to reproduce:
    - Fill out the form (required fields only is fine)
    - Hit 'Submit form'
    - Hit 'Go back,' do not change anything (DOB field is still filled out)
    - Hit 'Submit form' again and then 'Yes, Submit Now'
    - sometimes, DOB does not get saved successfully in the DB. I'd be grateful @ttoomey if you could see if you can reproduce this as well, because it seems like it doesn't always happen, so maybe I'm missing something or doing something wrong!

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
ttoomey: So we shouldn't use that `CustomDataElementHelper` class here. I mentioned it only as an example but that was confusing, sorry about that! The reason it shouldn't be used is the loader code is only temporary; it will be removed after the current migration is complete.

I recommend that you extract the relevant code out of that helper. I think the only piece that's needed here is the behavior in the cde_value_fields() method that sets all the value_[type] attributes. That could probably be rolled into this class

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -33,29 +33,35 @@ def self.from_raw_data(raw_data, object_key:, last_modified:, form_definition:)
         definition_id: form_definition.id,
         raw_data: raw_data,
       }
-      submission.save!
-      submission.process_custom_data_elements!(form_definition: form_definition)
+      FormSubmission.transaction do
```
ttoomey: Thanks for adding some transactional safety! However it looks like we already wrap this in a tx where it's called from `ConsumeExternalFormSubmissionsJob` so we probably don't need both. Generally I think it's better to let the caller handle the tx if possible

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_data_element_helper.rb
```diff
@@ -6,17 +6,17 @@
 
 module HmisExternalApis::TcHmis::Importers::Loaders
   class CustomDataElementHelper
-    attr_accessor :data_source, :system_user, :today
+    attr_accessor :data_source_id, :system_user_id, :today
 
-    def initialize(data_source:, system_user:, today:)
-      self.data_source = data_source
-      self.system_user = system_user
+    def initialize(data_source_id:, system_user_id:, today:)
```
ttoomey: So this loader is used by our migration process. I'd prefer not to make changes to this code while we are in middle of data migration unless there a bug or some issue this addresses. If this change isn't necessary, can you back it out (and the change to BaseLoader)? Thanks!

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
ttoomey: I think we should probably pause on this; it looks like Gig is [working on the same thing](https://github.com/greenriver/hmis-warehouse/pull/4103)

ttoomey (changes requested):
Thanks Martha! I left a few comments where we might need changes. But I suggest we pause on this until https://github.com/greenriver/hmis-warehouse/pull/4103 is merged so we can leverage Gig's work for PublishExternalForms#update_cdeds

martha:
Thanks @ttoomey all your suggested changes make sense to me, but as you said, I will hold off on starting to implement them until after Gig's work is merged.

By any chance, were you able to replicate the bug I mentioned above with DOB collection in the public form? Do you think we should create a ticket for that?

ttoomey:
> By any chance, were you able to replicate the bug I mentioned above with DOB collection in the public form? Do you think we should create a ticket for that?

Oh sorry I missed that, thanks for flagging it!

I just tested it in chrome and I couldn't reproduce the issue but I could totally imagine there's something going wrong. Changing any of those date fields should update a hidden field in the form. Perhaps the change event is not firing for you. It could be there's some interaction with the browser's autocomplete or a browser extension. Maybe the behavior should be moved into the form's submit handler rather than on-change to be more reliable.

Regardless the date input could benefit from some more attention. One thing that's outstanding is that there's no validation on the input, the user  can enter an invalid date. I think making an issue in PT to capture both the buggy behavior and the need for validation would be great, thank you!

Relevant code:
drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_date.haml
drivers/hmis_external_apis/app/assets/javascripts/external_forms.js

martha:
Sounds good! Created https://www.pivotaltracker.com/story/show/187250024

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_data_element_helper.rb
```diff
@@ -51,26 +51,7 @@ def new_cde_record(value:, owner_type:, owner_id:, definition_key:, date_created
...
-    ].map { |v| [v, "value_#{v}"] }
-
-    # note, we need to set all fields- bulk insert becomes unhappy if the columns are not uniform
-    def cde_value_fields(definition, value)
-      result = {}
-      VALUE_FIELDS.map do |field_type, field_name|
-        result[field_name] = field_type == definition.field_type ? value : nil
-      end
-      result
+      }.merge(HmisExternalApis::ExternalForms::FormSubmission.cde_value_fields(definition, value))
```
martha: @ttoomey I know this is code used by the loader and we should avoid changing it, but I think your suggestion in a previous comment was to extract this into the FormSubmission. Let me know if you think the change I have made here looks ok or should be reverted (resulting in duplicated code, but will be removed when we remove all the loader code).

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
```diff
@@ -43,32 +43,9 @@ def data_source_id
...
-          data_source_id: data_source_id,
-          form_definition_identifier: definition.identifier,
-        }
-      end
-      if attrs
-        cded = cded_by_key[cded_key]
-        cded ? cded.update!(attrs) : missing.push(attrs)
-      end
+    Hmis::Hud::CustomDataElementDefinition.transaction do
+      definition.introspect_custom_data_element_definitions(set_definition_identifier: true).each(&:save!)
```
martha, line 47: Do we need to keep track of `seen_keys` here the way that we do in `setup.rake`? I was thinking it's not needed because this is only 1 form's worth of CDEDs, so we won't see duplicates across forms, but let me know if that is incorrect -- I know we may add more public forms later. Would it be worthwhile to extract that logic into a reusable place? Maybe it would live as another method on `Form::Definition` that calls `introspect_custom_data_element_definitions` internally?

drivers/hmis/lib/form_data/tarrant_county/external_forms/prevention_screening.json
```diff
@@ -10,48 +10,72 @@
           "link_id": "first_name",
           "required": true,
           "type": "STRING",
-          "text": "First Name"
+          "text": "First Name",
```
martha, line 13: I added `custom_field_key`s here because `introspect_custom_data_element_definitions` skips unless it exists https://github.com/greenriver/hmis-warehouse/blob/release-108/drivers/hmis/app/models/hmis/form/definition.rb#L483

Let me know if it would be preferable to modify that logic instead? Or to make these custom_field_keys more descriptive?

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -22,8 +22,13 @@ class HmisSchema::ExternalFormSubmission < Types::BaseObject
     custom_data_elements_field
 
     def custom_data_elements
-      cdeds = load_ar_association(definition, :custom_data_element_definitions, scope: Hmis::Hud::CustomDataElementDefinition.order(:key))
-      resolve_custom_data_elements(object, definition_scope: cdeds)
+      load_ar_association(
```
martha: I think this isn't correct and we want to still be using `resolve_custom_data_elements` here, but I was having some trouble getting that to work, since `resolve_custom_data_elements` no longer accepts a `definition_scope` and expects the `record` passed in to have a `custom_data_elements` association, which this form submission object (I think) doesn't have. I will keep working on this to fix

martha (commented):
this is still messy and I'm uncertain about much of it, also there are still bugs in the UI for reviewing external forms that need to be updated in accordance with Gig's recent changes to CDEDs; however I wanted to push what I have so that @ttoomey @gigxz can let me know if I am on the right track or going off in a wrongheaded direction

thank you!

drivers/hmis/app/graphql/types/hmis_schema/custom_data_element.rb
```diff
@@ -22,5 +22,11 @@ class HmisSchema::CustomDataElement < Types::BaseObject
     def activity_log_object_identity
       object.id
     end
+
```
martha: @gigxz let me know if re-adding this to get the tests to pass is a mistake; I saw that it was removed in your [recent PR](https://github.com/greenriver/hmis-warehouse/pull/4086/files#diff-993bc2a410d5610001ecf3c82471407961d7dbf7339b6163c31565d2af1aabd3L49-L54).

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
martha: ~Ok, maybe this is a bit better, but please see the change in [the concern](https://github.com/greenriver/hmis-warehouse/pull/4120/files#diff-5071bc8277a9fb37c81d88153d86fe66df65bc4e9cda29251bd3f8a7bb94441cR42-R46) and let me know if I am on the right track~ understood now that this whole thing should be deleted in favor of using the concern!

drivers/hmis/app/graphql/types/hmis_schema/custom_data_element.rb
martha: I'm seeing that this can be removed if we also remove [line 16 of](https://github.com/greenriver/hmis-warehouse/pull/4120/files#diff-993bc2a410d5610001ecf3c82471407961d7dbf7339b6163c31565d2af1aabd3L16) this same file, maybe this being left behind was a bug, but please let me know otherwise

martha: however, removing this also has bigger ramifications for the schema and frontend, so I'm truly not sure what to do :(

martha: hm, never mind - I see the problem and think it is resolved with [this removal](https://github.com/greenriver/hmis-warehouse/pull/4120/files#diff-070654ce76a94ac027302b87240008ae0e8dbd0de27d3626d38a9bb71af07066L24-L28)

drivers/hmis_external_apis/app/jobs/hmis_external_apis/publish_external_forms_job.rb
gigxz: Yeah I don't think we need it here, since we are only processing one FormDefinition.

drivers/hmis/app/models/hmis/hud/concerns/has_custom_data_elements.rb
```diff
@@ -8,7 +8,7 @@ module Hmis::Hud::Concerns::HasCustomDataElements
   extend ActiveSupport::Concern
 
   included do
-    has_many :custom_data_elements, as: :owner, dependent: :destroy
+    has_many :custom_data_elements, as: :owner, dependent: :destroy, class_name: 'Hmis::Hud::CustomDataElement'
     # All the CDEDs that have values for this record. Note it will retun non-distinct scope for any types that have multiple values.
     has_many :custom_data_element_definitions, through: :custom_data_elements, source: :data_element_definition
```
gigxz: You may need the class_name here too?

drivers/hmis/lib/form_data/tarrant_county/external_forms/prevention_screening.json
gigxz: I do think they should be descriptive, something like `tchc_prevention_first_name` is the convention we use elsewhere

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -532,7 +536,7 @@ def introspect_custom_data_element_definitions(set_definition_identifier: false)
       cded.label = ActionView::Base.full_sanitizer.sanitize(item.readonly_text || item.brief_text || item.text || key.humanize)
 
       # If specified, set the definition identifier to specify that this CustomDataElementDefinition is ONLY collected by this form type.
-      cded.definition_identifier = identifier if set_definition_identifier
```
martha, line 535: @gigxz , Theron and I realized we were confused about this - was it here proactively as a placeholder, or is `definition_identifier` actually something different?

gigxz: no I was just trying to be helpful! my bad

drivers/hmis/lib/form_data/tarrant_county/external_forms/prevention_screening.json
```diff
@@ -167,7 +212,10 @@
               "code": "no",
               "label": "No"
             }
-          ]
+          ],
+          "mapping": {
+            "custom_field_key": "tchc_prevention_screening_has_been_literally_homeless"
+          }
```
ttoomey, line 216: looks like there's one input that still needs a key on line 231:
```json
          "mapping": {
            "custom_field_key": "tchc_prevention_screening_how_many_months_is_your_lease"
          },
```

ttoomey (approved):
Looks great, thanks @martha!  🙌

PR #4128, opened 2024-03-15
Add mutations for service type/category creation, deletion, and renaming

martha:

## Description

PT issue: [187211692](https://www.pivotaltracker.com/story/show/187211692)

This PR adds creation, deletion, and renaming mutations for `ServiceCategories` and `ServiceTypes`. Frontend PR to come.

~One weird thing that I noticed is that Graphql seems to be eliding `Type` on the end of `ServiceType` when it generates input types and payloads, which we don't want, because a `Service` and a `ServiceType` are distinct concepts.
I know this isn't going to be the real solution, but I worked around it here by naming the mutations `CreateServiceTypeType`, with `Type` duplicated. That way the generated types end up with one `Type` instead of none at all. (Removing it even causes a conflict when generating the schema, because a `DeleteService` mutation already exists.) 
TODO: investigate a way to prevent this undesired behavior from graphql, or find a tidier workaround.~

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/mutations/create_service_category.rb
```diff
@@ -0,0 +1,32 @@
...
+
+module Mutations
+  class CreateServiceCategory < BaseMutation
+    argument :name, String, required: true
+
+    field :service_category, Types::HmisSchema::ServiceCategory, null: true
+    field :errors, [Types::HmisSchema::ValidationError], null: false, resolver: Resolvers::ValidationErrors
+
+    def resolve(name:)
+      raise 'access denied' unless current_user.can_configure_data_collection?
```
gigxz: We've been switching to this gradually, so that we can standardize access errors eventually 
```suggestion
      access_denied! unless current_user.can_configure_data_collection?
```

drivers/hmis/app/graphql/mutations/create_service_category.rb
```diff
@@ -0,0 +1,32 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Mutations
+  class CreateServiceCategory < BaseMutation
```
gigxz:
```suggestion
  class CreateServiceCategory < CleanBaseMutation
```

CleanBaseMutation was added when we realized we don't need the unnecessary Relay stuff that BaseMutation has. I recommend that going forward, we may clean up  old BaseMutation later

drivers/hmis/app/graphql/mutations/delete_service_category.rb
```diff
@@ -0,0 +1,27 @@
...
+  class DeleteServiceCategory < BaseMutation
+    argument :id, ID, required: true
+
+    field :service_category, Types::HmisSchema::ServiceCategory, null: true
+
+    def resolve(id:)
+      raise 'access denied' unless current_user.can_configure_data_collection?
+
+      service_category = Hmis::Hud::CustomServiceCategory.find_by(id: id)
+      raise HmisErrors::ApiError, 'Invalid service category ID' unless service_category
```
gigxz: This is fine, I also think it's okay to just do `Hmis::Hud::CustomServiceCategory.find(id)` which will raise its own exception if not found

drivers/hmis/app/graphql/mutations/delete_service_type_type.rb
```diff
@@ -0,0 +1,27 @@
...
+    field :service_type, Types::HmisSchema::ServiceType, null: true
+
+    def resolve(id:)
+      raise 'access denied' unless current_user.can_configure_data_collection?
+
+      service_type = Hmis::Hud::CustomServiceType.find_by(id: id)
+      raise HmisErrors::ApiError, 'Invalid service type ID' unless service_type
+
+      is_empty = service_type.custom_services.count == 0
+      raise HmisErrors::ApiError, 'Cannot delete a service type that has services' unless is_empty
```
gigxz: "exists?" does a faster query than `count`
```suggestion
      raise 'Cannot delete a service type that has services' if service_type.custom_service.exists?
```

The `HmisErrors::ApiError` doesn't really get us anything here, the user will see the "a server error occurred" message either way. Typically this sort of thing would warrant a user-facing ValidationError returned in the `{errors: ` object, but I think its correct to skip that nicety for now, since this is the first pass config tool. Eventually it should be fancier and tell you how many projects use this service, etc.

drivers/hmis/app/graphql/mutations/rename_service_type_type.rb
```diff
@@ -0,0 +1,25 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Mutations
+  class RenameServiceTypeType < BaseMutation
```
gigxz: I think you can add the `graphql_name` property here to specify `RenameServiceType` , and rename the class and file name. Looks like Mutation has it: https://graphql-ruby.org/api-doc/1.8.7/GraphQL/Schema/Mutation.html

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -1598,6 +1598,52 @@ type CreateScanCardCodePayload {
...
+  """
+  clientMutationId: String
+  errors: [ValidationError!]!
+  serviceCategory: ServiceCategory
+}
+
+"""
+Autogenerated input type of CreateServiceType
+"""
+input CreateServiceTypeInput {
```
gigxz: Switching to CleanBaseMutation should get rid of these noisy input wrapper types I think

drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb
```diff
@@ -0,0 +1,21 @@
...
+###
+
+module Types
+  class HmisSchema::ServiceTypeInput < BaseInputObject
+    description 'Create service type input'
+
+    argument :name, String, required: true
+    argument :custom_service_category_id, ID, required: true
+    argument :hud_record_type, ID, required: false
+    argument :hud_type_provided, ID, required: false
```
gigxz: Oh I forgot to mention these, we don't need them. The HUD service types are fixed, so users will only be managing custom services.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -89,6 +89,8 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
     :PROJECT_CONFIG,
     :CLIENT_ALERT,
     :FORM_DEFINITION,
+    :SERVICE_TYPE,
+    :SERVICE_CATEGORY,
```
gigxz: Are you using Static Forms for these? I don't see JSON files added. IMO these are simple enough (just a name?) that we can just as well implement them custom on the frontend. That lets us be a bit more flexible with the UI. (Not tied to StaticForm).

drivers/hmis/app/graphql/mutations/delete_service_category.rb
martha: Thanks, this is much tidier!

drivers/hmis/app/graphql/mutations/rename_service_type_type.rb
martha: Perfect!! Thank you

drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb
martha: Just confirming, we do still need `supports_bulk_assignment`, correct?

drivers/hmis/app/models/hmis/form/definition.rb
martha: Ok good plan, I will implement the forms directly in the frontend.

drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb
gigxz: yep!

drivers/hmis/app/graphql/mutations/base_mutation.rb
```diff
@@ -5,6 +5,8 @@
 ###
 
 module Mutations
+  # Generally, use CleanBaseMutation instead of this since we don't need Relay.
+  # We may clean up the old BaseMutation later
```
gigxz, line 9: Thanks

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -10057,6 +10111,15 @@ input ServiceTypeFilterOptions {
   searchTerm: String
 }
 
+"""
+Create service type input
+"""
+input ServiceTypeInput {
+  customServiceCategoryId: ID!
```
gigxz: We have been keeping the `custom` out of the graphql schema to date. So to be in keeping with that, this should probably be just `serviceCategoryId`.

gigxz (approved):
Looks great! I small comment about naming for consistency, feel free to merge after that

gigxz:
Actually, before merging, can you retarget this to release-109 @martha on the frontend and backend? thanks!

PR #4131, opened 2024-03-18
Hmis recent items and clients perf

ttoomey:

## Description

API performance on my local db with pg13 and ~400,000 clients and enrollments is quite poor. This PR helps some of the worst performing queries I see.

The change here is to the recent items and client authorization methods which have been adjusted to avoid plucking all visible client ids. Also added a limit to recent items, it should return a max of 10 items now.

### testing
* Recent items query should return in less than 3s (previously 6)
* Client pages should return more quickly

## Type of change
- [x] Performance fix

### DIFF
--- a/app/models/concerns/has_recent_items.rb
+++ b/app/models/concerns/has_recent_items.rb
@@ -13,16 +13,26 @@ module HasRecentItems
       self.class.recent_item_types
     end
 
-    def recent_items
-      viewable_recent_item_links.order(updated_at: :desc).preload(:item).map(&:item).compact
+    def recent_items(limit: 10)
+      items = recent_item_types.values.flat_map do |klass|
+        viewable_recent_items(klass, limit: limit)
+      end
+      items.sort_by { |item, date| [date, item.id] }.map(&:first).reverse!.take(limit)
     end
 
-    def viewable_recent_item_links
-      ids = recent_item_types.values.map do |item_class|
-        recent_item_links.where(item_type: item_class.name, item_id: item_class.viewable_by(self).pluck(:id)).pluck(:id)
-      end.flatten
+    # @returns [item, last_updated]
+    protected def viewable_recent_items(item_klass, limit:)
+      links_by_item = recent_item_links.
+        where(item_type: item_klass.sti_name).
+        order(updated_at: :desc).
+        limit(limit).
+        to_a.index_by(&:item_id)
 
-      recent_item_links.where(id: ids)
+      # restrict by visibility
+      item_klass.viewable_by(self).where(id: links_by_item.keys).map do |item|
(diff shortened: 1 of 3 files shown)


gigxz (approved):
looks good!

app/models/concerns/has_recent_items.rb
```diff
@@ -13,16 +13,26 @@ def recent_item_types
...
+        order(updated_at: :desc).
+        limit(limit).
+        to_a.index_by(&:item_id)
 
-      recent_item_links.where(id: ids)
+      # restrict by visibility
+      item_klass.viewable_by(self).where(id: links_by_item.keys).map do |item|
+        link = links_by_item.fetch(item.id)
+        [item, link.updated_at]
+      end
```
eanders, line 35: If you can't see any of the most recent 10, aren't you going to get nothing back from this?  (The limit is imposed on finding the most recent items before the view ability is enforced.)

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -115,11 +115,15 @@ class Hmis::Hud::Client < Hmis::Hud::Base
...
-    where(id: unenrolled_ids + enrolled_ids + wip_ids)
+    scopes = []
+    scopes << unenrolled.joins(:data_source).merge(GrdaWarehouse::DataSource.hmis(user)) if user.permissions?(*permissions, **kwargs)
+    scopes += [
+      joins(:projects).where(p_t[:id].in(pids)),
+      joins(:wip).where(wip_t[:project_id].in(pids)),
+    ]
+    sql = scopes.map { |s| s.select(c_t[:id].to_sql).to_sql }.join(' UNION ALL ')
+
+    where("#{c_t[:id].to_sql} IN (#{sql})")
```
eanders: can this not be written?
```suggestion
    where(c_t[:id].in(Arel.sql(sql)))
```

app/models/concerns/has_recent_items.rb
ttoomey: It's a good point. My thought was to avoid loading the ids for every item the user has ever viewed just to keep the top 10. I think it is probably okay to keep the limit where it is with the following assumptions:
a) it's pretty rare for someone to lose access to a recent client or project
b) the impact of not seeing those recent clients/projects is minor since this just a convenience and navigation shortcut

If that's not true I can move the limit to viewable items. It probably won't make much of a difference in most situations.

gigxz: I think it's fine.. I agree its a convenience, so we should focus on it being fast over being complete in an edge case

gigxz:
Can we merge this into release-109 or are you planning to do additional review @eanders ?

PR #4140, opened 2024-03-19
Health Database Optimizations

yakloinsteak:
## Description

https://github.com/greenriver/devops-requests/issues/739

I tested these changes against the problematic health database, but somebody with more knowledge of the application and business logic should also test this before merging. 

## Type of change
- [x] Query speed-ups

### DIFF
--- a/app/models/health/release_form.rb
+++ b/app/models/health/release_form.rb
@@ -37,7 +37,7 @@ module Health
     scope :recent, -> { order(signature_on: :desc).limit(1) }
     scope :reviewed, -> { where.not(reviewed_by_id: nil) }
     scope :valid, -> do
-      parent_ids = Health::ReleaseFormFile.where.not(parent_id: nil).pluck(:parent_id)
+      parent_ids = Health::ReleaseFormFile.where.not(parent_id: nil).select(:parent_id)
       where.not(file_location: [nil, '']).
         or(where(id: parent_ids))
     end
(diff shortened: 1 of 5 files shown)


awisema (approved):
Thanks Todd! These look reasonable to me.

yakloinsteak:
@eanders Can you take a look and merge it if you're okay with it as well?

app/models/health/release_form.rb
```diff
@@ -37,7 +37,7 @@ class ReleaseForm < HealthBase
     scope :recent, -> { order(signature_on: :desc).limit(1) }
     scope :reviewed, -> { where.not(reviewed_by_id: nil) }
     scope :valid, -> do
-      parent_ids = Health::ReleaseFormFile.where.not(parent_id: nil).pluck(:parent_id)
+      parent_ids = Health::ReleaseFormFile.where.not(parent_id: nil).select(:parent_id)
```
eanders, line 40: Pretty sure this is safe, but @awisema can you confirm these parent IDs are going to be used in the same database?  With so many cross-database situations, just want to double check a `select` is ok here.

eanders (approved):
Looks good if we confirm the select is in the same DB.

yakloinsteak:
I ran this code in production and it worked.

app/models/health/release_form.rb
awisema: I have confirmed that the parent of a `Health::ReleaseFormFile` is the `Health::ReleaseForm` it is uploaded against -- so, there should be no x-db issue here.

eanders: Thanks

PR #4141, opened 2024-03-19
Gig/hmis acl style updates

gigxz:


app/models/menu/menu.rb
```diff
@@ -502,59 +502,66 @@ def hmis_admin_menu
...
+      access_menu.add_child(
         Menu::Item.new(
           user: user,
           visible: ->(_user) { context.hmis_admin_visible? },
           path: hmis_admin_access_overviews_path,
           title: Translation.translate('Overview'),
         ),
       )
-      menu.add_child(
+      access_menu.add_child(
```
gigxz, line 523: I can back this out, it should match the warehouse. This is an order that makes sense to me, but that might just be me!

PR #4150, opened 2024-03-20
Add project services spec

gigxz:

## Description

* Add Project Services spec to test performance and N+1. The query tested is the query that the frontend performs on the Project Service page, once https://github.com/greenriver/hmis-frontend/pull/701 is merged.
* Deterministic sorting when sorting services by date provided

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/spec/requests/hmis/project_services_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/project_services_spec.rb
@@ -1,0 +1,135 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+require 'rails_helper'
+require_relative 'login_and_permissions'
+require_relative '../../support/hmis_base_setup'
+
+RSpec.describe Hmis::GraphqlController, type: :request do
+  include_context 'hmis base setup'
+  include_context 'hmis service setup'
+  let!(:access_control) { create_access_control(hmis_user, p1) }
+
+  describe 'project services query' do
+    before(:each) do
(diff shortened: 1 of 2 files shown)


drivers/hmis/spec/requests/hmis/project_services_spec.rb
```diff
@@ -0,0 +1,135 @@
...
+      before(:each) do
+        # 30 enrollments, each with 10 services (5 HUD, 5 Custom) (300 services total)
+        30.times do |index|
+          # create WIP or non-WIP Enrollment
+          enrollment_factory = index.even? ? :hmis_hud_enrollment : :hmis_hud_wip_enrollment
+          en = create(enrollment_factory, data_source: ds1, project: p1)
+
+          # create both HUD and Custom services
+          5.times { create(:hmis_hud_service, date_provided: Date.yesterday, data_source: ds1, enrollment: en, client: en.client) }
+          5.times { create(:hmis_custom_service, date_provided: Date.yesterday, data_source: ds1, enrollment: en, client: en.client) }
```
ttoomey, line 107: This is fine as it is. In the future I suggest preferring memoized variables for date/time in tests. This reduces the chance that tests will fail if the current date changes during the run. for example:

```ruby
let(:yesterday) { Date.yesterday }
```

ttoomey (approved):
Thanks for adding tests!!

drivers/hmis/spec/requests/hmis/project_services_spec.rb
gigxz: I'll keep that in mind! In this case it shouldn't fail if the date is different. I just set them to be the same so that the initial 50-record page has a variety of clients and enrollments.

PR #4152, opened 2024-03-21
Use HMIS_LOGO if provided

martha:

## Description

This PR updates `Hmis::AppSettingsController` to prefer `HMIS_LOGO` over `LOGO` if it is present in env.
The feature also depends on dev ops work (see ticket), but this is fine to merge independently from that work, as LOGO will still be used if HMIS_LOGO is not present.
PT issue: https://www.pivotaltracker.com/story/show/187220552

NB I tested this by just running the relevant code in rails console; is there another way we should test it?

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/controllers/hmis/app_settings_controller.rb
```diff
@@ -11,7 +11,8 @@ class Hmis::AppSettingsController < Hmis::BaseController
   def show
     okta_enabled = ENV['HMIS_OKTA_CLIENT_ID'].present? && ENV['OKTA_DOMAIN'].present?
 
-    logo = ENV['LOGO']
+    hmis_logo = ENV['HMIS_LOGO']
+    logo = hmis_logo && !hmis_logo.blank? ? hmis_logo : ENV['LOGO'] # prefer HMIS_LOGO if provided, otherwise LOGO
```
gigxz: Hmm I think this could be made more succinct. `presence` will convert `'' => nil` which is a bit neater:
```
# prefer HMIS_LOGO if provided, otherwise LOGO (which is shared with the warehouse)
logo = ENV['HMIS_LOGO']&.presence || ENV['LOGO']&.presence
if logo
  ...
```

gigxz:
>  I tested this by just running the relevant code in rails console; is there another way we should test it?

So yeah you should be able to test it! Here's what I have:
* A file at `open_path.svg` at `app/assets/images/theme/logo/open_path.svg` (I can share this file with you, or you can use any image)
* In my .env.local, `LOGO='open_path.svg'`

If you set that up and restart the backend, and pull in this frontend change https://github.com/greenriver/hmis-frontend/pull/702, then you should see the logo on the login page.

gigxz (approved):

PR #4153, opened 2024-03-21
Fix exit reason form logic for deceased client

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/187232760 

When the exit destination is set to Deceased, then don't show the Voluntary/Involuntary question.

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
```diff
@@ -106,17 +111,10 @@
...
               {
-                "question": "voluntary-exit-reason",
-                "operator": "EQUAL",
-                "answer_code": "Other"
-              },
-              {
-                "question": "involuntary-exit-reason",
-                "operator": "EQUAL",
-                "answer_code": "Other"
+                "question": "exit-reason",
```
martha, line 116: The reason for this change is to avoid the following bug:
1. user clicks any exit destination other than "Deceased"
2. user clicks either "voluntary" or "involuntary"
3. user selects "Other"
4. user returns to first question and now selects "Deceased"
5. the "Other" text field still appears and is still Required

Screenshot 2024-03-21 at 10 58 57 AM

I've added these enable_when conditions to work around this, on the assumption that we don't have any enable_behavior that can combine ANY and ALL approaches -- e.g. ALL of one condition and ANY of another set of conditions? but if we do have this, or you think I should implement it for this ticket, then I think that would be even better.

One thing that's not ideal about the way I have implemented it now, is that this GROUP is still shown when the exit reason exists, even when "Other" is not selected and the questions themselves don't show up, causing a bit of extra space:
Screenshot 2024-03-21 at 11 02 06 AM

drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
```diff
@@ -22,6 +22,14 @@
               {
                 "code": "Involuntary Termination"
               }
+            ],
+            "enable_behavior": "ALL",
```
martha, line 26: Is there any way to prevent the behavior of the question nesting in one level with a left grey bar, now that it's dependent on another question? If not, should I write some code to make it overrideable, or is it ok as-is?
Screenshot 2024-03-21 at 11 05 29 AM

drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
gigxz: Thats right, we don't have any way to combine ANY and ALL with the current system. It is definitely an annoyance, and there are other places  where we use groups to work around it (like you did here).

drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
gigxz: Ahmmm yeah, there is some logic like "show this vertical indent bar if this is a 'conditional' group", but it doesn't make sense in all cases, like here. I think we should consider removing that logic. [This is the code link](https://github.com/greenriver/hmis-frontend/blob/42a27d6fa82c27775ee729a21b8e566ece9fc997/src/modules/form/components/group/QuestionGroup.tsx#L22-L26). I think it makes sense to move to something more declarative, where the form author can specify if they want the item to appear "indented" or not. If you can put something  in the icebox with your screenshot and a link to that code, that would be great!

(And this is fine to leave as is, its not worth fiddling with imo!)

drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
gigxz: I [added a story](https://www.pivotaltracker.com/story/show/187289125). This approach looks good to me given the current constraints.

gigxz (approved):
Looks good! Sorry about the system constraints, & thank you for highlighting them so clearly.

drivers/hmis/lib/form_data/allegheny/fragments/patches/exit_reason.json
martha: Ok great, I created https://www.pivotaltracker.com/story/show/187289769 !

gigxz: Thanks!

PR #4160, opened 2024-03-22
186704906 public pages encrypted attrs

ttoomey:

## Description

Support encrypted spam scores.

Because recaptcha tokens only last a minute or two, we rely on the lambda to validate spam scores. That means the scores are returned to the client before upload to the s3 presigned url. Without encryption, we have to trust the client to post the score to s3 without manipulation. There's nothing to stop the client from manipulating the score, which kind of defeats the purpose.

**Note**: we could wait on this until we see if spam submissions are real problem. It might be fine; however without this it's possible for a spambot or bad-actor to flood submissions 

### env change
will need to create a new AES secret key as a remote crdential:
```ruby
    GrdaWarehouse::RemoteCredentials::SymmetricEncryptionKey.where(slug: 'external_forms_shared_key').first_or_create! do |record|
      record.algorithm = 'aes-256-cbc'
      record.key_hex = 'shared secret with the aws lamdba'
    end
```

## Type of change


lib/tasks/grda_warehouse.rake
```diff
@@ -320,7 +320,7 @@ namespace :grda_warehouse do
       Rake::Task['driver:hmis_external_apis:export:ac_clients'].invoke
     end
 
-    HmisExternalApis::ConsumeExternalFormSubmissionsJob.new.perform if HmisEnforcement.hmis_enabled? && GrdaWarehouse::DataSource.hmis.exists? && RailsDrivers.loaded.include?(:hmis_external_apis)
+    HmisExternalApis::ConsumeExternalFormSubmissionsJob.new.perform(encryption_key: ENV['EXTERNAL_FORMS_SHARED_AES_KEY']) if HmisEnforcement.hmis_enabled? && GrdaWarehouse::DataSource.hmis.exists? && RailsDrivers.loaded.include?(:hmis_external_apis)
```
eanders: Any reason this can't be stored in the remote credential for the s3 connection instead of the ENV?  We could just add another column to hold it.  I've found storing the config in the DB to be more maintainable long-term.

eanders (approved):
This looks good.  One request inline, if it seems reasonable, otherwise, feel free to merge!  Thanks @ttoomey

gigxz (approved):

lib/tasks/grda_warehouse.rake
ttoomey: sure thing!

ttoomey: @eanders I created a new type of credential "SymmetricEncryptionKey" and moved the decrypt method into it. Does that seem okay?

app/models/grda_warehouse/remote_credentials/symmetric_encryption_key.rb
```diff
@@ -0,0 +1,32 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module GrdaWarehouse
+  class RemoteCredentials::SymmetricEncryptionKey < GrdaWarehouse::RemoteCredential
+    # FIXME: we have a peculiar convention of repurposing the attributes of the remote credential class. Needs refactor
```
eanders:
```suggestion
    # FIXME: we have a convention of repurposing the attributes of the remote credential class. Needs refactor
```

eanders (approved):
Thanks for the ENV removal @ttoomey !  Looks good to me.

PR #4163, opened 2024-03-22
Proposed changes to viewing HMIS assessments in wh

gigxz:


app/views/clients/hud_assessments/show.haml
```diff
@@ -1,4 +1,4 @@
-- title = 'HUD Assessment'
+- title = 'HUD Coordinated Entry Assessment'
```
gigxz, line 1: intake/update/exit are referred to as "HUD Assessments" so I think this is clearer

app/views/clients/rollup/_assessments.html.haml
```diff
@@ -77,7 +77,7 @@
           - any_buttons = assessments.any?
           - cz = "assessment_type_#{i}"
           %tr{ class: ( 'assessment__new-type' if any_buttons )}
-            %td= link_to name, client_hud_assessment_path(@client, assessment), class: 'btn btn-primary btn-muted btn-xs', data: { loads_in_pjax_modal: true }
+            %td= link_to_if can_view_enrollment_details_tab?, name, client_hud_assessment_path(@client, assessment), class: 'btn btn-primary btn-muted btn-xs', data: { loads_in_pjax_modal: true }
```
gigxz, line 80: I think there are other links that need this too. When I turned off the permission, it would open an empty modal and then redirect me somewhere.

As an aside, the permission description states `Users with Can View Clients get this automatically.` which doesnt seem to be true, since I wasn't able to view these details.

drivers/hmis/app/views/hmis_client/assessments/show.haml
```diff
@@ -36,5 +36,11 @@
...
         %dd= value
-- else
+- elsif !@assessment.hud_assessment?
   .none-found No questions provided.
+-# Always link to HMIS for HUD Assessments, since the majority of data is non-Custom so wont be shown here.
+- if @assessment.hud_assessment?
+  .none-found
+    %a{href: @assessment.data_source.hmis_url_for(@assessment), target: :blank}
+      Open in HMIS
+      %i.icon-link-ext.ml-2{style: 'color: inherit; font-size: inherit;'}
```
gigxz, line 46: There typically won't be any data shown for HUD Assessments, so I included this link. I could take or leave it, though. If the Intake Assessment has any custom fields, they would be shown above as well.

Another option is to just not link to a modal for HUD Assessments.  I'm fine either way .

Screenshot 2024-03-22 at 10 01 31 AM

app/views/clients/rollup/_assessments.html.haml
eanders: Thanks!  pesky `&&` that should have been `||`

drivers/hmis/app/views/hmis_client/assessments/show.haml
eanders: I think that works well, thanks!

eanders (approved):

PR #4164, opened 2024-03-22
Fix duplicate enrollment bug

martha:

## Description
[PT issue](https://www.pivotaltracker.com/story/show/187248710)
This PR addresses and adds a regression test for the bug described in the PT ticket above. We added the `acts_as_paranoid` gem to `Hmis::Wips`, so deleted rows are only soft-deleted, but we needed to also filter out soft-deleted rows in the `hmis_client_projects` view, otherwise they will be returned when calling `project.enrollments_including_wip`.

Note, @gigxz I think while we were pairing on this bug we thought that the problem was with in-progress enrollments, but while I was writing the regression test, I learned that the problem is actually with not-in-progress enrollments.
- If the enrollment _used_ to be in progress, then it has a `hmis_wip` whose `deleted_at` date is not null. 
- Before the clients are merged, one enrollment is returned, thanks to that the first part of the `hmis_client_projects` view definition -- which joins client/project/enrollment tables and selects only rows where Client is not deleted
- During merge, one client is deleted, and if that client had an enrollment, it is updated to point at the non-deleted merged client.
- After the merge, that enrollment is still returned by that first part of the view definition, BUT, the second part of the definition which queries `hmis_wips` table now returns a record that still has the old (deleted) client ID attached. And because the client ID is different, it's not filtered out as a duplicate (which the UNION query would otherwise do).

## Type of change
- [x] Bug fix
- [ ] Dependency update


gigxz:
Thanks for the explanation @martha 💯 

I wonder about this piece:
> If the enrollment used to be in progress, then it has a hmis_wip whose deleted_at date is not null.

Do you know where/why that's happening? The below code is called when an Enrollment moves from WIP=>Non-WIP. It calls `destroy!` on the wip record, which should set deleted_at. (At least since acts_as_paranoid was added. Before that, the wip record was just hard-deleted).

So, I'm not following why there are undeleted wip records for non-wip enrollments. (and I'm not seeing any like that locally)

https://github.com/greenriver/hmis-warehouse/blob/5b708e9df524571f8e1f747a9dca3d7cc0fb1dc9/drivers/hmis/app/models/hmis/hud/enrollment.rb#L322-L328

db/views/hmis_client_projects_v03.sql
```diff
@@ -0,0 +1,38 @@
...
+    "Enrollment"."EnrollmentID",
+    "Enrollment"."HouseholdID",
+    "Enrollment"."data_source_id"
+  FROM
+    "hmis_wips"
+    INNER JOIN "Enrollment" ON "Enrollment"."DateDeleted" IS NULL
+    AND "Enrollment"."id" = "hmis_wips"."source_id"
+  WHERE
+    "hmis_wips"."source_type" = 'Hmis::Hud::Enrollment'
+  AND "hmis_wips"."deleted_at" IS NULL
```
gigxz: This line is the only change from `db/views/hmis_client_projects_v02.sql`, correct? It can be helpful to call that out since we don't get a nice diff with view changes.

gigxz: Prettifying indentation
```suggestion
    AND "hmis_wips"."deleted_at" IS NULL
```

martha:
@gigxz sorry this wasn't clear -- the `hmis_wip` records *are* deleted for not-in-progress enrollments! They still exist in the DB but with `deleted_at` date set, thanks to the acts_as_paranoid gem.

Then when the `hmis_wips` table is queried but without that filter on `deleted_at` that's added in this PR, that is what leads to the bug. My explanation was just meant to clarify that it's not currently-WIP enrollments that are the problem, but formerly-WIP enrollments that have this soft-deleted `hmis_wip` row.

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
```diff
@@ -145,6 +145,21 @@
...
+    let!(:p1) { create(:hmis_hud_project, data_source: data_source, organization: o1) }
+    let!(:e1) { create(:hmis_hud_wip_enrollment, client: client2, project: p1, data_source: data_source) }
+
+    # This is a regression test for a bug where deleted rows in hmis_wips were still returned by project.enrollments_including_wip.
+    # The bug was fixed by selecting only non-deleted rows in the hmis_client_projects view.
+    it 'does not result in duplicate enrollment records' do
+      e1.save_not_in_progress!
+      expect(p1.enrollments_including_wip.count).to eq(1)
+      Hmis::MergeClientsJob.perform_now(client_ids: [client1.id, client2.id], actor_id: actor.id)
+      expect(p1.enrollments_including_wip.count).to eq(1), 'it hould not show dupes'
```
gigxz:
```suggestion
      expect(p1.enrollments_including_wip.count).to eq(1), 'it should not show dupes'
```

db/views/hmis_client_projects_v03.sql
martha: Yes that's right! I'll add a comment

gigxz:
You were super clear @martha I just wasn't reading correctly! sorry about that! I deleted my comment. This looks great, thank you. 

Just for linking purposes-- this bug was introduced with https://github.com/greenriver/hmis-warehouse/pull/3863#discussion_r1440080460

gigxz (approved):
Looks great! Tagging @ttoomey as an FYI! (TLDR we added `acts_as_paranoid` to the Wip model without updating the view)

PR #4167, opened 2024-03-22
Add deleted_at clause to hmis_households view

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/187289783
Since `acts_as_paranoid` gem has been added to the Households model, we need to also add this filter to the view to ensure we never return duplicate rows. (However, see comment inline for why this bug was not surfaced)

## Type of change
- [x] Bug fix
- [ ] Dependency update


db/views/hmis_households_v06.sql
```diff
@@ -0,0 +1,80 @@
...
+      LEFT OUTER JOIN "Exit" ON "Exit"."EnrollmentID" = "Enrollment"."EnrollmentID"
+      AND "Exit"."data_source_id" = "Enrollment"."data_source_id"
+      AND "Exit"."DateDeleted" IS NULL
+      JOIN "hmis_wips" ON "hmis_wips"."source_id" = "Enrollment"."id"
+      AND "hmis_wips"."source_type" = 'Hmis::Hud::Enrollment'
+      JOIN "Project" ON "Project"."DateDeleted" IS NULL
+      AND "Project"."id" = "hmis_wips"."project_id"
+    WHERE
+      "Enrollment"."DateDeleted" IS NULL
+      AND "Enrollment"."ProjectID" IS NULL
```
martha, line 46: I spent a few minutes trying to see if I could reproduce a bug and create a regression test. I _think_ this line is the reason why the bug was not surfaced. `"Enrollment"."ProjectID"` is only NULL when the enrollment is WIP. When it's not, then even if the `hmis_wips` table still contains a now-deleted row for this enrollment, it wouldn't be returned by this query because non-WIPs are filtered out at this line.

gigxz (approved):

ttoomey (approved):
Looks good, thank you!

PR #4170, opened 2024-03-24
187269047-dangling-enrollments

ttoomey:

## Description

[PT Issue](https://www.pivotaltracker.com/story/show/187269047)

Expands our existing monitor reports to include invalid HUD references. Specifically:
* dangling client references by data source/table
* dangling and mismatched enrollment references by data_source/table
* dangling project references by data source/table
* duplicate hud_keys by data source/table

## Type of change


app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,165 @@
...
+      Hmis::Hud::Exit,
+      Hmis::Hud::Funder,
+      Hmis::Hud::HealthAndDv,
+      Hmis::Hud::HmisParticipation,
+      Hmis::Hud::IncomeBenefit,
+      Hmis::Hud::Inventory,
+      Hmis::Hud::ProjectCoc,
+      Hmis::Hud::Service,
+      Hmis::Hud::YouthEducationStatus,
+    ].flat_map do |model|
```
eanders: I think we should lean on some existing lists if possible, and not on the HMIS driver (if possible).
We might want to make a method somewhere in the HMIS driver that could return the custom data, or maybe we should just add it to the `by_hud_filename` method.  I'm not 100% sure where that's used or if it would matter, but I'd like to keep the lists of available classes from growing since they change every couple years and tracking them down can be a pain.
```suggestion
    hud_models = GrdaWarehouse::Hud.models.by_hud_filename.values
    if RailsDrivers.loaded.include?(:hmis)
      hud_models += [
        Hmis::Hud::CustomAssessment,
        Hmis::Hud::CustomCaseNote,
        Hmis::Hud::CustomClientAddress,
        Hmis::Hud::CustomClientContactPoint,
        Hmis::Hud::CustomClientName,
        Hmis::Hud::CustomService,
       ]
     end
     hud_models.sort.flat_map do |model|
```

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,165 @@
...
+      Hmis::Hud::Disability,
+      Hmis::Hud::EmploymentEducation,
+      Hmis::Hud::Enrollment,
+      Hmis::Hud::Event,
+      Hmis::Hud::Exit,
+      Hmis::Hud::HealthAndDv,
+      Hmis::Hud::IncomeBenefit,
+      Hmis::Hud::Service,
+      Hmis::Hud::YouthEducationStatus,
+    ].flat_map do |model|
```
eanders: I swear we have this somewhere as well but I can't find it.  Something like:
```
GrdaWarehouse::Hud.models.by_hud_filename.values.select { |model| model.column_names.include?('PersonalID') }
```
as a starting point would feel less brittle to me.

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,165 @@
...
+    # models that reference a project via (ProjectID)
+    [
+      Hmis::Hud::Affiliation,
+      Hmis::Hud::CeParticipation,
+      Hmis::Hud::Enrollment,
+      Hmis::Hud::Funder,
+      Hmis::Hud::HmisParticipation,
+      Hmis::Hud::Inventory,
+      Hmis::Hud::ProjectCoc,
+    ].flat_map do |model|
```
eanders: Same here, something like:
```
GrdaWarehouse::Hud.models.by_hud_filename.values.select { |model| model.column_names.include?('ProjectID') }
```

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,165 @@
...
+      Hmis::Hud::CustomService,
+      Hmis::Hud::Disability,
+      Hmis::Hud::EmploymentEducation,
+      Hmis::Hud::Event,
+      Hmis::Hud::Exit,
+      Hmis::Hud::HealthAndDv,
+      Hmis::Hud::IncomeBenefit,
+      Hmis::Hud::Service,
+      Hmis::Hud::YouthEducationStatus,
+    ].flat_map do |model|
```
eanders: Again:
```
GrdaWarehouse::Hud.models.by_hud_filename.values.select { |model| model.column_names.include?('EnrollmentID') }
```

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,165 @@
...
+          # PersonalID is different from enrollment.PersonalID ("mismatch")
+          mismatched_enrollments: scope.joins(enrollment_join_sql).where(arel_table[:personal_id].not_eq(arel.e_t[:personal_id])).count,
+        }
+      end
+    end
+  end
+
+  def arel
+    Hmis::ArelHelper.instance
+  end
```
eanders, line 114: I think I like this mechanism of bringing in Arel Helper.  Definitely compartmentalizes and clarifies the use.  Thanks!

eanders (commented):
@ttoomey this looks generally like a good idea and I think the SQL is good.  I'd like to avoid more creep of lists of classes if possible since those do change every few years.

If you think that's too much for this PR, we can just make a story to clean that up as tech debt in the future.

ttoomey:
> @ttoomey this looks generally like a good idea and I think the SQL is good. I'd like to avoid more creep of lists of classes if possible since those do change every few years.
> 
> If you think that's too much for this PR, we can just make a story to clean that up as tech debt in the future.

Sure, I removed the class lists and changed it to introspect on the descendants of the base model. Or would you rather I use by_hud_filename? Also added some `references_foo*?` helper methods. Does that seem okay?

app/models/grda_warehouse_base.rb
```diff
@@ -55,4 +55,16 @@ def paper_trail_meta_value(key)
       value
     end
   end
+
+  def self.references_hud_client?
+    columns_hash.key?('PersonalID') && reflect_on_association(:client)
+  end
+
+  def self.references_hud_enrollment?
+    columns_hash.key?('EnrollmentID') && reflect_on_association(:group)
```
eanders: Should this be `:group` or `:enrollment`?  I'm not familiar with the `:group` relation.

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,119 @@
...
+    new.duplicate_ids
+  end
+
+  def hud_models
+    Rails.application.eager_load!
+    seen = Set.new
+    results = []
+    [
+      GrdaWarehouse::Hud::Base,
+      Hmis::Hud::Base,
```
eanders, line 31: Thanks!  This seems much less likely to get out of sync in the future.

eanders (approved):
Thanks @ttoomey !

app/models/grda_warehouse_base.rb
gigxz: I think this should be `:enrollment`

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,119 @@
...
+  end
+
+  def client_references
+    hud_models.filter(&:references_hud_client?).flat_map do |model|
+      data_sources.map do |data_source|
+        scope = model.where(data_source: data_source)
+        {
+          data_source_id: data_source.id,
+          table_name: model.table_name,
+          dangling_clients: scope.left_outer_joins(:client).where(arel.c_t[:id].eq(nil)).count,
```
gigxz, line 69: I'm able to run this, but where does the `arel` variable come from?

app/models/grda_warehouse_base.rb
```diff
@@ -55,4 +55,16 @@ def paper_trail_meta_value(key)
       value
     end
   end
+
+  def self.references_hud_client?
+    columns_hash.key?('PersonalID') && reflect_on_association(:client)
```
gigxz: `?` convention should return a boolean (I think) but this returns the reflection or `false` or `nil`

app/models/app_resource_monitor/hud_references_inspector.rb
```diff
@@ -0,0 +1,119 @@
...
+
+  def self.client_references
+    new.client_references
+  end
+
+  def self.duplicate_ids
+    new.duplicate_ids
+  end
+
+  def hud_models
```
gigxz, line 25: Should the results of this be stored in an instance variable? It seems unnecessary to recalculate it (and re-eager-load) for each reference check call.

app/models/grda_warehouse_base.rb
ttoomey: Yes, should be `:enrollment`, sorry I missed this typo. Thanks for catching!

app/models/grda_warehouse_base.rb
ttoomey: good thought, the type could be tighter, will fix

app/models/app_resource_monitor/hud_references_inspector.rb
ttoomey: it's just a method down on line 112
```ruby
  def arel
    Hmis::ArelHelper.instance
  end
```

gigxz: oops, thanks

gigxz (approved):
Looks great once the enrollment reference is fixed! Thank you!!

app/models/app_resource_monitor/hud_references_inspector.rb
ttoomey: Yes, memoizing this would make the it marginally more performant; probably save a few 100ms? I would totally do that if this was used in a request context. But this is called from a cron at long intervals so it's hard for me to justify the time for micro optimizations

PR #4180, opened 2024-03-26
187269047-exclude-hmis-services-for-invalid-enrollments

ttoomey:

## Description
[PT issue](https://www.pivotaltracker.com/story/show/187269047)
Speculative solution to for services linked to invalid enrollments. Assumes relatively few invalid enrollments so breaking up into two queries might be faster than an additional join

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/project.rb
+++ b/drivers/hmis/app/models/hmis/hud/project.rb
@@ -69,7 +69,10 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   # minutes. It needs optimization; for now we use a class method instead of a AR association
   # has_many :hmis_services, through: :enrollments_including_wip
   def hmis_services
-    Hmis::Hud::HmisService.joins(:project).where(Hmis::Hud::Project.arel_table[:id].eq(id))
+    invalid_enrollment_ids = enrollments.with_invalid_references.pluck(e_t[:enrollment_id])
+    Hmis::Hud::HmisService.joins(:project).
+      where(Hmis::Hud::Project.arel_table[:id].eq(id)).
+      where(Hmis::Hud::HmisService.arel_table[:EnrollmentID].not_in(invalid_enrollment_ids))
   end
 
   has_and_belongs_to_many :project_groups,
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -69,7 +69,10 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   # minutes. It needs optimization; for now we use a class method instead of a AR association
   # has_many :hmis_services, through: :enrollments_including_wip
   def hmis_services
-    Hmis::Hud::HmisService.joins(:project).where(Hmis::Hud::Project.arel_table[:id].eq(id))
+    invalid_enrollment_ids = enrollments.with_invalid_references.pluck(e_t[:enrollment_id])
+    Hmis::Hud::HmisService.joins(:project).
+      where(Hmis::Hud::Project.arel_table[:id].eq(id)).
+      where(Hmis::Hud::HmisService.arel_table[:EnrollmentID].not_in(invalid_enrollment_ids))
```
eanders, line 75: why the `not_in` here, instead of an in with a `not_eq(nil)` in the scope on enrollment?  I'm not too familiar with the performance around a left outer join, but generally postgres seems faster at dealing with non-null values.  (Of course it also has to excluded those.)

gigxz (approved):
works for me as a temporary fix. there is a relatively high number of mismatched enrollments in some installations, but we plan to clean those up, so once the clean-up is done the number should be low.

drivers/hmis/app/models/hmis/hud/project.rb
ttoomey: Sorry I'm probably misunderstanding but are you suggesting inverting this (plucking all the valid ids out rather than the invalid ones)? Assuming the set of valid enrollments is large compared to the invalid ones, my intuition is that any gains in query efficiency would be offset by selecting a much larger set of ids. Perhaps my intuition is wrong on that though; it seems pretty dependent on the shape of the data in production.

eanders: Oh, right, I forgot you had done a pluck above, any reason not to make that a select/sub query?
I agree, the number of invalid _should_ be much smaller than the number of valid.  Let's proceed with what you have and revisit if there are performance issues.

ttoomey: yes a subquery might well be faster; we can try that if this approach isn't performant

ttoomey:
@gigxz should I merge this into release-110 or 109? Not sure which is best for launch

gigxz:
109! thanks

PR #4185, opened 2024-03-27
CoC and date range enforcement

eanders:

## Description

This does two things:
1. Enforces CoC code the HMIS CSV exporter
 a. if you choose a CoC Code and projects that don't operate in that CoC, the exporter will ignore the projects
 b. if you choose a CoC Code and project operates in more than one CoC, then only the data (Project CoC and Enrollment) for the chosen CoC Code will be exported.
2. For the LSA (and probably in the future all HUD reports) we now limit the project IDs to those that operate during the reporting range OR have ongoing enrollments that overlap the reporting range.  That way if you pick a project with no enrollments that is closed, it will be ignored. 
 
## Type of change
- [x] Bug fix

### DIFF
--- a/app/models/filters/hud_filter_base.rb
+++ b/app/models/filters/hud_filter_base.rb
@@ -70,7 +70,7 @@ module Filters
         ids_with_enrollments = GrdaWarehouse::Hud::Project.
           where(id: effective_project_ids).
           joins(:enrollments).
-          merge(GrdaWarehouse::Hud::Enrollemnt.open_during_range(effective_range)).
+          merge(GrdaWarehouse::Hud::Enrollment.open_during_range(effective_range)).
           distinct.
           pluck(:id)
         ids_for_open_projects = GrdaWarehouse::Hud::Project.
--- a/app/models/filters/hud_filter_base.rb
+++ b/app/models/filters/hud_filter_base.rb
@@ -59,6 +59,16 @@ module Filters
       @effective_project_ids.uniq.reject(&:blank?)
     end
 
+    # Limit the effective project ids to only those with enrollments that overlap the report range
+    def effective_project_ids_with_ongoing_enrollments
+      @effective_project_ids_with_ongoing_enrollments ||= GrdaWarehouse::Hud::Project.
+        where(id: effective_project_ids).
+        joins(:enrollments).
+        merge(GrdaWarehouse::Hud::Enrollemnt.open_during_range(range)).
+        distinct.
+        pluck(:id)
+    end
+
     def apply(scope, except: [])
       # @filter is required for these to work
       @filter = self
--- a/app/models/filters/hud_filter_base.rb
+++ b/app/models/filters/hud_filter_base.rb
@@ -60,13 +60,25 @@ module Filters
     end
 
     # Limit the effective project ids to only those with enrollments that overlap the report range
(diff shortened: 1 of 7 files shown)


app/models/filters/hud_filter_base.rb
```diff
@@ -59,6 +59,28 @@ def effective_project_ids
...
+    # Projects with operating dates overlapping the range.
+    # Looking for both of these will enable reporting on non-HMIS participating projects open during
+    # the report range, and will allow the data quality checks for enrollments that are open
+    # outside of the operating end dates
+    def effective_project_ids_during_range(effective_range)
+      @effective_project_ids_during_range ||= begin
+        ids_with_enrollments = GrdaWarehouse::Hud::Project.
+          where(id: effective_project_ids).
+          joins(:enrollments).
+          merge(GrdaWarehouse::Hud::Enrollemnt.open_during_range(effective_range)).
```
gigxz:
```suggestion
          merge(GrdaWarehouse::Hud::Enrollment.open_during_range(effective_range)).
```

It doesn't look like this was caught by your tests - maybe worth adding test coverage? Not sure what kind of tests we have for Filter currently, but I thought we had some.

app/models/filters/hud_filter_base.rb
```diff
@@ -59,6 +59,28 @@ def effective_project_ids
...
     end
 
+    # Limit the effective project ids to only those with enrollments that overlap the report range
+    # OR
+    # Projects with operating dates overlapping the range.
+    # Looking for both of these will enable reporting on non-HMIS participating projects open during
+    # the report range, and will allow the data quality checks for enrollments that are open
+    # outside of the operating end dates
+    def effective_project_ids_during_range(effective_range)
+      @effective_project_ids_during_range ||= begin
```
gigxz: This seems like it could  lead to errors down the road - if this is called repeatedly with different `effective_range`s, it will always return the first one. Are we confident that  won't happen?
And semi related question, why can't we pull the `effective_range` from the start/end dates on the Filter?

app/models/filters/hud_filter_base.rb
eanders: Ooh, thanks.  Yeah, I'm testing this manually right now, but I'll see if we have any HUD Filter tests.  Even some simple unit tests here would be nice.

app/models/filters/hud_filter_base.rb
eanders: So, I don't think we'd ever call this with different ranges in the same iteration, but good point, I'd added the memoization before realizing we needed to add the range.  I'll adjust to memoize based on range.

We need the range because the issue it's fixing is related to the HIC, where HUD guidance of "run a 1 day LSA" isn't actually what they mean.  They mean, pull the data for a 1 day range and then run the LSA against it.  Our 1 day LSA still had a 7 year look-back. (Which is totally unnecessary for the HIC.)

gigxz: Gotcha. Sounds good. Thanks!

dtgreiner (approved):

gigxz (approved):

eanders:
Closing in favor of https://github.com/greenriver/hmis-warehouse/pull/4187 which is the same changes against 108 so we can deploy sooner.

PR #4191, opened 2024-03-27
Add cleanup function for fixing incorrect PersonalID references, other clean fn improvements and tests

gigxz:

## Description

* Add `fix_incorrect_personal_id_references!` cleanup util [PT STORY](https://www.pivotaltracker.com/n/projects/2591838/stories/187269047)
* Make some existing utils more performant (hopefully!)
* Add test coverage (incomplete)

## Type of change

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/concerns/shared.rb
+++ b/drivers/hmis/app/models/hmis/hud/concerns/shared.rb
@@ -21,33 +21,45 @@ module Hmis::Hud::Concerns::Shared
       "GrdaWarehouse::Hud::#{self.class.name.demodulize}".constantize.find(id)
     end
 
-    def self.hud_class_names
+    def self.enrollment_related_hud_class_names
       [
-        'Export',
-        'Organization',
-        'Project',
-        'Client',
         'Disability',
         'EmploymentEducation',
-        'Enrollment',
-        'HmisParticipation',
-        'CeParticipation',
         'Exit',
-        'Funder',
         'HealthAndDv',
         'IncomeBenefit',
-        'Inventory',
-        'ProjectCoc',
-        'Affiliation',
         'Service',
         'CurrentLivingSituation',
         'Assessment',
         'AssessmentQuestion',
         'AssessmentResult',
         'Event',
-        'User',
         'YouthEducationStatus',
       ].freeze
+    end
+
+    def self.hud_class_names
+      [
+        *enrollment_related_hud_class_names,
+        'Export',
+        'Organization',
+        'Project',
+        'Client',
+        'Enrollment',
+        'HmisParticipation',
+        'CeParticipation',
+        'Funder',
+        'Inventory',
+        'ProjectCoc',
+        'Affiliation',
+        'User',
+      ].freeze
+    end
+
+    def self.hmis_enrollment_related_classes
+      enrollment_related_hud_class_names.map do |name|
+        "Hmis::Hud::#{name}".constantize
+      end
     end
 
     def self.hmis_classes
--- a/drivers/hmis/app/models/hmis/hud/current_living_situation.rb
+++ b/drivers/hmis/app/models/hmis/hud/current_living_situation.rb
@@ -22,7 +22,7 @@ class Hmis::Hud::CurrentLivingSituation < Hmis::Hud::Base
   after_commit :warehouse_trigger_processing
 
   private def warehouse_trigger_processing
-    return unless warehouse_columns_changed?
+    return unless enrollment && warehouse_columns_changed?
 
     # NOTE: we only really need to do this for SO at the moment, but this is future-proofing against
     # pre-processing CLS in other enrollments
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -73,7 +73,7 @@ module HmisDataCleanup
     end
 
     # Fix any instances of enrollment-related records where the PersonalID does nots match the Enrollment's PersonalIDs
-    def self.fix_incorrect_personal_id_references!(classes: nil)
+    def self.fix_incorrect_personal_id_references!(classes: nil, dry_run: false)
       classes&.each do |klass|
         raise "Invalid class: #{klass.name}" unless Hmis::Hud::Enrollment.hmis_enrollment_related_classes.include?(klass)
       end
@@ -84,6 +84,16 @@ module HmisDataCleanup
       Hmis::Hud::Base.transaction do
         classes.each do |klass|
           Rails.logger.info "[#{klass.name}] Processing"
+
+          if dry_run
+            records_needing_update = klass.where(data_source_id: data_source_id).
+              left_outer_joins(:enrollment).
+              where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+              count
+
+            Rails.logger.info "[#{klass.name}] #{records_needing_update} records with bad PersonalID"
+            next
+          end
 
           klass.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
             where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -17,18 +17,9 @@ module HmisDataCleanup
     # Remove all ExportIDs from HMIS Enrollments.
     # This should be done after an HMIS migration, otherwise ServiceHistoryService generation will behave incorrectly
     def self.clear_enrollment_export_ids!
-      Hmis::Hud::Enrollment.hmis.where.not(ExportID: nil).find_in_batches(batch_size: 5_000) do |batch|
-        batch.each { |enrollment| enrollment.ExportID = nil }
-
-        result = Hmis::Hud::Enrollment.import(
-          batch,
-          validate: false,
-          timestamps: false,
-          on_duplicate_key_update: { conflict_target: [:id], columns: [:ExportID] },
-        )
-        raise "Failed: #{result.failed_instances}" if result.failed_instances.present?
-
-        Rails.logger.info "Nullified ExportID on #{result.ids.count} Enrollments"
+      without_papertrail_or_timestamps do
+        rows_affected = Hmis::Hud::Enrollment.hmis.where.not(ExportID: nil).update_all(ExportID: nil)
+        Rails.logger.info "Nullified ExportID on #{rows_affected} Enrollments"
       end
     end
 
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -72,38 +72,39 @@ module HmisDataCleanup
       end
     end
 
-    # UNTESTED!!!
-    def self.fix_bad_personal_id_references_on_services!
+    # Fix any instances of enrollment-related records where the PersonalID does nots match the Enrollment's PersonalIDs
+    def self.fix_incorrect_personal_id_references!(classes: nil)
+      classes&.each do |klass|
+        raise "Invalid class: #{klass.name}" unless Hmis::Hud::Enrollment.hmis_enrollment_related_classes.include?(klass)
+      end
+
+      classes ||= Hmis::Hud::Enrollment.hmis_enrollment_related_classes
       data_source_id = GrdaWarehouse::DataSource.hmis.first.id
-      # Find and fix Services where the PersonalID does not match the PersonalID on the referenced Enrollment
-      # We may want to expand this to other Enrollment-related record types
-      GrdaWarehouse::Hud::Service.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
-        where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
-        find_in_batches do |batch|
-        # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
-        eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
-          data_source_id: data_source_id,
-          EnrollmentID: batch.pluck(:EnrollmentID),
-        ).pluck(:EnrollmentID, :PersonalID).to_h
 
-        # update PersonalID on each Service
-        batch.each do |service|
-          real_personal_id = eid_to_pid[service.EnrollmentID]
-          unless real_personal_id
-            puts "Enrollment not found: Service##{service.id}, EnrollmentID##{service.EnrollmentID}"
-            next
+      Hmis::Hud::Base.transaction do
+        classes.each do |klass|
+          Rails.logger.info "[#{klass.name}] Processing"
+
+          klass.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
+            where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+            find_in_batches do |batch|
+            Rails.logger.info "[#{klass.name}] Processing batch"
+
+            # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of records
+            eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
+              data_source_id: data_source_id,
+              EnrollmentID: batch.map(&:EnrollmentID),
+            ).pluck(:EnrollmentID, :PersonalID).to_h
+
+            # values = []
+            batch.each do |record|
+              real_personal_id = eid_to_pid[record.EnrollmentID]
+              next unless real_personal_id
+
+              record.update_columns(PersonalID: real_personal_id)
+            end
           end
-
-          service.PersonalID = real_personal_id
         end
-        result = GrdaWarehouse::Hud::Service.import(
-          batch,
-          validate: false,
-          timestamps: false,
-          on_duplicate_key_update: { conflict_target: [:id], columns: [:PersonalID] },
-        )
-
-        raise "Failed: #{result.failed_instances}" if result.failed_instances.present?
       end
     end
 
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -72,6 +72,41 @@ module HmisDataCleanup
       end
     end
 
+    # UNTESTED!!!
+    def self.fix_bad_personal_id_references_on_services!
+      data_source_id = GrdaWarehouse::DataSource.hmis.first.id
+      # Find and fix Services where the PersonalID does not match the PersonalID on the referenced Enrollment
+      # We may want to expand this to other Enrollment-related record types
+      GrdaWarehouse::Hud::Service.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
+        where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+        find_in_batches do |batch|
+        # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
+        eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
+          data_source_id: data_source_id,
+          EnrollmentID: batch.pluck(:EnrollmentID),
+        ).pluck(:EnrollmentID, :PersonalID).to_h
+
+        # update PersonalID on each Service
+        batch.each do |service|
+          real_personal_id = eid_to_pid[service.EnrollmentID]
+          unless real_personal_id
+            puts "Enrollment not found: Service##{service.id}, EnrollmentID##{service.EnrollmentID}"
+            next
+          end
+
+          service.PersonalID = real_personal_id
+        end
+        result = GrdaWarehouse::Hud::Service.import(
+          batch,
+          validate: false,
+          timestamps: false,
+          on_duplicate_key_update: { conflict_target: [:id], columns: [:PersonalID] },
+        )
+
+        raise "Failed: #{result.failed_instances}" if result.failed_instances.present?
+      end
+    end
+
     def self.delete_duplicate_bed_nights!
       duplicates = Hmis::Hud::Service.hmis.where(record_type: 200).
         where.not(date_provided: nil).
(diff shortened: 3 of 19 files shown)


drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,73 @@ def self.clear_enrollment_export_ids!
...
+      data_source_id = GrdaWarehouse::DataSource.hmis.first.id
+      # Find and fix Services where the PersonalID does not match the PersonalID on the referenced Enrollment
+      # We may want to expand this to other Enrollment-related record types
+      GrdaWarehouse::Hud::Service.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
+        where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+        find_in_batches do |batch|
+        # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
+        eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
+          data_source_id: data_source_id,
+          EnrollmentID: batch.pluck(:EnrollmentID),
```
ttoomey: batch is an array here- pluck works but it's a bit confusing imo
```suggestion
          EnrollmentID: batch.map(&:EnrollmentID),
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,73 @@ def self.clear_enrollment_export_ids!
...
+    # UNTESTED!!!
+    def self.fix_bad_personal_id_references_on_services!
+      data_source_id = GrdaWarehouse::DataSource.hmis.first.id
+      # Find and fix Services where the PersonalID does not match the PersonalID on the referenced Enrollment
+      # We may want to expand this to other Enrollment-related record types
+      GrdaWarehouse::Hud::Service.where(data_source_id: data_source_id).left_outer_joins(:enrollment).
+        where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+        find_in_batches do |batch|
+        # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
+        eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
```
ttoomey: indentation/rubocop ?
```suggestion
        find_in_batches do |batch|
          # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
          eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,73 @@ def self.clear_enrollment_export_ids!
...
+          having(nf('COUNT', [:HouseholdID]).eq(1)).
           where.not(relationship_to_hoh: 1).
           update_all(relationship_to_hoh: 1) # skips callbacks
+
+        Rails.logger.info "Set HoH in #{rows_affected} single-member households"
+      end
+    end
+
+    # UNTESTED!!!
+    def self.fix_bad_personal_id_references_on_services!
```
ttoomey: maybe run the whole thing in a tx?

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,73 @@ def self.clear_enrollment_export_ids!
...
+        where(GrdaWarehouse::Hud::Enrollment.arel_table[:id].eq(nil)).
+        find_in_batches do |batch|
+        # map {EnrollmentID=>PersonalID} for each enrollment referenced by this batch of services
+        eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
+          data_source_id: data_source_id,
+          EnrollmentID: batch.pluck(:EnrollmentID),
+        ).pluck(:EnrollmentID, :PersonalID).to_h
+
+        # update PersonalID on each Service
+        batch.each do |service|
```
ttoomey: I think this works but the upsert on id including all the service attributes seems like it's doing a lot of extra stuff. Wondering if we could just update records that need to be fixed here, maybe would be much smaller record set so we could do individual updates:
```ruby
batch.each do |service|
  real_personal_id = eid_to_pid[service.EnrollmentID]
  if real_personal_id &&  service.PersonalID != real_personal_id
    service.update_columns(PersonalID: real_personal_id)
  end
end
```

ttoomey (commented):
reviewed fix_bad_personal_id_references_on_services, it makes sense to me, thanks!

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: That what I had originally, but I thought it might not be performant enough.
Would it be most performant to use an upsert of the style `Model.import column_names, array_of_values` so we are only specifying the value we care about?

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: rubocop wants them to be on the same line? if I indent it it complains about indentation

drivers/hmis/lib/hmis_data_cleanup/util.rb
ttoomey: @gigxz 
Yeah sorry I missed there were 140k records. That's too many for individual updates.

I'd suggest something like this
* filter the records to update just the ones that have a mismatched personal id
* narrow the columns to just the personal id and PK

```ruby
  values = []
  batch.each do |service|
    real_personal_id = eid_to_pid[service.EnrollmentID]
    unless real_personal_id
      puts "Enrollment not found: Service##{service.id}, EnrollmentID##{service.EnrollmentID}"
      next
    end
    values.push({id: service.id, PersonalID: real_personal_id}) if service.PersonalID != real_personal_id
  end
  if values.any?
    result = GrdaWarehouse::Hud::Service.import(
      values,
      validate: false,
      timestamps: false,
      on_duplicate_key_update: { conflict_target: [:id], columns: [:PersonalID] },
    )

    raise "Failed: #{result.failed_instances}" if result.failed_instances.present?
  end
```

drivers/hmis/app/models/hmis/hud/concerns/shared.rb
```diff
@@ -21,35 +21,47 @@ def as_warehouse
       "GrdaWarehouse::Hud::#{self.class.name.demodulize}".constantize.find(id)
     end
 
+    def self.enrollment_related_hud_class_names
```
gigxz, line 24: I don't love this approach, but I figured this was a place where we already listed them out.

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -17,8 +17,18 @@ class Util
     # Remove all ExportIDs from HMIS Enrollments.
     # This should be done after an HMIS migration, otherwise ServiceHistoryService generation will behave incorrectly
     def self.clear_enrollment_export_ids!
-      without_papertrail_or_timestamps do
-        Hmis::Hud::Enrollment.hmis.update_all(ExportID: nil)
+      Hmis::Hud::Enrollment.hmis.where.not(ExportID: nil).find_in_batches(batch_size: 5_000) do |batch|
+        batch.each { |enrollment| enrollment.ExportID = nil }
+
+        result = Hmis::Hud::Enrollment.import(
```
gigxz: Maybe I should switch this back to update_all or update_columns for perf reasons? this one will be large as its nearly all enrollments.

I'm not sure what is the most performant way to do that. I thought it was import. @ttoomey @eanders 
 thoughts on this? In this case it is nullifying one column.

gigxz:
need to let tests run through to see if factory changes broken anything. hopefully not.

drivers/hmis/app/models/hmis/hud/concerns/shared.rb
eanders: I'm ok with this here.  Probably eventually they should move to `HudUtility*` since they vary by year.

drivers/hmis/lib/hmis_data_cleanup/util.rb
eanders: This should be fine, `import` with a conflict key (maybe that's missing, or maybe it defaults to id?) will do an upsert/update_all.

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: Okay, sounds good. `if service.PersonalID != real_personal_id` should be unnecessary, the `batch` only includes records where that would be the case, because of the left outer join. but i can add it as an extra check.

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,87 @@ def self.clear_enrollment_export_ids!
...
     def self.assign_missing_household_ids!
-      scope = Hmis::Hud::Enrollment.hmis.where(household_id: nil)
-      Rails.logger.info "Assigning household id to #{scope.size} enrollments"
-      scope.each do |enrollment|
-        hh_id = Hmis::Hud::Base.generate_uuid
-        without_papertrail_or_timestamps do
-          enrollment.update_columns(household_id: hh_id) # skips callbacks
-        end
+      Hmis::Hud::Enrollment.hmis.where(household_id: [nil, '']).find_in_batches do |batch|
+        batch.each { |enrollment| enrollment.HouseholdID = Hmis::Hud::Base.generate_uuid }
```
eanders: This is fine, but if you want it to be consistent over time, something like https://github.com/greenriver/hmis-warehouse/blob/30d691487c2fae4f56815034f24594a66fa3357a/drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/fix_blank_household_ids.rb#L14 might be better.

drivers/hmis/lib/hmis_data_cleanup/util.rb
eanders: oh, the `conflict_target` is there, I just missed it.

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,87 @@ def self.clear_enrollment_export_ids!
...
-          where(household_id: single_member_households).
+        rows_affected = Hmis::Hud::Enrollment.hmis.where(household_id: single_member_households).
           where.not(relationship_to_hoh: 1).
           update_all(relationship_to_hoh: 1) # skips callbacks
+
+        Rails.logger.info "Set HoH in #{rows_affected} single-member households"
+      end
+    end
+
+    # Fix any instances of enrollment-related records where the PersonalID does nots match the Enrollment's PersonalIDs
```
eanders:
```suggestion
    # Fix any instances of enrollment-related records where the PersonalID does not match the Enrollment's PersonalIDs
```

eanders (approved):
Looks good to me.  Thanks!

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: updated thanks!

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+        and change(GrdaWarehouse::Hud::Exit, :count).by(0).
+        and change(GrdaWarehouse::Hud::HealthAndDv, :count).by(0).
+        and change(GrdaWarehouse::Hud::IncomeBenefit, :count).by(0).
+        and change(GrdaWarehouse::Hud::Service, :count).by(0).
+        and change(GrdaWarehouse::Hud::CurrentLivingSituation, :count).by(0).
+        and change(GrdaWarehouse::Hud::Assessment, :count).by(0).
+        and change(GrdaWarehouse::Hud::AssessmentQuestion, :count).by(0).
+        and change(GrdaWarehouse::Hud::AssessmentResult, :count).by(0).
+        and change(GrdaWarehouse::Hud::Event, :count).by(0).
+        and change(GrdaWarehouse::Hud::YouthEducationStatus, :count).by(0)
```
gigxz: Alright switch back to `import`ing records, and here we ensure that no new records were created

gigxz: I've had trouble with `not_to change` so I do change by 0.. not as readable but seems to be happy

drivers/hmis/spec/factories/hmis/hud/hmis_base_factory.rb
```diff
@@ -0,0 +1,18 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+FactoryBot.define do
+  factory :hmis_base_factory, class: 'Hmis::Hud::Base' do
```
ttoomey, line 8: Nice! Should this get set as the parent for the rest of the factories in `drivers/hmis/spec/factories/hmis/hud`? (maybe in another pr)

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
```diff
@@ -22,7 +22,7 @@ class Hmis::Hud::CurrentLivingSituation < Hmis::Hud::Base
   after_commit :warehouse_trigger_processing
 
   private def warehouse_trigger_processing
-    return unless warehouse_columns_changed?
+    return unless enrollment && warehouse_columns_changed?
```
ttoomey, line 25: Question about the change to `warehouse_trigger_processing` on these models: we weren't testing for enrollment before but we were invoking `enrollment.invalidate_processing!` right? So I guess there were cases were it was missing and then this callback would crash?

drivers/hmis/lib/hmis_data_cleanup/util.rb
ttoomey: I'd recommend reverting this change. Using `Hmis::Hud::Enrollment.hmis.update_all(ExportID: nil)` is just one sql update statement. It doesn't get much more performant than that.

I think the only reason to use a technique like this "upsert on PK conflict" is if the values for ExportID were all different (like in `fix_incorrect_personal_id_references` below). Since the value is all the same here, a simple update statement seems much better to me. It's less code/complexity and will be more performant

I'd keep the logging though, that seems good

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,104 @@ def self.clear_enrollment_export_ids!
...
-    # Note: If this gets to be a large number, an upsert is probably worth doing.
-    # We could `scope.find_in_batches` and then enrollment.assign_attributes and finally "import" the batch with a conflict key of id.
     def self.assign_missing_household_ids!
-      scope = Hmis::Hud::Enrollment.hmis.where(household_id: nil)
-      Rails.logger.info "Assigning household id to #{scope.size} enrollments"
-      scope.each do |enrollment|
-        hh_id = Hmis::Hud::Base.generate_uuid
-        without_papertrail_or_timestamps do
-          enrollment.update_columns(household_id: hh_id) # skips callbacks
+      Hmis::Hud::Enrollment.hmis.where(household_id: [nil, '']).find_in_batches do |batch|
```
ttoomey, line 38: Assuming we can just use any random UUID for the initial household id, it seems like this could all be done in the server. This might be preferable if we are trying to make the operation as performant as possible. Maybe something like this?

```ruby
rows_affected = Hmis::Hud::Enrollment.hmis.
  where(household_id: [nil, '']).
  update_all('"HouseholdID" = gen_random_uuid()')
Rails.logger.info "Assigned HouseholdID to #{rows_affected} Enrollments"

```

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -27,43 +37,104 @@ def self.clear_enrollment_export_ids!
...
+            eid_to_pid = GrdaWarehouse::Hud::Enrollment.where(
+              data_source_id: data_source_id,
+              EnrollmentID: batch.map(&:EnrollmentID),
+            ).pluck(:EnrollmentID, :PersonalID).to_h
+
+            values = []
+            batch.each do |record|
+              real_personal_id = eid_to_pid[record.EnrollmentID]
+              next unless real_personal_id # enrollment not found, so we cant update the PersonalID
+              next if record.PersonalID == real_personal_id # this shouldn't be true, but check anyway
```
ttoomey, line 109: Oh I see, this is redundant with the Enrollment.id = nil condition on the scope. Yeah it should be find to remove the extra check, sorry I think I missed that earlier

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+      ds = create(ds_factory)
+      proj = create(:hud_project, data_source_id: ds.id)
+      client = create(:grda_warehouse_hud_client, data_source_id: ds.id)
+      5.times do
+        create(:grda_warehouse_hud_enrollment, data_source_id: ds.id, client: client, project: proj, DateCreated: 1.year.ago, DateUpdated: 1.year.ago)
+      end
+    end
+  end
+
+  def expect_leaves_non_hmis_data_alone(&block)
```
ttoomey, line 34: I don't know that this is testing exactly what is suggested given the name of the method:
* the scopes aren't restricted to non-hmis data source (like where.not(data_source: hmis_ds))
* it might not be reliable to use DateUpdated to detect changes; I believe scope.update_all(...) doesn't change timestamps 

How about something like this where it does a full comparison of all attributes of the models?
```ruby
  def expect_leaves_non_hmis_data_alone(&block)
    expect do
      yield block
    end.to not_change {
      [
        GrdaWarehouse::Hud::Client,
        GrdaWarehouse::Hud::CurrentLivingSituation,
        GrdaWarehouse::Hud::Enrollment,
        GrdaWarehouse::Hud::Service,
      ].flat_map do |scope|
        scope.order(:id).where.not(data_source: hmis_ds).map do |record|
          record.attributes
        end
      end
    }.and not_change(GrdaWarehouse::Version, :count)
  end
```

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -0,0 +1,222 @@
...
+      # none of the utils should set timestamps on any records
+      and change(GrdaWarehouse::Hud::Enrollment.where(DateUpdated: 5.minutes.ago..), :count).by(0).
+      and change(GrdaWarehouse::Hud::Service.where(DateUpdated: 5.minutes.ago..), :count).by(0).
+      and change(GrdaWarehouse::Hud::CurrentLivingSituation.where(DateUpdated: 5.minutes.ago..), :count).by(0).
+      and change(GrdaWarehouse::Hud::Client.where(DateUpdated: 5.minutes.ago..), :count).by(0)
+  end
+
+  context 'clear_enrollment_export_ids' do
+    it 'works' do
+      GrdaWarehouse::Hud::Enrollment.update_all(ExportID: 'XYZ')
```
ttoomey: minor style suggestion for these context blocks: maybe move shared context into a before each block ( also repeated in 'leaves no trace')

```suggestion
  context 'enrollments with export ids' do
    before(:each)  { GrdaWarehouse::Hud::Enrollment.update_all(ExportID: 'XYZ') }
    it 'clears the exports' do
```

ttoomey (approved):
Looks great! Thanks so much for the test improvements 🙌 

I left a bunch of suggestions but I think the only recommendations are:
* back out the change to `clear_enrollment_export_ids`
* adjustment to expect_leaves_non_hmis_data_alone

drivers/hmis/spec/factories/hmis/hud/hmis_base_factory.rb
gigxz: Yes I can do it separately!

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
gigxz: I believe this came up from the factory constructing the object.

I hadn't seen any errors like this come through sentry. This gets invoked after commit, and it wouldn't be possible to edit any of these records in the app because the app would fail to resolve them.

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: that makes perfect sense, thanks, fixed! I think I was concerned about the size of the operating, so thought batching would help. sounds like this should be fine though!

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: I liked @eanders suggestion of making it consistent based on the enrollment ID, because then if we need to run it repeatedly (say, the nightly imports are still going ) then it generates consistent IDs. Not that that really matters, I don't think, but it seems appropriate. I don't think performance on this one is huge because it should be a relatively smaller subset of enrollments.

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
gigxz: Thanks for this suggestion! Definitely wanted to beef it up, I will add the attribute comparison.

 I wanted to check DateUpdated on **everything** too because the cleanup routines should not set timestamps on any of the records that they do touch.

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
gigxz: thank you! always appreciate style suggestions on rspec, somehow nothing I write in rspec ever feels neat.

drivers/hmis/lib/hmis_data_cleanup/util.rb
eanders: Oh, yeah, sorry, I mis-read that and thought we were doing more than just setting all the ExportID to nil.  I agree, we don't even need to batch this, just 
```
Hmis::Hud::Enrollment.hmis.where.not(ExportID: nil).update_all(ExportID: nil)
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
eanders: I _think_ we're trying to be consistent with the new importer extension, which uses the MD5 pattern below.  We could probably figure out how to do that in the DB, but I'm not sure it's worth it.

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: it's updated

eanders (approved):

PR #4193, opened 2024-03-27
Get existing e2e tests running

martha:

PT ticket https://www.pivotaltracker.com/story/show/187307089
This PR
- updates our existing e2e tests that have suffered some decay since they aren't run automatically as part of our CI
- adds some new test coverage
- adds documentation for e2e tests

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/bin/run_hmis_system_tests.sh
+++ b/bin/run_hmis_system_tests.sh
@@ -47,9 +47,10 @@ set -e
 # Clone the specific branch from the repository
 git clone --depth 1 --branch "$BRANCH_NAME" "$REPO_URL" .
 
+CWD=$(pwd)
 yarn config set ignore-engines true
-yarn install
-yarn build
+yarn --cwd $CWD install
+yarn --cwd $CWD build
 
 # unset fails on exit
 set +e
@@ -58,7 +59,7 @@ set +e
 HOSTNAME=`hostname`
 
 # Start in the background
-SERVER_HTTPS=false HMIS_SERVER_URL="http://localhost:4444" HMIS_HOST=$HOSTNAME yarn preview &
+SERVER_HTTPS=false HMIS_SERVER_URL="http://localhost:4444" HMIS_HOST=$HOSTNAME yarn --cwd $CWD preview &
 SERVER_PID=$!
 sleep 5
 
--- a/spec/support/E2E_README.md
+++ b/spec/support/E2E_README.md
@@ -1,0 +1,62 @@
+# E2E Testing
+
+We have a few end-to-end tests located in [drivers/hmis/spec/system/hmis](drivers/hmis/spec/system/hmis).
+We are working on adding more coverage and running them in CI!
+Below are instructions for getting set up to run and write E2E tests locally.
+
+## Background and Testing Philosophy
+
+The goal of these tests is to catch regressions, especially in the UI. The tests use [Capybara](https://github.com/teamcapybara/capybara), which allows us to test our [frontend](https://github.com/greenriver/hmis-frontend/) using our existing factories to mock the backend models. This means these tests are testing both UI and API behavior.
+
+Since E2E tests are expensive -- slow to run and fiddly to write/update -- we should avoid testing parts of the interface where we anticipate major redesign or other churn, and we should avoid using E2E tests to test logic that could be tested more efficiently in another way. For example, we would use an E2E test to verify that form validation messages are displayed appropriately to the user, but we would not write E2E tests exercising all the different validations on every field; this kind of logic should be tested using an API or model test.
+
+## Run E2E tests locally
+
+1. Start the docker container for headless Chrome. This is in our `docker-compose.yml` file, but it's part of the `test` profile, so it isn't started when running `docker-compose up` normally.
+    ```bash
+    docker-compose up chrome
+    ```
+
+2. In another tab, open a Docker shell.
+    ```bash
+    docker-compose run --rm shell  # `dcr shell` if you have that alias
+    ```
+   
+3. In the Docker shell, run the `run_hmis_system_tests` script, indicating the branch of the `hmis-frontend` repo you want to test:
+    ```bash
+    BRANCH_NAME=branch-name-to-test bin/run_hmis_system_tests.sh
+    ```
+
+### Local development
+
+This script is designed to be run as part of a CI pipeline. It clones the frontend repo into a temporary directory, serves the frontend using yarn preview, runs the system tests against that frontend, and then shuts everything down and cleans up after the tests finish running. But when working on E2E tests locally, we don't want to repeatedly clone and clean up the whole repo on each run. Instead,
+
+4. Open a new terminal tab and connect to the same Docker container from step 2. For example,
+    ```bash
+    docker ps # grab the container ID for the Docker shell
+    docker exec -it $CONTAINER_ID /bin/bash # with the container ID from above
+    ```
+   You should be connected to the same Docker container in two tabs: one to serve the frontend, and the other to run the tests.
+   
+5. In the first tab, clone the repo and serve the frontend. You can do this by running the `bin/run_hmis_system_tests.sh` with the following modifications:
+    - comment out the `trap cleanup EXIT` line (we don't want to clean up after exiting the script)
+    - comment out everything after the `yarn preview` line
+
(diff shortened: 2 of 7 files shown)


spec/support/e2e_setup.rb
```diff
@@ -48,6 +55,20 @@ def mui_select(choice, from:)
     find('li[role=option]', text: choice).click
   end
 
+  def mui_table_select(choice, row:, column:)
```
ttoomey, line 49: neat!

spec/support/e2e_setup.rb
```diff
@@ -48,6 +55,20 @@ def mui_select(choice, from:)
...
+    scroll_to(row_label, align: :center)
+    column_label = find('th', text: column)
+    find("[aria-labelledby='#{row_label['id']} #{column_label['id']}']").click
+    find('li[role=option]', text: choice).click
+  end
+
+  def mui_date_select(label, date:)
+    field = find("[aria-label='#{label}']")
+    field.click
+    field.native.send_keys(:left, :left, :backspace, date.strftime('%m/%d/%Y'))
```
ttoomey, line 62: This key sequence seems like a bit of magic. Maybe add a comment?

spec/support/e2e_setup.rb
```diff
@@ -38,6 +38,13 @@ def mui_choose(choice, from:)
     end
   end
 
+  def mui_radio_choose(choice, from:)
```
ttoomey, line 32: This replaces mui_choose? If so maybe delete the old function or comment it out?

ttoomey (commented):
Nice work! I haven't run the tests yet but the code looks good to me. Thanks!

spec/support/e2e_setup.rb
martha: Ah true, you're right, it can be removed

spec/support/e2e_setup.rb
martha: Yes you're right -- sorry, I thought I'd left a PR comment to ask you about this, but I'd left it as a draft review!

I was stuck on this for some time, trying different things, and wasn't able to find another good way to interact with the datepicker. I thought it might be related to this issue [https://github.com/mui/mui-x/issues/4644] but it's not quite the same problem.

spec/support/E2E_README.md
martha, line 1: Is this the right place for this?

spec/support/E2E_README.md
```diff
@@ -0,0 +1,62 @@
+# E2E Testing
+
+We have a few end-to-end tests located in [drivers/hmis/spec/system/hmis](drivers/hmis/spec/system/hmis).
+We are working on adding more coverage and running them in CI!
+Below are instructions for getting set up to run and write E2E tests locally.
+
+## Background and Testing Philosophy
+
+The goal of these tests is to catch regressions, especially in the UI. The tests use [Capybara](https://github.com/teamcapybara/capybara), which allows us to test our [frontend](https://github.com/greenriver/hmis-frontend/) using our existing factories to mock the backend models. This means these tests are testing both UI and API behavior.
```
martha, line 9: Should we add more detail here - for example talking about Cuprite and linking to the Evil Martians document?

spec/support/E2E_README.md
```diff
@@ -0,0 +1,62 @@
...
+    HOSTNAME=`hostname`
+    unset HMIS_OKTA_CLIENT_ID
+    unset OKTA_DOMAIN
+    ```
+    Then, replacing `/*` as needed with the specific test file/line that you want to run,
+    ```bash
+    RUN_SYSTEM_TESTS=true RAILS_ENV=test CAPYBARA_APP_HOST="http://$HOSTNAME:5173" rspec drivers/hmis/spec/system/hmis/*
+    ```
+
+TODO: Should we make a separate script for local development (perhaps reusable by the CI script?)?
```
martha, line 56: Let me know what you think about this

bin/run_hmis_system_tests.sh
```diff
@@ -47,9 +47,10 @@ set -e
 # Clone the specific branch from the repository
 git clone --depth 1 --branch "$BRANCH_NAME" "$REPO_URL" .
 
+CWD=$(pwd)
```
martha, line 50: I spent a bit of time this afternoon trying to debug my yarn config issue, but without real progress. 

**Summary of the issue**
Yarn's `cwd` is always set to `/app`, even after running `yarn config set cwd $(pwd)`. Yarn installs/builds/etc in the `/app` directory instead of the e2e tests' tmp directory tests. We found that adding `--cwd $CWD` to every yarn command is a viable workaround.

This seems to be impacting just me and not @ttoomey, so I'm curious about the experience of other developers.

**What I tried/learned this afternoon**
- I noticed an interesting difference: `yarn config current` shows `cwd = /app` no matter what (reflecting how yarn actually behaves), whereas `yarn config list` shows `cwd` with the new assigned value after I run `yarn config set cwd $(pwd)` (but, again, yarn completely ignores this and continues using `/app` as the working directory). As far as I could tell in [the yarn docs](https://classic.yarnpkg.com/lang/en/docs/cli/config/), `yarn config current` is not documented, so I'm not sure what the difference is.
- I found [this issue](https://github.com/yarnpkg/yarn/issues/8397) and thought my problem might be related, but there has not been further dialog there since 2020. I did post a question on the thread, we'll see if it leads anywhere.
- After reading a little more about the world of Yarn and learning that yarn 1.x is in maintenance-only mode, I tried upgrading to yarn 3.6.4. I was able to do the upgrade, but this problem remains.

I wonder if this has to do with the `hmis-warehouse-yarn` docker container? https://github.com/greenriver/hmis-warehouse/blob/29f8a66c9cbc1cb8f102c30023adc7ae02740b3d/docker/docker-compose.yml#L107

martha: Overall, though, if others agree, I'd be ok with moving on from this curious issue and leaving the current workaround in place, where we set the `--cwd` flag explicitly on every yarn command.

martha:
On the topic of system/UI tests, what's the status of our Cyprus e2e tests mentioned [here](https://github.com/greenriver/hmis-frontend/?tab=readme-ov-file#cypress-e2e-tests) in the frontend readme? Are they still in commission or should that be removed?

bin/run_hmis_system_tests.sh
ttoomey: Yes it's probably not worth spending more time on it since it's a non-blocking issue that only impacts the development env. Thanks for looking into it!

Re yarn upgrade, my understanding is that newer versions of yarn are not backwards compatible, yarn 1 and 2 are essentially different tools. The upgrade path from yarn "classic" isn't clear yet but it hasn't reached EOL so we can kick that can down the road. We may end up just moving to npm though

spec/support/E2E_README.md
ttoomey: Good question. It seems like we have docs living in the top-level `/docs` dir which might be the right place for this. Tagging @eanders - what's the preferred place for developer docs?

Wherever the doc ends up, I'd suggest adding a link to it in a comment at the top of the E2eTests class so it's easier to discover

spec/support/E2E_README.md
ttoomey: Yes I think it would be good to improve the ergonomics of this down the road. Maybe we'll extract the front-end app out of the existing run_hmis_system_tests so it can run independently of the tests, it could probably run as another docker service. Could be a good issue to add to the icebox

spec/support/E2E_README.md
ttoomey: I think inline comments may be best for that kind of thing
* looks like we have that evilmartians article linked at the top of the e2e_tests file 
* for documentation of gem dependencies, I'd suggest annotating the Gemfile

ttoomey (approved):
Looks great, thanks for fixing up the tests and adding documentation!

spec/support/E2E_README.md
eanders: I think where you have it is fine, but maybe a link in the main developer docs in /docs so we don't lose it?

spec/support/E2E_README.md
martha: Ok great, I added https://www.pivotaltracker.com/story/show/187338459

PR #4198, opened 2024-03-28
HMIS links

eanders:

## Description

This:
1. Adds some checks to the links generated for linking off to HMIS to try and ensure someone will be able to access the item when they get there.
2. Standardizes the `to_warehouse` and `to_hmis` methods and makes it so you _can_ call them without needing a DB query

## Type of change


app/models/grda_warehouse/data_source.rb
```diff
@@ -574,6 +574,33 @@ def hmis_url_for(entity)
     # For any other Enrollment-related record, link to the enrollment page
     url ||= "#{base}/client/#{entity.client&.id}/enrollments/#{entity.enrollment&.id}" if entity.respond_to?(:enrollment) && entity.respond_to?(:client)
 
+    # If we don't have the HMIS driver we probably aren't here, but we need to check for the next section
+    # If we don't have a user, just return the URl (backwards compatibility)
+    return url if user.blank? || ! HmisEnforcement.hmis_enabled?
```
gigxz: This is ok but might be nicer to do
`return unless hmis? && HmisEnforcement.hmis_enabled?` on line 557 instead

eanders: Will do!  Thanks.

app/models/grda_warehouse/data_source.rb
```diff
@@ -574,6 +574,33 @@ def hmis_url_for(entity)
...
+      'Hmis::Hud::Enrollment' => [:can_view_projects, :can_view_enrollment_details],
+      'Hmis::Hud::CustomAssessment' => [:can_view_projects, :can_view_enrollment_details],
+    }
+
+    perms = known_permissions[entity.class.name]
+    # If we can't determine if we can see this in HMIS, just go ahead and show the link,
+    # HMIS will handle access
+    return url if perms.blank?
+    # If we can see this in HMIS, don't bother linking to it
+    return nil unless hmis_user.permissions_for?(hmis_entity, perms, mode: :all)
```
gigxz: @eanders this branch isn't working for me I wasn't seeing links anywhere even though I have access. I finally tracked down that this method fails silently if passed a `perms` array... it's supposed to be:

```suggestion
    return nil unless hmis_user.permissions_for?(hmis_entity, *perms, mode: :all)
```

With that fix, at least perms are attempting to resolve, but I get an error:

Screenshot 2024-03-28 at 4 23 23 PM


Sorry that it wasn't so straightforward to figure out the access. I don't mind if this goes into release-110.

app/models/grda_warehouse/data_source.rb
```diff
@@ -574,6 +574,33 @@ def hmis_url_for(entity)
...
+    else
+      entity
+    end
+    hmis_user = user.related_hmis_user(self)
+
+    known_permissions = {
+      'GrdaWarehouse::Hud::Project' => [:can_view_project],
+      # 'GrdaWarehouse::Hud::Organization' => [:can_view_organization],
+      'GrdaWarehouse::Hud::Client' => [:can_view_clients],
+      'Hmis::Hud::Enrollment' => [:can_view_projects, :can_view_enrollment_details],
```
gigxz, line 593:
```suggestion
      'GrdaWarehouse::Hud::Enrollment' => [:can_view_projects, :can_view_enrollment_details],
```
I dont think there is any place we are passing an HMIS enrollment to this? but I could be wrong

gigxz: I'm specifically looking at https://hmis-warehouse.dev.test/source_data/59064?type=Enrollment which is a grdawarehouse model

app/models/grda_warehouse/data_source.rb
gigxz: We have that `persisted?` check there as a regression check measure because we had some sort of bug around permissions on unpersisted records.  Maybe you can trick the model into knowing its persisted in `as_hmis` somehow..

eanders:
@gigxz sorry this was such a mess.  I _think_ it's sorted now.  I've reverted to always loading the associated record form the DB since that seems necessary for the permissions to work, and have, hopefully, cleaned things up a bit.

PR #4203, opened 2024-03-29
Add search filter for Organizations query

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/186494330

This PR adds the capability to add a search term on the Organizations query.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/models/hmis/filter/organization_filter.rb
+++ b/drivers/hmis/app/models/hmis/filter/organization_filter.rb
@@ -1,0 +1,20 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::Filter::OrganizationFilter < Hmis::Filter::BaseFilter
+  def filter_scope(scope)
+    scope = ensure_scope(scope)
+    scope.
+      yield_self(&method(:with_search_term)).
+      yield_self(&method(:clean_scope))
+  end
+
+  protected
+
+  def with_search_term(scope)
+    with_filter(scope, :search_term) { scope.matching_search_term(input.search_term) }
+  end
+end
--- a/drivers/hmis/app/models/hmis/filter/organization_filter.rb
+++ b/drivers/hmis/app/models/hmis/filter/organization_filter.rb
@@ -6,7 +6,8 @@
 
 class Hmis::Filter::OrganizationFilter < Hmis::Filter::BaseFilter
   def filter_scope(scope)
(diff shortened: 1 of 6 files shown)


drivers/hmis/app/models/hmis/hud/organization.rb
```diff
@@ -31,6 +31,14 @@ class Hmis::Hud::Organization < Hmis::Hud::Base
     where(id: ids, data_source_id: user.hmis_data_source_id)
   end
 
+  scope :matching_search_term, ->(search_term) do
+    return none unless search_term.present?
+
+    search_term.strip!
+    query = "%#{search_term.split(/\W+/).join('%')}%"
+    where(o_t[:OrganizationName].matches(query).or(o_t[:id].eq(search_term)).or(o_t[:organization_id].eq(search_term)))
```
martha, line 39: Just to check, should we search by OrganizationCommonName here too? I noticed Projects also has a CommonName field, but doesn't filter by it [here](https://github.com/greenriver/hmis-warehouse/blob/29f8a66c9cbc1cb8f102c30023adc7ae02740b3d/drivers/hmis/app/models/hmis/hud/project.rb#L135), which is why I did not add that here.

gigxz: Good Q, no need to search the common name field. That's from the HUD CSV spec but we don't use it.

drivers/hmis/app/models/hmis/filter/organization_filter.rb
```diff
@@ -0,0 +1,20 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class Hmis::Filter::OrganizationFilter < Hmis::Filter::BaseFilter
+  def filter_scope(scope)
+    scope = ensure_scope(scope)
+    scope.
+      yield_self(&method(:with_search_term)).
+      yield_self(&method(:clean_scope))
```
gigxz: We don't really need this, since the search term scope doesn't do a `join`. It was added in some places to deal with that, because the join would lead to duplicate search results. This is a pretty small table, so the performance impact is probably not noticable, but I would err on the side of removing it since it's not needed.

gigxz (approved):
One minor suggestion, otherwise looks good to me!

drivers/hmis/app/models/hmis/filter/organization_filter.rb
martha: Removing this caused an issue, I'm still investigating

gigxz (approved):

PR #4214, opened 2024-04-02
Add permission for client name and contact info

martha:

## Description

PT issues: 
- Name: https://www.pivotaltracker.com/story/show/186494363
- Contact info: https://www.pivotaltracker.com/story/show/186494368

This PR adds a permission for viewing client names and contact info.

@gigxz as we discussed yesterday - I think there were a lot of distinct places where this needed to be added, and I'm not 100% sure I got all of them, so when you review, please let me know if you think I missed anything! (~For example, I was wondering if it needed to be added [here](https://github.com/greenriver/hmis-warehouse/blob/29f8a66c9cbc1cb8f102c30023adc7ae02740b3d/drivers/hmis/app/graphql/types/hmis_schema/client.rb#L150-L171) on the client graphql schema object's access_field attribute. I think no, because it's not used from the frontend for any queries, but let me know if I am wrong about that?~ Disregard -- I think it will be necessary to implement Elliot's suggestion [here](https://www.pivotaltracker.com/story/show/186494363/comments/240792009))

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/types/application/client_access_summary.rb
+++ b/drivers/hmis/app/graphql/types/application/client_access_summary.rb
@@ -14,6 +14,8 @@ module Types
     field :client_name, String, null: true
 
     def client_name
+      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
+
       client&.brief_name
     end
 
--- a/drivers/hmis/app/graphql/types/application/client_access_summary.rb
+++ b/drivers/hmis/app/graphql/types/application/client_access_summary.rb
@@ -14,8 +14,7 @@ module Types
     field :client_name, String, null: true
 
     def client_name
-      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
-
+      # not checking for name permission access, because can_audit_users is required here
       client&.brief_name
     end
 
--- a/drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
+++ b/drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
@@ -34,8 +34,7 @@ module Types
     end
 
     def client_name
-      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
-
+      # not checking for name permission access, because can_audit_users is required here
       client&.brief_name
     end
 
--- a/drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
+++ b/drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
@@ -34,6 +34,8 @@ module Types
     end
 
     def client_name
+      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
+
       client&.brief_name
     end
 
--- a/drivers/hmis/app/graphql/types/base_audit_event.rb
+++ b/drivers/hmis/app/graphql/types/base_audit_event.rb
@@ -6,7 +6,7 @@
 
 module Types
   class BaseAuditEvent < BaseObject
-    def self.build(node_class, field_permissions: nil, object_permissions: nil, excluded_keys: nil, transform_changes: nil)
+    def self.build(node_class, excluded_keys: nil, transform_changes: nil)
       dynamic_name = "#{node_class.graphql_name}AuditEvent"
       klass = Class.new(self) do
         graphql_name(dynamic_name)
@@ -23,26 +23,6 @@ module Types
           return transform_changes.call(object, changes) if transform_changes.present?
 
           changes
-        end
-
-        define_method(:authorize_field) do |record, key|
-          return true unless field_permissions[key].present?
-
-          # If no record provided, then it's a deleted record, so check general permissions
-          return current_user.permission?(*Array.wrap(field_permissions[key])) unless record
-
-          # Check if user has permission to view audit history for this particular field (for example SSN/DOB on Client)
-          current_user.permissions_for?(record, *Array.wrap(field_permissions[key]))
-        end
-
-        define_method(:authorize_object) do |item_type, record|
-          return true unless object_permissions
-
-          return true unless object_permissions[item_type].present?
-
-          return current_user.permission?(*Array.wrap(object_permissions[item_type])) unless record
-
-          return current_user.permission_for?(record, *Array.wrap(object_permissions[item_type]))
         end
       end
 
@@ -194,12 +174,6 @@ module Types
 
         # Skip if changes are empty, or if the change is `nil=>99` or `99=>nil`. This is not meaningful to show in the UI.
         next if values.map { |v| v == 99 ? nil : v }.compact.empty?
-
-        # hide certain changes (SSN/DOB) if unauthorized
-        values = 'changed' unless authorize_field(changed_record, key)
-
-        # hide all changes on certain objects (address, phone) if unauthorized
-        values = 'changed' unless authorize_object(object.item_type, changed_record)
 
         [
           field_name,
--- a/drivers/hmis/app/graphql/types/base_audit_event.rb
+++ b/drivers/hmis/app/graphql/types/base_audit_event.rb
@@ -28,6 +28,9 @@ module Types
         define_method(:authorize_field) do |record, key|
           return true unless field_permissions[key].present?
 
+          # If no record provided, then it's a deleted record, so check general permissions
+          return current_user.permission?(*Array.wrap(field_permissions[key])) unless record
+
           # Check if user has permission to view audit history for this particular field (for example SSN/DOB on Client)
           current_user.permissions_for?(record, *Array.wrap(field_permissions[key]))
         end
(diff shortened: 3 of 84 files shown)


drivers/hmis/app/graphql/types/application/client_access_summary.rb
```diff
@@ -14,6 +14,8 @@ class Application::ClientAccessSummary < Types::BaseObject
     field :client_name, String, null: true
 
     def client_name
+      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
+
       client&.brief_name
```
gigxz: @eanders @martha I'm a but unsure on this one (General comment, not just for this line).

Should we support scenarios where:
* User CAN **audit clients**, but CANNOT see client [names | dob | ssn | contact info]
* User CAN **audit users**, but CANNOT see client [names | dob | ssn | contact info]

These use-cases both feel highly unlikely to me.
We already have some code to pull DOB/SSN from audit history to support that first case. I wonder if the extra code complexity is worth it. I am leaning towards pulling out this complexity, and instead updating the audit descriptions to say something like:
> This permission allows the user to view changes to any client details, including DOB, SSN, and name, even if they don't have the specific permissions to view those fields on the Client.

Those audit permissions are already marked as **administrative** so this seem pretty reasonable to me. I think I would prefer it just for the sake of simpler code. What do you think? I apologize for not having made that decision together prior to implementing -- now that I see the complexity, I am leaning more towards the simplified approach.


(The complexity around `name` for audit is particularly dicey since we have both the name table and the name fields on the client table. It doesn't seem worth it to me.)

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -220,6 +220,13 @@ def self.permissions_with_descriptions
         category: 'Client Access',
         sub_category: 'Access',
       },
+      can_view_client_name: {
+        description: 'Access to view Client Name.',
```
gigxz:
```suggestion
        description: 'Access to view client names',
```

drivers/hmis/app/graphql/types/application/client_access_summary.rb
eanders: I think that makes sense.  If you are auditing, you can see some raw back-end data, which we don't fully filter.  Alternatively, we could require that if you can audit clients you must also have the other "view" permissions...

eanders, line 18: This `client&.brief_name` made me wonder if instead of hiding the client name completely, we could show the first letter of each name (essentially initials) if you don't have permission to see the client name.  That way we could leave it in the search results, and there would be something to pass along when discussing details of a client.

gigxz: I don't like that approach as much. For gr staff for example, it may be useful to enable auditing for some specific debugging process, but to leave the overall viewing perms turned _off_ to reduce exposure elsewhere in the application.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -216,6 +216,10 @@ def build_primary_custom_client_name
     )
   end
 
+  def masked_name
+    "Client <#{id}>"
+  end
```
eanders: As noted above, I wonder if this should actually be initials....

eanders: Or maybe configurable by site to either be initials _or_ client id as you have now.  I can see situations where you wouldn't want to expose anything, not sure if initials would be sufficient in those cases.

drivers/hmis/app/graphql/types/application/client_access_summary.rb
eanders: Yeah, I agree, let's just make it clear what is being exposed if you have the audit permission.

martha: Okay sounds good, I agree too - as you can see in the contact info permissions [PR](https://github.com/greenriver/hmis-warehouse/pull/4220#discussion_r1548565518), that change added even more complexity to the audit field permissions.

drivers/hmis/app/models/hmis/hud/client.rb
martha: @gigxz what do you think about this? I'm happy to implement it either way. It seems like if the primary use case, for now, is GR devs poking around prod to investigate issues while minimizing exposure, then having the masked name just point to the client ID is enough. But if this ever gets used by non-GR users, then maybe it's more friendly to have a masked name that shows the initials -- it's a bit aesthetically similar to masking all but last 4 digits of ssn.

Also, what would be the best way to implement this as a per-site configurable attribute? A new env var?

drivers/hmis/app/graphql/types/application/client_access_summary.rb
martha: I made this update and just wanted to note that in doing so, I merged in the client contact info permissions (just to avoid a merge conflict) - so this PR now includes new permissions for both name and contact info.

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: Hm lets stick with `Client id` for now. I think this should just be `"Client #{id}"`.

I could see initials being useful in some case, but initials are still PII so I think we should err on the side of privacy. The `id` is still useful when communicating, because users can search for it and see it on the client profile. We should favor communicating with IDs over communicating with initials.


`GrdaWarehouse::Config` class in the warehouse is used to manage "site settings" which can be configured on the warehouse admin page. SO far, we have avoided using it for HMIS, but we probably will need site settings some point. I think that will probably want to be a different table. Def out of scope here though.

drivers/hmis/app/graphql/types/application/client_access_summary.rb
gigxz: sounds good! thanks Martha

drivers/hmis/app/graphql/types/application/client_access_summary.rb
```diff
@@ -14,6 +14,8 @@ class Application::ClientAccessSummary < Types::BaseObject
     field :client_name, String, null: true
 
     def client_name
+      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
+
```
gigxz:
```suggestion
      # not checking for name permission access, because can_audit_users is required here
```

drivers/hmis/app/graphql/types/application/enrollment_access_summary.rb
```diff
@@ -34,6 +34,8 @@ def client_id
     end
 
     def client_name
+      return client&.masked_name unless current_permission?(permission: :can_view_client_name, entity: client)
```
gigxz: Same here re: no need to mask since user has audit permission

drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -180,9 +173,6 @@ def object_changes
         # Skip if changes are empty, or if the change is `nil=>99` or `99=>nil`. This is not meaningful to show in the UI.
         next if values.map { |v| v == 99 ? nil : v }.compact.empty?
 
-        # hide certain changes (SSN/DOB) if unauthorized
-        values = 'changed' if changed_record && !authorize_field(changed_record, key)
```
gigxz, line 184: I looked through environments to confirm that there are no current users who were hitting this behavior. E.g. no users who lack SSN/DOB access but have Audit access. This is safe to remove, and we are updating the description of the Audit perm accordingly. 👍

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -271,7 +268,31 @@ def dob
...
+    def middle_name
+      object.middle_name if current_permission?(permission: :can_view_client_name, entity: object)
+    end
+
+    def last_name
+      object.last_name if current_permission?(permission: :can_view_client_name, entity: object)
+    end
+
+    def name_suffix
+      object.name_suffix if current_permission?(permission: :can_view_client_name, entity: object)
```
gigxz: Style suggestion:
```
    def middle_name
      object.middle_name if can_view_name
    end

   def last_name
      object.last_name if can_view_name
    end

    def names
      # initialize a dummy CustomClientName with masked name
      return [object.names.new(first: object.masked_name)] unless can_view_name
    ...

   private def can_view_name
       current_permission?(permission: :can_view_client_name, entity: object)
   end

```

gigxz (commented):
Looks great!!

drivers/hmis/app/models/hmis/hud/client.rb
eanders: Sounds like a plan.  I hadn't realized that this was all collected into one method when I started that thought, seems like it would be super easy to adjust in the future if we need to.

gigxz (approved):
great! excited to put this into practice with developer roles 🎊

PR #4215, opened 2024-04-02
Utils for deduplicating HouseholdIDs and deleted Enrollments

gigxz:

## Description

* `reassign_duplicate_household_ids!` [story](https://www.pivotaltracker.com/n/projects/2591838/stories/187345937)
* `hard_delete_duplicate_deleted_enrollments!` should resolve [merge bug](https://www.pivotaltracker.com/story/show/187346580) (pending agreement on approach)

## Type of change

### DIFF
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -208,38 +208,6 @@ module HmisDataCleanup
       enrollments_to_delete.each(&:really_destroy!)
     end
 
-    # Transform all-uppercase client names to camel-cased client names
-    # WARNING! Doesn't skips callbacks. Should probably be modified if we ever need to use this again.
-    def self.humanize_client_names!
-      name_fields = [:first_name, :middle_name, :last_name, :name_suffix]
-      scope = Hmis::Hud::Client.hmis.where('"Client"."FirstName" = upper("Client"."FirstName")')
-      # scope = Hmis::Hud::Client.hmis
-      scope.in_batches(of: 5_000) do |batch|
-        Rails.logger.info 'Next batch..'
-        values = []
-        batch.pluck(:id, *name_fields).each do |id_and_names|
-          id = id_and_names.first
-          names_arr = id_and_names.drop(1)
-          names = name_fields.zip(names_arr).to_h.compact_blank
-          next unless names.any?
-
-          # puts "old: #{names.values.join(' ')}"
-          names = names.transform_values { |s| s.downcase.gsub(/(\s+\w)|(^\w)/, &:upcase) }
-          # puts "new: #{names.values.join(' ')}"
-          names[:id] = id
-          values << names
-        end
-
-        grouped_clients = values.index_by { |client| client[:id] }
-
-        # without papertrail so it doesn't show up in Client Audit History
-        without_papertrail_or_timestamps do
-          Hmis::Hud::Client.update(grouped_clients.keys, grouped_clients.values)
-          Rails.logger.info "Updated #{grouped_clients.size} clients"
-        end
-      end
-    end
-
     # Change ProjectID of a project. Useful in staging environment if need a project ID to match prod.
     def self.change_project_id!(old_id, new_id, force: false)
       return if Rails.env.production? && !force
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -205,7 +205,7 @@ module HmisDataCleanup
         end
       end
 
-      enrollments_to_delete.each(&:really_destroy!)
+      GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
     end
 
     # Change ProjectID of a project. Useful in staging environment if need a project ID to match prod.
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -209,7 +209,7 @@ module HmisDataCleanup
         end
 
         Rails.logger.info "Deleting duplicate ids: #{enrollments_to_delete.inspect}"
-        GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
+        GrdaWarehouse::Hud::Enrollment.only_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
       end
     end
 
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -52,6 +52,24 @@ module HmisDataCleanup
       end
     end
 
+    # Reassign household IDs that are duplicated across projects.
+    def self.reassign_duplicate_household_ids!
+      hh_ids = Hmis::Hud::Enrollment.hmis.group(:household_id).
+        having('count(distinct "Enrollment"."ProjectID") > 1').
+        pluck(:HouseholdID)
+
+      hh_id_to_enrollments = Hmis::Hud::Enrollment.hmis.where(household_id: hh_ids).group_by(&:household_id)
+
+      hh_id_to_enrollments.each do |hh_id, enrollments|
+        puts "Processing HouseholdID #{hh_id}..."
+        enrollments.group_by(&:project_id).each do |project_id, enrollments_in_project|
+          id = Hmis::Hud::Base.generate_uuid
+          puts "[HouseholdID #{hh_id}][ProjectID #{project_id}] Reassigning Enrollments with ids: [#{enrollments_in_project.map(&:id).join(',')}] to new HouseholdID (#{id})"
+          enrollments_in_project.each { |en| en.update_columns(household_id: id) }
+        end
+      end
+    end
+
     # Find any single-member households that do not have a HoH, and make that person the HoH
     def self.make_sole_member_hoh!
       single_member_households = Hmis::Hud::Enrollment.hmis.where.not(household_id: nil).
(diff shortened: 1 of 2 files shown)


drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -155,6 +173,41 @@ def self.delete_duplicate_exit_records!
...
+        if enrollments.any? { |e| e.DateDeleted.nil? }
+          enrollments_to_delete.push(*enrollments.select { |e| e.DateDeleted.present? })
+        else
+          # otherwise they are all deleted, so keep 1 of them
+          to_keep = enrollments.max_by(&:DateUpdated)
+          enrollments_to_delete.push(*enrollments.reject { |e| e.id == to_keep.id })
+        end
+      end
+
+      enrollments_to_delete.each(&:really_destroy!)
```
gigxz: this deletes related records too. maybe we should just do `delete`

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -155,6 +173,41 @@ def self.delete_duplicate_exit_records!
...
     end
 
+    # This cleanup routine was added because duplicate EnrollmentIDs were leading to uniqueness violation constraints
+    # when Client records were merged.
+    #
+    # 1) Hard-delete deleted enrollments where there is any NON-deleted Enrollment that shares an EnrollmentID
+    # 2) Hard-delete deleted all but 1 enrollment where there is a set of deleted Enrollments that share an EnrollmentID
+    def self.hard_delete_duplicate_deleted_enrollments!
+      data_source_id = GrdaWarehouse::DataSource.hmis.first.id
+      result = GrdaWarehouse::Hud::Enrollment.with_deleted.
```
eanders, line 183: This seems like a really complex way to do this.  Could we not just do something like?
```
open_enrollments = GrdaWarehouse::Hud::Enrollment.where(data_source_id: data_source_id).pluck(:enrollment_id)
deleted_enrollments = GrdaWarehouse::Hud::Enrollment.only_deleted.where(data_source_id: data_source_id).pluck(:enrollment_id)

duplicates = open_enrollments & deleted_enrollments

GrdaWarehouse::Hud::Enrollment.only_deleted.where(data_source_id: data_source_id, enrollment_id: duplicates).delete_all
```

That's untested, but I _think_ it's roughly equivalent.

gigxz: I think thats equivalent, except for the fact that it will dedup even in cases where the ProjectID doesn't match. (but that could easily be included into your approach).  I haven't seen cases like that (same enrollment ID but diff project id) but it seems like something we'd want to retain.

It also isn't equivalent in that it retains existing duplicates among deleted enrollments. but again, that could be incorporated with this array approach.

I like your approach for readability but can we pluck all the Enrollment IDs in the whole data source?   That seemed like a lot. It's only a one-time thing, so maybe it's fine..

eanders: Yeah, I didn't account for ProjectID, we probably should.  And good point about not checking for duplicates within the deleted records.  

I don't think plucking all the ids will be an issue.

Either approach is fine with me, I was just having some trouble following yours.

gigxz: I was having trouble writing and following it too tbh... I will update to something more readable.

gigxz: I'm having a hard time making this more readable using pluck keys, when project IDs and deduping deleted ones come into play. It feels a lot harder to reason about because we don't have access to the database Ids.



```
      active_enrollments = GrdaWarehouse::Hud::Enrollment.where(data_source_id: data_source_id).pluck(:enrollment_id, :project_id)
      deleted_enrollments = GrdaWarehouse::Hud::Enrollment.only_deleted.where(data_source_id: data_source_id).pluck(:enrollment_id, :project_id)

      duplicates = active_enrollments & deleted_enrollments
      duplicates = duplicates.to_set

      # Hard-delete deleted enrollments where there is any NON-deleted Enrollment that shares the same EnrollmentID+ProjectID
      GrdaWarehouse::Hud::Enrollment.only_deleted.
        where(data_source_id: data_source_id, enrollment_id: duplicates.map(&:first)).
        select { |e| duplicates.include?([e.enrollment_id, e.project_id]) }.
        each(&:really_destroy!)

      # deleted enrollments that are duplicated, and DON'T match existing enrollments
      deleted_enrollments_dups = deleted_enrollments.group_by(&:itself).
        filter_map { |_, v| v if v.size > 1 } - active_enrollments

      # now need to delete only the dups of the above (leave 1)
      # todo
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -155,6 +173,41 @@ def self.delete_duplicate_exit_records!
...
+        if enrollments.any? { |e| e.DateDeleted.nil? }
+          enrollments_to_delete.push(*enrollments.select { |e| e.DateDeleted.present? })
+        else
+          # otherwise they are all deleted, so keep 1 of them
+          to_keep = enrollments.max_by(&:DateUpdated)
+          enrollments_to_delete.push(*enrollments.reject { |e| e.id == to_keep.id })
+        end
+      end
+
+      GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
```
gigxz: this is inefficient but the only way I could figure out how to hard-delete without destroying. `enrollments_to_delete.each(&:delete)` doesn't work.

drivers/hmis/lib/hmis_data_cleanup/util.rb
ttoomey: I don't have a problem with the "non-pluck keys" version you have currently.

My only concern is that it sounded like there are ~10k duplicates; that seems like kind of a lot of enrollments to load into memory. Maybe that's okay but might have to make some adjustments to accommodate that. Seems like we could use batching or something though

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -155,6 +173,41 @@ def self.delete_duplicate_exit_records!
...
+
+        # if any are NOT deleted, hard-delete all the deleted ones
+        if enrollments.any? { |e| e.DateDeleted.nil? }
+          enrollments_to_delete.push(*enrollments.select { |e| e.DateDeleted.present? })
+        else
+          # otherwise they are all deleted, so keep 1 of them
+          to_keep = enrollments.max_by(&:DateUpdated)
+          enrollments_to_delete.push(*enrollments.reject { |e| e.id == to_keep.id })
+        end
+      end
```
ttoomey: It might be necessary to batch this if memory is an issue. Maybe something like this:
```ruby
# untested
result.map(&:enrollment_ids).in_groups_of(100).each do |group|
    enrollments_by_id = Hmis::Hud::Enrollment.with_deleted.
        where(id: group.flat_map(&:enrollment_ids)).index_by(&:id)
    enrollments_to_delete = []
    result.each do |res|
      # ... etc
    end
    GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
 end
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
ttoomey: This is just "delete from Enrollment where id in x,yz" right? That should be fast (much faster than `each(&:delete)`).

It is packing like 10k ids into one sql statement though but that would also be fixed by batching it. 

I also wonder if we should log the ids?
```suggestion
      Rails.logger.info "Deleting duplicate ids: #{enrollments_to_delete.inspect}"
      GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: enrollments_to_delete.map(&:id)).delete_all
```

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -306,4 +309,84 @@ def expect_leaves_non_hmis_data_alone(&block)
       end
     end
   end
+
+  context 'with household ids duplicated across projects' do
+    before(:each) do
+      p2 = create(:hmis_hud_project, data_source: hmis_ds, organization: o1)
+      e1.update(household_id: 'ABCDEF', project: p1)
+      e2.update(household_id: 'ABCDEF', project: p2)
+      e3.update(household_id: 'ABCDEF', project: p2)
```
ttoomey: probably want to get exception if this fails
```suggestion
      e1.update!(household_id: 'ABCDEF', project: p1)
      e2.update!(household_id: 'ABCDEF', project: p2)
      e3.update!(household_id: 'ABCDEF', project: p2)
```

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -306,4 +309,84 @@ def expect_leaves_non_hmis_data_alone(&block)
...
+  context 'with duplicate deleted EnrollmentIDs' do
+    let(:dup_key) { 'ABCDEF' }
+    before(:each) do
+      e1.update(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p1)
+      e2.update(EnrollmentID: dup_key, DateDeleted: nil, project: p1)
+      e3.update(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p1)
+
+      # duplicate in a different project
+      p2 = create(:hmis_hud_project, data_source: hmis_ds, organization: o1)
+      e4.update(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p2)
```
ttoomey: probably want to get exception if this fails
```suggestion
      e1.update!(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p1)
      e2.update!(EnrollmentID: dup_key, DateDeleted: nil, project: p1)
      e3.update!(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p1)

      # duplicate in a different project
      p2 = create(:hmis_hud_project, data_source: hmis_ds, organization: o1)
      e4.update!(EnrollmentID: dup_key, DateDeleted: 1.week.ago, project: p2)
```

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -306,4 +309,84 @@ def expect_leaves_non_hmis_data_alone(&block)
...
+          change { Hmis::Hud::Enrollment.hmis.only_deleted.where(EnrollmentID: dup_key).count }.to(1), # the one in another project
+        ].reduce(&:and),
+      )
+
+      expect(GrdaWarehouse::Hud::Enrollment.with_deleted.where(id: [e1.id, e3.id])).to be_empty
+      expect(e2.reload).to be_present
+    end
+
+    it 'keeps 1 deleted enrollment if ALL the dups are deleted' do
+      e2.update(DateDeleted: 2.days.ago)
```
ttoomey:
```suggestion
      e2.update!(DateDeleted: 2.days.ago)
```

ttoomey (approved):
Looks good to me, thanks! It's too bad we can't just adjust the unique constraint on Enrollment to ignore deleted records.

eanders (approved):

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
gigxz: whoops thanks

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: yeah, added logging

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: thanks, updated to groups of 500

PR #4219, opened 2024-04-02
TC only import new assessments

gigxz:

## Description

So that we can import missing records without `clobber`.

## Type of change

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
@@ -115,7 +115,7 @@ module HmisExternalApis::TcHmis::Importers::Loaders
         end
 
         assessment_hud_key = row_assessment_id(row)
-        if @existing_assessment_hud_keys.include?(assessment_hud_key)
+        if existing_assessment_hud_keys.include?(assessment_hud_key)
           # should only happen if clobber:false
           log_info "Skipping CustomAssessmentID that has already been imported: #{assessment_hud_key}"
           next
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
@@ -113,6 +113,14 @@ module HmisExternalApis::TcHmis::Importers::Loaders
           log_skipped_row(row, field: ENROLLMENT_ID_COL)
           next # early return
         end
+
+        assessment_hud_key = row_assessment_id(row)
+        if @existing_assessment_hud_keys.include?(assessment_hud_key)
+          # should only happen if clobber:false
+          log_info "Skipping CustomAssessmentID that has already been imported: #{assessment_hud_key}"
+          next
+        end
+
         actual += 1
 
         assessment_date = row_assessment_date(row)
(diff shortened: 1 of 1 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
```diff
@@ -113,14 +113,22 @@ def create_assessment_records(rows)
           log_skipped_row(row, field: ENROLLMENT_ID_COL)
           next # early return
         end
+
+        assessment_hud_key = row_assessment_id(row)
+        if @existing_assessment_hud_keys.include?(assessment_hud_key)
+          # should only happen if clobber:false
```
eanders, line 119: I'm missing something, but is it that `@existing_assessment_hud_keys` will be empty if called with `clobber: true`?  That's probably outside the scope of the pr.

gigxz: yes, the `clobber` option deletes any existing assessments that match the excel file. so the `existing_assessment_hud_keys` set should not contain any keys that overlap with the excel file, in that case.

I wasn't too worried about it since we do not plan to run this with `clobber` ever again

eanders (approved):

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
```diff
@@ -113,14 +113,22 @@ def create_assessment_records(rows)
           log_skipped_row(row, field: ENROLLMENT_ID_COL)
           next # early return
         end
+
+        assessment_hud_key = row_assessment_id(row)
+        if @existing_assessment_hud_keys.include?(assessment_hud_key)
```
ttoomey: I might be missing something but I think this instance variable will not be defined at this point; I don't see where we are calling the `existing_assessment_hud_keys()` function that would build the instance variable. 

It seems like the intention of `@existing_assessment_hud_keys` is to memoize the value of `existing_assessment_hud_keys()` - in that case I would suggest not using the instance variable directly (the ivar seems like more of a side-effect/function internal). Instead suggest always calling the `existing_assessment_hud_keys()` function

```suggestion
        if existing_assessment_hud_keys.include?(assessment_hud_key)
```

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
```diff
@@ -153,6 +161,7 @@ def create_form_processor_records(rows)
       rows.each do |row|
         assessment_id = owner_id_by_row_id[row_assessment_id(row)]
         next unless assessment_id
+        next if @existing_assessment_hud_keys.include?(assessment_hud_key) # already imported
```
ttoomey: probably:
```suggestion
        next if existing_assessment_hud_keys.include?(assessment_hud_key) # already imported
```

ttoomey (approved):
Looks good, I think there's one small bug to fix, see note. Should be good to merge afterwards. Thanks!

drivers/hmis_external_apis/app/models/hmis_external_apis/tc_hmis/importers/loaders/custom_assessment_loader.rb
gigxz: whoops, totally my intention to use the method name, thanks!

PR #4220, opened 2024-04-02
Add permission for client contact

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/186494368

This adds a new permission for restricting access to client contact info. I also added a little extra groundwork around the audit event permission parsing - will put more detail in an inline comment. 

I'm opening this PR against the Client Name permission PR (#4214) because I built it on top of that PR to avoid merge conflicts.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/types/base_audit_event.rb
+++ b/drivers/hmis/app/graphql/types/base_audit_event.rb
@@ -6,7 +6,7 @@
 
 module Types
   class BaseAuditEvent < BaseObject
-    def self.build(node_class, field_permissions: nil, excluded_keys: nil, transform_changes: nil)
+    def self.build(node_class, field_permissions: nil, object_permissions: nil, excluded_keys: nil, transform_changes: nil)
       dynamic_name = "#{node_class.graphql_name}AuditEvent"
       klass = Class.new(self) do
         graphql_name(dynamic_name)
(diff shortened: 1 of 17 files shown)


drivers/hmis/app/graphql/types/base_audit_event.rb
```diff
@@ -34,6 +34,16 @@ def self.build(node_class, field_permissions: nil, excluded_keys: nil, transform
           # Check if user has permission to view audit history for this particular field (for example SSN/DOB on Client)
           current_user.permissions_for?(record, *Array.wrap(field_permissions[key]))
         end
+
+        define_method(:authorize_object) do |item_type, record|
```
martha, line 38: This enables us to hide all fields on a given item type, rather than making the determination key-by-key. Previously, we only needed key-by-key masking because the only keys we masked out were DOB and SSN. Now that we are masking out contact info, this is coming from the generic `CustomClientContactPoint` object, and has keys like `system` and `value` -- and I felt nervous that there might be overlap with keys in some other object that we would not want to mask. (This also enables the removal of the `CustomClientName` fields from `FIELD_PERMISSIONS_FOR_AUDIT` in drivers/hmis/app/graphql/types/hmis_schema/client.rb.) Please let me know if this addition looks ok!

PR #4223, opened 2024-04-03
Throw error if bulk assign causes overlapping enrollments

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/187288068

This PR fixes the bug where a user could end up with overlapping enrollments through the new bulk services interface. It also adds a regression test.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
+++ b/drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
@@ -54,7 +54,7 @@ module Mutations
             raise 'bulk service assignment generated invalid enrollment' unless enrollment.valid?
 
             entry_date_errors = Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date(enrollment)
-            raise HmisErrors::ApiError, entry_date_errors.first.full_message unless entry_date_errors.empty?
+            error_out(entry_date_errors.first.full_message) unless entry_date_errors.empty?
 
             # Attempt to assign this enrollment to a unit if this project has units. This is AC-specific for now, and does
             # not support specifying the unit type. Needs improvement if/when we expand unit capabilities.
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -53,7 +53,8 @@ def resolve(input:)
             access_denied! unless can_enroll_clients
             raise 'bulk service assignment generated invalid enrollment' unless enrollment.valid?
 
-            # TODO(#187288068) warn about overlapping enrollments using Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date
+            entry_date_errors = Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date(enrollment)
+            raise HmisErrors::ApiError, entry_date_errors.first.full_message unless entry_date_errors.empty?
```
gigxz: This has a pretty unintuitive API, you have to pass `display_message` like [this](https://github.com/greenriver/hmis-warehouse/blob/fb5fe917564ac0ea99e453c6b451f035731cce39/drivers/hmis/app/graphql/mutations/bulk_assign_service.rb#L105) if you want the message to actually be visible to the user in production, which we do in this case. You can mimic production behavior(sort of) by adjusting this code locally to test: https://github.com/greenriver/hmis-warehouse/blob/fb5fe917564ac0ea99e453c6b451f035731cce39/drivers/hmis/app/controllers/hmis/graphql_controller.rb#L103

martha: Ah thank you! It looks like there is already a helper `error_out` that does this, so I changed it to use that.

Screenshot 2024-04-03 at 5 00 31 PM

gigxz: nice, thanks!

PR #4224, opened 2024-04-04
Rails 7 upgrade

ttoomey:

## Description

[Rails upgrade ](https://github.com/open-path/Green-River/issues/20)
#### Dependencies / Notes
* use version 4 of devise-two-factor that should be compatible with rails 7 to defer having to migrate encrypted fields
* Sticking with HAML v5 for now. It seems like V6 is not safe given our usage
* kept rgeo at old version to avoid build errors in CI
* monkey patch active record to work around deadlock in tests with multiple dbs in rails 7.0

#### Class loader issues
* fix "HasRecentItems" so it uses strings to reference class names, avoid loader problems
* [use custom rails_drivers branch I'm hacking for compat with rails 7
](https://github.com/greenriver/rails_drivers/tree/rails-7)
* rename/move some classes around to match zeitwerk requirements
* fix classes that double-include acts_as_paranoid (it's already included in parent)

#### Deprecation
* we will continue to use the now-deprecated date.to_s and time.to_s. [See issue](https://github.com/open-path/Green-River/issues/4897)
* change date.to_s(fmt) to date.to_fs(fmt)
* Faker::IDNumber -> Faker::IdNumber
* s/request.content_type/media_type/

#### Housekeeping
* rm docker_fs_fix, I don't think it's needed anymore

## Type of change
- [x] Dependency update


lib/docker_fs_fix.rb
```diff
@@ -1,15 +0,0 @@
-# https://github.com/docker/for-linux/issues/1015
-
-# puts "Patching https://github.com/docker/for-linux/issues/1015"
-module FileUtilsPatch
```
ttoomey, line 4: probably not necessary now

drivers/public_reports/config/initializers/public_reports_feature.rb
```diff
@@ -12,42 +12,44 @@
...
-    external_url: "#{GrdaWarehouse::Help::DEFAULT_HELP_URL}/Homeless-Populations-Report-Generator",
-  },
-  {
-    controller_path: 'public_reports/warehouse_reports/state_level_homelessness',
-    action_name: 'index',
-    external_url: "#{GrdaWarehouse::Help::DEFAULT_HELP_URL}/State-Level-Homelessness-Report-Generator",
-  },
-].each do |report|
-  Rails.application.config.help_links << report
+Rails.application.config.to_prepare do
```
ttoomey, line 15: wrap this, not supposed to reference classes like GrdaWarehouse in the initializer context

drivers/hmis/app/models/hmis/form/instance_enrollment_match.rb
```diff
@@ -6,6 +6,7 @@
 
 # determine if enrollment is a valid match for a form intance
 #
+require 'memery'
```
ttoomey, line 9: for some reason we have 'require: false' in the gemfile and then require memery when needed. Not sure why...

config/routes.rb
```diff
@@ -7,9 +7,6 @@
   match "/422", to: "errors#unacceptable", via: :all
   match "/500", to: "errors#internal_server_error", via: :all
 
-  require 'rails_drivers/routes'
-  RailsDrivers::Routes.load_driver_routes
```
ttoomey, line 11: this is now part of the rails_driver railtie in the GR rails-7 branch

config/initializers/time.rb
```diff
@@ -1,4 +1,5 @@
 # Rails.logger.debug "Running initializer in #{__FILE__}"
 
-Date::DATE_FORMATS[:default] = '%b %e, %Y'
-Time::DATE_FORMATS[:default]= '%b %e, %Y %l:%M %P'
```
ttoomey: This behavior is deprecated. Probably should use rails i18n instead

config/environments/production.rb
```diff
@@ -93,7 +93,7 @@
 
   cache_ssl = (ENV.fetch('CACHE_SSL') { 'false' }) == 'true'
   cache_namespace = "#{ENV.fetch('CLIENT')}-#{Rails.env}-hmis"
-  redis_config = Rails.application.config_for(:cache_store).merge({ expires_in: 8.hours, raise_errors: false, ssl: cache_ssl, namespace: cache_namespace })
+  redis_config = Rails.application.config_for(:cache_store).merge({ expires_in: 8.hours, race_condition_ttl: 1.minute, ssl: cache_ssl, namespace: cache_namespace })
```
ttoomey, line 98: `raise_errors` not valid option anymore. Adding `race_condition_ttl` per recommendations in documentation

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2024/dq/question_three.rb
```diff
@@ -57,7 +57,7 @@ def q3_veteran_row(sheet, universe_members)
       )
 
       total_cell = sheet.update_cell_value(cell: 'E2', value: [dkpntr_cell, missing_cell, issue_cell].map(&:value).sum)
-      total_cell.add_members([dkpntr_cell, missing_cell, issue_cell].map(&:members).sum)
+      total_cell.add_members([dkpntr_cell, missing_cell, issue_cell].map(&:members).sum([]))
```
ttoomey, line 60: using sum() on arrays with non-numeric values is deprecated unless an initial value is provided

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,141 @@
...
+Rails.application.config.action_view.apply_stylesheet_media_default = false
+
+# Change the digest class for the key generators to `OpenSSL::Digest::SHA256`.
+# Changing this default means invalidate all encrypted messages generated by
+# your application and, all the encrypted cookies. Only change this after you
+# rotated all the messages using the key rotator.
+#
+# See upgrading guide for more information on how to build a rotator.
+# https://guides.rubyonrails.org/v7.0/upgrading_ruby_on_rails.html
+# Rails.application.config.active_support.key_generator_hash_digest_class = OpenSSL::Digest::SHA256
```
ttoomey, line 26: This change might impact active storage but we don't use it, right?

config/initializers/new_framework_defaults_7_0.rb
```diff
@@ -0,0 +1,141 @@
...
+
+# Change the digest class for ActiveSupport::Digest.
+# Changing this default means that for example Etags change and
+# various cache keys leading to cache invalidation.
+# Rails.application.config.active_support.hash_digest_class = OpenSSL::Digest::SHA256
```
ttoomey, line 31: It's a little unclear to me what the impact of this change outside of etags. Maybe session cookies?

Gemfile
```diff
@@ -69,10 +70,7 @@ gem 'image_processing'
 gem 'ruby-filemagic' unless ENV['SKIP_FILEMAGIC'].to_s == 'true'
 gem 'mini_magick'
 gem 'mimemagic'
-# there are no obvious breaking changes but
-# since there are no tests for this
-# it should be tested manually
-gem 'acts-as-taggable-on', '~> 7.0'
+gem 'acts-as-taggable-on', '~> 9.0.1'
```
eanders, line 76: Making a note, we should double check this is working.  We use very little of the official API, but it is embedded fairly deeply in client file uploads.

app/controllers/health/careplans_controller.rb
```diff
@@ -57,7 +57,7 @@ def index
 
     def show
       pdf = careplan_combine_pdf_object
-      file_name = "care_plan_#{@careplan.updated_at.to_s(:db)}"
+      file_name = "care_plan_#{@careplan.updated_at.to_fs(:db)}"
```
eanders, line 60: @ttoomey I'm slightly confused by [this deprecation note](https://guides.rubyonrails.org/7_0_release_notes.html#active-support-deprecations).  Is `to_fs(:db)` acceptable going forward or is there additional cleanup we need to do with this?

app/models/concerns/hud_concerns/enrollment.rb
```diff
@@ -49,7 +49,9 @@ def self.invalidate_processing!
 
     def invalidate_processing!
       # use update_columns to avoid stale object errors when this method is called from Service after_commit hooks
-      update_columns(processed_as: nil, processed_hash: nil)
+      self.class.without_optimistic_locking do
```
eanders, line 52: This feels right here.  The invalidate processing shouldn't fail just because the record is stale.  Thanks.

config/environments/development.rb
```diff
@@ -1,4 +1,5 @@
 I18n.config.available_locales = :en
+require 'active_support/core_ext/integer/time'
```
eanders, line 2: Do we need this in other environments?

Gemfile
ttoomey: Thanks, would it be practical to write a test for this behavior? I could do that potentially

app/controllers/health/careplans_controller.rb
ttoomey: Yes, my understanding is `date_or_time.to_fs(format)` is what we should use going forward. It's the override of ruby's native `to_s` that is deprecated.

app/models/concerns/hud_concerns/enrollment.rb
ttoomey: Yes, unfortunately it's not an exception, it's actually a no-op (`Update Enrollments set.... where lock_version <= {stale_version}`)

config/initializers/cookie_rotator.rb
eanders, line 1: @ttoomey is my reading correct, in that this will rotate us from `:marshal` based cookies to `:json` cookies?  In theory, we only need to run this for a day or two right since our sessions are so short?  Or is there a situation where a cookie might live much longer, so better to give it plenty of time to upgrade them?

drivers/hmis/app/graphql/hmis_schema.rb
```diff
@@ -11,6 +11,8 @@ class HmisSchema < GraphQL::Schema
   trace_with(GraphqlTraceBehavior)
 
   # For batch-loading (see https://graphql-ruby.org/dataloader/overview.html)
+  # - after upgrade to Rails 7.1 we could replace Dataloader with AsyncDataloader for performance
+  #   https://graphql-ruby.org/dataloader/async_dataloader
```
eanders, line 15: Should we add a TodoOrDie here that checks for rails 7.1?  Or if this is just for performance, maybe this is enough?

drivers/hmis/app/models/hmis/form/instance_enrollment_match.rb
eanders: Historically, we've tried to work that way with some gems to decrease app boot time.  Not sure it actually helps though.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2024/dq/question_three.rb
eanders: Is this solved with a `compact`?  or are we getting strings and things back?

eanders (commented):
Whoops, just noticed I never actually submitted my comments, adding those now, continuing to review.

eanders:
@ttoomey I think the `Date.to_s` problem is going to be a problem.  Just loading any page throws hundreds of deprecation warnings to slack. [See this slack channel.](https://greenriver.slack.com/archives/C031PSXMW5Q)  
Some of them are from things like:
```
row.values_at(*@keys)
```

Which are going to be really hard to fine, and even harder to fix.  Seems like there must be a different way around this.

config/initializers/cookie_rotator.rb
ttoomey: This migrates formats but it also changes encryption (rails 7 uses sha256 now). It seems unnecessary to leave this in place for longer than the max cookie lifetime. However I was unsure if we had other cookies that might have longer lifetimes than session.

Looking at this again, it seems were were already using the json cookie format, we can just that config now since the default is json

drivers/hmis/app/graphql/hmis_schema.rb
ttoomey: My understanding is it's just performance related. Probably not worth halting CI for it but I'll put an issue in the backlog to track it

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2024/dq/question_three.rb
ttoomey: it looks like we are  getting arrays back here which requires a default arg

ttoomey:
> I think the Date.to_s problem is going to be a problem. Just loading any page throws hundreds of deprecation warnings to slack. Which are going to be really hard to find, and even harder to fix. Seems like there must be a different way around this.

Thanks @eanders - this was a first stab at the problem, we have options. I think we discussed this before but we could simply extract rails old to_s behavior and keep using it as-is; we could choose not to address this deprecation at this time. That may be the least burdensome way to go.

However I have some ideas about the notification messages:
* We have two consumers of deprecation events, one for our slack notifier and one for sentry. The sentry notifier has more aggressive debounces to avoid this kind of flood. Perhaps it would be sufficient to just rely on Sentry and we could remove the slack notification for deprecations?
* The warnings in the slack channel are all for `ApplicationController#cache_grda_warehouse_base_queries` which does make it hard to track down. I guess the deprecation payload[:callstack] ends here for some reason (maybe the cache interacts with the callstack)? I don't recall seeing the equivalent of `cache_grda_warehouse_base_queries` in other applications. I guess the intention is to share sql query results across requests. I'd be interested to know more about this cache and perhaps consider if it's still necessary.

Maybe we can check in on this next week

>Some of them are from things like: row.values_at(*@keys)

That's odd, maybe that's an old deprecation? Could also be it's writing dates to csv and converting them to strings, which would trigger it

ttoomey:
adjusted to continue using our custom default formatting with date/time to_s. See config/initializers/legacy_rails_conversions.rb

app/models/grda_warehouse/census/project_batch.rb
```diff
@@ -40,7 +40,9 @@ def build_census_batch
                 (inventory.InventoryEndDate.blank?)) ||
             ((inventory.InformationDate.blank? && inventory.InventoryStartDate.present? && inventory.InventoryStartDate < date) &&
                 (inventory.InventoryEndDate.present? && inventory.InventoryEndDate > date))
-          end.sum(&:beds) rescue 0
+          end
+          # preserve behavior of returning 0 if any bed values are nil (rather than compact.sum)
+          census_item.beds = filtered_beds.all?(&:present?) ? filtered_beds.sum : 0
```
ttoomey, line 45: @eanders - I'm guessing that "beds" is sometimes nil here which is triggering the deprecation warning in sentry. I assume the rescue was there to recover from trying to add nil values. I replaced it with a conditional which seems equivalent.

eanders: Thank you!

PR #4159, opened 2024-03-21
Expose HMIS custom assessments on the client dashboard

eanders:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

This exposes HMIS custom assessments on the client dashboard in the warehouse.  These are added as rows to the Assessments card and have a link to the assessment in the HMIS as well as a modal that shows a very basic read-only view of the assessment in the same manner as ETO assessments or other similar items have been shown in the past.

Assessments are grouped by name with the most-recent by AssessmentDate visible, previous assessments are hidden behind the +.

### Custom Assessment in HMIS
Screenshot 2024-03-21 at 6 24 02 PM

### Client Dashboard in the Warehouse
Screenshot 2024-03-21 at 6 21 31 PM

### Client Dashboard expanded
Screenshot 2024-03-21 at 6 21 23 PM

### Assessment modal in the warehouse (note it integrates some CE Assessment information at the top if a CE assessment is attached.)
Screenshot 2024-03-21 at 6 20 19 PM

## Type of change
[//]: # 'remove options that are not relevant'
- [x] New feature (adds functionality)

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4230, opened 2024-04-04
Show custom services in warehouse UI

martha:

## Description

PT issue: https://www.pivotaltracker.com/story/show/186985555
To show a more holistic picture of the client, we want to add the details of custom services rendered to the client's enrollment details page in the warehouse.

This is a first pass based on my understanding and digging around from the guidelines outlined in the ticket. I'll add a few specific questions I have inline.

Screenshot 2024-04-04 at 4 52 37 PM
Screenshot 2024-04-04 at 4 53 08 PM

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
```diff
@@ -12,6 +12,7 @@ module ClientExtension
       has_many :custom_client_addresses, **Hmis::Hud::Base.hmis_relation(:PersonalID, 'CustomClientAddress'), inverse_of: :client
       has_many :hmis_custom_assessments, through: :enrollments
       has_many :hmis_source_custom_assessments, through: :source_enrollments, source: :hmis_custom_assessments
+      has_many :custom_services, through: :source_enrollments
```
martha, line 16: Elliot's [PR for custom assessments](https://github.com/greenriver/hmis-warehouse/pull/4159/files#diff-736cbdcc0e0094b4fd523aff48ebfb121f27f1e3c09ae5c1e32653760059b561) linked in the PT ticket includes a new controller. Should we also add a new controller for custom services?

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+        %th
+    %tbody
+      - custom_services.each do |service|
+        %tr
+          %td
+            .ds{class: "ds-color-#{service.data_source_id}"}= ds_short_name_for(service.client.id)
+          %td= service.project.name
+          %td= service.enrollment.entry_date
+          %td= service.date_provided
+          %td= service.service_type.custom_service_category.name == service.service_name ? service.service_name : "#{service.service_type.custom_service_category.name} - #{service.service_name}"
```
martha: This unwieldy line is meant to match the HMIS frontend's behavior [here](https://github.com/greenriver/hmis-frontend/blob/325fa54119f0346a9a520318303acda99eaf82e5/src/modules/enrollment/components/dashboardPages/EnrollmentServicesPage.tsx#L30-L32), but I wanted to check and confirm that it's a good idea/worth the complexity?

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+          %td.text-right
+            - if @client.hmis_source_visible_by?(current_user)
+              = link_to 'HMIS', source_datum_path(service.enrollment.id, type: 'Enrollment'), class: 'btn btn-xs btn-secondary btn-muted btn-hmis mb-2'
```
martha: Does this link look correct? I copied/modified from the [Income Benefit](https://github.com/greenriver/hmis-warehouse/blob/042bd5b6f97fc572ba4c4888b5e013db80ac0785/app/views/clients/rollup/_income_benefits.html.haml#L38) and [Enrollment table](https://github.com/greenriver/hmis-warehouse/blob/042bd5b6f97fc572ba4c4888b5e013db80ac0785/app/views/clients/_enrollment_table.haml#L134) examples. I was wondering if we want it to link directly to the Services page of the Enrollments dashboard.

drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
martha, line 1: @gigxz one thing you wrote about in the PR that I didn't understand and wanted to ask about:
> You can use the link on the Client Profile ID card in the HMIS to link over to the warehouse profile from HMIS.

I must be missing something silly but couldn't find this link?

drivers/hmis/app/views/hmis_client/services/_table.haml
eanders: Can that be added as a method on `service`?  I'm a big fan of fat models, thin views.

martha: Thanks, that's much better! 😄

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,13 @@
+:ruby
+  custom_services = @client.custom_services.
+    joins(enrollment: :client).
+    joins(custom_service_type: :custom_service_category).
+    merge(GrdaWarehouse::Hud::Enrollment.visible_to(current_user)).
+    preload( :enrollment, :project, :client, :custom_service_type ).
+    preload(custom_service_type: :custom_service_category).
+    order(date_provided: :desc)
```
eanders: Maybe this should be a method on Client (or a class method on CustomService) something like `services_for_rollup` This is a lot of logic to hide in the view.

drivers/hmis/app/views/hmis_client/services/_table.haml
eanders: Hmm, so that's going to link to the enrollment.  Usually we'd link to the object in question (the service in this case) but the point of these links is generally so you can see ALL of the source data on the record.   I wonder if it would be better to just link off to the services page in the HMIS in this case.  @gigxz, thoughts on that?

drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
eanders: I don't think we need a controller in this case.

drivers/hmis/app/views/hmis_client/services/_table.haml
gigxz: The most helpful place is probably `/client/:id/enrollments/:id/services` because that's where you can edit the service records.

drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
gigxz: Oh I was just mentioning this. no action needed, just helpful for working on HMIS+Warehouse
Screenshot 2024-04-05 at 9 30 36 AM

app/views/clients/rollup/_custom_services.html.haml
gigxz: Suggest adding a secondary sort on `:id` to make sure it is always sorted consistently

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,13 @@
...
+- if custom_services.any?
+  = render 'hmis_client/services/table', custom_services: custom_services
```
gigxz: I don't see much benefit of rendering another partial here as opposed to just writing the table view directly in this file. It seems like we do that for many other rollups. But that might just be me, it is always hard for me to find referenced views in my IDE.

drivers/hmis/app/models/hmis/hud/custom_service.rb
```diff
@@ -38,6 +38,12 @@ def within_range?(range)
     date_provided.between?(range.begin, range.end)
   end
 
+  def display_name
+    return service_name if service_type.custom_service_category.name == service_name
+
+    "#{service_type.custom_service_category.name} - #{service_name}"
+  end
```
gigxz: I think we should probably use `service_type.name` instead of `service_name` because we don't keep that field up to date.

drivers/hmis/app/models/hmis/hud/custom_service.rb
```diff
@@ -38,6 +38,12 @@ def within_range?(range)
     date_provided.between?(range.begin, range.end)
   end
 
+  def display_name
+    return service_name if service_type.custom_service_category.name == service_name
```
gigxz: Could add a `has_one :custom_service_category, through: :custom_service_type` to reference the category directly

app/views/clients/rollup/_custom_services.html.haml
gigxz: Oh I see, its because its in the driver. Even harder to find 😬  but I appreciate sticking to a pattern, so its OK with me.

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,13 @@
+:ruby
+  custom_services = @client.custom_services.
+    joins(enrollment: :client).
+    joins(custom_service_type: :custom_service_category).
+    merge(GrdaWarehouse::Hud::Enrollment.visible_to(current_user)).
+    preload( :enrollment, :project, :client, :custom_service_type ).
+    preload(custom_service_type: :custom_service_category).
```
gigxz: I don't think it affects the sql at all, but you can do this in one line:
```
preload(enrollment: [:project, :client], custom_service_type: [:custom_service_category])
```

I don't know if we really need to preload the client though..it seems silly since we are already on the client dashboard. It looks like it's used for the data source display somehow? I will leave a comment down there :)

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+        %th Project
+        %th Entry Date
+        %th Service Date Provided
+        %th Service Name
+        %th
+    %tbody
+      - custom_services.each do |service|
+        %tr
+          %td
+            .ds{class: "ds-color-#{service.data_source_id}"}= ds_short_name_for(service.client.id)
```
gigxz, line 15: 1) @eanders do you think we should have this DS column or no? I added it to the spec for consistency but I could go either way
2) we preload the client just so we can do this `ds_short_name_for` thing. Seems a bit unnecessary to me, since we already know the data source id, but I'm not sure what all its doing.

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,13 @@
+:ruby
+  custom_services = @client.custom_services.
+    joins(enrollment: :client).
+    joins(custom_service_type: :custom_service_category).
```
gigxz: I think this join can be removed

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,13 @@
+:ruby
+  custom_services = @client.custom_services.
+    joins(enrollment: :client).
```
gigxz: Does this need to join client? I think we only need enrollment in order to do the merge with the visible enrollment scope.

drivers/hmis/app/views/hmis_client/services/_table.haml
eanders: I think the consistency is worth it on this page.

gigxz: Okay. I forgot that that data source pill thing shows the PersonalID and other info when you hover on it, so this makes sense to me now.

drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
martha: Of course 🤦‍♀️ thanks!

drivers/hmis/app/views/hmis_client/services/_table.haml
martha: I agree that's probably the most helpful place to link to here. I'm not sure I've done the linking correctly -- I couldn't find a similar example except in the function `hmis_url_for` on the data source, so I kinda-sorta copied that. Let me know if there is a better way or somewhere else I can look to for a better example.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1721,6 +1721,13 @@ def date_of_last_homeless_service
       processed_service_history&.last_homeless_date
     end
 
+    def services_for_rollup
```
eanders, line 1717: Thank you!

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+          %td= service.project.name
+          %td= service.enrollment.entry_date
+          %td= service.date_provided
+          %td= service.display_name
+          %td.text-right
+            - if @client.hmis_source_visible_by?(current_user)
+              = link_to 'HMIS', "https://#{service.enrollment.data_source.hmis}/client/#{service.enrollment.client.id}/enrollments/#{service.enrollment.id}/services", class: 'btn btn-xs btn-secondary btn-muted btn-hmis mb-2'
```
eanders: I _think_ we want to use `service.data_source.hmis_url_for(service, user: current_user)`  The `hmis_url_for` method will need a case statement for the service, but that should be pretty straight forward with what you have.

martha: Ah wonderful. Thanks!

drivers/hmis/app/views/hmis_client/services/_table.haml
martha: I think the data source pill would only have a tooltip if we add this:
https://github.com/greenriver/hmis-warehouse/blob/042bd5b6f97fc572ba4c4888b5e013db80ac0785/app/views/clients/_enrollment_table.haml#L51

It doesn't seem like the other tables in the Enrollment Details tab have that, so I didn't add it here. Should I add it? Should the other tables get it as well?

eanders: I don't think we need the tooltips on the enrollment details page.  Thanks for checking!

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,7 @@
+:ruby
+  custom_services = @client.services_for_rollup.merge(GrdaWarehouse::Hud::Enrollment.visible_to(current_user))
```
gigxz: @eanders all the other rollups like this use a cache like this https://github.com/greenriver/hmis-warehouse/blob/5a6075ace45ed8bb213b5d33fcc4c86f92d139a0/app/views/clients/rollup/_income_benefits.html.haml#L1-L9

Shouldn't we use the same behavior here?

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+        %th Entry Date
+        %th Service Date Provided
+        %th Service Name
+        %th
+    %tbody
+      - custom_services.each do |service|
+        %tr
+          %td
+            .ds{class: "ds-color-#{service.data_source_id}"}= ds_short_name_for(service.client.id)
+          %td= service.project.name
```
gigxz: I think this needs to be `service.project.as_warehouse.name(current_user)` to respect confidential projects, right @eanders? That might also benefit from some caching to store a map of projects, though, since `as_warehouse` does a lookup currently.

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,7 @@
+:ruby
```
gigxz: Screenshot 2024-04-10 at 11 26 09 AM

@eanders I pushed a fix for Employment Education since those labels were wrong. Just wanted to point out that the Service table is so inconsitent with the others. ("Locations" insetad of "Project", different column order, etc). I didn't touch it though since it looked like it was done relatively recently.

I also pushed relabeling some of the CUstom Service table headers for consistency @martha

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,27 @@
...
+          %td= service.enrollment.entry_date
+          %td= service.date_provided
+          %td= service.display_name
+          %td.text-right
+            - if @client.hmis_source_visible_by?(current_user)
+              = link_to 'HMIS', service.data_source.hmis_url_for(service, user: current_user), class: 'btn btn-xs btn-secondary btn-muted btn-hmis mb-2'
```
gigxz: I think this should be the "Open in HMIS" link style,. since that is always what we use to link to HMIS. This button style usually links to a different warehouse page.

drivers/hmis/app/views/hmis_client/services/_table.haml
eanders: That sounds correct to me, and also too bad that it will cause N+1s.  I guess we could cache the answer to `as_warehouse` so at least we would only have to pull it once per project.

app/views/clients/rollup/_custom_services.html.haml
eanders: Thanks @gigxz !  I think we should probably align them all, though as long as the headers match the data below, that's ok for now.

app/views/clients/rollup/_custom_services.html.haml
eanders: yes, definitely.  Thanks for catching that.

martha: @gigxz thanks for pointing that out! Does this look correct? I was not sure exactly what to put in the cache keys array. I was especially wondering about adding `can_view_enrollment_details?` (copying from `can_view_projects?`) in the example you linked.

drivers/hmis/app/views/hmis_client/services/_table.haml
martha: Something along these lines?

app/views/clients/rollup/_custom_services.html.haml
```diff
@@ -0,0 +1,8 @@
+= cache([@client, 'enrollments', can_view_enrollment_details?, controller_path, current_user], expires_in: 1.hours) do
```
gigxz: I'm not sure about what needs to be in this cache key @eanders  can you take a look?

I think it should say something unique/descriptive like `'custom_services'` instead of `'enrollments'`.

drivers/hmis/app/views/hmis_client/services/_table.haml
```diff
@@ -0,0 +1,31 @@
...
+        %th Service Type
+        %th
+    %tbody
+      - custom_services.each do |service|
+        %tr
+          %td
+            .ds{class: "ds-color-#{service.data_source_id}"}= ds_short_name_for(service.client.id)
+          %td
+            = cache(service.project.as_warehouse) do
+              = service.project.as_warehouse.name(current_user)
```
gigxz: The thing we want to cache is the `as_warehouse` lookup, so that lookup should be in the yielded block, not the cache key. I think that defeats the purpose.   https://apidock.com/rails/ActionView/Helpers/CacheHelper/cache

@eanders what do you recommend for railsy way to do the cache key?

maybe something like `[service, current_user, can_view_confidential_project_names?], expires_in: 4.hours` ?

app/views/clients/rollup/_custom_services.html.haml
gigxz: I'm not totally sure,  tagged @eanders with a question below

app/views/clients/rollup/_custom_services.html.haml
eanders: pretty sure that should be 
```suggestion
= cache([@client, 'custom_services', can_view_enrollment_details?, controller_path, current_user], expires_in: 1.hours) do
```

drivers/hmis/app/views/hmis_client/services/_table.haml
eanders: Hmm, I'm not really a fan of the partial caching, I wonder if we can just add memost to the `as_warehouse` method.  Should be fine to do the lookup once per page per project right?  I think the goal is to not look up the `GrdaWarehouse::Hud::Project` record for every service.

I need to think on this one a bit more.

gigxz: 2 other ideas:
* add an extension for warehouse_project, and have the rollup method `preload(:warehouse_project)`. and this call `service.warehouse_project.name(user)`. Extensions are at `drivers/hmis/extensions/grda_warehouse/hud`
* move confidential name logic to a shared concern, so we can call `service.project.confidentialized_warehouse_name(user)` or whatnot

eanders: I like option 1, but option 2 would be fine with me as well.  Thanks @gigxz !

martha: Thank you both! @gigxz could you expand on the reason to put it in an extension rather than just putting it on the model? Also, since the model we are extending is (I think) `Hmis::Hud::CustomService`, would the extension go in `drivers/hmis_external_apis/extensions/hmis/hud/` (rather than `drivers/hmis/extensions/grda_warehouse/hud`)?

eanders: @martha, pretty sure that should be a relation on `GrdaWarehouse::Hud::Project`.  Though maybe it would be cleaner on service, in which case, yeah, I think it would just go on the model.

app/views/clients/rollup/_custom_services.html.haml
martha: Is there a good way to test the caching behavior other than just playing around with it? (e.g. how graphql apollo has dev tools where you can check out what's in the cache?)

For example, I noticed while testing the confidential project names (below thread) that when I updated a project to be confidential, I could still see the name until I cleared out this cache, after which it showed up as 'Confidential Project' which I think is the intended behavior. I was wondering if that means we need to add Projects to this cache key, or if that is such an infrequent use case that it's not worthwhile.

drivers/hmis/app/views/hmis_client/services/_table.haml
martha: Ah so, if it should be an extension on Project rather than CustomService, then instead of `service.warehouse_project.name(user)` as Gig wrote above, it would be `service.project.warehouse_project.name(user)`?

gigxz: Yeah, my comment about extensions didn't make much sense, now that I think about it 👀  my bad. I agree with @eanders  just throw a `has_one :warehouse_project, class_name: 'GrdaWarehouse::Hud::Project',, through: :client_project` on the CustomService model. I _think_ that should let you do `Hmis::Hud::CustomService.last.warehouse_project` -> should return an instance of `GrdaWarehouse::Hud::Project`.

We only use extensions when we want to add HMIS-related stuff to the GrdaWarehouse models.

martha: Okay! Makes sense, thank you.

drivers/client_access_control/app/views/client_access_control/clients/enrollment_details/_enrollment_rollups.haml
```diff
@@ -15,6 +15,9 @@
     %article.rollup.o-dashboard__block.o-dashboard__block--secondary{data: {partial: 'services-all'}}
       %header
         %h1#full_services.h3 Services
+    %article.rollup.o-dashboard__block.o-dashboard__block--secondary{data: {partial: 'custom_services'}}
+      %header
+        %h1#full_services.h3 Custom Services
```
gigxz: This should probably be wrapped in an HMIS check, so its only shown if this warehouse installation has an accompanying HMIS.

```suggestion
    - if HmisEnforcement.hmis_enabled?
      %article.rollup.o-dashboard__block.o-dashboard__block--secondary{data: {partial: 'custom_services'}}
        %header
          %h1#full_services.h3 Custom Services
```

This would still show the "Custom Services" section for a non-HMIS Client in an installation that does support HMIS, but I think that's okay.

gigxz (approved):
Looks good to me!
One last comment about `HmisEnforcement` below.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1721,6 +1721,14 @@ def date_of_last_homeless_service
...
+      custom_services.
+        preload(enrollment: [:project, :client], custom_service_type: [:custom_service_category]).
+        preload(:warehouse_project).
```
eanders: This is probably ok, not sure if preload just added to previous calls or if it merges them, but maybe:
```suggestion
        preload(:warehouse_project, enrollment: [:project, :client], custom_service_type: [:custom_service_category]).
```
just to be safe.

eanders: As an alternative the the above, and the need to add the relation to the custom service class, it looks like you _could_ just 
```
preload(enrollment: [:client, {project: :warehouse_project], custom_service_type: [:custom_service_category]).
```

I'm not sure that's any better, and you'd have to refer to `enrollment.project.warehouse_project.name(current_user)` to get the appropriate name, but it would be one less cross-over between HMIS & warehouse namespaces.  
I'm happy as it is, just a thought.

eanders (approved):
@martha this looks great to me.  Thanks for all the iteration!

martha:
Thank you both for your feedback and your patience! :)

app/models/grda_warehouse/hud/client.rb
martha: I committed your suggestion, but would be happy to make this change as well if you think it's preferable to have one less crossover!

eanders (approved):
Looks fantastic.

PR #4231, opened 2024-04-05
App config & external forms fixes

ttoomey:

## Description
* fix a few bugs for external form jobs
* change test filenames to correspond to the respective job classes
* add error handling around `ConsumeExternalFormSubmissionsJob` so that the hourly task can continue if errors occur
* adds a generic key value store for application config (`AppConfigProperty`)
* replaces hard-coded/env form config with aforementioned db-backed config

## Type of change
- [x] Bug fix


drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/config.rb
```diff
@@ -5,24 +5,24 @@
...
 
 module HmisExternalApis::ExternalForms
-  Config = Struct.new(:site_title, :site_logo_url, :site_logo_dimensions, :recaptcha_key, :presign_url, keyword_init: true) do
+  class Config
+    attr_reader :site_title, :site_logo_url, :site_logo_dimensions, :recaptcha_key, :presign_url, :csp_content
+
+    def initialize
+      # This could be dryed up if we repeat this pattern
+      attrs = [:site_title, :site_logo_url, :site_logo_dimensions, :recaptcha_key, :presign_url, :csp_content]
+      attr_keys = attrs.map { |attr| "external_forms/#{attr}" }
```
eanders: This is probably a better way to do it, but I noticed you made the value column a `jsonb`, so you _could_ store the config as one key in a hash.  Do you think we'll ever want more than one `external_forms` implementation?  Would it make sense to add another variable to the key name to specify _which_ external form this is for?

eanders (approved):

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/config.rb
gigxz: +1 We will definitely need to support multiple external forms with different configurations.

I wondered too about using a prefix `external_forms/` approach as opposed to just one key with json.

eanders: If we need to support multiple external forms, I guess I'd lean on renaming this file to represent which form it backs, something like `intake.rb` and then the keys could be `external_forms/intake/...`.  Just a thought though, happy to entertain other ideas.

ttoomey: Thanks for the feedback! So a couple of things here:
* My intention is that this configuration is common for all external forms (there's only one lambda, logo, captcha key, etc). Each form still has it's own form definition though. Maybe there are use-cases where that wouldn't work; can you help me understand what per-form configuration is needed that we can't just represent in the form definition?
* My reason for using json to store the config props was to provide a convenient typing for the values (strings, numbers, arrays, etc rather than needing to store/manage a value_type on the property).
* Although you could pack all the external form settings as one hash with this schema, I think it's more manageable to make them distinct properties (although perhaps this depends on how they will be managed). I think that if I'm changing the site logo, it doesn't really make sense for me to be updating the recaptcha key or lamdba url at the same time. Keeping them separate means we have more granular timestamps, auditing, and could support access control mechanisms (if we wanted to allow end-users to manage a subset of these properties).

gigxz: I was wondering about `site_title` specifically - is that like the HTML title? i would think it should match the name of the form, rather than being the same across all sites.

ttoomey: Yes, we currently use `site_title` as the HTML title. That's a good point, I agree it would be better to use the form definition title; it should be easy to adjust the layout to make that change. However we shouldn't need form-specific config properties for that since we can just use the definition.

gigxz: Ok, that makes sense. In that case I think they can share the rest of these configurations!

gigxz (approved):
thanks!

ttoomey:
Great, thanks! I'll make that change to the form definition and retarget this to release-111 assuming we'll deploy that to Tarrant prod next week.

gigxz (approved):

PR #4247, opened 2024-04-12
187373375 improve wip enrollments perf

ttoomey:

## Description
[Issue](https://github.com/open-path/Green-River/issues/5699)
Refactor how HMIS handles 'wip' project/enrollment relationship by adding a direct relationship between enrollments and projects. Review and testing needed to ensure this change doesn't introduce regressions, particularly around inclusion of WIP enrollments.

### Details
* User facing change: Project household view should be more performant
* replace conditional code and extra associations with simpler project/enrollment relation
* project.enrollment and project.households now include WIP enrollments
* simplify hmis_households view, add direct relationship to households
* remove hmis_client_projects view
* enhance project/enrollment validation
* move `belongs_to enrollment` to ClientProjectEnrollmentRelated mixin
* disables achmis importer specs

### Testing/ deployment

A data migration is needed to populate the new association. The HMIS is non-functional until this is run:
```
rails driver:hmis:data_migration:update_project_pk
```

__Question:__ should we adjust this for multi-stage deployment to reduce disruption? This branch could be split so that the association could be deployed and populated first.  Then the code change could be deployed after.

## Type of change
- [x] Performance fix


db/warehouse/migrate/20240409215111_add_project_to_enrollment.rb
```diff
@@ -0,0 +1,7 @@
+class AddProjectToEnrollment < ActiveRecord::Migration[6.1]
+  def change
+    safety_assured do
+      add_reference :Enrollment, :actual_project, foreign_key: { to_table: :Project, name: 'fk_rails_enrollment_actual_project_di' }
```
ttoomey: Maybe better to call this `project_pk` instead of `actual_project_id`

eanders: I think I prefer `project_pk`  It'll make it harder to mistake it for `project_id`

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -49,12 +49,12 @@ def load_open_enrollment_for_client(client, project_id:, open_on_date:)
...
       client,
       :enrollments,
-      scope: Hmis::Hud::Enrollment.viewable_by(current_user).preload(:client_project, :exit),
+      scope: Hmis::Hud::Enrollment.viewable_by(current_user).preload(:exit),
     )
 
     # Filter down by project and date
     enrollments.filter do |en|
-      en.open_on_date?(open_on_date) && en.client_project.project_id.to_s == project_id.to_s
+      en.open_on_date?(open_on_date) && en.actual_project_id.to_s == project_id.to_s
```
eanders: wouldn't this now reference the `project.id` instead of `project_id`  something like;
```suggestion
      en.open_on_date?(open_on_date) && en.actual_project_id == project.id
```

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -169,7 +163,7 @@ class << self
     hud_org_ids = tuples.map(&:second)
     raise 'orgs are in multiple data sources' if ds_ids.size > 1
 
-    joins(:projects_including_wip).where(p_t[:organization_id].in(hud_org_ids).and(p_t[:data_source_id].eq(ds_ids.first)))
+    joins(:projects).where(p_t[:organization_id].in(hud_org_ids).and(p_t[:data_source_id].eq(ds_ids.first)))
   end
```
eanders, line 165: I'm probably missing something, but this looks like a more complicated version of:

```
joins(projects: :organization).merge(Hmis::Hud::Organization.where(id: organization_ids))
```

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
```diff
@@ -110,7 +110,8 @@ def self.hmis_validate(record, role: nil, **_)
 
   def validate(record)
     super(record) do
-      record.errors.add :project_id, :required if record.project_id.nil? && record.wip.nil?
+      # enrollment.project_id should match the project.project_id. The rails association is through actual_project_id
+      record.errors.add :project_id, 'must match project pk' if record.ProjectID && record.ProjectID != record.project.ProjectID
```
eanders: Is the error string correct?  I feel like the error is catching something different.
```
'must match project pk'
```

drivers/hmis/lib/tasks/data_migration.rake
```diff
@@ -0,0 +1,30 @@
...
+      UPDATE "Enrollment"
+        SET actual_project_id = "Project".id
+      FROM "Project"
+        WHERE "Project"."data_source_id" = "Enrollment"."data_source_id"
+        AND "Project"."ProjectID" = "Enrollment"."ProjectID"
+        AND "Enrollment"."data_source_id" = #{data_source.id}
+    SQL
+    connection.execute('VACUUM ANALYZE "Enrollment"')
+  end
+end
```
eanders: Probably worth using the `TaskQueue` for this one.  See `drivers/hmis_csv_twenty_twenty_four/config/initializers/hmis_csv_twenty_twenty_four_feature.rb` for an example.  Basically, the site is aware of a queue of one-time run tasks (usually data migrations).  They get run as part of the hourly task if they haven't been run before.

eanders (commented):
Looks generally good I think.  Thanks!

drivers/hmis/lib/tasks/data_migration.rake
ttoomey: Thanks @eanders, I'm not familiar with the TaskQueue mechanism, good to know it's there. If this branch were deployed as-is, the HMIS would be non-functional until this data migration runs. So it may not be desirable to wait for cron.

However we may want to use a two-step deployment for this, where we add and populate the new association field in one deploy, then roll out the relationship change in another. But that's extra work. I'll check with Gig next week

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
ttoomey: Oh yeah, I think I adjusted the condition here; I'll fix up the description to match

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
ttoomey: Hmm I don't think project is defined in this graphql helper method. Also the cast to string appears to be needed by the active_enrollment resolver on the client graphql type.

drivers/hmis/app/models/hmis/hud/client.rb
ttoomey: Good catch! It does look like it could be simplified. @gigxz - does that seem right to you?

drivers/hmis/spec/support/ac_hmis_loader_helpers.rb
```diff
@@ -4,9 +4,13 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
-# frozen_string_literal: true
-
 module AcHmisLoaderHelpers
+  def self.included(base)
```
ttoomey, line 8: disable broken tests. Should be deleted in the near future but that probably should happen in another branch

drivers/hmis/app/graphql/mutations/submit_form.rb
```diff
@@ -162,7 +162,7 @@ def _resolve(input:, record_lock_version: nil)
       when 'Hmis::Hud::Funder', 'Hmis::Hud::ProjectCoc', 'Hmis::Hud::Inventory', 'Hmis::Hud::CeParticipation', 'Hmis::Hud::HmisParticipation'
         [project, klass.new({ project_id: project&.project_id, **ds })]
       when 'Hmis::Hud::Enrollment'
-        [project, klass.new({ project_id: project&.project_id, personal_id: client&.personal_id, **ds })]
+        [project, klass.new({ project_pk: project&.id, personal_id: client&.personal_id, **ds })]
```
ttoomey: only assigning the project_pk will create a WIP enrollment initially. That's probably what we want?

drivers/hmis/app/models/hmis/hud/assessment.rb
```diff
@@ -20,7 +20,6 @@ class Hmis::Hud::Assessment < Hmis::Hud::Base
     date_updated: 'Last Updated: Most Recent First',
   }.freeze
 
-  belongs_to :enrollment, **hmis_enrollment_relation, optional: true
```
ttoomey, line 23: moved to ClientProjectEnrollmentRelated because order matters for `has_many :through`

drivers/hmis/app/models/hmis/wip.rb
```diff
@@ -5,20 +5,6 @@
 ###
 
 class Hmis::Wip < Hmis::HmisBase
+  TodoOrDie('Remove Hmis::WIP and db table', by: '2024-10-01')
```
ttoomey, line 8: left model in place to remind us to remove the table

drivers/hmis/lib/tasks/data_migration.rake
```diff
@@ -0,0 +1,31 @@
+namespace :data_migration do
+  # rails driver:hmis:data_migration:update_project_pk
+  desc 'One time task to set Enrollment.project_pk'
+  task :update_project_pk, [] => [:environment] do
```
ttoomey, line 4: I haven't tested this well, probably could use additional tests

drivers/hmis/app/graphql/mutations/submit_form.rb
gigxz: That logic of whether it should be saved as WIP or non-WIP is handled by this method:
https://github.com/greenriver/hmis-warehouse/blob/85d50fadd0639c95d25824bdd5657ddc32782761/drivers/hmis/app/models/hmis/hud/enrollment.rb#L291-L302

So I think here you can set both ProjectID and project_id, and `ProjectID` will get nulled out by  `save_in_progress!` if its called.

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
```diff
@@ -42,9 +42,6 @@ def perform(force: false)
             joins(:project).
             pluck(e_t[:id], p_t[:id])
```
gigxz: This could be simplified to the below, right? Since we don't need to join anymore, we don't need the `unscoped` piece either.
```
project_id_map  = Hmis::Hud::Enrollment.with_deleted.where(id: enrollment_ids).pluck(:id, :project_pk).to_h
```

drivers/hmis/lib/tasks/data_migration.rake
```diff
@@ -0,0 +1,31 @@
...
+      UPDATE "Enrollment"
+        SET project_pk = "hmis_wips".project_id
+      FROM "hmis_wips"
+        WHERE "hmis_wips"."source_id" = "Enrollment"."id"
+        AND "hmis_wips"."source_type" = 'Hmis::Hud::Enrollment'
+        AND "Enrollment"."data_source_id" = #{data_source.id}
+        AND "Enrollment"."ProjectID" IS NULL
+    SQL
+
+    puts 'updating enrollments from project'
```
gigxz, line 20: Is this going to fail on the foreign key constraint if there are Enrollments that have ProjectIDs that cant be resolved? If so we should add some code to address those, if they exist. I have a feeling that exists, among deleted records, but I haven't checked.

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: yep Elliot's suggestion seems correct

drivers/hmis/app/models/hmis/hud/client_access_loader.rb
```diff
@@ -14,17 +14,16 @@ def fetch(items)
     client_ids = items.map { |i| i.first.id }.compact.uniq
 
     group_view_t = Hmis::GroupViewableEntity.arel_table
-    client_project_t = Hmis::Hud::ClientProject.arel_table
     orphan_client_ids = Hmis::Hud::Client.
-      left_outer_joins(:client_projects).
+      left_outer_joins(:enrollments).
       where(id: client_ids).
-      where(client_project_t[:project_id].eq(nil)).
+      where(arel.e_t[:project_id].eq(nil)).
```
gigxz: I don't think this one is correct. `orphan_client_ids` should represent "unenrolled clients", as the comment below indicates ("client is not associated with a project").

We want clients that have NO enrollments, which I believe is:
```suggestion
      where(arel.e_t[:id].eq(nil)).
```

We should probably add a test to verify this behavior (if this didn't cause a failure) but feel free to bump that to a new issue if its getting to be too much scope-creep.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -115,11 +123,8 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
   scope :with_access, ->(user, *permissions, **kwargs) do
     return none unless user.permissions?(*permissions)
 
-    project_ids = Hmis::Hud::Project.with_access(user, *permissions, **kwargs).pluck(:id, :ProjectID)
-    viewable_wip = wip_t[:project_id].in(project_ids.map(&:first))
-    viewable_enrollment = e_t[:ProjectID].in(project_ids.map(&:second))
-
-    left_outer_joins(:wip).where(viewable_wip.or(viewable_enrollment))
+    project_ids = Hmis::Hud::Project.with_access(user, *permissions, **kwargs).order(:id).pluck(:id)
```
gigxz, line 126: just wondering, why `order` before pluck?

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -131,14 +136,11 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
...
+    full_access_project_ids = Hmis::Hud::Project.with_access(user, :can_view_enrollment_details, :can_view_project, mode: 'all').pluck(:id)
     # Projects where the user has limited enrollment access
-    limited_access_project_ids = Hmis::Hud::Project.with_access(user, :can_view_limited_enrollment_details).pluck(:id, :ProjectID)
+    limited_access_project_ids = Hmis::Hud::Project.with_access(user, :can_view_limited_enrollment_details).pluck(:id)
 
-    viewable_wip = wip_t[:project_id].in(full_access_project_ids.map(&:first) + limited_access_project_ids.map(&:first))
-    viewable_enrollment = e_t[:ProjectID].in(full_access_project_ids.map(&:second) + limited_access_project_ids.map(&:second))
-
-    left_outer_joins(:wip).where(viewable_wip.or(viewable_enrollment))
+    where(project_pk: (full_access_project_ids + limited_access_project_ids).uniq.sort)
```
gigxz, line 143: much simpler!! 👍 

I have the same question about the sorting- perhaps the db has an easier time if the list of ids is sorted in db order, which would make sense.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -242,10 +238,6 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
   after_create :warehouse_trigger_processing
   after_update :warehouse_trigger_processing
 
-  def project
-    super || Hmis::Hud::Project.find_by(id: wip.project_id)
-  end
```
gigxz, line 247: woohooo

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -302,34 +294,20 @@ def save_new_enrollment!
   end
 
   def save_in_progress!
-    # "WIP" Enrollments have a null ProjectID column, so that they don't
-    # get pulled into reporting. The Project database ID is stored in the hmis_wips table.
-    saved_project_id = project.id
-
+    # "WIP" Enrollments have a null ProjectID column, so that they aren't included in reporting.
     self.project_id = nil
```
gigxz, line 300: Should we raise an error here if `project_pk` is unset? It's not expected, but we should guard against it here so we don't lose both the project links

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -302,34 +294,20 @@ def save_new_enrollment!
...
-      save!
-    end
+    self.project_id = project.project_id
+    save!
   end
   alias save_not_in_progress save_not_in_progress!
 
   def in_progress?
-    project_id.nil?
+    read_attribute(:ProjectID).nil?
```
gigxz: Whats the difference here?

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -33,54 +33,29 @@ class Hmis::Hud::Project < Hmis::Hud::Base
...
-      where(Hmis::Hud::Enrollment.arel_table[:project_id].eq(project_id)).
-      or(Hmis::Hud::CustomAssessment.joins(:enrollment).left_outer_joins(enrollment: :wip).
-        where(Hmis::Wip.arel_table[:project_id].eq(id)))
-  end
+  has_many :households, foreign_key: :project_pk, inverse_of: :project
+  has_many :custom_assessments, through: :enrollments
+  has_many :services, through: :enrollments
+  has_many :custom_services, through: :enrollments
+  has_many :clients, through: :enrollments
+  has_many :hmis_services, through: :enrollments
```
gigxz, line 53: awesome

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
```diff
@@ -110,7 +110,13 @@ def self.hmis_validate(record, role: nil, **_)
 
   def validate(record)
     super(record) do
-      record.errors.add :project_id, :required if record.project_id.nil? && record.wip.nil?
+      if record.project
+        # enrollment.project_id should match the project.project_id. The rails association is through project_pk
+        record.errors.add :project_id, 'reference does not match DB PK' if record.ProjectID && record.ProjectID != record.project.ProjectID
+        record.errors.add :project_id, 'must match enrollment data source' if record.project.data_source_id != record.data_source_id
+      else
+        record.errors.add :project_id, 'must be present'
```
gigxz: use the default message for a required field
```suggestion
        record.errors.add :project_id, :required
```

gigxz: although kinda confused about this part, we would enter this case if `project_pk` is nil or invalid. So really the error message should say that `project_pk` is required, right?

gigxz:
This looks great @ttoomey!

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: It's a habit I have to try and keep the query consistent every time. The ids are likely to be in the same order most of the time anyway but using an order guarantees it.  My understanding is that if the query text is the same, postgres can use a cached query plan so it's a bit of a nano-optimization.

I suppose there could also be a benefit for rails application level query caching by ensuring consistent queries.

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: Good idea! We could also enforce this with a database level constraint on Enrollment. We could add that after the data migration is complete

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: Just trying to distinguish it from `project.id`, which would go through project_pk and always be true. Maybe it's better as it was though... idk

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
ttoomey: good catch, I'll change it to use project_pk. I just wanted something to validate that the project was present

drivers/hmis/lib/tasks/data_migration.rake
ttoomey: I don't think it will fail, but `Enrollment.project_pk` would just be NULL if the ProjectID is not valid

drivers/hmis/app/models/hmis/hud/client_access_loader.rb
ttoomey: Great catch! This doesn't seem to be covered by current tests. Actually I'm not quite sure where to add a test for this as the root level client resolver uses viewable_by which catches this before the access loader. Maybe we need unit tests for the access loaders

drivers/hmis/app/jobs/hmis/activity_log_processor_job.rb
ttoomey: Nice! Thanks!

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Oh interesting, thanks

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Okay I see. I think its fine I was just wondering if `read_attribute` had some impact like not reading from local changes on the record or something like that.  Fine with this, or `ProjectID.nil?` to emphasize the casing..

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
```diff
@@ -14,7 +14,6 @@ class Hmis::Hud::CurrentLivingSituation < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::ServiceHistoryQueuer
   include ::Hmis::Hud::Concerns::HasCustomDataElements
 
-  belongs_to :enrollment, **hmis_enrollment_relation, optional: true
```
eanders, line 17: Why are you removing these? It's probably fine, but these (CLS, Disability, Employment Education, etc.) are all tied to enrollments.  Maybe I just missed the concern that's now getting pulled in somewhere?

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -302,34 +294,22 @@ def save_new_enrollment!
...
   end
   alias save_in_progress save_in_progress!
 
   def save_not_in_progress!
-    transaction do
-      self.project_id = project_id || project.project_id
-      wip&.destroy!
-      save!
-    end
+    self.project_id = project.project_id
```
eanders, line 306: maybe reuse this method from above?
```suggestion
    set_hud_project_id_from_project_pk
```

eanders (approved):
Thanks @ttoomey !  Looks good to me, dropped a couple questions inline.

gigxz (approved):
looks good! as discussed lets put this in release-113 so we have a little more time to QA.

gigxz:
@eanders mind if I cut 113 already?

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
ttoomey: It was moved into the ClientProjectEnrollmentRelated mixin. It was necessary to move the enrollment association before the project association as it's now `has_one :project, through: :enrollment` and order matters in that case

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: set_hud_project_id_from_project_pk won't change project_id (wip) status so probably not quite right here

gigxz (changes requested):
Test failures `uninitialized constant Hmis::Hud::Enrollment::ProjectID`

ttoomey:
> Test failures `uninitialized constant Hmis::Hud::Enrollment::ProjectID`

@gigxz Thanks, should be fixed now

PR #4250, opened 2024-04-12
Add VA Dashboard

dtgreiner:

## Description

Add new dashboard for the VA implementation that utilizes a new implied consent model.

## Type of change


app/models/consent/default.rb
```diff
@@ -0,0 +1,65 @@
...
+    'None on file'
+  end
+
+  def revoked_consent_string
+    ''
+  end
+
+  def full_consent_string
+    ''
+  end
```
eanders: What's the difference between `full_consent_string` and `full_release_string`?  I _think_ those should be two words for the same thing.

app/models/consent/default.rb
```diff
@@ -0,0 +1,65 @@
...
+    'Limited CAS Release'
+  end
+
+  def full_release_string
+    'Full HAN Release'
+  end
+
+  def release_current_status
+    consent_text = if @client.housing_release_status.blank?
+      @client.class.no_release_string
```
eanders: Should this use the local `no_release_string` method?

app/models/consent/implied.rb
```diff
@@ -0,0 +1,76 @@
...
+
+  def full_release_string
+    'Expanded Consent'
+  end
+
+  def release_current_status
+    # TODO: COMPLETE THIS
+    consent_text = @client.class.no_release_string
+    consent_text = @client.class.full_consent_string if current_consent_type == full_release_string
+    consent_text = @client.class.revoked_consent_string if current_consent_type == revoked_consent_string
```
eanders: This _looks_ right to me.  Maybe no longer a TODO?

eanders: Oh, except, I _think_ you want to use the local methods instead of reaching out to the client:
```suggestion
    consent_text = no_release_string
    consent_text = full_consent_string if current_consent_type == full_release_string
    consent_text = revoked_consent_string if current_consent_type == revoked_consent_string
```

app/models/consent/default.rb
dtgreiner: I believe we can merge these into one, I had initially kept them separated as the full_consent_string was new in the implied model, but they ended up being the same thing in that class. I'll do some cleanup with these.

app/models/consent/default.rb
dtgreiner: yes, will add this in the next commit

app/models/consent/implied.rb
```diff
@@ -0,0 +1,76 @@
...
+    permission = if va_revoked_consent
+      :can_view_clients
+    else
+      :can_view_client_enrollments_with_roi
+    end
+
+    va_scope = @client.service_history_enrollments.
+      entry.
+      hud_residential
+    va_scope.joins(:project).merge(GrdaWarehouse::Hud::Project.viewable_by(user, permission: permission))
```
eanders:
```suggestion
    revoked_consent = release_current_status == revoked_consent_string

    permission = if revoked_consent
      :can_view_clients
    else
      :can_view_client_enrollments_with_roi
    end

    scope = @client.service_history_enrollments.
      entry.
      hud_residential
    scope.joins(:project).merge(GrdaWarehouse::Hud::Project.viewable_by(user, permission: permission))
```

app/models/consent/implied.rb
dtgreiner: yes, the TODO should be removed, and we should use the local methods here. I missed the change after adding the local methods.

app/models/consent/implied.rb
```diff
@@ -0,0 +1,76 @@
...
+    permission = if va_revoked_consent
+      :can_view_clients
+    else
+      :can_view_client_enrollments_with_roi
+    end
+
+    va_scope = @client.service_history_enrollments.
+      entry.
+      hud_non_residential
+    va_scope.joins(:project).merge(GrdaWarehouse::Hud::Project.viewable_by(user, permission: permission))
```
eanders: I think this can be DRYed up with the above in addition to just making it less VA specific.
```suggestion
    revoked_consent = release_current_status == revoked_consent_string

    permission = if revoked_consent
      :can_view_clients
    else
      :can_view_client_enrollments_with_roi
    end

    scope = @client.service_history_enrollments.
      entry.
      hud_non_residential
    scope.joins(:project).merge(GrdaWarehouse::Hud::Project.viewable_by(user, permission: permission))
```

app/models/grda_warehouse/hud/client.rb
```diff
@@ -984,16 +984,20 @@ def scope_for_ongoing_residential_enrollments
...
+    def scope_for_residential_enrollments(user)
+      active_consent_model.scope_for_residential_enrollments(user)
+    end
+
+    def active_consent_model
+      @active_consent_model ||= GrdaWarehouse::Config.active_consent_class.new(client: self)
+    end
+
+    def self.active_consent_model
+      @active_consent_model ||= GrdaWarehouse::Config.active_consent_class.new(client: self)
```
eanders: This won't work (maybe isn't in use?) since `client:` should be an instance of a client, AND it's memoizing at the class level, which is almost never what you want.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1033,25 +1037,26 @@ def show_health_hpc_for?(user)
...
+      Translation.translate(active_consent_model.full_release_string)
+      active_consent_model.full_release_string
     end
 
     def self.partial_release_string
       # Return the untranslated string, but force the translator to see it
-      Translation.translate('Limited CAS Release')
-      'Limited CAS Release'
+      Translation.translate(active_consent_model.partial_release_string)
+      active_consent_model.partial_release_string
```
eanders: To make this work, I think we want class methods for `partial_release_string` on the consent models.  The instance methods can just return the class method value.

```
def self.partial_release_string
  'the string'
end

def partial_release_string
  self.class.partial_release_string
end
```

That way we can just call 
```
GrdaWarehouse::Config.active_consent_class.partial_release_string
```

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1033,25 +1037,26 @@ def show_health_hpc_for?(user)
...
+      active_consent_model.revoked_consent_string
+    end
+
+    def self.full_consent_string
+      active_consent_model.full_consent_string
+    end
 
-      'None on file'
+    def self.no_release_string
+      active_consent_model.no_release_string
```
eanders: These will need the same treatment.

app/views/clients/va/_consent_form_flags.haml
```diff
@@ -0,0 +1,3 @@
+.client-consent-status.mb-4
+  %h3.mt-0 Consent
```
eanders: I wonder if this should be `Release of Information` or something other than just `Concent` (maybe client consent).  We've historically referred to the release, but we've also been inconsistent.

app/views/clients/va/_tab_navigation.haml
```diff
@@ -0,0 +1,40 @@
+- unless ajax_modal_request?
+  :ruby
+    show_health_pilot = @client.show_health_pilot_for?(current_user) rescue false
+    show_health_hpc = @client.show_health_hpc_for?(current_user) rescue false
```
eanders: Pretty sure the can be removed (and any associated code).

app/views/clients/va/_tab_navigation.haml
```diff
@@ -0,0 +1,40 @@
...
+    show_demographic = @client.show_demographics_to?(current_user) rescue false 
+    show_expanded_va_dashboard = @client.release_current_status == @client.class.full_consent_string
```
eanders: Since were already in the context of VA here, let's just refer to this as 
```suggestion
    show_expanded_dashboard = @client.release_current_status == @client.class.full_consent_string
```

drivers/client_access_control/app/views/client_access_control/clients/_va.haml
```diff
@@ -1,7 +1,4 @@
-- if can_access_some_version_of_clients? && ! GrdaWarehouse::Config.get(:multi_coc_installation)
+- if can_access_some_version_of_clients?
   %section.o-dashboard__block--primary.client__summary
-    = render 'client_access_control/clients/client_summary'
-- if can_view_full_client_dashboard?
-  = render 'client_access_control/clients/rollups'
-- elsif can_view_limited_client_dashboard?
-  = render 'client_access_control/clients/rollups_limited'
+    = render 'client_access_control/clients/va_client_summary'
+= render 'client_access_control/clients/va_rollups'
```
eanders: can we move these into:
```suggestion
    = render 'client_access_control/clients/va/client_summary'
= render 'client_access_control/clients/va/rollups'
```

dtgreiner:
@eanders The requested changes have been added to the PR.

eanders (approved):
Thanks @dtgreiner this seems to work as designed and looks great to me!

PR #4252, opened 2024-04-13
CAS contact info 

eanders:

## Description

This pulls the following for CAS:
1. client email & phone records from HMIS client contacts
2. assessor email from the most-recent CE assessment

## Type of change
- [x] Bug fix

### DIFF
--- a/app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
+++ b/app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
@@ -327,7 +327,7 @@ module GrdaWarehouse::CasProjectClientCalculator
       # if we don't know the assessor, and the assessment was added by the system, ignore it
       return nil if email == User.system_user.email
 
-      email
+      Array.wrap(email)
     end
   end
 end
--- a/app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
+++ b/app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
@@ -34,6 +34,7 @@ module GrdaWarehouse::CasProjectClientCalculator
       [
         :ssvf_eligible,
         :child_in_household,
+        :default_shelter_agency_contacts,
       ].freeze
     end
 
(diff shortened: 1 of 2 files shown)


gigxz:
I think this closes https://github.com/open-path/Green-River/issues/5702? if so, great

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1428,6 +1428,9 @@ def hmis_client_response
     end
 
     def email
+      # Fetch the data from the source clients if we are a destination client
+      return source_clients.map(&:email).reject(&:blank?).first if destination?
```
gigxz, line 1432: Is this only needed because of OP HMIS? Are the `hmis_client_response`s present on the destination client? Because if they're only on destination client, then they would be missed with this change.

app/models/grda_warehouse/cas_project_client_calculator/tc_hmis_hat.rb
```diff
@@ -318,5 +320,14 @@ def most_recent_assessment_for_destination(client)
...
         where(she_t[:age].lt(18).or(she_t[:other_clients_under_18].eq(true))).exists?
     end
+
+    private def default_shelter_agency_contacts(client)
+      # Email of most recent assessor
+      email = cas_assessment(client)&.user&.user_email
+      # if we don't know the assessor, and the assessment was added by the system, ignore it
+      return nil if email == User.system_user.email
+
+      email
```
gigxz: CAS appears to expect this to be an array. It is a jsonb column.
https://github.com/open-path/Green-River/issues/5696#issuecomment-2056858864

https://github.com/greenriver/boston-cas/blob/9d5ffbb0a071d1f4bd0c1f178c80cce62f4754a1/app/views/clients/_contact.haml#L36

That view should probably be more permissive since it can be anything that fits in jsonb

eanders: Maybe, the use of the jsonb was less about being permissive, and more about being able to support multiple email addresses.  I'll Array.wrap this for now, which should be fine.  We should confirm with the client what the expected behavior is (where these emails should come from and how many to include.)

app/models/grda_warehouse/hud/client.rb
eanders: Looks like these are joined to source clients (always, as far as I can tell).  I'm unclear how it was working before.

gigxz (approved):
looks good, confirmed on staging!

PR #4264, opened 2024-04-16
Add task for one-off date updates

martha:

## Description

GH issue https://github.com/open-path/Green-River/issues/5720
This PR adds a script to do the data cleanup described in the ticket, and a spec test.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
martha, line 1: I put this in this directory, rather than just in the rake file alongside the task definition, in order to more easily test it. But please let me know if this is the wrong place and I will happily move it!

drivers/hmis/spec/factories/hmis/hud/enrollments.rb
```diff
@@ -31,7 +31,16 @@
       exit_date { nil }
     end
     after(:create) do |enrollment, evaluator|
-      enrollment.exit = create(:hmis_hud_exit, exit_date: evaluator.exit_date, enrollment: enrollment, data_source: enrollment.data_source, client: enrollment.client) if evaluator.exit_date
+      if evaluator.exit_date
+        enrollment.exit = create(
+          :hmis_base_hud_exit,
```
martha, line 36: A few changes here:
- changed the `enrollment` factory to use `hmis_base_hud_exit` instead of `hmis_hud_exit`, which does some manipulation with the exit date provided. I *think* I tracked down the other usages of this exit date callback, and they are not impacted negatively.
  - client_active_enrollment_spec.rb
  - bulk_assign_service_spec.rb
  - open_enrollment_summary_spec.rb
- also added the `without_validations` trait to the creation call. If you think it's worth the effort, maybe it would be good to make this depend on a `without_validations` trait passed in to the `enrollment` factory?

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+    puts "Updating (normal): #{project_ids.count} projects"
+
+    exits_to_update = Hmis::Hud::Exit.
+      joins(:enrollment).
+      where(e_t[:project_id].in(project_ids)).
+      where(e_t[:entry_date].eq(ex_t[:exit_date]))
+
+    puts "Found #{exits_to_update.count} enrollments with equal entry and exit dates"
+
+    Hmis::Hud::Exit.transaction do
```
martha, line 82: Let me know if we should add some logging to describe whether there are any enrollments in the unexpected situations that we're _not_ correcting for in this script -- for example, exit before entry in a normal (non special treatment) project, or exit before entry in a non-residential project, or etc.

Also, while testing this out locally I was adding the line `Rails.logger.level = 1` in order to see the logs more clearly, without all the output from sql. Would that be helpful to add to the committed code?

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+  def ex_t = Hmis::Hud::Exit.arel_table
+  def sv_t  = Hmis::Hud::Service.arel_table
+  def p_t = Hmis::Hud::Project.arel_table
+
+  def perform
+    puts "Fixing dates. Special treatment project: #{special_treatment_project_id}"
+    special_project = Hmis::Hud::Project.find_by(project_id: special_treatment_project_id)
+    raise 'Special project not found' unless special_project
+
+    normal_project_ids = Hmis::Hud::Project.where.not(project_id: special_treatment_project_id).pluck(:project_id)
```
gigxz: add `hmis` to make sure we're only looking at the HMIS data source, and add the filter for residential project types
```suggestion
    normal_project_ids = Hmis::Hud::Project.hmis.where.not(project_id: special_treatment_project_id).where(project_type: HudUtility2024.residential_project_type_ids).pluck(:project_id)
```

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+    fix_dates(normal_project_ids)
+  end
+
+  def fix_dates_special(project_id)
+    puts "Updating (special): #{project_id}"
+
+    exits_to_update = Hmis::Hud::Exit.
+      joins(:enrollment, :project).
+      where(p_t[:project_id].eq(project_id)).
+      where(p_t[:project_type].in(HudUtility2024.residential_project_type_ids)).
```
gigxz: This isn't needed here, since we're working on the "special" project

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+
+    normal_project_ids = Hmis::Hud::Project.where.not(project_id: special_treatment_project_id).pluck(:project_id)
+    fix_dates_special(special_treatment_project_id)
+    fix_dates(normal_project_ids)
+  end
+
+  def fix_dates_special(project_id)
+    puts "Updating (special): #{project_id}"
+
+    exits_to_update = Hmis::Hud::Exit.
```
gigxz: This is at risk of including Exits from other data sources. Add `hmis` to filter to HMIS data source only
```suggestion
    exits_to_update = Hmis::Hud::Exit.hmis.
```

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+        s.date_provided = new_date
+        s.save!
+      end
+    end
+  end
+
+  def fix_dates(project_ids)
+    puts "Updating (normal): #{project_ids.count} projects"
+
+    exits_to_update = Hmis::Hud::Exit.
```
gigxz:
```suggestion
    exits_to_update = Hmis::Hud::Exit.hmis.
```

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
gigxz: I think this is fine as is, we can run it with extra logging if needed

drivers/hmis/lib/tasks/fix_enrollment_dates_20240416.rake
```diff
@@ -0,0 +1,11 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+require_relative '../hmis_data_cleanup/fix_enrollment_dates_20240416'
+
+desc 'Fix entry and exit dates'
+# rails driver:hmis:fix_enrollment_dates_20240416[222]
+task :fix_enrollment_dates_20240416, [:special_treatment_project_id] => :environment do |_task, args|
+  FixEnrollmentDates20240416.new(special_treatment_project_id: args.special_treatment_project_id).perform
+end
```
gigxz: imo we can remove this and just run cleanup function in the console. I usually feel more comfortable doing that, so I can validate the before and after counts right there in the console.

drivers/hmis/spec/factories/hmis/hud/exits.rb
```diff
@@ -11,6 +11,10 @@
     sequence(:ExitID, 50)
     destination { 1 }
     exit_date { Date.today }
+
+    trait :without_validations do
```
gigxz: This factory should already accept trait `skip_validate` that it inherits from parent `hmis_base_factory`

drivers/hmis/spec/factories/hmis/hud/enrollments.rb
gigxz: Good catch, thanks! I think it's fine to skip validation here because the caller is explicitly passing an exit date.

gigxz: We can use `skip_validate`, though, see below

drivers/hmis/lib/spec/fix_enrollment_dates_spec.rb
```diff
@@ -0,0 +1,75 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+require 'rails_helper'
+require_relative '../hmis_data_cleanup/fix_enrollment_dates_20240416'
+
+RSpec.describe FixEnrollmentDates20240416 do
```
gigxz, line 9: nice tests! thanks

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  attr_accessor :special_treatment_project_id
+  def initialize(special_treatment_project_id:)
+    self.special_treatment_project_id = special_treatment_project_id
+  end
+
+  def e_t = Hmis::Hud::Enrollment.arel_table
+  def ex_t = Hmis::Hud::Exit.arel_table
+  def sv_t  = Hmis::Hud::Service.arel_table
+  def p_t = Hmis::Hud::Project.arel_table
```
gigxz: `include ArelHelper` or `include HmisArelHelper` should get you all of these, with slightly different names.

gigxz: The only difference will be sv_t=>s_t

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
gigxz: makes sense to me!

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  def fix_dates_special(project_id)
+    puts "Updating (special): #{project_id}"
+
+    exits_to_update = Hmis::Hud::Exit.
+      joins(:enrollment, :project).
+      where(p_t[:project_id].eq(project_id)).
+      where(p_t[:project_type].in(HudUtility2024.residential_project_type_ids)).
+      where(ex_t[:exit_date].extract('year').eq(2020)).
+      where(e_t[:entry_date].gteq(ex_t[:exit_date]))
```
gigxz, line 41: nice arel!!

gigxz (commented):
looks great @martha! a few comments inline.

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
martha: Oh good!

I noticed the `e_t` defined in `ArelHelper` is `GrdaWarehouse::Hud::Enrollment.arel_table`, not `Hmis::Hud::Enrollment.arel_table` (the definition that I removed from this file after adding that include statement). Does it still work because these two models use the same underlying table?

gigxz: Yep exactly, it will behave the same way

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  include ArelHelper
+  attr_accessor :special_treatment_project_id
+  def initialize(special_treatment_project_id:)
+    self.special_treatment_project_id = special_treatment_project_id
+  end
+
+  def perform
+    puts "Fixing dates. Special treatment project: #{special_treatment_project_id}"
+    special_project = Hmis::Hud::Project.find_by(project_id: special_treatment_project_id)
```
gigxz: Missed one -- there could be projects in other data sources with the same `ProjectID`. The field, like all the HUD keys, are only guaranteed to be unique within the data source.
```suggestion
    special_project = Hmis::Hud::Project.hmis.find_by(project_id: special_treatment_project_id)
```

martha: @gigxz is this also needed for the Service query on line 54?

gigxz: oh yep, it is. It's needed because of using `ProjectID` to scope down the Services

drivers/hmis/lib/hmis_data_cleanup/fix_enrollment_dates_20240416.rb
```diff
@@ -0,0 +1,91 @@
...
+    Hmis::Hud::Exit.transaction do
+      exits_to_update.each do |e|
+        new_date = e.exit_date.change(year: 2023)
+        puts "Updating exit #{e.id} exit date to #{new_date}"
+        e.exit_date = new_date
+        e.save!
+      end
+    end
+
+    services_to_update = Hmis::Hud::Service.
```
gigxz:
```suggestion
    services_to_update = Hmis::Hud::Service.hmis.
```

gigxz (approved):

PR #4271, opened 2024-04-17
GPGME health integration

awisema:

## Description

Support GPG decryption of Health imports

## Type of change


eanders (approved):
@awisema looks reasonable to me!  Thanks!

app/models/health/tasks/import_epic.rb
```diff
@@ -164,7 +176,14 @@ def fetch_files
       notify 'Health data downloaded'
     end
 
-    def read_csv_file path:
+    def read_csv_file(path:)
+      # Decrypt the file if it is '*.gpg'
+      if path =~ /.gpg$/
```
eanders: maybe?
```suggestion
      if path =~ File.extname(path) == '.gpg'
```

app/models/health/tasks/import_epic.rb
```diff
@@ -164,7 +176,14 @@ def fetch_files
...
+        data = @config.decrypt(File.read(path))
+        path = path[...-4] # Trim off '.gpg'
```
eanders: I feel like there must be a more ruby way to do this, but I'm not finding a great one.  something like `path.sub(/.gpg$//)`  (I don't think that's correct though.)

awisema: How about `File.join(File.dirname(path), File.basename(path, '.gpg'))`?

eanders: Yep, that's the most Rails-y I think!

PR #4273, opened 2024-04-18
Add caseworker exit survey assessment

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/5925
This PR adds a json form definition for a custom form, Caseworker Exit Survey.

I'm not certain whether this should in fact be added to the repo at all? but I had a few specific questions and wanted to ask in code review format.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
+++ b/drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
@@ -16,14 +16,6 @@
           "assessment_date": true
         },
         {
-          "text": "Exit Date",
-          "type": "DATE",
-          "link_id": "caseworker_exit_date",
-          "mapping": {
-            "custom_field_key": "caseworker_exit_date"
-          }
-        },
-        {
           "text": "Phone Number",
           "size": "SMALL",
           "type": "STRING",
@@ -109,10 +101,9 @@
           }
         },
         {
-          "text": "Percent of GAS Change",
-          "type": "INTEGER",
+          "text": "<b>Percent of GAS Change</b>: ${value}%",
+          "type": "DISPLAY",
           "link_id": "caseworker_exit_gas_score_percent_change",
-          "read_only": true,
           "autofill_values": [
             {
               "formula": "(caseworker_exit_end_gas_score - caseworker_exit_beginning_gas_score) / 40 * 100",
@@ -120,10 +111,7 @@
               "autofill_behavior": "ALL",
               "autofill_readonly": true
             }
-          ],
-          "mapping": {
-            "custom_field_key": "caseworker_exit_gas_score_percent_change"
-          }
+          ]
         },
         {
           "text": "Where will the client be residing?",
@@ -276,6 +264,7 @@
           "type": "INTEGER",
           "link_id": "caseworker_exit_gained_employment",
           "read_only": true,
+          "hidden": true,
           "autofill_values": [
             {
               "value_number": 1,
--- a/drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
+++ b/drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
@@ -1,0 +1,611 @@
+{
+  "item": [
+    {
+      "text": "Client Data",
+      "type": "GROUP",
+      "link_id": "caseworker_exit_client_data",
+      "item": [
+        {
+          "text": "Assessment Date",
+          "type": "DATE",
+          "link_id": "caseworker_assessment_date",
+          "required": true,
+          "mapping": {
+            "field_name": "assessmentDate"
+          },
+          "assessment_date": true
+        },
+        {
+          "text": "Exit Date",
+          "type": "DATE",
+          "link_id": "caseworker_exit_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_date"
+          }
+        },
+        {
+          "text": "Phone Number",
+          "size": "SMALL",
+          "type": "STRING",
+          "link_id": "caseworker_exit_phone",
+          "component": "PHONE",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_phone"
+          }
+        },
+        {
+          "text": "Email",
+          "type": "STRING",
+          "link_id": "caseworker_exit_email",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_email"
+          }
+        },
+        {
+          "text": "Number of weeks in program",
+          "type": "INTEGER",
+          "required": true,
+          "link_id": "caseworker_exit_weeks_in_program",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_weeks_in_program"
+          }
+        },
+        {
+          "text": "Beginning GAS Score",
+          "helper_text": "Target 40.00",
+          "type": "NUMBER",
+          "link_id": "caseworker_exit_beginning_gas_score",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_beginning_gas_score"
+          }
+        },
+        {
+          "text": "Beginning GAS Score Date",
+          "type": "DATE",
+          "link_id": "caseworker_exit_beginning_gas_score_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_beginning_gas_score_date"
+          }
+        },
+        {
+          "text": "Exit GAS Score",
+          "helper_text": "Target 40.00",
+          "type": "INTEGER",
+          "link_id": "caseworker_exit_end_gas_score",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_end_gas_score"
+          }
+        },
+        {
+          "text": "Exit GAS Score Date",
+          "type": "DATE",
+          "link_id": "caseworker_exit_end_gas_score_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_end_gas_score_date"
+          }
+        },
+        {
+          "text": "Difference/Change in GAS score",
+          "type": "INTEGER",
+          "link_id": "caseworker_exit_gas_score_change",
+          "read_only": true,
+          "autofill_values": [
+            {
+              "formula": "caseworker_exit_end_gas_score - caseworker_exit_beginning_gas_score",
+              "autofill_when": [
+                {
+                  "operator": "EXISTS",
+                  "question": "caseworker_exit_end_gas_score",
+                  "answer_boolean": true
+                },
+                {
+                  "operator": "EXISTS",
+                  "question": "caseworker_exit_beginning_gas_score",
+                  "answer_boolean": true
+                }
+              ],
+              "autofill_behavior": "ALL",
+              "autofill_readonly": true
+            }
+          ],
+          "mapping": {
+            "custom_field_key": "caseworker_exit_gas_score_change"
+          }
+        },
+        {
+          "text": "Percent of GAS Change",
+          "type": "INTEGER",
+          "link_id": "caseworker_exit_gas_score_percent_change",
+          "read_only": true,
+          "autofill_values": [
+            {
+              "formula": "(caseworker_exit_end_gas_score - caseworker_exit_beginning_gas_score) / 40 * 100",
+              "autofill_when": [
+                {
+                  "operator": "EXISTS",
+                  "question": "caseworker_exit_end_gas_score",
+                  "answer_boolean": true
+                },
+                {
+                  "operator": "EXISTS",
+                  "question": "caseworker_exit_beginning_gas_score",
+                  "answer_boolean": true
+                }
+              ],
+              "autofill_behavior": "ALL",
+              "autofill_readonly": true
+            }
+          ],
+          "mapping": {
+            "custom_field_key": "caseworker_exit_gas_score_percent_change"
+          }
+        },
+        {
+          "text": "Where will the client be residing? (Please check one).",
+          "type": "CHOICE",
+          "required": true,
+          "pick_list_options": [
+            {
+              "code": "Rental Home/Apartment"
+            },
+            {
+              "code": "Family/Friends (Permanent)"
+            },
+            {
+              "code": "Family/Friends (Temporary)"
+            },
+            {
+              "code": "Hotel/Motel"
+            },
+            {
+              "code": "Other Permanent Destination"
+            },
+            {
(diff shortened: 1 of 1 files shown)


drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+              "question": "caseworker_exit_employed_at_exit",
+              "answer_boolean": true
+            }
+          ],
+          "mapping": {
+            "custom_field_key": "caseworker_exit_hourly_rate_at_exit"
+          }
+        },
+        {
+          "text": "Auto populated. If they gained employment in shelter, this should say \"1\" (1 = they gained employment while in shelter, 0 = they either lost employment while in shelter, was employed the entire time in shelter, or never was employed while in shelter. This helps us track how many clients got employment while in shelter.)",
```
martha, line 263: Should this field be hidden? (It's not in ETO)

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+              "autofill_behavior": "ALL",
+              "autofill_readonly": true
+            }
+          ],
+          "mapping": {
+            "custom_field_key": "caseworker_exit_gas_score_change"
+          }
+        },
+        {
+          "text": "Percent of GAS Change",
```
martha: Would it be desirable/worthwhile to create a new form component for percentages, so we can display a % symbol after this field?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+          "type": "DATE",
+          "link_id": "caseworker_exit_beginning_gas_score_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_beginning_gas_score_date"
+          }
+        },
+        {
+          "text": "Exit GAS Score",
+          "helper_text": "Target 40.00",
+          "type": "INTEGER",
```
martha, line 65: It seems we don't have a FLOAT type yet, only INTEGER, which can cause problems with the form if the users enter non-int values or the calculated values end up being non-integers (see e.g. https://greenriver.slack.com/archives/C0610D9HSSH/p1713471971947969)

This needs to be fixed somehow for this form, but I wasn't sure if the right solution is to make a new FLOAT type now, or just have these as integers and convert the readonly auto-filled fields to strings, or some other route?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+          "mapping": {
+            "custom_field_key": "caseworker_exit_end_gas_score_date"
+          }
+        },
+        {
+          "text": "Difference/Change in GAS score",
+          "type": "INTEGER",
+          "link_id": "caseworker_exit_gas_score_change",
+          "read_only": true,
+          "autofill_values": [
```
martha, line 91: Autofill behavior currently shows "Data not collected" instead of "0" when the two values are the same. Should we pull a fix for that into the scope of this ticket, or make a separate ticket to fix that?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+            {
+              "formula": "caseworker_exit_end_gas_score - caseworker_exit_beginning_gas_score",
+              "autofill_when": [
```
martha: I'm also seeing that `autofill_when` doesn't correctly update when a field is filled in but then deleted, as `enable_when` does. Same question here, should I debug and fix now or create a separate ticket?

gigxz:
@martha this shouldn't be merged into the codebase, since it will be managed in the form builder like the rest of the custom forms. but we can discuss here and close the pr after

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: I believe https://github.com/greenriver/hmis-frontend/pull/741 addresses that

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: Yep sounds like a good idea. Can you add an issue to github?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+          "required": true,
+          "link_id": "caseworker_exit_weeks_in_program",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_weeks_in_program"
+          }
+        },
+        {
+          "text": "Beginning GAS Score",
+          "helper_text": "Target 40.00",
+          "type": "NUMBER",
```
gigxz:
```suggestion
          "type": "INTEGER",
```

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+          "assessment_date": true
+        },
+        {
+          "text": "Exit Date",
+          "type": "DATE",
+          "link_id": "caseworker_exit_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_date"
+          }
+        },
```
gigxz: This is a little weird, I'd think we could just show the `Enrollment.exitDate` in the interface as read-only, or remove this if its visible elsewhere. Let's clarify.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+        {
+          "text": "Email",
+          "type": "STRING",
+          "link_id": "caseworker_exit_email",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_email"
+          }
+        },
+        {
+          "text": "Number of weeks in program",
```
gigxz, line 37: It would be cool if this was calculated `(exit - entry)/7` but I don't think that `formula` currently handles local constants, which is what those are. let's leave it as, but just a thought on what might be useful for others.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+          "type": "INTEGER",
+          "required": true,
+          "link_id": "caseworker_exit_weeks_in_program",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_weeks_in_program"
+          }
+        },
+        {
+          "text": "Beginning GAS Score",
+          "helper_text": "Target 40.00",
```
gigxz: maybe update like so, since they are integers
```suggestion
          "helper_text": "Target: 40",
```

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+              "autofill_behavior": "ALL",
+              "autofill_readonly": true
+            }
+          ],
+          "mapping": {
+            "custom_field_key": "caseworker_exit_gas_score_percent_change"
+          }
+        },
+        {
+          "text": "Where will the client be residing? (Please check one).",
```
gigxz: This feels more appropriate since it's a single-select dropdown
```suggestion
          "text": "Where will the client be residing?",
```

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,611 @@
...
+            "custom_field_key": "caseworker_exit_end_gas_score"
+          }
+        },
+        {
+          "text": "Exit GAS Score Date",
+          "type": "DATE",
+          "link_id": "caseworker_exit_end_gas_score_date",
+          "mapping": {
+            "custom_field_key": "caseworker_exit_end_gas_score_date"
+          }
```
gigxz: this should probably have a value_bound minimum of `caseworker_exit_beginning_gas_score_date`

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: This would also work:
```
          "text": "<b>Change in GAS Score</b>: ${value}%",
          "type": "DISPLAY",
          "link_id": "caseworker_exit_gas_score_percent_change",
```
Screenshot 2024-04-19 at 9 42 40 AM


Maybe we should use that approach for both difference and percentage, so they look consistent.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: @martha good question. I ran the export for this one and all the values are integers, they don't actually input fractional values here. So I think we're good to keep this as an integer

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: Good find, a separate ticket would be great. I'm not sure exactly how it should behave, that's a tricky one. Although in this case if you leave `autofill_when` empty, it is probably better, because it reacts to when one or both values are removed.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: This would work visually, but will the value still be saved as a custom data element? (as the ticket mentions "All fields should be saved. Including calculated fields like "difference between savings"") I can test this out to check.

martha: Here's what I found:
- If the field is display type, then the script for creating custom data element definitions will skip it, logging, "Skipping unexpected custom_field_key on non-question item. Unable to determine type."
- However, if the CDED already exists (we would have to create it manually), then the field value will get saved correctly when the form is submitted.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: It can still cause errors if they input only integers, due to the percent change field. For example, inputting 39 in the beginning score and 40 in the exit score results in a % change of 2.5, and an error
`Error processing field 'CustomAssessment.caseworker_exit_gas_score_percent_change': unexpected value "2.5"`

I guess we could consider converting the percent CDED type to a string?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: Added to questions for meeting agenda

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: Yes perfect! :D

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: Ok sounds good, I created https://github.com/open-path/Green-River/issues/5954 and will make that change to this form

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: Right. We have a few places where we actually store values on a DISPLAY type. I see how it's useful, but I don't love it, for this reason. It also complicates the form builder because the data type would need to be specified separately.

I think I'd prefer this type of thing (formatted value display) is implemented with something like `type: INTEGER, component: FORMATTED_DISPLAY, display_formula: "something: ${value}"`, or similar. I'll add it as a discussion topic to raise at standup.

I think for now lets leave what you have as-is.

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
gigxz: String sounds like a good idea for storing the percent change. I _think_ we will need a fix [here](https://github.com/greenriver/hmis-warehouse/blob/b89d9d6c73741c007de9adce3da1478899f9f5a4/drivers/hmis/app/models/hmis/hud/processors/base.rb#L251-L258), it should probably check for `Numeric` instead of just `Integer.

martha: Decision from dev standup: Ignore persisting the percentage entirely, for now. To do: create ticket to revisit integer/float/percent/currency storage and display

Edit: Created https://github.com/open-path/Green-River/issues/5962

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
```diff
@@ -0,0 +1,563 @@
+{
+  "item": [
+    {
+      "text": "Client Data",
+      "type": "GROUP",
+      "link_id": "caseworker_exit_client_data",
+      "item": [
+        {
+          "text": "Assessment Date",
```
martha, line 9: We confirmed this should be allowed post exit; should there be some post-exit bound such as 90 days?

drivers/hmis/lib/form_data/tarrant_county/assessments/caseworker_exit.json
martha: I think we never created a follow-up ticket for this idea -- typed display elements is a different issue from the integer/float issue captured here https://github.com/open-path/Green-River/issues/5962, if I am correctly untangling, right?

Edit: Created https://github.com/open-path/Green-River/issues/5964 to hopefully help us remember the conversation we had about this today later on when we pick it up again.

PR #4277, opened 2024-04-19
Enable filtering by specific custom assessment type

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/5918

This PR adds a new filter option to Assessments, allowing us to filter by the form definition. The motivation for this change is that from the frontend Assessment tables, we want to be able to filter not just by "Custom Form" (form role), but rather by the specific type of custom form (e.g. HAT, UHA, etc.)

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -76,6 +76,10 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
     joins(:project).merge(Hmis::Hud::Project.where(id: project_ids))
   end
 
+  scope :with_form_definition_identifier, ->(form_identifiers) do
```
martha: One risk I haven't wrapped my head all the way around yet: the form definition identifier is not unique, because forms can have multiple versions. I think this okay, right? We would just return all assessments whose definition identifier matches, regardless of version, which is the desired behavior from the frontend perspective?

I'd like to add spec tests for the changes I'm making before this PR gets merged, but I would like to confirm the general approach first.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -447,5 +449,11 @@ def self.available_units_for_enrollment(project, household_id: nil)
           }
         end
     end
+
+    def self.assessment_types_for_project(project)
+      project.custom_assessments.joins(:form_processor, :definition).pluck(:identifier, :title).uniq.map do |id, title|
```
martha: One downside to this change is that, since the pick list now only returns options if a custom assessment exists with that identifier for this project, then the pick list *won't* return options for assessments that are enabled for this project, but none exist yet. 

Maybe this is acceptable, and even positive? But I did think it may confuse users, so let me know if I should seek a different approach.

drivers/hmis/app/models/hmis/filter/assessment_filter.rb
```diff
@@ -20,6 +21,10 @@ def with_role(scope)
     with_filter(scope, :type) { scope.with_role(input.type) }
```
martha: I wasn't sure about whether to attempt to completely remove this old `type` filter in favor of the newly added filter. It seemed to cause other issues (the route `/client/:id/enrollments/:id/intake` failed, for example), but I didn't look deeply into why, and maybe it's a simple fix.

gigxz: @martha we want to adjust the current filter instead of adding a new one. Sorry that wasn't clear! 
The desired dropdown list should look like the below, except `Custom Assessement` is removed and replaced with one item per Custom Assessment Form Definition.

If `project_id` is provided to the picklist query, it should only show Form Definitions that are "available" in the proejct. If `project_id` is not provided (like when rendered on the Client Assessments page) then all custom assessment types should be listed. Does that make sense? I'll address the question of "available" forms on your other comment.


Screenshot 2024-04-22 at 9 47 49 AM

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: I don't think that's ideal UX, and it will be slow to look through all those assessments.  This also needs to work if project is _not_ specified, because the Client assessments table has a type filter too.

Below is a suggestion. It's a little odd to combine the "roles" (eg `INTAKE`) with the identifiers (eg `housing_needs_assessment`), but I think we need to do that in order to get the desired behavior. The "Intake" option should show _all_ Intakes, regardless of what form they used. 

Thoughts? Happy to discuss more at standup

```
      # get all form rules for custom assessments (active and inactive)
      scope = Hmis::Form::Instance.with_role(:CUSTOM_ASSESSMENT)

      # filter down to rules that match this project, if project is specified
      scope = scope.filter { |fi| fi.project_match(project) } if project

      # { code: definition.identifier, label: definition.title }
      custom_options = scope.map(&:to_pick_list_option).uniq.sort_by { |opt| opt[:label] }

      hud_options = Hmis::Form::Definition::FORM_DATA_COLLECTION_STAGES.excluding(:CUSTOM_ASSESSMENT).keys.
        map { |k| { code: k.to_s, label: k.to_s.humanize } }

      hud_options + custom_optionss

```

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
gigxz: Sorry I missed this comment the first time around somehow! 

Yes, that is correct. If a client has 2 assessments, one that was collected with SPDAT V1 and another that was collected with SPDAT V2, they should both be returned when they select Filter option "SPDAT" for assessment type.

Right now this means looking at the form_processor.definition_id, but if/when https://github.com/open-path/Green-River/issues/5709 is implemented (which I [propose](https://docs.google.com/document/d/1iL6yr_2clGEW1scCrnO0mt9uZmb_CbP7q7D0c5pqeUE/edit#heading=h.gsie1zz6or3n) we do..) then this query would be simplified to just look at the custom_assessment.definition_identifier.

martha: Makes sense. I checked out the linked ticket and one of the action items is to check the code for TODO(187248703) (referring to the old PT issue number), so I added a couple of those TODOs here so we remember to make this update

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: Yes this makes sense! Thanks for setting me straight.

After implementing this change I was wondering if it's worthwhile to create a new enum type in graphql, or should this input just be a list of strings?
- There is already an `AssessmentType` enum, from [HUD](https://github.com/greenriver/hmis-warehouse/blob/656e7c43e3a3e3d061f082d6eaf34296361d8c29/lib/util/concerns/hud_lists_2024.rb#L1121-L1123), so the names would conflict. We could name it something like `AssessmentRoleOrIdentifier`
- Could add a new `use_enum_with_same_key` on `Hmis::Form::Definition` that returns all of the identifiers for forms with role CUSTOM_ASSESSMENT
- maybe would need to make changes to `Hmis::FieldMap` to enable functionality like merging, and excluding specific keys?

drivers/hmis/app/graphql/types/forms/enums/pick_list_type.rb
```diff
@@ -15,6 +15,7 @@ class Forms::Enums::PickListType < Types::BaseEnum
     value 'ENROLLABLE_PROJECTS', 'Projects that the User can enroll Clients in'
     value 'RESIDENTIAL_PROJECTS', 'Residential Projects'
     value 'ORGANIZATION', 'All Organizations that the User can see'
+    value 'ASSESSMENT_TYPES'
```
gigxz: Can you add a description here? Maybe: "Assessment types in the system, including custom assessments and assessment types that are inactive. If a project is specified, the list is limited to assessments that exist in the project (both active and inactive)."

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -65,7 +65,14 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
...
     stages = Array.wrap(role).map { |r| Hmis::Form::Definition::FORM_DATA_COLLECTION_STAGES[r.to_sym] }.compact
-    where(data_collection_stage: stages)
+    # joining form processor and definition here so that this scope is compatible to be OR-ed with with_form_definition_identifier.
+    # TODO(#187248703): remove `joins(:form_processor, :definition).`
+    joins(:form_processor, :definition).where(data_collection_stage: stages)
+  end
+
+  scope :with_form_definition_identifier, ->(form_identifiers) do
+    # TODO(#187248703): simplify this query to just look at custom_assessment.definition_identifier and avoid the join
+    joins(:form_processor, :definition).merge(Hmis::Form::Definition.where(identifier: form_identifiers))
```
gigxz: There's no need to do an explicit join with `:form_processor`. Joining with `:definition` is sufficient, the sql will join through form processor anyway:


```
irb(main):021:0>  Hmis::Hud::CustomAssessment.joins(:definition).merge(Hmis::Form::Definition.where(identifier: ["base-annual"])).to_sql
=> "SELECT \"CustomAssessments\".* FROM \"CustomAssessments\" INNER JOIN \"hmis_form_processors\" ON \"hmis_form_processors\".\"custom_assessment_id\" = \"CustomAssessments\".\"id\" INNER JOIN \"hmis_form_definitions\" ON \"hmis_form_definitions\".\"deleted_at\" IS NULL AND \"hmis_form_definitions\".\"id\" = \"hmis_form_processors\".\"definition_id\" WHERE \"CustomAssessments\".\"DateDeleted\" IS NULL AND \"hmis_form_definitions\".\"deleted_at\" IS NULL AND \"hmis_form_definitions\".\"identifier\" = 'base-annual'"

vs

irb(main):022:0>  Hmis::Hud::CustomAssessment.joins(:definition, :form_processor).merge(Hmis::Form::Definition.where(identifier: ["base-annual"])).to_sql
=> "SELECT \"CustomAssessments\".* FROM \"CustomAssessments\" INNER JOIN \"hmis_form_processors\" ON \"hmis_form_processors\".\"custom_assessment_id\" = \"CustomAssessments\".\"id\" INNER JOIN \"hmis_form_definitions\" ON \"hmis_form_definitions\".\"deleted_at\" IS NULL AND \"hmis_form_definitions\".\"id\" = \"hmis_form_processors\".\"definition_id\" INNER JOIN \"hmis_form_processors\" \"form_processors_CustomAssessments\" ON \"form_processors_CustomAssessments\".\"custom_assessment_id\" = \"CustomAssessments\".\"id\" WHERE \"CustomAssessments\".\"DateDeleted\" IS NULL AND \"hmis_form_definitions\".\"deleted_at\" IS NULL AND \"hmis_form_definitions\".\"identifier\" = 'base-annual'
```

gigxz: Also, just as a comment:
`Hmis::Hud::CustomAssessment.joins(:definition).where(fd_t[:identifier].in(form_identifiers)).to_sql` is equivalent. That or using `merge` is just a style question, as far as I can tell (in this case).

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -65,7 +65,14 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
 
   scope :with_role, ->(role) do
     stages = Array.wrap(role).map { |r| Hmis::Form::Definition::FORM_DATA_COLLECTION_STAGES[r.to_sym] }.compact
-    where(data_collection_stage: stages)
+    # joining form processor and definition here so that this scope is compatible to be OR-ed with with_form_definition_identifier.
+    # TODO(#187248703): remove `joins(:form_processor, :definition).`
+    joins(:form_processor, :definition).where(data_collection_stage: stages)
```
gigxz: I don't like doing a `join` in this scope, because the scope itself does not require it. We may use this scope in other places where join is not necessary.

Does it work to move the join the the filter code? So like:
`scope.joins(:definition).with_role(hud_types).or(scope.with_form_definition_identifier(custom_types))

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: It can't be an enum because the form `identifier`s are dynamic basic on what's in the database. I think what you have with strings makes sense.
Maybe we should call it `assessment_name`? Leaving it as type seems okay too. It might conflict with my comments on the frontend though..

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
gigxz: thank you!

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: I'm good with switching to calling it "assessment name." I can make that change on the frontend too.

martha: Although, I'm seeing we use the language of 'Assessment Type' (rather than 'Name') on the frontend in a user-facing way. I guess it's two slightly different concepts,
- Assessment Type could be Intake, Exit ... Custom and corresponds to hud data collection stage
- Assessment Name could be Intake, Exit, but also SPDAT, "Martha's Custom Assessment," etc.

Do you think users of the app will understand this distinction? Maybe it doesn't really matter?

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -296,9 +296,9 @@ type AssessmentEligibility {
 }
 
 input AssessmentFilterOptions {
+  assessmentName: [String!]
```
martha, line 299: What is the best way to roll out this breaking API change? Maybe we need to take a phased approach, adding the new `assessmentName` and then removing the old `type` in a future release? From your comment [here](https://github.com/greenriver/hmis-frontend/pull/744#discussion_r1578513276) it sounds like you weren't too worried about the breaking change and we can just push this as-is?

gigxz: IMO we can live with this because there are no pages that use this filter by default. The user has to go out of their way to apply the filter. So the breaking change would only cause an issue if a user with a stale frontend during a deploy happens to apply the filter. I think we can just let that happen and proceed with the change.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: I think either one is fine. They are both better than just "type". If we get feedback around wording later, we can override the label on the frontend.

gigxz (approved):

gigxz:
wait to merge until https://github.com/greenriver/hmis-frontend/pull/744 goes in

PR #4282, opened 2024-04-25
Make geocode nullable

gigxz:

## Description

Support resolving null geocode. It is allowed to be null in the database, so we should support resolving that. It was previously non-null because HUD spec requires it.

https://green-river.sentry.io/issues/5249984838/?project=4504006381273088&query=is%3Aunresolved+geocode&referrer=issue-stream&statsPeriod=90d&stream_index=0

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/schema.graphql
+++ b/drivers/hmis/app/graphql/schema.graphql
@@ -8362,7 +8362,7 @@ type ProjectCoc {
   dateCreated: ISO8601DateTime
   dateDeleted: ISO8601DateTime
   dateUpdated: ISO8601DateTime
-  geocode: String!
+  geocode: String
   geographyType: GeographyType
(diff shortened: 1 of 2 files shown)


martha (approved):
Thanks! Just curious - is it a problem that HUD spec requires it but our db does not?

gigxz:
Good q @martha ! In this case, no.  Since the warehouse imports HMIS data from lots of external systems, it needs to be forgiving with what it can accept. We don't expect external systems to match the HUD spec perfectly. It's not a big deal if geocode is missing -- thats the kind of thing that should be highlighted in a data quality report.

There are some cases where the db is potentially _too_ lenient, we have some discussion on a ticket here https://github.com/open-path/Green-River/issues/977.

PR #4294, opened 2024-04-29
Send errors to Sentry

eanders:
…the delayed job worker

## Description

There are errors that for whatever reason Delayed Job is just eating.   Until we figure out where that's occurring, send the errors directly to sentry.

## Type of change
- [x] Bug fix

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -234,6 +234,8 @@ gem 'business_time', '~> 0.10.0'
 gem 'cable_ready', '>= 5.0.0.rc2'
 gem 'graphql', '~> 2.0'
 gem 'sentry-rails', '~> 5.5'
+gem 'sentry-ruby'
+gem 'sentry-delayed_job'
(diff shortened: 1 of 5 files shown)


ttoomey (approved):
Probably fine although would be better to just get sentry working and not have to rescue

It looks like we don't have the "sentry-delayed_job" gem installed. Adding that might fix sentry notifications

Maybe it would be better to avoid the 'foo.delay.bar' interface for delayed job and instead use proper job classes? That allows overriding the max_attempts to turn of retry (looks like we have the default set to 3)

eanders:
This now uses the sentry DJ gem and successfully [sent an error to sentry.](https://green-river.sentry.io/issues/5282644319/?project=4504017584783360&query=is%3Aunresolved&referrer=issue-stream&statsPeriod=24h&stream_index=0)

PR #4306, opened 2024-05-01
5961 hmis replace comp keys with generated slugs

ttoomey:
# _Please squash merge_

## Description

[Issue 5961](https://github.com/open-path/Green-River/issues/5961) 

replace composite keys with generated 'slug' column between enrollments, services, customServices, and customServiceTypes. This is a performance optimization

## Type of change
- [x] Performance fix


db/warehouse/migrate/20240430045111_custom_service_type_slug.rb
```diff
@@ -0,0 +1,28 @@
+class CustomServiceTypeSlug < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured do
+      # change_column :Services, :id, :bigint
+      change_column_null :Services, :EnrollmentID, false
+      change_column_null :Services, :PersonalID, false
```
ttoomey, line 6: matches CustomServiceType nullability

db/warehouse/migrate/20240414183410_add_enrollment_slug_to_services.rb
```diff
@@ -0,0 +1,33 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class AddEnrollmentSlugToServices < ActiveRecord::Migration[6.1]
+  def up
+    safety_assured do
```
ttoomey, line 8: So this may be a problem for deployment since the tables are locked while this occurs (same in the CustomServiceTypeSlug migration below). Should we try and find a way to reduce service interruption or just plan to deploy during off-hours?

db/warehouse/migrate/20240414183410_add_enrollment_slug_to_services.rb
```diff
@@ -0,0 +1,33 @@
...
+      [
+        ['Enrollment', 'slug'],
+        ['Services', 'enrollment_slug'],
+        ['CustomServices', 'enrollment_slug'],
+      ].each do |table, field|
+        execute <<~SQL
+          ALTER TABLE "#{table}" ADD COLUMN #{field} VARCHAR GENERATED ALWAYS AS (data_source_id::text || ':' || "EnrollmentID") STORED
```
ttoomey, line 15: not including personal id here since enrollments may have mismatched personal ids due to merge

db/warehouse/migrate/20240430045111_custom_service_type_slug.rb
```diff
@@ -0,0 +1,28 @@
...
+      execute <<~SQL
+        ALTER TABLE "Services" ADD COLUMN custom_service_type_slug VARCHAR GENERATED ALWAYS AS (data_source_id::text || ':' || "TypeProvided"::text || ':' || "RecordType"::text) STORED
+      SQL
+      execute <<~SQL
+        ALTER TABLE "CustomServiceTypes" ADD COLUMN slug VARCHAR GENERATED ALWAYS AS (data_source_id::text || ':' || "hud_type_provided"::text ||':'|| "hud_record_type"::text) STORED
+      SQL
+      execute %{CREATE INDEX idx_custom_service_type_slug ON "CustomServiceTypes" (slug)}
```
ttoomey, line 10: could add an index on services.custom_service_type_slug but I'm not sure it's worth it, service type seems low cardinality and local perf seems okay without it at least.

db/warehouse/migrate/20240414183410_add_enrollment_slug_to_services.rb
eanders: We can probably plan for off-hours.  I think the other solution would be to duplicate the table, add the column and then swap in the new table, but while it might not lock the table, it risks losing data...  Looking around, another possible solution would be to create a trigger that filled the column instead of leaning on the generated column, but I'd prefer to use the generated column if we can stomach the table locks.

@gigxz ?

eanders (approved):
I like this solution better if we can handle the locked tables, or work around that somehow.

db/warehouse/migrate/20240430045111_custom_service_type_slug.rb
```diff
@@ -0,0 +1,25 @@
...
+      execute %(VACUUM ANALYZE "Enrollment")
+      execute %(VACUUM ANALYZE "CustomServices")
+      execute %(VACUUM ANALYZE "Services")
+    end
+  end
+
+  def down
+    safety_assured do
+      change_column_null :Services, :EnrollmentID, true
+      change_column_null :Services, :PersonalID, true
```
gigxz, line 20: What's this for? I don't see anywhere we make them required on `up`

db/views/hmis_services_v05.sql
```diff
@@ -0,0 +1,48 @@
...
+		"DateProvided",
+		"Services"."UserID" :: varchar AS "UserID",
+		"Services"."DateCreated" AS "DateCreated",
+		"Services"."DateUpdated" AS "DateUpdated",
+		"Services"."DateDeleted" AS "DateDeleted",
+		"Services"."data_source_id" AS "data_source_id"
+	FROM
+		"Services" JOIN "CustomServiceTypes"
+		ON "CustomServiceTypes"."slug" = "Services"."custom_service_type_slug"
+		AND "CustomServiceTypes"."DateDeleted" IS NULL
```
gigxz, line 22: I'm assuming this join is causing performance issues,  which is why the `custom_service_type_slug` is being proposed?

Have you considered that we can get rid of this join entirely? It was added as a convenience so that the view has `custom_service_type_id` on each "row" regardless of whether it's HUD or Custom. 

But, an alternative would be for the view to resolve the `RecordType` and `TypeProvided` from the Services table instead.
The resulting view would have an optional `custom_service_type_id` (if it's a Custom Service) an optional `record_type`/`type_provided` (if it's a HUD Service).

The `HmisService` model could handle translating the `record_type`/`type_provided` into a "custom service type" as needed, we can use a cache for that lookup too because it should never change. I think we over-engineered for configurability with the HUD Services by storing them as CustomServiceTypes, they could just as well be in code.


I just want to throw that out there. Maybe we can huddle to discuss further. I worry about adding this `custom_service_type_slug` because it's further entrenching/complicating a design decision that is questionable imo. (The decision to store HUD Service types in the CustomServiceTypes table).

db/warehouse/migrate/20240414183410_add_enrollment_slug_to_services.rb
ttoomey: To reduce potential downtime, this would probably have to be two deploys: a initial migration just adds the new slugs (trigger probably). Once the slugs are built we'd update the view and change the rails associations with another deploy

db/warehouse/migrate/20240430045111_custom_service_type_slug.rb
ttoomey: Ah oops, I forgot to take this out. I had initially changed the nullability of Services to match CustomServices but I don't think that matters for performance

db/views/hmis_services_v05.sql
ttoomey: oh yeah, thanks for calling that out! The service type join seemed to be contributing to performance issues here. I think if we can simplify the relationship as you suggest, that would be much better than adding another column.

I'll give that a try

ttoomey:
[more minimal approach here](https://github.com/greenriver/hmis-warehouse/pull/4318), let's try that first

PR #4310, opened 2024-05-01
Create scope and update schema to support form versioning

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/5953

This PR adds a new scope to FormDefinitions, which returns one definition per identifier. It also adds a new frontend schema type, FormIdentifier.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -268,6 +268,17 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
     where(identifier: instance_scope.pluck(:definition_identifier))
   end
 
+  scope :current_versions, -> do
```
martha: @ttoomey, I'd like to ask for your rails expertise on this new scope, or maybe we could pair this afternoon or tomorrow. Ideally for performance purposes we would make this query using only one hit to the DB, possibly using ActiveRecord's `group` instead of ruby's `group_by`? But Gig and I weren't able to figure that out on a brief pairing session earlier today.

Another issue with the code as written is that this scope doesn't chain in an expected way with other scopes. `Hmis::Form::Definition.current_versions.non_static` returns different results from `Hmis::Form::Definition.non_static.current_versions`.

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,45 @@
...
+    def id
+      object.id
+    end
+
+    def current_version
+      object
+    end
+
+    def draft
+      Hmis::Form::Definition.find_by(identifier: object.identifier, status: 'draft')
```
gigxz: This should be accompanied by a database constraint that there can only be 0-1 draft per identifier. I pulled that into its own story for more discussion and accompanying index changes: https://github.com/open-path/Green-River/issues/6006

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,45 @@
...
+    # TODO: add summary information so the UI can resolve top-level detail such as 'active in 15 projects'
+
+    field :id, String, null: false
+    field :identifier, String, null: false
+    field :role, Types::Forms::Enums::FormRole, null: false
+    field :title, String, null: false
+
+    field :current_version, Types::Forms::FormDefinition, null: false
+    field :draft, Types::Forms::FormDefinition, null: true
+    field :retired_versions, [Types::Forms::FormDefinition], null: false
```
gigxz: It's not really clear based on the designs yet, but I think we'll want to paginated this. So the type would be `Types::Forms::FormDefinition.page_type`. I also could see it being nice to include _all_ the versions in the table (not just the retired ones) but we can always adjust that later on a second pass.

drivers/hmis/app/models/hmis/form/definition.rb
ttoomey: Hi @martha 
I think you could do achieve this with postgres "DISTINCT ON" syntax (also possible with "MAX/GROUP BY" but I think distinct on is more readable). I think you could use something like the following to push this behavior into a single database query. Using a subquery is less likely to break active record's expectations than say manipulating the select clause of the query itself

```ruby
 # the id of the row with the largest version by identifier  
 CURRENT_VERSION_SQL = <<~SQL
    SELECT DISTINCT ON (identifier) id
    FROM hmis_form_definitions
    ORDER BY identifier, version DESC
  SQL
  scope :current_versions, -> {
    where("hmis_form_definitions.id IN (#{CURRENT_VERSION_SQL})")
  }
```

Regarding your second question about the difference in behavior, I think it may be because you are returning a new scope and not "chaining" from the current scope (the current scope is `self` in this context). So you'd probably want:
```diff
-  Hmis::Form::Definition.where(id: ids)
+  self.where(id: ids)
```

Happy to pair on this if you have questions

martha: Wonderful @ttoomey thanks for taking a look and putting me on the right track!

eanders: @martha we have arel helpers for this, checkout `one_for_column`.
https://github.com/greenriver/hmis-warehouse/blob/0212c91c3a96a3e9f02a20e7c82bed7791ffcdfd/app/models/concerns/arel_helper.rb#L179-L216

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    # TODO: add summary information so the UI can resolve top-level detail such as 'active in 15 projects'
+
+    field :id, String, null: false
+    field :identifier, String, null: false
+    field :role, Types::Forms::Enums::FormRole, null: false
+    field :title, String, null: false
+
+    field :current_version, Types::Forms::FormDefinition, null: false
+    field :draft, Types::Forms::FormDefinition, null: true
```
gigxz: If a form hasn't been published yet (so there is only a draft version), how does it resolve? I see the current_version is non-nullable.

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,45 @@
...
+    def current_version
+      object
+    end
+
+    def draft
+      Hmis::Form::Definition.find_by(identifier: object.identifier, status: 'draft')
+    end
+
+    def retired_versions
+      Hmis::Form::Definition.where(identifier: object.identifier, status: 'retired').order(version: :desc)
```
gigxz: It feels like the statuses should be scopes so we don't have raw strings like this. Like: `Hmis::Form::Definition.where(identifier: object.identifier).retired`


I also added spec to this story to use AR enums for the statuses: https://github.com/open-path/Green-River/issues/6006.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -416,6 +416,26 @@ def form_definitions(filters:)
       scope.order(updated_at: :desc)
     end
 
+    field :form_identifier, Types::Forms::FormIdentifier, null: true do
+      argument :identifier, String, required: true
+    end
+    def form_identifier(identifier:)
+      raise 'Access denied' unless current_user.can_configure_data_collection?
+
+      Hmis::Form::Definition.where(identifier: identifier).current_versions.first
```
gigxz: It looks like with the proposed behavior, if a form is marked "retired" and there is no published or draft form to replace it, it completely disappears from the config tool interface. (neither `formIdentifiers` or `formIdentifeir` query will resolve it).

 I don't think that is correct, we still need to be able to view those forms (and create new drafts of them to publish again). I didn't think about that edge case when spec'ing.   Do you agree w that desired behavior, and am I understanding this code correctly?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -268,6 +268,20 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
...
+    SELECT DISTINCT ON (identifier) id
+    FROM hmis_form_definitions
+    WHERE status IN ('published', 'draft')
+    ORDER BY identifier, CASE
+      WHEN status = 'published' THEN 1
+      WHEN status = 'draft' THEN 2
+      ELSE 3
+    END, version DESC
+  SQL
+  scope :current_versions, -> do
```
gigxz: This scope probably warrants an explanatory comment. I could see the name `current_versions` being misinterpreted as _active/published versions_ that can be used in the UI, which it's not (since some are draft).

gigxz (commented):
looks good so far!

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -416,6 +416,26 @@ def form_definitions(filters:)
...
+    end
+
+    field :form_identifiers, Types::Forms::FormIdentifier.page_type, null: false do
+      filters_argument Forms::FormIdentifier
+    end
+    def form_identifiers(filters: nil)
+      raise 'Access denied' unless current_user.can_configure_data_collection?
+
+      scope = Hmis::Form::Definition.non_static.current_versions
+      scope = scope.apply_filters(filters) if filters
```
martha, line 435: @gigxz this filter which accepts `search_term` doesn't return hits if the search term matches previous versions. If an assessment used to be named 'XYZ' and its name changes to 'ABC', it won't be returned anymore if the user searches 'XYZ'.

To implement that I think we could
- make a new `Hmis::Filter::FormIdentifierFilter` (right now this is using the existing `Hmis::Filter::FormDefinitionFilter`)
- have it search all FormDefinitions for the search term
- then return the current version _for each of those ids_, a scope that doesn't exist currently

Do you agree we should implement this behavior and with this approach?

drivers/hmis/app/graphql/types/forms/form_identifier.rb
martha: Hm, right now `current_versions` returns a draft version if no published version exists. Maybe this is too confusing. I guess the reason I thought to do this is because we basically want a query that returns one FormIdentifier per row that we want to display in the form table, and we want to display both published and draft-only forms there.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Yes to both. Maybe this would make `current_versions` even more confusing, but it sounds like we should also return the most recent retired version if no published or draft version exists.

drivers/hmis/app/models/hmis/form/definition.rb
martha: Agree this needs a comment, it's even feeling to me like we should consider renaming `current_versions` entirely. A couple ideas:
- `one_per_identifier`
- `per_identifier`
- `identifier_representatives`
- `representative_versions` * my favorite

drivers/hmis/app/graphql/types/forms/form_identifier.rb
martha: Hm, agree, and maybe it's better to make that change now to avoid a schema update later? That does seem to be what the design indicates, so I can go ahead and make that change

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,45 @@
...
+    skip_activity_log
+    description 'FormIdentifier'
+
+    available_filter_options do
+      arg :search_term, String
+      # ADD: role
+      # ADD: status
+    end
+
+    # TODO: add summary information so the UI can resolve top-level detail such as 'active in 15 projects'
```
martha: @gigxz should this update, and other updates to basically make the form versions table match the new [design](https://www.figma.com/file/F9EHR2IZYuMAlOJ9aOG2IL/OP-%2F-HMIS-%2F-2024-%2F-Q1-%2B-Q2-(Jan-1---June-30)?type=design&node-id=2611-31087&mode=design&t=s8EuJodzWywXzhGa-0), be in-scope for this ticket?

gigxz: the "summary" piece needs to be spec'd still, so no. We need to think more about what is helpful to show imo. If there are UI-only changes on the forms table, feel free to make those.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Yeah, I think so. I don't think those should just disappear.

drivers/hmis/app/models/hmis/form/definition.rb
martha: Hm. Another idea is to scrap the complicated logic relating to statuses, and just simply have this scope be called `latest_versions` and return the latest version (by version number) for each form.

That makes the code and variable naming simpler. It also enables us to use the `one_for_column` arel helper Elliot mentioned above, without modification. 

The result would be that in the frontend table of form definitions, if a user changes a form's title/role/etc., then that change would still show up in the table before the user has hit publish.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: As discussed we can skip this, the current behavior is sufficient!

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: ok huddle summary>>
* use latest_version as suggested
* gql type handles choosing which title to display
* don't worry about search scope for now
* use data loaders for versions
* use identifier as cache key

drivers/hmis/app/graphql/types/forms/form_identifier.rb
gigxz: as discussed, lets rename to published_version and make nullable

drivers/hmis/app/models/hmis/form/definition.rb
martha: Marvelous, thank you Elliot. I think going with this helper simplified the code significantly.

(We decided against some complicated logic creating a scope that returns a published version for that identifier if it exists, else a draft if it exists, else a retired version; and instead just implemented a scope that uses `one_for_column` out-of-the-box to return the latest version per identifier.)

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -416,6 +416,26 @@ def form_definitions(filters:)
       scope.order(updated_at: :desc)
     end
 
+    field :form_identifier, Types::Forms::FormIdentifier, null: true do
+      argument :identifier, String, required: true
+    end
+    def form_identifier(identifier:)
+      raise 'Access denied' unless current_user.can_configure_data_collection?
+
+      Hmis::Form::Definition.where(identifier: identifier).published.first
```
ttoomey: perhaps order? 
```suggestion
      Hmis::Form::Definition.where(identifier: identifier).order(version: :desc).published.first
```

Or maybe it should use the same scope as the the `form_identifiers` resolver
```suggestion
      Hmis::Form::Definition.non_static.latest_versions.where(identifier: identifier).first
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -46,6 +46,9 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   has_many :external_form_publications, class_name: 'HmisExternalApis::ExternalForms::FormPublication', dependent: :destroy
   has_many :custom_data_element_definitions, class_name: 'Hmis::Hud::CustomDataElementDefinition', dependent: :nullify,
                                              primary_key: 'identifier', foreign_key: 'form_definition_identifier'
+  has_one :published_version, -> { published }, class_name: 'Hmis::Form::Definition', primary_key: 'identifier', foreign_key: 'identifier'
+  has_one :draft_version, -> { draft }, class_name: 'Hmis::Form::Definition', primary_key: 'identifier', foreign_key: 'identifier'
```
ttoomey: It looks like if there were multiple draft/published versions you might get an inconsistent record. Perhaps we should have a constraint to enforce that, otherwise adding an order should fix it
```suggestion
  has_one :published_version, -> { order(version: :desc).published }, class_name: 'Hmis::Form::Definition', primary_key: 'identifier', foreign_key: 'identifier'
  has_one :draft_version, -> { order(version: :desc).draft }, class_name: 'Hmis::Form::Definition', primary_key: 'identifier', foreign_key: 'identifier'
```

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,60 @@
...
+    end
+
+    def id
+      object.identifier
+    end
+
+    def title
+      # If there is a published version corresponding to this identifier, return its title. Otherwise
+      # (meaning there exist only draft or retired versions for this identifier), return whatever title we have
+      published&.title ? published.title : object.title
```
ttoomey: fwiw, this could be golfed down a bit:
```ruby
 published&.title || object.title
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Good idea, I think I like using the same scope as the `form_identifiers` resolver. In theory they should all resolve to similar objects anyway with the same drafts/published version/etc, but it's good to make it more predictable and consistent

drivers/hmis/app/models/hmis/form/definition.rb
martha: I think we plan to have a uniqueness constraint - @gigxz added a note to do that on this ticket https://github.com/open-path/Green-River/issues/6006

However, I'm happy to add the order here too just to be sure.

martha: I think we plan to have a uniqueness constraint - @gigxz added a note to do that on this ticket https://github.com/open-path/Green-River/issues/6006

However, I'm happy to add the order here too just to be sure.

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -3981,6 +3981,32 @@ type FormDefinitionsPaginated {
...
+"""
+FormIdentifier
+"""
+type FormIdentifier {
+  allVersions(limit: Int, offset: Int): FormDefinitionsPaginated!
+  displayVersion: FormDefinition!
+  draft: FormDefinition
+  id: String!
+  identifier: String!
+  published: FormDefinition
```
gigxz: It seems inconsistent that `displayVersion` and `allVersion` have "version" in the name and they resolve FormDefinitions, whereas `draft` and `published` _don't_ have `Version` in the name, even though they resolve the same thing. I would have thought `draft`/`published` were booleans based on the name.

 I suggest standardizing, to all use `_version[s]` suffix (or `_form`, but maybe we have enough "form" language 😅 )

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,55 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Types
+  class Forms::FormIdentifier < Types::BaseObject
+    skip_activity_log
+    description 'FormIdentifier'
```
gigxz: Comment below is helpful but it would be nice to update the description too
```suggestion
    description 'Type representing one form Identifier, which collects all versioned FormDefinitions for the same identifier'
```

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,55 @@
...
+      # ADD: status
+    end
+
+    field :id, String, null: false
+    field :identifier, String, null: false
+
+    field :published, Types::Forms::FormDefinition, null: true
+    field :draft, Types::Forms::FormDefinition, null: true
+    field :all_versions, Types::Forms::FormDefinition.page_type, null: false
+    field :display_version, Types::Forms::FormDefinition, null: false
```
gigxz: Suggest adding description to clarify that this field should _not_ be used for rendering forms in the application 
```suggestion
    field :display_version, Types::Forms::FormDefinition, null: false, description: 'Form version to use for display in the configuration tool interface. The form itself may be draft, status, or retired.'`
```

drivers/hmis/app/graphql/types/forms/form_identifier.rb
```diff
@@ -0,0 +1,55 @@
...
+    def published
+      load_ar_association(object, :published_version)
+    end
+
+    def draft
+      load_ar_association(object, :draft_version)
+    end
+
+    def all_versions
+      load_ar_association(object, :all_versions)
```
gigxz, line 45: We will probably want some consistent sort to this by descending version. Put that can be deferred to https://github.com/open-path/Green-River/issues/5982

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -416,6 +416,26 @@ def form_definitions(filters:)
       scope.order(updated_at: :desc)
     end
```
gigxz, line 418: Can we add a `deprecation_reason` to the `form_definitions` query, since it is not used anymore?

gigxz (approved):
Looks good to me!! 👏

PR #4311, opened 2024-05-02
4818-add-client-name-permission

ttoomey:

## Description
[Issue 4818](https://github.com/open-path/Green-River/issues/4818)

Adds new abstractions 
* PiiProvider, accessed as client.pii_provider(user:) to centralize pii handling
* Authorization policies, user.policies.for_client, project, etc. Obeys access controls when relevant instead of assuming global permissions
* CoreDemographicsReport:: DetailsColumn to handle enforcing pii in cols

## Details
* update client search and client dashboard to redact names
* update core demographics details to redact names and dob
* Add ClientPII and client.pii_provider(user:) to centralize pii handling
* add can_view_client_name permission
* add documentation to current permission models
* cache viewable_project_ids

## Type of change


app/models/grda_warehouse/client_pii.rb
```diff
@@ -0,0 +1,66 @@
...
+
+  def full_name
+    [first_name, middle_name, last_name].compact.join(' ').presence
+  end
+
+  def brief_name
+    [first_name, last_name].compact.join(' ').presence
+  end
+
+  DATE_FMT = '%m/%d/%Y'.freeze
```
eanders: On the warehouse side, we've always used `Aug 30, 2021` as the date format.  It's non-US standard, but I don't think we should enforce a different format here.

app/models/grda_warehouse/client_pii.rb
```diff
@@ -0,0 +1,66 @@
...
+    name_part(client.first_name.presence)
+  end
+
+  def last_name
+    name_part(client.last_name.presence)
+  end
+
+  def middle_name
+    name_part(client.middle_name.presence)
+  end
```
eanders: I think these should access the name part with the hash key format (`client[:FirstName]`) this would allow us to pass in a "client" that is really a hash so we can use this for situations where we've plucked the data and don't actually have an instance of `Client`.

The same would need to happen for where it calls `client.ssn` and `client.dob`. 

Alternatively we could new up a struct or open struct and pass it in with the appropriately named attributes, but that feels like it will be a bigger list and use more memory.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1403,9 +1406,15 @@ def presented_with_family?(after: nil, before: nil)
       end
     end
 
+    def pii(user: nil)
+      GrdaWarehouse::ClientPii.new(user: user, client: self)
+    end
+    memoize :display
```
eanders: should this be `memoize :pii` ?

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1403,9 +1406,15 @@ def presented_with_family?(after: nil, before: nil)
       end
     end
 
+    def pii(user: nil)
```
eanders: I wonder if this should be named `masked_pii` or `handle_pii`.  I think `pii` is fine, but doesn't imply that it does any manipulation.

eanders (approved):
Thanks @ttoomey

app/models/grda_warehouse/hud/client.rb
ttoomey: Yes, I considered other names, naming is hard as always. I think "PII" is a reasonable choice since it accurately describes the return value (a bag of pii accessors)
- "masked_pii" this name implies that the pii is always masked which is not true
- "handle_pii" this name suggests an event handler or action to me, but these are just accessors

Maybe 'pii' is too short and hard to grep for, we could go with something like "client.pii_accessor()" or pii_provider

app/models/grda_warehouse/client_pii.rb
ttoomey: I opted for to wrap attributes up in a struct. This does add some small memory overhead but makes the class cleaner imo. Could revisit if there's a need for micro-optimizations around memory usage

app/views/clients/_match_results.html.haml
```diff
@@ -24,15 +24,9 @@
               = link_to(client_name, client_path(c), target: "_blank")
           - if sc_count == 1
             %td
-              - if can_view_full_dob?
-                = c.DOB
-              - else
-                = c.age
+              = c.pii(current_user).dob_or_age
```
eanders: I think these need to be:
```suggestion
              = c.pii(user: current_user).dob_or_age
```

app/models/grda_warehouse/hud/client.rb
gigxz: I like pii_accessor and pii_provider

app/models/grda_warehouse/client_pii.rb
```diff
@@ -0,0 +1,103 @@
...
+
+  def ssn(force_mask: false)
+    value = record.ssn.presence
+    mask = force_mask || !policy.can_view_full_ssn?
+    format_ssn(value, mask: mask) if value
+  end
+
+  def client_id_as_name
+    "Client ID: #{record.id || 'N/A'}"
```
gigxz: It might be nice for this to be consistent with the HMIS.
HMIS currently does:
https://github.com/greenriver/hmis-warehouse/blob/f83d7e658ad43e7a1dc7bd4a1d596d5bcd7138ed/drivers/hmis/app/models/hmis/hud/client.rb#L215-L217

app/models/grda_warehouse/client_pii.rb
```diff
@@ -0,0 +1,103 @@
...
+  def format_full_ssn(value)
+    value.gsub(SSN_RGX, '\1-\2-\3')
+  end
+
+  def name_part(value, initial)
+    return nil unless value.present?
+
+    return value if policy.can_view_client_name?
+
+    return initial == 'M' ? initial : "#{initial}****"
```
gigxz: So a client name would appear as `F*** M L****` if the user lacks access to view names?
I would assume that means they have a first name that starts with "F" and a last name that starts with "L".

I think we should decide on a pattern and use it in each system. In the HMIS we mask client names as just "Client :id" as the first name.

ttoomey: Yeah this is confusing. How about we go with "Name Redacted" instead?

app/models/user.rb
```diff
@@ -69,18 +70,6 @@ def load_health_effective_permissions
       send(permission)
     end
 
-    define_method("#{permission}_for?") do |entity|
```
ttoomey, line 72: I think these aren't used(?)

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,86 @@
...
+  memoize def legacy_user_role_policy
+    GrdaWarehouse::AuthPolicies::LegacyUserRolePolicy.new(user: user)
+  end
+
+  # Needs review
+  def visible_client_project_ids(client_id)
+    p_t = GrdaWarehouse::Hud::Project.arel_table
+    GrdaWarehouse::Hud::Enrollment
+      .visible_to(user, client_ids: [client_id])
+      .joins(:project).order(p_t[:id]).pluck(p_t[:id])
```
ttoomey: @eanders this probably isn't right. Can you point me in the right direction? Given an acl-using user, how should I find the set of authorized projects relevant to the client_id?

app/controllers/clients_controller.rb
```diff
@@ -223,6 +223,7 @@ def chronic_days
   end
 
   private def client_scope(id: nil)
+    # this is probably not used
```
eanders:
```suggestion
    # this is probably not used since it is overridden in the client_access_control driver.
```

app/models/grda_warehouse/auth_policies/any_policy.rb
```diff
@@ -0,0 +1,14 @@
+class GrdaWarehouse::AuthPolicies::AnyPolicy
```
eanders: These files are missing our copyright notice.
```
###
# Copyright 2016 - 2024 Green River Data Analysis, LLC
#
 ###
```

app/models/grda_warehouse/auth_policies/policy_provider.rb
eanders: @ttoomey I think we'll need to crib from `drivers/client_access_control/extensions/grda_warehouse/hud/client_extension.rb` and `drivers/client_access_control/app/models/client_access_control/enrollment_arbiter.rb`.  Specifically, I think `ClientAccessControl::EnrollmentArbiter#clients_destination_visible_to` and `ClientAccessControl::EnrollmentArbiter#clients_source_visible_to`.  The trick is that we've never done per-client per-field permissions, so there's some ambiguity (or not yet determined functionality) around how we handle things like releases of information.  If a client has an ROI on file, should you be able to see their name even if otherwise you wouldn't be able to even see their record at all?

eanders: I _think_ this ends up being something like:
```
arbiter = GrdaWarehouse::Config.arbiter_class.new(user)
arbiter..enrollments_visible_to(user, client_ids: some_set_of_client_ids_for_performance).
  joins(:project).
  pluck(p_t[:id])
```

The issue with the above is that it's missing the ability to pass in the related permission.  Maybe that's ok because you're checking the permission against the `project_id` in `client_project_policy` but it negates the ROI since the ROI doesn't automatically add projects to your list.  I think we need a method where you can ask "For which clients do I have this permission?"

I'll push a possible solution to the enrollment arbiter class.

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,108 @@
...
+  end
+
+  protected
+
+  def handle_legacy_unauthorized
+    raise 'legacy authorization not performed'
+  end
+
+  memoize def viewable_project_ids
+    user.viewable_project_ids(:can_view_projects)
```
eanders: This is the issue I was hinting at.  The permission that you need to check is the permission in question (maybe `can_view_client_names`)  Can view projects is only to be used in the context of if you can see the project.

eanders: That's compounded by the concept of ROI which can release a client regardless of project relationship.  BUT we don't have a definition of _what_ is being released, so maybe we shouldn't release name or SSN or HIV status....

app/models/grda_warehouse/hud/client.rb
```diff
@@ -815,6 +819,7 @@ def self.extended_age_groups
     end
 
     def source_clients_searchable_to(user)
+      # FIXME: looks like we intended to memoize, it's not working due to missing conditional assignment
```
eanders: I think that's a correct assessment.  This should be `||=`  I'm not seeing an obvious reason it couldn't be memoized that way.

drivers/core_demographics_report/app/models/core_demographics_report/details_column.rb
```diff
@@ -0,0 +1,46 @@
...
+      pii_value(raw_value, policy)
+    end
+
+    protected
+
+    REDACTED = 'Redacted'.freeze
+    def pii_value(raw_value, policy)
+      case label
+      when 'First Name', 'Last Name'
+        policy.can_view_client_name? ? raw_value : REDACTED
```
eanders: This should probably be something like `policy.name_or_redacted(:first_name, :can_view_client_name?)`

ttoomey: I don't think display logic should go in the policy object (separation of concerns and all that).
It could be rolled up into PiiProvider, maybe a class method would work for this case where we are iterating over cells 
`PiiProvider.first_name(raw_value, policy: policy)`

app/controllers/application_controller.rb
```diff
@@ -291,4 +291,10 @@ def set_sentry_user
   def set_app_user_header
     response.headers['X-app-user-id'] = current_user&.id
   end
+
+  # TODO: START_ACL remove when ACL transition complete
+  def set_legacy_implicitly_assume_authorized_access
+    current_user.policies.legacy_implicitly_assume_authorized_access = true
```
eanders, line 299: maybe:
```suggestion
    return if current_user.using_acls?
    
    current_user.policies.legacy_implicitly_assume_authorized_access = true
```

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,99 @@
...
+  end
+
+  # TODO: START_ACL remove after ACL migration is complete
+  memoize def legacy_user_role_policy
+    handle_legacy_unauthorized unless legacy_implicitly_assume_authorized_access
+    GrdaWarehouse::AuthPolicies::LegacyUserRolePolicy.new(user: user)
+  end
+  # END_ACL
+
+  # Needs review
```
eanders: I believe this is correct because it enforces `can_view_clients` which in the ACL world is needed for both viewing a client and searching. (see note on `can_search_own_clients` in the role class.)

ttoomey: I wonder, maybe it's better to just query the projects for all the client's enrollments since the checks are performed in ProjectPolicy?
```
  memoize def client_project_policy(client_id)
    c_t = GrdaWarehouse::Hud::Client.arel_table
    client_project_scope = GrdaWarehouse::Hud::Project.joins(:clients).where(c_t[:id].eq(client_id))
    project_policies = client_project_scope.pluck(:id).map do |project_id|
      for_project(project_id)
    end
    GrdaWarehouse::AuthPolicies::AnyPolicy.new(policies: project_policies)
  end
```

ttoomey:
@elliot - follow up our conversation:
* I tried to memoize user.viewable_project_ids but that caused failures in other tests so I rolled it back
* I tried to adjust the can_search_own_clients factory role to include can_view_clients- that fixed the failing tests but resulted in many other test failures. ~~I had to find a different solution. I ended up replacing the policy_provider.visible_client_project_ids see 19e062d5316dcb7d7a1c0b5addbe25dde203fe60~~

ttoomey:
Even though tests seem to be passing, this needs work. PolicyProvider.client_project_policy is not right, it should not include window_data_source_ids. Need to figure a better way to determine the appropriate role for clients without enrollments in a windowed data source

eanders:
@ttoomey I've been [playing around with this a bit here](https://github.com/greenriver/hmis-warehouse/pull/4401).  I think the test failures (at least some in drivers/client_access_control/spec/requests/client_visibility_spec.rb) are related to how we're determining if you can see the name.  (That's probably obvious, but I don't have a better way of describing it.)  Basically, in the current version: 
1. if you can see any of the source clients, you should get the name on the destination client
2. We check if you can see a client in two different ways (one for search, one for show).  These are encapsulated in:
  a. `enrollment_arbiter.clients_source_searchable_to` (search)
  b. `enrollment_arbiter.clients_source_visible_to` (show)

I wonder if going forward we just use the searchable context for deciding if you can see a client when determining if you can see their name.  That is generally the most relaxed context, and as long as we can re-check permission to see the actually PII value, _should_ be a reasonable approach.

ttoomey:
Thanks for working on it @eanders - I pushed up adjustments to this branch over the weekend that seem to work. Sorry I should have pinged you about that and saved you the time.

I think the missing piece was to query the relevant data sources for a client. So the policy for a client is a combination of the project permissions and the permissions they have for those data sources.

As an aside, I had also worked up a version that just uses [collection-based policies](https://github.com/greenriver/hmis-warehouse/blob/4818-add-client-name-permission-with-collection-view/app/models/grda_warehouse/auth_policies/policy_provider.rb#L61). It is much simpler for me to get my head around but it doesn't re-use the existing code and I suspect it's too much of a reach

> I wonder if going forward we just use the searchable context for deciding if you can see a client when determining if you can see their name. That is generally the most relaxed context, and as long as we can re-check permission to see the actually PII value, should be a reasonable approach.


So my understanding is that client name visibility is always based on the can_view_client_name permission rather than inclusion in the visible/searchable client scope. The logic that determines the scopes uses the visible/searchable flags in combination with ROI, etc. My mental model is that these scopes are different/independent from the permissions on a given record. Also name is just one permission that's easy to test with our existing specs but I imagine we want the same logic to apply for all permissions (we don't want special case handling for name?)

eanders:
No worries, on not pinging me, I've been following along.  Just trying to lean on as much existing logic as possible.  Let's revisit this week, sounds like you might have a solid solution, and mine is definitely half baked.

ttoomey:
follow up PR
https://github.com/greenriver/hmis-warehouse/pull/4406

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+    else
+      # TODO: START_ACL remove after ACL migration is complete
+      legacy_user_role_policy
+      # END_ACL
+    end
+  end
+
+  protected
+
+  memoize def for_project_using_acls(project_id)
```
eanders, line 56: @ttoomey does this describe this correctly?  I think my disconnect on the system collection before was that it was just looking for all collections involving the project, regardless of if they were involved in the user's access controls.  That check is performed in `CollectionPolicy`
```suggestion
  # Find all collections that might possibly involve this project,
  # return a collection policy that includes all collections involving the project cross-checked against the user 
  memoize def for_project_using_acls(project_id)
```

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+    p_t = GrdaWarehouse::Hud::Project.arel_table
+    collection_ids = GrdaWarehouse::ProjectCollectionMember.where(project_id: project_id).pluck(:collection_id)
```
eanders: This seems to be assuming the project would be a member of the collection directly.  I'm probably missing it, but where do we check if the project is included in the collection if its organization or data source is included? (CoC Code is covered below.)  If this is already handled, great!  If not, I wonder if we should make a class method or scope on `Collection` that takes a project ID and returns the collections that contain it, just so we can collect up the logic for that in the place where the assignment takes place.

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -0,0 +1,127 @@
...
+    collection_ids += Collection.for_coc_codes(coc_codes).pluck(:id) if coc_codes.any?
+
+    # collections for the client's authoritative data source. Needed for clients records that do not have enrollments
+    collection_ids += GrdaWarehouse::DataSource.authoritative.not_hmis.
+      joins(:group_viewable_entities, :clients).
+      where(gve_t[:collection_id].not_eq(nil)).
+      where(c_t[:id].eq(client_id)).
+      pluck(gve_t[:collection_id])
+
+    collection_ids += system_collection_ids(:data_sources)
```
eanders: Same as with projects, I _think_ this should move to `Collection`, and probably needs to handle the scenario where the entity assigned to the collection is an organization, project group, or data source.

spec/models/auth_policies/policy_provider_spec.rb
```diff
@@ -0,0 +1,199 @@
...
+    def create_authorized_collection(user:, role:)
+      user_group = create(:user_group)
+      user_group.add(user)
+      collection = create(:collection)
+      create :access_control, role: role, collection: collection, user_group: user_group
+      collection
+    end
+
+    shared_examples 'expect authorized access' do
+      it 'has allows access to authorized project' do
```
eanders:
```suggestion
      it 'allows access to authorized project' do
```

app/models/grda_warehouse/auth_policies/policy_provider.rb
ttoomey: The `ProjectCollectionMember` view should be handling that, it collections related through associated entities (joins project to orgs, data source, etc). I added test cases to prove it out in `policy_provider_spec.rb` - do those tests describe the desired behavior?

app/models/grda_warehouse/auth_policies/policy_provider.rb
ttoomey: Yes that's right, the CollectionPolicy has the responsibility for limiting the collections

Maybe I could just reorganize the existing code into `all_collection_ids_for_project` and `all_collection_ids_for_client`? That might make the division of responsibility more clear

```ruby
def for_project_using_acls(project_id)
   # all collections that include the project
   collection_ids = all_collection_ids_for_project(project_id: project_id)
   # policy constrains the collection_ids to only those accessible to the user
   GrdaWarehouse::AuthPolicies::CollectionPolicy.new(user: user, collection_ids: collection_ids)
end

def all_collection_ids_for_project(project_id: project_id)
#...
end

```

app/models/grda_warehouse/auth_policies/policy_provider.rb
ttoomey: > Same as with projects, I think this should move to Collection

I could see an argument for putting the code to gather collections in a model. I'm a little reluctant to do that since there's so many auth methods scattered around, I find it hard to follow the the thread. Keeping the code internal to the class is a trade off, I find it to be more readable but it's definitely less re-usable. I guess I could go either way

> probably needs to handle the scenario where the entity assigned to the collection is an organization, project group, or data source.

That should be handled by ProjectCollectionMember up on line 74

app/models/grda_warehouse/auth_policies/policy_provider.rb
eanders: Yes, that's definitely nicer.  Thanks!

app/models/grda_warehouse/auth_policies/policy_provider.rb
eanders: Right!  I forgot you'd encoded that in the view, perfect solution.

app/models/grda_warehouse/auth_policies/policy_provider.rb
eanders: Yep, thanks, for the reminder!

eanders (approved):

PR #4318, opened 2024-05-03
5961 hmis services hud type idx

ttoomey:

## Description
[Issue 5961](https://github.com/open-path/Green-River/issues/5961) 

More minimal approach to improving hmis service performance

* add index to hud type cols on services
* remove custom service type join from hmis_services_view
* remove custom service association from hmis services and adjust mutations around service creation
* run analyze on tables relevant hud tables 

## Type of change
- [x] Performance fix


drivers/hmis/app/models/hmis/hud/hmis_service.rb
```diff
@@ -75,15 +80,6 @@ class Hmis::Hud::HmisService < Hmis::Hud::Base
...
   end
 
-  scope :matching_search_term, ->(search_term) do
-    return none unless search_term.present?
-
-    search_term.strip!
-    query = "%#{search_term}%"
-    joins(:custom_service_type, :custom_service_categories).
-      where(cst_t[:name].matches(query).or(csc_t[:name].matches(query)))
-  end
```
ttoomey, line 85: I don't think this is used

drivers/hmis/app/graphql/types/hmis_schema/service.rb
```diff
@@ -56,7 +56,18 @@ def sub_type_provided
     end
 
     def service_type
-      load_ar_association(object, :custom_service_type)
+      custom_service_types_scope = Hmis::Hud::CustomServiceType.where(data_source_id: object.data_source_id)
+      case object.owner_type
+      when 'Hmis::Hud::Service'
+        # ideally we'd lazy load this
+        custom_service_types_scope.to_a.detect do |cst|
```
ttoomey, line 63: because we load all custom service types here, the query is the same so rails query cache should mean we aren't actually hitting the db. However this could be made more efficient with a custom data loader

ttoomey:
Test failures showed there's some dependencies on HmisService.custom_service_type in  form_processor. I may need some advice on how to untangle it

drivers/hmis/app/graphql/types/hmis_schema/service.rb
```diff
@@ -8,6 +8,15 @@
 
 module Types
   class HmisSchema::Service < Types::BaseObject
+    # FIXME: order dependent, filter options must be first statement in class
```
ttoomey: @gigxz I started encountering an error running these tests. Have you ever seen this kind of thing?
> "Variable $filters of type ServicesForEnrollmentFilterOptions! was provided invalid value for serviceType (Field is not defined on ServicesForEnrollmentFilterOptions)"

Seems like BaseFilterOptions is getting built before the available_filter_options method is defined. Probably just a local development issue but I thought it was odd that it is only manifesting now.

gigxz: Weird, I've never seen that before. I can reproduce it locally with test `rspec ./drivers/hmis/spec/requests/hmis/enrollment_service_filter_spec.rb` if I move the filter declaration down to where it used to be. I have no idea why it would be coming up now

gigxz (commented):
This looks great @ttoomey thanks!
Looks like there is one more test failure for last_service_date, I went ahead and proposed an approach here: https://github.com/greenriver/hmis-warehouse/pull/4333

drivers/hmis/app/graphql/types/hmis_schema/service.rb
gigxz: Not sure if this should be a `FIXME` comment, though, maybe just a comment

gigxz (approved):
🙌  nice work

PR #4326, opened 2024-05-06
Add status uniqueness and enum constraints to form definitions

martha:

## Description

This PR adds uniqueness constraints based on published/draft status to the Form Definitions table, and constrains statuses to be one of (published, draft, retired) in the model and the schema code. (Though the status attribute of a form definition is not used anywhere in the frontend yet; form definitions are just organized by status when they are returned hanging off of the new form identifier schema object.) 

GH issue: https://github.com/open-path/Green-River/issues/6006
I'm opening this PR against the existing PR for form versioning for now, since there were some TODOs in that code related to this change.

We can merge that one first and then retarget this one.

GH issue open question:
> We may also want to put a where deleted_at IS NULL on those constraints. We don't really have a plan for whether/how to allow Form Definitions to be deleted. (If we want to support deletion at all).

~I'd be happy to add this, but perhaps it's preferable to leave it out to reduce confusion, since we don't currently support deletion at all?~ Edit from checkin: Decided to add this condition; we don't have form definition deletion enabled right now but hypothetically we do usually want the user to be able to undo anything they do in the app.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/hmis_schema/enums/form_status.rb
```diff
@@ -0,0 +1,14 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class HmisSchema::Enums::FormStatus < Types::BaseEnum
+    graphql_name 'FormStatus'
+
+    Hmis::Form::Definition::STATUSES.each_with_index do |status, index|
+      value status, description: "#{index + 1} - #{status.titleize}"
```
gigxz: What's the reasoning behind the numeric value in the description here?
The HUD enums have it because they are stored as integers, but in this case I don't see any relation to this number and the status.

drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -29,6 +29,7 @@ class Forms::FormDefinition < Types::BaseObject
     field :definition, Forms::FormDefinitionJson, null: false
     field :raw_definition, JsonObject, null: false
     field :system, Boolean, null: false
+    field :status, HmisSchema::Enums::FormStatus, null: false, default_value: Hmis::Form::Definition::DRAFT
```
gigxz: Status is non-nullable in the database, so this doesn't need a default value.

We could implement some behavior whereby it resolves "draft" or "invalid" if the stored status is not recognized. I don't think that's necessary, though, since we know that everything is "draft" right now and we will be validating that going forward.

db/warehouse/migrate/20240506204908_add_form_uniqueness_constraints.rb
```diff
@@ -0,0 +1,19 @@
+class AddFormUniquenessConstraints < ActiveRecord::Migration[7.0]
+  def up
+    safety_assured do
+      remove_index :hmis_form_definitions, name: :uidx_hmis_form_definitions_identifier
+      add_index :hmis_form_definitions, [:identifier, :version], unique: true, where: "deleted_at IS NULL", name: :uidx_hmis_form_definitions_identifier
+      add_index :hmis_form_definitions, [:identifier], unique: true, where: "status = 'draft' AND deleted_at IS NULL", name: :uidx_hmis_form_definitions_one_draft_per_identifier
+      add_index :hmis_form_definitions, [:identifier], unique: true, where: "status = 'published' AND deleted_at IS NULL", name: :uidx_hmis_form_definitions_one_published_per_identifier
```
gigxz, line 7: This looks great to me!  Tagging @ttoomey for another pair of eyes. This should represent:
1. `identifier`+`version` are unique together
2. Enforce 0-1 `draft` form per identifier
3. Enforce 0-1 `published` form per identifier

gigxz (approved):
Looks good! Just a couple minor comments.

drivers/hmis/app/graphql/types/hmis_schema/enums/form_status.rb
martha: Good point, I probably just copied and pasted this from somewhere, but you're right, it is neater without the indexes! :)

PR #4327, opened 2024-05-07
Gig/5961 hmis services hud type idx

gigxz:
proposed changes to include in https://github.com/greenriver/hmis-warehouse/pull/4318

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/hmis_service.rb
+++ b/drivers/hmis/app/models/hmis/hud/hmis_service.rb
@@ -90,12 +90,18 @@ class Hmis::Hud::HmisService < Hmis::Hud::Base
     true
   end
 
+  # Helper to initialize the "owner" (Service or CustomService) based on a specified custom_service_type_id.
+  # Initialization happens in SubmitForm when submitting a new service.
   private def initialize_owner
-    raise 'Cannot initialize HmisService without a CustomServiceType' unless custom_service_type.present?
+    raise 'Cannot initialize HmisService without a custom_service_type_id' unless custom_service_type_id.present?
 
+    custom_service_type = Hmis::Hud::CustomServiceType.find(custom_service_type_id)
     attrs = [:enrollment_id, :personal_id, :user_id, :data_source_id, :date_provided].map { |k| [k, send(k)] }.to_h
     if custom_service_type.hud_service?
-      self.owner = Hmis::Hud::Service.new(**attrs)
+      # If this is a HUD service, set RecordType and TypeProvided on the HUD Service record
+      record_type = custom_service_type.hud_record_type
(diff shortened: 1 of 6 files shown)


drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
```diff
@@ -46,12 +46,6 @@ def assign_metadata
       hmis_service = @processor.owner_factory
       hmis_service.assign_attributes(user: @processor.hud_user)
       hmis_service.owner.assign_attributes(user: @processor.hud_user)
-      return unless hmis_service.hud_service?
-
-      hmis_service.owner.assign_attributes(
-        record_type: hmis_service.service_type&.hud_record_type,
-        type_provided: hmis_service.service_type&.hud_type_provided,
-      )
```
gigxz, line 54: Moved this logic to the HmisService initializer. We don't let you change the service type after the fact, so it doesn't need to be here anymore

drivers/hmis/app/models/hmis/hud/hmis_service.rb
```diff
@@ -90,12 +90,18 @@ def readonly?
     true
   end
 
+  # Helper to initialize the "owner" (Service or CustomService) based on a specified custom_service_type_id.
+  # Initialization happens in SubmitForm when submitting a new service, and in BulkAssignService.
   private def initialize_owner
-    raise 'Cannot initialize HmisService without a CustomServiceType' unless custom_service_type.present?
+    raise 'Cannot initialize HmisService without a custom_service_type_id' unless custom_service_type_id.present?
```
gigxz, line 96: The service type is used to determine whether we need to initialize a Service or a CustomService. Changed it to use `custom_service_type_id` since the belongs_to relationship was removed.

drivers/hmis/spec/requests/hmis/submit_form_spec.rb
```diff
@@ -41,7 +41,7 @@
   end
   let!(:hmis_hud_service1) do
     hmis_service = Hmis::Hud::HmisService.find_by(owner: s1)
-    hmis_service.custom_service_type = Hmis::Hud::CustomServiceType.find_by(hud_record_type: s1.record_type, hud_type_provided: s1.type_provided)
+    hmis_service.custom_service_type_id = Hmis::Hud::CustomServiceType.find_by(hud_record_type: s1.record_type, hud_type_provided: s1.type_provided).id
     hmis_service
```
gigxz, line 45: Test setups could definitely be made nicer, but this works

gigxz:
fixes all test failures except the filter tests, didn't look at that one yet.

ttoomey (approved):
Looks great, thank you!

PR #4333, opened 2024-05-08
Fix last_service_date

gigxz:
fixing final test failure `drivers/hmis/spec/requests/hmis/enrollment_last_service_date_spec.rb`


ttoomey (approved):
great, thanks!

PR #4339, opened 2024-05-08
Allow multiple state configuration

eanders:

## Description

This enables the warehouse to support multiple-state installations.  It updates the code around importing and accessing shapes for maps and updates two reports that use those maps.

## Type of change


app/models/grda_warehouse/shape/state_scopes.rb
```diff
@@ -9,8 +9,8 @@ module Shape
     module StateScopes
       extend ActiveSupport::Concern
       included do
-        scope :my_state, -> { where(statefp: my_fips_state_code) }
-        scope :not_my_state, -> { where.not(statefp: my_fips_state_code) }
+        scope :my_states, -> { where(statefp: my_fips_state_codes) }
```
yakloinsteak, line 12: It might be helpful to make sure all the relevant tables index statefp, but I'm not sure if it's actually beneficial.

lib/tasks/us_census_api.rake
```diff
@@ -43,14 +43,18 @@ namespace :us_census_api do
 
   desc "Get the available variables from the US Census"
   task :vars, [] => [:setup, :environment] do
-    importer = GrdaWarehouse::UsCensusApi::Importer.new(years: @years, datasets: @datasets, state_code: @state_code, levels: @levels)
-    importer.bootstrap_variables!
+    @state_codes.each do |state_code|
```
yakloinsteak: I think for the purposes of adding the variables, you don't need to do this twice.

dtgreiner (approved):

shape_files/CoC/make.inserts
```diff
@@ -4,17 +4,17 @@ require 'fileutils'
 require 'pathname'
 require 'zip'
 
-if ARGV[0].to_s.length != 2
-  puts "Set first arg to the two-letter state postal code"
+if ARGV[0].to_s.split(',').map(&:length).uniq != [2]
+  puts 'Set first arg to the two-letter state postal code, or comma delimited list of postal codes'
```
yakloinsteak: I think the added text sounds like you're expecting zip codes.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -234,7 +234,8 @@ def self.coc_picklist(selected_project)
     end
 
     def self.geocodes_picklist
-      state = ENV['RELEVANT_COC_STATE']
+      # NOTE: HMIS currently only supports one state installations
+      state = ENV['RELEVANT_COC_STATE']&.split(',')&.first
```
yakloinsteak: Maybe it's too late, but `ENV['RELEVANT_COC_STATE']&.split(',')` could be a constant somewhere.

yakloinsteak (approved):
This all looks reasonable to me.

app/models/grda_warehouse/shape/state_scopes.rb
eanders: Hmm, maybe, though with a max of 3 states right now, I'm not totally convinced.  Also, I _think_ we may join on the state column (maybe?), but we aren't limiting on it, we're just showing all of the relevant states.  Looks like we already have one on `shape_cocs`, `shape_states`, `shape_counties`, and `state_zip_codes` so mostly covered.  Let's try without the index and revisit if we have any performance issues.

PR #4341, opened 2024-05-08
New CAS Prioritized Client Columns

dtgreiner:
Populate new CAS columns when sync to cas is run. 

## Type of change


app/models/grda_warehouse/tasks/push_clients_to_cas.rb
```diff
@@ -89,6 +89,9 @@ def sync!
               project_client.enrolled_in_rrh = client.enrolled_in_rrh(enrollments)
               project_client.enrolled_in_psh = client.enrolled_in_psh(enrollments)
               project_client.enrolled_in_ph = client.enrolled_in_ph(enrollments)
+              project_client.ongoing_es_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_es&.map { |e| e['project_name'] + ':' + e['date'] }.presence
+              project_client.ongoing_so_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_so&.map { |e| e['project_name'] + ':' + e['date'] }.presence
```
eanders: Should these have a space to the right of the colon?  Or maybe the date should be labeled somehow?

dtgreiner: They should, I was trying to match how the project/date was shown elsewhere, but missed the space. It also looks like I missed some date formatting. With that said, I am open to suggestions on better ways to display the information

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1785,6 +1785,22 @@ def last_projects_served_by(include_confidential_names: false)
...
     end
 
+    def last_intentional_contacts_ignore_visibile_status(include_confidential_names: false, skip_confidential_projects: false, include_dates: false)
+      contacts = processed_service_history&.last_intentional_contacts
+      return [] unless contacts.present?
+
+      contacts = JSON.parse(contacts)
+
+      contacts.sort_by { |c| c['date']&.to_date || 5.years.ago }.reverse.map do |contact|
+        project = GrdaWarehouse::Hud::Project.find(contact['project_id'])
```
eanders: This is going to introduce an N+1 (and sometimes we're loading data for 60k+ clients when pushing them to CAS, I don't have a great solution right now, other than fetching all the projects and passing in a hash of them.  Open to other ideas.

eanders (changes requested):
We need to solve the N+1, but otherwise I think this is good.

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
eanders: Let's go with `: ` for now.  Thanks!

dtgreiner: I've pushed a new commit to address the ":" and date format

app/models/grda_warehouse/hud/client.rb
dtgreiner: I've pushed a new commit to pull the projects ahead of time to clean the N+1

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1785,6 +1785,24 @@ def last_projects_served_by(include_confidential_names: false)
       end.uniq.sort
     end
 
+    def last_intentional_contacts_ignore_visibile_status(include_confidential_names: false, skip_confidential_projects: false, include_dates: false)
+      contacts = processed_service_history&.last_intentional_contacts
+      return [] unless contacts.present?
+
+      projects = GrdaWarehouse::Hud::Project.all.index_by(&:id)
```
eanders: This will need to go outside of the method call or we'll just load them all for each client.  Also, we should confirm `project_id` from `last_intentional_contacts` is `project.id` not `project.ProjectID`

dtgreiner: The id should be correct. This method is set up mimicking the last_intentional_contacts method except with the user parameter pulled out.

I'll plan to move the project data up higher in the method stack to clean it up a bit too.

```
def last_intentional_contacts(user, include_confidential_names: false, skip_confidential_projects: false, include_dates: false)
  contacts = processed_service_history&.last_intentional_contacts
  return [] unless contacts.present?

  visible_projects = user.visible_projects_by_id
  visible_project_ids = user.visible_project_ids_enrollment_context
  contacts = JSON.parse(contacts)

  contacts.sort_by { |c| c['date']&.to_date || 5.years.ago }.reverse.map do |contact|
    project_id = contact['project_id']
    next if project_id.present? && !project_id.in?(visible_project_ids)

    project = visible_projects[project_id]
    next if project&.confidential? && skip_confidential_projects

    name = project&.name(ignore_confidential_status: include_confidential_names) || contact['project_name']
    name += ': ' + contact['date']&.to_date.to_s if include_dates
    name
  end.compact.uniq
end
```

```
def visible_projects_by_id
    @visible_projects_by_id ||= GrdaWarehouse::Hud::Project.find(visible_project_ids_enrollment_context).index_by(&:id)
end
```

dtgreiner: @eanders the project names are being pulled into the method as a parameter based on out call. There were some additional changes needed too as all of the contacts only have the project_id and date. The project names being passed in are having their safe name checked and set ahead of time.

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
```diff
@@ -67,6 +67,10 @@ def sync!
                 :source_eccovia_case_managers,
               ]
             end
+            safe_project_names = {}
+            GrdaWarehouse::Hud::Project.all.each do |project|
```
eanders: Small memory tweak:
```suggestion
            GrdaWarehouse::Hud::Project.find_each do |project|
```

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
```diff
@@ -89,6 +93,9 @@ def sync!
               project_client.enrolled_in_rrh = client.enrolled_in_rrh(enrollments)
               project_client.enrolled_in_psh = client.enrolled_in_psh(enrollments)
               project_client.enrolled_in_ph = client.enrolled_in_ph(enrollments)
+              project_client.ongoing_es_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_es&.map { |e| e['project_name'] + ': ' + e['date']&.to_date.to_s }.presence
+              project_client.ongoing_so_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_so&.map { |e| e['project_name'] + ': ' + e['date']&.to_date.to_s }.presence
```
eanders: Should these project names be made safe?  Maybe we should repeat the logic for these as well, something like:
```suggestion
              project_client.ongoing_es_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_es&.map { |e| safe_project_names[e['project_id']] + ': ' + e['date']&.to_date.to_s }.presence
              project_client.ongoing_so_enrollments = client.processed_service_history&.cohorts_ongoing_enrollments_so&.map { |e| safe_project_names[e['project_id']] + ': ' + e['date']&.to_date.to_s }.presence
```
Unless those are already "safe".

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
dtgreiner: I will have this included in the upcoming commit

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
dtgreiner: These names are already safe (set in GrdaWarehouse::WarehouseClientsProcessed.last_seen_in_type(type)), but setting them here would be more clear and provide some more consistency. I'll run a few quick tests and as long as this comes through as expected, i'll include it in the upcoming commit.

dtgreiner:
@eanders the most recent update requests have been added to the PR.

eanders (approved):
Looks great, thanks @dtgreiner !

PR #4343, opened 2024-05-08
Rename renameServiceType to updateServiceType

martha:

## Description
https://github.com/open-path/Green-River/issues/6012
This PR
- renames the existing `rename_service_type` mutation to `update_service_type`
- adds the ability to update the `supports_bulk_assignment` attribute, which wasn't editable before.

Open question:
- Should we use the [ServiceTypeInput](https://github.com/greenriver/hmis-warehouse/blob/0212c91c3a96a3e9f02a20e7c82bed7791ffcdfd/drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb#L8) here? I opted against it in this first pass because it contains only 3 attribures, 1 of which isn't relevant -- we don't want to let users update the service type's service_category_id after the fact, if I understand correctly.

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/hmis_schema/mutation_type.rb
```diff
@@ -41,7 +41,7 @@ class HmisSchema::MutationType < Types::BaseObject
 
     field :create_service_type, mutation: Mutations::CreateServiceType
     field :delete_service_type, mutation: Mutations::DeleteServiceType
-    field :rename_service_type, mutation: Mutations::RenameServiceType
+    field :update_service_type, mutation: Mutations::UpdateServiceType
```
gigxz: Does it work to do something like this? We should be in the habit of deprecating instead of removing/renaming as much as possible
```
    field :rename_service_type, mutation: Mutations::UpdateServiceType, deprecation_reason: "Renamed to updateServiceType"
    field :update_service_type, mutation: Mutations::UpdateServiceType
```

drivers/hmis/app/graphql/mutations/update_service_type.rb
```diff
@@ -4,19 +4,21 @@
...
+    graphql_name 'UpdateServiceType'
 
     argument :id, ID, required: true
     argument :name, String, required: true
+    argument :supports_bulk_assignment, Boolean, required: true
 
     field :service_type, Types::HmisSchema::ServiceType, null: true
 
-    def resolve(id:, name:)
+    def resolve(id:, name:, supports_bulk_assignment:)
```
gigxz, line 16: This looks great to me (re: your question about using a separate input type)

drivers/hmis/app/graphql/types/hmis_schema/mutation_type.rb
martha: Yes this works great. Good idea to just use the same UpdateServiceType mutation.

gigxz (approved):
👍

PR #4129, opened 2024-03-16
187087981 link id migration script

ttoomey:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

[PT issue](https://www.pivotaltracker.com/story/show/187087981)

* Migrate LinkIDs so they can be used as valid identifiers. This allows us to use the identifiers as selectors and variable names in conditional statements.
* Adds validation to the form definition json schema to try and enforce valid identifiers going forward

### Questions:
Our seed data files will also need to be transformed. Right now the seed files in drivers/hmis/lib/form_dat are not valid. I could parse and update them but it's bit complicated due to fragments. Or we could leave the ids in the seed files as-is and do the transform as part of the `HmisUtil::JsonForms`. That's probably less work but might be confusing to maintain

I picked went with the following transforms but these need review for any unintended side-effects. Do these transforms seem okay?
* dots become underscores. HUD ref numbers are prefixed with 'hud' `4.04.3 => hud_4_04_3`
* hyphens become camel case `ssn-group => ssnGroup`
* numbers are prefixed with 'Q' `1 => q_1`

## Type of change
[//]: # 'remove options that are not relevant'
- [x] New feature (adds functionality)

## Checklist before requesting review
- [ ] I have performed a self-review of my code
- [ ] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [ ] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [ ] My code follows the style guidelines of this project (rubocop)
- [ ] I have updated the documentation (or not applicable)
- [ ] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4344, opened 2024-05-09
Script to migrate LinkIDs for form files in the repo

gigxz:
## Description

Changes to merge into https://github.com/greenriver/hmis-warehouse/pull/4129

* add code that updates the form JSON files that are committed to the repo (`update_form_data_files`). The diff looked okay to me (didn't commit it here but you can see the diff on https://github.com/greenriver/hmis-warehouse/pull/4344/commits/5f85452c1cac6e8390f59eb0a38015ccd6a2ff17)
* change to snake case instead of camel casing

## Type of change


drivers/hmis/spec/support/form_helpers.rb
```diff
@@ -35,317 +35,194 @@ def add_item_to_definition(definition, item)
     definition
   end
 
-  def completed_form_values_for_role(role)
-    yield(COMPLETE_VALUES[role]) if block_given?
+  # get mocked "hud_values" (keyed by field name) and generate "values" (keyed by link id)
+  def mock_form_values_for_definition(definition)
```
gigxz: These changes can be pushed totally separately, I'm just trying to do the most low-effort way to remove the dependency on link ids from this test. This seems to work.

drivers/hmis/lib/tasks/one_time_migration_20240303.rake
```diff
@@ -74,8 +98,8 @@ class OneTimeMigration20230303
     value = value =~ /\A[0-9]+\z/ ? "q_#{value}" : value
     # leading number with a period seems to be HUD ref numbers like "2.02.6"
     value = value =~ /\A[^a-z]+\./i ? "hud_#{value}" : value
-    # camelize hyphenated strings or spaces
-    value = value.gsub(/(-| )+([a-z])/) { Regexp.last_match(2).upcase }
+    # snakecase hyphenated strings or spaces
+    value = value.gsub(/(-| )+/, '_')
     # replace dots and lingering hyphens with an underscore
     value = value.gsub(/[-.]/, '_')
```
gigxz, line 104: I felt like it was unnecessarily inconsistent to camelcase some and snake case others, so just made it snake case on both. That ok?

gigxz:
Tests fail because:
* https://github.com/greenriver/hmis-warehouse/pull/4129 implements the validation change, and I didn't comment the form changes (see reverted commits)
* ~~Also, some tests still rely on link ids (submit form spec (pr up) and definition spec)~~

drivers/hmis/lib/tasks/one_time_migration_20240303.rake
ttoomey: Probably is fine- my thinking was that it would be nice if it was possible to reverse. If both '.' and '-' become underscores than it's ambiguous what the original key value was. Maybe that doesn't matter though and the consistency has more value.

drivers/hmis/lib/tasks/one_time_migration_20240303.rake
```diff
@@ -15,6 +15,26 @@ class OneTimeMigration20230303
     update_form_definitions
   end
 
+  # Migrate LinkIDs for all files in drivers/hmis/lib/form_data. Write changes to files.
+  # Doesn't catch everything, a few things need to be manually fixed (client-id and veteran-status)
+  def update_form_data_files
```
ttoomey, line 20: this isn't called from perform, we'd just call this manually and commit the changes

ttoomey (approved):
Looks good, thank you!

PR #4366, opened 2024-05-14
Require client alerts to have expiry dates

martha:

## Description

This PR implements TCHC's requested changes: https://github.com/open-path/Green-River/issues/6044
- Make expiration date for alerts required
- Give existing non-expiring alerts expiration dates of 1 month after alert creation 
- Set a max on alert expiration dates to 10 years in the future

## Type of change
- [ ] Bug fix
- [ ] Dependency update


db/warehouse/migrate/20240514131401_require_client_alert_expiry.rb
```diff
@@ -0,0 +1,17 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class RequireClientAlertExpiry < ActiveRecord::Migration[7.0]
+  def up
+    unexpired = Hmis::ClientAlert.with_deleted.where(expiration_date: nil)
+    unexpired.update_all("expiration_date=created_at + interval '1 month'")
+
+    safety_assured { change_column_null :hmis_client_alerts, :expiration_date, false }
```
martha: I wanted to double check about this `safety_assured` -- I know that's something we have done in the past, but would also be happy to implement it as suggested by strong migrations (add a check constraint and validate it in a separate migration). Would those migrations need to be in two different releases?

db/warehouse_structure.sql
```diff
@@ -150,8 +150,8 @@ CREATE TYPE public.record_type AS ENUM (
 
 CREATE FUNCTION public.f_unaccent(text) RETURNS text
     LANGUAGE sql IMMUTABLE STRICT PARALLEL SAFE
-    AS $_$
-SELECT public.unaccent('public.unaccent', $1)  -- schema-qualify function and dictionary
+    AS $_$
```
martha: These changes are unrelated, I was trying to figure out how to revert them but they aren't showing up as changes at all in my local ide! Anyway, I think nothing is really changing here and not sure why github detects it as changed.

db/warehouse/migrate/20240514131401_require_client_alert_expiry.rb
gigxz: Hmm. This migration would be fine, but I think this database change is a little too high-commitment, actually.

If down the line we have another installation that _doesn't_ want to use Expiration dates on alerts at all, we will want to accommodate that alongside this by using a different client_alert form (i.e. changing `"required": true`). (I'm not suggesting making that configurability change now, but we may want to make it later. I know right now the form is `static`).

Does that make sense? I think as a lower-commitment change, we can just manually validate the presence of the expiration date in `CreateClientAlert`. Then it'll be easier later on to make the expiration date required-ness configurable.

db/warehouse_structure.sql
gigxz: It shouldn't be anything to worry about as long as the tests are still passing. We all have minor differences in this file and it gets formatted differently if you open it in an ide, etc.

drivers/hmis/app/graphql/types/hmis_schema/client_alert.rb
```diff
@@ -13,7 +13,7 @@ class HmisSchema::ClientAlert < Types::BaseObject
     field :note, String, null: false
     field :created_at, GraphQL::Types::ISO8601DateTime, null: false
     field :created_by, Types::Application::User, null: true
-    field :expiration_date, GraphQL::Types::ISO8601Date, null: true
+    field :expiration_date, GraphQL::Types::ISO8601Date, null: false
```
gigxz: Assuming the db change is reverted, this should also remain nullable. Following the "strict-on-input"/"lax-on-output" approach.

db/warehouse/migrate/20240514131401_require_client_alert_expiry.rb
martha: Thanks, good point - I kind of wondered about this, should have brought it up before implementation! 🤦

gigxz (approved):
thanks!

PR #4393, opened 2024-05-23
Fix bugs in find_definition_for_role logic

gigxz:

## Description

Update FormDefinition selection to take into account:
* forms that have only `inactive` rules should never be chosen
* forms that have non-`system` default rules should be chosen over forms that have `system` default rules
* forms with multiple versions (soon this should only look at `published` but for now it looks at latest versions of each form. see TODOs for https://github.com/open-path/Green-River/issues/6147)

Code cleanup:
* removed unnecessary `version:` arg for form selection. We never want/need to specify which version to use.

Closes https://github.com/open-path/Green-River/issues/6143

## Type of change
- [x] Bug fix


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -233,34 +234,37 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
...
-    else
-      # order so that detect_best_instance_for_project will use version as a tie-breaker if multiple instances match
-      definition_scope = definition_scope.order(version: :desc, id: :desc)
-    end
+  #  * return the published version of the definition that is referenced by that instance
+  def self.for_project(project:, role:, service_type: nil)
+    # Consider all Active, Published Forms for this Role
+    definition_scope = Hmis::Form::Definition.with_role(role).
+      active. # Drop definitions that have no active rules
+      latest_versions # TODO(#6147): Switch from `latest_versions` to `published` scope
```
ttoomey, line 244: should this use TodoOrDie for 2 months from now or something?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -233,34 +234,37 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
...
+    # Filter by Service Type if specified
     definition_scope = definition_scope.for_service_type(service_type) if service_type.present?
-    base_scope = Hmis::Form::Instance.joins(:definition).merge(definition_scope)
 
-    # Choose the first scope that has any records. Prefer more specific instances.
-    instance = base_scope.detect_best_instance_for_project(project: project)
-    instance ? where(identifier: instance.definition_identifier) : none
-  end
+    # Scope of active Instances (Rules) that are are associated with the eligible definitions
+    instance_scope = Hmis::Form::Instance.joins(:definition).merge(definition_scope).active
```
ttoomey, line 250: So definition_scope has the active scope applied (line 243) meaning it's joined to instances already. Then we join that into instances again here. This should work but the sql seems a bit hard to read.

Maybe a subquery would be better than a join for the active scope? This also avoids getting duplicates as a side-effect from join 
```ruby
  scope :active, -> do
    where(identifier: Hmis::Form::Instance.active.select(:definition_identifier))
  end
```

ttoomey (approved):
Looks great, thanks!

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: I added a specific note to grep for `TODO(#6147)` in this issue https://github.com/open-path/Green-River/issues/6147 so I think we should be okay. There are 3 places that use this comment tag.

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: @ttoomey great point, thanks for catching that. I like making the definition `active` scope use a sub-query. Updated

PR #4397, opened 2024-05-24
Add SSVF sections to the Data Quality Tool report

dtgreiner:

## Description

Add SSVF enrollment & client data to the Data Quality Tool report.

## Type of change


drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -209,7 +220,7 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       report_item.data_source_id = enrollment.data_source_id
       report_item.project_id = enrollment.project.id
       report_item.project_name = enrollment.project.name(report.user)
-      project_type =  enrollment.project.project_type_to_use
+      project_type = enrollment.project.project_type_to_use
```
eanders: this can just be `enrollment.project.project_type` now.  (With the last iteration of overrides, `project_type_to_use` became a long convoluted way of getting to `project_type`.)

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -231,6 +242,13 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       report_age_date = [enrollment.EntryDate, report.filter.start].max
       report_item.age = enrollment.client.age_on(report_age_date)
       report_item.ch_at_entry = enrollment.chronically_homeless_at_start?
+      report_item.percent_ami = enrollment.PercentAMI
+      report_item.vamc_station = enrollment.VAMCStation
+      report_item.veteran = client.veteran?
```
eanders: pretty sure `report_item.veteran` is an integer, so probably:
```suggestion
      report_item.veteran = client.veteran_status
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -365,6 +390,31 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
...
+      return false if enrollment.PrisonDischarge.blank? || enrollment.PrisonDischarge == 99 # V7.J
+      return false if enrollment.SexOffender.blank? || enrollment.SexOffender == 99 # V7.K
+      return false if enrollment.DisabledHoH.blank? || enrollment.DisabledHoH == 99 # V7.L
+      return false if enrollment.CurrentPregnant.blank? || enrollment.CurrentPregnant == 99 # V7.M
+      return false if enrollment.SingleParent.blank? || enrollment.SingleParent == 99 # V7.N
+      return false if enrollment.DependentUnder6.blank? || enrollment.DependentUnder6 == 99 # V7.O
+      return false if enrollment.HH5Plus.blank? || enrollment.HH5Plus == 99 # V7.P
+      return false if enrollment.CoCPrioritized.blank? || enrollment.CoCPrioritized == 99 # V7.Q
+      return false if enrollment.HPScreeningScore.blank? # V7.R
+      return false if enrollment.ThresholdScore.blank? # V7.S
```
eanders: Where does this logic come from?  Can you add a comment pointing to the source doc on the HUD Exchange maybe?  Also, is there any utility in checking it with something like:
```
enrollment.TimeToHousingLoss.in?(known_values)
```
That would catch blanks, 99s, and any value that is irrelevant.  (Maybe we're doing that somewhere else?)

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -278,8 +299,12 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       annual_income_assessment = annual_assessment(enrollment, hoh.first_date_in_program)
       exit_income_assessment = enrollment.income_benefits_at_exit
 
+      report_item.total_monthly_income_at_entry = entry_income_assessment&.TotalMonthlyIncome
+      report_item.total_monthly_income_from_source_at_entry = entry_income_assessment&.all_sources_and_amounts&.sum { |_k, amount| amount || 0 }
```
eanders, line 321: Do we need to check if any of these sources are blank?  Sometimes a missing value is indicative of a failure to collect the info rather than a `0`  (Not sure that's relevant here though.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -250,6 +268,9 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       report_item.head_of_household_count = hh&.select(&:head_of_household?)&.count || 0
       report_item.household_type = household_type(report_item.household_min_age, report_item.household_max_age)
 
+      report_item.hh_veteran_count = hh.select { |hm| hm.client.veteran? }.count || 0
+      report_item.hoh_veteran = hh&.select { |hm| hm.head_of_household && hm.client.veteran? }&.count&.> 0 || false
```
eanders: Is this the same?
```suggestion
      report_item.hoh_veteran = hh&.any? { |hm| hm.head_of_household && hm.client.veteran? }
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -439,6 +489,10 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       false
     end
 
+    def self.adult?(item)
```
eanders, line 513: If we're going to define this, let's use it in the `hoh_or_adult?` method just below.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: I pulled the fields from V7 HP Targeting Criteria in the [FY 2024 HMIS Data Dictionary](https://files.hudexchange.info/resources/documents/HMIS-Data-Dictionary-2024.pdf), then mapped the fields to the enrollment data based on [drivers/hmis/lib/form_data/default/fragments/v7_hp_targeting_criteria.json](https://github.com/greenriver/hmis-warehouse/blob/ad2ee5ab286e771bea1746c90fd024e1698f1448/drivers/hmis/lib/form_data/default/fragments/v7_hp_targeting_criteria.json). The json file also includes the reference to the hud list with the relevant responses.

I'll add a comment around this in the method.

For the known_values, the values are all from the HudLists2024 data which includes the 99 values. I may be missing something, but to do this we would be recreating a bunch of the lists. We could possibly create a method that takes a relevant hud list as an argument and strips the '99' responses so we only get the valid codes.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: I believe it is, I'll give it a test run to verify

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: My interpretation of the check this is being used in was that it is only looking to see if the total monthly amount matched the sum from the sources, so nils would be the same as zero's. I could definitely be wrong and happy to check more into it.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: will include this in the next commit

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: will update `hoh_or_adult?` to use this in the next commit

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+            ].flatten.include?(item.project_type)
+          },
+          limiter: ->(item) {
+            return false unless hoh?(item)
+            return false unless [
+              HudUtility2024.performance_reporting[:rrh],
+              HudUtility2024.performance_reporting[:prevention],
+            ].flatten.include?(item.project_type)
+
+            item.percent_ami.blank?
```
eanders: this needs to check for `99` as well, which is a valid value, but indicates it's missing.  Maybe better to check for it being in an accepted value.

```
! item.percent_ami.in?(HudUtility2024.percent_amis.keys - [99])
```
or something like that.

We could adjust the description, to say missing, not collected, or not an accepted value.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          detail_columns: default_detail_columns + [
+            :target_screen_required,
+            :target_screen_completed,
+          ],
+          denominator: ->(item) { hoh?(item) && HudUtility2024.performance_reporting[:prevention].include?(item.project_type) },
+          limiter: ->(item) {
+            return false unless hoh?(item)
+            return false unless HudUtility2024.performance_reporting[:prevention].include?(item.project_type)
+            # Screening Required flag is missing
+            return true if item.target_screen_required.blank?
```
eanders: This also needs to check for 99 (or better yet that it's not in the acceptable values)

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+              HudUtility2024.performance_reporting[:es_entry_exit],
+              HudUtility2024.performance_reporting[:th],
+              HudUtility2024.performance_reporting[:psh],
+              HudUtility2024.performance_reporting[:so],
+              HudUtility2024.performance_reporting[:sh],
+              HudUtility2024.performance_reporting[:oph],
+              HudUtility2024.performance_reporting[:rrh],
+            ].flatten.include?(item.project_type)
+
+            item.vamc_station.blank?
```
eanders: This also needs to check for 99 (or better yet that it's not in the acceptable values)

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          detail_columns: default_detail_columns + [
+            :living_situation,
+          ],
+          denominator: ->(item) { hoh_or_adult?(item) },
+          limiter: ->(item) {
+            return false unless hoh_or_adult?(item)
+
+            item.living_situation.blank? || item.living_situation == 99
+          },
+        },
```
eanders: Pretty sure this is already captured in `living_situation_issues`.  Let's adjust that logic a bit to exclude 99 from the current check, which currently counts it as acceptable I think.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+        homeless_living_situation_issues: {
+          title: 'HP - Homeless Living Situation',
+          description: 'HP adult has a homeless prior living situation',
```
eanders:
```suggestion
          description: 'Adults in Homeless Prevention should not be indicating they are homeless prior immediately prior to entry.',
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          limiter: ->(item) {
+            return false unless adult?(item)
+            return false unless HudUtility2024.performance_reporting[:prevention].include?(item.project_type)
+
+            HudUtility2024.homeless_situations(as: :prior).include?(item.living_situation)
+          },
+        },
+        non_homeless_living_situation_issues: {
+          title: 'RRH - Non-Homeless Living Situation',
+          description: 'RRH adult has a non-homeless prior living situation',
```
eanders:
```suggestion
          description: 'Adults entering RRH must indicate they were homeless prior to entry.',
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          required_for: 'Adults in RRH at entry',
+          detail_columns: default_detail_columns + [
+            :living_situation,
+          ],
+          denominator: ->(item) { adult?(item) && HudUtility2024.performance_reporting[:rrh].include?(item.project_type) },
+          limiter: ->(item) {
+            return false unless adult?(item)
+            return false unless HudUtility2024.performance_reporting[:rrh].include?(item.project_type)
+
+            HudUtility2024.situation_type(item.living_situation) == 'Homeless'
```
eanders: For consistency, isn't this the same (and similar to the HP check?)
```suggestion
            ! HudUtility2024.homeless_situations(as: :prior).include?(item.living_situation)
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          ],
+          denominator: ->(item) { item.income_at_entry_expected == true },
+          limiter: ->(item) {
+            return false unless item.income_at_entry_expected == true
+
+            has_any_source_of_income_at_entry = item.income_from_any_source_at_entry.present? && item.income_from_any_source_at_entry == 1
+            has_monthly_income_at_entry = item.total_monthly_income_at_entry.present? && item.total_monthly_income_at_entry > 0
+
+            # any source at entry XOR reported monthly income at entry
+            has_any_source_of_income_at_entry ^ has_monthly_income_at_entry
```
eanders: why not? 
```
has_any_source_of_income_at_entry != has_monthly_income_at_entry
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          },
+        },
+        no_veteran_in_household: {
+          title: 'No Veteran in Household',
+          description: 'No member of the household is a veteran',
+          required_for: 'Household',
+          detail_columns: default_detail_columns + [
+            :hh_veteran_count,
+          ],
+          denominator: ->(_item) { true },
```
eanders: I _think_ this is only valid for some funding sources.  So the denominator will need to account for that.  Not every enrollment ever needs a veteran.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +674,224 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+          },
+        },
+        hoh_not_veteran: {
+          title: 'Head of Household is not a Veteran',
+          description: 'The Head of Household is is not a veteran',
+          required_for: 'Household',
+          detail_columns: default_detail_columns + [
+            :relationship_to_hoh,
+          ],
+          denominator: ->(_item) { true },
```
eanders: Same here, the denominator needs to limit by funding source.

eanders (changes requested):
@dtgreiner looks great! I dropped a bunch of questions inline, let me know if you want to discuss synchronously.

dtgreiner:
@eanders I've pushed a new commit that includes the suggested changes and also accounts for the different funders on each item

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -209,8 +228,9 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       report_item.data_source_id = enrollment.data_source_id
       report_item.project_id = enrollment.project.id
       report_item.project_name = enrollment.project.name(report.user)
-      project_type =  enrollment.project.project_type_to_use
+      project_type = enrollment.project.project_type
       report_item.project_type = project_type
+      report_item.funders = enrollment.project.funders.pluck(:Funder).map(&:to_i)
```
eanders: Hmm, I'm not sure how rails handles a pluck like this.  I know you can pluck from an array (and a scope) so it might do the right thing and just grab it out of the array, but it might be safer (from an N+1 perspective, since you've already preloaded) to just:
```suggestion
      report_item.funders = enrollment.project.funders.map(&:Funder).reject(&:blank).map(&:to_i)
```
I'm also not sure you want to call `to_i` on it without first rejecting the empty values.  If you don't you'll probably get some unintentional `0`s, since `nil.to_i` and `''.to_i` result in `0`.

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -250,6 +286,9 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
       report_item.head_of_household_count = hh&.select(&:head_of_household?)&.count || 0
       report_item.household_type = household_type(report_item.household_min_age, report_item.household_max_age)
 
+      report_item.hh_veteran_count = hh.select { |hm| hm.client.veteran? }.count || 0
```
eanders: Looking up a couple of lines, I think we need to catch the situation where `hh` is `nil.
```suggestion
      report_item.hh_veteran_count = hh&.select { |hm| hm.client.veteran? }&.count || 0
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -365,6 +408,32 @@ def report_item_fields_from_enrollment(report_items:, enrollment:, report:) # ru
...
+      return true unless (HudUtility2024.incarcerated_adults.keys - [99]).include?(enrollment.IncarceratedAdult) # V7.I
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.PrisonDischarge) # V7.J
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.SexOffender) # V7.K
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.DisabledHoH) # V7.L
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.CurrentPregnant) # V7.M
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.SingleParent) # V7.N
+      return true unless (HudUtility2024.dependent_under_6_options.keys - [99]).include?(enrollment.DependentUnder6) # V7.O
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.HH5Plus) # V7.P
+      return true unless (HudUtility2024.yes_no_missing_options.keys - [99]).include?(enrollment.CoCPrioritized) # V7.Q
```
eanders: Let' DRY this up just a wee bit:
```suggestion
      valid_yes_no_missing_options = HudUtility2024.yes_no_missing_options.keys - [99]
      return true unless (HudUtility2024.time_to_housing_losses.keys - [99]).include?(enrollment.TimeToHousingLoss) # V7.A
      return true unless (HudUtility2024.annual_percent_amis.keys - [99]).include?(enrollment.AnnualPercentAMI) # V7.B
      return true unless (HudUtility2024.literal_homeless_histories.keys - [99]).include?(enrollment.LiteralHomelessHistory) # V7.C
      return true unless valid_yes_no_missing_options.include?(enrollment.ClientLeaseholder) # V7.D
      return true unless valid_yes_no_missing_options.include?(enrollment.HOHLeaseholder) # V7.E
      return true unless valid_yes_no_missing_options.include?(enrollment.SubsidyAtRisk) # V7.F
      return true unless (HudUtility2024.eviction_histories.keys - [99]).include?(enrollment.EvictionHistory) # V7.G
      return true unless valid_yes_no_missing_options.include?(enrollment.CriminalRecord) # V7.H
      return true unless (HudUtility2024.incarcerated_adults.keys - [99]).include?(enrollment.IncarceratedAdult) # V7.I
      return true unless valid_yes_no_missing_options.include?(enrollment.PrisonDischarge) # V7.J
      return true unless valid_yes_no_missing_options.include?(enrollment.SexOffender) # V7.K
      return true unless valid_yes_no_missing_options.include?(enrollment.DisabledHoH) # V7.L
      return true unless valid_yes_no_missing_options.include?(enrollment.CurrentPregnant) # V7.M
      return true unless valid_yes_no_missing_options.include?(enrollment.SingleParent) # V7.N
      return true unless (HudUtility2024.dependent_under_6_options.keys - [99]).include?(enrollment.DependentUnder6) # V7.O
      return true unless valid_yes_no_missing_options.include?(enrollment.HH5Plus) # V7.P
      return true unless valid_yes_no_missing_options.include?(enrollment.CoCPrioritized) # V7.Q
```

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -620,6 +693,297 @@ def self.sections(report) # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticC
...
+            funding_sources = [
+              HudUtility2024.funder_components['VA: SSVF'],
+            ].flatten
+            project_types = [
+              HudUtility2024.performance_reporting[:rrh],
+              HudUtility2024.performance_reporting[:prevention],
+            ].flatten
+
+            # Only hohs that include the above funder(s) and project type(s)
+            hoh?(item) && ! (funding_sources & item.funders).empty? && project_types.include?(item.project_type)
```
eanders: Is this the same as:
```suggestion
            hoh?(item) && project_types.include?(item.project_type) && (funding_sources & item.funders).any? 
```
Two proposed changes there, one to push the project type check up since I _think_ that's less work for ruby, and the second to use `any?` instead of `! (..).empty?)` just to try and make it a bit more readable.

dtgreiner:
@eanders I have the new suggestions updated and pushed to the PR. Thanks!

eanders (approved):
Thanks @dtgreiner !

PR #4400, opened 2024-05-24
Resolve referral_date as date

gigxz:
## Description

`referral_date` column is a `date` in the db. We were resolving it as a datetime which resolves as `2024-05-24T00:00:00+00:00` which makes it show up as `5/23/24` in the UI (which I believe is a bug https://github.com/greenriver/hmis-frontend/pull/784). But, we should be resolving it as a date in the first place.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/mutations/ac_hmis/create_outgoing_referral_posting.rb
+++ b/drivers/hmis/app/graphql/mutations/ac_hmis/create_outgoing_referral_posting.rb
@@ -36,7 +36,7 @@ module Mutations
 
       referral = HmisExternalApis::AcHmis::Referral.new(
         enrollment: enrollment,
-        referral_date: Time.current,
+        referral_date: Date.current,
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/graphql/mutations/ac_hmis/create_outgoing_referral_posting.rb
```diff
@@ -36,7 +36,7 @@ def resolve(input:)
 
       referral = HmisExternalApis::AcHmis::Referral.new(
         enrollment: enrollment,
-        referral_date: Time.current,
+        referral_date: Date.current,
```
gigxz, line 39: This was never being stored as a Time anyway, since the column is a date

eanders (approved):

ttoomey (approved):
thanks for fixing this

PR #4401, opened 2024-05-26
Client name permissions - alternate caching approach

eanders:

## Description

WIP

## Type of change


app/models/user.rb
```diff
@@ -132,9 +121,11 @@ def populate_external_reporting_permissions!
   def viewable_project_ids(context)
     return GrdaWarehouse::Hud::Project.project_ids_viewable_by(self, permission: context) if Rails.env.test?
 
-    Rails.cache.fetch("#{user_project_id_prefix}_#{context}", expires_in: EXPIRY_MINUTES.minutes) do
+    @viewable_project_ids ||= {}
+    @viewable_project_ids[context] ||= Rails.cache.fetch("#{user_project_id_prefix}_#{context}", expires_in: EXPIRY_MINUTES.minutes) do
       GrdaWarehouse::Hud::Project.project_ids_viewable_by(self, permission: context)
     end
+    @viewable_project_ids[context]
```
eanders, line 128: @ttoomey I'm going to close this PR since yours appears to work just fine, but do you think it makes sense to test this change to `viewable_project_ids` (in memory memoization without memoist?)  I think the report version is probably issuing a query to redis for this data for each client row.

ttoomey: @eanders - sorry I forgot to mention it but I did actually memoize `user.viewable_project_ids` as part of 4818. I used `memoize` rather than an instance variable. Is there reason not use use `memoize`?

See [`app/models/user.rb:218`](https://github.com/greenriver/hmis-warehouse/pull/4311/files#diff-9802ca3c9c4cf89904fd44bc114e35ebdf2c5dd3d5b645491e2b253e1afef29bR121)

The memoization did cause initial test failures; I addressed those by calling `user.clear_memery_cache!` in [`setup_access_control`](https://github.com/greenriver/hmis-warehouse/pull/4311/files#diff-b729166ea878c77af1f6a135521edd6e32221056db63201e41fa4de9b90ed7dcR11)

eanders: Ah, I thought you'd pulled it out because of the test failures.  My fear of using memoize is that I'm unclear how long that cache lasts.  If someone changes the set of projects associated with the user, how long does it take for that to be reflected?  My understanding is that memoist caches that for an indeterminate amount of time, but I could be very wrong.  We want to be sure we fetch the set of project_ids on each page load so that if someone changes the associated collection, we get the new set of ids.  What's the chance this is memoized beyond a single page load?

[I hadn't looked at the memoist source code before](https://github.com/tycooon/memery/blob/master/lib/memery.rb), but I think it's basically doing the same thing, so it's probably fine, just wanted to double check.

eanders: Hmm, doesn't seem to behave as I'd expect.  Even after re-fetching the user object, it doesn't reload the project IDs.  
Screenshot 2024-06-01 at 5 15 32 PM
Can we add a `ttl` to the memoization of this method?  `1.minutes` should be plenty to prevent calling it tons of times, but small enough that no one would notice if they changed the associated project set.

eanders: Ugh, just realized there's a redis cache in there. Of course it wouldn't re issue the query.  I think we're all set then with just the memoize. Thanks

eanders:
Closing, memoize works as expected

app/models/user.rb
ttoomey: Yeah my understanding is that memery/memoize is an instance-variable cache. I would be very surprised if the cache persisted beyond of the object scope

PR #4403, opened 2024-05-28
Only alert if theres an error in unit availability job

gigxz:


drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/update_unit_availability_job.rb
```diff
@@ -40,7 +40,7 @@ def perform(force: false)
         end
       end
 
-      handle_alert("Failed to sync #{failed_updates} capacity updates to LINK.")
+      handle_alert("Failed to sync #{failed_updates} capacity updates to LINK.") unless failed_updates.zero?
```
eanders: Is `failed_updates` always an integer?  If so, maybe:
```suggestion
      handle_alert("Failed to sync #{failed_updates} capacity updates to LINK.") if failed_updates.positive?
```

If it can be `nil` this probably wants an `&`

eanders (approved):
Looks fine as is, but I think it'd be more readable as noted.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/update_unit_availability_job.rb
gigxz: Yes it's initialized to zero

PR #4404, opened 2024-05-28
Modify Form Definition update mutation

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6103

This PR edits the mutation `UpdateFormDefinition` to accept the frontend's input data format.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/mutations/update_form_definition_typed.rb
```diff
@@ -0,0 +1,42 @@
...
+    argument :id, ID, required: true
+    argument :definition, 'Types::Admin::FormDefinitionTypedInput', required: false
+    # TODO(#6005) - add role and title as inputs
+    # argument :role, Types::Forms::Enums::FormRole, required: false
+    # argument :title, String, required: false
+
+    field :form_definition, Types::Forms::FormDefinition, null: true
+
+    def resolve(id:, definition:)
+      raise 'not allowed' unless current_user.can_manage_forms?
```
gigxz:
```suggestion
      access_denied! unless current_user.can_manage_forms?
```

drivers/hmis/app/graphql/mutations/update_form_definition_typed.rb
```diff
@@ -0,0 +1,42 @@
...
+
+      form_definition = Hmis::Form::Definition.find_by(id: id)
+      raise 'not found' unless form_definition
+
+      form_definition.definition = definition
```
gigxz: Huh, I'm surprised that this works without calling `definition.to_params` to convert it from a `GraphQL::Schema::InputObject` into a ruby hash. I'm unsure how this behaves (is it documented?) so it might be safe to convert it to params first. (If we stick with this approach)

drivers/hmis/app/graphql/types/admin/form_item_input.rb
```diff
@@ -0,0 +1,43 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  # This is the GraphQL input version of Types::Forms::FormItem. This input type is needed because
+  # in graphql, attributes on input types can only be built-in scalar types or enums.
+  class Admin::FormItemInput < Types::BaseInputObject
+    argument :link_id, String, 'Unique identifier for item', required: false
```
gigxz: Ahh sigh, when we discussed this approach earlier, I didn't consider the fact that we can't actually re-use `Types::Forms::FormItem` as the input type. Part of the benefit, in my mind, is that we could use the existing type. But you're right that's [not possible](https://spec.graphql.org/October2021/#note-7a6d2).


It feels redundant to have the Form structure defined in three places:
1) The [JSON Schema](https://github.com/greenriver/hmis-warehouse/blob/release-118/drivers/hmis_external_apis/public/schemas/form_definition.json) which is used for validating the form structure
2) `Types::Forms::FormItem` which is used for resolving the form
3) `Types::Admin::FormItemInput` which is used for updating the form


I'm thinking that for _updating the form_, we already rely on the JSON Schema (1) to validate the input, when we call validate_schema. Is it really necessary to ALSO have the GraphQL layer validate the input shape (3)?

The old mutation (`UpdateFormDefinition`) just accepts the form as a string, and then validates it using the JSON Schema. I feel like we COULD stick with that approach, the only difference we need is for the mutation to `.underscore` all the keys in the json string that it receives, prior to validating and saving it.

That would let us: (1) continue to rely on the JSON Schema for validation, and (2) let the frontend continue to use the GraphQL Typed (camelCased) form shape. The frontend just needs to stringify it before sending it to the mutation. Is that a viable path, or am I missing something? Sorry for the waffling here

gigxz (commented):
some comments 👀 maybe worth a huddle tomorrow @martha!

drivers/hmis/app/graphql/types/admin/form_item_input.rb
gigxz: If that's not possible, I also looked into generating the `Admin::FormItemInput` class which seems very doable

martha: Hmm! Yes I see what you mean. Lemme investigate and see if the path you're describing is possible. For sure, it would be good to pair later today.

martha: Ok, I took a stab at implementing your suggestion. It is a bit more complicated than simply calling .underscore, because of having to remove junk like `__typename` (which previously I was doing on the frontend here: https://github.com/greenriver/hmis-frontend/pull/785/commits/bb35bc177bf36507e53bb77a7dfec92b339eba2f#diff-7d33d3e9a02ac3ebc1f517b289868d09de568774095409317f30b95d4c1d1341R278-R306)

I'd be ok to keep doing that on the frontend, if you prefer the backend should never know about that cruft, but I think I prefer it this way where only one or the other (backend or frontend) have to implement this recursive validation. What do you think?

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -12,13 +12,14 @@ class UpdateFormDefinition < CleanBaseMutation
...
     def resolve(id:, input:)
-      raise 'not allowed' unless current_user.can_manage_forms?
+      access_denied! unless current_user.can_manage_forms?
 
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition
       raise 'not allowed to change identifier' if input.identifier.present? && input.identifier != definition.identifier
 
       definition.assign_attributes(**input.to_attributes)
+      definition.definition = convert_form_definition(JSON.parse(input.definition))
```
martha: To do: Write some comments explaining how we landed here

drivers/hmis/spec/requests/hmis/update_form_definition_spec.rb
```diff
@@ -0,0 +1,50 @@
...
...bleBehavior":"ANY","enableWhen":null,"autofillValues":null},{"__typename":"FormItem","item":null,"linkId":"emergency_contact_name","type":"STRING","component":null,"prefix":null,"text":"Emergency Contact Name 1","briefText":null,"readonlyText":null,"helperText":null,"required":false,"warnIfEmpty":false,"hidden":false,"readOnly":false,"repeats":false,"mapping":{"__typename":"FieldMapping","recordType":null,"fieldName":null,"customFieldKey":"emergency_contact_name"},"pickListReference":null,"size":null,"assessmentDate":null,"prefill":false,"bounds":null,"pickListOptions":null,"initial":null,"dataCollectedAbout":null,"disabledDisplay":"HIDDEN","enableBehavior":"ANY","enableWhen":null,"autofillValues":null}],"linkId":"event-group","type":"GROUP","component":null,"prefix":null,"text":"SPDAT","briefText":null,"readonlyText":null,"helperText":null,"required":false,"warnIfEmpty":false,"hidden":false,"readOnly":false,"repeats":false,"mapping":null,"pickListReference":null,"size":null,"assessmentDate":null,"prefill":false,"bounds":null,"pickListOptions":null,"initial":null,"dataCollectedAbout":null,"disabledDisplay":"HIDDEN","enableBehavior":"ANY","enableWhen":null,"autofillValues":null}]}',
```
gigxz: For the sake of https://github.com/open-path/Green-River/issues/5916, lets make these `linkId`s compliant with the  constraints we plan to add. So no periods, no hyphens, and no starting with numbers.

drivers/hmis/app/graphql/mutations/update_form_definition.rb
gigxz: I don't think we need the extra `convert_form_definition` function, since it's doing the same thing. This should suffice:
```suggestion
      definition.definition = convert_form_item(JSON.parse(input.definition))
```

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -35,5 +36,41 @@ def resolve(id:, input:)
...
+      {
+        item: definition['item'].map do |i|
+          convert_form_item(i)
+        end,
+      }
+    end
+
+    def convert_form_item(item)
+      converted = basic_convert(item)
+      converted['mapping'] = basic_convert(item['mapping']) if item['mapping']
```
gigxz: This needs to account for all other attributes that can contain nested objects, so they get transformed correctly. This should be all of them:

mapping
bounds
pick_list_options
enable_when
initial
autofill_values
autofill_values => autofill_when

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -35,5 +36,41 @@ def resolve(id:, input:)
...
+
+      input.keys.each do |key|
+        next if key == '__typename'
+        next if input[key].nil?
+
+        converted[key.underscore] = input[key]
+      end
+
+      converted
+    end
```
gigxz: This method could be simplified a bit by using rails hash methods:
```
      input.excluding('__typename').
        compact. # drop keys with nil values
        transform_keys(&:underscore) # snake_case keys
```

I'd also suggest renaming `basic_convert` to indicate that it is for hashes specifically, maybe `transform_hash` or something.

It would also be possible to use `.deep_transform_keys(&:underscore)` which can be called once and transform the keys on the whole definition. But we need to traverse it anyway to do the `compact`, so doing the key transformation here seems fine. There is no handy helper for doing a ["deep compact"](https://stackoverflow.com/questions/56457020/ruby-how-do-i-remove-nil-empty-values-from-a-nested-hash-and-why-arent-my-atte). Doing it the way you have -- with separate cases for transforming `mapping` and any other nested objects -- seems like a fine approach to me!

gigxz (commented):
@martha this looks good to me!
Left a few suggestions on cleaning up the transformation code, and request updating it to transform all nested objects on items (not just `mapping`).

drivers/hmis/app/graphql/mutations/update_form_definition.rb
martha: Wow beautiful! Thanks for the simplification suggestion!

drivers/hmis/app/graphql/mutations/update_form_definition.rb
martha: I ended up feeling like a transformation function that's totally agnostic to the specific keys it needs to transform was a bit more graceful, because we separately check the input against the expected json schema. Let me know if you think this looks ok!

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -35,5 +43,34 @@ def resolve(id:, input:)
...
+      # First drop unneeded keys and transform them all to snake case
+      converted = transform_hash(form_element)
+
+      # Then map through all the elements in the result object
+      converted.keys.each do |key|
+        if converted[key].is_a?(Array)
+          # If it's an array, recursively transform each element in the array
+          converted[key] = converted[key].map do |element|
+            recursively_transform(element)
+          end
```
gigxz: I don't think this will handle [an array of strings](https://github.com/greenriver/hmis-warehouse/blob/0f38f55bdaea72c641137925bfbfd0984bffe258/drivers/hmis/app/graphql/types/forms/enable_when.rb#L21) well. Maybe if  `recursively_transform` receives an `element` that is not a hash, it should return the element directly? Or, only call `recursively_transform` if element is a hash.

drivers/hmis/app/graphql/mutations/update_form_definition.rb
gigxz: looks good to me. I think there was just the issue of handling primitive arrays. I'm also not sure how it will handle empty arrays/objects, and whether that matters.

drivers/hmis/app/graphql/mutations/update_form_definition.rb
martha: Oh true, thanks for pointing out the issue with arrays of strings. I updated it and added a regression test too. Hopefully this version is more robust!

drivers/hmis/app/graphql/mutations/update_form_definition.rb
martha: I think empty arrays/objects should be okay, the loop on the array (or the hash's keys) will just do nothing

gigxz (approved):
looks good! thanks!

gigxz:
Looks like there were some test failures on `update_form_definition_spec` @martha

drivers/hmis/spec/requests/hmis/update_form_definition_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+    expect(result.dig('errors')).to be_empty
+    expect(result.dig('data', 'updateFormDefinition', 'errors')).to be_empty
+  end
+
+  it 'should work when the definition does not need to be converted' do
+    input = {
+      definition: '{"item":[{"item":[{"text":"Assessment Date","type":"DATE","hidden":false,"link_id":"link_1","prefill":false,"repeats":false,"required":false,"read_only":false,"warn_if_empty":false,"assessment_date":true,"enable_behavior":"ANY","disabled_display":"HIDDEN"},{"text":"Emergency Contact Name","type":"STRING","hidden":false,"link_id":"emergency_contact_name","mapping":{"custom_field_key":"emergency_contact_name"},"prefill":false,"repeats":false,"required":false,"read_only":false,"warn_if_empty":false,"enable_behavior":"ANY","disabled_display":"HIDDEN"}],"text":"SPDAT","type":"GROUP","hidden":false,"link_id":"event_group","prefill":false,"repeats":false,"required":false,"read_only":false,"warn_if_empty":false,"enable_behavior":"ANY","disabled_display":"HIDDEN"}]}',
+    }
+    _response, result = post_graphql(id: 30, input: input) { mutation }
+    expect(result.dig('errors')).to be_empty
```
gigxz: I have found `expect(response.status).to eq(200), result.inspect` is a nice approach, it will log the result if the response is not 200:

```
     Failure/Error: expect(response.status).to eq(200), result.inspect
       {"errors"=>[{"message"=>"not found", "backtrace"=>["/app/drivers/hmis/app/graphql/mutations/update_form_definition.rb:18:in `resolve'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/schema/resolver.rb:119:in `public_send'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/schema/resolver.rb:119:in `call_resolve'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/schema/mutation.rb:69:in `call_resolve'
...
```

It's not finding a definition with id `30`. You'll need to create a definition using the `hmis_form_definition` factory

martha: Thank you. 🙏 did you find this by running it locally? It succeeded for me locally and I'm curious about why that is -- even when I added 
```
  before(:all) do
    cleanup_test_environment
  end
```

gigxz: Probably because the `fd1` fixture happened to have database id `30` for you locally. I'm guessing thats how you came up with the number 30 in the first place :)

The IDs on those fixtures will not be consistent.
One way you can clear out the forms in your test db is `rails c -e test`, `Hmis::Form::Definiton.delete_all`.

gigxz (approved):

PR #4406, opened 2024-05-28
4818 add client name permission with collection view

ttoomey:
A conceptually simpler approach to policies that does not wrap enrollment arbiter.


ttoomey:
@eanders follow up on our conversations today
*  There's a test that suggests the system "all data sources" data source is relevant for determining permission. See ./drivers/client_access_control/spec/requests/client_visibility_spec.rb:38. Calling maintain_system_groups doesn't seem to matter here. The test is adding roles directly to the system data source; perhaps this test is incorrect? The system data source seems to be used in the ProjectCoc.viewable_by scope
* Unfortunately we can't include the Collection.coc_codes in the project_collection_members view because it crosses the database boundary. So we need additional queries to to get the collections related to projects via coc_code

eanders:
@ttoomey that test is correct, it is explicitly testing that a user with **full** access to all data sources:
```
setup_access_control(user, can_view_clients, Collection.system_collection(:data_sources))
setup_access_control(user, can_search_own_clients, Collection.system_collection(:data_sources))
```
Can see all the clients.

So, yes the system collection for all data sources is used for granting permission, but it should not be assumed to always be present when checking access to a client.

Too bad about the CoC Codes, but that makes sense.  Thanks for checking.

app/models/grda_warehouse/auth_policies/policy_provider.rb
```diff
@@ -53,33 +53,41 @@ def for_project(project_or_id)
...
 
   memoize def for_project_using_acls(project_id)
-    GrdaWarehouse::AuthPolicies::ProjectPolicy.new(user: user, project_id: project_id)
+    p_t = GrdaWarehouse::Hud::Project.arel_table
+    collection_ids = GrdaWarehouse::ProjectCollectionMember.where(project_id: project_id).pluck(:collection_id)
+    coc_codes = GrdaWarehouse::Hud::ProjectCoc.
+      joins(:project).
+      where(p_t[:id].eq(project_id)).
+      pluck(:coc_code)
+    collection_ids += Collection.including_coc_codes(for_coc_codes) if coc_codes.any?
```
eanders: maybe?
```suggestion
    collection_ids += Collection.including_coc_codes(for_coc_codes).pluck(:id) if coc_codes.any?
```

ttoomey: ah oops, thanks!

ttoomey:
> So, yes the system collection for all data sources is used for granting permission, but it should not be assumed to always be present when checking access to a client.

Thanks - is there a condition where we wouldn't want to include it? It sounds like what I had before might have been correct (if the user has can_view/search and view_name permission for the system data source then they can view/search and see the name for all clients in the system)

eanders:
> > So, yes the system collection for all data sources is used for granting permission, but it should not be assumed to always be present when checking access to a client.
> 
> Thanks - is there a condition where we wouldn't want to include it? It sounds like what I had before might have been correct (if the user has can_view/search and view_name permission for the system data source then they can view/search and see the name for all clients in the system)
We would only want to include it if the user had access to that collection through one of their Access Controls with a role granting the appropriate permission.

ttoomey:
> We would only want to include it if the user had access to that collection through one of their Access Controls with a role granting the appropriate permission.

Great, I think what we have is right then

PR #4412, opened 2024-05-30
Show headshot from HMIS in WH

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6074

This PR adds code that returns the source client image, if one has been uploaded to the HMIS. If both exist, it should still return the warehouse image.

## Type of change
- [ ] Bug fix
- [ ] Dependency update

- [ ] My code follows the style guidelines of this project (rubocop) - there is a non-rubocop-compliant line in this file that I haven't changed, and am a bit hesitant to mess with


app/models/concerns/client_image_consumer.rb
```diff
@@ -50,6 +50,9 @@ def local_client_image_data
       headshot = uploaded_local_image
```
martha, line 35: I was curious about the requirement that an ETO api be available, in this file on line 18?

app/models/concerns/client_image_consumer.rb
```diff
@@ -84,6 +87,10 @@ def set_local_client_image_cache(image_data) # rubocop:disable Naming/AccessorMe
       client_files.window.tagged_with('Client Headshot').order(updated_at: :desc).limit(1)&.first
     end
 
+    private def source_local_image
```
martha: I was also wondering whether this should be made available as an attribute on `GrdaWarehouse::Hud::Client`?

app/models/concerns/client_image_consumer.rb
eanders: Ah, that's because that's the only API we have historically integrated with.  Looks like we check it again on line 27.  We probably need to pull out the call on line 18.  Thanks!  I'd prefer to wrap the ETO specific logic into a method or ruby class.  But we can do that later.

eanders: Oh, I see, sorry, to waffle so much, that was limiting it so no images show up unless they "have" the ETO API...  There are definitely some installations where it doesn't make sense to have a fake images.  Let's try this for now, but we may need some other configuration to check if we should just hide images, even in staging.

martha: Ah, ok. Should I revert that removal?

app/models/concerns/client_image_consumer.rb
```diff
@@ -10,46 +10,43 @@ module ClientImageConsumer
...
-    def image(cache_for = 10.minutes) # rubocop:disable Lint/UnusedMethodArgument
-      return nil unless GrdaWarehouse::Config.get(:eto_api_available)
-
-      # Use an uploaded headshot if available
-      faked_image_data = local_client_image_data || fake_client_image_data
-      return faked_image_data unless Rails.env.production?
-
-      # Use the uploaded client image if available, otherwise use the API, if we have access
+    # Returns actual image bytes. Cache time limit not yet implemented.
+    def image(_cache_for = 10.minutes)
```
martha, line 14: @eanders after staring at the code for a few more minutes, I was wondering whether this method `image` should actually use `image_for_source_client` below as a subroutine? They seem to repeat some of the same logic.

This is basically untested pseudocode but I can fix it up if you think it's a good idea -- something like,
```
def image
  image_data = local_client_image_data
  image_data ||= image_for_source_client
  image_data || self.class.no_image_on_file_image
```

eanders: I think the reason we avoided that is because `image` is executed against a destination client, so we need to loop over the source clients.  Might be cleaner to replace:
```
 # Check for HMIS uploaded images
       # Queries client_files once for every source client - not ideal, but there are never that many
       image_data ||= source_clients.detect(&:local_hmis_image)&.as_thumb

       # If we still don't have an image at this point, look it up in the ETO API
       image_data ||= source_clients.detect(&:eto_source_image_data)
```
with 
```
image_data ||= source_clients.detect(&:image_for_source_client)
```

I'm not sure if detect could match `''` or if we'd need to return `nil` from `image_for_source_client`.

That does seem somewhat cleaner though, so feel free to take a stab at it!

eanders (approved):
Thanks @martha, this looks great to me, a huge improvement for readability!

gigxz (approved):
nice!

eanders:
@martha do you think it's ok to target this to 119?

PR #4413, opened 2024-05-30
Fix bug with large integer search

martha:

## Description

GH ticket: https://github.com/open-path/Green-River/issues/6068

This PR changes the behavior of Project, Organization, and Enrollment search so that, when the input is an integer, it will only try to match that to a primary key of the entity being searched IF the integer is small enough that it could actually be a primary key (less than 4 bytes).

This behavior was already present on the client search, [here](https://github.com/greenriver/hmis-warehouse/blob/47cfee51730ef5fa07ba5c0557440d4f1d1579bc/app/models/concerns/client_search.rb#L26-L27), which is why the client search functionality wasn't impacted by this bug, as the ticket notes. I did touch that code in order to move everything into a central helper, which I put on `Hmis::Hud::Base` but would be happy to move if there's a better place for it.

For testing, verify that the 4 cases identified in the ticket linked above now work correctly.

I was also wondering, could you help me identify where in the UI we have a free text search for Enrollments that uses the Enrollment search term filter? I'd like to test this too.

## Type of change
- [x] Bug fix
- [ ] Dependency update


app/models/concerns/client_search.rb
```diff
@@ -22,9 +22,7 @@ def self.text_searcher(text, sorted:, resolve_for_join_query: false)
 
       # should never match
       never_cond = sa[:id].eq(nil)
-
-      max_pk = 2_147_483_648 # PK is a 4 byte signed INT (2 ** ((4 * 8) - 1))
-      term_is_possibly_pk = numeric ? text.to_i < max_pk : false
+      term_is_possibly_pk = possibly_pk?(text)
```
gigxz: This concern is shared by the Warehouse and HMIS `Client` models -- so I think this will break for the warehouse search, since `possibly_pk?` won't be defined.

I can't think of a good place for this method to go for code-sharing, except maybe the [ApplicationHelper](https://github.com/greenriver/hmis-warehouse/blob/release-118/app/helpers/application_helper.rb). (@eanders thoughts on that?)

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -148,11 +148,10 @@ def set_hud_project_id_from_project_pk_unless_wip
...
 
     alpha_numeric = /[[[:alnum:]]-]+/.match(search_term).try(:[], 0) == search_term
-    numeric = /[\d-]+/.match(search_term).try(:[], 0) == search_term
 
-    # If numeric, check if it's an Enrollment primary key
-    if numeric
-      matching_enrollments = where(id: search_term)
+    # If it's a possible PK, check if it's an Enrollment primary key
+    if possibly_pk?(search_term)
+      matching_enrollments = where(id: search_term.to_i)
```
gigxz, line 154: > I was also wondering, could you help me identify where in the UI we have a free text search for Enrollments that uses the Enrollment search term filter? I'd like to test this too.

I believe this is when you go to the Project Dashboard => Enrollments => Clients and search. (And maybe you can do it on households table too, I forget).

drivers/hmis/app/models/hmis/hud/organization.rb
```diff
@@ -36,7 +36,12 @@ class Hmis::Hud::Organization < Hmis::Hud::Base
 
     search_term.strip!
     query = "%#{search_term.split(/\W+/).join('%')}%"
-    where(o_t[:OrganizationName].matches(query).or(o_t[:id].eq(search_term)).or(o_t[:organization_id].eq(search_term)))
+
+    where(
+      o_t[:OrganizationName].matches(query).
+        or(o_t[:id].eq(possibly_pk?(search_term) ? search_term : '')).
+        or(o_t[:organization_id].eq(search_term)),
+    )
```
gigxz: This feels wrong to be checking `o_t[:id].eq('')` as a no-op. I think it would be clearer to just omit that condition if `search_term` is not a possible pk. If you build up an array of conditions, you can chain them like this `where([ cond1, cond2, cond3].inject(&:or))`

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -121,7 +121,12 @@ class Hmis::Hud::Project < Hmis::Hud::Base
 
     search_term.strip!
     query = "%#{search_term.split(/\W+/).join('%')}%"
-    where(p_t[:ProjectName].matches(query).or(p_t[:id].eq(search_term)).or(p_t[:project_id].eq(search_term)))
+
+    where(
+      p_t[:ProjectName].matches(query).
+        or(p_t[:id].eq(possibly_pk?(search_term) ? search_term : '')).
```
gigxz: same comment hear on neatness of `''` usage

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -152,6 +157,7 @@ def self.sort_by_option(option)
   end
 
   def self.apply_filters(input)
+    # binding.pry
```
gigxz: this got left behind

drivers/hmis/spec/requests/hmis/organization_spec.rb
```diff
@@ -68,5 +68,12 @@
       expect(orgs.length).to eq(1)
       expect(orgs.first['organizationName']).to eq(o1.organization_name)
     end
+
+    it 'does not error when a large integer is passed' do
+      response, result = post_graphql(filters: { search_term: '73892738928' }) { query }
+      expect(response.status).to eq 200
+      orgs = result.dig('data', 'organizations', 'nodes')
+      expect(orgs.length).to eq(0)
```
gigxz, line 76: thanks for the reg tests!

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Thanks! I was confused about this because I couldn't see the bug get reproduced, even before I made the change in `enrollment.rb`.

I just stepped through it in a debugger and realized that this line,
```
matching_enrollments = where(id: search_term.to_i)
```
doesn't throw an error when search_term is '73892738928'. I was surprised by that because I couldn't see any difference in the database definitions of the ID columns between the Enrollment and Project models. Am I missing something very obvious?

app/models/concerns/client_search.rb
martha: Ah, I see. Thanks for pointing that out.

I will revert it to a non-shared version for now, I think, but would be happy to move into ApplicationHelper if you think that is the best move here.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: @martha good point. I can't find the rails documentation for this, but it does look throwing in a `to_i` on the project search (`p_t[:id].eq(search_term.to_i)`) fixes the issue for project search as well (i.e. no error is thrown).

I strangely can't get a minimal repro working for this. The query that fails in the UI works in the console. But both fail with `.to_sql`.

```
irb(main):007:0> Hmis::Hud::Project.where(Hmis::Hud::Project.arel_table[:id].eq('73892738928'))
=> []
irb(main):008:0> Hmis::Hud::Project.where(Hmis::Hud::Project.arel_table[:id].eq('73892738928')).to_sql
/bundle/ruby/3.1.0/gems/activemodel-7.0.8.1/lib/active_model/type/integer.rb:52:in `ensure_in_range': 73892738928 is out of range for ActiveModel::Type::Integer with limit 4 bytes (ActiveModel::RangeError)
irb(main):009:0> Hmis::Hud::Project.where(Hmis::Hud::Project.arel_table[:id].eq('73892738928'.to_i)).to_sql
/bundle/ruby/3.1.0/gems/activemodel-7.0.8.1/lib/active_model/type/integer.rb:52:in `ensure_in_range': 73892738928 is out of range for ActiveModel::Type::Integer with limit 4 bytes (ActiveModel::RangeError)
```

Soo... not totally sure what's going on, but I think the approach of a `possibly_pk?` method is good/safest.

app/models/concerns/client_search.rb
gigxz: I think ApplicationHelper would make sense for the below fn, we seem to put misc things in there. I don't know if we have guidance around that @eanders @ttoomey  so I would say go for it.
```
def possibly_pk?(search_term) # could add optional arg for 4 byte vs 8 byte, if needed later
    numeric = /[\d-]+/.match(search_term).try(:[], 0) == search_term
    return false unless numeric

    max_pk = 2_147_483_648 # PK is a 4 byte signed INT (2 ** ((4 * 8) - 1))
    search_term.to_i < max_pk
end
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Huh, yeah that is surprising. I have no idea why either. One interesting note is that in the rails console the first query returns `[]`, but when stepping through in a debugger, calling the same query will return an `ActiveRecord_Relation` object, which only throws an error when you try to call something on it (like `.count` or `.compact`).

I did also notice in the debugger that calling `.to_i` in the context of Project still results in this error. So I'm really not sure what is different about Enrollments.

Confusing :/ but anyway, I agree that it's best to consistently use the `possibly_pk` approach for all 3.

app/models/concerns/client_search.rb
martha: Seems like `ApplicationHelper` is only automagically included in Rails views. There are a couple places where we explicitly `include ApplicationHelper`, but that's not quite working here -- getting the error,
```
undefined method `possibly_pk?' for #<ActiveRecord::Relation ... >
```

gigxz (changes requested):
LGMT without https://github.com/greenriver/hmis-warehouse/pull/4413/commits/fa13319dbceae39162008158e0e689e09610f782

Requested review from Elliot to see if he has additional feedback about where the code could live as a shared function

app/helpers/application_helper.rb
```diff
@@ -464,4 +464,13 @@ def bracket_small_population(value, mask: true)
...
   end
+
+  # Determine whether the given search term is possibly a Primary Key (it's numeric and less than 4 bytes)
+  def possibly_pk?(search_term) # could add optional arg for 4 byte vs 8 byte, if needed later
+    numeric = /[\d-]+/.match(search_term).try(:[], 0) == search_term
+    return false unless numeric
+
+    max_pk = 2_147_483_648 # PK is a 4 byte signed INT (2 ** ((4 * 8) - 1))
+    search_term.to_i < max_pk
+  end
```
eanders: Maybe this should go on `ApplicationRecord` as a class method?  Then you could call it:
```
ApplictaionRecord.possibly_pk?(1234)
```
wherever you need it.  That also means any class that inherits from ApplicationRecord would have the method, so enrollment (for instance) could just call `possibly_pk?(1234)` in the context of a  scope or class method.

ApplicationHelper is really intended for use in views.

martha: That's a good point. Thanks for helping us find a place for this @eanders !

app/models/application_record.rb
```diff
@@ -16,6 +16,15 @@ def self.needs_migration?
     ActiveRecord::Migration.check_pending!
   end
 
+  # Determine whether the given search term is possibly a Primary Key (it's numeric and less than 4 bytes)
+  def self.possibly_pk?(search_term) # could add optional arg for 4 byte vs 8 byte, if needed later
+    numeric = /[\d-]+/.match(search_term).try(:[], 0) == search_term
```
ttoomey: I could be missing something but this is trying to see if the whole term is numeric right? I think it would be clearer to write this as:
```suggestion
    numeric = search_term =~ /\A\d+\z/
```

or maybe more concisely:
```ruby
  MAX_PK = 2_147_483_648 
  def self.possibly_pk?(search_term)
   search_term =~ /\A\d+\z/ && search_term.to_i < MAX_PK
  end
```

app/models/application_record.rb
```diff
@@ -16,6 +16,15 @@ def self.needs_migration?
     ActiveRecord::Migration.check_pending!
   end
 
+  # Determine whether the given search term is possibly a Primary Key (it's numeric and less than 4 bytes)
+  def self.possibly_pk?(search_term) # could add optional arg for 4 byte vs 8 byte, if needed later
```
ttoomey: I don't think this should go in ApplicationRecord. That seems like a very general bucket as every model in the app inherits from it. Also our tables have inconsistencies around pks so this may not be the correct logic for every model.

Since it's so simple I'd be tempted to either inline it into the scopes where it's used or drop it in a bag of utility functions like `HudUtility.term_is_possibily_pk?(term)`

eanders: How about `Hmis::Hud::Base` or `Hmis::Hud::Concerns::Shared`?  This appears to be something we'd only use in the `Hud` namespace, and isn't used on the warehouse side right?

gigxz: It is used on the warehouse side because the client search concern uses it.Or we could just inline that specific use-case.

eanders: Thanks then I _think_ `HmisStructure::Base` fits the bill.

app/models/application_record.rb
martha: Thanks @ttoomey good point, I lifted this uncritically from client_search here: https://github.com/greenriver/hmis-warehouse/blob/47cfee51730ef5fa07ba5c0557440d4f1d1579bc/app/models/concerns/client_search.rb#L19

Do you think that both should be cleaned up? as it stands this PR doesn't touch client_search, but that was an oversight on my part; if we do decide to put this code somewhere in common per the discussion in the thread below, then I think that client_search should be updated too.

ttoomey: Ah, that makes sense! The ClientSearch concern is quite complex; the current pattern is including negative numbers for some reason.

You could try DRYing it up if you have high confidence that it's not changing behavior. Personally I think leaving it unchanged also reasonable given that it is complex and a bit risky to change

martha: Thanks, Theron. I wouldn't say I have high confidence, so I do think I'd like to leave it unchanged if other reviewers are okay with that solution.

app/models/application_record.rb
martha: Based on the proposal to leave the client search concern unchanged, per the thread above, I've moved it to `Hmis::Hud::Base`.

gigxz (approved):

eanders (approved):
Thanks @martha !

ttoomey (approved):
looks good, thanks!

PR #4416, opened 2024-05-31
Ignore null DateProvided in AutoExitJob

gigxz:

## Description

Bug: AutoExitJob would throw an error if Service.DateProvided was nil
Fix: ignore services where date provided is nil

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -28,10 +28,10 @@ module Hmis
           most_recent_contact = if project.es_nbn? # Night-by-night Emergency Shelter
             # For NBN shelters, the most recent contact is the last bed night.
             # If the client had no bed nights, use the enrollment (entry date) as the last contact.
-            enrollment.services.bed_nights.order(:date_provided).last || enrollment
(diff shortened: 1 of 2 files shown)


martha (approved):

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -28,10 +28,10 @@ def perform
           most_recent_contact = if project.es_nbn? # Night-by-night Emergency Shelter
             # For NBN shelters, the most recent contact is the last bed night.
             # If the client had no bed nights, use the enrollment (entry date) as the last contact.
-            enrollment.services.bed_nights.order(:date_provided).last || enrollment
+            enrollment.services.bed_nights.where.not(date_provided: nil).order(:date_provided).last || enrollment
```
eanders, line 31: I wonder if we should add a `valid_service` scope to Service (or something like that)?

gigxz: It might be hard to pin down what a "valid" service is. For example HUD Services that have record type 12 or 13 (deprecated), or type provided that doesn't match, etc. In this case, I think those should still be considered contacts based on their date.

eanders: Sorry, I was thinking "structurally valid"

PR #4420, opened 2024-05-31
Persist form processor for non-assessment records

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6111

* add polymorphic owner
* make custom_assessment_id optional, and stop using it (remove later)
* update SubmitForm to persist the `FormProcessor`
* update `ServiceProcessor` to process underlying Service/CustomService records, rather than HmisService records. This was a necessary change because we need to store the relationship to the service as the `owner` of the FormProcessor
* move service init out of HmisService view
* fix issue with delete assessment, switch to destroy_dependent_records fn

- [x] add tests for persisting related records on non-assessment owners

## Type of change


drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -37,11 +38,16 @@ class Hmis::Form::FormProcessor < ::GrdaWarehouseBase
...
 
-  attr_accessor :owner, :hud_user, :current_user
+  attr_accessor :hud_user, :current_user
+
+  def custom_assessment
+    owner if owner_type == Hmis::Hud::CustomAssessment.sti_name
+  end
 
-  def run!(owner:, user:)
+  # TODO probably want to persist the user_id too
```
gigxz: x

db/warehouse_structure.sql
```diff
@@ -3855,426 +3117,6 @@ CREATE VIEW public."bi_Organization" AS
   WHERE ("Organization"."DateDeleted" IS NULL);
 
 
---
--- Name: bi_Project; Type: VIEW; Schema: public; Owner: -
```
gigxz, line 3859: should I have all these bi views locally?

db/warehouse_structure.sql
```diff
@@ -18387,15 +17226,17 @@ CREATE TABLE public.hmis_form_processors (
...
     hud_values jsonb,
     youth_education_status_id integer,
     employment_education_id integer,
     current_living_situation_id integer,
     ce_assessment_id bigint,
-    ce_event_id bigint
+    ce_event_id bigint,
+    owner_type character varying NOT NULL,
+    owner_id bigint NOT NULL
 );
```
gigxz, line 17240: relevant portion

gigxz:
This is tentatively ready for review. Waiting to see if there are any tests that I missed.

drivers/hmis/app/models/hmis/hud/concerns/has_custom_data_elements.rb
```diff
@@ -8,6 +8,9 @@ module Hmis::Hud::Concerns::HasCustomDataElements
   extend ActiveSupport::Concern
 
   included do
+    # throwing this in here because if you have CDEs you are most likely submittable via a form- maybe move to its own concern
+    has_one :form_processor, class_name: 'Hmis::Form::FormProcessor', as: :owner, dependent: :destroy
+
```
gigxz, line 13: do one of
* remove comment
* change name of concern to something like `FormSubmittable` that encompasses both CDEs and form processor
* move to a new concern

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -213,15 +236,15 @@ def ce_assessment_factory(create: true)
 
     return ce_assessment if ce_assessment.present? || !create
 
-    self.ce_assessment = enrollment_factory.assessments.build(user_id: custom_assessment&.user_id)
+    self.ce_assessment = enrollment_factory.assessments.build
```
gigxz, line 241: These all were redundant because assign_metadata handles setting the user_id based on the current user that is submitting.

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -154,20 +161,34 @@ def current_living_situation_factory(create: true)
...
+  # Enrollment is a special case, because it is not referenced by the FormProcessor.
+  # Enrollment can be the owner, or it can be related to the owner.
+  # Examples of forms that may update Enrollment:
+  #  - set EntryDate from intake assessment (owner_type=CustomAssessment)
+  #  - enroll an existing client (owner_type=Enrollment)
+  #  - update Move-in Date at occurrence (owner_type=Enrollment)
+  #
+  # In some other cases, the enrollment_factory can be used indirectly, to determine the relationship between records.
+  # An example is a CustomCaseNote form that generates a CurrentLivingSituation. The form does not update Enrollment directly,
+  # but it relies on the enrollment_factory to determine which enrollment to use when generating a new CLS (current_living_situation_factory)
```
gigxz: This is weird so figured I'd explain it. Maybe I could come up with something cleaner. I hesitated to dive to far into "storing related records for non-assessment form submissions" since that's **not** what this ticket was originally about, but I did add one test for it to `submit_form_spec` (which tests the CaseNote scenario described above). 🌀

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -135,12 +140,14 @@ def ce_assessment?
   end
 
   def store_assessment_questions!
+    return unless owner_type == Hmis::Hud::CustomAssessment.sti_name && ce_assessment?
```
eanders:
```suggestion
    return unless  custom_assessment? && ce_assessment?
```

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -401,6 +426,25 @@ def related_records
     end.compact.uniq
   end
 
+  def destroy_related_records!
```
eanders, line 431: Might want to add a comment here that this will need to be updated along with the data standards.  I'm tempted to suggest we move this list into `HudUtility2024` to enforce that. Thoughts?

db/warehouse/migrate/20240531152432_add_form_processor_owner.rb
```diff
@@ -0,0 +1,10 @@
+class AddFormProcessorOwner < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction! # needed for index creation
+
+  # rails db:migrate:up:warehouse VERSION=20240531152432
+  # rails db:migrate:down:warehouse VERSION=20240531152432
+  def change
+    # note: owner will be made non-nullable at a later step
+    add_reference :hmis_form_processors, :owner, polymorphic: true, index: { algorithm: :concurrently, unique: true }, null: true
```
ttoomey: This is probably okay but it might better to add the index in the last migration in the series. My understanding is that pg will try and build this index async, while the other migrations are running. Even though the col is all NULL, I think it still does a full scan to build the index. So it could end up running the updates in `20240603190227` while it tries to index the same data. That's likely less efficient that updating the col and then indexing

drivers/hmis/app/models/hmis/hud/concerns/has_custom_data_elements.rb
ttoomey: Seems reasonable to me to have them in the same concern, renaming to `FormSubmittable` would be tidy

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -401,6 +426,25 @@ def related_records
...
+      :hiv_aids,
+      :mental_health_disorder,
+      :substance_use_disorder,
+      :exit,
+      :youth_education_status,
+      :employment_education,
+      :current_living_situation,
+      :ce_assessment,
+      :ce_event,
+    ].map { |assoc| send(assoc)&.destroy! }.compact
```
ttoomey: is the return value significant? Looks like it would either all succeed or raise on failure
```suggestion
    ].each { |assoc| send(assoc)&.destroy! }
```

ttoomey (approved):
This is much cleaner, thank you!!

* looks like there are still a few test failures to be addressed
* I think we will need to coordinate with the data team on this. DBT has a dependency on the "hmis_form_processors"."custom_assessment_id" association; we'll need to make corresponding changes for that and do synchronized deployments so as to not break exports

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: No, updated

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: Not sure about that one, I don't see why we should limit ourselves to say that only HUD records can be updated from forms. You might want to have an assessment that generates a Referral for example, or some other kind of custom record.

db/warehouse_structure.sql
eanders: It's fine if you don't. `Bi::ViewMaintainer.create_views` will give them to you.  I'm pretty sure no one is using them any more, and it looks like the migration that created them may have been removed (with the last migration roll-up).

drivers/hmis/app/models/hmis/form/form_processor.rb
eanders: I wasn't proposing that they'd have to be HUD tables, only that we _know_ the HUD ones change on a schedule, so maybe we should move those there.  something like:
```
tables = [
  :referrals,
] + HudUtility2024.enrollment_dependent_hmis_tables
tables.map { |assoc| send(assoc)&.destroy! }.compact
```
maybe?

db/warehouse/migrate/20240531152432_add_form_processor_owner.rb
gigxz: Ok interesting, thank you for the explanation. I will make that change!

gigxz: https://github.com/greenriver/hmis-warehouse/pull/4420/commits/284c92316ea06ee62c602ae15acc2fc0d34b3c31

gigxz:
The test failure should have been fixed by https://github.com/greenriver/hmis-warehouse/pull/4426, waiting for the latest test suite to run.

drivers/hmis/app/models/hmis/hud/concerns/has_custom_data_elements.rb
gigxz: https://github.com/greenriver/hmis-warehouse/pull/4420/commits/850f43db90a4cbd23cbf2c4d70c0031129125939

PR #4422, opened 2024-06-01
5916-stage-1-link-ids

ttoomey:

## Description

[GH issue 5916](https://github.com/open-path/Green-River/issues/5916)

5916 stage 1, 
*   backup form definition and processor columns in case we need to rollback
* update link ids in form processor records
* update link ids in form definition db records 
* update link ids in json form definitions

stage 3: https://github.com/greenriver/hmis-warehouse/pull/4424
## Type of change
- [x] housekeeping


gigxz (approved):

ttoomey:
@gigxz per our convo last week, this should be ready for QA

db/warehouse/migrate/20240531020035_migrate_hmis_form_data.rb
```diff
@@ -0,0 +1,20 @@
+class MigrateHmisFormData < ActiveRecord::Migration[7.0]
+  def up
+    raise unless Hmis::Form::FormProcessor.column_names.include?('backup_values')
+    raise unless Hmis::Form::Definition.column_names.include?('backup_definition')
+
+    # Rake::Task.clear # clear previously loaded tasks
+    # Rails.application.load_tasks # reload tasks from the application
+
+    Rake::Task['driver:hmis:migrate_hmis_link_ids'].invoke
```
gigxz, line 9: I did encounter an error locally on an inability to transform the link id `"0354bf9d-f658-4689-9e55-8fcc2058f252"`. I believe it's because the regex says that it should not start with a number, but `transform_identifier` does not add the q_ prefix in this case

gigxz: Was able to run successfully with this change:

```diff
-    # value is just a number
-    value = value =~ /\A[0-9]+\z/ ? "q_#{value}" : value
+    # value starts with a number
+    value = value =~ /\A[0-9]/ ? "q_#{value}" : value
```

gigxz (commented):
Looks good to me, pending change from https://github.com/greenriver/hmis-warehouse/pull/4422/files#r1633284274. Thank you @ttoomey!

Testing I did locally:
* Check out `drivers/hmis_external_apis/public/schemas/form_definition.json` from `5916-stage-3-link-ids` https://github.com/greenriver/hmis-warehouse/pull/4424
* Validate my local forms `Hmis::Form::Definition.all.map { |fd| Hmis::Form::Definition.validate_schema(fd.definition) }.flatten`
* Modify `env_key` in JsonForms to test all possible `CLIENT` values, re-run seed_definitions  and run schema validator for each one to ensure that no link ids are invalid

db/warehouse/migrate/20240531020035_migrate_hmis_form_data.rb
ttoomey: Great catch, that regex was too strict. Thanks!

gigxz (approved):

gigxz:
retarget to 120 and to qa on it next week (or later this week)

gigxz (approved):
Looks good. I pushed a few commits during QA. I'd say this is ready to merge.

db/warehouse/migrate/20240531020035_migrate_hmis_form_data.rb
```diff
@@ -0,0 +1,20 @@
...
+  end
+
+  def down
+    raise 'refusing destructive rollback in production' if Rails.env.production?
```
gigxz, line 13: @ttoomey maybe we should wrap this `MigrateHmisFormData` migration in a `HmisEnforcement.hmis_enabled?` check so it skips altogether on other environments. What do you think ( cc @eanders ) ?

ttoomey: @gigxz  I guess that should be fine to add an early-return so the migration runs but it's a no-op.

It probably doesn't matter though since it looks like non-hmis installations wouldn't even have form definition records to migrate; it seems like they aren't seeded unless `ENV['ENABLE_HMIS_API'] == 'true'`

gigxz:
@eanders do you want to review before merging?

eanders:
Sorry I missed the request to review before!  I don't have time to review right now, but took a quick look at what's being touched, and it seems safe enough if you're good with the HMIS side of things.

PR #4425, opened 2024-06-01
4818-client-name-permission-data-migration

ttoomey:

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/4818)

Migration to set the can_view_client_name to true on existing roles that might be used to grant access to a client

## Type of change
- [x] Bug fix


db/migrate/20240601143623_client_name_permission_data_migration.rb
```diff
@@ -0,0 +1,24 @@
...
+    view_cond = <<~SQL
+      can_view_clients = true OR
+      can_edit_clients = true OR
+      can_search_clients_with_roi = true OR
+      can_search_window = true OR
+      can_search_own_clients = true OR
+      can_search_all_clients = true
+    SQL
+    Role.where(view_cond).
+      where.not('name ILIKE ?', '%Green River%').
```
ttoomey, line 16: intention is keep this off for GR staff

app/models/role.rb
```diff
@@ -149,6 +149,7 @@ def self.administrative? permission:
     permissions_with_descriptions.merge(health_permissions_with_descriptions)[permission][:administrative] rescue true # rubocop:disable Style/RescueModifier
   end
 
+  # appears to be broken, the :access property is not present in permissions_with_descriptions
   def self.permissions_for_access(access)
     permissions_with_descriptions.select { |_k, attrs| attrs[:access].include?(access) }.keys
   end
```
ttoomey: Noticed this issue while investigating:
This is called by `with_viewable_permissions` but throws an exception, maybe it's rotted away and should be removed?

eanders: Thanks, looks like there are corollary versions in the HMIS side that _are_ used.  Removing this and a couple other unused methods that called it.

eanders (approved):
Thanks @ttoomey !  Tested locally with and without ACLs and at least on the surface it appears to work correctly.

PR #4426, opened 2024-06-02
Fix bug in warehouse_trigger_processing

gigxz:

## Description

Resolves https://github.com/open-path/Green-River/issues/6170

3 main issues here:
* `submit_form_spec` was not properly testing queuing of processing jobs, because they were already queued from when the factory created the jobs
* CustomClientName performed an `update` inside of an after_save hook, which was causing issues with the processing hook
* We were using `after_commit` which can't reliable track `changes` when there are multiple saves happening in a transaction. I switch to `after_save` instead. The downside of that is that it would trigger even if the transaction rolls back, but that is unlikely (we don't save unless its valid) and worst case it just triggers unnecessary processing.

## Type of change
- [x] Bug fix


drivers/hmis/app/models/hmis/hud/custom_client_name.rb
```diff
@@ -52,13 +52,14 @@ def full_name
...
 
   def update_client_name
-    client.update(
-      first_name: first,
-      last_name: last,
-      middle_name: middle,
-      name_suffix: suffix,
-      name_data_quality: name_data_quality || 99,
-    )
+    # Removed calling update from within an after_save callback is dangerous(?) and caused warehouse processing job not to trigger for Client
```
gigxz: discuss- what is a better way to do this

eanders: I think the problem with calling `update` within an `after_save` hook, is that it will cause a `save` event, so can loop out of control.  
What about calling it in `before_save` and just setting the attributes?
Alternatively, you could dirty the record with a new column) and then have a background process that looks for dirty records every few minutes and fixes them up.

gigxz: I like the idea of a separate process that reconciles them, but it doesn't help in cases where the propagation really should happen immediately (like during a mutation).

I considered the `before_save` option, but it doesn't feel right since then the caller would then be expected to call `name.client.save!` after, which is unlikely. The other direction-- before_save on the Client-- doesn't seem quite right either, since you also need to call it in a specific way.

There are only 2 places where this propagation is needed, so I don't think it's _too_ much to just ask the caller to handle it. (Which is what I propose in this pr with `assign_primary_name_fields`).
1) When a Client is created through an application form submission. This always goes through the `ClientProcessor` which handles the name propagation.
2) The Referral creation API endpoint, which creates a CustomClientName for a new Client.
~~3) Client Merge~~ it isn't actually needed here, since we only mark a name as primary if it already matches the hud Client's name fields.

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -120,11 +120,12 @@ def merge_and_find_primary_name
       name_scope.sort_by(&:id).each do |name|
         client_val = [client_to_retain.first_name, client_to_retain.middle_name, client_to_retain.last_name, client_to_retain.name_suffix]
         custom_client_name_val = [name.first, name.middle, name.last, name.suffix]
+        # consider this name "primary" it matches the name on the client_to_retain's Client record
         primary = (client_val == custom_client_name_val) && !primary_found
 
         name.client = client_to_retain
         name.primary = primary ? true : false
-        name.save!(validate: false) # if primary, this save will update the Client name attributes (FirstName, LastName, etc)
+        name.save!(validate: false)
```
gigxz, line 128: This comment was irrelevant. The values would already match if its primary

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
```diff
@@ -18,7 +18,10 @@ class Hmis::Hud::CurrentLivingSituation < Hmis::Hud::Base
   belongs_to :user, **hmis_relation(:UserID, 'User'), optional: true
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
 
-  after_commit :warehouse_trigger_processing
+  # Trigger warehouse processing after create/update if certain fields changed.
+  # This is called even if the transaction that created/updated the record is rolled back, so it could queue some unnecessary processing.
+  # Not using the after_commit hook because it doesn't reliably pick up the saved_changes from a transaction.
+  after_save :warehouse_trigger_processing
```
gigxz, line 24: I think my comment here is accurate, but please let me know if what I'm saying is incorrect.

For a while I was looking at this as a reference https://github.com/rails/rails/issues/51046 but I think it's actually unrelated because it's 7.1, and we are on 7.0.

drivers/hmis/app/models/hmis/hud/service.rb
```diff
@@ -28,7 +28,7 @@ class Hmis::Hud::Service < Hmis::Hud::Base
 
   scope :bed_nights, -> { where(RecordType: 200) }
 
-  after_commit :warehouse_trigger_processing
+  after_save :warehouse_trigger_processing
```
gigxz, line 31: This one was actually passing tests with `after_commit`, unlike CLS, I'm not sure why. Still updated it though for consistency.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
```diff
@@ -146,10 +146,9 @@ def setup_client_name(client, attrs)
       new_ccn.save!
 
       # ensure the ccn is the only primary
-      client.names.where.not(id: new_ccn.id).each do |ccn|
-        # update on each record for lifecycle hooks
-        ccn.update!(primary: false)
-      end
+      client.names.where.not(id: new_ccn.id).update_all(primary: false)
+      # propagate the primary name to the client record
+      client.assign_primary_name_fields
```
eanders, line 151: Where does `client.save` get called? (I'm probably missing something, but `client.assign_primary_name_fields` doesn't seem to be persisting these fields.)

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -132,8 +132,7 @@ def self.input_to_multi_fields(input_field, enum_map, not_collected_key, none_fi
       end
 
       primary = values.find { |v| v['primary'] == true }
-      # Build attributes for Client based on the Primary name. This already happens in the after_save hook on CustomClientName,
-      # but we need it here so that the correct values are present on the unpersisted record for validation.
+      # Build attributes for Client based on the Primary name
       client_attributes = if primary.present?
```
eanders, line 136: I can't seem to comment on it, but 137 - 142 are roughly equivalent to `client.rb#426-430`, I wonder if it's worth DRYing that up a bit.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -416,4 +417,17 @@ def self.reset_source_hashes!(data_source_id)
       )
     end
   end
+
+  # Assign the HUD Client name fields (FirstName, LastName) from the primary CustomClientName record
+  def assign_primary_name_fields
```
eanders, line 423: Is this only called in the create referral job?

drivers/hmis/app/models/hmis/hud/current_living_situation.rb
ttoomey: Looks accurate to me.

It's a bit confusing that the delayed job is called even if the tx is rolled back (took me a minute to realize we back it in a different database). But that's probably TMI for this comment

ttoomey (approved):
Looks good, thanks for fixing!

I noticed `Exit` has the same hook. Maybe it should match for consistency:
```diff
-  after_commit :warehouse_trigger_processing
+  after_save :warehouse_trigger_processing
```

Maybe out of scope but I'd also suggest renaming `warehouse_columns_changed?` to `warehouse_columns_saved?` - I think that's a more accurate description of what it's doing.

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
gigxz: Roughly, yes. They're different though, in this case the "primary name" that we're working with is not actually a CustomClientName object, it's just a hash map of values from the form submission.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
gigxz: Line 94 here https://github.com/greenriver/hmis-warehouse/blob/0850eb5fe207315020c91a04cb0cb9a6fa27cfcd/drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb#L84-L94 and there is test coverage for it. I'll add a comment

gigxz:
https://github.com/greenriver/hmis-warehouse/pull/4426/commits/63f0323044d9b1f0ab3772effe642e479463dd0f feedback addressed:
* use after_save on Exit too
* clarification comment in referral job

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: yes

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
eanders: Thanks

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
eanders: yeah, I noticed that, but the primary name object should respond just fine to `['first']`, etc.  If you want to clean that up....  Maybe just a comment to note that it would need to be changed there if this is ever changed.

eanders (approved):

eanders:
@gigxz can we retarget this to 119?  It's getting close to Thursday and this seems large-ish.

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
gigxz: Ok, I added comments in both places

PR #4428, opened 2024-06-03
Actiontext CVE

eanders:

## Description

Security update for action text

## Type of change
- [ ] Dependency update


Gemfile
```diff
@@ -190,6 +190,7 @@ gem 'net-ssh', '~> 7', require: false
 gem 'net-http'
 gem 'addressable' # normalize uris
 gem 'redis-actionpack'
+gem 'actiontext', '~> 7.0.8.3'
```
ttoomey: Housekeeping suggestion: since we have the exact versions in Gemfile.lock, it's probably not necessary to repeat them in the Gemfile too, right? Or if we do, maybe set the version on the `rails` gem itself.

```suggestion
```

ttoomey (approved):
Thanks for patching!

Gemfile
eanders: My understanding was that adding it here was what forced the update in the lock file.

ttoomey: You can force an update that way. However running `bundle update rails` is the approach I have seen more commonly. Once the Gemfile.lock is updated, the additional dependency in the Gemfile isn't required to use the new version.

I think including dependencies for the side-effect could be confusing to other engineers. To me, a dependency like this communicates that this application has a special dependency on specific versions of actiontext, which isn't the case (I don't think we even use actiontext)

eanders: Thanks.  Updated with `bundle update rails`

PR #4431, opened 2024-06-04
6000 prune csv staging and source tables

ttoomey:

## Description
prune csv staging and source tables

[GH Issue](https://github.com/open-path/Green-River/issues/6000)

Note: 
* we are not handling tables for previous hud specs (2022, 2020)
* deletion is not yet performed, it's a separate job that is commented out of our cron. When ready uncomment the job in `app/jobs/importing/run_daily_imports_job.rb`

## Type of change
- [x] Data clean-up / housekeeping


ttoomey:
@eanders this is work-in-progress but wanted to check if [the expiration logic](https://github.com/greenriver/hmis-warehouse/pull/4431/files#diff-bd1e1eefe312ac5c8d87643ec82faf91fb38f9c74dc75e8efb4e75f4a206c5b7R1) looks like it's on the right track

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
```diff
@@ -0,0 +1,55 @@
...
+        puts "#{model.name} expired: #{candidates_for_expiration.where(expired: true).size}"
+      end
+
+      total
+    end
+
+    protected
+
+    def recent_import_ids
+      @recent_import_ids ||= HmisCsvImporter::Loader::LoaderLog.order(created_at: :desc).limit(@recent_imports_to_keep).pluck(:id)
```
eanders: There's some complexity here we need to account for.  These will need to be per-data source, and it will need to multiply the `@recent_imports_to_keep` by `HmisImportConfig.file_count` if there is a config for that data source.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
```diff
@@ -0,0 +1,55 @@
...
+    end
+
+    def expired?(record, destination)
+      date_deleted = destination&.DateDeleted || record.DateDeleted
+      return true if date_deleted && date_deleted < @expire_deleted_records_before
```
eanders:
```suggestion
      # The record is no longer attached to a destination and not included in the recent loaders (via batch selection)
      return true if destination.blank?
      # The destination record (or this record) has been deleted over 2 weeks ago
      return true if date_deleted && date_deleted < @expire_deleted_records_before
```

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      false
+    end
+
+    def models
+      # TODO: need to handle importer tables too
+      all = [
+        HmisCsvImporter::Loader::Loader.loadable_files,
+        #  #HmisCsvImporter::Importer::Importer.importable_files,
```
eanders: These should exclude the `Export` table.  We should keep those indefinitely (there's only ever one row of metadata.)  And there's no `DateDeleted` so it breaks all sorts of other things.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
```diff
@@ -0,0 +1,55 @@
...
+      @now = now.to_datetime
+      @expire_deleted_records_before = now - keep_deleted_records_for
+      @recent_imports_to_keep = recent_imports_to_keep
+
+      total = 0
+      models.each do |model|
+        # FIXME, should include select clause to improve performance
+        candidates_for_expiration = model.
+          where.not(loader_id: recent_import_ids).
+          preload(:destination_record_with_deleted)
```
eanders: This probably isn't quite right, but it should be close.
```suggestion
          preload(:destination_record_with_deleted).
          select(:id, model.hud_key, data_source_id, :DateDeleted, :loader_id)
```

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
```diff
@@ -0,0 +1,55 @@
+module HmisCsvImporter
+  class MarkExpiredJob < ApplicationJob
+    def perform(now: DateTime.current, recent_imports_to_keep: 10, keep_deleted_records_for: 2.weeks)
```
eanders: I think this should take a `data_source_id` as well.  We can kick off the DJ at the end of an import.  (See below for where that's necessary.)

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/mark_expired_job.rb
eanders: LoaderLogs should have the associated `importer_log_id` which I think you'll need to target the tables like `hmis_2024_enrollments`.   That probably means `recent_import_ids` is actually `recent_loader_ids` and you'll need a second method to pull the importer ids.

eanders (commented):
@ttoomey this looks great.  I _think_ it's 90% there, dropped a bunch of thoughts inline.

eanders:
@ttoomey thinking about this a bit more.  I think we should add some tests.  Specifically, imports to two data sources where there are sufficient imports to one (but no the other) to trigger cleanup.  We should make sure data is removed/retained as expected.  In the test env we could set `recent_imports_to_keep` to 2 or 1.

We'll probably want to make the count variable by data source, I can see an argument for indefinite retention of some data, though I don't have a use-case yet, adding a field on data source, exposed on the edit screen, probably makes the most sense.

We should chat about the expiration date added to records where the warehouse copy is marked deleted.  Why do those need to be treated differently?  Could it just look at the warehouse records, including deleted to determine which records to remove from the imported data?  It could use the same "not in the last x imports" pattern.  I'm probably forgetting something, since when I first read through the code that made sense, but it feels like I'm missing something.  Generally speaking, the only way something would be deleted from the warehouse/destination side, is if it was removed from an import.  

Finally, I'm concerned about an edge case, where generally the data is automated, but sometimes people load data to fix a particular issue.  If they loaded data for only one project at a time, but imported 10-20 files in a row, it would remove all the automated source data regardless.  This makes the problem more tricky, and maybe isn't common enough to worry about, but if you have thoughts on how to easily identify the last 10 imports that _should_ have included the data (as in a combination of data source/project relation) it might be worth exploring.  Alternatively, using 20 instead of 10 as a default would be safer in this scenario.

ttoomey:
Thanks for following up @eanders - I'm working on some changes to address your feedback, I'll get it pushed up this week

> I think we should add some tests. Specifically, imports to two data sources where there are sufficient imports to one (but no the other) to trigger cleanup. We should make sure data is removed/retained as expected. In the test env we could set recent_imports_to_keep to 2 or 1.

Definitely, will add tests

> We'll probably want to make the count variable by data source, I can see an argument for indefinite retention of some data, though I don't have a use-case yet, adding a field on data source, exposed on the edit screen, probably makes the most sense.

Sounds reasonable, should we wait until there's a use-case?

> We should chat about the expiration date added to records where the warehouse copy is marked deleted. Why do those need to be treated differently? Could it just look at the warehouse records, including deleted to determine which records to remove from the imported data? It could use the same "not in the last x imports" pattern. I'm probably forgetting something, since when I first read through the code that made sense, but it feels like I'm missing something. Generally speaking, the only way something would be deleted from the warehouse/destination side, is if it was removed from an import.

Yeah let's chat about it. Maybe we can simplify as you suggest

> Finally, I'm concerned about an edge case, where generally the data is automated, but sometimes people load data to fix a particular issue. If they loaded data for only one project at a time, but imported 10-20 files in a row, it would remove all the automated source data regardless. This makes the problem more tricky, and maybe isn't common enough to worry about, but if you have thoughts on how to easily identify the last 10 imports that should have included the data (as in a combination of data source/project relation) it might be worth exploring. Alternatively, using 20 instead of 10 as a default would be safer in this scenario.

Perhaps we need a time condition in there too (keep records from the last X imports and last N weeks). Or perhaps there's a way to identify such small "partial" imports and ignore those for the purposes of this calculation. I'll give it some thought

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_importers_job.rb
```diff
@@ -0,0 +1,75 @@
...
+          SELECT importer_log_id, "#{key_field}", row_number() OVER (PARTITION BY "#{key_field}" ORDER BY id DESC) AS row_num
+          FROM "#{model.table_name}"
+          WHERE id IN (#{batch.map(&:id).join(',')})
+        ) subquery
+        WHERE subquery.row_num <= #{@retain_log_count}
+      SQL
+
+      results = model.connection.select_rows(sql)
+      results.each_with_object({}) do |ary, h|
+        key, id = ary
```
eanders: I'm probably reading this wrong, but could this be written:
```
results.each_with_object({}) do |(key, id), h|
```
I find that easier to follow than the destructuring

eanders (changes requested):
@ttoomey I think this is an elegant solution!
I was able to follow the logic and I like the approach of looking at the loader and importer tables over the warehouse itself.  However, this misses one case (at least I think it does).  The reason we were starting from the warehouse was because there may be data there that hasn't been imported in a long time (say over a year) but is still valid, and still needs its importer and loader records.  I think the cleanest way to do that is to have another check in the loops that add `valid_ids` that pulls records in the "destination" class for that data source, and generates a set of hud_keys that are all considered valid.

Let me know if that doesn't make sense or if you want to discuss.

ttoomey:
>  The reason we were starting from the warehouse was because there may be data there that hasn't been imported in a long time (say over a year) but is still valid, and still needs its importer and loader records. 

Thanks @eanders - I think the approach here should handle that case already but maybe I'm misunderstanding. I should add a test-case to demonstrate. But let's try and check in on it to confirm

drivers/hmis_csv_importer/spec/jobs/hmis_csv_importer/cleanup/shared.rb
```diff
@@ -0,0 +1,54 @@
...
+  let(:now) { DateTime.current }
+
+  let(:data_source) do
+    GrdaWarehouse::DataSource.create(name: 'Green River', short_name: 'GR', source_type: :sftp)
+  end
+
+  def import_csv_records(run_at:)
+    Timecop.freeze(run_at) do
+      import_hmis_csv_fixture(
+        'drivers/hmis_csv_importer/spec/fixtures/files/twenty_twenty_four/allowed_projects',
```
ttoomey, line 19: Is a good fixture to use? Maybe there's one that's more comprehensive

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -201,6 +201,12 @@ def pre_process_class!(file_name, klass)
       end
     end
 
+    def self.expiring_models
+      importable_files.values.filter do |model|
+        model.column_names.include?('expired')
```
ttoomey: could also write this as excluding the "Export" class, opting to select only models with an 'expired' column

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -0,0 +1,109 @@
...
+module HmisCsvImporter::Cleanup
+  class ExpireBaseJob < ApplicationJob
+    include ReportingConcern
+
+    # @param data_source_id [Integer] the ID of the data source
+    # @param retain_log_count [Integer] the number of retained log records
+    # @param retain_after_date [DateTime] the date after which records are retained
+    def perform(data_source_id: nil, retain_log_count: 10, retain_after_date: DateTime.current)
+      @data_source_id = data_source_id
+      @retain_log_count = retain_log_count * files_per_log
```
eanders: So, I'm afraid I lead us down an unnecessary path.  When we were basing the retention on the existence of data in the warehouse, we needed to know the data source and we needed to be careful about the number of files that were loaded at import time, but now that you've pivoted to looking at the loader and importer records, we can just preserve the last N (and if we want to make sure we have at least X weeks worth that's fine too.)  

We no longer need to care about the data source or number of files per import because a given row will only exist in one of those anyway, so by keeping the last N we'll always have the last N for any given row, which is sufficient, and way less complicated, than what's there now.

Let me know if that doesn't make sense, or if you think we still need this complexity.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: This definitely seems safer.  Thanks!

eanders: Though, maybe we should exclude `Export` as well, just in case HUD ever adds an `expired` field, or we decide to for some reason.

drivers/hmis_csv_importer/app/views/hmis_csv_importer/imported/show.haml
```diff
@@ -13,11 +13,13 @@
           %tr
             - @klass.hmis_structure.each_key do |header|
               %th= header
+          %th Expired
         %tbody
           - @data.each do |row|
             %tr
               - row.hmis_data.each do |key, value|
                 %td= link_to_if key.to_sym == @klass.hud_key, value, source_data_path(search: {type: entry_name, id: value, data_source_id: row.data_source_id })
+              %td= row.expired ? 'Y' : 'N'
```
eanders: We have a helper for this:
```suggestion
              %td= yes_no(row.expired)
```
or if you want to get fancy, we have `checkmark_or_x`

drivers/hmis_csv_importer/spec/jobs/hmis_csv_importer/cleanup/shared.rb
eanders: yep, that should be fine.  There are probably others with more data, but as long as we have a few projects/organizations, it should be fine.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
ttoomey: No worries, this doesn't seem like much complexity. I'll remove the files_per_log multiplier. 

The data source can be removed from the deletion job but I think keeping it for the expiration jobs makes sense. We need it to partition the records by hud_key anyway (key is not unique over data sets). So might as well pass it in explicitly to the job

ttoomey:
maybe I should target r121 with this?

eanders:
> maybe I should target r121 with this?

Yes please, but that feels very plausible.

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -168,6 +168,14 @@ def perform
         create_statistical_matches
         generate_logging_info
 
+        GrdaWarehouse::DataSource.importable.with_deleted.each do |data_source|
```
eanders: I _think_ we want:
```suggestion
        GrdaWarehouse::DataSource.with_deleted.each do |data_source|
```
We now have the situation where a previously importable data source has been converted to an HMIS data source, and we should probably clean those up as well.  There are some data sources (health for instance) that will never have had imports, but that should be a really quick run.  
Alternatively, we could call it once for the non-importable data sources since they shouldn't be changing.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -0,0 +1,105 @@
...
+#   end
+#
+#   def log_model
+#   end
+#
+#   def models
+#   end
+#
+module HmisCsvImporter::Cleanup
+  class ExpireBaseJob < ApplicationJob
```
eanders: Thanks for DRYing this up.  I think there's a solid argument for running it once against `HmisCsvTwentyTwentyTwo.importable_files_map` and `HmisCsvTwentyTwenty.importable_files_map` to enforce the same retention policy against those tables.  Perhaps in the `TaskQueue`.  Definitely doesn't need to hold up this PR.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -201,6 +201,12 @@ def pre_process_class!(file_name, klass)
       end
     end
 
+    def self.expiring_models
+      importable_files.values.filter do |model|
+        model.column_names.include?('expired') && model.name.demodulize != 'Export'
```
eanders: One more thought here.  I _think_ we should also keep Project.  The definition of the export is essentially "the details in Export.csv + the projects in Project.csv."  So keeping those records which get repeated in pretty much every import, would allow us to know if some data should or should not have been included in the import (within the last 5 of any given record.)
```suggestion
        # Don't expire or remove Export or Project records since they define the scope of a given import
        model.column_names.include?('expired') && ! model.name.demodulize.in?(['Export', 'Project'])
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/loader/loader.rb
```diff
@@ -100,6 +100,12 @@ def importer_log
       @importer&.importer_log
     end
 
+    def self.expiring_models
+      loadable_files.values.filter do |model|
+        model.column_names.include?('expired') && model.name.demodulize != 'Export'
```
ttoomey, line 105:
```suggestion
        model.column_names.include?('expired') && ! model.name.demodulize.in?(['Export', 'Project'])
```

eanders (approved):
I think this is good now thanks @ttoomey !

ttoomey:
> I think this is good now thanks @ttoomey !

Thanks! I'll leave this unmerged for now assuming we might want to selectively deploy to staging for testing

db/warehouse/migrate/20240628180603_index_import_tables.rb
```diff
@@ -0,0 +1,30 @@
+class IndexImportTables < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
```
ttoomey: @eanders we aren't using async (concurrent indexes). Probably want this in a ddl tx?

eanders: So, I've had terrible deployment experiences with that recently.  (And couldn't get it working correctly locally either.) I'm not sure if it's our deployment infrastructure or what, but It tends to throw errors that it can't get a lock unless we add indexes synchronously.  Maybe we need to mess with the lock timeouts?

ttoomey: Oh yeah, I just meant that there's no reason to run this outside of a ddl tx right? So we can remove this line
```suggestion
```

eanders: Yes! that was probably left over from trying to follow the strong migrations guidance.

db/warehouse/migrate/20240628180603_index_import_tables.rb
```diff
@@ -0,0 +1,29 @@
...
+      execute "SET statement_timeout TO '0s'"
+
+      tables.each do |table|
+        add_index table, :data_source_id, if_not_exists: true
+      end
+
+      execute "SET statement_timeout TO #{quote(old_value)}"
+    end
+  end
+end
```
eanders: @ttoomey with the changes coming in https://github.com/greenriver/hmis-warehouse/pull/4502, maybe we don't need this?  I think we're no longer leaning on Data Source ID anywhere.  Do you agree?

ttoomey: Good catch, thanks! We do group on data_source_id in the partition but the index doesn't seem to impact performance. I'll remove this migration

ttoomey:
@eanders- okay to merge this?

eanders:
@ttoomey I'll deploy to QA first just to make sure we didn't break something.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -0,0 +1,138 @@
...
+    private def relevant_id_query(model)
+      key_field = model.hud_key
+      <<~SQL
+        SELECT id, #{log_id_field} FROM (
+          SELECT id, #{log_id_field}, row_number() OVER (
+            PARTITION BY "#{key_field}", data_source_id ORDER BY id DESC
+          ) AS row_num
+          FROM #{model.quoted_table_name}
+        ) subquery
+        WHERE subquery.row_num > #{@retain_item_count}
```
eanders: @ttoomey what do you think about adding an `AND subquery.expired != true` here, and the associated select of the `expired` column.  That would allow this to be resumable and would cut down significantly on writes, which is one of the issues we're seeing with the burst balance in RDS.  Sadly, the update being in a transaction means those writes aren't recorded until the transaction commits.  Is there any way we can take that out of the transaction?  Then it really would be resumable, and I don't think there's any reason it would be a problem if we marked things expired in some incomplete way.

eanders: OH!  Or maybe since we have the temp table (if we add back in the ID column) we could easily batch the update with some simple pagination.

ttoomey: Great, thanks for adding that!

PR #4433, opened 2024-06-04
Remove CAS section from client search in VA

dtgreiner:

## Description

## Type of change


drivers/client_access_control/app/views/client_access_control/clients/_client_card.haml
```diff
@@ -6,7 +6,7 @@
 -# Contain content blocks that are later yielded in
 -# the "client_card_content-secondary__#{id}" content block
 -# within `clients/client_card_wrapper` partial
-= render 'clients/search/cas', client: client
+= render 'clients/search/cas', client: client unless GrdaWarehouse::Config.get(:client_dashboard).to_s == 'va'
```
eanders: Hmm, this is probably fine, but I wonder if we should create a client card for each dashboard?  What do you think?  Essentially this could be:
```
render "clients/search_cards/#{GrdaWarehouse::Config.get(:client_dashboard)}, client: client"
```
and then encapsulate the existing card as `_default.haml` and `_va.haml` would hold only those bits they want.

dtgreiner: I was looking at doing something similar off the bat, but wasn't sure how often these cards would change. If its very often, that may end up with a change being missed on one or the other cards. With that said, I definitely like the idea of having separate cards per dashboard. It feels like it would be a bit cleaner in the long run. 

I can add it to my list for tomorrow morning to switch this PR over to use the `clients/search_cards/#{GrdaWarehouse::Config.get(:client_dashboard)}` method

eanders: Thanks.  Yeah, they don't change often, but I like the idea that we have an easy knob for each client dashboard implementation.

dtgreiner:
@eanders I've pushed a change that will split out the search cards by the dashboard types. Thanks!

eanders (approved):

PR #4447, opened 2024-06-11
Update  system pathways report equity section

awisema:

## Description

Update the System Pathways Report equity section to be based on the Race and Equity categorizations introduced in the 2024 PIT counts.

## Type of change


drivers/system_pathways/app/models/system_pathways/chart_base.rb
```diff
@@ -141,13 +144,84 @@ def average(value, count)
...
+    private def filter_for_race_with_details_filter(scope, race_filter)
+      return scope.where(id: multi_racial_clients(scope).select(:id)) if race_filter.races.include?('MultiRacial')
+
+      query = races.except('MultiRacial', 'RaceNone').keys.map { |k| [k.underscore.to_sym, false] }.to_h
+      race_filter.races.each do |column|
+        next if column.in?(['MultiRacial', 'RaceNone'])
+
+        query[column.underscore.to_sym] = true
+      end
+      scope.where(query)
```
eanders, line 159: Does this need:
```suggestion
      scope.where(**query)
```
Also, what happens if the race column is nil?  I think this will explicitly select only those that are false or true.

drivers/system_pathways/app/models/system_pathways/chart_base.rb
```diff
@@ -141,13 +144,84 @@ def average(value, count)
...
+    private def filter_for_ethnicity_with_details_filter(scope, filter)
+      unknown_query = races.except('MultiRacial', 'RaceNone').keys.map { |k| [k.underscore.to_sym, false] }.to_h
+
+      # These are mutually exclusive, so, although the filter is an array, ignore multiples
+      case filter.ethnicities.first
+      when :hispanic_latinaeo
+        scope.where(hispanic_latinaeo: true)
+      when :non_hispanic_latinaeo
+        scope.where(hispanic_latinaeo: false).
+          where.not(unknown_query).except(:hispanic_latinaeo) # Must have at least one race response
```
eanders: This doesn't look correct, should the `except` go on the `unknow_query`?  It looks like active record has an `except` [method](https://apidock.com/rails/v7.0.0/ActiveRecord/SpawnMethods/except) but I don't think it's doing what you intend.
Also, shouldn't unknown be something like `RaceNone in (8, 9, 99)`?

drivers/system_pathways/app/models/system_pathways/chart_base.rb
```diff
@@ -141,13 +144,84 @@ def average(value, count)
...
+      when :unknown
+        scope.where(unknown_query)
+      end
+    end
+
+    private def filter_for_race_and_ethnicity_with_details_filter(scope, filter)
+      base_query = races.except('MultiRacial', 'RaceNone').keys.map { |k| [k.underscore.to_sym, false] }.to_h
+      combination = filter.race_ethnicity_combinations.first
```
eanders, line 201: Why `.first` here? Can you only choose one combination at a time?

drivers/system_pathways/app/models/system_pathways/chart_base.rb
```diff
@@ -141,13 +144,84 @@ def average(value, count)
...
+
+      query = case combination
+      when :am_ind_ak_native
+        base_query.merge(am_ind_ak_native: true)
+      when :am_ind_ak_native_hispanic_latinaeo
+        base_query.merge(am_ind_ak_native: true, hispanic_latinaeo: true)
```
eanders: Would there be any utility/cleanup/DRY/ease of reading if this were a hash?  Something like:
```
combinations = {
  am_ind_ak_native: {am_ind_ak_native: true},
  am_ind_ak_native_hispanic_latinaeo: {am_ind_ak_native: true, hispanic_latinaeo: true),
  asian: {asian: true},
  asian_hispanic_latinaeo: {asian: true, hispanic_latinaeo: true},
...
}
query = base_query.mege(combinations[combination])
```
Multi-racial would need some special casing, but otherwise i think it would work.  
OR
We could even build the hash from an array of the fields with a  simple map:
```
combinations = {} 
[:am_ind_ak_native, :asian].map do |race|
  combinations[race] = base_query.merge(race: true)
combinations["#{race}_#{hispanic_latinaeo}".to_sym] = base_query.merge(race: true, hispanic_latinaeo: true)
end
combinations[:multi_racial] = { hispanic_latinaeo: false }
...
```

drivers/system_pathways/app/models/system_pathways/chart_base.rb
awisema: It doesn't seem to matter as the signature of `where` is `*args`:
```ruby
def foo(*args)
  puts args.inspect
end

foo(a: 1, b: 2)
[{:a=>1, :b=>2}]

foo({a: 1, b: 2})
[{:a=>1, :b=>2}]

```
But, if it seems more readable, I'm happy to change it.

As to the race columns, they are [true, false] by construction:
```ruby
am_ind_ak_native: client.am_ind_ak_native == 1,
asian: client.asian == 1,
...
```
But, the DB does mark them as `nullable`, so we could clean it up if desired. I believe there are other places where there are non-null assumptions that we should look at if this should be changed.

drivers/system_pathways/app/models/system_pathways/chart_base.rb
awisema: Yes, that is a bug, thanks for catching it! As to `RaceNone` that doesn't get carried over into `SystemPathways::Client`, there are just the booleans as calculated in `SystemPathways::Report#create_universe`

drivers/system_pathways/app/models/system_pathways/chart_base.rb
awisema: I don't think so as I think the `case` is very explicit, and I think that would move the logic away from its use, but, I could easily be convinced as I'm almost always in favor of declarative code!

drivers/system_pathways/app/models/system_pathways/chart_base.rb
awisema: I added an explanatory comment.

drivers/system_pathways/app/models/system_pathways/equity/race_and_ethnicity.rb
```diff
@@ -0,0 +1,117 @@
...
+      am_ind_ak_native_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'am_ind_ak_native') },
+      asian: ->(race_data) { single_race(race_data, 'asian') },
+      asian_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'asian') },
+      black_af_american: ->(race_data) { single_race(race_data, 'black_af_american') },
+      black_af_american_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'black_af_american') },
+      hispanic_latinaeo: ->(race_data) do
+        race_data.select { |client| client['hispanic_latinaeo'] }.
+          map { |client| client.except('id').values }.
+          count { |m| m.count(true) == 1 }
+      end,
```
eanders: I think this needs a comment about what's going on.  I _think_ the `client` isn't actually a client, but some sort of object that has rows with each race key and 0 or 1?

drivers/system_pathways/app/models/system_pathways/equity/race_and_ethnicity.rb
```diff
@@ -0,0 +1,117 @@
...
+        race_data.reject { |client| client['hispanic_latinaeo'] }.
+          map { |client| client.except('id', 'hispanic_latinaeo').values }.
+          count { |m| m.count(true) > 1 }
+      end,
+
+      multi_racial_hispanic_latinaeo: ->(race_data) do
+        race_data.select { |client| client['hispanic_latinaeo'] }.
+          map { |client| client.except('id', 'hispanic_latinaeo').values }.
+          count { |m| m.count(true) > 1 }
+      end,
```
eanders: These could probably use a comment as well, or maybe just move the `hispanic_latinaeo:` down here, and add one comment for all 3.

drivers/system_pathways/app/models/system_pathways/time_chart/ethnicity.rb
```diff
@@ -0,0 +1,156 @@
...
+
+      ph_counts = ph_projects.map do |p_type, p_label|
+        counts = {}
+        ethnicities.each_key do |k|
+          counts[k] ||= 0
+        end
+        # Get rid of the distinct from node_clients
+        stay_length_col = sp_e_t[:days_to_move_in]
+        scope = SystemPathways::Enrollment.joins(:client).where(id: node_clients(p_type).select(:id)).where(stay_length_col.not_eq(nil))
+        race_data = pluck_to_hash(race_columns.except('race_none', 'multi_racial').map { |k, v| [sp_c_t[k], v] }.to_h.merge(stay_length_col => 'Days to Move-In'), scope)
```
eanders: maybe someday we should add a method `known_individual_race_column` so we don't have to keep except-ing `race_none` and `multi_racial`, but probably ok for now.

drivers/system_pathways/app/models/system_pathways/time_chart/race_and_ethnicity.rb
```diff
@@ -0,0 +1,193 @@
...
+
+  private def lookups
+    {
+      am_ind_ak_native: ->(race_data, stay_length_col) { single_race(race_data, 'am_ind_ak_native', stay_length_col) },
+      am_ind_ak_native_hispanic_latinaeo: ->(race_data, stay_length_col) { single_race_latinaeo(race_data, 'am_ind_ak_native', stay_length_col) },
+      asian: ->(race_data, stay_length_col) { single_race(race_data, 'asian', stay_length_col) },
+      asian_hispanic_latinaeo: ->(race_data, stay_length_col) { single_race_latinaeo(race_data, 'asian', stay_length_col) },
+      black_af_american: ->(race_data, stay_length_col) { single_race(race_data, 'black_af_american', stay_length_col) },
+      black_af_american_hispanic_latinaeo: ->(race_data, stay_length_col) { single_race_latinaeo(race_data, 'black_af_american', stay_length_col) },
+      hispanic_latinaeo: ->(race_data, stay_length_col) do
```
eanders: Same here, let's group this with `multi_racial` and add a comment about what's going on in the select.

drivers/system_pathways/app/models/system_pathways/equity/disabling_condition.rb
```diff
@@ -12,10 +12,10 @@ module SystemPathways::Equity::DisablingCondition
       chart: 'disabling_condition',
       config: {
         size: {
-          height: 800,
+          height: node_names.count * 30,
```
dtgreiner, line 15: Do we need a default height for this (and other similar height calculations) in case it ever equates to zero?

drivers/system_pathways/app/models/system_pathways/equity/race_and_ethnicity.rb
```diff
@@ -0,0 +1,120 @@
...
+      black_af_american: ->(race_data) { single_race(race_data, 'black_af_american') },
+      black_af_american_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'black_af_american') },
+      mid_east_n_african: ->(race_data) { single_race(race_data, 'mid_east_n_african') },
+      mid_east_n_african_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'mid_east_n_african') },
+      native_hi_pacific: ->(race_data) { single_race(race_data, 'native_hi_pacific') },
+      native_hi_pacific_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'native_hi_pacific') },
+      white: ->(race_data) { single_race(race_data, 'white') },
+      white_hispanic_latinaeo: ->(race_data) { single_race_latinaeo(race_data, 'white') },
+
+      # For interdependent race information, counting is u sed to determinethe number of true values
```
dtgreiner:
```suggestion
      # For interdependent race information, counting is used to determine the number of true values
```

dtgreiner (approved):

drivers/system_pathways/app/models/system_pathways/equity/disabling_condition.rb
eanders: Good question.  It default to 100 in the calling file.  I guess if there are no nodes (which can't happen) it might end up not showing up, but the node set is fixed, so I think that's ok.

PR #4449, opened 2024-06-11
Support draft to publish workflow for forms

martha:

## Description

GH issue https://github.com/open-path/Green-River/issues/5858

This PR
- adds a new mutation `CreateDraftForm` that creates a new draft version of an existing published form. Let me know what you think about implementing this as its own mutation. The validations here could be moved into a special case of `CreateFormDefinition` if you think that is neater.
- adds a new mutation `PublishForm` that publishes an existing draft. Again, this could have been a special case of `UpdateFormDefinition` but I thought it was a bit neater in its own mutation.
- creates new forms in draft mode
- disables updating non-draft forms
- updates the schema
- adds api tests

The diff is not showing up quite correctly, because I believe release-120 and release-119 aren't in sync, but to see the actual diff for this PR you can check the [commit](https://github.com/greenriver/hmis-warehouse/commit/24d45966de04fe69eb78f1e80290ec2da300324c).

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/mutations/publish_form.rb
```diff
@@ -0,0 +1,38 @@
...
+        identifier: definition.identifier,
+        status: Hmis::Form::Definition::PUBLISHED,
+      )
+
+      previous.update(status: 'retired')
+
+      definition.status = 'published'
+
+      if definition.valid?
+        definition.save!
```
gigxz: Looks good! Just a couple things on this mutation:
* The `update` and `save` should happen in a transaction so that they happen atomically. It would be bad if we retired the old form, but werent able to publish the new one.
* We should also call `update!` instead of `update`, so that it raises an error if the update failed.
* Lets use the constants for published and retired instead of the strings

drivers/hmis/app/graphql/mutations/publish_form.rb
```diff
@@ -0,0 +1,38 @@
...
+    field :form_definition, Types::Forms::FormDefinition, null: true
+
+    def resolve(id:)
+      access_denied! unless current_user.can_manage_forms?
+
+      definition = Hmis::Form::Definition.find_by(id: id)
+      raise 'not found' unless definition
+      raise 'only draft forms can be published' unless definition.status == Hmis::Form::Definition::DRAFT
+
+      previous = Hmis::Form::Definition.where(
```
gigxz: There can only be one published record for a given identifier, so this can be a `find_by`. That will make it so `previous` is a record instead of a scope. It might be clearer to rename to `previous_published_form` to clarify that.

drivers/hmis/app/graphql/mutations/publish_form.rb
```diff
@@ -0,0 +1,38 @@
...
+        identifier: definition.identifier,
+        status: Hmis::Form::Definition::PUBLISHED,
+      )
+
+      previous.update(status: 'retired')
```
gigxz: If this is the first time this form identifer is published, then there wouldn't be a previous record. So we should use a safe accessor here: `previous&.update!(status: 'retired')`

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -16,6 +16,7 @@ def resolve(id:, input:)
 
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition
+      raise 'only allowed to modify draft forms' unless definition.status == Hmis::Form::Definition::DRAFT
```
gigxz: might be nice to have a helper like `definition.draft?` since we check this in both mutations

drivers/hmis/app/graphql/mutations/create_draft_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      last = Hmis::Form::Definition.find(definitions.max_by(&:version).id)
+
+      # Duplicate the last version, incrementing the version number and setting the status to draft
+      definition = last.dup
+      definition.version = last.version + 1
+      definition.status = Hmis::Form::Definition::DRAFT
+
+      if definition.valid?
+        definition.save!
+        { form_identifier: Hmis::Form::Definition.non_static.latest_versions.where(identifier: identifier).first }
```
gigxz: It feels a little strange to return the form identifier rather than the form definition that was created. I assume that's because we want to update the Form Identifier in the cache to have the `draftVersion` field populated?

drivers/hmis/app/graphql/mutations/create_draft_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      else
+        errors.add_ar_errors(definition.errors)
+        { errors: errors }
+      end
```
gigxz: I don't think we want to return validation errors here. `definition.valid?` will just validate the status and identifier uniqueness, it does not validate the form structure itself. When we return `errors: [...]` they are meant to be validation errors that are correctible by the user. In this example, there are no errors that are fixable by the user. 

If there is an "unfixable" error, we prefer to raise that as an exception so that it gets sent to Sentry. `save!` will raise an exception if validation fails.

(We are not consistent about this, we often do what you propose here, so I see why you added it).


So my proposal would be to simplify to:

```suggestion
      definition.save!
      
      # return the form identifier, which will resolve the newly created draft as the draft_version
      { form_identifier: Hmis::Form::Definition.latest_versions.where(identifier: identifier).first }
```

Let me know if that makes sense / if you have questions

drivers/hmis/app/graphql/mutations/publish_form.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      definition.status = 'published'
+
+      if definition.valid?
+        definition.save!
+        { form_definition: definition }
+      else
+        errors.add_ar_errors(definition.errors)
+        { errors: errors }
```
gigxz: Same comment here with regards to raising an exception rather than returning validation errors.

We could do a secondary validation step here (eg call `validate_schema`/`validate_json`) and return a nice message if the form is invalid, but the draft _shouldn't_ have been able to save if it was invalid.

drivers/hmis/app/graphql/mutations/create_draft_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      raise 'not allowed' unless current_user.can_manage_forms?
+
+      definitions = Hmis::Form::Definition.
+        exclude_definition_from_select.
+        where(identifier: identifier)
+
+      raise 'not found' if definitions.empty?
+
+      existing_drafts = definitions.where(status: Hmis::Form::Definition::DRAFT)
+      raise 'not allowed to create draft if one already exists' unless existing_drafts.empty?
```
gigxz: This seems fine, another approach would be to just return the draft that is already present. The scenario would occur if a different user created a draft while you had the page open, so the page lets you hit "new draft" even though the draft already exists in the db. We haven't really tackled the question of multiple editors yet- I created https://github.com/open-path/Green-River/issues/6214 so we could discuss it separately.

drivers/hmis/app/graphql/types/hmis_schema/mutation_type.rb
```diff
@@ -64,6 +64,8 @@ class HmisSchema::MutationType < Types::BaseObject
     field :create_form_definition, mutation: Mutations::CreateFormDefinition
     field :update_form_definition, mutation: Mutations::UpdateFormDefinition
     field :delete_form_definition, mutation: Mutations::DeleteFormDefinition
+    field :publish_form, mutation: Mutations::PublishForm
+    field :create_draft_form, mutation: Mutations::CreateDraftForm
```
gigxz: I don't see a meaningful difference between when we use the word `form_definition` vs `form`. Maybe we can use `PublishFormDefinition` insetad of `PublishForm` for the new one.

It's also unclear from the names whats different between `CreateFormDefinition` (brand new form definition with no content) and `CreateDraftForm` (dup an existing form into a draft).

Maybe CreateDraftForm can be called `CreateNextDraftFormDefinition` or something?

drivers/hmis/app/graphql/mutations/create_draft_form.rb
martha: That was the initial reason, but I couldn't quite figure out how to get the cache to auto update -- maybe because the query to update and the initial query were happening in different components? (see around [here](https://github.com/greenriver/hmis-frontend/pull/804/files#r1635507884) in the frontend PR). As the code exists now, it's not auto-updating, so I think it's neater to just return the `FormDefinition`, and can make that update.

gigxz: Hmm, the cache should update as long as you are resolving all the fields and the cache keys. So it looks like it resolves the `FormDefinitionMetadata` fragment, so all those fields should update. It doesn't resolve the form definition JSON on the draft, so it makes sense if _that_ isn't loaded automatically. Maybe the mutation response wants to look something like:
```
  createDraftForm(identifier: $identifier) {
    formIdentifier {
      id
      identifier
      draftVersion {
          id
          cacheKey
          ... FormDefinitionFieldsForEditor // or whatever is needed for the Form Builder - I forget if its this fragment
       }
    }
  }
```
That way, as soon as the mutation resolves, you already have everything you need to render the Form Builder for the "draft version" that was created. Does that make sense?


It would also be fine to update this to resolve the draft definition alone, but then we'd want to manually update the cache so that `draftVersion` points to it, I think.

martha: Aha, thanks for this, yes that was exactly the problem with the auto-updates -- not all the fields were being returned.

I updated the publish mutation as well so that it returns 3 things -- the previously published (now retired) version, the newly published version, and the form identifier -- and can see in the apollo inspector that this is working correctly. If you prefer, I'd also be willing to update that query to return just 1 thing and then use `cache.modify` in the callback to update the other 2.

drivers/hmis/app/graphql/mutations/create_draft_form.rb
martha: Yes! I was sorta thinking something along these lines already (from my question [here](https://github.com/greenriver/hmis-frontend/pull/804/files#r1635505852) on the frontend PR) -- so it totally makes sense that the fix is actually here upstream.

drivers/hmis/app/graphql/mutations/create_draft_form.rb
gigxz: Ok great, glad that worked!

I see, so for **publish**, the frontend is now resolving:
```
  publishFormDefinition(id: $id) {
    newlyPublished {
      ...FormDefinitionFieldsForEditor
    }
    newlyRetired {
      ...FormDefinitionMetadata
    }
    formIdentifier {
      ...FormIdentifierDetails
    }
  }
```

I'm not exactly following the need for the separate fields, or why `cache.modify` would be necessary. I can try to repro later to see whats happening. But--without having fiddled with it-- a couple thoughts:

1) Why do we need to update the cache for the "newlyRetired" form? From my understanding, it is no longer needed to render the page.  What happens if we just remove it from the cache with `cache.evict`? (Aside: The Version History table, once it exists, should probably be forced to re-fetch once the publish action completes. Then you will see the "newly retired" form at the top of that table).

2) Why does `newlyPublished` need to be returned separately, as opposed to resolving it on the formIdentifer? Like: `formIdentifier {  publishedVersion { ...fieldsWeNeed } } `

martha: Aha, sensible. I pushed up changes accordingly. (2) is definitely better. The only challenge with (1) is that at this point in the publish workflow I think we don't have a reference/id for the just-retired form version. Maybe the best solution is that the publish query should resolve `allVersions` on the form identifier? (You are right, the retired version is not needed to render the page, I was just worried about having stale cache data where we have 2 published versions.)

martha: I just gave this idea a shot and it seems to work okay. https://github.com/greenriver/hmis-frontend/pull/804/files#diff-e38721dd872401d491c291485f2e562aac599fcdf386e9284d0e424b38d1c662R49-R54

We are resolving more data than we really need, especially if the list of previously retired versions is very long, but I think that's okay. Particularly if hitting Publish from the Preview page will take you back to the View Form Details screen -- which it doesn't right now, but that seems like a sensible workflow to me -- then we need this info anyway to render the version history table there.

An alternative would be to just return the `id` and `status` of the versions from `allVersions` and call `cache.evict` on all the ones with status=retired. Then the Version History Table would be responsible for re-fetching, as you described.

gigxz: Okay, awesome. As long as it's not causing issues in the page, I don't think we _need_ to worry about it. Once we implement the Version History table, it will re-fetch it. I added a note to https://github.com/open-path/Green-River/issues/5982 about that.

gigxz: Oops I commented before seeing your latest

gigxz (approved):
Looks good! Thanks for the test coverage! 🎁

PR #4451, opened 2024-06-13
Additional fields for form version table

gigxz:

## Description

* Resolve timestamps and user on form definition
* Sort versions by most recently updated
* Use PaperTrail `save_with_version` to persist a version when the draft is updated, despite the fact that we `ignore` the `definition` field. At some point we made the decision to ignore this field because it is large and we don't really need to persist the version history of changes to `draft`s.

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/809

## Type of change

### DIFF
--- a/drivers/hmis/app/graphql/mutations/update_form_definition.rb
+++ b/drivers/hmis/app/graphql/mutations/update_form_definition.rb
@@ -35,15 +35,16 @@ module Mutations
         builder.validate_definition(definition.definition) { |err| errors.add(:definition, message: err) }
       end
 
+      # Return user-facing validation errors for the form content
       return { errors: errors } if errors.present?
 
-      if definition.valid?
-        definition.save!
-        { form_definition: definition }
-      else
-        errors.add_ar_errors(definition.errors)
-        { errors: errors }
-      end
+      # Raise if the definition is not valid (invalid role/status/identifier; not expected)
+      raise "Definition invalid: #{definition.errors.full_messages}" unless definition.valid?
+
+      # Manually save a PaperTrail version, so we know who made the change. Because we `skip` the definition
+      # field, a PaperTrail record will not automatically get created if only the definition changed.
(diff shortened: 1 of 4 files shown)


drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -35,15 +35,16 @@ def resolve(id:, input:)
...
-      else
-        errors.add_ar_errors(definition.errors)
-        { errors: errors }
-      end
+      # Raise if the definition is not valid (invalid role/status/identifier; not expected)
+      raise "Definition invalid: #{definition.errors.full_messages}" unless definition.valid?
+
+      # Manually save a PaperTrail version, so we know who made the change. Because we `skip` the definition
+      # field, a PaperTrail record will not automatically get created if only the definition changed.
+      definition.paper_trail.save_with_version
```
gigxz, line 46: https://www.rubydoc.info/gems/paper_trail/11.0.0/PaperTrail%2FRecordTrail:save_with_version

needed because of https://github.com/greenriver/hmis-warehouse/blob/72dcf983ff7338a4ef85814419e446da8ff71244/drivers/hmis/app/models/hmis/form/definition.rb#L35

`save_with_version` seems to work OK, but we don't use this method anywhere else. Is there a better way to do this?

We _could_ stop `skip`ing the definition field, but I do still think it would be unnecessary to cram all those definition jsons into the version table.

gigxz:
Screenshot 2024-06-13 at 3 11 31 PM

drivers/hmis/app/graphql/mutations/update_form_definition.rb
ttoomey: using `save_with_version` seems reasonable to me, it's part of paper_trail's documented api

ttoomey (approved):
Looks good, thanks!

PR #4455, opened 2024-06-13
Access Control Importer

eanders:

## Description

* UI and import tooling to import Excel files with users, collections, and roles.
* Background processing of uploaded Excel and import
* Download of sample file
* Fixes for user edits when the user is using Access Controls (maintain system user groups through saves)
* Fixes to the invitations controller for less-used options
* Default invited users to using Access Controls if anyone else already is using access controls
* Feature to allow using titles on permissions that aren't just the humanized version of the column
* Limit a few drop-downs to only users using Access Controls or only non-system items
* Rubocop cleanups to routes file

## Type of change


app/views/admin/groups/download.xlsx.axlsx
```diff
@@ -2,7 +2,7 @@ wb = xlsx_package.workbook
 wb.add_worksheet(name: 'Users') do |sheet|
   sheet.add_row(['First Name', 'Last Name', 'Email', 'Agency', 'Role', 'Account Created', 'Last Login', 'Active'])
   User.joins(:agency).preload(:legacy_roles, :agency).find_each do |user|
-    user.legacy_roles.uniq.each do |role|
+    user.legacy_roles.editable.uniq.each do |role|
```
ttoomey: probably don't need the to preload legacy_roles if we querying them again anyway with `editable`

app/models/concerns/user_concern.rb
```diff
@@ -144,11 +152,11 @@ def self.available_permission_contexts
     end
 
     def self.anyone_using_acls?
-      active.not_system.where(permission_context: 'acls').exists?
+      active.not_system.using_acls.exists?
     end
 
     def self.all_using_acls?
-      ! active.not_system.where(permission_context: [nil, 'role_based']).exists?
+      ! active.not_system.using_role_based.exists?
```
ttoomey, line 159: small style note, in my experience it's uncommon in ruby to use a space after negation "!" - should we try and keep this consistent?
```suggestion
      !active.not_system.using_role_based.exists?
```

app/views/admin/access_controls/_filter.haml
```diff
@@ -10,7 +10,7 @@
     = simple_form_for :filter, method: :get do |f|
       .o-filtration__option.row
         .col-12
-          = f.input :user_id, as: :select_two, collection: User.active.not_system.order(:first_name, :last_name), selected: params.dig(:filter, :user_id).to_i, required: false, include_blank: 'Any user'
+          = f.input :user_id, as: :select_two, collection: User.active.using_acls.not_system.order(:first_name, :last_name), selected: params.dig(:filter, :user_id).to_i, required: false, include_blank: 'Any user'
```
ttoomey: Ensure consistent order. Maybe this should be last, first?
```suggestion
          = f.input :user_id, as: :select_two, collection: User.active.using_acls.not_system.order(:first_name, :last_name, :id), selected: params.dig(:filter, :user_id).to_i, required: false, include_blank: 'Any user'
```

app/views/admin/acl_imports/_complete_import.haml
```diff
@@ -0,0 +1,5 @@
+- unless @import.imported?
+  .d-flex.well
+    %p Proposed changes based on this file are outlined below.  If everything looks good, you can import them with the Complete Import button.
+    = link_to admin_acl_import_path, method: :patch, class: 'btn btn-primary ml-auto' do
+      Complete Import
```
ttoomey: this is fine although it seems like button_to would be better since this is an action, not navigation. Maybe also add a confirmation prompt? ("Are you sure.. this cannot be undone, etc")

app/views/admin/acl_imports/_roles.haml
```diff
@@ -0,0 +1,27 @@
...
+        - @import.roles.each do
+          %th Existing
+          %th
+          %th.border-right Incoming
+    %tbody
+      - Role.permissions(exclude_health: true).each do |column|
+        %tr
+          %th.border-right= Role.title_for_column(column)
+          - @import.roles.each do |role|
+            - perm = role[:permissions].try(:[], column)
```
ttoomey: I think `dig()` is equivalent and maybe easier to read
```suggestion
            - perm = role.dig(:permissions, column)
```

app/views/admin/acl_imports/_collections.haml
```diff
@@ -0,0 +1,29 @@
...
+        %th CoCs
+        %th Cohorts
+        %th Reports
+        %th Project Groups (for project groups)
+        %th Create new?
+    %tbody
+      - @import.collections.each do |collection|
+        %tr
+          %th= collection[:name]
+          - @import.collection_relations.each do |relation, _|
```
ttoomey:
```suggestion
          - @import.collection_relations.each_key do |relation|
```

app/views/admin/acl_imports/_users.haml
```diff
@@ -0,0 +1,103 @@
...
+        - @import.users.each do |user|
+          - next if user[:existing_user_id]
+
+          %tr
+            %td= user[:first_name]
+            %td= user[:last_name]
+            %td= user[:email]
+            %td= user[:agency]
+.mb-4
+  %h2 Agencies
```
ttoomey, line 68: maybe this and User Groups should have their own tabs? Otherwise someone would have to scroll through all the users to the bottom

app/views/admin/acl_imports/_table.haml
```diff
@@ -0,0 +1,24 @@
...
+      %thead
+        %tr
+          %th Uploaded
+          %th Uploaded By
+          %th Status
+          %th
+      %tbody
+        - @imports.each do |import|
+          %tr
+            %td= link_to_if import.accessible?, import.created_at.to_date, admin_acl_import_path(import)
```
ttoomey, line 15: I wonder if rather than linking the date on these two columns, it would be clearer to show the file name and make that the link text:
```
            %td= link_to_if import.accessible?, import.display_file_name, admin_acl_import_path(import)
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+  private def valid_collections?
+    collections_worksheet.cell_at('A1').value == 'Collection' && collections_worksheet.cell_at('B1').value == 'Data Sources'
+  end
+
+  private def existing_role(role_name)
+    existing_roles[role_name]
+  end
+
+  private def existing_roles
+    @existing_roles ||= Role.editable.index_by(&:name)
```
ttoomey: fwiw, not sure that role has a unique constraint on name (or even a rails validation). Don't need to fix that but maybe just order them so the behavior is reproducible (same on these collections and user_groups)
```suggestion
    @existing_roles ||= Role.editable.order(:id).index_by(&:name)
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+    end
+  end
+
+  private def user_data
+    users_worksheet.map do |row|
+      # First row is headers
+      next unless row.index_in_collection.positive?
+
+      # NOTE: order based on template
+      email = row.cells[2].value
```
ttoomey, line 207: I feel like it's safer and easer to use header names rather than positional indexes to reference columns. That's pretty easy with roo iirc; not sure about rubyXL though. Probably not worth changing at this point

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+      user.restore if user.deleted?
+      user.save!
+    end
+  end
+
+  private def import_user_groups!
+    user_groups.values.each do |item|
+      group = UserGroup.where(name: item[:name]).first_or_create!
+      user_ids = User.where(email: item[:users]).pluck(:id)
+      group.update(user_ids: user_ids)
```
ttoomey:
```suggestion
      group.update!(user_ids: user_ids)
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+    agency_ids = Agency.pluck(:name, :id).to_h
+    users.each do |item|
+      user = User.with_deleted.where(email: item[:email]).first_or_initialize do |u|
+        u.password = Faker::Internet.password(min_length: 16)
+      end
+      user.first_name = item[:first_name]
+      user.last_name = item[:last_name]
+      user.agency_id = agency_ids[item[:agency]]
+      user.permission_context = 'acls'
+      user.restore if user.deleted?
```
ttoomey, line 284: Are there any potential unexpected side effects of restoring a user? Maybe it would be safer to create a new user record?

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+  end
+
+  private def import_collections!
+    collections.each do |item|
+      collection = Collection.where(name: item[:name], collection_type: collection_type(item)).first_or_create!
+      collection_relations.each_key do |relation|
+        names = item[relation].select { |m| m[:found] }.map { |r| r[:name] }
+        case relation
+        when :coc_codes
+          collection.update(coc_codes: names) if names.present?
```
ttoomey:
```suggestion
          collection.update!(coc_codes: names) if names.present?
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+      end
+    end
+  end
+
+  private def permissions_for_row(row, existing_role)
+    {}.tap do |p|
+      Role.permissions(exclude_health: true).count.times do |i|
+        title = titles[i]
+        value = case row[2 + i]&.value
+        when 'X', 'x', 'Y', 'y', 'TRUE', 'true'
```
ttoomey: maybe use regex. Also "yes" is valid?
```suggestion
        when /\A(?:x|y|true|yes)\z/i
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+          true
+        else
+          false
```
ttoomey: maybe check for 'negative' values too and raise if we get nonsense
```suggestion
        when /\A(?:n|no|false)\z/
          false
        when nil
          false
        else
          # or track this as an error somehow
          raise ArgumentError, "Unknown value: #{row[2 + i]&.value}"
        end
```

app/models/access_control_upload.rb
```diff
@@ -0,0 +1,463 @@
...
+        h << title if title.present?
+      end
+    end
+  end
+
+  private def permissions_for_row(row, existing_role)
+    {}.tap do |p|
+      Role.permissions(exclude_health: true).count.times do |i|
+        title = titles[i]
+        value = case row[2 + i]&.value
```
ttoomey:
```suggestion
        value = case row[2 + i]&.value&.strip&.presence
```

app/controllers/admin/acl_imports_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+  end
+
+  def create
+    @import = import_scope.create!(import_params.merge(user_id: current_user.id, status: AccessControlUpload::UPLOADED))
+    @import.delay(queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)).pre_process!
+    respond_with(@import, location: admin_acl_imports_path)
+  end
+
+  # Completes the import using the file specified
+  def update
```
ttoomey, line 29: should this bail if @import.status != 'processed'?

app/controllers/admin/acl_imports_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    @import.delay(queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)).import!
+    @import.update(status: AccessControlUpload::IMPORTING)
+    respond_with(@import, location: admin_acl_imports_path)
+  end
+
+  def destroy
+    @import.destroy
```
ttoomey:
```suggestion
    @import.destroy!
```

ttoomey (approved):
Looks good, thanks for the cleanups! I left some suggestions and notes but nothing that should be a blocker

app/views/admin/groups/download.xlsx.axlsx
eanders: Hmm, yeah, or more to the point, I should probably use the preloaded roles.  Thanks!

app/models/concerns/user_concern.rb
eanders: Hmm, so, OP has actually used the space after the bang for negation since it's much easier to see the negation.  Perhaps we should discuss this and standardize.

app/models/access_control_upload.rb
eanders: Agreed.  I didn't see a built-in way to do this with RubyXL.  Holding off for now.

app/models/access_control_upload.rb
eanders: The system won't allow multiple users with the same email, and for audit purposes we don't want to completely delete previous users.  There shouldn't be unexpected side-effects, except, maybe we want to force a password reset...

app/models/access_control_upload.rb
eanders: Yeah, that definitely has occurred to me.  Project doesn't either, but we're using it as a primary key for importing collections.  We'll have to see if that becomes an issue.

app/views/admin/acl_imports/_users.haml
eanders: I stuck them together because there's shared context between them all that can be helpful to see.  If we disconnect the agencies from the users, then we can't tell which row in the spreadsheet caused the agency to b added, same for user groups.  I definitely waffled on this though, so I'll leave it for now, but we can ask for feedback as it rolls out.

PR #4457, opened 2024-06-13
Patch client headshot issue by returning null instead of empty string

gigxz:

## Description

[This change](https://github.com/greenriver/hmis-warehouse/pull/4412/files) made it so that the function `def image(_cache_for = 10.minutes)` would return an empty string `''` in some cases (the empty string comes from the function `image_for_source_client`).

That resulted in the below line evaluating, which results in [Argument Error :type option required](https://green-river.sentry.io/issues/5195363249/?alert_rule_id=12523843&alert_type=issue&environment=production&notification_uuid=5fad2771-6712-4eee-92ee-1c3bddab8db2&project=4503977346662400&referrer=slack) because `::MimeMagic.by_magic('')` returns nil.
https://github.com/greenriver/hmis-warehouse/blob/3b2b80c82b9eb47cb32e00a46c3f79412ca6a9f6/drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb#L120

This changes makes it so the `image` function returns nil, which results in `image && ! Rails.env.test?` evaluating to falsy, and this the `send_data` method is not invoked. (Which is correct, because there is no image). Tested with the client id from the linked Sentry.

## Type of change
- [x] Bug fix

### DIFF
--- a/app/controllers/source_clients_controller.rb
+++ b/app/controllers/source_clients_controller.rb
@@ -62,7 +62,7 @@ class SourceClientsController < ApplicationController
     expires_in max_age, public: false
     image = @client.image_for_source_client(max_age)
     # NOTE: The test environment is really unhappy when there's no image
-    if image && ! Rails.env.test?
+    if image.present? && ! Rails.env.test?
       send_data image, type: ::MimeMagic.by_magic(image), disposition: 'inline'
     else
       head(:forbidden)
--- a/app/controllers/source_clients_controller.rb
+++ b/app/controllers/source_clients_controller.rb
@@ -62,7 +62,7 @@ class SourceClientsController < ApplicationController
     expires_in max_age, public: false
     image = @client.image_for_source_client(max_age)
     # NOTE: The test environment is really unhappy when there's no image
-    if !image.blank? && !Rails.env.test?
+    if !image.empty? && !Rails.env.test?
       send_data image, type: ::MimeMagic.by_magic(image), disposition: 'inline'
     else
       head(:forbidden)
--- a/app/controllers/source_clients_controller.rb
+++ b/app/controllers/source_clients_controller.rb
@@ -62,7 +62,7 @@ class SourceClientsController < ApplicationController
     expires_in max_age, public: false
     image = @client.image_for_source_client(max_age)
     # NOTE: The test environment is really unhappy when there's no image
-    if image.present? && ! Rails.env.test?
+    if !image.blank? && !Rails.env.test?
       send_data image, type: ::MimeMagic.by_magic(image), disposition: 'inline'
     else
       head(:forbidden)
--- a/app/models/concerns/client_file_base.rb
+++ b/app/models/concerns/client_file_base.rb
@@ -64,13 +64,23 @@ module ClientFileBase
     def as_preview
       return client_file.download unless client_file.variable?
 
-      client_file.variant(resize_to_limit: [1920, 1080]).processed.download
+      begin
+        client_file.variant(resize_to_limit: [1920, 1080]).processed.download
+      rescue ActiveStorage::FileNotFoundError
+        Rails.logger.warn('Could not find client file')
+        return nil
+      end
     end
 
     def as_thumb
       return nil unless client_file.variable?
 
-      client_file.variant(resize_to_limit: [400, 400]).processed.download
+      begin
+        client_file.variant(resize_to_limit: [400, 400]).processed.download
+      rescue ActiveStorage::FileNotFoundError
+        Rails.logger.warn('Could not find client file')
+        return nil
+      end
     end
   end
 end
--- a/app/models/concerns/client_image_consumer.rb
+++ b/app/models/concerns/client_image_consumer.rb
@@ -18,7 +18,7 @@ module ClientImageConsumer
 
       # Check first for locally uploaded image or a cached ETO image
       # Otherwise, check for an image connected to any source client
-      local_client_image_data || source_clients.detect(&:image_for_source_client)&.image_for_source_client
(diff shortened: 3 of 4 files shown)


app/models/concerns/client_image_consumer.rb
```diff
@@ -14,7 +14,7 @@ module ClientImageConsumer
     def image(_cache_for = 10.minutes)
       # Check first for locally uploaded image or a cached ETO image
       # Otherwise, check for an image connected to the source client
-      local_client_image_data || source_clients.detect(&:image_for_source_client)&.image_for_source_client
+      local_client_image_data.presence || source_clients.detect(&:image_for_source_client)&.image_for_source_client.presence
     end
 
     def image_for_source_client(_cache_for = 10.minutes)
```
gigxz, line 24: I'm not sure why this method returns an empty string when there is no image, but it seems like that's a holdover from `stable`.

Here is the code from stable, which explicitly returns `''` in some cases and `nil` in others. @eanders Do you know why that is?

https://github.com/greenriver/hmis-warehouse/blob/6702d55a7a29cba68b29bbf0342cadd83de7754a/app/models/grda_warehouse/hud/client.rb#L1484-L1505

app/controllers/source_clients_controller.rb
```diff
@@ -62,7 +62,7 @@ def image
     expires_in max_age, public: false
     image = @client.image_for_source_client(max_age)
     # NOTE: The test environment is really unhappy when there's no image
-    if image && ! Rails.env.test?
+    if image.present? && ! Rails.env.test?
       send_data image, type: ::MimeMagic.by_magic(image), disposition: 'inline'
     else
```
gigxz: I haven't seen this one appear in Sentry, but I would imagine it has the same issue. `'' && true` evaluates to true.

In this case I am changing it here within the controller rather than returning nil from `image_for_source_client`, because I am uncertain about whether there is significance to returning '' from that method (See comment on https://github.com/greenriver/hmis-warehouse/pull/4457/files#r1639054025).

Note: the similar code in the ClientsController is:
https://github.com/greenriver/hmis-warehouse/blob/3b2b80c82b9eb47cb32e00a46c3f79412ca6a9f6/drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb#L118-L121

app/models/concerns/client_image_consumer.rb
eanders: Yeah, thanks, we were continuing to return an empty string since that's what it used to do, but we didn't investigate further to determine if that was the correct behavior!  We should probably convert them all to nils, but what you have should work for now.  Thanks!

gigxz: What I had actually didn't work, calling `.presence` on the image string will fail with `ArgumentError: invalid byte sequence in UTF-8`. You can call `blank?` or `empty?` but not `present?`... the `client_visibility_config_variations_legacy_spec` helped me figure that out. I will push another change

app/models/concerns/client_image_consumer.rb
```diff
@@ -10,11 +10,15 @@ module ClientImageConsumer
...
-    # Returns actual image bytes. Cache time limit not yet implemented.
+    # Cache time limit not yet implemented
+    # @return [String, nil] actual image bytes.
     def image(_cache_for = 10.minutes)
+      # If we call `image` on a source client, return the attached image
+      return image_for_source_client if source?
+
       # Check first for locally uploaded image or a cached ETO image
-      # Otherwise, check for an image connected to the source client
-      local_client_image_data || source_clients.detect(&:image_for_source_client)&.image_for_source_client
```
eanders, line 17: This was the one root cause.  Turns out detect was allowing an empty string to pass for a `true`.

app/models/concerns/client_file_base.rb
```diff
@@ -64,13 +64,23 @@ def self.available_tags
     def as_preview
       return client_file.download unless client_file.variable?
 
-      client_file.variant(resize_to_limit: [1920, 1080]).processed.download
+      begin
+        client_file.variant(resize_to_limit: [1920, 1080]).processed.download
+      rescue ActiveStorage::FileNotFoundError
```
ttoomey, line 69: is it right to hide these exceptions? It seems like something has gone wrong if this is occurring. Maybe send to sentry?

app/models/concerns/client_image_consumer.rb
```diff
@@ -10,11 +10,15 @@ module ClientImageConsumer
...
+      # Otherwise, check for an image connected to any source client
+      local_client_image_data || source_clients.detect { |sc| sc.image_for_source_client.length > 100 }&.image_for_source_client
```
ttoomey: looks like image_for_source_client can return nil, maybe protect against that? otherwise calling length will crash
```suggestion
      local_client_image_data || source_clients.detect { |sc| (sc.image_for_source_client || '').length > 100 }&.image_for_source_client
```

app/models/concerns/client_image_consumer.rb
ttoomey: good catch!

I wonder if we should try and set the encoding correctly on these data strings to avoid this type of exception. For example `image_for_source_client` should call something like `data.force_encoding(Encoding::BINARY))` before returning

ttoomey (approved):
Looks good, I think the only issue is that you might still crash if `image_for_source_client` is nil (see comment). Maybe that doesn't happen in practice though

app/models/concerns/client_file_base.rb
eanders: I've only ever seen this development (because we clear our minio storage accidentally.)  Given that this is just the client headshot, I think it's ok.

PR #4463, opened 2024-06-17
Access Control system project group and cohort collection bug fix

eanders:

## Description

This:
1. Fixes a bug in the `system` scope on Role
2. This bug introduced duplicate system Cohorts and Project Groups, so a migration is included to cleanup the duplicates
3. Adds missing `collection_type` assignment to system Collections
4. Attempts to assign `collection_type` to existing system Collections 
5. Tightens when a session is deemed `invalid` by JavaScript.  This change doesn't affect how session validity is determined, only if the browser throws up a big warning that your session is invalid in one scenario where it often is still valid

## Type of change
- [x] Bug fix

### DIFF
--- a/app/assets/javascripts/inactive-session-modal-controller.es6
+++ b/app/assets/javascripts/inactive-session-modal-controller.es6
@@ -119,7 +119,6 @@ App.StimulusApp.register(
       // Note, when you logout state.userId is set to undefined, there seems to be
       // some situations where it is returned as the string "null", maybe a race condition?
       // For now, don't mark those as invalid
-      console.log(state)
       if (state.userId !== this.initialUserIdValue && state.userId !== 'null') {
         // another tab has logged out or changed user
         state.invalid = true;
--- a/app/assets/javascripts/inactive-session-modal-controller.es6
+++ b/app/assets/javascripts/inactive-session-modal-controller.es6
@@ -116,7 +116,11 @@ App.StimulusApp.register(
         this.mainLoopInterval = setTimeout(() => this.mainLoop(), timeout);
       }
 
-      if (state.userId !== this.initialUserIdValue) {
+      // Note, when you logout state.userId is set to undefined, there seems to be
+      // some situations where it is returned as the string "null", maybe a race condition?
+      // For now, don't mark those as invalid
+      console.log(state)
+      if (state.userId !== this.initialUserIdValue && state.userId !== 'null') {
         // another tab has logged out or changed user
         state.invalid = true;
       } else if (state.remaining === 0) {
--- a/app/models/role.rb
+++ b/app/models/role.rb
@@ -41,7 +41,7 @@ class Role < ApplicationRecord
   end
 
   replace_scope :system, -> do
-    not_system.where(system: true)
+    where(system: true)
   end
 
   scope :not_system, -> do
--- a/app/views/admin/collections/index.haml
+++ b/app/views/admin/collections/index.haml
@@ -13,7 +13,7 @@
       %span.icon-plus
       Add a Collection
     .w-50.ml-auto
-      - @prompt = 'Search user groups...'
+      - @prompt = 'Search collections...'
       = render 'search_form'
   - if @pagy.count.positive?
     = render 'common/pagination_top', item_name: 'group'
--- a/app/views/admin/collections/index.haml
+++ b/app/views/admin/collections/index.haml
@@ -13,7 +13,7 @@
       %span.icon-plus
       Add a Collection
     .w-50.ml-auto
-      - @prompt = 'Search collections...'
+      - @prompt = 'Search collections…'
       = render 'search_form'
   - if @pagy.count.positive?
     = render 'common/pagination_top', item_name: 'group'
--- a/db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
+++ b/db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
@@ -1,0 +1,59 @@
+class CleanupSystemRolesAndAccessControls < ActiveRecord::Migration[7.0]
+  # Cleanup a bug in the system scope on Role
+  def up
+    [
+      'System Role - Can Edit Project Groups',
+      'System Role - Can Participate in Cohorts',
+      'System Role - Can View Cohorts',
+    ].each do |name|
+      keep_id = Role.where(name: name, system: true).minimum(:id)
+      delete_ids = Role.where(name: name, system: true).where.not(id: keep_id).pluck(:id)
+      # Find the unique access controls based on collection_id, and user_group_id that use the
+      # roles in deleted_ids.
+      # Keep the first of each of those, updating the role_id to be keep_id, unless one already exists
+      # using role_id: keep_id
+      # Remove any Access Controls that use delete_ids
+      combinations = {}
+      AccessControl.where(role_id: delete_ids).find_each do |ac|
+        combinations[[ac.collection_id, ac.user_group_id]] ||= []
+        combinations[[ac.collection_id, ac.user_group_id]] << ac
+      end
+      combinations.each do |(collection_id, user_group_id), access_controls|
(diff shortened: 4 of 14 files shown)


app/assets/javascripts/inactive-session-modal-controller.es6
```diff
@@ -116,7 +116,10 @@ App.StimulusApp.register(
         this.mainLoopInterval = setTimeout(() => this.mainLoop(), timeout);
       }
 
-      if (state.userId !== this.initialUserIdValue) {
+      // Note, when you logout state.userId is set to undefined, there seems to be
+      // some situations where it is returned as the string "null", maybe a race condition?
+      // For now, don't mark those as invalid
+      if (state.userId !== this.initialUserIdValue && state.userId !== 'null') {
```
eanders, line 122: @ttoomey let me know if you have a better approach for this, in testing, this seems to fix the issue I've been seeing in development.

app/models/role.rb
```diff
@@ -41,7 +41,7 @@ def role_name
   end
 
   replace_scope :system, -> do
-    not_system.where(system: true)
```
eanders, line 44: This was the root cause of the bug since it resulted in `where system = false and system = true`.

app/assets/javascripts/inactive-session-modal-controller.es6
ttoomey: I'm guessing that null is getting serialized and read back from localStorage as 'null'  on line 104. Maybe we could avoid serializing falsy values with:
```diff
 const shared = {
   saveValue: (key, value) => {
-    window.localStorage.setItem(key, String(value));
+    if (value) {
+      window.localStorage.setItem(key, String(value));
+    } else {
+      window.localStorage.removeItem(key);
+    }
   },
```

app/views/admin/collections/index.haml
```diff
@@ -13,7 +13,7 @@
       %span.icon-plus
       Add a Collection
     .w-50.ml-auto
-      - @prompt = 'Search user groups...'
+      - @prompt = 'Search collections...'
```
ttoomey: maybe better to use the ellipsis char?
```suggestion
      - @prompt = 'Search collections…'
```

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      collection_type = case types.first
+      when 'GrdaWarehouse::ProjectGroup'
+        'Project Groups'
+      when 'GrdaWarehouse::Cohort'
+        'Cohorts'
+      when 'GrdaWarehouse::DataSource', 'GrdaWarehouse::Hud::Organization', 'GrdaWarehouse::Hud::Project'
+        'Projects'
+      end
+      collection.update(collection_type: collection_type)
```
ttoomey:
```suggestion
      collection.update!(collection_type: collection_type)
```

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
```diff
@@ -0,0 +1,59 @@
...
+    end
+
+    # Cleanup any exact duplicate Access Controls (this is probably just a side-effect of running the
+    # migration a bunch of times, but there's no harm making sure)
+    controls = AccessControl.system.group_by { |ac| [ac.collection_id, ac.user_group_id, ac.role_id] }
+    controls.each do |(collection_id, user_group_id, role_id), access_controls|
+      next if access_controls.count == 1
+
+      # Keep the one with the lowest ID
+      access_controls.sort_by(&:id).drop(1).each(&:destroy)
```
ttoomey:
```suggestion
      access_controls.sort_by(&:id).drop(1).each(&:destroy!)
```

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+      combinations.each do |(collection_id, user_group_id), access_controls|
+        # We already have one that uses the role we want to keep
+        next if access_controls.any? { |ac| ac.role_id == keep_id }
+
+        puts "Found #{access_controls.count} Access Controls that should have been using the role #{keep_id}"
+        access_controls.min_by(&:id).update(role_id: keep_id)
+      end
+      AccessControl.where(role_id: delete_ids).destroy_all
+      Role.where(id: delete_ids).destroy_all
```
ttoomey:
```suggestion
      Role.where(id: delete_ids).each(&:destroy!)
```

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
```diff
@@ -0,0 +1,59 @@
...
+      AccessControl.where(role_id: delete_ids).find_each do |ac|
+        combinations[[ac.collection_id, ac.user_group_id]] ||= []
+        combinations[[ac.collection_id, ac.user_group_id]] << ac
+      end
+      combinations.each do |(collection_id, user_group_id), access_controls|
+        # We already have one that uses the role we want to keep
+        next if access_controls.any? { |ac| ac.role_id == keep_id }
+
+        puts "Found #{access_controls.count} Access Controls that should have been using the role #{keep_id}"
+        access_controls.min_by(&:id).update(role_id: keep_id)
```
ttoomey:
```suggestion
        access_controls.min_by(&:id).update!(role_id: keep_id)
```

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
```diff
@@ -0,0 +1,59 @@
...
+      end
+      AccessControl.where(role_id: delete_ids).destroy_all
+      Role.where(id: delete_ids).destroy_all
+    end
+
+    # Cleanup any exact duplicate Access Controls (this is probably just a side-effect of running the
+    # migration a bunch of times, but there's no harm making sure)
+    controls = AccessControl.system.group_by { |ac| [ac.collection_id, ac.user_group_id, ac.role_id] }
```
ttoomey, line 34: Maybe there should be a unique constraint to prevent duplicates like this?

ttoomey (approved):
I think I'm following this except I'm not quite sure about collection.collection_type

Maybe we intended to support heterogeneous collections at one point but then decided not to so added this as a restriction? The old untyped collections are left in-place as 'legacy' collections? (different from user-role "legacy" permissions). Or maybe we only care about collection type for 'system' collections?

Perhaps we should validate that new collections have a non-null collection-type that is part of 'collection type'.

Also we use human labels for 'collection_type' instead of class names for STI, I guess that's just for display?

db/migrate/20240617132532_cleanup_system_roles_and_access_controls.rb
eanders: I'll add that to the backlog.  Thanks.

eanders:
> I think I'm following this except I'm not quite sure about collection.collection_type
> 
> Maybe we intended to support heterogeneous collections at one point but then decided not to so added this as a restriction? The old untyped collections are left in-place as 'legacy' collections? (different from user-role "legacy" permissions). Or maybe we only care about collection type for 'system' collections?
> 
> Perhaps we should validate that new collections have a non-null collection-type that is part of 'collection type'.
> 
> Also we use human labels for 'collection_type' instead of class names for STI, I guess that's just for display?

The short version is that collections were migrated from `AccessGroup` initially, which didn't have a specific type.  We later decided the UI/UX was better if we could show just one type at a time, so we added `collection_type` as a way to differentiate in the UI.  However, some Access Groups granted access to more than one type of thing, so to keep that functionality, we leave the `collection_type` blank.  

I agree, we should add validation on create that `collection_type` is present (and maybe in the known types.)  Thanks for the suggestion!  `collection_type` was never intended to be the `type` column, so does not imply STI, though it does use the suffix `_type`.

PR #4474, opened 2024-06-20
Fix issue where acts as paranoid and paper trail weren't persisting correctly for changes made to user groups

gigxz:

## Description

Ensure that UserGroupMember changes are tracked in paper trail and that they don't get hard-deleted.

To serve https://github.com/open-path/Green-River/issues/5790 eventually

## Type of change
- [x] Bug fix


app/models/user_group.rb
```diff
@@ -53,6 +53,7 @@ def self.text_search(text)
   def add(users)
     # Force individual queries for paper_trail
     Array.wrap(users).uniq.each do |user|
+      # should we use with_deleted to restore the user_group_member if it was previously deleted?
```
gigxz: Is this preferred?I think we use this pattern for group viewable entities, so I can update this for consistency if desired. Currently it will create a new record if user is removed and re-added to the group.

app/controllers/admin/user_groups_controller.rb
```diff
@@ -30,8 +30,16 @@ def edit
...
-      @group.save
+      users = User.where(id: group_params[:user_ids])
+      users_to_remove = @group.users - users
+      @group.add(users) # add with paper trail
+      @group.remove(users_to_remove) # destroy with paranoia
+
+      if group_params[:name]
+        @group.name = group_params[:name]
+        @group.save!
+      end
```
eanders: maybe the following, just to catch the situation where we add other fields that should be saved?
(untested)
```suggestion
      @group.update!(group_params.except(:user_ids))
```

app/models/user_group.rb
eanders: if we have paper trail enabled, then yeah, I vote for `with_deleted` so we re-use the previous member.

drivers/hmis/app/controllers/hmis_admin/user_groups_controller.rb
```diff
@@ -31,8 +31,15 @@ def edit
...
-    @user_group.save
+    users = Hmis::User.where(id: user_group_params[:user_ids])
+    users_to_remove = @user_group.users - users
+    @user_group.add(users) # add with papertrail
+    @user_group.remove(users_to_remove) # destroy with paranoia
+
+    if user_group_params[:name].present?
+      @user_group.name = user_group_params[:name]
+      @user_group.save!
+    end
```
eanders: Same as above here:

```suggestion
    @group.update!(group_params.except(:user_ids))
```

drivers/hmis/app/models/hmis/user_group.rb
```diff
@@ -42,6 +42,7 @@ def self.system_user_group
   def add(users)
     # Force individual queries for paper_trail
     Array.wrap(users).uniq.each do |user|
+      # should we use with_deleted to rescucitate the user_group_member if it was previously deleted?
```
eanders: Yes please!

app/models/user_group.rb
gigxz: ok, updated

app/models/user_group.rb
```diff
@@ -53,7 +53,9 @@ def self.text_search(text)
   def add(users)
     # Force individual queries for paper_trail
     Array.wrap(users).uniq.each do |user|
-      user_group_members.where(user_id: user.id).first_or_create!
+      member = user_group_members.with_deleted.where(user_id: user.id).first_or_create!
+      member.restore if member.deleted?
```
eanders, line 57: @gigxz do you think we should have some tests for this scenario where there's a deleted record being restored?

eanders (approved):
@gigxz feel free to merge if you don't think we want to invest in additional tests at this time.

app/models/user_group.rb
gigxz: I added some assertions to https://github.com/greenriver/hmis-warehouse/pull/4474/commits/6c55ed4cfe002db73aacf9c15e5a3100e47d4804

eanders: Thank you!

PR #4475, opened 2024-06-20
Split out `custom_rule` from hud `rule` field

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6164

- [x] add rest of rules to module (once agreed on approach)
- [x] add enforcement logic for Data Collected About
- [x] add validation for missing link ids in HUD assessments
- [x] add test:
  - [x] overwriting rule
  - [x] failing if link id is missing
  - [x] overwriting DCA

## Type of change


drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -19,6 +19,8 @@ def resolve(id:, input:)
       raise 'only allowed to modify draft forms' unless definition.draft?
       raise 'not allowed to change identifier' if input.identifier.present? && input.identifier != definition.identifier
 
+      # TODO: enforce that you can't change the rule fields?
+
```
gigxz: I can't think of an elegant way to do this. My only thought is to recur on the submitted form, comparing each `link_id =>rule` and make sure nothing has changed.  Or I could "overlay" the rules from the "old" form onto the submitted one (again, based on link id).

drivers/hmis/app/models/hmis/form/definition_item_filter.rb
```diff
@@ -59,6 +62,9 @@ def eval_rule(rule)
     # but should always be shown when creating/editing a Client outside of a project context.
     return true if project.nil?
 
+    # TODO we should probably add an "always show" override. there are workarounds e.g. selecting all project types
+    # return true if rule.fetch('all_projects', false) == true
```
gigxz: Thoughts on this? Didn't do it directly because it was complicated to add to the JSON Schema

gigxz:
@ttoomey I'd appreciate a look at my TODOs here before I go further. Thank you!

drivers/hmis/app/graphql/mutations/update_form_definition.rb
ttoomey: It seems like we could just modify `recursively_transform()` so that it overwrites the `rules` property since we are walking the tree anyway. However what happens if an element is removed by the user and then added back (or some other weirdness). Feels like in that case we'd want to pull the rules from a static source instead of the previously persisted form definition. I guess you'd want a lookup for [form_role, link_id] => {rules...}. This is sort of a combo of options 3 and 4 in 6164

drivers/hmis/app/models/hmis/form/definition_item_filter.rb
ttoomey: One way to to do this would be to add a 'MATCH' operator? Then we'd get this as a side-effect with:
```ruby
{ variable: 'projectType', operator: 'MATCH', value: ".*" }
```

I am a little surprised that it's a common need to ignore the hud rules

drivers/hmis/app/models/hmis/form/definition_item_filter.rb
```diff
@@ -32,7 +32,10 @@ def eval_items(items)
     # return items if Rails.env.development?
 
     items.filter do |item|
-      if eval_rule(item['rule'])
+      true_for_hud_rule = eval_rule(item['rule'])            # HUD system rules that cannot be overridden
+      true_for_custom_rule = eval_rule(item['custom_rule'])  # Custom rules that can be managed by admins
+
+      if true_for_hud_rule || true_for_custom_rule
```
ttoomey: micro performance optimization would be to use lazy eval here so that we only need to run the first eval_rule if it's true
```suggestion
      if eval_rule(item['rule'])  || eval_rule(item['custom_rule'])
```

drivers/hmis/app/models/hmis/form/definition_item_filter.rb
gigxz: Actually we probably don't need this. We only do it for the QA forms where we want to show everything all the time.  I'll remove this comment

drivers/hmis/app/graphql/mutations/update_form_definition.rb
gigxz: Here is an example where we move the "Static source" from JSON files into code: https://github.com/greenriver/hmis-warehouse/pull/4478/files -- is that sort of what you had in mind?

I guess the other point here is that we could rely on the frontend to ensure that `rule`s aren't changed. (It should not provide an interface for editing them). Obviously someone could change them with the API in that case, but that's not a dangerous operation, just not ideal for HUD Compliance. So if we're fine with that, we could just this as-is.

The other field that's important is `data_collected_about`. That field can be changed by the user but only expanded beyond the HUD required "minimum", so we still need a mechanism for that. The linked branch has an approach, but there are other ways we could do it.

ttoomey: > Here is an example where we move the "Static source" from JSON files into code: https://github.com/greenriver/hmis-warehouse/pull/4478/files -- is that sort of what you had in mind?

Yes, that's what I was thinking.
> I guess the other point here is that we could rely on the frontend to ensure that rules aren't changed.

The goal is to prevent folks from accidentally getting out of compliance when customizing their forms right? I could see an approach where we assume the front-end will always preserve the hidden "hud_rules". If an admin removes required elements or there's a client bug that clobbers parts of the form, then we could just decide not to handle that since it is unlikely.

But I don't quite understand the consequences of someone falling out of compliance. I'm guessing it could be kind of a big problem particularly if nobody catches it for a while. It seems like we could ward this off with a combination of validation and forcing to known-good values:
* validate that the appropriate fields are present in the form given the role
* validate or force hud_rules and data_collected_about to known-good values (maybe others)

If this feels tricky to do on every save, maybe it could be something that occurs during publication. For example you can't publish a form unless it passes a hud-compliance check

gigxz: Yeah I agree with that -- it could be a problem if non-compliant for a while. I think it makes sense to ensure compliance on publish. I'll move forward with that.

drivers/hmis/lib/hmis_util/hud_form_rules_2024.rb
```diff
@@ -0,0 +1,333 @@
...
+  class HudFormRules2024
+    # Keys match Link IDs in our Form Definition fragments for HUD data elements
+    HUD_LINK_ID_RULES = {
+      # CoC Code for Client Location
+      q_3_16: {
+        stages: [:INTAKE],
+        data_collected_about: :HOH,
+      },
+      # Prior Living Situation A
+      q_3_917A: {
```
gigxz: It works, but it feels like an anti-pattern to rely on a certain link id outside of the form logic. The alternative would be to add some new item attribute like `"hud_data_element_id": "3.917"` and do the checks based off that.

drivers/hmis/spec/models/hmis/form/definition_hud_requirements_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+        expect(fd.definition['item'][0]['data_collected_about']).to eq(expected_dca)
+      end
+    end
+  end
+
+  # do nothing for different role
+
+  # do nothing for non-hud fields
+
+  # error if hud field is missing from the form
```
gigxz: xx

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -257,6 +257,10 @@ def load_definition(form_definition:, identifier:, role:, title: nil)
       record.title = title if title.present?
       record.status = Hmis::Form::Definition::PUBLISHED
 
+      # Ensure HUD rules are set
+      changed_link_ids = record.set_hud_requirements # return chagned link ids
+      raise "hud rules wrong for #{identifier}: #{changed_link_ids}" if changed_link_ids.any?
+
```
gigxz: This is a check to ensure I didn't break anything. Once comfortable with this I can remove all the `"rule"` fields from the JSON in form_data, and we can rely on the `set_hud_requirements` method to set them according to what's in the module.

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2024.rb
```diff
@@ -0,0 +1,1134 @@
...
+                     [
+                       { 'variable' => 'projectFunderComponents', 'operator' => 'INCLUDE', 'value' => 'HHS: RHY' },
+                       { 'variable' => 'projectType', 'operator' => 'NOT_EQUAL', 'value' => 4 },
+                     ] },
+          { '_comment' => 'Included in YHDP Supplemental CSV. Recommended for YHDP projects.',
+            'variable' => 'projectFunders',
+            'operator' => 'INCLUDE',
+            'value' => 43 },
+        ] } },
+      # :V1=>
```
gigxz: commented out since its not part of hud assessments (its on client record)

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -23,7 +23,9 @@ def resolve(id:)
       )
 
       definition.status = Hmis::Form::Definition::PUBLISHED
-
+      # Ensure HUD requirements are set correctly (if applicable). This could mutate the definition.
+      definition.set_hud_requirements
```
gigxz, line 27: We have no validation to ensure that the content of the "hud fragments" are not modified inappropriately. We rely on the frontend to make sure that doesn't happen.

Looking at this with fresh eyes, it feels kind of arbitrary that we decided to strictly enforce/check that `rule` doesn't change, but we don't bother to check if the actual structure of the group (or item) matches what is expected. (See for example my test where I make "c3 education status" be a boolean).

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2024.rb
```diff
@@ -0,0 +1,1134 @@
...
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisUtil
+  class HudAssessmentFormRules2024
+    # Keys match Link IDs in our default HUD Assessments.
+
+    # def walk_definition_item(node, &block)
```
gigxz: I can get rid of this code comment, just showing how I generated the object below.

drivers/hmis/spec/models/hmis/form/definition_validator_spec.rb
```diff
@@ -0,0 +1,121 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+require 'rails_helper'
+require_relative '../../../support/hmis_base_setup'
+
+RSpec.describe Hmis::Form::DefinitionValidator, type: :model do
```
gigxz, line 10: these tests are not very comprehensive at all. could probably use more investment

ttoomey: There are some quick tests in `drivers/hmis/spec/models/hmis/form/json_schema_spec.rb` that run directly against the JsonValidator. Those could probably get removed; they are obviated by these tests

drivers/hmis/spec/models/hmis/form/definition_hud_requirements_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+          {
+            'type': 'DISPLAY',
+            'link_id': 'q_3_917A', # HUD link id
+            'label': 'fake',
+          },
+          { 'type': 'DISPLAY', 'link_id': 'foo', 'label': 'non-hud field' },
+        ],
+      }
+      fd = create(:hmis_form_definition, definition: mock_definition, role: :INTAKE, identifier: 'test_intake_form')
+      fd
```
ttoomey:
```suggestion
      create(:hmis_form_definition, definition: mock_definition, role: :INTAKE, identifier: 'test_intake_form')
```

drivers/hmis/spec/models/hmis/form/definition_hud_requirements_spec.rb
```diff
@@ -0,0 +1,86 @@
...
+    it 'does nothing if role is not required by hud' do
+      fd.update!(role: :UPDATE)
+
+      perform_and_reload
+
+      expect(fd.definition['item'][0]['data_collected_about']).to be_nil
+      expect(fd.definition['item'][0]['rule']).to be_nil
+    end
+
+    # Test against the PLS field which requires 'HOH_AND_ADULTS'
```
ttoomey: clarify acronym
```suggestion
    # Test against the Primary Language spoken (PLS) field which requires 'HOH_AND_ADULTS'
```

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2024.rb
ttoomey: I wonder if it would be worth it to use the json seed files rather than this HUD_LINK_ID_RULES constant. That would probably mean building this lookup table at run-time. But seems like it might be easier to maintain if there's one source of truth

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -578,4 +593,47 @@ def introspect_custom_data_element_definitions(set_definition_identifier: false)
...
+    rule_module = HmisUtil::HudAssessmentFormRules2024.new
+
+    walk_definition_nodes do |item|
+      link_id = item['link_id']
+
+      # Check if there is a required data collection rule for this link_id for this data collection stage (ie role)
+      hud_rule = rule_module.hud_data_element_rule(role, link_id)
+      if hud_rule
+        # If the rule is different from what's on the item, update it
+        difference = Hashdiff.diff(hud_rule, item['rule'], ignore_keys: '_comment')
```
ttoomey, line 615: Recommend adding the hashdiff gem as an explicit dependency to the Gemfile. It's in the project already but only because it's a secondary dependency of another gem.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -507,15 +507,30 @@ def external_form_object_key=(value)
     super(value.blank? ? nil : value.strip)
   end
 
-  def walk_definition_nodes(&block)
-    definition_struct&.item&.each do |node|
-      walk_definition_node(node, &block)
+  # Walk definition items either as open structs or hashes
+  def walk_definition_nodes(as_open_struct: false, &block)
+    if as_open_struct
```
ttoomey: This is fine, I guess I'm not sold that the utility of the struct variant is worth the additional complexity but we can leave it as is and refactor later

ttoomey (approved):
Looking good! I left a few suggestions

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2024.rb
gigxz: I was planning to remove the `rule`s from the JSON files so that we have one source of truth (per my comment here https://github.com/greenriver/hmis-warehouse/pull/4475#discussion_r1657141923). I guess it could go the other direction, but it feels like that might be more challenging to debug/understand since the values are spread across many files.

drivers/hmis/spec/models/hmis/form/definition_validator_spec.rb
gigxz: I moved one of those test cases to this file (link id pattern validation) and removed the json_schema_spec file

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2024.rb
gigxz: Rules are removed with https://github.com/greenriver/hmis-warehouse/pull/4475/commits/6262cebab93b74d6eb5f26e6f6893b8a34109647
Let me know if you had something else in mind @ttoomey.

gigxz:
@ttoomey Let me know if you want to re-review at all, otherwise I will merge tomorrow. Thanks!

PR #4480, opened 2024-06-24
Generate CustomDataElementDefinitions on publish

gigxz:

## Description

https://github.com/open-path/Green-River/issues/5776

Add missing `custom_field_key`s to new questions when published, and generate new CustomDataElementDefinitions for them.

- [x] add tests

## Type of change


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -549,39 +563,25 @@ def introspect_custom_data_element_definitions(set_definition_identifier: false)
 
     cded_records = []
     walk_definition_nodes do |item|
-      # skip unless custom_field_key specified for this item
+      # Skip non-questions items (Groups and Display items)
```
martha, line 555: I think the comments here are tripping me up a bit --
- line 566 says "skip non-question items", but line 567 skips _unless_ it's a non-question item, maybe this is a typo?
- line 572 says "skip items that already map to a custom data element," but I think what this line is actually doing is skipping items that don't specify a custom mapping in the json definition. Then on line 576 `cdeds_by_key[key]` that's where it is finding out whether this item already maps to an existing CDED. Am I understanding that correctly?

martha (approved):
Looks great, just made a note of some confusing comments in the existing routine for generating cdeds, but maybe that doesn't really matter since that code will likely be deleted soon?

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: good catch, thank you! updated to `next if` instead of `next unless`. This function (`introspect_custom_data_element_definitions`) does not have any tests and likely won't be used at all going forward, but I think it might make sense to keep it around for a bit in case its helpful for debugging issues

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -579,4 +576,28 @@ def introspect_custom_data_element_definitions(set_definition_identifier: false)
...
+    when 'CURRENCY'
+      'float'
+    else
+      raise "unable to determine cded type for #{item_type}"
+    end
+  end
+
+  # Helper for determining CustomDataElementDefinition attributes
+  def self.generate_cded_field_label(item)
+    label = item.readonly_text || item.brief_text || item.text || key.humanize
```
ttoomey: can we have empty strings here? Maybe should check with:
```suggestion
    label = item.readonly_text.presence || item.brief_text.presence || item.text.presence || key.humanize
```

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -25,13 +25,81 @@ def resolve(id:)
...
+    end
+
+    def ensure_unique_key(owner_type, key)
+      return key unless Hmis::Hud::CustomDataElementDefinition.exists?(owner_type: owner_type, key: key)
+
+      count = 1
+      possible_key = key
+      while Hmis::Hud::CustomDataElementDefinition.exists?(owner_type: owner_type, key: possible_key)
+        count += 1
+        possible_key = "#{key}_#{count}"
```
ttoomey, line 101: maybe add safety:
```suggestion
        possible_key = "#{key}_#{count}"
        raise if count > 50
```

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -25,13 +25,81 @@ def resolve(id:)
...
 
       Hmis::Form::Definition.transaction do
+        # Retire the previously published version
         previous_published_form&.update!(status: Hmis::Form::Definition::RETIRED)
+
+        # Add any missing custom field keys to the form definition (for new questions)
+        new_cdeds = add_missing_custom_field_keys(definition)
+        # Save any new CustomDataElementDefinitions
+        new_cdeds.each(&:save!)
+        # TODO: call new validation function, which should also validate that all referenced CDEDs exist
```
ttoomey, line 35: So there are attributes that are assigned to the CDED when the field is first published but it looks like they would not be updated on future publications: repeats, field_type, and label.

Maybe this is part of the TODO around validation but should we handle changes to the attributes on existing CDEDs on publish? Maybe label update would be harmless probably but attempts to change the `field_type` and `repeats` attrs could fail a validation check

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -25,13 +25,81 @@ def resolve(id:)
...
         definition.save!
       end
 
       {
         form_identifier: definition,
       }
     end
+
+    # Adds missing `{mapping: {custom_field_key: '...'}}` to all questions in the form definition
+    # Returns list of new CustomDataElementDefinitions to be saved
```
ttoomey:
```suggestion
    # mutates definition.items, adds `mapping.custom_field_key`
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -508,26 +511,37 @@ def external_form_object_key=(value)
     super(value.blank? ? nil : value.strip)
   end
 
+  # Walk definition items as OpenStructs
```
ttoomey: We have both walk_definition_nodes and walk_definition_items. Could we remove walk_definition_nodes at this point? Looks like we'd just need to adjust `introspect_custom_data_element_definitions`

ttoomey (approved):
Looks good! Left a few suggestions

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: The open struct options is nice for read-only traversal, which I feel like might be useful elsewhere. I combined them into one outer method https://github.com/greenriver/hmis-warehouse/pull/4480/commits/89bd6ae152de698ac33866e499fd5ea18b41619c how's that?

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
gigxz: Yep, validating field_type/repeats/owner_type is part of the TODO!

I was planning to just leave the `label` is it was when first published. Maybe later we introduce a way to change it manually if the label is important in reports or something.

I will probably want to target the CDED validations against https://github.com/greenriver/hmis-warehouse/pull/4487, so I'll open a separate PR.

PR #4487, opened 2024-06-26
Move all validations to DefinitionValidator

gigxz:

## Description

* Stop relying on the `JsonForms` module for validation. Instead, use the DefinitionValidator. Add instance method `validate_json_form` on the Definition that invokes the validator. This could also be done in a validation context, but I left it as a manual call
* Return `HmisErrors::Error`s from the validator so they can be resolved with the api. I could move this transformation logic somewhere else if it feels wrong for the DefinitionValidator @ttoomey -- but my thought was we might want to make use of the additional attributes that the err object provides (eg severity level)
* Add missing reference validation for `autofill_when`

## Type of change
- [x] Bug fix


gigxz:
@ttoomey I separated this out from my work on:
* https://github.com/greenriver/hmis-warehouse/pull/4475 which will validate that HUD assessments are HUD conformant
* https://github.com/greenriver/hmis-warehouse/pull/4480 which will validate that CDEDs exist and are valid (re: your comment there) (might be a follow-up pr)

because I do not like the pattern that we have mutations using the validation in `JsonForms` util module which really exists to seed the db. It didn't feel like the right flow of things. I moved all the validations into DefinitionValidator, and have it return HmisErrors which can be resolved and can have additional context on them (like severity level, for example).

What do you think?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -336,11 +336,10 @@ def self.find_definition_for_service_type(service_type, project:)
...
-  # Validate JSON definition when loading, to ensure no duplicate link IDs
-  def self.validate_json(...)
-    Hmis::Form::DefinitionValidator.perform(...).each do |issue|
-      yield issue
-    end
+  # returns list of HmisErrors::Error objects
+  # rename to like validate_form or something?
+  def validate_json
+    Hmis::Form::DefinitionValidator.perform(definition, role)
   end
```
gigxz: +gig update comment to be more descriptive.
This is the main validation method for validating the form content, including it's JSON structure, HUD conformance (if applicable), etc.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -71,11 +95,41 @@ def check_references(document, all_ids)
...
         link_check.call(child_item)
       end
     end
     link_check.call(document)
   end
+
+  # Introspect on GraphQL schema to get a superset of allowed values for `pick_list_reference`.
+  # This list includes ALL enums in the HUD schema, so it includes some enums
+  # that don't make sense as pick lists.
+  def allowed_pick_list_references
```
gigxz, line 112: copied from JsonForms

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -10,23 +10,47 @@ def self.perform(...)
...
+
+  def check_json_schema(document)
+    schema_errors = Hmis::Form::Definition.validate_schema(document)
+    schema_errors.each do |e|
+      # Try to figure out which Link ID the error is on
+      item_path = /(\/item\/[0-9]){1,}/.match(e.to_s)&.try(:[], 0)&.split('/')&.
+        compact_blank&.
+        map { |s| Integer(s, exception: false) || s }
+      link_id = document.dig(*item_path, 'link_id') if item_path
+      if link_id
```
gigxz, line 44: copied and slightly amended from JsonForms. This shouldn't happen with the form builder once its stable, but until then its helpful to see the link id with the error

martha (approved):
This change makes sense to me, thanks for the clear commenting Gig!

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -10,23 +10,47 @@ def self.perform(...)
   end
 
   # @param [Hash] document is a form definition document {'item' => [{...}] }
-  # @param [Hash] valid_pick_lists validate that pick list references are in this set
-  def perform(document, valid_pick_lists: [])
-    @issues = []
-    all_ids = check_ids(document, valid_pick_lists.to_set)
+  # @param [role] role of the form as string ('INTAKE', 'CLIENT', etc). If not provided, HUD rule validation will not occur.
+  def perform(document, _role = nil)
```
ttoomey, line 14: suggest taking out role since it's kind of misleading (unless we are implementing it very soon)

ttoomey (approved):
Looks great, thanks!

ttoomey:
>  I moved all the validations into DefinitionValidator, and have it return HmisErrors which can be resolved and can have additional context on them (like severity level, for example).

Makes sense to me. That's a solid improvement, thank you!

drivers/hmis/app/models/hmis/form/definition_validator.rb
ttoomey: Oh I see, we have this in 4475.  Ignore my comment above

PR #4494, opened 2024-06-27
Add `projectMatches` attribute to `FormDefinition`

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6108

This PR adds a `projectMatches` attribute to FormDefinitions, in order to display project applicability info. See [design](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/Form-Builder-Workflow?node-id=863-8373&t=4UXaYginq4JoKNQE-0)

Per our discussion, this returns a match for every project even if it's overridden by a more specific match in another project.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -48,6 +49,24 @@ def definition
...
+      # ...then forms A and B would still return a match for Project X.
+      object.instances.
+        active.
+        map(&:project_matches).
+        flatten.
+        sort_by(&:rank).
+        reverse. # lower rank means a better match
+        map { |match| [match.project.id, match] }. # per project ID, the later added (lower ranked) matches overwrite the earlier ones
+        to_h.values
+    end
```
gigxz, line 98: Nice!

This looks right. As discussed, the fact that `project_matches` checks against all projects is not performant, so that could be further optimized (ended up with an idea below, which doesn't seem too cumbersome).

We should also attempt to reduce N+1s on project when resolving, which we can do with a preload.

It also seems unnecessary that  `Hmis::Hud::Project.all` is queried multiple times. below is a rough suggestion of passing a `scope` that should reduce that:

```
instance_scope = object.instances

# fetching all projects is slow, but at least it happens only once. preload organization because we are going to resolve it
# `.hmis` scope is necessary to exclude unrelated warehouse projects
project_scope = Hmis::Hud::Project.hmis.preload(:organization)

# preload funders only if any of these rules are funder-based (micro-optimizing)
project_scope = project_scope.preload(:funders) if instance_scope.where.not(funder: nil).exists?

# future optimization: further refine project_scope down to only projects that could be matched by any of the instances

# returns `Hmis::Form::InstanceProjectMatch`
object.instances.active.map { |instance| instance.project_matches(scope: project_scope) }.flatten...etc
```

And here's a rough (untested) pass at how we could reduce the size of the project_scope:
```
    matches_any_project = p_t[:id].in(instances.select { |inst| inst.entity_type == 'Hmis::Hud::Project' }.map(&:entity_id))
    matches_org = o_t[:id].in(instances.select { |inst| inst.entity_type == 'Hmis::Hud::Organization' }.map(&:entity_id))
    matches_funders = f_t[:funder].in(instances.map(&:funder))
    matches_project_types = f_t[:project_type].in(instances.map(&:project_type))

    possible_refined_project_scope = if instances.defaults.exists?
      # there is a default so all projects are fair game
      Hmis::Hud::Project.hmis.all
    else
      # glom together all the projects that might match any of the rules
      # this still would need a second pass with the match class to express the full logic (eg "funder AND project type" match) and to get the rank
      Hmis::Hud::Project.hmis.joins(:organization).left_outer_joins(:funders).where(
        matches_any_project.or(matches_funders).or(matches_project_types).or(matches_org),
      )
    end
```

drivers/hmis/app/graphql/types/forms/project_match.rb
```diff
@@ -0,0 +1,34 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class Forms::ProjectMatch < Types::BaseObject
+    # object is a Hmis::Form::InstanceProjectMatch
+    graphql_name 'FormProjectMatch'
+    skip_activity_log
+    description 'Project match for a form, including information about which clients in this project the form is applicable to.'
```
gigxz, line 11:
```suggestion
    description 'Project match for a form, including information about which client groups in this project the form is applicable to.'
```

drivers/hmis/spec/requests/hmis/form_definition_lookup_spec.rb
```diff
@@ -135,6 +135,44 @@
...
+      GRAPHQL
+    end
+
+    let!(:o1) { create :hmis_hud_organization, data_source: ds1 }
+    let!(:p1) { create :hmis_hud_project, data_source: ds1, organization: o1 }
+    let!(:p2) { create :hmis_hud_project, data_source: ds1, organization: o1 }
+
+    let!(:form) { create :hmis_form_definition, identifier: 'form-def' }
+    let!(:all_clients) { create :hmis_form_instance, definition_identifier: 'form-def', entity: o1, active: true, data_collected_about: 'ALL_CLIENTS' }
+    let!(:hoh_and_adults) { create :hmis_form_instance, definition_identifier: 'form-def', entity: p1, active: true, data_collected_about: 'HOH_AND_ADULTS' }
```
gigxz: this factory does accept a definition, so it might be cleaner to just use that. The identifier itself is not really relevant to this test

```
    let!(:form) { create :hmis_form_definition }
    let!(:all_clients) { create :hmis_form_instance, definition: form, entity: o1, active: true, data_collected_about: 'ALL_CLIENTS' }
    let!(:hoh_and_adults) { create :hmis_form_instance, definition: form, entity: p1, active: true, data_collected_about: 'HOH_AND_ADULTS' }
``

File: drivers/hmis/app/graphql/types/forms/form_definition.rb (hunk 1 above)
Comment (martha) [reply drivers/hmis/app/graphql/types/forms/form_definition.rb:98]: thanks @gigxz this is great, I'd like to add more tests but it seems to be working ok against the existing test!

Comment (gigxz) [reply drivers/hmis/app/graphql/types/forms/form_definition.rb:98]: cool! more test cases would be great

File: drivers/hmis/app/graphql/types/forms/form_definition.rb [hunk 4]
```diff
@@ -48,6 +51,54 @@ def definition
...
+      project_scope = project_scope.preload(:funders) if instance_scope.where.not(funder: nil).exists?
+
+      # Further refine project_scope down to only projects that could be matched by any of the instances
+
+      unless instance_scope.defaults.exists?
+        # If there is a default, then all projects are fair game.
+        # Otherwise, glom together all the projects that might match any of the rules.
+
+        matches_any_project = p_t[:id].in(instance_scope.select { |inst| inst.entity_type == 'Hmis::Hud::Project' }.map(&:entity_id))
+        matches_org = o_t[:id].in(instance_scope.select { |inst| inst.entity_type == 'Hmis::Hud::Organization' }.map(&:entity_id))
```
gigxz: wasn't intentional naming on my part, could be `matches_any_org`, `matches_any_funder` etc

drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -48,6 +51,54 @@ def definition
...
+      # - Intake form C has a rule that specifies that it's used for Project X, an ES project
+      # ...then forms A and B would still return a match for Project X.
+
+      instance_scope = object.instances.active
+
+      # Fetching all projects is slow, so do it only once and pass the scope to project_matches below.
+      project_scope = Hmis::Hud::Project.hmis.preload(:organization)
+
+      # Preload funders only if any of these rules are funder-based (micro-optimizing)
+      project_scope = project_scope.preload(:funders) if instance_scope.where.not(funder: nil).exists?
```
gigxz: micro optimizing the micro-optimization: `instance_scope.find { |inst| inst.funder.present? }`

We hit the database once to get all the instances (line 63). We expect the list of instances to be relatively small, so we can operate on it in-memory. (As opposed to hitting the database again to ask if any of them have a funder, which is what the `exists?` query will do).

There are plenty of places below where we choose whether to hit the db or use memory:
`instance_scope.defaults.exists?` <= db
`instance_scope.map(&:project_type)` <= memory (db would be to use `pluck`)
`instance_scope.select ... ` <= memory (db would be `where`)


I hope I didn't over-explain that... just taking the opportunity to point out something it can be helpful to think about when writing code in the graphql resolvers. None of these examples here seem that important to me -- but it can be problematic if we are really hammering the API for something that we could easily do in-memory, for example.

drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -48,6 +51,54 @@ def definition
...
+      # For example, if:
+      # - Intake form A has a rule that specifies that it's used for all projects
+      # - Intake form B has a rule that specifies that it's used for all Emergency Shelter projects
+      # - Intake form C has a rule that specifies that it's used for Project X, an ES project
+      # ...then forms A and B would still return a match for Project X.
+
+      instance_scope = object.instances.active
+
+      # Fetching all projects is slow, so do it only once and pass the scope to project_matches below.
+      project_scope = Hmis::Hud::Project.hmis.preload(:organization)
```
gigxz: the `preload`s should probably be moved to the end once project_scope has been refined

drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -33,6 +35,7 @@ class Forms::FormDefinition < Types::BaseObject
     field :date_updated, GraphQL::Types::ISO8601DateTime, null: false, method: :updated_at
     field :date_created, GraphQL::Types::ISO8601DateTime, null: false, method: :created_at
     field :updated_by, Types::Application::User, null: true
+    field :project_matches, [Types::Forms::ProjectMatch], null: false
```
gigxz: There can be hundreds of matching projects, so we should employ pagination https://github.com/greenriver/hmis-frontend/pull/826/files#r1657702742

drivers/hmis/app/graphql/types/forms/form_definition.rb
martha: Yes this makes sense, noo worries about over-explaining, the level of explanation is perfect! ☺️ 

I've made the update you suggested here. I'm assuming that it's not worth it to replace the `instance_scope.defaults.exists?` db call with an equivalent in-memory matcher. But would be happy to do so if you disagree.

drivers/hmis/app/graphql/types/paginated_array.rb
```diff
@@ -0,0 +1,13 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+#
+module Types
+  class PaginatedArray < Types::PaginatedScope
```
gigxz, line 8: Nice solution!

gigxz (approved):
Great!

PR #4479, opened 2024-06-23
6215 enhance hmis form definition validation

ttoomey:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6215)

* update json schemer (validation gem)
* improve json schema to better describe our form definition
* extract form definition out to another class since it was getting kind of long
* check validity of link id references (bounds, enable_when, etc)
* check link id format (obviates [4424](https://github.com/greenriver/hmis-warehouse/pull/4424))
* fixes validation issues that have crept into our json form docs

Note: original issue described need to check for cyclic deps. Looks like we currently do have those deps but I guess they are supported by our front end. So I'm leaving that part off.

## Type of change
[//]: # 'remove options that are not relevant'
- [x] New feature (adds functionality)
- [x] Dependency update

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4498, opened 2024-06-28
stricter-hmis-form-json-schema

ttoomey:
reverts "5e178173494a217c09692f8835e46b1fc05d5609"

## Description
[GH issue](https://github.com/open-path/Green-River/issues/6215)

Depends on:
* fixes to front-end editor, currently the editor saves invalid properties

Previous PR: https://github.com/greenriver/hmis-warehouse/pull/4479
Frontend PR for enforcing shape when building form: https://github.com/greenriver/hmis-frontend/pull/825

## Type of change


drivers/hmis_external_apis/public/schemas/form_definition.json
```diff
@@ -94,107 +231,596 @@
...
+        },
+        "answer_number": {
+          "type": "integer"
+        },
+        "answer_boolean": {
+          "type": "boolean"
+        },
+        "answer_group_code": {
+          "type": "string"
+        }
```
gigxz, line 448: @ttoomey do you know how to make it so exactly one of these is required? 
I tried the below, but it still lets me input values for multiple. 
```
        "oneOf": [
          { "required": [ "answer_code" ] },
          { "required": [ "answer_codes" ] },
          { "required": [ "answer_number" ] },
          { "required": [ "answer_boolean" ] },
          { "required": [ "answer_group_code" ] },
          { "required": [ "compare_question" ] }
        ],
```

We should also have exactly one of `question` and `local_constant`

drivers/hmis_external_apis/public/schemas/form_definition.json
```diff
@@ -94,107 +231,596 @@
...
       }
     },
-    "autofill_values": {
+    "autofillValues": {
```
gigxz, line 451: This needs one-of logic as annotated here https://github.com/greenriver/hmis-warehouse/blob/087bbb6f3398bc90430bc7ea16562140170e5222/drivers/hmis/app/graphql/types/forms/autofill_value.rb#L14-L20

drivers/hmis_external_apis/public/schemas/form_definition.json
gigxz: I'm just going to try adding these validations in the code.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -20,6 +20,8 @@ def perform(document, role = nil)
     all_ids = check_ids(document)
     # Check references
     check_references(document, all_ids)
+    # Check mutually exclusive attributes ("one of" on conditional objects)
+    check_mutually_exclusive_attributes(document)
```
gigxz, line 24: @ttoomey I'm not able to request your review, but can you take a look at the changes I made to this file? I wasn't able to express it in JSON schema.

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -59,7 +59,8 @@ def recursively_transform(form_element)
       converted = form_element.
         excluding('__typename', ''). # drop typescript artifacts and empty keys
         compact. # drop keys with nil values
-        delete_if { |_key, value| value.is_a?(Array) && value.empty? }. # drop empty arrays
+        # drop empty arrays and empty strings
+        delete_if { |_key, value| (value.is_a?(Array) || value.is_a?(String)) && value.empty? }.
```
gigxz, line 63: Ensure empty strings are dropped from form on update

drivers/hmis_external_apis/public/schemas/form_definition.json
```diff
@@ -203,84 +829,269 @@
...
         },
-        "type": {
+        {
+          "$ref": "#/$defs/dataCollectedAbout"
+        },
+        {
+          "type": "object",
+          "$comment": "child item array should be required, but cant because when you add a new group to the form builder its empty and we validate on save. maybe add an exception?",
+          "required": [
+          ],
```
gigxz, line 848: Note to self to update/remove this comment. @ttoomey it's correct that groups _should_ have children, but the Form Builder needs to maintain the ability to add a group, save, and then add children to it.

The form renderer is capable of rendering childless groups, so I think it's OK to keep them un-required.

ttoomey: Fair, I think that's fine.

But I wonder if we even need to validate the schema of a draft. If we didn't validate, that would allow us to be more strict with the validation of published forms while still supporting intermediate states needed during editing. Trying to serve both the draft and the published definitions with the same schema might reduce the value of the schema.

If it was useful to have some validation on the drafts, we could build a simple and  permissive 'draft' schema to guard against basic structural errors. That would be pretty easy to create if we left the object properties open and didn't enumerate the properties beyond say 'item', 'link_id', and 'type'

drivers/hmis/app/models/hmis/form/definition_validator.rb
ttoomey: I think you can express this in JSON schema with something like the diff below. However the validation messages may not be as nice, you'd get something to like "Schema error on item 'foo': property '/item/1/bounds/0' is invalid: error type=one of is invalid" 

If that seems okay, I could take a crack at representing the checks in `check_mutually_exclusive_attributes` in the schema, it seems like it wouldn't be too hard. Or we could just stick with the validation function you have implemented here.


```diff
@@ -101,9 +101,14 @@
     "bounds": {
       "type": "object",
       "additionalProperties": false,
-      "required": [
-        "id",
-        "type"
+      "allOf": [
+        { "required": [ "id", "type" ]},
+        { "oneOf": [
+          { "required": [ "value_number" ] },
+          { "required": [ "value_date" ] },
+          { "required": [ "value_local_constant" ] },
+          { "required": [ "question" ] }
+        ]}
       ],
       "properties": {
         "_comment": {
@@ -127,6 +132,9 @@
         "value_number": {
           "type": "integer"
         },
+        "value_date": {
+          "type": "string"
+        },
         "offset": {
           "type": "integer"
         },
```

drivers/hmis_external_apis/public/schemas/form_definition.json
gigxz: The point of this is to ensure that our frontend form builder is only producing valid shapes. So I definitely think we should validate the schema when you save changes to an item (on a draft). If we don't, it's possible our form builder is allowing you to create invalid forms, and then when you go to publish you're faced with a bunch of (possibly un-fixable) errors and need to track them down. It will also make QA a lot harder if we can't surface issues until publish. So I would like to keep validation on save.

Requiring children in groups is the only thing I can think of to omit from the draft stage. As mentioned I don't feel like it's important enough to warrant building a specific workaround check for it. (Basically, nothing 'bad' happens from having an empty group)

gigxz:
Oops I didn't mean to push https://github.com/greenriver/hmis-warehouse/pull/4498/commits/92bad41a8d759c8ce9e160765d9ecc9edf94c717, I'll pull that back

drivers/hmis/app/models/hmis/form/definition_validator.rb
gigxz: Ah okay, cool.

Well, maybe we should just leave it in the code because we could add more advanced checks later that relate to the dependent item. For example if you are saying "enable when question X is > 10", we could validate that question X is in fact numeric. _That_ certainly isn't expressible in JSON Schema, I don't think.

gigxz (approved):

gigxz:
https://github.com/greenriver/hmis-warehouse/pull/4498/commits/b5b6e6163918f482cb39987b888c20b72f128f56 added:
* Only allow `bounds` on numeric and date fields
* Only allow `assessment_date` on date fields
* Add back `custom_rule`

gigxz:
Can I go ahead and merge this @ttoomey?

drivers/hmis/app/models/hmis/form/definition_validator.rb
ttoomey: Sounds good, seems fine to stick with code for now. Probably is possible to express it in json schema but it could get kind of ugly and hard to maintain.

ttoomey:
> Can I go ahead and merge this @ttoomey?

Looks good! I merged it in

PR #4499, opened 2024-06-30
HMIS notes on custom services in wharehouse

eanders:

## Description

This:
1. Exposes details about custom services in the warehouse
2. Standardizes the order of some columns on the enrollment details tab
3. Adds a link to the HMIS from the data source list page (and hides visible in window since that's not really relevant to anyone anymore, especially if they have an HMIS installation.)

## Type of change

### DIFF
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -36,7 +36,7 @@ module GrdaWarehouse::Hud
     self.table_name = :Client
     self.sequence_name = "public.\"#{table_name}_id_seq\""
 
-    CACHE_EXPIRY = if Rails.env.production? then 4.hours else 30.seconds end
+    CACHE_EXPIRY = if Rails.env.production? then 1.hours else 30.seconds end
 
     has_many :client_files
     has_many :health_files
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -1717,7 +1717,12 @@ module GrdaWarehouse::Hud
 
     def services_for_rollup
       custom_services.
-        preload(:warehouse_project, enrollment: [:project, :client], custom_service_type: [:custom_service_category]).
+        preload(
+          :warehouse_project,
+          enrollment: [:project, :client],
+          custom_service_type: [:custom_service_category],
+          form_processor: :definition,
+        ).
         order(date_provided: :desc).
         order(id: :desc)
     end
--- a/app/views/data_sources/index.haml
+++ b/app/views/data_sources/index.haml
@@ -23,7 +23,11 @@
         %th Short Name
         %th Client Records
         %th Project Count
-        %th Visible in Window
+        %th
+          - if hmis_enabled?
+            OP HMIS
+          - else
+            Visible in Window
         %th Authoritative?
         %th Last Import Run
         %th
@@ -43,7 +47,13 @@
             .ds{class: "ds-color-#{data_source.id}"}= data_source.short_name
           %td= number_with_delimiter(@client_counts[data_source.id])
           %td= number_with_delimiter(@project_counts[data_source.id])
-          %td.text-center= checkmark(data_source.visible_in_window)
+          %td.text-center
+            - if hmis_enabled?
+              - if data_source.hmis?
+                = link_to "https://#{data_source.hmis}" do
+                  %i.icon-link-ext
+            - else
+              = checkmark(data_source.visible_in_window)
           %td.text-center= checkmark(data_source.authoritative)
           %td
             = most_recently_completed
--- a/app/views/data_sources/index.haml
+++ b/app/views/data_sources/index.haml
@@ -23,11 +23,11 @@
         %th Short Name
         %th Client Records
         %th Project Count
+        - if hmis_enabled?
+          %th
+            HMIS
         %th
-          - if hmis_enabled?
-            OP HMIS
-          - else
-            Visible in Window
+          Visible in Window
         %th Authoritative?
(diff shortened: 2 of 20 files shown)


drivers/hmis/app/views/hmis_client/services/show.haml
```diff
@@ -0,0 +1,34 @@
...
+      - values_arr = cdes.map(&:value)
+      - display_value = values_arr.join(', ')
+      - display_value = yes_no(values_arr.first) if [true, false].include?(values_arr.first) && values_arr.size == 1
+      %dl
+        %dt= cded.label # label of the CustomDataElementDefinition
+        %dd= display_value
+  .none-found
+    %a{href: @service.data_source.hmis_url_for(@service), target: :blank}
+      Open in HMIS
+      %i.icon-link-ext.ml-2{style: 'color: inherit; font-size: inherit;'}
```
gigxz: On the client dashboard, the link uses class `.text-teeny` to match the font size of the "created"/"updated" text it sits next to. Can we make that consistent (either way)? I don't know if there's an easy way in rails to pull this into a "component" that can be shared. Maybe we could make a variation of the `hmis_url_for` function that returns an HTML snippet?

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1717,7 +1717,12 @@ def date_of_last_homeless_service
 
     def services_for_rollup
       custom_services.
-        preload(:warehouse_project, enrollment: [:project, :client], custom_service_type: [:custom_service_category]).
+        preload(
```
gigxz, line 1720: I'm not seeing services show up when I add new ones, is there a cache I need to clear for the page somehow?

app/views/data_sources/index.haml
```diff
@@ -43,7 +47,13 @@
...
           %td= number_with_delimiter(@client_counts[data_source.id])
           %td= number_with_delimiter(@project_counts[data_source.id])
-          %td.text-center= checkmark(data_source.visible_in_window)
+          %td.text-center
+            - if hmis_enabled?
+              - if data_source.hmis?
+                = link_to "https://#{data_source.hmis}" do
+                  %i.icon-link-ext
+            - else
+              = checkmark(data_source.visible_in_window)
```
gigxz: Would the OP HMIS never be configured as "visible in the window?

app/views/data_sources/index.haml
```diff
@@ -23,7 +23,11 @@
         %th Short Name
         %th Client Records
         %th Project Count
-        %th Visible in Window
+        %th
+          - if hmis_enabled?
+            OP HMIS
```
gigxz: Is this the language we are using to refer to the HMIS from within the Warehouse? In other places we just say "HMIS." I was under the impression that "OP" was currently meaningless to many users, so, maybe we should consider which terminology we want to use.

drivers/hmis/app/views/hmis_client/services/show.haml
```diff
@@ -0,0 +1,34 @@
...
+      %h3 Service Details
+      %dl
+        %dt Project
+        %dd= @service.enrollment.project.warehouse_project.name(current_user)
+        %dt Date Provided
+        %dd= @service.DateProvided
+        %dt Created
+        %dd= @service.DateCreated
+        %dt Updated
+        %dd= @service.DateUpdated
```
gigxz, line 28: We should include, if present:
`@service.FAAmount`
`@service.FAStartDate`
`@service.FAEndDate`

We may want to use this modal for HUD Services, too.. as far as I can tell there is no way currently to see the FA fields on a HUD Service?

The HMIS can also support recording custom fields on HUD services. So it seems reasonable we could use this detail modal for both record types.

gigxz (commented):
Looks great to me after adding the rest of the HUD Fields (3 `FA` fields)! Thanks!

drivers/hmis/app/views/hmis_client/services/show.haml
eanders: It might be slightly painful, but you can see these values for HUD services under the little HMIS button.

Screenshot 2024-07-01 at 11 42 22 AM

drivers/hmis/app/views/hmis_client/services/show.haml
eanders: We could make a helper method.  I actually think that's incorrect in this case since this is a BIG button/link, but I updated the table to use the appropriate class.  Thanks!

drivers/hmis/app/views/hmis_client/services/show.haml
gigxz: Ah okay

drivers/hmis/app/views/hmis_client/services/show.haml
gigxz: Oh my comment was meant for the other template, then!

app/views/data_sources/index.haml
eanders: So, once everyone is on Access Controls, there should be no reason to mark a data source as "Visible in the Window."  I went with OP HMIS because it seems odd to have a column called HMIS where 90% of those data sources represent an HMIS.  

Happy to entertain other wording.  Perhaps it's too early to remove the "window" column.  I'll add that back for now.  I guess if we're only showing the HMIS column to installations that have at least one "OP" HMIS, then it's probably fine to just call it HMIS.

eanders:
@gigxz I tried to impose a bit of order, added a helper method for linking out to HMIS and tried to use it in most places.  Standardized the client roll-up caching length...

gigxz (approved):
great!

PR #4502, opened 2024-07-02
6000 prune csv staging and source tables tt

ttoomey:
* advisory locks on each job
* pivot from running each job on a ds to running each job on a table
* jobs re-enqueue themselves to process the next table in series
* include 2020 an 2022 in processing
* seems like batches aren't going to be straight forward. Take the out since I don't think they were buggy. We can try running the update in a single tx and see if it works
* fixup logic so tests pass again- change logic so that retained count includes items that are retained due to recentness


app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -171,8 +171,8 @@ def perform
         # Cleanup old import data
         # Disabled until we're confident on timing and performance
         # GrdaWarehouse::DataSource.with_deleted.pluck(:id).each do |ds_id|
-        #   HmisCsvImporter::Cleanup::ExpireLoadersJob.perform_later(data_source_id: ds_id)
-        #   HmisCsvImporter::Cleanup::ExpireImportersJob.perform_later(data_source_id: ds_id)
+        #   HmisCsvImporter::Cleanup::ExpireLoadersJob.perform_later
+        #   HmisCsvImporter::Cleanup::ExpireImportersJob.perform_later
         # end
```
eanders: If we aren't running per-data source, we should remove the DS loop.

eanders (approved):

eanders:
@ttoomey this looks great, thanks! I was going to merge, but it looks like you still had it marked as draft, so holding off.  Feel free to merge if you think it's ready.

PR #4506, opened 2024-07-03
Expire imported data w/tmp table

eanders:

## Description

## Type of change
- [ ] Bug fix
- [ ] Dependency update


config/schedule.rb
```diff
@@ -25,16 +25,15 @@
...
 daily_schedule = ENV['DAILY_SCHEDULE'] || '3:10 am'
 import_schedule = ENV['IMPORT_SCHEDULE'] || '5:30 pm'
-glacier_import_schedule = ENV['IMPORT_SCHEDULE'] || '5:30 pm'
+
 export_schedule = if ENV['DAILY_EXPORT_SCHEDULE'].nil? || ENV['DAILY_EXPORT_SCHEDULE'].empty? then (Time.parse(daily_schedule) - 2.hours).strftime('%I:%M %P') else ENV['DAILY_EXPORT_SCHEDULE'] end
 file_cleaning_schedule = (Time.parse(daily_schedule) - 5.minutes).strftime('%I:%M %P')
 import_prefetch_schedule = (Time.parse(import_schedule) - 4.hours).strftime('%I:%M %P')
 census_schedule = (Time.parse(import_schedule) - 5.hours).strftime('%I:%M %P')
-# database_backup_time = Time.parse(import_schedule) - 3.hours
+import_cleanup_time = Time.parse(import_schedule) + 9.hours
```
ttoomey: this is isn't used?

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -81,11 +87,64 @@ def process_model(model)
...
+          SELECT id, #{log_id_field}, row_number() OVER (
+            PARTITION BY "#{key_field}", data_source_id ORDER BY id DESC
+          ) AS row_num
+          FROM #{model.quoted_table_name}
+        ) subquery
+        WHERE subquery.row_num > #{@retain_item_count}
+      SQL
+    end
+
     private def mark_expired_query(model)
```
ttoomey: this is unused now, remove?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/temp_table.rb
```diff
@@ -0,0 +1,44 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisCsvImporter
+  class TempTable
+    # temporary table functionality from https://github.com/zdennis/ar-extensions/blob/master/ar-extensions/lib/ar-extensions/temporary_table.rb
```
ttoomey: This seems to work but the referenced code is from 15 years ago, I think in modern rails this can be much simpler. I'm not sure that we even need a class for this. And I don't think there's a need to wrap it in an AR model

```ruby
GrdaWarehouseBase.connection.create_table table_name, temporary: true do |t|
  t.references : source, null: false
  t.references : source, null: false
end
```

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -81,11 +87,64 @@ def process_model(model)
...
+    private def expire_data(model)
+      tmp_table_prefix = model.table_name.downcase
+      tmp_table_name = "#{tmp_table_prefix}_tmp_exp"
+      tmp_model_name = "#{model.class.name.demodulize}Expire"
+      tmp_class = HmisCsvImporter::TempTable.create_temporary_table(table_name: tmp_table_name, model_name: tmp_model_name)
+      begin
+        populate_tmp_table(model, tmp_table_name)
+        write_expirations(model, tmp_table_name)
+      ensure
+        tmp_class.drop
```
ttoomey: temporary tables are automatically dropped at the end of the session. Probably don't need this?

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -48,11 +48,17 @@ def perform(model_name: nil, retain_item_count: 5, retain_after_date: DateTime.c
...
         model_name: next_model.name,
         retain_item_count: retain_item_count,
         retain_after_date: retain_after_date,
-      )
+        delay: delay,
+      }
+      if delay
+        self.class.perform_later(**options)
+      else
+        self.class.perform_now(**options)
```
ttoomey: Assume this is for development? Not sure that supporting the delay option is worth it just for dev. Also performing the job inline means you'd get a stack of these jobs (like ~50) and could get weird interactions cause they are all pending cleanup. Suggest just firing up a local worker instead for development

ttoomey (approved):
Thank @eanders - looks good! I left some suggestions

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
ttoomey: oh I see, this is because we aren't running in dj anymore

eanders: Yeah, this isn't necessary, but seemed like a reasonable way to let us play around with DJ vs cron without having to push a code change.

ttoomey:
@eanders going to push some changes up and merge this into 6000

eanders:
@ttoomey thanks!  I didn't realize you could create the temp tables with standard active record, that's great!  In the past we needed a class to back it, but I don't think we do here, so your solution looks great if it works.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -52,13 +52,8 @@ def perform(model_name: nil, retain_item_count: 5, retain_after_date: DateTime.c
...
         retain_item_count: retain_item_count,
         retain_after_date: retain_after_date,
-        delay: delay,
       }
-      if delay
-        self.class.perform_later(**options)
-      else
-        self.class.perform_now(**options)
-      end
+      self.class.perform_later(**options)
```
eanders: This needs to be `perform_now` or else it will end up being run as a DJ.   (Note the default was `delay: false`.)  I actually think we should keep this around until we're happy with the behavior in production.

ttoomey:
changes made:
* I don't think that adding an index to the tmp table helps performance. Testing locally, the time needed to build the index offsets the reduction in query time, there's not much difference in runtime. This makes sense to me since you'd need to do a table scan either way. Maybe this is different on very large tables the index is helpful there?
* if we aren't indexing the batch_id, there's no reason to include it in the temp table. Once these tables are pruned, I think adding an index to the import/log ids on all loader/importer tables might be worth doing
* the delayed param was set to false, so this was being invoked as one big job that calls itself recursively. If we want this to run synchronously for the task manager, I think we should structure it so that the jobs run discretely in series (model.each { |model| MyJob.perform_now(model) }
* tidy up tmp table management code

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -127,8 +124,8 @@ def process_model(model)
 
     private def populate_tmp_table_query(model, tmp_table_name)
       <<~SQL
-        INSERT INTO #{tmp_table_name} (source_id, batch_id)
-        SELECT id, #{log_id_field} from (#{relevant_id_query(model)}) as relevant_ids
+        INSERT INTO #{tmp_table_name} (source_id)
+        SELECT id FROM (#{relevant_id_query(model)}) as relevant_ids
```
eanders, line 109: Pretty sure we still need the `log_id_field` for performance until we've added the index to the non-temporary table.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -88,36 +83,38 @@ def process_model(model)
...
     end
 
+    # this could be batched if needed
     private def expiration_update_query(model, tmp_table_name)
       <<~SQL
         UPDATE #{model.quoted_table_name} source_table
         SET expired = true
         FROM #{tmp_table_name} tmp_table
         WHERE tmp_table.source_id = source_table.id
-        AND tmp_table.batch_id < #{min_age_protected_id}
```
eanders: Without this, we need to make sure this filter gets applied on insert/select.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
eanders: The idea is that this _should_ be called in sequence within the one cron task.  Cron tasks can run indefinitely (I believe) but delayed jobs have a max runtime.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
ttoomey: ah oops, dropped that by mistake. Thanks!

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
ttoomey: That's fine but I think having the job call itself recursively for each model like this is not ideal. I'll adjust this to run over all models discretely in series

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
ttoomey: where you able to see a difference in the total run time from adding the index? Benchmarking locally I didn't see much difference (1 mil rows)

eanders: I haven't had a chance to benchmark. It definitely helped on the non-temp tables, but with lack of iops balance, nothing is fast in QA right now. Happy to try without, and add to the import/loader tables after the first round of deletes.

PR #4509, opened 2024-07-03
EKS Prep

yakloinsteak:
## Description

These are changes that should:
1. Not break our ECS deployments
2. Make things work or work better on kubernetes

## Type of change

### DIFF
--- a/lib/rds_sql_server/rds.rb
+++ b/lib/rds_sql_server/rds.rb
@@ -8,8 +8,6 @@ class Rds
 
   REGION             ||= 'us-east-1'.freeze
   AVAILABILITY_ZONE  ||= 'us-east-1a'.freeze
-  ACCESS_KEY_ID      ||= ENV.fetch('RDS_AWS_ACCESS_KEY_ID')
-  SECRET_ACCESS_KEY  ||= ENV.fetch('RDS_AWS_SECRET_ACCESS_KEY')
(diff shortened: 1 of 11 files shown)


lib/rds_sql_server/rds.rb
```diff
@@ -37,19 +35,7 @@ def self.timeout
...
-                                           region: REGION,
-                                           access_key_id: ACCESS_KEY_ID,
-                                           secret_access_key: SECRET_ACCESS_KEY,
-                                         })
-    else
-      self.client = Aws::RDS::Client.new({
-                                           region: REGION,
-                                         })
-    end
+    self.client = Aws::RDS::Client.new
```
yakloinsteak, line 38: We should abandon long-lived keys, and spot checking a few roles in ECS, it looks like these keys aren't serving a purpose. Let me know if there's some reason to keep the ability to inject different static creds into this code or if I'm mistaken.

eanders: Hmm, we might still need the region (see line 50).  I'm pretty sure we fixed up the roles, but we should definitely test that.

yakloinsteak:
@eanders Could I deploy this to QA on ECS? Is release-121 okay? I'll need help with the LSA/RDS QA, though.

yakloinsteak:
All seems well in the staging environment.

PR #4511, opened 2024-07-03
Add CDED validation to form publish

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6316

- Adds a new flag `is_publishing` to the `validate_json_form` method.
- Additionally adds the `owner_type` as an argument to the `Hmis::Form::DefinitionValidator.perform` method. This is based on the `role` which `perform` already receives, but passing it in from the `FormDefinition` allows us to avoid repeating the logic to generate it (from `FormDefinition.owner_class`)
- Adds logic for validating CDEDs to the `DefinitionValidator`
  - CDED for the item should exist, by `key` and `owner_type`
  - CDED should have a compatible type with the item type
- Adds spec tests

How I've tested:
- On the release branch, create a form item, save and publish (this generates the CDED)
- Change to this branch, and in the db, modify the CDED:
  - by changing its type
  - by changing its owner_type
  - finally by deleting it
- Then return to the form and try to save (any random change) and publish it again. Each case should result in the appropriate error. These are runtime errors because they aren't fixable by the user; they would indicate a bug in our form builder logic or a weird database state.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -29,7 +29,7 @@ def resolve(id:)
       validation_errors = definition.validate_json_form
       return { errors: validation_errors } if validation_errors.any?
 
-      validation_errors = definition.validate_json_form
+      validation_errors = definition.validate_json_form(is_publishing: true)
```
martha: Just to make sure I have my head correctly wrapped around this:

The validator now will expect CDEDs to already exist for any form item that has a mapping with a `custom_field_key`. That could've thrown an error in the past, back when we used to add `custom_field_key` to the form and then generate the CDED (at first by hand, later using the rake task, but either way it was a two-step operation).

Now, though, as long as all forms currently published have their CDEDs generated already*, we should be OK, because on publish, the form builder will 
1) validate that all existing items that have mapping fields correspond to CDEDs that match, and then 
2) create new CDEDs for any new fields that need them, and add that mapping field to the json definition.

In other words, from now on, an item should never exist in an in-between state where it has a `custom_field_key` mapping, but no CDED.

Is there anything I'm missing?
*is this an okay assumption?

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -29,7 +29,7 @@ def resolve(id:)
       validation_errors = definition.validate_json_form
       return { errors: validation_errors } if validation_errors.any?
```
gigxz: Whoops, this is a mistaken merge resolution, this call should not be duplicated

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -11,7 +11,9 @@ def self.perform(...)
 
   # @param [Hash] document is a form definition document {'item' => [{...}] }
   # @param [role] role of the form as string ('INTAKE', 'CLIENT', etc). If not provided, HUD rule validation will not occur.
-  def perform(document, role = nil)
+  # @param [owner_type] owner_type of the form, such as Hmis::Hud::CustomAssessment
+  # @param [boolean] is_publishing if true, validate CDEDs
+  def perform(document, role = nil, owner_type = nil, is_publishing = false)
```
gigxz: And can we adjust [this code](https://github.com/greenriver/hmis-warehouse/blob/26ef8c27b51dba80b065dfa9a3034722f165a15e/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L243) as well to skip the cded check? This runs from the JSON Editor so we don't want that to fail when adding a new field to a draft.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -11,7 +11,9 @@ def self.perform(...)
 
   # @param [Hash] document is a form definition document {'item' => [{...}] }
   # @param [role] role of the form as string ('INTAKE', 'CLIENT', etc). If not provided, HUD rule validation will not occur.
-  def perform(document, role = nil)
+  # @param [owner_type] owner_type of the form, such as Hmis::Hud::CustomAssessment
+  # @param [boolean] is_publishing if true, validate CDEDs
```
gigxz: General note: for readability, I would suggesting using an argument name that is descriptive of how that argument modifies behavior, as opposed to a name that describes when it should be passed (or something about the caller). (Ex "is_publishing" describes when to pass the argument, but it doesn't provide any hints about what the argument actually does. As opposed to "skip_cded_validation" or "validate_cded_mappings" which describe how the argument will modify the validation routine).

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
gigxz: Yeah that's correct. There is only one situation when we we expect the CDEDs don't exist, and that's when the form is in `draft` status.

IMO we should move the validation step below the CDED creation step, though, which means moving the call to `validate_json_form` down into the transaction where we have a `TODO`. That way we don't get into a situation where the `add_missing_custom_field_keys` code generates faulty CDEDs in a way that actually fails validation. (Currently we would never see that, since we validate before creating the new CDEDs).

I also think we should probably invert the boolean, i.e. make it so the  _default_ behavior when running the DefinitionValidator should be to validate everything, but include a toggle to skip certain validation.  That way we will still run the full validation [on seed](https://github.com/greenriver/hmis-warehouse/blob/26ef8c27b51dba80b065dfa9a3034722f165a15e/drivers/hmis/lib/hmis_util/json_forms.rb#L264) for example, which we want.

One option is to push this logic to the `definition.validate_json_form` logic itself:

```
  def validate_json_form
    # Skip validation of CustomDataElementDefinitions on draft form, because new CDEDs won't be created yet
    Hmis::Form::DefinitionValidator.perform(definition, role, skip_cded_validation: draft?)
  end
```
or, we could keep it exposed for the caller to specify:
```
  def validate_json_form(skip_cded_validation: false)
    Hmis::Form::DefinitionValidator.perform(definition, role, skip_cded_validation: skip_cded_validation)
  end

...

// in UpdateFormDefinition
# Skip validation of CustomDataElementDefinitions because this is a draft form, so new CDEDs won't be created yet
validation_errors = definition.validate_json_form(skip_cded_validation: true)
```

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -11,7 +11,9 @@ def self.perform(...)
 
   # @param [Hash] document is a form definition document {'item' => [{...}] }
   # @param [role] role of the form as string ('INTAKE', 'CLIENT', etc). If not provided, HUD rule validation will not occur.
-  def perform(document, role = nil)
+  # @param [owner_type] owner_type of the form, such as Hmis::Hud::CustomAssessment
```
gigxz: Owner can be inferred from the role, maybe we can do that instead of passing it in?

One way we do this sometimes is to define a method as both a class method and instance method:
```
// defintion.rb
  def self.owner_class_for_role(role)
    return Hmis::Hud::CustomAssessment if ASSESSMENT_FORM_ROLES.include?(role.to_sym)

    return unless FORM_ROLE_CONFIG[role.to_sym].present?

    FORM_ROLE_CONFIG[role.to_sym][:owner_class].constantize
  end

  def owner_class
    self.class.owner_class_for_role(role)
  end


//definition_validator.rb

owner_type = Hmis::Form::Definition.owner_class_for_role(role).sti_name
```
(untested code)

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -146,4 +150,52 @@ def allowed_pick_list_references
...
+  def check_cdeds(document, owner_type)
+    cded_check = lambda do |item|
+      (item['item'] || []).each do |child_item|
+        cded_check.call(child_item)
+
+        mapping = child_item['mapping']
+        next unless mapping&.key?('custom_field_key')
+
+        cded_key = mapping['custom_field_key']
+        cded = Hmis::Hud::CustomDataElementDefinition.where(key: cded_key, owner_type: owner_type).first
```
gigxz: Possible performance improvement: do CDED lookups once, instead of in each iteration.
```
cdeds_by_key = Hmis::Hud::CustomDataElementDefinition.where(owner_type: owner_type).index_by(&:key)

or

cded_key_type_hash = Hmis::Hud::CustomDataElementDefinition.where(owner_type: owner_type).pluck(:key, :field_type).to_h
```

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
martha: ah sorry- thank you!

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
martha: All this makes sense, but I'm having a bunch of errors locally when I introduce the validation on seed. Probably my test DB is just out of date, so I wonder if this could be easily fixed by just running the rake task to generate cdeds? But the rake task to generate cdeds also has errors for me locally now. Maybe we could pair on this.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -591,7 +596,8 @@ def self.infer_cded_field_type(item_type)
     when 'CURRENCY'
       'float'
     else
-      raise "unable to determine cded type for #{item_type}"
+      'string'
```
martha: Sorry this is an artifact from debugging, I'll remove it after getting to the bottom of the issues with the cded generating rake task.

martha: (After this, the next error is
```
rake aborted!
ActiveRecord::NotNullViolation: PG::NotNullViolation: ERROR:  null value in column "key" of relation "CustomDataElementDefinitions" violates not-null constraint
DETAIL:  Failing row contains (714, Hmis::Hud::Client, null, boolean, null, See DOB, f, 3, 3800b2c8fe534842a7a38740d5d3aaf5, 2024-07-08 14:28:32.607318, 2024-07-08 14:28:32.607318, null, f, null).
```
)

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
gigxz: Yep happy to look!

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -146,4 +149,54 @@ def allowed_pick_list_references
       enums.to_set
     end
   end
+
+  def check_cdeds(document, role)
+    owner_type = role ? Hmis::Form::Definition.owner_class_for_role(role)&.sti_name : nil
```
gigxz: if owner_type is `nil`, we can't really check anything.  We probably shouldn't call this function if we don't have a role.

gigxz (approved):

PR #4512, opened 2024-07-03
ensure migrations are runnable

ttoomey:

## Description
 
Per slack Convo on 6/1 with Sandy, migrations should be runnable 
> Migrations are not running for me because MigrateHmisFormData references OneTimeMigration20230303 via migrate_hmis_link_ids which uses validate_definition which it notes ‘no longer exists’ — I’m going to just comment the line out for my dev, but, is that the correct solution in general so that our codebase is installable?
## Type of change
- [x] Bug fix


gigxz (approved):
whoops, thanks

PR #4523, opened 2024-07-09
Bug fix: dont delete instances when deleting a form draft

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6331

Adjust `dependent` behavior on relations to Definition.

This fixes a bug where calling `destroy!` on a definition would delete all the Form Instances that apply to it, even if there are other forms with the same `identifier` that are still present and governed by those instances.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/mutations/delete_form_definition.rb
+++ b/drivers/hmis/app/graphql/mutations/delete_form_definition.rb
@@ -17,7 +17,16 @@ module Mutations
       raise 'not found' unless definition
       raise 'can only delete draft forms' unless definition.draft?
 
-      definition.destroy!
+      has_other_versions = definition.all_versions.count > 1
+
+      Hmis::Form::Definition.transaction do
+        # If this is the last remaining version of this form, remove all Form Instances.
+        definition.instances.each(&:destroy!) unless has_other_versions
+
+        # Soft-delete the Form Definition. This will error if there are any form_processors or external_form_submissions
+        # that reference this particular form, which there shouldn't be because it's a Draft.
+        definition.destroy!
+      end
 
       {
         form_definition: definition,
--- a/drivers/hmis/app/models/hmis/form/definition.rb
+++ b/drivers/hmis/app/models/hmis/form/definition.rb
@@ -40,13 +40,15 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   # convenience attr for passing graphql args
   attr_accessor :filter_context
 
-  has_many :instances, foreign_key: :definition_identifier, primary_key: :identifier, dependent: :restrict_with_exception
+  # --- Relations by id ----
   has_many :form_processors, dependent: :restrict_with_exception
-  has_many :custom_service_types, through: :instances, foreign_key: :identifier, primary_key: :form_definition_identifier
   has_many :external_form_submissions, class_name: 'HmisExternalApis::ExternalForms::FormSubmission', dependent: :restrict_with_exception
   has_many :external_form_publications, class_name: 'HmisExternalApis::ExternalForms::FormPublication', dependent: :destroy
-  has_many :custom_data_element_definitions, class_name: 'Hmis::Hud::CustomDataElementDefinition', dependent: :nullify,
-                                             primary_key: 'identifier', foreign_key: 'form_definition_identifier'
+
+  # --- Relations by identifier ----
+  has_many :instances, foreign_key: 'definition_identifier', primary_key: 'identifier'
+  has_many :custom_service_types, through: :instances, foreign_key: 'identifier', primary_key: 'form_definition_identifier'
+  has_many :custom_data_element_definitions, class_name: 'Hmis::Hud::CustomDataElementDefinition', primary_key: 'identifier', foreign_key: 'form_definition_identifier'
   has_one :published_version, -> { order(version: :desc).published }, class_name: 'Hmis::Form::Definition', primary_key: 'identifier', foreign_key: 'identifier'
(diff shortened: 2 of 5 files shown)


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -40,13 +40,15 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
...
   attr_accessor :filter_context
 
-  has_many :instances, foreign_key: :definition_identifier, primary_key: :identifier, dependent: :restrict_with_exception
+  # --- Relations by id ----
   has_many :form_processors, dependent: :restrict_with_exception
-  has_many :custom_service_types, through: :instances, foreign_key: :identifier, primary_key: :form_definition_identifier
   has_many :external_form_submissions, class_name: 'HmisExternalApis::ExternalForms::FormSubmission', dependent: :restrict_with_exception
   has_many :external_form_publications, class_name: 'HmisExternalApis::ExternalForms::FormPublication', dependent: :destroy
-  has_many :custom_data_element_definitions, class_name: 'Hmis::Hud::CustomDataElementDefinition', dependent: :nullify,
-                                             primary_key: 'identifier', foreign_key: 'form_definition_identifier'
```
gigxz, line 49: I removed `dependent: :nullify` from this. We can't nullify because that would cause the column to nullify when a draft is deleted, even if there is a published version that still references those definitions.

It may be desirable to nullify (manually) when the "last" version of this form is deleted, but I didn't add that to the mutation because (1) It's not expected, because if it's the draft and theres no other versions that it shouldnt have any CDEDs yet, and (2) in the event that it DOES occur, if we nullified then we would have these floating CDEDs that we don't know where they came from

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -40,13 +40,15 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
...
+
+  # --- Relations by identifier ----
+  has_many :instances, foreign_key: 'definition_identifier', primary_key: 'identifier'
```
gigxz, line 49: Removed `dependent: :restrict_with_exception`

gigxz:
address test failures https://github.com/greenriver/hmis-warehouse/actions/runs/9861244157/job/27229263496

update: addressed with https://github.com/greenriver/hmis-warehouse/pull/4523/commits/956ac5f0c4e8b8de6ae7c83f56d0350fcff44d93. I also added a before_destroy hook to prevent destruction of non-draft forms.

martha (approved):
everything looks great to me!

drivers/hmis/app/graphql/mutations/delete_form_definition.rb
```diff
@@ -17,7 +17,16 @@ def resolve(id:)
...
+      has_other_versions = definition.all_versions.count > 1
+
+      Hmis::Form::Definition.transaction do
+        # If this is the last remaining version of this form, remove all Form Instances.
+        definition.instances.each(&:destroy!) unless has_other_versions
+
+        # Soft-delete the Form Definition. This will error if there are any form_processors or external_form_submissions
+        # that reference this particular form, which there shouldn't be because it's a Draft.
+        definition.destroy!
+      end
```
ttoomey, line 29: Looks like there is a race condition here. Maybe not worth the trouble since it's unlikely to be triggered, but could fix it by wrapping it in an advisory lock. 

Normally you could use definition.with_lock {} but the logic here involves multiple FD rows

```ruby
      Hmis::Form::Definition.with_advisory_lock("hmis-fd-delete-mutation") do
        has_other_versions = definition.all_versions.count > 1
        Hmis::Form::Definition.transaction do
          # If this is the last remaining version of this form, remove all Form Instances.
          definition.instances.each(&:destroy!) unless has_other_versions

          # Soft-delete the Form Definition. This will error if there are any form_processors or external_form_submissions
          # that reference this particular form, which there shouldn't be because it's a Draft.
          definition.destroy!
        end
      end
```

ttoomey (approved):
Looks good, thanks for fixing!

drivers/hmis/app/graphql/mutations/delete_form_definition.rb
gigxz: Huh I don't see where the race condition is, can you point it out?

gigxz: I'm going to merge so I can get this up on QA, but will open a second PR if needed (And would love to see what I'm missing!)

ttoomey: > I'm going to merge so I can get this up on QA, but will open a second PR if needed (And would love to see what I'm missing!)

Sounds good! So I think it's around the conditional delete on has_other_versions. In this case this condition is unlikely to occur, maybe even vanishingly unlikely so perhaps this is just academic.

But Let's say we receive two simultaneous requests and they are processed in parallel to delete two different FDs records with the same identifier. Then let's say thread 1 queries for has_other_versions, then thread 2 queries has_other_versions at the same time, before any delete occurs. They both get a count  of 2 so neither of them destroys instances. But if they ran in series the second request would see a count of 1 and the delete would happen. I guess the outcome is we end up with orphan form instances which maybe doesn't matter than much.

I'd be fine with ignoring this since it's probably not going to occur in production

gigxz: Ahh I see, thanks for the explanation. I think we can ignore it, since deleting a form draft is an uncommon action, and the negative outcome (orphaned instances) is not that bad.

PR #4526, opened 2024-07-09
Prevent re-referring when enrollment exists

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6282

This PR
- Throws errors when someone tries to accept a referral that would create an enrollment that conflicts with another existing enrollment (which is not HUD compliant). Step 2 of the linked ticket
- Issues a warning when someone tries to refer such a client in the first place. Step 1 of the linked ticket

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -446,6 +446,33 @@ def project_configs
       Hmis::ProjectConfig.all
     end
 
+    field :can_project_accept_referral, Boolean, null: false do
```
martha: This query ended up here, rather than on the Project, because any query to the Project fails if the user doesn't have permission to view that project. That doesn't work for this purpose, since we want users to be able to make this query even if they don't have project viewing permission.

I'd be glad to further improve this -- do you think it would rather live on the referring Enrollment, for example?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -446,6 +446,33 @@ def project_configs
...
+      argument :project_id, ID, required: true
+      argument :referrer_enrollment_id, ID, required: true
+    end
+    def can_project_accept_referral(project_id:, referrer_enrollment_id:)
```
martha: I'd also be grateful for feedback about the query name -- I think `can_project_accept_referral` is nice because it's simple, but it might be a bit confusing because there are other conditions for accepting referrals which we don't check here, like:
- project has CoC (?)
- project has available units

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Ah good point. I think keeping on the root query is just fine. The "can_" prefix is really established as a permissions convention, so, I think maybe this should have another name. What about this:

```
    field :project_can_accept_referral, Boolean, 'Whether the destination project is able to accept a referral for the client(s) belonging to the source enrollment', null: false do
      argument :source_enrollment_id, ID, required: true
      argument :destination_project_id, ID, required: true
    end
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: I suggested `project_can_accept_referral` in another comment. I think this actually does check all the conditions. Having a CoC or Units is not a requirement for being able to accept a referral. (In situations where they do use units, units are sometimes required on the sending form. That would be hard to know, from here)

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -162,6 +167,24 @@ def self.apply_filters(input)
...
+    # NOTE: is not optimized, could be refactored if performance is an issue. Used this approach to minimize
+    # duplication of project_match logic.
+    Hmis::Hud::Project.open_on_date(Date.current).select do |project|
+      instance_scope.any? { |instance| instance.project_match(project) }
+    end.map(&:id)
+  end
+
+  def receiving_referrals?
+    Hmis::Hud::Project.receiving_referrals_project_ids.include?(id)
+  end
```
gigxz: Maybe "receives_referrals?" no strong feeling there though

`Hmis::Form::Instance.active.with_role(:REFERRAL).any? { |instance| instance.project_match(self) }` would be more efficient

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -131,6 +131,11 @@ class Hmis::Hud::Project < Hmis::Hud::Base
     )
   end
 
+  scope :receiving_referrals, -> do
+    # See note below about performance
```
gigxz: Maybe we can move the logic into this scope since the `receiving_referrals?` check doesn't need to evaluate against all the open projects

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+    let!(:c2) { create :hmis_hud_client, data_source: ds1 }
+
+    # p1 is the referring project
+    let!(:p1_ac) { create_access_control(hmis_user, p1) }
+    let!(:referrer_enrollment) { create :hmis_hud_enrollment, client: c2, project: p1, data_source: ds1 }
+
+    it 'raises an error when the project doesnt accept referrals' do
+      response, result = post_graphql(project_id: p1.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
+      expect(response.status).to eq 500
+      expect(result.dig('errors')[0]['message']).to eq 'access denied'
```
gigxz: This helper should work:
```suggestion
      expect_access_denied post_graphql(project_id: p1.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
```

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+      expect(result.dig('data', 'canProjectAcceptReferral')).to be_truthy
+    end
+
+    it 'returns false when the client has an existing referral' do
+      create :hmis_hud_enrollment, client: c2, project: p2, data_source: ds1
+      response, result = post_graphql(project_id: p2.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
+      expect(response.status).to eq 200
+      expect(result.dig('data', 'canProjectAcceptReferral')).to be_falsey
+    end
+  end
```
gigxz: can we also test the case where a household member has an _exited_ enrollment in the receiving project? (should succeed)

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+
+    it 'raises an error when the current user doesnt have can_manage_outgoing_referrals permission' do
+      remove_permissions(access_control, :can_manage_outgoing_referrals)
+      remove_permissions(p1_ac, :can_manage_outgoing_referrals)
+      response, result = post_graphql(project_id: p2.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
+      expect(response.status).to eq 500
+      expect(result.dig('errors')[0]['message']).to eq 'access denied'
+    end
+
+    it 'returns true when the client has no existing referral' do
```
gigxz:
```suggestion
    it 'returns true when the client does not have an open enrollment at the receiving project' do
```

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+      response, result = post_graphql(project_id: p2.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
+      expect(response.status).to eq 200
+      expect(result.dig('data', 'canProjectAcceptReferral')).to be_truthy
+    end
+
+    it 'returns false when the client has an existing referral' do
```
gigxz:
```suggestion
    it 'returns false when the client has an open enrollment at the receiving project' do
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -446,6 +446,33 @@ def project_configs
...
+      raise 'access denied' unless current_user.can_manage_outgoing_referrals?
```
gigxz: So this needs to ensure that the caller can refer out from the specific source project, not just that they can refer out from _any_ project. The tests should reflect this as well.

You can check it with: `current_permission?(permission: :can_manage_outgoing_referrals, entity: source_enrollment.project)`

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+
+    it 'raises an error when the project doesnt accept referrals' do
+      response, result = post_graphql(project_id: p1.id, referrer_enrollment_id: referrer_enrollment.id) { can_accept_referral }
+      expect(response.status).to eq 500
+      expect(result.dig('errors')[0]['message']).to eq 'access denied'
+    end
+
+    it 'raises an error when the current user doesnt have can_manage_outgoing_referrals permission' do
+      remove_permissions(access_control, :can_manage_outgoing_referrals)
+      remove_permissions(p1_ac, :can_manage_outgoing_referrals)
```
gigxz: It's a little hard to tell what's going on with these tests when there are 2 access controls active. Unless we're explicitly trying to test how the access controls behave together, I would change this to only worry about 1 access control. (Or two, if we need to track 1 ACL tied to the source project, and 1 ACL tied to the destination project). In this case the `access_control` is tied to the whole data source, which makes it a little hard to follow.

Since this query no longer has to do with the project query, maybe we can just move this `describe` block to its own spec file? That might make it easier to manage the access controls.

drivers/hmis/spec/requests/hmis/project_spec.rb
```diff
@@ -240,6 +231,52 @@
...
+      GRAPHQL
+    end
+
+    let!(:p2) { create :hmis_hud_project, data_source: ds1, organization: o1, user: u1 }
+    let!(:referral_instance) { create :hmis_form_instance, role: :REFERRAL, entity: p2 }
+
+    let!(:c2) { create :hmis_hud_client, data_source: ds1 }
+
+    # p1 is the referring project
+    let!(:p1_ac) { create_access_control(hmis_user, p1) }
```
gigxz: I would say naming the vars "source_project"/"destination_project"  (or sending/receiving) is clearer than using the p1/p2 syntax. (I know I am guilty of this a lot😬 ).

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -446,6 +446,33 @@ def project_configs
...
+
+      # Can't accept the referral if any client in the household has an existing open enrollment in the project.
+      personal_ids = Hmis::Hud::Enrollment.
+        viewable_by(current_user).
+        find(referrer_enrollment_id).
+        household.clients.pluck(:personal_id)
+
+      # This doesn't check viewable_by either, see comment above
+      enrollments = load_ar_association(project, :enrollments).open_including_wip.where(personal_id: personal_ids)
+      enrollments.empty?
```
gigxz: Calling a scope on the result of a `load_ar_association` bypasses the benefit of the data loader, because it performs another query. Since we're not calling this query in batch, I think we can simplify this to:
```
! project.enrollments. open_including_wip.where(personal_id: personal_ids).exists?

File: drivers/hmis/app/graphql/types/hmis_schema/query_type.rb [hunk 13]
```diff
@@ -446,6 +446,33 @@ def project_configs
...
+      # - it is only available to users who can manage outgoing referrals
+      # - it doesn't expose any info about the actual enrollment(s), just a yes/no
+      project = Hmis::Hud::Project.find_by(id: project_id)
+      raise 'access denied' unless project.receiving_referrals?
+
+      # Can't accept the referral if any client in the household has an existing open enrollment in the project.
+      personal_ids = Hmis::Hud::Enrollment.
+        viewable_by(current_user).
+        find(referrer_enrollment_id).
+        household.clients.pluck(:personal_id)
```
gigxz: Untested suggestion that might be a bit more performant (no need to load the household view)
```
source_enrollment = Hmis::Hud::Enrollment.viewable_by(current_user).find(referrer_enrollment_id)
personal_ids = source_enrollment.household_members.pluck(:PersonalID)
```

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
```diff
@@ -48,6 +49,21 @@ def resolve(id:, input:)
           end
           raise ActiveRecord::Rollback if errors.any?
 
+          # This is similar to the logic in query_type.rb:can_project_accept_referral, but differs in a couple ways:
+          # - Here, we do use viewable_by when querying Enrollments. (We already threw an error above if the user doesn't have this permission)
+          # - We return information about each conflicting enrollment, so the user can fix the errors.
+          personal_ids = Hmis::Hud::Client.where(id: posting.referral.household_members.pluck(:client_id)).pluck(:PersonalID)
+          # use project_pk so that wip enrollments are included
+          conflicting_enrollments = Hmis::Hud::Enrollment.viewable_by(current_user).where(project_pk: posting.project.id, personal_id: personal_ids)
```
gigxz: We can just use `project` to hide these implementation details. The `project` belongs_to relation uses the project_pk under the hood.
```
          conflicting_enrollments = Hmis::Hud::Enrollment.viewable_by(current_user).where(project: posting.project, personal_id: personal_ids)
```

Although, I think it would be clearer to just directly query the enrollments on the project. There is no need to check viewable_by because we already raise an error if the user cant view enrollments in the receiving project.

```suggestion
          conflicting_enrollments = posting.project.enrollments.open_including_wip.where(personal_id: personal_ids)
```
It also looks like this would error if the client had an exited enrollment. Adding `open_including_wip` above should address that.

gigxz: Maybe a further improvement, to avoid the plucking, would be:
```
household_members = posting.referral.household_members.map(&:client)
conflicting_enrollments = posting.project.enrollments.open_including_wip.where(client: household_members)
```

drivers/hmis/spec/requests/hmis/update_referral_posting_spec.rb
```diff
@@ -56,4 +56,34 @@
...
+  context 'when an enrollment already exists' do
+    let!(:coc1) { create :hud_project_coc, data_source: ds1, project_id: p1.project_id, state: 'MA' }
+    let!(:access_control) { create_access_control(hmis_user, p1) }
+
+    let!(:c2) { create(:hmis_hud_client, data_source: ds1) }
+    let!(:r2) { create :hmis_external_api_ac_hmis_referral }
+    let!(:r2_hhm) { create :hmis_external_api_ac_hmis_referral_household_member, client: c2, referral: r2 }
+    let!(:rp2) { create :hmis_external_api_ac_hmis_referral_posting, identifier: nil, referral: r2, project: p1, status: 'assigned_status', unit_type: nil }
+
+    it 'should error on both WIP and non-WIP enrollment' do
```
gigxz: can we add a test case to ensure that it doesn't error if a household member has an Exited enrollment in the destination project?

gigxz (commented):
Looks good! Comments below :)

drivers/hmis/app/models/hmis/hud/project.rb
martha: Thanks, yeah this is a good point!

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
martha: I think the error here is with the enrollment <> client relationship, I'd like to understand better what is going on so I'll dig a bit more.
```
> c = Hmis::Hud::Client.first
> e = Hmis::Hud::Enrollment.where(client: c)

/bundle/ruby/3.1.0/gems/activemodel-7.0.8.4/lib/active_model/attribute_methods.rb:469:in `respond_to?': [:data_source_id, :PersonalID] is not a symbol nor a string (TypeError)

```

in context,
```
Error: the evaluation of `posting.project.enrollments.open_including_wip.where(client: household_members)` failed with the exception '[:data_source_id, :PersonalID] is not a symbol nor a string
```

gigxz: Ah okay, my bad. I thought our composite key relations supported that, but I guess they don't. I'm seeing the same behavior as you.

gigxz: Sorry for the confusion 😅 
We can't query like that because of the composite key relations that those fields use. https://github.com/greenriver/hmis-warehouse/blob/bcca50d001347b39946bab623eb17b95db09973f/drivers/hmis/app/models/hmis/hud/base.rb#L32-L69

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
```diff
@@ -49,12 +49,10 @@ def resolve(id:, input:)
...
           raise ActiveRecord::Rollback if errors.any?
 
-          # This is similar to the logic in query_type.rb:can_project_accept_referral, but differs in a couple ways:
-          # - Here, we do use viewable_by when querying Enrollments. (We already threw an error above if the user doesn't have this permission)
-          # - We return information about each conflicting enrollment, so the user can fix the errors.
-          personal_ids = Hmis::Hud::Client.where(id: posting.referral.household_members.pluck(:client_id)).pluck(:PersonalID)
-          # use project_pk so that wip enrollments are included
-          conflicting_enrollments = Hmis::Hud::Enrollment.viewable_by(current_user).where(project_pk: posting.project.id, personal_id: personal_ids)
+          # Similar to query_type.rb:project_can_accept_referral, but returns info about each conflicting enrollment, so the user can fix the errors.
+          personal_ids = posting.referral.household_members.map(&:client).pluck(:PersonalID)
```
gigxz: It's safe to use the PersonalID that is on the Enrollment record. We can be sure it will match the PersonalID on the Client, because it's used in the composite key relation (see my link in my previous comment).

For example: `posting.referral.household_members.pluck(:PersonalID)`.

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
```diff
@@ -48,6 +49,19 @@ def resolve(id:, input:)
...
+          # no need to check viewable_by on the enrollments, since we would have already raised 'access denied' above
+          conflicting_enrollments = posting.project.enrollments.open_including_wip.where(personal_id: personal_ids)
+
+          unless conflicting_enrollments.empty?
+            conflicting_enrollments.each do |e|
+              errors.add :base, :invalid, full_message: "#{e.client.full_name} already has an open enrollment in this project (entry date: #{e.entry_date}). Please exit the client if the enrollment is invalid or out-of-date, and otherwise deny the referral."
+            end
+
+            raise ActiveRecord::Rollback
```
gigxz: This is okay, but I also think this check doesn't _need_ to happen inside the transaction. It would be OK to check for conflicting enrollments first (if `posting_status_change == ['assigned_status', 'accepted_pending_status']`), and return immediately if there are errors due to the conflicting enrollments. Then we don't need to worry about rolling anything back.

I think that is generally safer, because we don't need to rely on the db transaction rollback behavior. The downside is that you only get the "first" error, for example if there are conflicting enrollments AND not enough units available, you wouldn't see the unit error until you resolved the conflicting enrollment errors.

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
martha: Ah I see :( thanks for looking into it though!

drivers/hmis/app/graphql/mutations/ac_hmis/update_referral_posting.rb
martha: It looks like the `ReferralHouseholdMember` type does not have a `PersonalID` column, only `client_id`. And looks like we can query the `Enrollments` table by this `client_id` only if we join to `clients` first. I'm not sure which one would be more performant, is it worth making a performance benchmark to find out?

gigxz: Oh I'm sorry, again! I was thinking it was enrollment.household_members, which are Enrollments. Nope this seems good!!

gigxz (approved):
👍

martha:
Thanks @gigxz for the thoroughness of your review!!! 🍝 => 🚄

PR #4531, opened 2024-07-10
Add 'Super Admin' mode

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6218
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/840

This PR
- Adds a new super-admin permission, `can_administrate_config`. I'm open to updating this name -- I decided not to literally use the phrase `super_admin` because it might be confusing to existing admins, but I'd be open to updating the language to just use it if we think it'd be clearer.
- checks for this permission on queries and mutations relating to forms, when the form role in question isn't in the list of non-admin roles (right now this is just SERVICE and CUSTOM_ASSESSMENT).
- adds spec tests.

I still have some todos and questions that I'll mark inline.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -407,25 +411,31 @@ def external_form_definition(identifier:)
...
       scope = Hmis::Form::Definition.non_static.valid.latest_versions
+      scope = scope.with_role(Hmis::Form::Definition::NON_ADMIN_FORM_ROLES) unless current_user.can_administrate_config?
       scope = scope.apply_filters(filters) if filters
       scope.order(updated_at: :desc)
     end
 
     form_rules_field
     def form_rules(**args)
-      raise 'Access denied' unless current_user.can_configure_data_collection?
+      access_denied! unless current_user.can_configure_data_collection?
```
martha, line 429: I still need to update this to filter out non-admin form roles if the current user lacks `can_administrate_config` permission. It should be intersected with the `form_type` filter argument if it's passed.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -436,7 +446,15 @@ def form_rules(**args)
...
       raise 'not allowed' unless current_user.can_configure_data_collection?
 
-      Hmis::Form::Instance.find_by(id: id)
+      instance = Hmis::Form::Instance.find_by(id: id)
+      ensure_form_role_permission(instance.definition.role)
+
+      instance
+    end
+
+    field :non_admin_form_role, Forms::Enums::NonAdminFormRole, null: false
```
martha: I'd like to get rid of this. It's here because unless the type `Forms::Enums::NonAdminFormRole` is used somewhere in the api, it won't be part of the code gen process that makes it available in `HmisEnums` in the frontend, to be used in the dropdown when creating/editing a form. Is there a better way to convince the graphql codegen step that this should be part of the api?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Ok, I've taken a stab at this, which I think is possibly not ideal because it's duplicating logic in the [FormInstanceFilter](https://github.com/greenriver/hmis-warehouse/blob/4c4a9e76d153de126451670502e2db9b30a60b10/drivers/hmis/app/models/hmis/filter/form_instance_filter.rb#L22-L28). This is not just bad because of the duplicated logic, but also because of the double query for form definition identifiers if the user passes in additional form roles to filter by.

However, the code is potentially clearer this way, and to be honest, I wasn't easily able to figure out how to overwrite the filters object with additional filters to pass (maybe using a [preprocessing step](https://graphql-ruby.org/fields/arguments.html#preprocessing) it would be possible?)

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
```diff
@@ -0,0 +1,16 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module ConfigToolPermissionHelper
+  extend ActiveSupport::Concern
+
+  def ensure_form_role_permission(role)
```
gigxz: So typically with something like this we would put the check as a special method on the on the `User` model. For example:

```
// true if user has can_manage_forms and this role is allowed, or if user has can_administrate_config
access_denied! unless user.can_manage_form_role?(role)
```

Can you try a refactor for that approach? It does mean that in some cases we need to look up the definition _before_ doing the permission check (so we can find the role) but I think it will be easier to reason about the mutation logic, if the permission check is just encapsulated in one method call.

drivers/hmis/app/graphql/mutations/create_form_definition.rb
```diff
@@ -6,12 +6,16 @@
...
+
     argument :input, Types::Admin::FormDefinitionInput, required: true
 
     field :form_definition, Types::Forms::FormDefinition, null: true
 
     def resolve(input:)
-      raise 'not allowed' unless current_user.can_manage_forms?
+      access_denied! unless current_user.can_manage_forms?
+
+      ensure_form_role_permission(input.role) if input.role # (if form_role isn't provided, will hit 'Definition invalid' below)
```
gigxz: Hmm so since this mutation uses a static form with a custom mutation, it doesn't have the built-in validation that we get when using SubmitForm (which validates the response against the fields that are `required` in the form definition. This is probably pointing to a larger issue, and perhaps when we move to RHF we could have this required-ness validation enforced by the frontend. However, for now, lets return user-facing messages:
```
      errors = HmisErrors::Errors.new
      errors.add(:role, :required) if input.role.blank?
      errors.add(:title, :required) if input.title.blank?
      errors.add(:identifier, :required) if input.identifier.blank?

      # do permission check on input.role
```

drivers/hmis/lib/form_data/static/form_definition.json
```diff
@@ -7,9 +7,9 @@
       "mapping": {
         "field_name": "role"
       },
-      "_comment": "FIXME this dropdown should be restrictied to form types that can be created/managed through the UI. Maybe start small initially.",
-      "pick_list_reference": "FormRole",
-      "required": true
+      "pick_list_reference": "NonAdminFormRole",
```
gigxz: What happens when a "super admin" tries to create a new form? They should see the full list of roles. I think we'll need to use a backend picklist for that.

drivers/hmis/app/graphql/mutations/create_form_definition.rb
```diff
@@ -23,6 +27,9 @@ def resolve(input:)
         **attrs,
       )
 
+      # Raise if the definition model is not valid (invalid role/status/identifier; not expected)
+      raise "Definition invalid: #{definition.errors.full_messages}" unless definition.valid?
```
gigxz, line 35: we can keep this in addition to the above suggestion

drivers/hmis/app/graphql/mutations/create_next_draft_form_definition.rb
```diff
@@ -22,10 +24,15 @@ def resolve(identifier:)
...
-      return { form_identifier: existing_drafts.first } unless existing_drafts.empty?
+
+      unless existing_drafts.empty?
+        ensure_form_role_permission(existing_drafts.first.role)
+        return { form_identifier: existing_drafts.first }
+      end
 
       # Re-fetch the most recent version (could be published or retired) to get the full form definition
       last = Hmis::Form::Definition.find(definitions.max_by(&:version).id)
+      ensure_form_role_permission(last.role)
```
gigxz: It is a little confusing that this check happens in two places. Related to my other suggestion about moving the check to the user, we can just check the permissions one time against `definitions.first.role`

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -14,8 +16,13 @@ class UpdateFormDefinition < CleanBaseMutation
     def resolve(id:, input:)
       access_denied! unless current_user.can_manage_forms?
 
+      ensure_form_role_permission(input.role) if input.role
+
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition
+
+      ensure_form_role_permission(definition.role)
```
gigxz: same suggestion here about moving permission check down. It looks like we explicitly want to check against the old role AND the new role, if it has changed, so maybe add a comment about that? (The UI does currently allow changing the role).

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: This looks good to me @martha . Instance has the `with_role` scope too, so we could use that in the same way you did for the form identifiers query 
```
scope = Hmis::Form::Instance.all
scope = scope.with_role(Hmis::Form::Definition::NON_ADMIN_FORM_ROLES) unless current_user.can_administrate_config?
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -218,6 +218,12 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   }.freeze
   NON_QUESTION_ITEM_TYPES = ['DISPLAY', 'GROUP'].freeze
 
+  # NON_ADMIN_FORM_ROLES are editable by everyone with can_manage_forms permission
+  NON_ADMIN_FORM_ROLES = [
```
gigxz, line 223: This naming has been tripping me up a little bit. I think it might be clearer to just name it in the way you did in the code comment. `USER_EDITABLE_FORM_ROLES`? I think the confusing part is that the people editing them are still "admins", and the var name doesn't mention "editable" or "configurable".

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -397,6 +397,13 @@ def self.permissions_with_descriptions
         category: 'Administration',
         sub_category: 'Admin Tools',
       },
+      can_administrate_config: {
+        description: 'Ability to use the JSON form editor, full Form Builder, and fragment management.',
```
gigxz:
```suggestion
        description: 'Ability to use the JSON form editor and administer all form types [GR Staff Only]',
```
?
(don't want it to sound too enticing 😄 )

gigxz (commented):
looking good! comments below

drivers/hmis/app/graphql/mutations/create_form_definition.rb
gigxz: https://github.com/greenriver/hmis-warehouse/pull/4540/files#r1676087231

drivers/hmis/app/models/hmis/form/definition.rb
martha: Ah, yeah I struggled a lot with this naming. I had `EDITABLE_FORM_ROLES` initially and would be fine with changing it back to something similar. The reason I moved away from that is that I started to confuse myself about this group of form rules being non-editable for some users. For example, in the [`form_definition`](https://github.com/greenriver/hmis-warehouse/blob/4c4a9e76d153de126451670502e2db9b30a60b10/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L385-L395) and [`form_identifier(s)`](https://github.com/greenriver/hmis-warehouse/blob/4c4a9e76d153de126451670502e2db9b30a60b10/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L406-L424) queries, this list is used to determine which models are available to view for users who have `can_configure_data_collection` but not `can_administer_config` -- but they aren't necessarily able to edit, which is determined by `can_manage_forms`.

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
martha: I'm happy to refactor to put this role-check permission on the `User` model (or, the [UserPermissions](https://github.com/greenriver/hmis-warehouse/blob/4c4a9e76d153de126451670502e2db9b30a60b10/app/models/concerns/user_permissions.rb#L4) concern?), but want to note that I specifically did not make `ensure_form_role_permission` check for `can_manage_forms`, because `ensure_form_role_permission` is also used to check access (not edit) permissions for form definitions/identifiers, which the user only needs `can_configure_data_collection` to view. (This is a bit related to my other comment [here](https://github.com/greenriver/hmis-warehouse/pull/4531#discussion_r1676377813))

martha: Maybe the naming is more confusing than it needs to be? I'll think more on the permission names and at the very least add some clarifying comments

martha: Maybe the `UserPermissions` concern should get 2 new methods, `can_view_form_role` and `can_edit_form_role`

martha: I guess that naming still doesn't fully clarify it, because `can_configure_data_collection` is an editing permission. So e.g. in the `CreateFormRule` mutation we want to check `can_configure_data_collection` combined with `can_administrate_config` if the `role` is one of the restricted roles, and we don't want to just call that `can_view_form_role`, because they are making a mutation, but it doesn't require full `can_manage_forms` permission either.

martha: I also think there is an advantage to checking `can_manage_forms?` separately before `can_administrate_config` in the cases where we have to query for a definition/instance in order to get the form role. If we bundle them together and only check after querying for the definition, we may return 'not found' when the user should actually see 'access denied', which would be a permission leak, right?

gigxz: Alright gotcha. It does make it confusing. I think I'm a little confused by the way that `ensure_form_role_permission` is used separately from the permission checks. That makes the mutation code hard to follow, for me anyway.

Maybe I am over-simplifying here, but can we just create a separate method for the "configure data collection" perm?
```
user.can_manage_forms_for_role?(role)?  // true if can_manage_forms && role is permitted
user.can_configure_data_collection_for_role?(role)?  // true if can_configure_data_collection && role is permitted
```

Then we only need to wrap our heads around those 2 permissions. This is similar conceptually to how permissions apply differently based on the project, so we have a mental model for it. (Eg the user has this permission in the context of this "entity", which in this case is a role, sort of).


```
# form mutations
access_denied! unless current_user.can_manage_forms_for_role?(input.role)

# form rule mutations
access_denied! unless current_user.can_configure_data_collection_for_role(definition.role)

```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -389,9 +390,12 @@ def form_definition(id:)
       # NOTE: this query is only used for form management. It probably should
       # not be used for the application, because there is no project context passed
       # to the definition.
-      raise 'Access denied' unless current_user.can_configure_data_collection?
+      access_denied! unless current_user.can_configure_data_collection?
+
+      definition = Hmis::Form::Definition.find(id)
+      ensure_form_role_permission(definition.role)
```
gigxz: I don't think we need to bother adding a check to this query. It's OK for these users to _see_ the forms for other roles (the forms aren't secret or anything). For now we are deciding not to expose them in the main UI, which you've already implemented below by filtering down the results returned from form_identifiers. 

So, it doesn't hurt to add it, but it's additional complexity that we don't really need imo

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -407,27 +411,36 @@ def external_form_definition(identifier:)
...
     end
     def form_identifier(identifier:)
-      raise 'Access denied' unless current_user.can_configure_data_collection?
+      access_denied! unless current_user.can_configure_data_collection?
 
-      Hmis::Form::Definition.non_static.latest_versions.where(identifier: identifier).first
+      definition = Hmis::Form::Definition.non_static.latest_versions.where(identifier: identifier).first
+      raise 'not found' unless definition
+
+      ensure_form_role_permission(definition.role)
```
gigxz: Same comment here about the complexity not being strictly necessary. If a user can navigate to `forms/client` or whatever and see the form info it's OK. The intended purposes is (1) make sure they can't edit those forms, and (2) for now, don't show them at all in the main form list, just to simplify the UI. Can take or leave this, just wanted to communicate that reasoning.

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
gigxz: Sorry I didn't see your last 3 comments when I posted @martha . Reading..

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Unless I am misunderstanding with my comment https://github.com/greenriver/hmis-warehouse/pull/4531#discussion_r1675962139 I think we need to replace this with a backend picklist anyway. GR staff should be allowed to create forms for all roles

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: Ah right, we're gating visibility on them too, right now. I guess NON_ADMIN_FORM_ROLES makes sense then!

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
gigxz: So, I _think_ my suggestion addresses your concern mentioned https://github.com/greenriver/hmis-warehouse/pull/4531#discussion_r1676396640, but let me know if it does not.

And to address this one:

> I also think there is an advantage to checking can_manage_forms? separately before can_administrate_config in the cases where we have to query for a definition/instance in order to get the form role. If we bundle them together and only check after querying for the definition, we may return 'not found' when the user should actually see 'access denied', which would be a permission leak, right?

That's a good thought about the leak, it's true it exposes some more information. In this case, that information is just about the existence of a certain form identifier, or form role, or form id. None of those are really leaks that concern me. They will see the these forms when interacting with the application, as do all users. Forms are essentially _config_, not data. Anyone can see it, but only certain people can edit it. That is the philosophy I would stick with, anyway. (As you can see, we have _no_ permission checks on [this](https://github.com/greenriver/hmis-warehouse/blob/51b622d79a3b74bcc64c304fa3539ee653554cec/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L178) or [this](https://github.com/greenriver/hmis-warehouse/blob/51b622d79a3b74bcc64c304fa3539ee653554cec/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L232) query for fetching FDs. At some point we decided to add permission checks for the FD queries that are used just for configuration, mainly just as an implementation note that they _should only_ be used for configuration screens, not for application screens. That may have been a confusing choice.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Sounds good! This needed a small tweak [here](https://github.com/greenriver/hmis-warehouse/pull/4531/files#diff-209b289761c1bb46bd55cd35081e14b1912ec5704296c27b7fcca123ef8d41f6R41) to accept a list of roles instead of a single role

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
martha: Okay, thanks for this and sorry for getting so bogged down, thanks for helping me sort it out! I think we settled on similar ideas at the same time where `UserPermissions` gets 2 more permissions that each take `role` as an input, one for management and one for data configuration. I think it's much more clear now

drivers/hmis/app/graphql/mutations/create_next_draft_form_definition.rb
martha: Sounds good. I've added ordering (by version descending) to the initial query, that way we always know the first definition in the returned results will be the most recent version which is most relevant to this mutation.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Both sound good. I'm happy to revert to the simpler version.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -129,6 +129,14 @@ def self.static_options_for_type(pick_list_type, user:)
         client_audit_event_record_type_picklist
       when 'PROJECTS_RECEIVING_REFERRALS'
         projects_receiving_referrals
+      when 'FORM_TYPES'
```
martha, line 132: Thanks @gigxz you are right this needs to be a backend picklist. I'm not sure I have totally done this right; I wanted to reuse the logic about capitalizing 'CE' which seems to live in `use_enum_with_same_key`

gigxz: This seems right. Can you add a comment explaining why the extra permission is checked?

drivers/hmis/app/graphql/concerns/config_tool_permission_helper.rb
gigxz: Ok, great, thanks!

app/models/concerns/user_permissions.rb
```diff
@@ -215,6 +217,14 @@ def can_edit_some_project_groups
       can_edit_assigned_project_groups? || can_edit_project_groups?
     end
 
+    def can_manage_forms_for_role(role)
+      can_manage_forms? && (Hmis::Form::Definition::NON_ADMIN_FORM_ROLES.include?(role) || can_administrate_config?)
```
gigxz: To make this slightly more robust, we can do `role.to_s` to ensure that `user.can_manage_forms_for_role?(:SERVICE)` works the same way as `user.can_manage_forms_for_role?('SERVICE')`

gigxz (approved):
Looks good! Thanks!

app/models/concerns/user_permissions.rb
martha: Good point thank you

PR #4534, opened 2024-07-10
Add missing types from form schema, don't require autofill conditions

gigxz:

## Description

1. add group components that were missing from schema
2. don't require conditions on autofill. We have forms that have autofills that should always run, for example when summing up dependent fields. Currently those forms set dummy values. Update the schema to just allow them to be nullable. I kept the GraphQL schema as-is and added default_value behavior, so as not to cause a breaking change. This does not affect behavior, because the frontend already treats `autofillBehavior: 'ANY'. autofillWhen: []` as "always autofill"

## Type of change
- [x] Bug fix


drivers/hmis/app/graphql/types/forms/autofill_value.rb
```diff
@@ -19,9 +19,9 @@ class Forms::AutofillValue < Types::BaseObject
     field :sum_questions, [String], 'Link IDs of numeric questions to sum up and set as the value if condition is met', null: true
     field :formula, String, 'Expression with mathematical or logical function defining the value', null: true
 
-    # Condition specifying when to perform this autofill
-    field :autofill_behavior, Types::Forms::Enums::EnableBehavior, null: false
-    field :autofill_when, [Types::Forms::EnableWhen], null: false
+    # Condition specifying when to perform this autofill. If not provided, the autofill will always run.
+    field :autofill_behavior, Types::Forms::Enums::EnableBehavior, null: true
+    field :autofill_when, [Types::Forms::EnableWhen], null: true
```
gigxz: actually this could cause a bug when deployed without the frontend change, hold off

gigxz: https://github.com/greenriver/hmis-frontend/pull/843

gigxz:
this could be re targeted to 124

martha (approved):

ttoomey (approved):
looks good, sorry I overlooked those component types

PR #4541, opened 2024-07-12
Update SPM by Subpop to use PIT races

awisema:

## Description

Use the PIT races instead of the HMIS race columns

## Type of change


app/models/concerns/filter/filter_scopes.rb
```diff
@@ -158,6 +159,83 @@ module Filter::FilterScopes
...
+        when :multi_racial_hispanic_latinaeo
+          race_ethnicity_alternative(scope, :MultiRacial, true)
+        when :race_none
+          race_ethnicity_alternative(scope, :RaceNone)
+        end
+        race_ethnicity_scope = add_alternative(race_ethnicity_scope, alternative)
+      end
+
+      scope.joins(:client).merge(race_ethnicity_scope)
+    end
```
eanders: Rather than a case statement why not something declarative:
```suggestion        
        options = race_alteratives[combination.to_sym]
        alternative = race_ethnicity_alternative(scope, **options)
        race_ethnicity_scope = add_alternative(race_ethnicity_scope, alternative)
      end

      scope.joins(:client).merge(race_ethnicity_scope)
    end

    def race_alteratives 
      @race_alteratives ||= {
        am_ind_ak_native: [:AmIndAKNative],
        am_ind_ak_native_hispanic_latinaeo: [:AmIndAKNative, true],
        asian: [:Asian],
        asian_hispanic_latinaeo: [:Asian, true],
        black_af_american [:BlackAfAmerican],
        black_af_american_hispanic_latinaeo: [:BlackAfAmerican, true],
        hispanic_latinaeo: [:HispanicLatinaeo],
        mid_east_n_african: [:MidEastNAfrican],
        mid_east_n_african_hispanic_latinaeo: [:MidEastNAfrican, true],
        native_hi_pacific: [:NativeHIPacific],
        native_hi_pacific_hispanic_latinaeo: [:NativeHIPacific, true],
        white: [:White],
        white_hispanic_latinaeo: [:White, true],
        multi_racial: [:MultiRacial],
        multi_racial_hispanic_latinaeo: [:MultiRacial, true],
        race_none: [:RaceNone]
      }
    end
```

Or better yet, if there's some way we can lean on `HudUtility2024.race_ethnicity_combinations` that would be really nice.

app/models/concerns/filter/filter_scopes.rb
```diff
@@ -158,6 +159,83 @@ module Filter::FilterScopes
...
+    private def race_ethnicity_alternative(scope, key, hispanic_latinaeo = false)
+      columns = {
+        AmIndAKNative: 0,
+        Asian: 0,
+        BlackAfAmerican: 0,
+        NativeHIPacific: 0,
+        White: 0,
+        HispanicLatinaeo: 0,
+        MidEastNAfrican: 0,
+      }
```
eanders: I'm hoping we can decrease the proliferation of race (and gender) column names.  Maybe?
```suggestion
      columns = (HudUtility2024.race_fields - [:RaceNone]).map { |k| [k, 0] }.to_h
```

drivers/homeless_summary_report/app/models/homeless_summary_report/client.rb
```diff
@@ -22,14 +22,21 @@ class Client < GrdaWarehouseBase
...
+      :black_af_american_hispanic_latinaeo,
+      :hispanic_latinaeo,
+      :mid_east_n_african,
+      :mid_east_n_african_hispanic_latinaeo,
+      :native_hi_pacific,
+      :native_hi_pacific_hispanic_latinaeo,
       :white,
+      :white_hispanic_latinaeo,
       :multi_racial,
+      :multi_racial_hispanic_latinaeo,
```
eanders: Not sure if we can easily pull these from `HudUtility2024.race_ethnicity_combinations`, but if we can, we should.

drivers/homeless_summary_report/app/models/homeless_summary_report.rb
```diff
@@ -8,4 +8,6 @@ module HomelessSummaryReport
   def self.table_name_prefix
     'homeless_summary_report_'
   end
+
+  REVISION_DATE = '2024-07-18'.to_date
```
eanders, line 12: Does this need to reflect the date it is deployed to production?

drivers/homeless_summary_report/app/models/homeless_summary_report/report.rb
```diff
@@ -823,70 +823,19 @@ def self.available_variants
     end
 
     def self.demographic_variants
```
eanders, line 825: What do you think about moving this to the `HudUtility2024` class? Then it could feed the various filter scopes, and other bits, and can be expanded with new keys as necessary.

eanders (changes requested):
Thanks @awisema this looks like a really good start.  Before we merge I'd like to DRY up some more of the race boilerplate so we can re-use it more easily in the future.

drivers/homeless_summary_report/app/models/homeless_summary_report.rb
awisema: Yes -- I considered just adding a `version` column to the reports table, and we could still do that, but it would require changing the message to reference a report version instead of a date, which seemed less helpful to the end user without more thought (e.g., do we show a version label on the tables?).

Is there a way we can programmatically determine the data a particular version of the code base was deployed?

drivers/homeless_summary_report/app/models/homeless_summary_report/report.rb
awisema: It seems very tightly tied to `Filter::FilterScopes`, so I am hesitant to tie it to `HudUtility2024` -- So, instead, I created a utility module `Filters::DemographicFilters`

eanders: Looks good, thank you.

eanders (approved):
Thanks @awisema! As far as I can tell this is working as expected.

PR #4542, opened 2024-07-12
Dg/datalab_test_kit_2024-07-11

dtgreiner:

## Description

Open up Organization J for the testkits and clean up some CAPER Q6 issues

## Type of change
- [X] Bug fix


drivers/hud_apr/app/models/hud_apr/generators/shared/fy2024/dq/question_three.rb
```diff
@@ -110,8 +110,11 @@ def q3_hoh_relationship_row(sheet, universe_members)
     def q3_client_location_row(sheet, universe_members)
       hoh_scope = universe_members.where(hoh_clause)
 
+      valid_cocs = HudUtility2024.cocs.keys
+      valid_cocs -= ['XX-502', 'XX-518'] # remove "invalid" testkit CoCs
```
eanders: Huh, this is tricky.  I don't think we want to remove these because the test kit considers `XX-500` and `XX-501` valid.  These follow the same pattern, so I'd expect them to be valid.  Maybe we should check that the format is valid (AA-DDD) and the enrollment_coc is in the CoCs that are selected in the filter?  I don't really understand how we're getting these values at all, doesn't the report filter to only the chosen CoC?

dtgreiner: The gist of what I've been getting is that Organization J has XX-518 data included in their results and counting the records in all of the result cells. I've set the Organization J tests to include this  CoC so we can verify additional tests, however, this question is looking for CoC data that isn't valid and is flagging XX-518 as such.

This is definitely a tricky one and I'm happy to hold off on it until we can dig more into it later.

eanders: Let's discuss when I get back.

dtgreiner:
@eanders This PR for the testkits should be ready as long as all of the tests sill pass in the github action

PR #4543, opened 2024-07-12
Update e2e test script to work for local development

gigxz:

## Description

* Use arm 64 compatible chrome image
* Add `--dev` flag to script
* Update documentation and move it from `spec/support/E2E_README.md` into the main HMIS README

I have sign_in_spec and intake_assessment_spec working. The enrollment management spec is giving me trouble. I will address separately

## Type of change


spec/support/E2E_README.md
```diff
@@ -21,42 +21,48 @@ Since E2E tests are expensive -- slow to run and fiddly to write/update -- we sh
...
+  If you want to restart with a different branch of the frontend, or need to re-pull the branch: exit and create a new shell (`dcr shell`), and re-start the run_hmis_system_tests script.
 
 ### Debugging
 The `debug` helper, defined in [e2e_tests.rb](e2e_tests.rb), enables pausing the driver and inspecting what is going on at the moment in the browserless Chrome.
 - Add `debug` on its own line. The test will pause execution at that line, output "Cuprite execution paused" to stdout, and output a localhost link to open the Chrome inspector. Open that localhost link.
 - Click the icon for Current Sessions (left sidebar between `<>` and the Settings gear icon).
 - Under "Current Sessions" there should be a link to "Open Path HMIS." Clicking this link will take you to the current session, in which you can interact and inspect the DOM.
+
+
+## Run Full E2E test suite locally
```
gigxz, line 51: I moved this section to the end, since it doesn't seem like something we will do often.

spec/support/E2E_README.md
```diff
@@ -21,42 +21,48 @@ Since E2E tests are expensive -- slow to run and fiddly to write/update -- we sh
...
+4. Once that process has completed, you can kick off your `rspec` test(s).
     ```bash
-    HOSTNAME=`hostname`
-    unset HMIS_OKTA_CLIENT_ID
-    unset OKTA_DOMAIN
+    HOSTNAME=`hostname` RUN_SYSTEM_TESTS=true RAILS_ENV=test CAPYBARA_APP_HOST="http://$HOSTNAME:5173" rspec -f d -P "drivers/hmis/spec/system/hmis/*"
     ```
-    Then, replacing `/*` as needed with the specific test file/line that you want to run,
+
+    Note: if you'd like, you can do this step in a separate terminal tab. To do so, you must connect to the same tab where you have the web server running (from step 3). For example:
```
gigxz, line 36: I kept this note here, even though I don't really see much of a need for running them in two separate shells. 
The approach I take is to just start the frontend server in the background, and then you can kick off your tests.

spec/support/E2E_README.md
```diff
@@ -21,42 +21,48 @@ Since E2E tests are expensive -- slow to run and fiddly to write/update -- we sh
...
-5. In the first tab, clone the repo and serve the frontend. You can do this by running the `bin/run_hmis_system_tests.sh` with the following modifications:
-    - comment out the `trap cleanup EXIT` line (we don't want to clean up after exiting the script)
-    - comment out everything after the `yarn preview` line
 
-6. In the other tab, run the `rspec` tests. Choosing just what we need from `bin/run_hmis_system_tests.sh`,
+4. Once that process has completed, you can kick off your `rspec` test(s).
     ```bash
-    HOSTNAME=`hostname`
-    unset HMIS_OKTA_CLIENT_ID
-    unset OKTA_DOMAIN
```
gigxz, line 49: This seems unnecessary to call out, since our default test env already does it: https://github.com/greenriver/hmis-warehouse/blob/8e0ec4f64eba4c9287f8e51e84a61fdc08924eb8/.env.test#L39-L40
@ttoomey maybe you have a local override file that was setting them, or something?

spec/support/E2E_README.md
```diff
@@ -21,42 +21,48 @@ Since E2E tests are expensive -- slow to run and fiddly to write/update -- we sh
...
     ```bash
-    docker ps # grab the container ID for the Docker shell
-    docker exec -it $CONTAINER_ID /bin/bash # with the container ID from above
+    BRANCH_NAME=release-X ./bin/run_hmis_system_tests.sh --dev
     ```
-   You should be connected to the same Docker container in two tabs: one to serve the frontend, and the other to run the tests.
-   
-5. In the first tab, clone the repo and serve the frontend. You can do this by running the `bin/run_hmis_system_tests.sh` with the following modifications:
-    - comment out the `trap cleanup EXIT` line (we don't want to clean up after exiting the script)
-    - comment out everything after the `yarn preview` line
```
gigxz, line 43: these are replaced with passing the `--dev` flag

spec/support/E2E_README.md
ttoomey: Good spot, yes I think these env vars can be removed, it's probably not needed since we unset them in env.test. I don't remember why we had these originally

ttoomey (approved):
Thanks for improving the docs and test runner!

PR #4553, opened 2024-07-17
6232 external client identification

ttoomey:

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6232)

* adds job to process and match external client ids. It's not actually hooked up to anything yet, we'll need to know the remote credentials and decide on when/how the job should run 
* per spec in issue, I extracted and reused the client lookups from `GrdaWarehouse::Tasks::IdentifyDuplicates`. I did end up refactoring the lookup code quite a bit. The tests pass but this definitely needs another set of eyes and some QA

## Type of change


app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,142 @@
...
+  def upload_name(filename)
+    dir = File.dirname(filename)
+    dir = dir == '.' ? nil : dir
+    base = File.basename(filename, '.*')
+    ext = File.extname(filename)
+    [dir, "#{base}-results#{ext}"].compact.join('/')
+  end
+
+  private def build_lookups
+    @name_lookup = GrdaWarehouse::ClientMatcherLookups::ProperNameLookup.new(transliterate: true)
```
ttoomey, line 63: Maybe we don't want this but it seems like transliteration might be helpful ("Jose" == "José")

app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,142 @@
...
+    @ssn_lookup = GrdaWarehouse::ClientMatcherLookups::SSNLookup.new(format: :last_four)
+    @dob_lookup = GrdaWarehouse::ClientMatcherLookups::DOBLookup.new
+
+    clients = GrdaWarehouse::Hud::Client.all
```
ttoomey: do we want to match all clients or just destinations?

app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,142 @@
...
+    match ? { 'GHOCID' => ghocid, 'Client ID' => match } : nil
+  end
+
+  # id is found in at least two match sets
+  def get_first_match(matches, required_number: 2)
+    count = Hash.new(0)
+    matches.each do |match|
+      match.each do |id|
+        count[id] += 1
+        return id if count[id] >= required_number
```
ttoomey: perhaps we should improve this to iterate over all three match sets in case there's a match on all 3, then do some kind of tie breaker

app/jobs/identify_external_clients_job.rb
eanders: That's great, thanks!

app/jobs/identify_external_clients_job.rb
eanders: I believe we only want to match source clients, but return the destination ID for the matching source client.  Destinations potentially have amalgamated PII, which we probably don't want to match.

app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
```diff
@@ -0,0 +1,41 @@
...
+      key = [first_name, last_name]
+      @values[key] ||= []
+      @values[key].push(client.id)
+    end
+
+    protected
+
+    def normalize(str)
+      str ? str.downcase.strip.gsub(/[^a-z0-9]/i, '').presence : nil
+      transliterate ? I18n.transliterate(str) : str
```
eanders: The logic here looks good, but I feel like it'd be easier to read as:
```suggestion
      return nil unless str.present?
      
      str = str.downcase.strip.gsub(/[^a-z0-9]/i, '').presence
      return str unless transliterate
      
      I18n.transliterate(str)
```

app/models/grda_warehouse/client_matcher_lookups/ssn_lookup.rb
```diff
@@ -0,0 +1,43 @@
...
+
+    def get_ids(ssn:)
+      return [] unless valid_ssn?(ssn)
+
+      @values[ssn]&.uniq || []
+    end
+
+    def add(client)
+      key = client.ssn
+      key = key[-4..] if key && format == :last_four
```
eanders: Any reason not to use?
```suggestion
      key = key.last(4) if key && format == :last_four
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -290,58 +268,40 @@ def all_source_clients
...
-      @destination_clients_grouped_by_dob ||= all_destination_clients.group_by { |_, _, _, dob, _| dob }.
-        transform_values { |values| values.map(&:last) }
+    private def build_source_lookups
+      @source_name_lookup = GrdaWarehouse::ClientMatcherLookups::ProperNameLookup.new
+      @source_ssn_lookup = GrdaWarehouse::ClientMatcherLookups::SSNLookup.new
+      @source_dob_lookup = GrdaWarehouse::ClientMatcherLookups::DOBLookup.new
+      @source_clients = []
+
+      clients = GrdaWarehouse::Hud::Client.joins(:warehouse_client_source).source
+      id_field = Arel.sql(wc_t[:destination_id].to_sql)
```
eanders, line 293:
```suggestion
      # Use the Destination Client's ID, but the source client's PII
      id_field = Arel.sql(wc_t[:destination_id].to_sql)
```

app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,142 @@
...
+    [dir, "#{base}-results#{ext}"].compact.join('/')
+  end
+
+  private def build_lookups
+    @name_lookup = GrdaWarehouse::ClientMatcherLookups::ProperNameLookup.new(transliterate: true)
+    @ssn_lookup = GrdaWarehouse::ClientMatcherLookups::SSNLookup.new(format: :last_four)
+    @dob_lookup = GrdaWarehouse::ClientMatcherLookups::DOBLookup.new
+
+    clients = GrdaWarehouse::Hud::Client.all
+    GrdaWarehouse::ClientMatcherLookups::ClientStub.from_scope(clients) do |client|
```
eanders:
```suggestion
    clients = GrdaWarehouse::Hud::Client.joins(:warehouse_client_source).source
    id_field = Arel.sql(wc_t[:destination_id].to_sql)
    GrdaWarehouse::ClientMatcherLookups::ClientStub.from_scope(clients, id_field: id_field) do |client|
```

app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,142 @@
...
+      log("Argument error (possibly encoding related): #{e.message}", object_key: key)
+    rescue StandardError => e
+      log("An unexpected error occurred: #{e.message}", object_key: key)
+    end
+    results
+  end
+
+  def process_row(row)
+    first_name, last_name, ssn4, dob, ghocid = row.values_at('first_name', 'last_name', 'ssn4', 'dob', 'ghocid')
+    return if ghocid.blank?
```
eanders: Can we rename `ghocid` to `external_id` to make this more generic?  We'll need to figure out how to pass that in (maybe from the remote credential?  The Remote Credential will be specific to the installation, so could have a less-generic word like `ghocid`.

app/jobs/identify_external_clients_job.rb
eanders: Hmm, I like the idea of prioritizing matches on 3 fields over 2 fields.  We would need a different approach here.  If I'm reading this correctly, we have:
matches: `[[1,2,3], [2, 3], [3]]`

I think we'd just have to do all the counting, filter out any where `value < required_number`, sort by values, take the one with the highest value.  (I'm open to other approaches, that might be somewhat expensive, though all the arrays _should_ be very small.)

eanders (changes requested):
@ttoomey this looks great!  Thanks.  I think the cleanup of identify duplicates and collection of data into the client matcher lookups classes is a huge improvement.

I noted a couple places I think need adjustment, let me know if you have questions.

app/jobs/identify_external_clients_job.rb
ttoomey: should be done, added quick tests to verify

app/jobs/identify_external_clients_job.rb
ttoomey: added as external_id_field as a param to the job

app/jobs/identify_external_clients_job.rb
```diff
@@ -0,0 +1,152 @@
...
+    match = get_best_match(matches)
+    match ? { external_id_field => external_id, 'Client ID' => match } : nil
+  end
+
+  # matching id is found in at least two match sets
+  def get_best_match(match_sets, min_rank: 2)
+    ranks = Hash.new(0)
+    match_sets.each do |matches|
+      matches.each { |match| ranks[match] += 1 }
+    end
```
eanders: Pretty sure this is the same as:
```suggestion
    ranks =  match_sets.flatten.tally
```


```
irb(main):004:0> match_sets = [[1,2,3], [2, 3], [3]]
=> [[1, 2, 3], [2, 3], [3]]
irb(main):005:0> ranks = Hash.new(0)
=> {}
irb(main):006:1* match_sets.each do |matches|
irb(main):007:1*        matches.each { |match| ranks[match] += 1 }
irb(main):008:0> end
=> [[1, 2, 3], [2, 3], [3]]
irb(main):009:0> ranks
=> {1=>1, 2=>2, 3=>3}
irb(main):010:0> match_sets.flatten.tally
=> {1=>1, 2=>2, 3=>3}
irb(main):011:0>
```

eanders (approved):
Thanks!  One tiny yak to shave (maybe), otherwise, I think we should merge it.

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -13,6 +13,8 @@ def initialize(run_post_processing: true)
       setup_notifier('IdentifyDuplicates')
       @run_post_processing = run_post_processing
       super()
+      build_source_lookups
+      build_destination_lookups
```
gigxz, line 17: The way this job gets run from HMIS is `GrdaWarehouse::Tasks::IdentifyDuplicates.new.delay(queue: :long_running).run!`

This change introduced some heavy computation to the initialize step which should really be part of the queued job.

Is there any reason we can't make this into a job class that inherits from `BaseJob`, so we can queue it normally?

gigxz: @ttoomey @eanders @martha 

Martha pointed out that the HMIS Client after_create/after_update [hooks](https://github.com/greenriver/hmis-warehouse/blob/6f696f489da1a48f926f6841900b85e94490c2a0/drivers/hmis/app/models/hmis/hud/client.rb#L374-L385) run a bunch of computation synchronously locally, which is how I tracked this down.

This is happening in prod too and likely having a significant perf impact. (`GrdaWarehouse::Tasks::IdentifyDuplicates.new` now takes 4 seconds to run on QA).


Issue for tracking https://github.com/open-path/Green-River/issues/6740

gigxz: opened https://github.com/greenriver/hmis-warehouse/pull/4737 as a quick fix.

ttoomey: Thanks so much @gigxz! I somehow missed that we were calling this in lifecycle hooks

PR #4555, opened 2024-07-17
Patch publish form bugs

gigxz:

## Description

Patch 2 bugs:
1. When publishing a form, if the form had any validation errors it would throw an exception (And not show the user the errors)
2. When publishing a SERVICE form, it would create CDEDs with owner_type `Hmis::Hud::HmisService` which is not valid (would not work to save the fields, since it's not the owner_type that the processor is looking for). More details on  https://github.com/open-path/Green-River/issues/6362

## Type of change
- [x] Bug fix


drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -8,7 +8,7 @@ module Mutations
   class PublishFormDefinition < CleanBaseMutation
     argument :id, ID, required: true
 
-    field :form_identifier, Types::Forms::FormIdentifier, null: false
+    field :form_identifier, Types::Forms::FormIdentifier, null: true
```
gigxz, line 11: Fixes https://green-river.sentry.io/issues/5615330060/?project=4504006381273088&query=&referrer=issue-stream&statsPeriod=1h&stream_index=0

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -61,9 +61,13 @@ def add_missing_custom_field_keys(definition)
...
       data_source = GrdaWarehouse::DataSource.hmis.first
-      # FIXME: some definitions support multiple owner_types (SERVICE, NEW_CLIENT_ENROLLMENT).
-      # This needs to be updated to account for that. The mapping.record_type field should specify the record type.
+
+      # FIXME(#6362): For some Definition types (SERVICE, NEW_CLIENT_ENROLLMENT), the desired owner_type may be ambiguous.
+      # This code does not handle that ambiguity.
+      # Instead, it assumes that the CDED "owner_type" should be the same as the definition's "owner_class".
+      # For SERVICE forms, it assumes that the CDED owner should be `CustomService` (as opposed to HUD `Service`).
       owner_type = definition.owner_class.sti_name
+      owner_type = 'Hmis::Hud::CustomService' if owner_type == 'Hmis::Hud::HmisService'
```
gigxz, line 70: Add this workaround, so that we're not generating CDEDs for "HmisService". The "real" fix will be addressed with https://github.com/open-path/Green-River/issues/6362.

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
martha: Hm, should we add a `field :errors` which seems to be our pattern elsewhere?

martha (approved):

drivers/hmis/app/graphql/mutations/publish_form_definition.rb
martha: Oh, I see this has been added by the frontend schema. But don't quite understand how if it was not added here? https://github.com/greenriver/hmis-frontend/pull/851/files#diff-e38721dd872401d491c291485f2e562aac599fcdf386e9284d0e424b38d1c662R59-R61

gigxz: parent class `CleanBaseMutation` sets the `errors` field!

PR #4548, opened 2024-07-15
Update enrollment management system test

gigxz:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

* Update test that was failing because it didn't wait until the enrollment mutation completed before asserting. (new assertion to wait for dialog to closes fixes that).
* Debug logs when running the preview server
* Stop using xpath for hoh change (so we don't depend on order), though it still needs to be improved on the frontend side

## Type of change
- [x] Bug fix

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4562, opened 2024-07-18
5755 run e2e tests in ci

ttoomey:

## Description
[Issue 5755](https://github.com/open-path/Green-River/issues/5755)

* Runs e2e tests in CI
* pulls out common dependencies, dry up repeated deps
* adding caching to speed up CI
* only run graphql schema generation on e2e tests
* update action version to address warnings related to node deprecations 

## Type of change


ttoomey:
@gigxz there's some loose ends here but I think this is in good shape if you want to take a look

config/initializers/active_record.rb
```diff
@@ -1,7 +1,8 @@
-if Rails.env.test?
+if Rails.env.test? && !ENV['RUN_SYSTEM_TESTS']
```
gigxz, line 1: Ahhh this appears to fix https://github.com/open-path/Green-River/issues/6363... great. I am going to make this change on my branch https://github.com/greenriver/hmis-warehouse/pull/4548.

.github/workflows/rails_tests.yml
```diff
@@ -117,36 +126,31 @@ jobs:
...
-          gem install bundler --version=2.5.7
-
       - name: Install gems
         run: |
+          gem install bundler --version=2.5.7
           bundle config set --local without 'production staging development'
           bundle install --jobs 4 --retry 3
 
+      - name: Cache Node.js modules
+        uses: actions/cache@v4
```
gigxz, line 152: So this is introducing cacheing for the warehouse node modules, right? Is it related to the e2e tests somehow, or just incidental improvement?

Just wondering because I have had issues with the actions cache for npm modules in the past, so wondering if its actually entangled to this change or can be considered separately.

.github/workflows/rails_tests.yml
```diff
@@ -164,10 +168,22 @@ jobs:
...
+          restore-keys: |
+            ${{ runner.os }}-assets-
+
+      - name: Precompile assets
+        # if: steps.cache-assets.outputs.cache-hit != 'true'
+        run: bundle exec rails assets:precompile
 
       # Regenerate the graphql schema and fail if it has changed
       - name: 'Check HMIS GraphQL schema'
+        if: matrix.test_group == 'system_tests'
```
gigxz, line 184: Why does this step need to be tied to a test group? Would it run repeatedly if we didn't have this check?

.github/workflows/rails_tests.yml
```diff
@@ -117,36 +126,31 @@ jobs:
...
-            curl libcurl curl-dev \
-            imagemagick \
-            libmagic file-dev file \
-            build-base libxml2-dev libxslt-dev postgresql-dev \
-            libgcc libstdc++ libx11 glib libxrender libxext libintl ttf-dejavu ttf-droid ttf-freefont ttf-liberation ttf-freefont \
-            chromium nss freetype freetype-dev harfbuzz ca-certificates ttf-freefont \
-            lftp postgresql tmux postgis geos geos-dev \
-            shared-mime-info gpg gpg-agent
-
+          apk add --no-cache $(cat .github/dependencies.txt)
```
gigxz, line 132: This seems alright, minor concern about (1) making it harder to find the git history around why these things were added, and (2) no longer able to add comments alongside each dependency (not that we were doing that :) )

.github/workflows/rails_tests.yml
ttoomey: There's no dependency on the e2e tests, we could take it out if it's too much. I was just taking the opportunity to improve performance since my head was in there.

This is also only changing the warehouse assets. It's not related to the front end

.github/workflows/rails_tests.yml
ttoomey: I could back this out. It's just that without the condition it runs for each element in the matrix which doesn't seem useful. It's not expensive though, we could leave it as is.

.github/workflows/rails_tests.yml
ttoomey: These deps are repeated in several of the workflow files, my plan was to DRY up those files too but I could back this out if that's a concern. Also, having the deps in a file lets generate a checksum and cache the apks to avoid needing to download them on every run

.github/workflows/rails_tests.yml
gigxz: Okay gotcha, thanks for clarifying. This seems like an improvement overall so I say lets keep it!  thanks

.github/workflows/rails_tests.yml
eanders: Given that each dependency is now on it's own line, could we not add a bash style comment if we wanted to for any line?  I'm unclear how apk add would handle that, but I'm sure we could put some awk/sed in place to strip them if we wanted the. I like that all of the dependencies will stay in sync, so at least that's a win, and they ARE still in git, so we could have commit comments if nothing else.

.github/workflows/rails_tests.yml
eanders: It seems somewhat odd to group the system tests into this file if were just adding if statements for all the test sections to decide if it should run or not, maybe it should just be it's own file?

eanders (approved):
This looks generally good to me. Dropped a couple questions inline, but I don't have strong feelings about either one right now.  Thanks @ttoomey !

.github/workflows/rails_tests.yml
ttoomey: Since we don't actually have any comments I was going to leave it as-is until it's a need. But we could use sed/awk or just make it a script itself (`sh ./github/install-deps.sh`)

.github/workflows/rails_tests.yml
ttoomey: We could make additional workflow files. I used conditional steps because it seemed more DRY, otherwise would copy/paste the entire workflow.

However there is something called a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows). I haven't tried this before but it seems like it could work. We'd probably extract the common config from our current workflows. Maybe worth doing but I'd suggest we tackle it in another feature

.github/workflows/rails_tests.yml
ttoomey: [tracking issue here](https://github.com/open-path/Green-River/issues/6409)

.github/workflows/rails_tests.yml
ttoomey: [tracking issue here](https://github.com/open-path/Green-River/issues/6409)

PR #4563, opened 2024-07-19
Retain MCI Unique ID on merge (handle edge case)

gigxz:
## Description

https://github.com/open-path/Green-River/issues/6384

Handle the following Scenario:

* Client A has no mci unique id
* Client B does have an mci unique id
* In the UI, use manually performs merge for Client A and Client B
Client A is retained, and Client B is destroyed.
  * Previous behavior: do not transfer the MCI Unique ID over to the retained client. Client A ends up with no MCI Unique Id.
  * Proposed behavior: transfer over the MCI Unique ID to the retained client

**This scenario is unlikely to occur**, because the merge process chooses the oldest client to keep. Typically its only newer clients that don't have mci unique Ids yet, because the nightly api job hasn't run. Despite that fact, I think it's safer to address this logic in the merge job anyway.

## Type of change
- [x] Bug fix


drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -223,6 +224,27 @@ def merge_mci_ids
...
+      return if client_to_retain.ac_hmis_mci_unique_id.present?
+
+      # If retained client does not have an MCI Unique ID, try to find one to keep from the merged clients
+      mci_unique_id_to_keep = HmisExternalApis::ExternalId.mci_unique_ids.
+        where(source: clients_needing_reference_updates).
+        max_by(&:updated_at)
+      return unless mci_unique_id_to_keep
+
+      # Re-assign this MCI Unique ID to the retained client
+      mci_unique_id_to_keep.update!(source: client_to_retain)
```
eanders, line 245: Is there any argument here for keeping the old ExternalID and creating a new one for the new client (with the old MCI ID) or is there no reason to retain that history?

eanders (approved):
Thanks @gigxz looks great!

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
gigxz: I don't think that's necessary- this here matches the behavior for MCI ID so I think better to keep them consistent

PR #4564, opened 2024-07-19
Implement staff assignment for Caseload

martha:

## Description
Ticket: https://github.com/open-path/Green-River/issues/6335
Spec doc: https://docs.google.com/document/d/17o9wMCrYR1Wo4vX5iLUqYMe1b7pZQjDdhFCQx6_T5-M/edit

- Adds a migration to create the tables indicated in the spec doc
- Adds new models and schema objects for the new types
- Adds a new project config type for indicating whether projects have staff assignment enabled
- Adds assign/unassign mutations

To test: It's probably easiest to test this in conjunction with the frontend changes, which are still a draft, but I've pushed what I have so far here: https://github.com/greenriver/hmis-frontend/pull/856

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -144,6 +148,22 @@ def self.static_options_for_type(pick_list_type, user:)
       end
     end
 
+    def self.eligible_staff_assignment_user_picklist(current_user)
+      return [] unless current_user
+
+      # Check that the user has can_edit_enrollments permission on at least one project with staff_assignment_enabled
```
martha: to discuss: Is this permission check correct / is it overkill?

drivers/hmis/app/graphql/mutations/assign_staff.rb
```diff
@@ -0,0 +1,42 @@
...
+        staff_assignment_type: assignment_type,
+      )
+
+      if assignment.valid?
+        assignment.save!
+        { staff_assignment: assignment }
+      else
+        errors = HmisErrors::Errors.new
+        errors.add_ar_errors(assignment.errors&.errors)
+        { errors: errors }
```
gigxz: Are these expected to be "fixable" validation errors? If not, we should raise instead

drivers/hmis/app/graphql/mutations/unassign_staff.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    field :staff_assignment, Types::HmisSchema::StaffAssignment, null: true
+
+    def resolve(id:)
+      record = Hmis::StaffAssignment.preload(household: :project).find_by(id: id)
+      raise 'Not found' unless record
+
+      access_denied! unless current_user.permissions_for?(record.household.project, :can_edit_enrollments)
+
+      default_delete_record(record: record, field_name: :staff_assignment)
```
gigxz: I don't think this is giving us much, imo clearer just to write out the `destroy!`

drivers/hmis/app/models/hmis/staff_assignment.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class Hmis::StaffAssignment < Hmis::HmisBase
+  acts_as_paranoid
+
+  belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
+  belongs_to :user, class_name: 'Hmis::User'
+  belongs_to :household, class_name: 'Hmis::Hud::Household', primary_key: [:data_source_id, :HouseholdID], foreign_key: [:data_source_id, :household_id], inverse_of: :staff_assignments, optional: true
+  belongs_to :staff_assignment_type, foreign_key: :hmis_staff_assignment_type_id, class_name: 'Hmis::StaffAssignmentType'
+
+  validates :household_id, uniqueness: { scope: [:user_id, :data_source_id, :hmis_staff_assignment_type_id] }
```
martha: If the user tries to submit a non-unique combination of [household, user, assignment type], the error message they see is "Household has already been taken." This isn't the most intuitive error message - the problem is not specifically with household, but with the combo of the 3 fields. There's probably a better railsy way to do this, maybe a way to apply a validation that is not tied to a specific field? To discuss/investigate docs more

drivers/hmis/app/graphql/types/hmis_schema/household.rb
```diff
@@ -46,5 +49,16 @@ def enrollments
     def assessments(**args)
       resolve_assessments(**args)
     end
+
+    def staff_assignments(is_currently_assigned: true)
```
martha, line 53: I haven't validated any permissions here. Is it safe to assume that anyone who has view permission on the household is fine to view the caseload info about that household, or should this be more restricted?

db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb
```diff
@@ -0,0 +1,29 @@
...
+
+class AddStaffAssignmentTables < ActiveRecord::Migration[7.0]
+  def change
+    create_table(:hmis_staff_assignment_types) do |t|
+      t.string :name, null: false, comment: 'name of role, such as "Case Manager" or "Housing Navigator"', index: { unique: true }
+      t.timestamps
+      t.timestamp :deleted_at
+    end
+    create_table(:hmis_staff_assignments) do |t|
+      # Actually setting the foreign key to the User table caused an exception in rspec when truncating tables from
```
martha: I'll remove this comment before merging - let's discuss

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: This pick list type should accept project as an argument, and the returned list should only include users with access to edit this enrollment.

This means we can get rid of the with_staff_assignments_enabled scope

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -207,6 +208,10 @@ def incoming_referral_postings(**args)
       )
     end
 
+    def staff_assignments_enabled
+      load_ar_association(object, :project_staff_assignment_configs).exists?
```
martha: We don't really need the data loader here, since this isn't called in batch. Call the class method and add a comment that indicates it shouldn't be used in batch

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -247,6 +252,10 @@ def data_collection_features
     end.compact
   end
 
+  def staff_assignments_enabled?
+    Hmis::ProjectStaffAssignmentConfig.where(project: self).exists?
```
martha: This should use `detect_best_config_for_project`

drivers/hmis/app/graphql/types/hmis_schema/household.rb
martha: discussed, fine to leave as is

martha: add pagination (assume users will add more data than we expect)
for the frontend card of current assignees, we don't have a design for pagination, so just limit the query to 100

drivers/hmis/app/models/hmis/staff_assignment.rb
martha: remove this validation (it's already checked on the db), instead add a check to the mutation where we can customize the error message

drivers/hmis/app/graphql/mutations/assign_staff.rb
martha: Once we move the uniqueness check in here (above), we should just get rid of the .valid? check, since there is no other user-resolvable error possible

martha:
@ttoomey, I would still like to put some time into [this todo](https://github.com/greenriver/hmis-warehouse/pull/4564/files#diff-46d7e0c7dbbcaec4b6c665a361a4af72207a48232a9aa1e844cb98995f852202R93), but otherwise would be very grateful for your review! thanks!

ttoomey:
> @ttoomey, I would still like to put some time into [this todo] but otherwise would be very grateful for your review! thanks!

Sorry I missed this PR request, thanks for the nudge. I'll get it reviewed asap!

drivers/hmis/app/graphql/mutations/assign_staff.rb
```diff
@@ -0,0 +1,45 @@
...
+module Mutations
+  class AssignStaff < CleanBaseMutation
+    argument :household_id, ID, required: true
+    argument :assignment_type_id, ID, required: true
+    argument :user_id, ID, required: true
+
+    field :staff_assignment, Types::HmisSchema::StaffAssignment, null: true
+
+    def resolve(household_id:, assignment_type_id:, user_id:)
+      household = Hmis::Hud::Household.preload(:project).find_by(household_id: household_id, data_source_id: current_user.hmis_data_source_id)
```
ttoomey: Since we only load a single record, I think the preload isn't needed. Also probably should use the viewable_by scope so we never even load a record the user can't see
```suggestion
      household = Hmis::Hud::Household.viewable_by(current_user).find_by(household_id: household_id, data_source_id: current_user.hmis_data_source_id)
```

drivers/hmis/app/graphql/mutations/assign_staff.rb
```diff
@@ -0,0 +1,45 @@
...
+      errors.add(:user_id, :invalid, message: "#{user.name} is already assigned as #{assignment_type.name} for this household") if existing.exists?
+      return { errors: errors } if errors.any?
+
+      assignment = Hmis::StaffAssignment.new(
+        household: household,
+        user: user,
+        staff_assignment_type: assignment_type,
+      )
+
+      assignment.save!
```
ttoomey: This is a small thing but enclosing the check in a lock forces duplicate requests to run in series so the the behavior is consistent (avoids race condition). 

Also, it's a little more concise to call `create!` rather than new, save!

```suggestion
      # Locking on assignment_type looks odd but we can't use user (different db) and household is a view
      assignment_type.with_lock do
        errors.add(:user_id, :invalid, message: "#{user.name} is already assigned as #{assignment_type.name} for this household") if existing.exists?
        return { errors: errors } if errors.any?

        assignment = Hmis::StaffAssignment.create!(
          household: household,
          user: user,
          staff_assignment_type: assignment_type,
        )
      end
```

drivers/hmis/app/graphql/mutations/assign_staff.rb
```diff
@@ -0,0 +1,45 @@
...
+    def resolve(household_id:, assignment_type_id:, user_id:)
+      household = Hmis::Hud::Household.preload(:project).find_by(household_id: household_id, data_source_id: current_user.hmis_data_source_id)
+      raise 'Not found' unless household
+
+      access_denied! unless current_user.permissions_for?(household.project, :can_edit_enrollments)
+
+      raise 'Staff Assignment not enabled' unless household.project.staff_assignments_enabled?
+
+      assignment_type = Hmis::StaffAssignmentType.find(assignment_type_id)
+      user = Hmis::User.find(user_id)
```
ttoomey: It seems like it should mirror what we have in the picklist?
```suggestion
      user = Hmis::User.can_edit_enrollments_for(household.project).find(user_id)
```

drivers/hmis/app/graphql/mutations/assign_staff.rb
```diff
@@ -0,0 +1,45 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Mutations
+  class AssignStaff < CleanBaseMutation
+    argument :household_id, ID, required: true
+    argument :assignment_type_id, ID, required: true
+    argument :user_id, ID, required: true
```
ttoomey: Perhaps this should be an input type? Input types tend to be easier to work with imo but perhaps this mutation is so simple it's not needed
```ruby
    argument :input, Types::HmisSchema::AssignStaffInput, required: true
```

db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb
```diff
@@ -0,0 +1,28 @@
...
+      # hmis_households is not actually a table, but a view; the source for HouseholdID is the Enrollments table,
+      # which is the reason to not use t.references here.
+      t.string :household_id
+
+      t.references :hmis_staff_assignment_type, null: false, foreign_key: { to_table: :hmis_staff_assignment_types }, index: false
+      t.references :data_source, null: false
+      t.timestamps
+      t.timestamp :deleted_at
+    end
+    add_index :hmis_staff_assignments, [:user_id, :hmis_staff_assignment_type_id, :household_id, :data_source_id], unique: true, where: 'deleted_at IS NULL', name: :uidx_hmis_staff_assignments
```
ttoomey: It seems like the two queries where we might want indexes for are user.staff_assignments and household.staff_assignments.

Postgres can only use a composite index from left to right, so it can't be used for the household query

Perhaps this we could reorder it and then add another index for the user_id:
```suggestion
    add_index :hmis_staff_assignments, [:data_source_id, :household_id, :user_id, :hmis_staff_assignment_type_id], unique: true, where: 'deleted_at IS NULL', name: :uidx_hmis_staff_assignments
    add_index :hmis_staff_assignments, :user_id
```

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -144,6 +148,17 @@ def self.static_options_for_type(pick_list_type, user:)
...
 
+    def self.eligible_staff_assignment_user_picklist(project)
+      return [] unless project&.staff_assignments_enabled?
+
+      Hmis::User.can_edit_enrollments_for(project).map do |user|
+        {
+          code: user.id.to_s,
+          label: user.full_name,
+        }
+      end.sort_by { |obj| obj[:code] }
```
ttoomey: if this is display order for the ui, perhaps order by last, first? or fullname?
`      Hmis::User.can_edit_enrollments_for(project).order(:last_name, :first_name,:id).map do |user|`

martha:
No rush @ttoomey I just requested your review yesterday -- it's been in draft mode! :)

drivers/hmis/app/graphql/mutations/unassign_staff.rb
```diff
@@ -0,0 +1,26 @@
...
+
+module Mutations
+  class UnassignStaff < CleanBaseMutation
+    argument :id, ID, required: true
+
+    field :staff_assignment, Types::HmisSchema::StaffAssignment, null: true
+
+    def resolve(id:)
+      record = Hmis::StaffAssignment.preload(household: :project).find_by(id: id)
+      raise 'Not found' unless record
```
ttoomey: This is fine; looks like we have the convention of raising a runtime error in this case. Some day we should improve this though, probably should just use scope.find() which raises ActiveRecord::RecordNotFound. Then handle that error generically so we get consistent behavior across the schema. Something for future housekeeping 🧹

drivers/hmis/spec/factories/hmis/hud/staff_assignments.rb
```diff
@@ -0,0 +1,23 @@
...
+    transient do
+      enrollment { association :hmis_hud_enrollment, data_source: data_source }
+    end
+    after(:build) do |record, evaluator|
+      record.household = evaluator.enrollment.household
+    end
+  end
+
+  factory :hmis_staff_assignment_type, class: 'Hmis::StaffAssignmentType' do
+    name { 'Case Manager' }
```
ttoomey: perhaps this should be a sequence so the factory can make multiple instances without violating the uniqueness constraint on name
```suggestion
    sequence(:name) { |n| "staff role #{n}" }
````

File: drivers/hmis/app/graphql/types/hmis_schema/household.rb [hunk 17]
```diff
@@ -46,5 +49,16 @@ def enrollments
...
+      # There's no current use case for returning all (both currently assigned and formerly assigned)
+      # in the same query, but we could update this to support that use case if it arises.
+      scope = load_ar_association(object, :staff_assignments).order(created_at: :desc)
+      if is_currently_assigned
+        scope
+      else
+        scope.with_deleted.where.not(deleted_at: nil).order(created_at: :desc, deleted_at: :desc)
```
ttoomey: small suggestion to use id rather than deleted_at as a tie breaker
```suggestion
      scope = load_ar_association(object, :staff_assignments).order(created_at: :desc, id: desc)
      if is_currently_assigned
        scope
      else
        scope.with_deleted.where.not(deleted_at: nil)
```

drivers/hmis/app/graphql/types/hmis_schema/staff_assignment.rb
```diff
@@ -0,0 +1,39 @@
...
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::StaffAssignment < Types::BaseObject
+    description 'Staff Assignment'
+    field :id, ID, null: false
+    field :user, Application::User, null: false
+    field :household, HmisSchema::Household, null: false
+    field :staff_assignment_type, String, null: false
```
ttoomey: Maybe this should be called something like "staff_assignment_name" or maybe just "role_name". In the context of a resolver, "type" suggests to me that this could be an enum

db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb
```diff
@@ -0,0 +1,28 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class AddStaffAssignmentTables < ActiveRecord::Migration[7.0]
+  def change
+    create_table(:hmis_staff_assignment_types) do |t|
```
ttoomey: I wonder if this would be better named "hmis_staff_assignment_roles" to match the inline comment. Also "type" is a somewhat overloaded term.

drivers/hmis/app/graphql/mutations/unassign_staff.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      access_denied! unless current_user.permissions_for?(record.household.project, :can_edit_enrollments)
```
ttoomey: I think it might be better to standardize on the `current_permission?` helper

```suggestion
      access_denied! unless current_permission?(permission: :can_edit_enrollments, entity: record.household.project)
```

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -88,6 +88,11 @@ def load_effective_permissions
       joins(:roles).
         where(roles: { permission => true })
     end
+
+    scope "#{permission}_for", ->(_entity) do
+      # todo @martha - we want to query for only users that have permission for this entity, but it's complicated
+      joins(:roles).where(roles: { permission => true }).group(:id)
```
ttoomey: it might be worth asking Elliot about this, we could well have something already implemented on the warehouse side, which uses similar models

ttoomey (approved):
This is looking great overall, thanks Martha! 

I left a bunch of suggestions that are all optional.

db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb
martha: thanks @ttoomey this is great 🎉

drivers/hmis/app/graphql/mutations/unassign_staff.rb
martha: Oh true, I'm happy to simplify it to use `find` now. And later we can handle that error generically, as you said.

drivers/hmis/spec/factories/hmis/hud/staff_assignments.rb
martha: Great 👍

drivers/hmis/app/graphql/types/hmis_schema/staff_assignment.rb
martha: I agree with you about "type" implying an enum, but I also remember Gig had concerns about "role" being overloaded too because it sounds like a permissions thing. "staff_assignment_title" (like job title?) is the only other good alternative that I have been able to come up with so far, but I feel hesitant to change things around in the spec that we agreed on before Gig left

drivers/hmis/app/models/hmis/user.rb
martha: @eanders , do you know if we do anything similar to this in the warehouse? From what I could find, not yet:
https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/app/models/user.rb#L73-L78
is the most similar, but doesn't include a scope that checks for all users with permission *for* a particular entity.

eanders: We usually do the check on the entity `scope :viewable_by, -> (user)`.  (See `app/models/grda_warehouse/hud/project.rb#viewable_by` for one example.)

The trick with going the other way is that you have to walk the inheritance chain, users can inherit access to a project from the project's organization for instance.

I think this would be something like (untested code):
```
scope "#{permission}_for", ->(entity) do
 user_ids = Hmis::AccessControl.joins(:role, :access_group, user_group: :users).
    preload(user_group: :user_group_members).
    merge(Hmis::Role.where(permission => true)).
    merge(Hmis::AccessGroup.contains(entity)). # I think this `contains` scope isn't complete since it doesn't walk the inheritance tree (but maybe it should?)
    select(Hmis::User.arel_table[:id]) # selecting the ids so we can ensure the returned scope doesn't include any of the complexity 
  User.where(id: user_ids)
end

```

Looking through the code, I think it would be helpful to add a `contains_with_inherited` scope on `Hmis::AccessGroup`.  That will be a bit of a pain to write, but should make the above useful.

martha: Thanks @eanders this is extremely helpful. Does this look like I am on the right track?
https://github.com/greenriver/hmis-warehouse/pull/4564/files#diff-c4a75867ebfe78fba438a93af5a60cddcd44463aaa2be29eac798cc3902d30a0R49-R62

- Do you think it is OK to leave the other entity types as 'not implemented' for now until a use case arises?

eanders: Hmm, so I think I must not have been very clear.  (Happy to huddle about it if that'd be easier.)  What you have still doesn't include the inheritance (as far as I can see.)  When checking for projects, you need to include the organizations and data sources that also contain that project since permission may be granted further up the chain.  

I'll drop a comment on that scope to see if I can get closer.

drivers/hmis/app/models/hmis/access_group.rb
```diff
@@ -46,6 +46,20 @@ class Hmis::AccessGroup < ApplicationRecord
...
+    when Hmis::Hud::Project
+      collection_ids = Hmis::Hud::Project.
+        joins(:group_viewable_entities).
+        where(id: entity.id).
+        pluck('hmis_group_viewable_entities.collection_id')
+      where(id: collection_ids)
+    else
+      # todo @martha
+      raise 'not implemented'
+    end
```
eanders:
```suggestion
  additional_entities = case entity
  when Hmis::Hud::Project
    {
      'Hmis::Hud::Organization' => entity.organization.id, 
      'GrdaWarehouse::DataSource' => entity.data_source.id,
    } 
  when Hmis::Hud::Organization
    {
      'GrdaWarehouse::DataSource' => entity.data_source.id,
    }
  when Hmis::DataSource
    {}
  else
    raise "Unsupported entity type: #{entity.class.name}"
  end
  a_t = Hmis::GroupViewableEntity.arel_table
  # Find the collections involving this entity directly
  query = a_t[:entity_type].eq(entity.class.name).and(a_t[:entity_id].eq(entity.id))
  # Add collections where this entity is a sub-entity of the assigned entity
  additional_entities.each do |klass, id|
    query = query.or(a_t[:entity_type].eq(klass).and(a_t[:entity_id].eq(id)))
  end
  # Fetch the collections that include this entity, either directly, or through inehritance
  collection_ids = Hmis::GroupViewableEntity.where(query).pluck(:collection_id)
  # Return a clean AccessGroup scope
  where(id: collection_ids)
```

martha: Thanks @eanders this makes sense and is very thorough, I'm happy to commit it, but I think maybe some of this work of finding the entities through inheritance is being done already through the `group_viewable_entity_projects` view? I think that's why the code in [ProjectAccessLoader](https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/drivers/hmis/app/models/hmis/hud/project_access_loader.rb#L16-L20) works, which is what my sketch was based on, even though it does not explicitly make queries for data source and organization based collections.

eanders: Oh!  Sorry, I'm not that familiar with the HMIS permission scheme, it was, at one point, very similar to the warehouse version, but seems to have matured.  Yes, `hmis_group_viewable_entity_projects` looks to include the same relationships I added.  Do you (or maybe @ttoomey) know how to take advantage of that table?

It looks like `Hmis::Hud::Project.joins(:group_viewable_entity_projects)` might be the ticket, but I'm definitely not following that completely.  Maybe you had the project version covered already.  The Data Sources are simple, but maybe we should build a new view for organizations?  (Or maybe we only need to determine which projects someone has access to, which might mean what you had was sufficient.)

martha: Hm, ok, yes I think you're right that `.joins(:group_viewable_entity_projects)` is the ticket, because of how this relationship is set up on the [Project model](https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/drivers/hmis/app/models/hmis/hud/project.rb#L46) with `has_many :group_viewable_entities, through: :group_viewable_entity_projects`.

I also discovered that it looks like there are already helpers for `includes_entities` with inheritance on the [GroupViewableEntity](https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/drivers/hmis/app/models/hmis/group_viewable_entity.rb#L53-L55), though they seemed to be a bit out-of-date and untested, so I fixed one or two bugs and added some spec tests to reassure us of the behavior.

martha: @ttoomey and/or @eanders (no rush) I've re-requested your review for this part specifically -- I'd be very glad for either one of you to give it a second pair of eyes before I merge, since I haven't done that much work on permissions yet.

db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb
eanders: @martha are these actually "relationships"?  I feel like both `roles` and `types` are heavily used, but if this is describing the relationship between the client and the staff member, something like **Case Manager** or **Housing Navigator**, I'd maybe go with `hmis_staff_assignment_relationships`, though `hmis_staff_assignment_relationship_types` also feels right to me...  Maybe just `hmis_staff_relationships`.

drivers/hmis/app/graphql/types/hmis_schema/staff_assignment.rb
eanders: Feel free to ignore my suggestion about `relationship` above (and here), but that's the other wording that feels vaguely right to me.

eanders (approved):
Looks good to me.  Thanks @martha !

drivers/hmis/app/graphql/types/hmis_schema/staff_assignment.rb
martha: Thanks, hm I like `relationship` too! I'll go ahead with the change

PR #4565, opened 2024-07-21
6397 rebalance test matrix

ttoomey:

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6397)

* should get our CI runs down to around 45 min.
* requires occasional manual profiling and re bucketing
* note, this requires adding some tags to our specs that will be managed automatically. Not every spec is tagged, only the slow ones (> 10 secs). Faster tests are all run in a default bucket together
* This surfaced some test failures which may be pre-existing issues in the tests related to missing HMIS forms. I think we see them here because the test order has changed. 

Rebalancing procedure:
1) run rspec with profiling enabled, output results to json (maybe this should happen on every run?)
2) download profiles from GH artifacts into rspec_profiles directory
3) run `rails ci:build_bucket_assignments[/tmp/rspec_profiles]` which updates `.github/rspec_buckets.json`
4) if the rake task added new buckets, add them to the github text matrix

## Type of change


drivers/hmis/spec/requests/hmis/form_definition_rules_spec.rb
```diff
@@ -11,6 +11,10 @@
 RSpec.describe Hmis::GraphqlController, type: :request do
   include_context 'hmis base setup'
 
+  before(:all) do
+    ::HmisUtil::JsonForms.seed_all
```
ttoomey, line 15: this started failing, seems like an order-dependent issue

eanders: Does it make sense to move this into `hmis base setup`?

lib/tasks/ci.rake
```diff
@@ -0,0 +1,97 @@
...
+    profile_groups = profile_groups.shuffle(random: rng)
+
+    # Categorize specs into buckets
+    profile_groups.each do |group|
+      total_time = group[:total_time]
+
+      # Skip specs that run for less than 10 seconds
+      next if total_time < 10
+
+      if current_bucket[:total_time] + total_time > (60 * 40)
```
eanders: is `60 * 40` 40 minutes?  Might be worth adding a comment or some variables to indicate what this is checking.

eanders (commented):
@ttoomey I _think_ this looks like a good approach.  
In some ways I'd prefer to drop the lists of tests per bucket into some sort of text or yaml file so we don't need to make code changes to tests and then guess/check/adjust those as new tests are added, but what you've written seems pretty light weight for now.

Happy to give this a whirl and revisit in a month or two.

drivers/hmis/spec/requests/hmis/form_definition_rules_spec.rb
ttoomey: Yes, I think so. I'll give it a try and and see what breaks

ttoomey: hmm that looks like a bigger change actually since there's like 90 tests that use that. I might hold off on that for now

ttoomey:
> @ttoomey I _think_ this looks like a good approach. In some ways I'd prefer to drop the lists of tests per bucket into some sort of text or yaml file so we don't need to make code changes to tests and then guess/check/adjust those as new tests are added, but what you've written seems pretty light weight for now.

Thanks @eanders - I changed the approach so that we don't need to commit the tags. Now it keeps the groups in a json file (.github/rspec_buckets.json) and modifies the tests in the CI container right before running them. Still plenty of room for improvement though

I also adjusted the bucket approach to try and get the runtime down closer to 30 minutes.

app/models/aws_s3.rb
```diff
@@ -148,11 +148,11 @@ def put(file_name:, prefix:)
...
-    if Rails.env.development? || Rails.env.test?
-      obj.put(body: content, content_type: content_type)
-    else
-      obj.put(body: content, server_side_encryption: 'AES256', content_type: content_type)
-    end
+    args = { body: content }
+    args.merge!(content_type: content_type) if content_type
+    # TODO: document why we skip encryption here
+    args.merge!(server_side_encryption: 'AES256') unless Rails.env.development? || Rails.env.test?
+    obj.put(**args)
```
ttoomey: @eanders 
This addresses a failure in `rspec ./spec/models/aws_s3_spec.rb:48` which was added [in this pr](https://github.com/greenriver/hmis-warehouse/pull/4553/files#diff-ded50fd5f0dc4498ae5af0ac5b785601d5319da6b6be9862628d235917ed1a3eR154)

app/models/aws_s3.rb
```diff
@@ -148,11 +148,11 @@ def put(file_name:, prefix:)
...
   def store(content:, name:, content_type: nil)
     obj = @bucket.object(name)
-    if Rails.env.development? || Rails.env.test?
-      obj.put(body: content, content_type: content_type)
-    else
-      obj.put(body: content, server_side_encryption: 'AES256', content_type: content_type)
-    end
+    args = { body: content }
+    args.merge!(content_type: content_type) if content_type
+    # TODO: document why we skip encryption here
```
eanders: So, I believe we're skipping server side encryption for test and development because minio didn't (or doesn't) support it, or it was fairly complex to setup and didn't seem necessary for local development or testing.  Might be better to inspect the hostname, and ensure it's enabled for real s3 urls.

app/models/aws_s3.rb
ttoomey: [extracted this fix to pr](https://github.com/greenriver/hmis-warehouse/pull/4572)

eanders (approved):
Pending verification of test counts

ttoomey:
I compared between the two branches. The total time is very similar but it looks like there's a small discrepancy, there are 4 more examples counted in release-125 than in this branch for some reason. @eanders - do you think I should try and track that down or is it close enough?

I could re-run profiling on r125 and this branch and diff the specs to track it down, might take an hour or so. Note, the last time I ran profiling, we had 5,148 total examples.

### This branch with new buckets
| bucket  | minutes | examples |
|---------|---------|-------|
| 1       | 30      | 903   |
| 2       | 29      | 691   |
| 3       | 26      | 131   |
| 4       | 30      | 906   |
| default | 20      | 2523  |
| system  | 3       | 10    |
| okta    | 1       | 21    |
| **Total** | **139**  | **5185** |

### release-125
| bucket | minutes | examples |
|--------|---------|-------|
| a-g    | 48      | 1265  |
| h-i    | 68      | 2512  |
| j-z    | 2       | 19    |
| spec   | 12      | 1362  |
| system | 3       | 10    |
| okta   | 0       | 21    |
| **Total** | **133**  | **5189** |

ttoomey:
the four tests were from this spec file
drivers/hmis/lib/spec/fix_enrollment_dates_spec.rb

It's outside of the expected directory pattern for specs so it's excluded. I think that's okay, it looks like it was a test for a one-time data fix

PR #4578, opened 2024-07-29
Add query to support User Dashboard for Caseload

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6337

This PR's base branch is the open PR for [#6335](https://github.com/open-path/Green-River/issues/6335) for now. It should be retargeted to the release branch after that one is merged.

To test this PR, it's probably easiest in conjunction with the frontend PR for the user dashboard: https://github.com/greenriver/hmis-frontend/pull/859

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -99,5 +102,16 @@ def enrollment_access_summaries(filters: nil)
           project_ids: filters&.project,
         )
     end
+
+    def staff_assignments
+      load_ar_association(object, :staff_assignments).
```
ttoomey: The reason to use data loader is to avoid n+1 queries. However since there's additional sql performed on the loaded scope, we have n+1 anyway (actually it's probably doing extra queries, since the resolved records are discarded)

This could be a performance concern for queries that resolve staff_assignments on a collection of users. But we are probably only resolving this on one user at a time in practice. So the simplest thing to do would probably be to just skip using the data loader in this case.

```suggestion
      # n+1 queries, needs optimization to be performant on collections
      object.staff_assignments.
```

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -9,6 +9,7 @@
 module Types
   class Application::User < Types::BaseObject
     include Types::HmisSchema::HasAuditHistory
+    include Hmis::Concerns::HmisArelHelper
```
ttoomey: This is fine, we do this in lots of places in the application.

However I think it would be better to use composition if possible. Suggest adding an arel helper to `GraphqlApplicationHelper`
```
  def arel
    Hmis::ArelHelper.instance
  end
```

Then adjust how we call these arel table aliases: s/hh_t/arel.hh_t/

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -99,5 +102,16 @@ def enrollment_access_summaries(filters: nil)
           project_ids: filters&.project,
         )
     end
+
+    def staff_assignments
```
ttoomey, line 105: do we need a permissions check here or is it assumed that if we can resolve the user, we have permission to resolve staff assignments too?

ttoomey (approved):
Looks good! I left a few suggestions. Happy to discuss the data loading issue in more detail

drivers/hmis/app/graphql/types/application/user.rb
martha: That makes sense, thank you for the explanation!

drivers/hmis/app/graphql/types/application/user.rb
martha: Thanks for the suggestion Theron, can you elaborate more on why this is better - is it for clarity of code or for another reason? In any case, I'm happy to implement your suggestion here if that's the pattern we should standardize around going forward.

https://github.com/greenriver/hmis-warehouse/pull/4578/commits/542ea94671f2108e5b6ce4c352ac3218fe4e41a6

drivers/hmis/app/graphql/types/application/user.rb
martha: Hm, that's a good question. If we were to add a permission check here, I think there's an easy way and a harder way:
- easy: You can only resolve this if you are the user making the request. This is more restrictive than necessary but would work, since currently this query is only used for populating "My Dashboard"
- harder: You can resolve all of the staff assignments for this user related to enrollments that you have permission to view. (We assume that if you are the current user then you have permission to view all of your assignees' enrollments)

I couldn't find anything about this in the [spec document](https://docs.google.com/document/d/17o9wMCrYR1Wo4vX5iLUqYMe1b7pZQjDdhFCQx6_T5-M/edit#heading=h.3trpwowucmf2) or the ticket, so I'll check in with Mary Ann

martha: Hm, on a little more investigation, the user [query](https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L321) is already gated on what I think is the appropriate permission here. You can resolve a user if it's yourself, or if you have "can audit users" or "can impersonate users" permission.

edit: And, I think we have made the decision before that we are OK with the "can audit users" permission granting permission to view client details that the user wouldn't otherwise have access to, since this is an admin level permission. https://github.com/greenriver/hmis-warehouse/blob/9755f5b77568378ef4f160713db8579848c4d63d/drivers/hmis/app/models/hmis/role.rb#L187

drivers/hmis/app/graphql/types/application/user.rb
ttoomey: Thanks Martha, good question! In this case, with our arel helpers, we just want access to a set of constants that define table names. By using an arel helper singleton, we avoid polluting the caller's namespace with these constants, reducing the changes of name collision (of particular concern with short, non-descriptive vars like 'hh_t')

More generally there's something of an adage in OOP: "favor composition over inheritance". It's not super relevant here but the idea is to encourage designs that compose behavior using smaller, loosely coupled components rather than mixing behavior together into one big class using inheritance or modules. Of course this is just a guideline, modules are frequently the most pragmatic approach

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -99,5 +101,17 @@ def enrollment_access_summaries(filters: nil)
...
+      # n+1 queries, needs optimization to be performant on collections
+      object.staff_assignments.
+        joins(:household).
```
ttoomey: I did a little testing locally and it seems like this query is kind of slow. My modestly sized db with 400k enrollments loads this in about 2 seconds. Probably worth optimizing as it's going to be the landing page for most folks. Maybe it could be done in a follow up issue

martha: That's a good idea @ttoomey and I'm happy to do it as part of this ticket rather than a follow-up issue. Could we pair on this perhaps?

PR #4580, opened 2024-07-29
Config for Synthetic CE Assessments for Enrollments

eanders:

## Description

Provides an interface to specify if a project should create synthetic CE assessments for all enrollments and the logic to create those as part of the `hud_sync` logic.
Screenshot 2024-07-29 at 3 52 52 PM
Screenshot 2024-07-29 at 3 53 02 PM

## Type of change


dtgreiner (approved):

drivers/synthetic_ce_assessment/spec/models/synthetic_ce_assessments_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+  let!(:enrollment2) { create :hud_enrollment, PersonalID: client.PersonalID, ProjectID: project2.ProjectID, data_source_id: data_source.id }
+
+  let!(:project_config) { create :synthetic_ce_assessment_project_config, project: project, assessment_type: 3, assessment_level: 1, prioritization_status: 2 }
+
+  it 'creates synthetic assessments for the enrollment where the project is flagged' do
+    SyntheticCeAssessment::EnrollmentCeAssessment.sync
+    aggregate_failures do
+      expect(SyntheticCeAssessment::EnrollmentCeAssessment.count).to eq(1)
+      expect(SyntheticCeAssessment::EnrollmentCeAssessment.first&.source).to eq(enrollment)
+    end
```
ttoomey: maybe use an expect block to ensure that sync actually creates the record (not tested)?
```suggestion
    assessments =  SyntheticCeAssessment::EnrollmentCeAssessment.all
    expect do
      SyntheticCeAssessment::EnrollmentCeAssessment.sync
    end.to change(scope, :count).by(1)
     expect(scope.first&.source).to eq(enrollment)
```

drivers/synthetic_ce_assessment/app/views/synthetic_ce_assessments/project_config/edit.haml
```diff
@@ -0,0 +1,13 @@
+= content_for :crumbs do
+  = link_to project_path(id: @project.id) do
+    = "« " + @project.name
+- content_for :title, 'Manage Synthetic CE Assessment Creation'
+%h1= content_for :title
+%p This form allows you to specify that a Coordinated Entry Assessment be created for every enrollment at the project #{@project.name}.  If the configuration is active, a synthetic assessment will be created for every enrollment in the project with the selected values.  These assessments will be created once daily, and there will be a delay between data arriving in the warehouse and the assessments being created.  Ff you change the values in the configuration, the changes will be applied to any existing synthetic assessments as well as future assessments.
```
ttoomey:
```suggestion
%p This form allows you to specify that a Coordinated Entry Assessment be created for every enrollment at the project #{@project.name}.  If the configuration is active, a synthetic assessment will be created for every enrollment in the project with the selected values.  These assessments will be created once daily, and there will be a delay between data arriving in the warehouse and the assessments being created.  If you change the values in the configuration, the changes will be applied to any existing synthetic assessments as well as future assessments.
```

app/models/grda_warehouse/synthetic/assessment.rb
```diff
@@ -80,14 +80,22 @@ def hud_assessment_hash
...
-        DateUpdated: source.updated_at,
+        DateCreated: created_at_from(source),
+        DateUpdated: updated_at_from(source),
         UserID: user_id,
         data_source_id: enrollment.data_source_id,
         synthetic: true,
       }
     end
 
+    private def created_at_from(source)
```
ttoomey, line 92: suggest adding a clarifying comment
```suggestion
    # overridden in subclass, SyntheticCeAssessment::EnrollmentCeAssessment
    private def created_at_from(source)
```

drivers/synthetic_ce_assessment/app/controllers/synthetic_ce_assessments/project_config_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    private def project
+      @project ||= project_scope.find(params[:project_id].to_i)
+    end
+
+    private def project_config
+      @project_config ||= project.synthetic_ce_project_config
+      return @project_config if @project_config.present?
+
+      @project_config = project.create_synthetic_ce_project_config(assessment_type: 2, assessment_level: 1, prioritization_status: 2)
```
ttoomey, line 52: Generally http gets should not cause state changes. Looks like this is intentional though, so I probably don't understand why this is necessary
```suggestion
      @project_config = project.build_synthetic_ce_project_config(assessment_type: 2, assessment_level: 1, prioritization_status: 2)
```

drivers/synthetic_ce_assessment/app/models/synthetic_ce_assessment/enrollment_ce_assessment.rb
```diff
@@ -0,0 +1,71 @@
...
+
+    def data_source
+      # TODO: is this correct? or should it be source.data_source or something else?
+      'Enrollment CE Assessment'
+    end
+
+    def self.sync
+      remove_orphans
+      add_new
+    end
```
ttoomey, line 41: It's kind of confusing to me to have all these class methods on an AR model that aren't scopes or similar. I guess it's fine since this is a small class but if there's a chance it could grow, I'd suggest pulling this out into something like a job class

```
class SyntheticCeAssessmentSyncJob
  def perform
      remove_orphans
      add_new
  end
  # etc...
end
```

drivers/synthetic_ce_assessment/app/models/synthetic_ce_assessment/enrollment_ce_assessment.rb
```diff
@@ -0,0 +1,71 @@
...
+      orphan_ids = pluck(:source_id) - active_ids
+      return unless orphan_ids.present?
+
+      where(source_id: orphan_ids).delete_all
+    end
+
+    def self.add_new
+      active_source_scope.preload(:client).where.not(id: self.select(:source_id)).find_each do |enrollment|
+        create(enrollment: enrollment, client: enrollment.client, source: enrollment)
+      end
```
ttoomey, line 54: are there a lot of these? Should it happen in a tx? Or maybe use active record import?

ttoomey (approved):
Looks good. I'm guessing we need a new driver because this will grow new functionality?

drivers/synthetic_ce_assessment/app/controllers/synthetic_ce_assessments/project_config_controller.rb
eanders: Since this is a `has_one` relationship, this allows us to not worry about what the ID of the config is, we can always point people to the `new` path and it ensures the config exists and then redirects to `edit`.  If you think it's too much of an anti-pattern, I can adjust the link in the view to go to new if the config hasn't been created and edit if it has.  I can't explain why I like this pattern more for `has_one` relationships, but I guess I do.

drivers/synthetic_ce_assessment/app/models/synthetic_ce_assessment/enrollment_ce_assessment.rb
eanders: There may be a bunch the first time it's run, but subsequently, only a handful a day.

drivers/synthetic_ce_assessment/app/models/synthetic_ce_assessment/enrollment_ce_assessment.rb
eanders: I suppose these could be pulled out.  This is following the pattern of synthetic events and assessments where the logic for maintaining them is encapsulated within the active record model.  We don't expect this to grow in the future (though there may be more of them.)

PR #4581, opened 2024-07-30
APR/CAPER pay for success

eanders:

## Description

* Updates to support Pay-For-Success in APR/CAPER
* This sets a new `pay_for_success` boolean on APR Clients that is true if the project is Services Only (7) and it has a funding source of Pay-For-Success (35)
* Questions are updated to match the programming instructions V1.6

## Type of change
- [x] Dependency update

### DIFF
--- a/app/models/grda_warehouse/hud/funder.rb
+++ b/app/models/grda_warehouse/hud/funder.rb
@@ -70,7 +70,7 @@ module GrdaWarehouse::Hud
     end
 
     def pay_for_success?
-      HudUtility2024.funding_source('HUD: Pay for Success', true).to_s == self.Funder.to_s
+      HudUtility2024.funding_source('HUD: Pay for Success', true, raise_on_missing: true).to_s == self.Funder.to_s
     end
 
     def operating_year
--- a/app/models/grda_warehouse/hud/funder.rb
+++ b/app/models/grda_warehouse/hud/funder.rb
@@ -69,6 +69,10 @@ module GrdaWarehouse::Hud
       HudUtility2024.funding_sources.keys.include?(funder)
     end
 
+    def pay_for_success?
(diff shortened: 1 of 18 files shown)


app/models/grda_warehouse/hud/funder.rb
```diff
@@ -69,6 +69,10 @@ def self.valid_funder_code? funder
       HudUtility2024.funding_sources.keys.include?(funder)
     end
 
+    def pay_for_success?
+      HudUtility2024.funding_source('HUD: Pay for Success', true).to_s == self.Funder.to_s
```
dtgreiner: Referencing this by the string name feels a bit fragile to me, but I'm not seeing a better way and don't think hard-coding the code is correct. It would be nice if we had something like HudUtility2024.funding_codes[:hud_pfs] in case HUD adjusts the name in a future update.

app/models/grda_warehouse/hud/project.rb
```diff
@@ -858,6 +858,13 @@ def participating_in_ce_on?(date)
       end.present?
     end
 
+    # NOTE: preload funders before calling this
+    def pay_for_success?
+      return false unless project_type == 7
```
dtgreiner: Something like this may help future-proof against for HUD changes. 

```suggestion
      return false unless HudUtility2024.performance_reporting[:other].include?(project_type)
```

dtgreiner (approved):

dtgreiner (approved):

app/models/grda_warehouse/hud/funder.rb
eanders: as discussed, added optional `raise_on_missing` to the HUD lookup methods so we can use the string and know tests will fail if HUD changes the strings.

PR #4587, opened 2024-08-01
Collect `VerifiedBy` on CLS for CE projects

martha:

## Description
GH issue: https://github.com/open-path/Green-River/issues/6272

This PR:
- Adds a new column to CLS table, `verified_by_project_id`, foreign key to `Project`
- Adds a new pick list for Continuum Projects
- Modifies the CLS collection form to collect a new value `verified_by_project_id`. It's populated by the new Continuum Projects pick list, and it collects Project ID
- Modifies the CLS form processor to get the project name for `verified_by_project_id` and copy the value onto the HUD field `VerifiedBy`
- Adds spec test

Further testing instructions on frontend PR https://github.com/greenriver/hmis-frontend/pull/862.

To discuss:
- As noted on the ticket, `VerifiedBy` is the HUD field and should collect project name. `verified_by_project_id` is just a helper field that we are adding to collect the project ID, since project names are not unique. I wasn't 100% sure about my approach of having the form collect onto the non-HUD field `verified_by_project_id`, especially since in the code, the form fields are referred to as "HUD values", so it's a bit confusing! Is this acceptable? Do we have any existing pattern for collecting non-HUD values directly onto HUD models (as opposed to onto CDEs?)

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -145,6 +145,11 @@ def self.static_options_for_type(pick_list_type, user:)
         end
 
         form_types.map { |ft| { code: ft[:value], label: ft[:desc] } }
+      when 'CONTINUUM_PROJECTS'
+        Hmis::Hud::Project.where(continuum_project: true).
```
gigxz: This project list should be limited to projects in the data source that is backing the HMIS that is making this request. Typically we rely on `viewable_by(user)` for this. In this case it sounds like we don't want the limit the list to viewable projects, but it should still be limited to continuum projects _in the HMIS_.

martha: Thanks for the catch, @gigxz

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -22,5 +22,15 @@ def schema
...
     end
+
+    def assign_metadata
+      super
+
+      current_living_situation = @processor.send(factory_name, create: false)
+      return unless current_living_situation.verified_by_project_id
+
+      project_name = Hmis::Hud::Project.find(current_living_situation.verified_by_project_id)&.name
+      current_living_situation.verified_by = project_name
```
gigxz: probably worth a comment here, similar to the one in the form structure, explaining that we copy the name into the HUD field `VerifiedBy`

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -22,5 +22,15 @@ def schema
     def dependent_destroyable?
       true
     end
+
+    def assign_metadata
+      super
+
+      current_living_situation = @processor.send(factory_name, create: false)
+      return unless current_living_situation.verified_by_project_id
```
gigxz: Hmm. I'm not sure how this should behave if `verified_by_project_id` is empty. On the one hand, I would expect that if I clear that value and save, that the VerifiedBy field is cleared.

On the other hand, if we have imported records with CurrentLivingSituation.VerifiedBy, I would expect that value to remain even when updating the form.
I might be forgetting here, did we discuss the desired behavior for migrated-in records with VerifiedBy values?

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -2004,6 +2004,19 @@ def process_assessment(new_hud_values: nil)
       expect(cls.information_date).to eq(information_date)
     end
 
+    it 'should work when verified by project ID is provided' do
+      hud_values = {
```
gigxz: could use `merge` with the `hud_values` defined for this block, so it's clear what's changing

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -2004,6 +2004,19 @@ def process_assessment(new_hud_values: nil)
...
+        'CurrentLivingSituation.informationDate' => information_date.strftime('%Y-%m-%d'),
+        'CurrentLivingSituation.currentLivingSituation' => 'SAFE_HAVEN', # 118
+        'CurrentLivingSituation.verifiedByProjectId' => p1.id,
```
gigxz: Might be worth adding a test case for re-processing an existing CLS record that already has `verifiedBy` and `verifiedByProjectId` set, and ensuring that verified by changes. (Use-case would be changing the "verified by" value on an existing CLS record that was created in HMIS). For that you can use the normal factory to create the CLS record.

gigxz (approved):
This looks good to me! I think my only outstanding question is how we want to handle migrated-in CLS records that have `VerifiedBy` values. I added it to the dev check-in agenda. I think what is proposed here in this PR is great, though, we may wanting some migration script to fill `verified_by_project_id` in addition.

db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb
```diff
@@ -0,0 +1,6 @@
+class AddVerifiedByProjectIdToCurrentLivingSituation < ActiveRecord::Migration[7.0]
+  def change
+    add_column :CurrentLivingSituation, :verified_by_project_id, :integer
+    add_foreign_key :CurrentLivingSituation, :Project, column: :verified_by_project_id, validate: false
```
eanders: This can't be a foreign key, can it?  I guess if it can be null, then that might be ok, but we won't have this for any non CE project (right?) and won't have it from imported data.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
eanders: Calling out that the length of these two fields `VerifiedBy` and `ProjectName` are not equivalent in the spec, so this could be problematic to store project name, or at least to assume it will match on imported data.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
eanders: If we pivot to storing one of the Project IDs in this field instead of name, could we?
1. Attempt to match a project and show the name if we can
2. Show the actual value if we can't

db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb
gigxz: It's a nullable foreign key, from my understanding that should be fine?

Correct we won't have it for non-CE projects or for imported data

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
gigxz: Thanks, that's a good catch. It looks like we actually have that constraint on the database -- max 100 chars on VerifiedBy. So, we should truncate here if it's longer than 100.

It's a little odd to me that we put that constraint on the DB as opposed to just truncating when we generate CSVs. But that's another question probably for the bigger db-conforming-to-csv-spec convo.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
gigxz: Are you suggesting storing ProjectID in the HUD `VerifiedBy` column? We did discuss that option. I'm not sure I quite understand. I can provide some more context around the decision to use a separate column, if that's what you're asking about

eanders: Yeah, my suggestion would be to store the `ProjectID` in `VerifiedBy`.  As long as we can provide a drop-down of projects to fill the field, I don't think HUD will ever verify if the field contains the name or ID.

gigxz: Okay, I provided some context on https://github.com/open-path/Green-River/issues/6272#issuecomment-2266013278 and suggested reopening this for discussion at the hmis check-in call. Definitely important to get agreement on the pattern we decide on, since it will surely come up again. (Like on CE Events).

db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb
ttoomey: Maybe a little more common to use add_reference. But perhaps we are using `:integer` here to match Project.id (which really should be a bigint) 
```suggestion
    add_reference :CurrentLivingSituation, :verified_by_project, foreign_key: { to_table: :Project }, index: true, null: true
```

db/warehouse/migrate/20240731145826_validate_current_living_situation_verified_by_project_id.rb
```diff
@@ -0,0 +1,10 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class ValidateCurrentLivingSituationVerifiedByProjectId < ActiveRecord::Migration[7.0]
+  def change
+    validate_foreign_key :CurrentLivingSituation, :Project
+  end
+end
```
ttoomey: Sorry, I'm confused as to what the intention of this migration is. Can you clarify? I confess I've never used `validate_foreign_key` so maybe I'm missing something

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -2004,6 +2004,19 @@ def process_assessment(new_hud_values: nil)
...
+      }
+      cls = Hmis::Hud::CurrentLivingSituation.new(client: c1, enrollment: e1, data_source: ds1)
+      process_record(record: cls, hud_values: hud_values, user: hmis_user, definition: definition)
+
+      expect(cls.verified_by_project_id).to eq(p1.id)
+      expect(cls.verified_by).to eq(p1.name)
```
ttoomey: recently I've been thinking we should make more use of expect blocks to check that the actions are actually changing the value. Perhaps:
```suggestion
      expect do
        process_record(record: cls, hud_values: hud_values, user: hmis_user, definition: definition)
      end.to change(cls, :verified_by_project_id).to(p1.id).
      and change(cls, :verified_by).to(p1.name)
```

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
ttoomey: it seems like something like this could work? 

```ruby
# if this was a normal mutation, could logic would be:
if input.verified_by_project_id
  # allow updating verified_by
  cls.verified_by = find_project_name(input.verified_by_project_id)
else if cls.verified_by_project_id
  # cls was set previously and is now cleared explicitly
  current_living_situation.verified_by = nil
else
  # current_living_situation.verified_by is preserved
end
```
So an imported cls.verified_by value would be unchanged since cls.verified_by_project_id would be null. But a user could still change the value (and once changed could then clear it as well)

db/warehouse/migrate/20240731145826_validate_current_living_situation_verified_by_project_id.rb
martha: I added the separate migration due this warning from `strong_migrations`:

```
=== Dangerous operation detected #strong_migrations ===

Adding a foreign key blocks writes on both tables. Instead,
add the foreign key without validating existing rows,
then validate them in a separate migration.

class AddVerifiedByProjectIdToCurrentLivingSituation < ActiveRecord::Migration[7.0]
  def change
    add_foreign_key :CurrentLivingSituation, :Project, column: :verified_by_project_id, validate: false
  end
end

class ValidateAddVerifiedByProjectIdToCurrentLivingSituation < ActiveRecord::Migration[7.0]
  def change
    validate_foreign_key :CurrentLivingSituation, :Project
  end
end

/app/db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb:4:in `change'
```

I guess it was just my instinct that it would be better to validate the foreign key, but we could leave it as `validate: false` without following up to validate in a separate migration. Is that our common practice?

db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb
martha: related to my comment below about `strong_migrations,` this also results in a similar warning:
```
=== Dangerous operation detected #strong_migrations ===

Adding a foreign key blocks writes on both tables. Instead, use:

class AddVerifiedByProjectIdToCurrentLivingSituation < ActiveRecord::Migration[7.0]
  disable_ddl_transaction!

  def change
    add_reference :CurrentLivingSituation, :verified_by_project, index: {algorithm: :concurrently}, null: true
  end
end

Then add the foreign key in separate migrations.

/app/db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb:3:in `change'
```

It looks like one practice we use is to just use `safety_assured` or `disable_check` with `add_reference`, so I would be fine with doing that here if that's the preferred method.

eanders: We do both.  I've had limited success with `algorithm: :concurrently` though I think we fixed something in the deployment process recently that should mean that might work.  When adding a reference to a small table, or a table we don't expect to be an issue `safety_assured` or `disable_check` is fine.  Seems like it would be NICE to use `algorithm: :concurrently` for a large table like CLS if possible.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
martha: Thanks, agree this is better!

db/warehouse/migrate/20240731144633_add_verified_by_project_id_to_current_living_situation.rb
martha: Makes sense, but I still don't feel that confident in my understanding. From what I read, the reason to prefer `add_reference` over `add_foreign_key` is because it's a shortcut for adding the column and the foreign key at the same time, which is what we want to do here; and because it's more common in our codebase as Theron pointed out in the initial comment on this thread. But to avoid blocking writes on both the Project and CLS table at the same time while adding the foreign key, StrongMigrations recommends adding the foreign key separately anyway; in that case, is there any advantage to using `add_reference` at all? Or should I revert this to what I had before?

martha: Resolved on a call - we will use `safety_assured` (ignore StrongMigrations) here because validating a column full of nulls should still be relatively fast.

db/warehouse/migrate/20240731145826_validate_current_living_situation_verified_by_project_id.rb
martha: Resolved on a call - ignoring StrongMigrations allows us to remove this extra migration

drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb
```diff
@@ -40,5 +41,17 @@ def enrollment
     def client
       load_ar_association(object, :client)
     end
+
```
martha, line 44: Hm, so of course, I wasn't quite right about my assertion on our call earlier that our existing form dropdown behavior will just work, but please let me know what you think about this hack and commentary.

martha:
Ok, retargeting to release-128 and marking as ready-for-review. 🙏 Thanks everyone for the great input and discussion so far.

drivers/hmis/lib/form_data/default/fragments/current_living_situation.json
```diff
@@ -64,15 +64,15 @@
       ]
     },
     {
-      "type": "STRING",
-      "required": false,
-      "hidden": true,
+      "type": "CHOICE",
+      "required": true,
```
martha, line 68: Should this field actually be required for CE projects? That was my impression from the ticket https://github.com/open-path/Green-River/issues/6272, but then on the call today we spent some time making sure we enabled the user to null out existing values, so just making sure.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -6,6 +6,37 @@
 
 module Hmis::Hud::Processors
   class CurrentLivingSituationProcessor < Base
+    def process(field, value)
+      attribute_name = ar_attribute_name(field)
+      current_living_situation = @processor.send(factory_name)
+
+      return super(field, value) unless attribute_name == 'verified_by_project_id'
+
+      verified_by_project_id = attribute_value_for_enum(graphql_enum(field), value)
```
gigxz: I was gonna say that we don't need `attribute_value_for_enum` since this isn't an enum field, but, I think we do need it because it converts `_HIDDEN`=>`nil`. We could skip the `graphql_enum` bit since we know that this isn't an enum field. Looks like we have this exact situation here: https://github.com/greenriver/hmis-warehouse/blob/aa9a48f3c1fc2ed68a7738a17ba4b119d9f7ec02/drivers/hmis/app/models/hmis/hud/processors/base.rb#L230

This points towards the need for a refactor, but for now maybe removing the `graphql_enum` and adding a comment is sufficient

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -6,6 +6,37 @@
...
+
+      if verified_by_project_id
+        # We collect project ID onto the non-HUD field `verified_by_project_id`,
+        # then copy the corresponding project name onto the HUD field `VerifiedBy` (aka `verified_by`).
+        project_name = Hmis::Hud::Project.find(verified_by_project_id).name
+
+        current_living_situation.assign_attributes(
+          verified_by_project_id: verified_by_project_id,
+          verified_by: project_name.truncate(100),
```
gigxz:
```suggestion
          verified_by: project_name.truncate(100), # HUD spec has 100 char limit, and so does the database
```

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -6,6 +6,37 @@
...
+      current_living_situation = @processor.send(factory_name)
+
+      return super(field, value) unless attribute_name == 'verified_by_project_id'
+
+      verified_by_project_id = attribute_value_for_enum(graphql_enum(field), value)
+
+      if verified_by_project_id
+        # We collect project ID onto the non-HUD field `verified_by_project_id`,
+        # then copy the corresponding project name onto the HUD field `VerifiedBy` (aka `verified_by`).
+        project_name = Hmis::Hud::Project.find(verified_by_project_id).name
```
gigxz: It might help for future debugging to raise something more explicit if the project id isn't found:

```
project = Hmis::Hud::Project.find_by(id: verified_by_project_id)
raise 'CLS Verified By project not found' unless project
```

drivers/hmis/lib/form_data/default/fragments/current_living_situation.json
gigxz: Yes looking at the dictionary, it should be required for CE. I still think we should support clearing the value, in case that is desired later on. For example if someone wanted to collect "verified by" in a different project type,  optionally. Or if the HUD spec changes and makes it optional.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
gigxz: Actually, I did encounter this error when editing the date on a "migrated-in" CLS record. In that situation the `verified_by_project_id` is submitted as whatever value was in `VerifiedBy`. Maybe this would be safer:

```
verified_by_project = Hmis::Hud::Project.find_by(id: verified_by_project_id) if verified_by_project_id

if verified_by_project
   # a project was selected, so assign attributes verified_by_project_id and  verified_by
elsif current_living_situation.verified_by_project_id
  # a project was previously selected and is now cleared. null out values
end
```
Thoughts?

drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb
gigxz: I think this is makes sense, and thanks for the explanatory comment!

gigxz:
Overall looks great to me. Just left a comment about one edge case with an error, that could use a test as well. Thank you!!

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
martha: Hm, I think that removing `graphql_enum` causes other problems in the `attribute_value_for_enum` procedure. `graphql_enum` returns `nil` for the type, since it doesn't respond to `value_for`; without taking that step, `attribute_value_for_enum` fails on line 109.

I'm proposing to divest from `attribute_value_for_enum` and just make the HIDDEN check directly

gigxz: Ok looks good, thanks!

gigxz (approved):

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
```diff
@@ -6,6 +6,40 @@
...
+      attribute_name = ar_attribute_name(field)
+      current_living_situation = @processor.send(factory_name)
+
+      return super(field, value) unless attribute_name == 'verified_by_project_id'
+
+      # convert HIDDEN to nil
+      verified_by_project_id = value == Base::HIDDEN_FIELD_VALUE ? nil : value
+
+      verified_by_project = Hmis::Hud::Project.find_by(id: verified_by_project_id) if verified_by_project_id
+      # Don't raise if we can't find the project, this can happen with migrated-in values, related to the hack
```
eanders: `verified_by_project_id` won't be migrated in right?  I'm not sure if we should or shouldn't hide the raise if we can't find the associated project.  If `verified_by_project_id` doesn't have a corresponding project, maybe we should blank it out?

eanders (approved):
Thanks @martha, this looks good, just one question about how we want to handle projects that were there, but aren't now.

drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb
martha: Yes that `verified_by_project_id` wouldn't be migrated in, but, see the hack described here: https://github.com/greenriver/hmis-warehouse/pull/4587/files#r1704580947

This was done in order to get the frontend behavior that we want in the form, which is: for migrated-in data, where `VerifiedBy` exists as a string while `verified_by_project_id` is null, the `VerifiedBy` string should be shown in the dropdown (but should not be selectable - only real projects are selectable options).

This means that when a CLS in that migrated-in state is re-submitted without any change to the `verified_by_project_id` dropdown, that string value of `VerifiedBy` is getting re-submitted in the `verified_by_project_id` field. I agree this is pretty confusing! 

I could either:
- Clarify my comments here to better explain what is going on
- Modify the form structure and stop trying to collapse these 2 fields into 1 visually. I think this would result in a less-good UX but clearer code

eanders: Ah, sorry, I missed that this was the same as `object` so it would potentially be set.  This is good as is  Thanks!

PR #4599, opened 2024-08-06
6362 improve cded validations

ttoomey:

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/6362)

* adjust `PublishFormDefinition.add_missing_custom_field_keys` to address TODO. Uses item.mapping&.record_type when creating cdeds
* `BaseProcessor#process_custom_field` raises if it can't find a cded
* `Hmis::Form::FormProcessor` calls `process_custom_field()` and `process()` conditionally, based on item.mapping.custom_field_key and mapping.field_name
* adjust `Hmis::Form::DefinitionValidator` to use mapping record_type if provided rather than just the owner_type inferred from role when validating cdeds
* create a `Hmis::Form::RecordType` collection helper, it's a bit more extensible than using a graphql enum
* change test factory `hmis_form_definition ` item mappings to use assessmentDate and custom fields
* adjust specs for new cded behavior

## Type of change


drivers/hmis/spec/requests/hmis/assessments/submit_assessment_spec.rb
```diff
@@ -246,7 +246,7 @@
         },
         {
           'fullMessage' => 'The Required Field must exist',
-          'attribute' => 'fieldOne',
+          'attribute' => 'base',
```
ttoomey: for custom fields, this is set to base I guess? Not sure if anything depends on that or we could just change ValidationError to return the attribute

drivers/hmis/app/models/hmis/form/record_type.rb
```diff
@@ -0,0 +1,81 @@
...
+        new(
+          id: 'ENROLLMENT',
+          owner_type: 'Hmis::Hud::Enrollment',
+          processor_name: 'Enrollment',
+        ),
+        new(
+          id: 'ENROLLMENT_COC',
+          # we don't have this HUD model in the hmis
+          # owner_type: 'Hmis::Hud::EnrollmentCoc',
+          owner_type: nil,
```
ttoomey: I'm not sure about this, there's no hmis model or processor

ttoomey (commented):
oops forgot to submit my self-review

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1211,6 +1211,44 @@ def process_record(record:, hud_values:, user:, definition:, save: true)
       expect(client.contact_points.pluck(:id)).not_to include(contact2.id)
       expect(client.contact_points.pluck(:value)).to contain_exactly('foo@bar.com', 'baz@boop.com')
     end
+
+    describe 'with unknown custom field' do
+      RSpec.shared_examples 'a processor that raises on unknown data elements' do
```
gigxz, line 1216: Cool, I didn't know about this rspec pattern. Seems really useful

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -142,7 +142,7 @@ def process_custom_field(field, value)
     return false unless record.respond_to?(:custom_data_elements)
```
gigxz: This should probably also raise:
`raise "Record does not support custom data elements" unless record.respond_to?(:custom_data_elements)`

(No need to specify the field or container, since it will be rescued and re-raised with that context

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -142,7 +142,7 @@ def process_custom_field(field, value)
...
 
     cded = Hmis::Hud::CustomDataElementDefinition.for_type(record.class.sti_name).find_by(key: field)
-    return false unless cded.present?
+    raise "Unknown Custom Data Element \"#{field}\" for record #{record.class.sti_name}" unless cded
```
gigxz: Could probably just raise "unknown custom data element" or "custom data element definition not found", since the exception should be captured and re-raised with context (see `Error processing field` err)

drivers/hmis/spec/models/hmis/form/definition_validator_spec.rb
```diff
@@ -235,7 +237,7 @@ def expect_validation_errors(definition:, role: nil, expected_errors: [])
 
   describe 'Validating custom data element definitions on publish' do
     let!(:definition) { create :hmis_form_definition, role: 'CUSTOM_ASSESSMENT' }
-    let!(:cded) { create :hmis_custom_data_element_definition, owner_type: 'Hmis::Hud::CustomAssessment' }
+    let!(:cded) { create :hmis_custom_data_element_definition, owner_type: 'Hmis::Hud::CustomAssessment', data_source: data_source }
```
gigxz, line 240: whoops - good that we catch these now, thanks

drivers/hmis/spec/requests/hmis/assessments/submit_assessment_spec.rb
```diff
@@ -21,7 +21,7 @@
   let(:c1) { create :hmis_hud_client, data_source: ds1, user: u1 }
   let!(:e1) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, user: u1, entry_date: 2.weeks.ago }
   let!(:fd1) do
-    ['informationDate', 'fieldOne', 'fieldTwo'].each do |key|
+    ['fieldOne', 'fieldTwo'].each do |key|
       create(:hmis_custom_data_element_definition, key: key, owner_type: Hmis::Hud::CustomAssessment.sti_name, data_source: ds1)
```
gigxz, line 27: Comment unrelated to this change:

On a draft PR I added a pattern for creating CDEDs on `after(:create)` of the definition factories. Left a comment [here](https://github.com/greenriver/hmis-warehouse/pull/4607/files#r1713723064) so I could link it. I _think_ it makes sense as a pattern, because if you are using the definition in a test than its most likely you will want the CDEDs to exist, too.

Just wanted to raise the idea. If you want we can wait for discussion until I get around to requesting review on that PR, or you can feel free to comment there.

drivers/hmis/spec/requests/hmis/assessments/submit_assessment_spec.rb
gigxz: Huh, yeah it should be fine to change it, just adding `|| item.mapping&.custom_field_key` to the end of [this line](https://github.com/greenriver/hmis-warehouse/blob/a74e8d2f57d159655f6923d3b1dc354eb692742b/drivers/hmis/app/models/hmis/form/definition.rb#L439). It shouldn't make any difference since the errors are "wired up" based on the `link_id` that is resolved on each ValidationError, not the `attribute` (which is this field). Not ideal that we check for `attribute` here but not `link_id`, that should probably be added to the assertion.

drivers/hmis/app/models/hmis/form/record_type.rb
gigxz: Yeah this can be removed, it's no longer supported. It was removed in the 2024 HUD Spec.

It should be safe to remove it from the graphql enum as well, there will be no forms submitting this.

drivers/hmis/app/models/hmis/form/record_type.rb
```diff
@@ -0,0 +1,81 @@
...
+        new(
+          id: 'CLIENT',
+          owner_type: 'Hmis::Hud::Client',
+          processor_name: 'Client',
+        ),
+        new(
+          id: 'CURRENT_LIVING_SITUATION',
+          owner_type: 'Hmis::Hud::CurrentLivingSituation',
+          processor_name: 'CurrentLivingSituation',
+        ),
```
gigxz, line 30: It feels a little off that we now need a mapping between the **processor name** and the **owner type**, since the processor is supposed to be responsible for "knowing" which record(s) to update (usually by setting `factory_name` which will use the appropriate factory to find/create the record it needs to process..).

Another way could be for the processor to still "own" the knowledge of which record it modifies, but expose that externally as a class method. Looking at the existing (non-class) methods, something in the neighborhood of:
```
processor_name = 'CurrentLivingSituation'
Hmis::Form::FormProcessor.valid_containers[processor_name].record_type
```


I don't know if that's necessarily better, but it does feel more inline with the initial intention of the Form Processor abstraction (though I admit it has gotten away from us).

Am I correct that `PublishFormDefinition` is the only place we currently look at this `owner_type` attribute on this object?

gigxz (commented):
This looks great. Left one comment inquiring into the RecordType helper class approach, but I'd also be happy with this as-is. Tested with publish=>save>submit and all works great.

This change also kicked up the dust on some (minor) _existing_ issues, most of my comments are about those. It would be great to get those into this branch, but if it feels too out of scope Ill make tickets for anything left.

thanks!

drivers/hmis/app/models/hmis/form/record_type.rb
ttoomey: > It feels a little off that we now need a mapping between the processor name and the owner type

Yes, that's a fair point. I think delegating the owner_type from the processor is appealing. I might be missing something but I'm not sure the design of our form processors will accommodate this very well.

The record_type.owner_type is used  in the context of form validation and creation of CDEDs during publication. In this context we want to know the owner_type (class).

However the form processors don't seem to provide a clean way to query the class of objects they might create; they just call back to factories to get an instance. Given that, it seems like you'd have to jump through some hoops to figure out the owner_type. Probably by instantiating a new dummy form processor, pass that to the field processor, call the field_processor.factory, and introspect on the returned class.

I suppose we could simply co-locate the owner_type as a constant on the field processor class. Maybe that's more inline with what you're suggesting?

gigxz: Lets keep the currently implementation as discussed. Doesn't seem worth the effort to co-locate the mappings as constants when they wouldn't actually be used during processor evaluation. Thanks!

gigxz (approved):
Looks good once HMIS tests pass! 👍

PR #4602, opened 2024-08-06
Quarterly full refresh upload to AC DW

gigxz:
## Description

https://github.com/open-path/Green-River/issues/6196

* Add a quarterly upload with a 10 year lookback of HMIS CSV
* Add `daily_uploads` mode so we don't need to maintain the list of uploads in the rake task, which was a little confusing

## Type of change

### DIFF
--- a/drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/data_warehouse_upload_job.rb
+++ b/drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/data_warehouse_upload_job.rb
@@ -12,39 +12,27 @@ module HmisExternalApis::AcHmis
 
     attr_accessor :state
 
-    def perform(mode)
-      setup_notifier("AC Data Warehouse upload (mode: #{mode})")
+    def perform(methods)
+      setup_notifier("AC Data Warehouse upload (methods: #{methods})")
       if Exporters::DataWarehouseUploader.can_run?
-        Rails.logger.info "Running #{mode} upload clients job"
+        Rails.logger.info "Running #{methods} DW upload job"
 
-        # The 'mode' indicates which upload(s) to run. Daily and quarterly uploads are run on a schedule.
-        # It's possible to run individual jobs for testing purposes or if we ever need to run manually.
-        case mode
-        when 'clients_with_mci_ids_and_address' then clients_with_mci_ids_and_address
-        when 'hmis_csv_export' then hmis_csv_export
-        when 'hmis_csv_export_full_refresh' then hmis_csv_export_full_refresh
-        when 'project_crosswalk' then project_crosswalk
-        when 'move_in_addresses' then move_in_address_export
-        when 'postings' then posting_export
-        when 'custom_fields' then custom_fields_export
-        when 'pathways' then pathways_export
-        when 'daily_uploads'
-          clients_with_mci_ids_and_address
-          hmis_csv_export
-          project_crosswalk
-          move_in_address_export
-          posting_export
-          custom_fields_export
-          pathways_export # we should remove this eventually, same data is included in custom_fields_export
-        when 'quarterly_uploads'
-          hmis_csv_export_full_refresh
-        else
-          raise "invalid item to upload: #{mode}"
+        Array.wrap(methods).each do |method|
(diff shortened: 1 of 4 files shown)


drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/data_warehouse_upload_job.rb
```diff
@@ -12,18 +12,32 @@ class DataWarehouseUploadJob < BaseJob
 
     attr_accessor :state
 
-    def perform(mode = 'clients_with_mci_ids_and_address')
+    def perform(mode)
```
eanders: maybe `mode` should be an array of method names?  (or I guess `daily_uploads` could just be a method that calls the whole set of known methods. 
```
def perform(methods)
   methods.each do |method|
     next unless known?(method) # maybe raise?
    
     call(method)
end

def known?(method)
  known_methods.include?(method)
end

def known_methods
  [
    'clients_with_mci_ids_and_address',
    ...
  ].freeze
end
```

Just to be clear, there's nothing wrong with what's here, but I think the potential to get misalignment between daily and all the methods is non-zero.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/hmis_export_fetcher.rb
```diff
@@ -15,7 +15,7 @@ class HmisExportFetcher
 
     delegate :content, to: :export
 
-    def run!
+    def run!(lookback_years: 3)
```
eanders: I'd probably just make this take a start date so you could send an arbitrary range if you wanted.
```suggestion
    def run!(start_date: 3.years.ago.to_date)
```

eanders (approved):
This looks good @gigxz !  I dropped a couple suggestions in there to try and future proof it a bit, but feel free to ignore them if you think it's better as is.

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/data_warehouse_upload_job.rb
gigxz: Yeah, definitely a chance for misalignment. I was trying to get away from the fact that the methods themselves need to be listed in `export.rake` which means that when you add a new method you need to add it to both files, which was always annoying (and also has a potential for misalignment).

I think with your suggestion, we would still need to list out each upload in the rake task? Or is there something I'm missing?

eanders: I think maybe the bets of both worlds can be achieved.  I'm probably missing something, and this adds a wee bit of duplication, but you could keep the same API in the rake task if you just add methods for daily and monthly. 

```
def perform(methods)
   methods.each do |method|
     next unless known?(method) # maybe raise?
    
     call(method)
end

def known?(method)
  known_methods.include?(method)
end

def known_methods
  [
    'clients_with_mci_ids_and_address',
    'daily_uploads',
    'quarterly_uploads',
    ...
  ].freeze
end

def daily_uploads
  known_methods - ['hmis_csv_export_full_refresh', 'daily_uploads', 'quarterly_uploads']
end

def quarterly_uploads
 'hmis_csv_export_full_refresh'
end
```

eanders: Alternatively, those could be special cased at the top of perform so you wouldn't have to remove the set from the known_methods array.

gigxz: Ah I see what you mean, thanks. I think I addressed this adequately with https://github.com/greenriver/hmis-warehouse/pull/4602/commits/524cbc6dcec0dc8de81ee6a7ac6a8e0834e83f1f

eanders (approved):
Thanks @gigxz, I don't know about you, but I find that way more readable.

gigxz:
Me too, thanks for the suggestsion!

PR #4603, opened 2024-08-07
Restrict 'My Dashboard' tab visibility

gigxz:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6337

## Type of change
- [x] Bug fix
- [ ] Dependency update


app/models/concerns/user_concern.rb
```diff
@@ -172,6 +172,13 @@ def can_access_project?(project, permission: :can_view_projects)
       cached_viewable_project_ids(permission: permission).include?(project.id)
     end
 
+    def can_view_my_dashboard?
```
gigxz: I don't think this should go in the user concern actually, since it's not applicable to the warehouse users. I think we can just put it on the HMIS User class, and perhaps inline the  `can_access_staff_assignment_projects?` logic.

app/models/concerns/user_concern.rb
```diff
@@ -172,6 +172,13 @@ def can_access_project?(project, permission: :can_view_projects)
...
+      Rails.cache.fetch('user/can_view_my_dashboard', expires_in: 1.minutes) do
+        # todo @martha - possible to manually expire cache when user impersonation changes?
```
gigxz: The cache key should contain the user id, since we only want to cache the result per user -- we don't want to cache it globally, which would return the same thing for everyone.

I like the approach taken elsewhere in this file: `key = [self.class.name, __method__, id]`. The first 2 items are to uniquely specify this use case –– its unlikely that we'll accidentally collide with some cache key elsewhere in the application. The last item is the user id, which ensures that each user will have their values cached separately.

Once the `id` is in the cache key, you shouldn't need to do anything special for impersonation. When impersonating someone else, the `current_user` will have a different `id`, so this will re-evaluate. Good thought to test that!

drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -5,6 +5,7 @@
 
 class Hmis::ProjectConfig < Hmis::HmisBase
   self.table_name = 'hmis_project_configs'
+  pc_t = Hmis::ProjectConfig.arel_table # todo @martha - is this ok?
```
gigxz: I think this is a little funny, I haven't seen us use this pattern. I would keep it as it was (re-declare in the other scope). You could also use `arel_table` directly (instead of `pct_t`, you should have access to that here since it's a class method on the class.

gigxz:
Whoops I can't approve this PR since I opened it technically, but looks good to me! I tested it out including with impersonation.

martha (approved):
😂

PR #4607, opened 2024-08-08
Resolve most recent Form Definition version for editing Assessments

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6112

Support new desired behavior:
* [unchanged] When you open an “old” Assessment for viewing, it is re-opened using the form that was most recently used to submit it (record.form_processor.definition_id). If no such reference exists, use the “most relevant” form.
* [new] When you open an “old” Assessment for editing, it is re-opened using the most recent version of the form that was last used to submit it. **This form is resolved as `upgradedDefinitionForEditing`**

Additional changes include:
* system test coverage
* simplify Custom Assessment factories
* update Definition factories to make their own CDEDs
* resolve `version` on definition (I think there was a request down the line to surface the version in the config tool form history. I'm not actually using it here)

Frontend PR has instructions on how to test: https://github.com/greenriver/hmis-frontend/pull/882

## Type of change


drivers/hmis/spec/factories/hmis/form/definitions.rb
```diff
@@ -171,4 +171,84 @@
...
+
+      Hmis::Hud::CustomDataElementDefinition.where(
+        owner_type: 'Hmis::Hud::CustomAssessment',
+        key: :veteran_hoh_custom_question,
+        label: 'Custom field for Veteran HoH only',
+        field_type: :string,
+        data_source: evaluator.data_source,
+        UserID: '1',
+      ).first_or_create!
+    end
```
gigxz: Pattern to create CDEDs alongside definition fixtures, which I think makes sense. Using first_or_create to avoid conflicts. (though that should be a non-issue. could change back to use creation factories instead).

Should we use this pattern for all the definition factory variants?

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
```diff
@@ -58,44 +59,51 @@ def role
...
+
+      previous_definition = definition
+      # If original form is not retired, then we should stil use it for editing.
+      return unless previous_definition.retired?
+
+      # Find the published version of the previous definition. If this resolves nil, then the original form will be used.
+      published_definition = previous_definition.published_version
+      published_definition&.filter_context = { project: project, active_date: object.assessment_date }
+      published_definition
+    end
```
gigxz, line 83: This is the meaningful change. Very open to suggestions on the naming of field `assessment.upgradedDefinitionForEditing { ... }`. I don't love it but was the best I could come up with

drivers/hmis/spec/system/hmis/assessment_definitions_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+      it 'uses old form version for editing' do
+        click_link old_definition.title
+        assert_text old_definition.title
+        expect(page).to have_field('Assessment Date', with: wip_assessment.assessment_date.strftime('%m/%d/%Y'))
+        assert_text 'Text on old form'
+      end
+    end
+  end
+
+  context 'Performing Household Intake assessments with various definition versions' do
```
gigxz, line 90: I would appreciate feedback on the structuring of this test. I set up one household, where each member has different states, and tested each member in a different test case.

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
martha: Seems fine to me - it is descriptive. Some other ideas - `latestDefinition`, `publishedDefinition`, `editingDefinition`, `definitionForEditing` all more concise but less descriptive.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -564,11 +568,10 @@ def walk_definition_nodes(as_open_struct: false, &block)
   # Find and/or initialize CustomDataElementDefinitions that are collected by this form.
   # This is used by PublishExternalFormsJob.
   # This should eventually be removed as we're now moving towards relying on Publish Form to generate CDEDs.
```
martha:
```suggestion
  # For internal forms, we now rely on Publish Form to generate CDEDs, but this is still used in test fixtures.
```

drivers/hmis/spec/factories/hmis/hud/custom_assessments.rb
```diff
@@ -27,31 +27,29 @@
...
+    # end
+    after(:create) do |assessment, _evaluator|
+      assessment.update!(assessment_date: assessment.enrollment.entry_date)
+      assessment.update!(date_created: assessment.enrollment.date_created)
+      assessment.update!(date_updated: assessment.enrollment.date_updated)
     end
   end
+
+  factory :hmis_wip_custom_assessment, parent: :hmis_custom_assessment do
+    wip { true }
```
martha, line 53: This no longer uses `assessment.save_in_progress`, which set the enrollment's `project_id` to nil, to match what in-progress assessments' enrollments really look like in our db -- is that understanding correct?

drivers/hmis/spec/system/hmis/assessment_definitions_spec.rb
```diff
@@ -0,0 +1,171 @@
...
+    let!(:old_definition) do
+      fd = create(:hmis_intake_assessment_definition, identifier: definition.identifier, title: 'Old Special Intake', version: 0, status: :retired)
+      fd.definition['item'][0]['item'] << { 'type': 'DISPLAY', 'link_id': 'old_message', 'text': 'Text on old form' }
+      fd.save!
+      fd
+    end
+
+    # e1 (HoH): Intake was Submitted with old form
+    let!(:c1) { create :hmis_hud_client, data_source: ds1, first_name: 'Parent', last_name: 'Jones' }
+    let!(:e1) { create :hmis_hud_wip_enrollment, data_source: ds1, project: p1, client: c1, entry_date: 2.months.ago, relationship_to_hoh: 1 }
```
martha: Maybe, change `e1` and `e2` to use `hmis_hud_enrollment` factory (rather than `wip`) since they both have completed intakes already?

drivers/hmis/spec/system/hmis/assessment_definitions_spec.rb
martha: I think this structure is good, especially if it's important that we test the new/old form definition behavior specifically in the context of a multi-member household.

If not, then maybe the test could be simplified and moved to a separate file where the focus is on testing the multi-member household intake process, rather than the old/new form definition behavior.

But if it is important, or if we just want to keep the existing test that you've written (because code wins 🚀 ), then it could be good to add a short comment explaining why the test is testing these 2 sorta-kinda distinct concepts.

martha (approved):

drivers/hmis/spec/factories/hmis/hud/custom_assessments.rb
gigxz: I think you're looking at the `enrollment.save_in_progress` function. The `assessment.save_in_progress` function _used to_ do something similar (null out an ID), but we simplified it so it just sets a `wip` flag to true on the CustomAssessment table. So, this factory should be producing the same result.

drivers/hmis/spec/system/hmis/assessment_definitions_spec.rb
gigxz: yes that was an accident, thanks!

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: good catch, thanks

drivers/hmis/spec/system/hmis/assessment_definitions_spec.rb
gigxz: Okay, thanks. I do think it's important that the household has a variety of assessment setups, so we can be sure it works in that situation. I added a couple comments to that effect!

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
gigxz: Okay, thanks for the ideas. I will leave it as-is!

drivers/hmis/spec/factories/hmis/hud/custom_assessments.rb
martha: Aha, makes sense! Thanks 👍

PR #4610, opened 2024-08-09
cleanup rendering of client homeless verification job

dtgreiner:

## Description

Cleanup the homless verificaiton pdf generation so it runs more in line with the current architecture.

## Type of change

### DIFF
--- a/app/models/client_history.rb
+++ b/app/models/client_history.rb
@@ -68,13 +68,14 @@ class ClientHistory
       pdf = io.read
     end
 
-    client_file = ::GrdaWarehouse::ClientFile.new
-    client_file.client_id = client.id
-    client_file.user_id = requesting_user&.id || user.id
-    client_file.note = "Auto Generated for prior #{years} years"
-    client_file.name = file_name
-    client_file.visible_in_window = true
-    client_file.effective_date = Date.current
+    client_file = ::GrdaWarehouse::ClientFile.new(
+    client_id: client.id,
+    user_id: (requesting_user&.id || user.id),
+    note: "Auto Generated for prior #{years} years",
+    name: file_name,
+    visible_in_window: true,
+    effective_date: Date.current,
+)
     client_file.tag_list.add(['Homeless Verification'])
     begin
       tmp_path = Rails.root.join('tmp', "service_history_pdf_#{client.id}.pdf")
--- a/app/models/client_history.rb
+++ b/app/models/client_history.rb
@@ -1,0 +1,147 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class ClientHistory
+  attr_reader :client, :user, :requesting_user, :years
+  def initialize(
+    client_id:,
+    user_id:,
+    years:
+  )
+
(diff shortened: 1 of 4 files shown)


drivers/client_access_control/app/controllers/client_access_control/history_controller.rb
```diff
@@ -54,134 +54,20 @@ def queue
     end
 
     def pdf
-      @user = User.setup_system_user
-      current_user ||= @user # rubocop:disable Lint/UselessAssignment
+      @client_history = ClientHistory.new(client_id: params[:client_id].to_i, user_id: params[:user_id].to_i, years: params[:years]&.to_i)
+      @client ||= @client_history.client
```
eanders, line 59: Don't need this any more right?
```suggestion
```

eanders: Oh, that's so the test will pass.  It's fine, but I wonder why it's needed for the test?

app/models/client_history.rb
```diff
@@ -0,0 +1,147 @@
...
+    end
+
+    client_file = ::GrdaWarehouse::ClientFile.new
+    client_file.client_id = client.id
+    client_file.user_id = requesting_user&.id || user.id
+    client_file.note = "Auto Generated for prior #{years} years"
+    client_file.name = file_name
+    client_file.visible_in_window = true
+    client_file.effective_date = Date.current
+    client_file.tag_list.add(['Homeless Verification'])
```
eanders: This is fine as is, but might be nicer as part of the new statement.  Any idea why we did it this way?
```suggestion
    client_file = ::GrdaWarehouse::ClientFile.new(
    client_id: client.id,
    user_id: (requesting_user&.id || user.id),
    note: "Auto Generated for prior #{years} years",
    name: file_name,
    visible_in_window: true,
    effective_date: Date.current,
)
    client_file.tag_list.add(['Homeless Verification'])
```

eanders (commented):
Thanks @dtgreiner this looks like a really nice cleanup.

app/models/client_history.rb
dtgreiner: This was the format being used in the previous file, and I just copied it forward. I like the cleanup though and will push it into the PR.

eanders (approved):
Thanks!  Just waiting on tests to pass before merging.

drivers/client_access_control/app/controllers/client_access_control/history_controller.rb
dtgreiner: The `pdf` method is calling the `show` method from within itself, and the `show` method utilizes the `@client` instance variable. `show` generally has the `set_client` method running before it, but it looks like since it is called this way, that's not happening.  We could specifically call that out, but then would have to set the `@user` and `current_user` variables before doing that so the `set_client` method doesn't break (or adjust that method a bit).

This way felt a bit cleaner when I was looking into it on Friday, but could be convinced otherwise too.

eanders: Yeah, this is fine.  Maybe a quick comment above?
```
# set @client for when this method is called outside of the controller context.
```

dtgreiner: just pushed it up!

PR #4612, opened 2024-08-10
Enforce `.rubocop.yml` Excluded files in the CI

eanders:

## Description

This copies the excluded file patterns (adjusted for grep) into the rubocop CI so we don't throw errors on files we aren't expecting.

## Type of change

### DIFF
--- a/.github/workflows/asset_compilation.yml
+++ b/.github/workflows/asset_compilation.yml
@@ -104,7 +104,7 @@ jobs:
     - name: Install gems
       if: always()
       run: |
-        gem install bundler --version=2.5.7
+        gem install bundler --version=2.5.17
         bundle config set --local without 'production staging development'
         bundle install --jobs 10 --retry 3
 
--- a/.github/workflows/rubocop.yml
+++ b/.github/workflows/rubocop.yml
@@ -46,7 +46,7 @@ jobs:
       - id: gems
         name: Install gems
         run: |
-          gem install bundler --version=2.5.7
+          gem install bundler --version=2.5.17
           bundle config set --local without 'production staging'
           bundle install --jobs 4 --retry 3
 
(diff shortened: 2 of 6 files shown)


.github/workflows/rubocop.yml
```diff
@@ -59,7 +59,8 @@ jobs:
         env:
           ALL_CHANGED_FILES: ${{ steps.changed-files.outputs.all_changed_files }}
         run: |
-          files=`echo "${ALL_CHANGED_FILES}" | tr " " "\n" | grep "**/*.rb" | tr "\n" " ./"`
+          ignore_patterns="db/*\|config/*\|config/environments/*\|config/initializers/*\|config/locales/*\|scripts/*\|bin/rails\|bin/rake\|bin/bundle\|bin/db_prep\|app/assets/*"
```
eanders: @ttoomey if you have any great ideas on how to pull this from `.rubocop.yml` it feels slightly brittle to hard-code this here.

ttoomey: Let me just make sure I understand the issue: we have these exclusions in `.rubocop.yml` which are files that we never want to lint(?). But we also only want to run rubocop on changed files, since there are many linter issues that we don't want to fix in other parts of the codebase. But if we change a file in one of these excluded directories, rubocop checks it anyway? So this is copying the exclude patterns out of rubocop to ensure those files are always excluded even if they are changed

I guess that works but maybe it would be better to just address the issues in the exclude list (or disable cops with a todo or magic comments). Then we wouldn't need a list at all. But I'm probably missing something.

More generally, it seems risky to me to exclude some of these files. The config files are sensitive, it seems like we'd want to ensure they pass a linter check at a minimum.

Also, it looks like `scripts` dir in the pattern above wouldn't be needed since that dir doesn't exist

.github/workflows/asset_compilation.yml
```diff
@@ -104,7 +104,7 @@ jobs:
     - name: Install gems
       if: always()
       run: |
-        gem install bundler --version=2.5.7
+        gem install bundler --version=2.5.17
```
ttoomey, line 107: thanks for fixing! Sorry I missed some of these hard-coded bundle versions.

ttoomey (approved):
Looks good, thanks for the improvements!

.github/workflows/rubocop.yml
eanders: That's basically the gist of it.  
It might be safe to just exclude `db/` files.  The issue is that some files are auto generated and trigger linting errors (`schema.rb`) in CAS is one big one, so even though all of **your** code passes, the rubocop CI fails.  Other files should probably be fixed as we tough them.  

Maybe we should just exclude files that are code gen'd where we don't control the code generator, or that are part of core rails and shouldn't be changed.

Mostly I was looking for a pattern, and it sounds like this is mostly OK as an approach.  I'll dial back what we're excluding and we can see what happens.

PR #4621, opened 2024-08-14
Time-based Annual Assessment Calculation

dtgreiner:

## Description

Update the calculation for if an annnual assessment is expected to use a time-based calculation regardless of if the service is a NBN. This will no longer count bed nights and instead go strictly by a year-based active enrollment anniversary.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/models/concerns/hud_reports/clients.rb
+++ b/app/models/concerns/hud_reports/clients.rb
@@ -32,24 +32,24 @@ module HudReports::Clients
       }
     end
 
-    # Assessments are expected if:
-    # You are the head of household.
-    # If the enrollment lasted 365 days or more
-    # and the reporting period end is 365 days or more since the beginning of the enrollment
-    # and the enrollment started 365 days ago
-    private def annual_assessment_expected?(enrollment)
+    # Calculate the head of household’s number of years in the project. This can be done using the same
+    # algorithm as for calculating a client’s age as of a certain date. Use the client’s [project start date] and the
+    # [report end date] as the two dates of comparison. This calculation is time-based, not service-based, so a
+    # client with a [project start date] that is active for a year or longer in a night-by-night emergency shelter
+    # would require an Annual Assessment, regardless of how many bed night dates were recorded. It is
+    # important to use the “age” method of determining client anniversaries due to leap years; using “one year =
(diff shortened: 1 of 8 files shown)


app/models/concerns/hud_reports/clients.rb
```diff
@@ -32,24 +32,24 @@ module HudReports::Clients
...
-        enrollment.enrollment.services.
-          bed_night.between(start_date: enrollment.first_date_in_program, end_date: end_date).
-          count >= 365
-      else
-        true
-      end
+      start_for_annual = enrollment.entry_date
+      end_date = [enrollment.last_date_in_program, report_end_date].compact.min
+      # Get difference in years, ignoring month/date
+      years_in_project = end_date.year - start_for_annual.year
```
eanders, line 48: I could be wrong, but I'm not sure this is what we are looking for.  `.year` will give you something like `2020`, so `'2020-12-10'.to_date.year` would result in `2020` and `'2021-01-05'.to_date.year` would result in `2021`, and `2021 - 2020` will give `1` but they aren't 1 year apart.  I think we need to lean on ruby for determining how many years apart they are, or maybe it would be safe to get the days difference and check for 365?  I feel like there was explicit advice not to do that.  Maybe best to add a year to the entry date and confirm that it is < the end_date?

app/models/concerns/hud_reports/clients.rb
```diff
@@ -32,24 +32,24 @@ module HudReports::Clients
...
+      # Remove 1 year if month/date of start is after the month/date of the end date. This will account for leap years.
+      years_in_project -= 1 if start_for_annual + years_in_project.years > end_date
```
eanders, line 50: Hmm, maybe this is doing that here....

dtgreiner: Yeah, line 48 is just getting the difference in the years, so it could be off by 1 depending on the month/day of both dates. Line 50 is the adjustment for this. It is checking if the start date happens later in the year than the end date and adjusting accordingly. 

The HUD documentation does specifically call out that we shouldn't use 1 year = 365 days due to leap years. 

Happy to look through it some more or do a quick call.

eanders (approved):

PR #4625, opened 2024-08-16
Zparks/8 9 24/update csg for multi project

ZachCMP:

## Description

Updates CSG import to allow importing multiple projects to a single CSG program. Also adds a test for the reporting logic

## Type of change


ZachCMP:
@eanders Looks like some other changes came through for the `warehouse.sql` file. Let me know if you want me to clean them up, or if this looks okay to you

drivers/ma_reports/app/models/ma_reports/csg_engage/report_components/program.rb
```diff
@@ -26,12 +26,12 @@ def initialize(program_mapping)
...
 
-    def project
-      @project ||= program_mapping.project
+    def project_ids
+      @project_ids ||= program.program_mappings.pluck(:project_id)
     end
 
     def households_scope
-      project.enrollments.heads_of_households.preload(project: [:project_cocs])
+      GrdaWarehouse::Hud::Enrollment.joins(:project).where(project: { id: project_ids }).heads_of_households.preload(project: [:project_cocs])
```
eanders, line 34: Where is household_scope getting used?  This change potentially will give you back households with the same `HouseholdID` in different projects in the same scope, so that should be disambiguated with `[HouseholdID, ProjectID, data_source_id]` if you are counting households.  That said, HouseholdID is _supposed_ to be unique across a data source, so that's a theoretical need.

eanders (approved):

ZachCMP:
@eanders This should be good to go now. I adjusted the household identifier to be an aggregate key of the enrollment id, project id, and data source id. To keep the IDs somewhat anonymized I converted them to hex before constructing the aggregate key. Let me know if you need anything else from me to get this on staging

drivers/ma_reports/app/models/ma_reports/csg_engage/report_components/household.rb
```diff
@@ -12,7 +12,9 @@ def initialize(hoh_enrollment)
       @hoh_enrollment = hoh_enrollment
     end
 
-    field('Household Identifier') { hoh_enrollment.id.to_s }
+    field('Household Identifier') do
+      [hoh_enrollment.id, hoh_enrollment.project.id, hoh_enrollment.data_source.id].map { |id| id.to_s(16) }.join('::')
```
eanders: So, in this case using `.id` on the enrollment would be sufficient, but I _think_ you want to tie it to eh `HouseholdID` field in case they change who is HoH.  Using `project.id` also means you don't need `data_source.id`.

We usually do something like this:
```suggestion
      Digest::MD5.hexdigest([hoh_enrollment.HouseholdID, hoh_enrollment.project.id].join('|'))
```

That should give a consistent value based on the project and HouseholdID and should be the correct length to fit in the HouseholdID field.

drivers/ma_reports/spec/models/ma_reports/csg_engage/report_components/report.rb
eanders, line 1: Should this file end in `_spec.rb`?  I'm not sure our CI will catch this without that naming convention.

eanders (changes requested):
Just a couple comments/questions.  Otherwise, I think this should be good.

drivers/ma_reports/spec/models/ma_reports/csg_engage/report_components/report.rb
ZachCMP: Fixed

drivers/ma_reports/app/models/ma_reports/csg_engage/report_components/household.rb
ZachCMP: @eanders The field length limit is 20 chars, and I think MD5 hashes are 32. I guess we could just truncate it, but technically that could cause collisions, though unlikely.

ZachCMP: @eanders This is fixed now

PR #4627, opened 2024-08-20
Add system test for form builder

martha:

## Description
GH issue: https://github.com/open-path/Green-River/issues/6340
Depends on frontend PR: https://github.com/greenriver/hmis-frontend/pull/897/files

This PR
- Adds system tests for the Form Builder "happy path."
  - See issue linked above, and todos inline, for more test cases that we could/should add. I've spent about 3.5h so far of a 5h estimate and I don't think that adding coverage for these other cases will be onerous, but
  - (Edit: Added additional test cases)
- This test also uncovers a clobbering bug, which is probably higher priority to investigate and solve! See comment thread inline... Edit: https://github.com/open-path/Green-River/issues/6555

## Type of change
- [ ] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/spec/system/hmis/form_builder_spec.rb
+++ b/drivers/hmis/spec/system/hmis/form_builder_spec.rb
@@ -1,0 +1,180 @@
+require 'rails_helper'
+require_relative '../../requests/hmis/login_and_permissions'
+require_relative '../../support/hmis_base_setup'
+
+# Custom matcher for asserting order of text in Capybara page body, helpful for testing reordering of form items.
+# https://stackoverflow.com/questions/64002940/is-it-possible-to-find-out-if-an-element-comes-before-the-second-element-in-capy
+RSpec::Matchers.define :appear_before do |later_content|
+  match do |earlier_content|
+    page.body.index(earlier_content) < page.body.index(later_content)
+  end
+end
+
+RSpec.feature 'HMIS Form Builder', type: :system do
+  include_context 'hmis base setup'
+
+  let!(:ds1) { create(:hmis_data_source, hmis: 'localhost') }
+  let!(:access_control) { create_access_control(hmis_user, p1) }
+  let!(:published) { create :custom_assessment_with_custom_fields_and_rules, title: 'System Test 1', identifier: 'system_test_1', data_source: ds1 }
+  let!(:draft) { create :custom_assessment_with_custom_fields_and_rules, title: 'System Test 2', identifier: 'system_test_2', status: 'draft', data_source: ds1 }
+  let!(:advanced_draft) { create :custom_assessment_with_field_rules_and_autofill, identifier: 'advanced_form', status: 'draft', data_source: ds1 }
+
+  context 'Form builder happy path' do
+    before(:each) do
+      sign_in(hmis_user)
+      disable_transitions
+      visit '/admin/forms'
+    end
+
+    it 'creates a new form with a draft' do
+      visit '/admin/forms'
+      click_button 'New Form'
+
+      assert_text 'Form Type'
+      mui_select 'Custom assessment', from: 'Form Type'
+      fill_in 'Form Title', with: 'System Test 3'
+      fill_in 'Form Identifier', with: 'system_test_3'
+      click_button 'Save'
+
+      assert_text 'SELECTED FORM System Test 3'
+      assert_text 'Edit Draft'
+    end
+
+    it 'creates a new draft' do
+      visit '/admin/forms/system_test_1'
+      click_button 'New Draft'
+
+      assert_text 'EDITING DRAFT System Test 1'
+    end
+
+    context 'with an existing form draft' do
+      before(:each) do
+        visit "/admin/forms/system_test_2/#{draft.id}/edit"
+      end
+
+      it 'edits a form item' do
+        find("button[aria-label='section_1 item actions']").click
+        find("li[aria-label='edit section_1']").click
+        assert_text 'EDIT FORM ITEM Test Custom Assessment'
+
+        fill_in 'Group Label', with: 'A New Section Title'
+        find("button[type='submit']").trigger('click') # for some reason, .trigger('click') works but .click does not?
+        assert_text 'Group: A New Section Title'
+      end
+
+      it 'adds new items' do
+        find("button[aria-label='Add Group item']").click
+        assert_text 'ADD NEW FORM ITEM Group Item'
+        assert_no_selector "[role='submit']" # can't submit without a Label
+        fill_in 'Group Label', with: 'Brand New Group'
+        find("button[type='submit']").trigger('click')
+        assert_text 'Group: Brand New Group'
+
+        find("button[aria-label='Add Display item']").click
+        assert_text 'ADD NEW FORM ITEM Display Item'
+        fill_in 'Display Text', with: 'This is a <b>displayable item</b>.'
+        find("button[type='submit']").trigger('click')
+        assert_text 'Display: This is a displayable item.'
+
+        find("button[aria-label='Add Text item']").click
+        assert_text 'ADD NEW FORM ITEM Text Item'
+        find("input[name='text']").fill_in(with: 'What is the answer to this question?')
+        find("button[type='submit']").trigger('click')
+        assert_text 'Text: What is the answer to this question?'
+
+        find("button[aria-label='Add Paragraph item']").click
+        assert_text 'ADD NEW FORM ITEM Paragraph Item'
+        find("input[name='text']").fill_in(with: 'What about this long paragraph?')
+        find("button[type='submit']").trigger('click')
+        assert_text 'Paragraph: What about this long paragraph?'
+
+        draft.reload
+        # todo @martha - replace with `dig` for code consistency
+        expect(draft.definition.dig('item', 1, 'text')).to eq('Brand New Group')
+        expect(draft.definition.dig('item', 2, 'text')).to eq('This is a <b>displayable item</b>.')
+        expect(draft.definition.dig('item', 3, 'text')).to eq('What is the answer to this question?')
+        expect(draft.definition.dig('item', 4, 'text')).to eq('What about this long paragraph?')
+
+        # todo @martha - add more item types
+      end
+
+      it 'reorders items' do
+        assert_text 'Group: Test Custom Assessment'
+        find("div[aria-label='item section_1']").click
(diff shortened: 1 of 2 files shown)


drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,179 @@
...
+
+        advanced_draft.reload
+        item = advanced_draft.definition.dig('item', 0, 'item', 2)
+        autofill = item.dig('autofill_values', 0)
+
+        expect(autofill.dig('value_code')).to eq('filled')
+        expect(autofill.dig('autofill_when', 0, 'question')).to eq('yes_or_no')
+
+        # todo @martha - this is a bug
+        expect(item.dig('custom_rule', 'value')).to eq('some-particular-project-id')
```
martha: @gigxz it's a good thing you called this out on the issue description; it turns out the form builder does clobber `custom_rule` and `rule` and, I believe, any other property that's allowed on the form definition json schema but not present on the `FormItem` graphql type. 

I'm a bit torn between possible approaches,
- we could add all possible properties to the form item [schema object](https://github.com/greenriver/hmis-warehouse/blob/11be216286e6b81d4e1367813d8ea53bb3a69b59/drivers/hmis/app/graphql/types/forms/form_item.rb), even if they may not be edited by the form builder. This could fail and silently start to clobber properties if we were to ever add a new property to the schema and forget to add it to the input object, but that seems unlikely?
- Maybe we should make it the job of the [UpdateFormDefinition mutation](https://github.com/greenriver/hmis-warehouse/blob/11be216286e6b81d4e1367813d8ea53bb3a69b59/drivers/hmis/app/graphql/mutations/update_form_definition.rb) to merge the old and new definitions? But to make sure it doesn't re-merge properties that have been intentionally deleted, we would need a list of properties that are in the json schema but not the FormItem. And then we would be back to needing to remember to add any new properties in 2 places, same concern as above. We could generate this list programmatically, I guess, but I'm not sure that seems worthwhile.

martha: Edit: Gig created a new ticket https://github.com/open-path/Green-River/issues/6555, we can consider this out of scope for this PR

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+RSpec.feature 'HMIS Form Builder', type: :system do
+  include_context 'hmis base setup'
+
+  let!(:ds1) { create(:hmis_data_source, hmis: 'localhost') }
+  let!(:access_control) { create_access_control(hmis_user, p1) }
+  let!(:published) { create :custom_assessment_with_custom_fields_and_rules, title: 'System Test 1', identifier: 'system_test_1', data_source: ds1 }
+  let!(:draft) { create :custom_assessment_with_custom_fields_and_rules, title: 'System Test 2', identifier: 'system_test_2', status: 'draft', data_source: ds1 }
+  let!(:advanced_draft) { create :custom_assessment_with_field_rules_and_autofill, identifier: 'advanced_form', status: 'draft', data_source: ds1 }
+
+  context 'Form builder happy path' do
```
gigxz: I woud suggest encapsulating the fixtures inside `context`s, so the tests are easier to read. That way, there are only 1 or 2 fixtures at play for a given test, and we don't need to worry about mapping back the identifiers like "system_test_1" vs "system_test_2" etc.


For example:

```
context 'with no existing form' do
   it 'creates a new draft' do
   end
end

context 'with an existing published form' do
   let!(:published) { create :custom_assessment_with_custom_fields_and_rules, data_source: ds1 }
     before(:each) do
        visit "/admin/forms/#{published.identifier}"
      end
   it 'creates a new draft' do
   end
end

context 'with an existing draft form' do
   let!(:published) { create :custom_assessment_with_custom_fields_and_rules, status: 'published', data_source: ds1 }
   let!(:draft) { create :custom_assessment_with_custom_fields_and_rules, identifier: published.identifier, status: 'draft', data_source: ds1 }

     before(:each) do
        visit "/admin/forms/#{draft.identifier}"
      end

   it 'does not create a new draft' do
   end

   it 'publishes the draft and retires the old form' do
   end
end
```

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+      click_button 'New Form'
+
+      assert_text 'Form Type'
+      mui_select 'Custom assessment', from: 'Form Type'
+      fill_in 'Form Title', with: 'System Test 3'
+      fill_in 'Form Identifier', with: 'system_test_3'
+      click_button 'Save'
+
+      assert_text 'SELECTED FORM System Test 3'
+      assert_text 'Edit Draft'
```
gigxz: Suggest wrapping with an expectation block, expecting `Hmis::Form::Definition.count` to change by 1?

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+    end
+
+    it 'creates a new draft' do
+      visit '/admin/forms/system_test_1'
+      click_button 'New Draft'
+
+      assert_text 'EDITING DRAFT System Test 1'
```
gigxz: same here with asserting that draft count changed by 1, and that the published form is still published

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+    end
+
+    context 'with an existing form draft' do
+      before(:each) do
+        visit "/admin/forms/system_test_2/#{draft.id}/edit"
+      end
+
+      it 'edits a form item' do
+        find("button[aria-label='section_1 item actions']").click
+        find('li', text: 'Edit', class: 'MuiMenuItem-root').trigger('click')
```
martha: Maybe it's confusing to add the  class here-- This was to disambiguate; capybara returns an ambiguity error without it. I was not sure what capybara thought was ambiguous, since I couldn't see any other obvious `li` with text "Edit" on the page

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+        assert_text 'Group: Test Custom Assessment'
+        click_button 'Preview / Publish'
+        assert_text 'PREVIEWING DRAFT System Test 2'
+        click_button 'Publish'
+        assert_text 'Are you sure you want to publish this form?'
+        click_button 'Confirm'
+        assert_text 'SELECTED FORM System Test 2'
+        assert_text 'Status Published'
+        draft.reload
+        expect(draft.status).to eq('published')
```
gigxz: Neater way is to expect `.to change(draft, :status).to('published')`

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+        find("button[type='submit']").trigger('click')
+
+        draft.reload
+        expect(draft.definition.dig('item', 0, 'item').size).to eq(1)
+      end
+
+      it 'publishes form' do
+        assert_text 'Group: Test Custom Assessment'
+        click_button 'Preview / Publish'
+        assert_text 'PREVIEWING DRAFT System Test 2'
```
gigxz: I think test becomes more maintainable if we use the fixture instead of repeating the title throughout:
`assert_text "PREVIEWING DRAFT #{definition.title}"`

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,257 @@
...
+        click_button 'Publish'
+        assert_text 'Are you sure you want to publish this form?'
```
gigxz: I think we can skip this assertion, and just expect that the "confirm" button is present. (In case we change the wording of this in the frontend, it doesn't need to break the test).

drivers/hmis/spec/system/hmis/form_builder_spec.rb
martha: This works with `reload`: `change { draft.reload.status }.to('published')`

drivers/hmis/spec/system/hmis/form_builder_spec.rb
```diff
@@ -0,0 +1,285 @@
...
+        expect(draft.definition.dig('item', 4, 'text')).to eq('What about this long paragraph?')
+        expect(draft.definition.dig('item', 5, 'text')).to eq('Which option?')
+        expect(draft.definition.dig('item', 6, 'text')).to eq('What day?')
+        expect(draft.definition.dig('item', 7, 'text')).to eq('How many?')
+        expect(draft.definition.dig('item', 8, 'text')).to eq('Yes or no?')
+      end
+
+      it 'reorders items' do
+        assert_text 'Group: Test Custom Assessment'
+        find("div[aria-label='item section_1']").click
```
martha, line 171: I'm not sure, but I think the test failures here are due to the ci running against a frontend based on the `stable` branch instead of `release-130`. Is there anything we should do to update that?

gigxz: Hmm, yeah.
`BRANCH_NAME: "${{ github.head_ref || github.ref_name }}:${{ github.base_ref || 'stable' }}:stable"`
is resolved as
`BRANCH_NAME: martha/6340-formbuilder-sys-test:stable:stable`

Which isn't right. We want `github.base_ref` to resolve `release-130` since the PR is opened against that. The problem is that this github action is trigger on push (NOT pull_request), so [base_ref](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/accessing-contextual-information-about-workflow-runs#github-context) isn't working in the way we want.

I can open a ticket, but as a workaround, you can open a branch called `martha/6340-formbuilder-sys-test` on the frontend and push it, and then re-run the tests on CI.

gigxz: https://github.com/open-path/Green-River/issues/6566

martha: Ok, makes sense. Thank you for creating the ticket!

gigxz (approved):
💯

PR #4628, opened 2024-08-20
Skip most validations for QAs from Epic to match old behavior

awisema:

## Description

The Epic QA importer relies on not enforcing validations on QAs, but that conflicts with using validations to only allow QAs within enrollment windows. This disables the other validations if the QA is coming from Epic.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/models/health/qualifying_activity.rb
+++ b/app/models/health/qualifying_activity.rb
@@ -191,6 +191,7 @@ module Health
 
     def contact_required?
       return false unless activity
+      return false if from_epic?
 
       !activity.to_sym.in?(qa_version.class::CONTACTLESS_ACTIVITIES)
     end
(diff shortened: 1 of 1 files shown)


app/models/health/qualifying_activity.rb
```diff
@@ -275,7 +277,13 @@ def client_reached_title key
       client_reached[key&.to_sym].try(:[], :title) || key
     end
 
+    def from_epic?
+      source_type == 'Health::EpicQualifyingActivity'
+    end
+
     def mode_of_contact_is_other?
+      return false if from_epic?
```
ttoomey, line 285: changing mode_of_contact_is_other also impacts the [display_sections method](https://github.com/greenriver/hmis-warehouse/blob/c7705c39ed69c782e2da319041112834b57a56c1/app/models/health/qualifying_activity.rb#L304) - there's also a view that seems to reference it ( app/views/health/sdh_case_management_notes/_activities.haml)

app/models/health/qualifying_activity.rb
```diff
@@ -284,6 +292,8 @@ def mode_of_contact_other_value
     end
 
     def reached_client_is_collateral_contact?
+      return false if from_epic?
```
ttoomey, line 295: reached_client_is_collateral_contact is referenced in views:
 app/views/health/sdh_case_management_notes/_activities.haml
app/models/health/qualifying_activity.rb

ttoomey (approved):
Caveat: I don't know anything about bhchp, I'm reviewing as this as it's an urgent issue

I think the changes to skip validations look resonable. 

The only thing I noticed is that `contact_required?`, `mode_of_contact_is_other` and `reached_client_is_collateral_contact` are also called in the views and display methods. So the changes to those methods could have side-effects beyond validation.  It's not clear to me if this is actually a problem, I just wanted to flag the side-effect

eanders (approved):
This looks generally good.  Thanks @ttoomey for the thorough review.  I'm guessing that might be important on the view side.  Can we determine `mode_of_contact_is_other?` for Epic in a different way? Same with `reached_client_is_collateral_contact?`.  Or maybe those are never true for Epic QA?

app/models/health/qualifying_activity.rb
awisema: Yes, these are also used to control visibility of the contingent data, but since, for an Epic QA, the contingent data will always be blank, I don't think that is an issue.

awisema:
@eanders I merged this so it doesn't get lost in the deployment cycle -- but will look next week -- Seems like I could just break out the validation methods so the view behavior doesn't change

PR #4629, opened 2024-08-20
Make referral posting endpoint idempotent

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6499
This PR makes the referral posting endpoint idempotent, and adds a spec test.

Questions about API surface:
- 'Posting already exists with a different project' - is it OK to introduce this new language to the API? Any updates to the wording?
- 'Referral Created' is still returned even if the referral hasn't been created because the referral posting already existed. Should we change this, and if so, to what?

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
+++ b/drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
@@ -33,6 +33,12 @@ module HmisExternalApis::AcHmis
       # transact assumes we are only mutating records in the warehouse db
       record = nil
       HmisExternalApis::AcHmis::Referral.transaction do
+        # Idempotent - Check for existing posting and return early if it exists
+        posting = existing_posting
+        return [posting.referral, errors] if posting
+        # Return early in case of errors too, so we don't proceed with trying to create the referral
+        return [nil, errors] unless errors.empty?
+
         referral = find_or_create_referral
         raise ActiveRecord::Rollback unless referral
 
@@ -45,6 +51,18 @@ module HmisExternalApis::AcHmis
         record = referral
       end
       [record, errors]
+    end
+
+    def existing_posting
+      existing = HmisExternalApis::AcHmis::ReferralPosting.find_by(identifier: params.fetch(:posting_id))
+
+      if existing.present?
+        project = mper.find_project_by_mper(params.fetch(:program_id))
+        return error_out('Project not found') unless project
+        return error_out('Posting already exists with a different project') unless existing.project == project
+      end
+
+      existing
     end
 
     def find_or_create_referral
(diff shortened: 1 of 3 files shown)


drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
```diff
@@ -47,6 +53,18 @@ def perform_in_tx(params:)
       [record, errors]
     end
 
+    def existing_posting
+      existing = HmisExternalApis::AcHmis::ReferralPosting.find_by(identifier: params.fetch(:posting_id))
+
+      if existing.present?
+        project = mper.find_project_by_mper(params.fetch(:program_id))
+        return error_out('Project not found') unless project
+        return error_out('Posting already exists with a different project') unless existing.project == project
```
gigxz, line 64: We should also make sure the referral identifier matches, just in case. Untested:
```
return error_out('Posting already exists with a different Referral ID') unless existing.referral.identifier == params.fetch(:referral_id)
```

drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/create_referral_job.rb
```diff
@@ -33,6 +33,12 @@ def perform_in_tx(params:)
       # transact assumes we are only mutating records in the warehouse db
       record = nil
       HmisExternalApis::AcHmis::Referral.transaction do
+        # Idempotent - Check for existing posting and return early if it exists
+        posting = existing_posting
+        return [posting.referral, errors] if posting
+        # Return early in case of errors too, so we don't proceed with trying to create the referral
+        return [nil, errors] unless errors.empty?
```
gigxz: This doesn't really -need- to be inside the transaction, since it's not writing anything to the database. Might be clearer to move it out, so it's more obvious that its not writing.

gigxz (commented):
Looks good! 

Yes, I think it would be a good idea to return a different message- maybe `"Referral Found"` instead of `"Referral Created"`? It would help us keep track of the frequency of this occurring (because our response gets logged in `ExternalRequestLog`).

You might be able to figure that out from the controller with something like `referral.identifier_previously_changed?` ([docs](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html#method-i-2A_previously_changed-3F)), if not then just returning the success message from `CreateReferralJob` sounds good.

martha:
> You might be able to figure that out from the controller with something like referral.identifier_previously_changed?

Cool idea, I tried this out here: https://github.com/greenriver/hmis-warehouse/commit/8f19a96b64562e3f057aac01b18c997556588340. But, `referral.identifier_previously_changed?` will also be `false` when posting to an existing closed referral. In my understanding, 'Referral Created' is still correct to return in this case, not 'Referral Found'.

gigxz (approved):
looks good!

PR #4633, opened 2024-08-21
Allow multiple versions of external forms

gigxz:

## Description

A few form fixes to support PIT now that we live in a versioned-forms world.

1. Remove uniqueness constraint on `external_form_object_key`. It's OK if there are multiple versions of the same form (i.e. same identifier) that share the same `external_form_object_key`. It's not OK if there are forms with different identifiers that share the same `external_form_object_key`. For some reason I was coming up a blank trying to express that as a db constraint, but I added a rails validation method, which I think is enough (and easier to change).
2. Require forms to be `published` (ideally via Form Builder) before they can be published to S3. That way we are relying on our usual publish validation and cded generation.
3. *[Not related to External Forms]*: Don't do role-based validation when creating a new form. This fixes a bug that occurs when you attempt to create a new `Intake` form (for example). It would show an error because the form doesn't yet have all the fields required for HUD intakes. It should be safe to remove this because we validate again on publish.

## Type of change
- [x] Bug fix

### DIFF
--- a/db/warehouse/migrate/20240821180638_update_external_form_object_key_index.rb
+++ b/db/warehouse/migrate/20240821180638_update_external_form_object_key_index.rb
@@ -1,0 +1,5 @@
+class UpdateExternalFormObjectKeyIndex < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :hmis_form_definitions, name: "index_hmis_form_definitions_on_external_form_object_key"
+  end
+end
--- a/drivers/hmis/app/models/hmis/form/definition.rb
+++ b/drivers/hmis/app/models/hmis/form/definition.rb
@@ -114,7 +114,6 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   ].freeze
 
   validates :role, inclusion: { in: FORM_ROLES.map(&:to_s) }
-  validates :identifier, uniqueness: { scope: :version }
 
   ENROLLMENT_CONFIG = {
     owner_class: 'Hmis::Hud::Enrollment',
--- a/drivers/hmis/app/models/hmis/form/definition.rb
+++ b/drivers/hmis/app/models/hmis/form/definition.rb
@@ -325,6 +325,16 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
     where(status: PUBLISHED)
   end
 
+  validate :validate_external_form_object_key
+  def validate_external_form_object_key
+    return unless role == 'EXTERNAL_FORM' && external_form_object_key.present?
+
+    # Ensure that external_form_object_key is not reused across forms with different identifiers
+    object_key_taken = Hmis::Form::Definition.where(external_form_object_key: external_form_object_key).
+      where.not(identifier: identifier).exists?
+    errors.add(:external_form_object_key, 'has already been taken') if object_key_taken
+  end
+
   def self.apply_filters(input)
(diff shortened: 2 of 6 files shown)


db/warehouse/migrate/20240821180638_update_external_form_object_key_index.rb
```diff
@@ -0,0 +1,5 @@
+class UpdateExternalFormObjectKeyIndex < ActiveRecord::Migration[7.0]
+  def change
+    remove_index :hmis_form_definitions, name: "index_hmis_form_definitions_on_external_form_object_key"
```
ttoomey: this is fine, maybe should use the key to allow this to rollback if needed (not tested)
```suggestion
    remove_index :hmis_form_definitions, [:external_form_object_key]
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -325,6 +325,16 @@ def self.for_project(project:, role:, service_type: nil)
...
 
+  validate :validate_external_form_object_key
+  def validate_external_form_object_key
+    return unless role == 'EXTERNAL_FORM' && external_form_object_key.present?
+
+    # Ensure that external_form_object_key is not reused across forms with different identifiers
+    object_key_taken = Hmis::Form::Definition.where(external_form_object_key: external_form_object_key).
+      where.not(identifier: identifier).exists?
+    errors.add(:external_form_object_key, 'has already been taken') if object_key_taken
+  end
```
ttoomey, line 335: This is fine. Maybe could be expressed more concisely with [active record validations](https://guides.rubyonrails.org/active_record_validations.html#uniqueness) but I haven't tested it
```suggestion
  validates :external_form_object_key, uniqueness: { scope: :identifier},
    if: -> { role == 'EXTERNAL_FORM' && external_form_object_key.present? }
```

ttoomey (approved):
looks great, thanks!

db/warehouse/migrate/20240821180638_update_external_form_object_key_index.rb
gigxz: Thanks, good idea. Updated and tested reversability

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: I tried to come up with something using active record validations, but couldn't get there. Your example doesn't quite hit the mark, because there CAN be multiple `external_form_object_key`s with the same identifier, they just can't be reused _across_ identifiers. (At least, I think that makes sense. It could get very confusing if you have different forms that both publish to the same place).

Valid:
```
my-identifier,object_key_1,retired,v1
my-identifier,object_key_1,published,v2
my-identifier,object_key_1,draft,v3
```

Invalid:
```
identifier-1,object_key_1,published,v1
identifier-2,object_key_1,draft,v2
```

gigxz:
Test is failing due to running against stable (https://github.com/open-path/Green-River/issues/6566)

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -114,7 +114,6 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   ].freeze
 
   validates :role, inclusion: { in: FORM_ROLES.map(&:to_s) }
-  validates :identifier, uniqueness: { scope: :version }
```
gigxz, line 117: Not needed because its a db constraint now

PR #4634, opened 2024-08-21
Collect w4 and w6 at annual for HOPWA projects

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6463

This PR:
- Factors out w4 and w6 into their own fragment, `hopwa_disability.json`
- Creates a new fragment `hopwa_disability_for_annual.json`, which incorporates that fragment and patches in a hidden item that sets `q_4_08_2` to `YES` so that the w4 and w6 questions can appear even when the rest of the disability table isn't part of the annual assessment.
  - This is fine to autofill since it's assumed to be true for anyone enrolled in a HOPWA program
  - This won't autofill in non-HOPWA projects, since the whole group is disabled unless `projectFunderComponents` includes `HUD: HOPWA`, because of the hud rules set in `hud_assessment_form_rules_2024.rb`.
  - Note, I used `initial` here rather than `autofill`, which didn't work (I didn't take the time to deeply investigate why not)

Question:
- Should we also add a hidden item to collect the overall disability status? Based on the AAQ response, it sounds like this isn't required, but would it be best practice?
- I have a sort of vague idea that the initial value in `q_4_08_2` could cause more problems later, once we enable expanding on HUD rules with custom rules in the form builder. If someone makes a custom rule that enables the HOPWA fragment in a non-HOPWA project, then we could be auto-collecting a YES value in a hidden field when we really should be providing a dropdown for it. Maybe that scenario is vanishingly unlikely?

## Type of change
- [x] Bug fix - it was a bug that we didn't collect these previously
- [ ] Dependency update


drivers/hmis/lib/form_data/default/fragments/hopwa_disability.json
```diff
@@ -0,0 +1,138 @@
...
+        "record_type": "DISABILITY_GROUP",
+        "field_name": "tCellCountAvailable"
+      },
+      "enable_behavior": "ALL",
+      "enable_when": [
+        {
+          "question": "q_4_08_2",
+          "operator": "EQUAL",
+          "answer_code": "YES"
+        }
```
gigxz: I agree that having the enable_when conditions across fragments is not ideal.

I think we can simplify this a lot if we assume that w4 and w6 questions should ALWAYS appear for HOPWA programs, which I think would be fine (especially since they all have data-not-collected options). With that approach, we can remove all the `enable_when` conditions in the hopwa_disability fragment.

drivers/hmis/lib/form_data/default/fragments/hopwa_disability_for_annual.json
```diff
@@ -0,0 +1,24 @@
+{
+  "fragment": "#hopwa_disability",
+  "item": [
+    {
+      "type": "CHOICE",
+      "link_id": "q_4_08_2",
```
gigxz: RE: other comment for removing enable_when's, we can give this a unique link_id for clarity, like `q_4_08_2_hiv_aids_hopwa`. I think we still want this, to ensure that `DisabilityResponse:1` gets set. (The alternative would be to infer it in the processor if any of the hiv aids fields are present).

drivers/hmis/lib/form_data/default/fragments/hopwa_disability.json
martha: Sounds good, I removed the `enable_when` conditions that referred out to a different fragment. I kept the ones that are internal to this fragment.

drivers/hmis/lib/form_data/default/fragments/hopwa_disability_for_annual.json
martha: Sounds good and makes sense. After submitting an annual assessment locally in a hopwa project, I checked in the Disabilities table and can see a row where DisabilityType is 8, DisabilityResponse is 1, DataCollectionStage is 5 (annual), and the hopwa questions (TCellCount, TCellSource etc) match what I entered, so I think this is working correctly. 👍

drivers/hmis/lib/form_data/default/fragments/hopwa_disability_for_annual.json
```diff
@@ -0,0 +1,24 @@
...
+      "link_id": "q_4_08_2_hiv_aids_hopwa",
+      "text": "HIV/AIDS",
+      "hidden": true,
+      "mapping": {
+        "record_type": "DISABILITY_GROUP",
+        "field_name": "hivAids"
+      },
+      "initial": [
+        {
+          "_comment": "Set to YES so that HOPWA questions' enable_when conditions are fulfilled, even though Annual does not collect the full disability table",
```
gigxz:
```suggestion
          "_comment": "Set to YES so that DisabilityResponse is stored as YES for HIV/AIDS Disability record, even though Annual does not collect the full disability table. This fragment is only included for HOPWA projects.",
```

gigxz (approved):
LGTM lets merge into 130

PR #4648, opened 2024-08-27
6399-hopwa-caper-report

ttoomey:

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6399)

## Type of change


drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/ethnicity_filter.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HopwaCaper::Generators::Fy2024::EnrollmentFilters
+  EthnicityFilter = Struct.new(:label, :code_name, keyword_init: true) do
+    def apply(scope)
+      code = HudUtility2024.race_field_name_to_id.fetch(code_name)
+      # Hisp and any other race/ethnicity
+      scope.where("races @> #{SqlHelper.quote_sql_array([code], type: :integer)}")
```
eanders: This is fine, but we do have the [ActiveRecordExtended](https://github.com/GeorgeKaraszi/ActiveRecordExtended?tab=readme-ov-file#contains) gem in the app, which I _think_ can do this.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HopwaCaper
+  class Enrollment < GrdaWarehouseBase
+    include SqlHelper
+    self.table_name = 'hopwa_caper_enrollments'
+
+    has_many :services, class_name: 'HopwaCaper::Service', foreign_key: [:hud_enrollment_id, :data_source_id, :report_instance_id], primary_key: [:hud_enrollment_id, :data_source_id, :report_instance_id]
```
eanders: Usually we just use the primary key from the enrollment in `hopwa_caper_enrollments` to make this connection.  Is there a reason you went with the composite key?   

Additionally, we usually just store `GrdaWarehouse::Hud::Enrollment.id` in the `hopwa_caper_enrollments` table to make the join back to the source data easier.  (Same with the service record, it can just store `GrdaWarehouse::Hud::Service.id`)

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+      all.map do |record|
+        ::HudReports::UniverseMember.new(
+          universe_membership_type: sti_name,
+          universe_membership_id: record.id,
+        )
+      end
+    end
+
+    scope :overlapping_range, ->(start_date:, end_date:) {
+      where('entry_date <= :end_date AND (exit_date >= :start_date OR exit_date IS NULL)', start_date: start_date, end_date: end_date)
```
eanders: We've worked really hard to keep raw SQL out of the code base unless absolutely necessary.  This is a pattern we have in many classes already, can we use arel for this? 
Cribbed from service history enrollment:
```
scope :overlapping_range, ->(start_date:, end_date:) do
  at = arel_table  
  where(at[:exit_date].gteq(start_date).or(at[:exit_date].eq(nil)).and(at[:entry_date].lteq(end_date)))
end
```

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    scope :latest_by_personal_id, -> {
+      select('DISTINCT ON (hopwa_caper_enrollments.data_source_id, hopwa_caper_enrollments.hud_personal_id) *').order(data_source_id: :asc, hud_personal_id: :asc, entry_date: :desc, id: :desc)
```
eanders: We also have the [ActiveRecordDistinctOn](https://github.com/alecdotninja/active_record_distinct_on) gem.   I believe this can be written as:
```
distinct_on(:data_rouce_id, :hud_personal_id).order(data_source_id: :asc, hud_personal_id: :asc, entry_date: :desc, id: :desc)
```

However, I think you should be storing the `client.id` and using that instead of needing the composite of `data_source_id` and `hud_personal_id`.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+      :PrivateDisability,
+      :VADisabilityService, :VADisabilityNonService,
+      :ChildSupport, :Alimony,
+      :WorkersComp,
+      :Unemployment,
+      :OtherIncomeSource
+    ].freeze
+    def self.from_hud_record(enrollment:, report:)
+      project = enrollment.project
+      hiv_disabilities = enrollment.disabilities.filter(&:hiv?).sort_by(&:id)
```
eanders, line 72: Not sure if the intention is to order these newest first (or last) but sorting by ID won't guarantee that in any way, you probably want to sort by `InformationDate` (or `information_date`)

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+        next unless record.InformationDate.in?(report_date_range)
+        INCOME_SOURCE_FIELDS.filter { |field| record[field] == 1 }
+      end
+      medical_insurance_types = enrollment.income_benefits.flat_map do |record|
+        next unless record.InformationDate.in?(report_date_range)
+        INSURANCE_FIELDS.filter { |field| record[field] == 1 }
+      end
+
+      exit = enrollment.exit if enrollment.exit&.exit_date&.<= report.end_date
+      client = enrollment.client
```
eanders: This is going to give you the source client, which is probably fine, but most (all?) other HUD reports operate on the destination client for demographic questions.  At the very least, you should store `warehouse_client_id: client.destination_client.id` in the enrollment for easy lookup/linking.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+        hud_household_id: enrollment.household_id,
+
+        first_name: client.first_name,
+        last_name: client.last_name,
+
+        age: client.age_on([report.start_date, enrollment.entry_date].max),
+        dob: client.DOB,
+        dob_quality: client.DOBDataQuality,
+        genders: client.gender_multi.sort,
+        races: client.race_multi.sort,
```
eanders: These items should all come from the destination client as that record will sometimes have better quality data than the source data from a given source.

eanders, line 101: This is probably fine, with the 2024 standards we've been moving toward storing each of the gender and race columns in the report tables as well, but I don't think there's any particular reason we can't store the integer representations.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+        veteran: client.veteran_status == 1,
```
eanders: This is also fine, but there is a method that does the same thing:
```
veteran: client.veteran?
```

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+        percent_ami: enrollment.percent_ami,
+
+        relationship_to_hoh: enrollment.relationship_to_hoh,
+        hud_project_id: project.project_id,
```
eanders: Might be cleaner to just store the `.id` of the project so we don't need composite joins to get back to the project.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+        project_funders: project.funders.map(&:funder).compact.sort,
+        entry_date: enrollment.entry_date,
+        exit_destination: exit&.destination,
+        exit_date: exit&.exit_date,
+        housing_assessment_at_exit: exit&.housing_assessment,
+        subsidy_information: exit&.enrollment,
```
eanders: What is this doing?  It looks like it's adding the enrollment only if the exit is present, but the whole enrollment?

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,130 @@
...
+    end
+
+    def self.detail_headers
+      special = ['hud_personal_id', 'first_name', 'last_name']
+      remove = ['id', 'created_at', 'updated_at']
+      cols = special + (column_names - special - remove)
+      cols.map do |header|
+        label = case header
+        when 'client_id'
+          'Warehouse Client ID'
```
eanders: This is the source client ID.  See the note above about storing the destination/warehouse client ID.

drivers/hopwa_caper/app/models/hopwa_caper/filters/hopwa_caper_filter.rb
```diff
@@ -0,0 +1,13 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HopwaCaper::Filters
+  class HopwaCaperFilter < ::Filters::HudFilterBase
+    def funder_ids
+      HudUtility2024.funder_components.fetch('HUD: HOPWA')
+    end
+  end
```
eanders: This is probably fine, but it is a pain point we've fought to avoid in the past.  Rather than extending `Filters::HudFilterBase` you can just set the `funder_ids` attribute.  Again, maybe it's fine, but we had an explosion of filter classes that ended up being really hard to maintain so we've tried to keep it just to the base filter and HUD base filter... somewhat unsuccessfully.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/age_filter.rb
eanders, line 1: I really like this pattern, at some point I'd like to revisit `Filter::FilterScopes` with something like this.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,162 @@
...
+
+    def self.default_project_type_codes
+      HudUtility2024.path_project_type_codes
+    end
+
+    def initialize(report)
+      super(report)
+      return unless report&.persisted?
+
+      Rails.logger.level = 1 if Rails.env.development?
```
eanders: This appears to be equivalent (and I think more clear what's going on)
```suggestion
      Rails.logger.level = :info if Rails.env.development?
```

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,162 @@
...
+      ]
+    end
+
+    protected
+
+    def service_history_enrollments
+      scope = GrdaWarehouse::ServiceHistoryEnrollment.
+        entry.
+        open_between(start_date: @report.start_date, end_date: @report.end_date)
+      # FIXME- I don't know how to create these from a spec
```
eanders: Generally we import an HMIS CSV set and post-process it something like the following from `drivers/hmis_csv_importer/spec/models/importer/twenty_twenty_four/base_spec.rb`.  Generating the source for the CSVs might be easiest through the HMIS and using the HMIS CSV exporter.

```
import_hmis_csv_fixture(
        'drivers/hmis_csv_importer/spec/fixtures/files/twenty_twenty_four/enrollment_test_files',
        version: 'AutoMigrate',
        run_jobs: true,
      )
```

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,162 @@
...
+          hud_enrollment = service_history_enrollment.enrollment
+
+          enrollment_rows << HopwaCaper::Enrollment.from_hud_record(report: report, enrollment: hud_enrollment)
+          service_rows += hud_enrollment.services.map do |hud_service|
+            HopwaCaper::Service.from_hud_record(report: report, enrollment: hud_enrollment, service: hud_service)
+          end
+        end
+
+        populate_hopwa_eligible(enrollment_rows)
+        ensure_uniform_client_attrs(enrollment_rows)
```
eanders: I think this is only ensuring within the batch, if a client exists in multiple batches, are they also being "made uniform"?  You might need to adjust the service history scope to batch by batches of client_ids to ensure a client is fully accounted for in each batch.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,162 @@
...
+        group.each { |e| e.attributes = uniform_attrs }
+      end
+    end
+
+    # try and figure out which person in a household is hopwa eligible
+    def populate_hopwa_eligible(enrollment_rows)
+      households = enrollment_rows.sort_by(&:id).group_by { |e| [e.data_source_id, e.hud_household_id] }.values
+      eligible_enrollments = households.map do |enrollments|
+        # if the hoh is hiv+
+        hohs = enrollments.filter { |e| e.relationship_to_hoh == 1 }
```
eanders, line 162: Might as well lean on existing helper methods.
```suggestion
        hohs = enrollments.filter(&:head_of_household?)
```

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,162 @@
...
+          ever_perscribed_anti_retroviral_therapy: groups.any?(:ever_perscribed_anti_retroviral_therapy),
+          viral_load_supression: groups.any?(:viral_load_supression),
+        }
+        group.each { |e| e.attributes = uniform_attrs }
+      end
+    end
+
+    # try and figure out which person in a household is hopwa eligible
+    def populate_hopwa_eligible(enrollment_rows)
+      households = enrollment_rows.sort_by(&:id).group_by { |e| [e.data_source_id, e.hud_household_id] }.values
```
eanders: Similar to the issue with batching and uniform client information, there's no guarantee an entire household will be in `enrollment_rows`. I believe you'll need to either do this calculation against the `hopwa_caper_enrollments` table after populating it with most of the data, or you'll have adjust the scope for `enrollment_rows` such that it always includes the entire household.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/ethnicity_filter.rb
ttoomey: Good to know! I'll give it a spin

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: Oh that's probably better, I'll switch to the enrollment pk.

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: yes, I'll make a pass back through and transform any stray sql strings. I find it much easier to work with strings than arel so I tend to start there and transform them before I hand it off

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: on interesting, thanks for pointing out that gem. Yes, client_id would be a simpler discriminator here

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: thanks for pointing that out. I wasn't aiming for chronological order, just deterministic since by default it would be unordered

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: ah yes thanks, I'll switch the the destination client. I had some struggles with that that

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: oh okay, I can use discrete cols for these demographics if that's more standard

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: oh woops, somehow I lost the method name there. Thanks! Not actually using this field anywhere yet

drivers/hopwa_caper/app/models/hopwa_caper/filters/hopwa_caper_filter.rb
ttoomey: got it, I can change that behavior

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: oh yeah, this was just temporary to keep the log chatter down to a minimum, I'll take it out

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: oh good point, I'll check this. Thanks!

ttoomey:
Thanks for the review @eanders!

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
eanders: I actually think this is probably fine, let's leave it as is for now.

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
```diff
@@ -0,0 +1,44 @@
...
+    = link_to(link_params.merge({format: :xlsx}),{class: 'btn btn-secondary'}) do
+      %i.icon-file-excel
+      Download
+
+
+
+- if @clients.blank?
+  .none-found No records found.
+- else
+  .card
```
eanders: It's probably worth adding the live search logic here, see `drivers/core_demographics_report/app/views/core_demographics_report/warehouse_reports/shared/_details.haml` for an example.  (Two bits, the search form and the JS)

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: enrollments here are hopwa cper enrollments, so the helper isn't available

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: seems like the solution we have from yesterday works in the tests for post processing (chronic, service history enrollments, destination clients)

app/views/hud_reports/index.haml
```diff
@@ -1,6 +1,6 @@
 %h1 HUD Reports
 
 - reports = Rails.application.config.hud_reports.values.uniq
-- reports.each do |r|
+- reports.sort_by {|r| r[:title] }.each do |r|
```
ttoomey, line 4: Maybe we don't want this but should we sort the hud reports alphabetically?

lib/util/concerns/hud_lists_2024.rb
```diff
@@ -2016,7 +2016,7 @@ def ssvf_services(id, reverse = false, raise_on_missing: false)
       _translate(ssvf_services_options, id, reverse, raise_on_missing: raise_on_missing)
     end
 
-    # V2.3
+    # W2.2
```
ttoomey, line 2019: I think this had the wrong code

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/prior_living_situation_filter.rb
```diff
@@ -0,0 +1,107 @@
...
+        new(
+          label: 'A permanent housing situation for formerly homeless persons?',
+          types: [
+            'Rental by client, with ongoing housing subsidy',
+          ],
+          rental_subsidy_types: [
+            'VASH housing subsidy',
+            'Permanent Supportive Housing',
+            'Other permanent housing dedicated for formerly homeless persons',
+          ],
```
ttoomey, line 53: seems like rental subsidy could be used this way, could use another set of eyes

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/exit_destination_filter.rb
```diff
@@ -0,0 +1,114 @@
...
+    def self.php_destinations
+      [
+        other_hopwa_program,
+        other_subsidy_program,
+        private_housing,
+      ]
+    end
+
+    def self.all_destinations
+      type_unknown = nil
```
ttoomey, line 65: we don't know how to calculate some questions, leaving as blank

app/models/hud_reports/generator_base.rb
```diff
@@ -42,16 +42,16 @@ def self.report_year_slug
...
     end
 
     def run!(email: true, manual: true)
       @report.state = 'Waiting'
       @report.question_names = self.class.questions.keys
       @report.manual = manual
-      @report.save
-      Reporting::Hud::RunReportJob.perform_now(self.class.name, @report.id, email: email)
+      @report.save!
+      Reporting::Hud::RunReportJob.new.perform(self.class.name, @report.id, email: email)
```
eanders: This makes me slightly nervous.  If memory serves `perform_now` and `new.perform` run in different contexts (ActiveJob vs. DelayedJob, I think).  What's the reasoning behind this change?

app/models/hud_reports/question_base.rb
```diff
@@ -42,10 +42,14 @@ def run!
       remaining_questions = @report.remaining_questions - [self.class.question_number]
       @report.update(remaining_questions: remaining_questions)
     rescue StandardError => e
+      # for debugging sql issues in tests, raise immediately since attempting further updates will crash in failed tx
+      # and we'd like to get the backtrace from the original exception
+      raise if Rails.env.test? && e.is_a?(ActiveRecord::StatementInvalid)
```
eanders, line 47: Nice, thank you!

app/views/hud_reports/index.haml
eanders: This is good, thanks.  I don't think the current sort order was intentional, and once you are in one, the side-menu is sorted alphabetically.

app/models/hud_reports/generator_base.rb
ttoomey: I was having trouble with the job swallowing the stack trace, not quite sure why. I'll change it back. The difference is that invoking it this way should avoid ActiveJob & DelayedJob entirely

eanders: That's totally fine!  Since we were running them `now` anyway, I think it makes more sense to do what you have.  Thanks!

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,140 @@
...
+             -> do
+               where(::HudReports::UniverseMember.arel_table[:universe_membership_type].eq('HopwaCaper::Enrollment'))
+             end,
+             inverse_of: :universe_membership,
+             class_name: 'HudReports::UniverseMember',
+             foreign_key: :universe_membership_id
+    has_many :services, class_name: 'HopwaCaper::Service', primary_key: :enrollment_id
+
+    def self.as_report_members
+      all.map do |record|
```
eanders: Isn't the `all` here going to override any calling scope?  
Should this start with something like `current_scope`?  Can you do that in a class method?

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
```diff
@@ -0,0 +1,140 @@
...
+    scope :overlapping_range, ->(start_date:, end_date:) {
+      table = arel_table
+      where(
+        table[:entry_date].lteq(end_date).and(
+          table[:exit_date].gteq(start_date).or(table[:exit_date].eq(nil)),
+        ),
+      )
+    }
+
+    scope :latest_by_personal_id, -> {
```
eanders: This scope name seems disconnected from the contents.  Maybe just `latest_by_id`?  Or `ordered`?

drivers/hopwa_caper/app/models/hopwa_caper/enrollment.rb
ttoomey: hmm, I don't think [scope.all()](https://github.com/rails/rails/blob/fc734f28e65ef8829a1a939ee6702c1f349a1d5a/activerecord/lib/active_record/scoping/named.rb#L22) overrides the calling scope. It's very similar to calling current_scope. I can use current_scope instead if you think it would be more clear

eanders: Interesting.  if `current_scope` works, that's what we've done in the past, so I guess that's preferable, (and wouldn't cause me to think twice about it.)  If it's not available, this is fine as is.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/income_benefit_level_filter.rb
```diff
@@ -0,0 +1,40 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HopwaCaper::Generators::Fy2024::EnrollmentFilters
+  IncomeBenefitLevelFilter = Struct.new(:label, :type, keyword_init: true) do
+    def apply(scope)
+      cond = HopwaCaper::Enrollment.group(:report_household_id).having('MAX(percent_ami) = ?', code)
```
eanders: Calling MAX here will return `99` if Data Not Collected for some clients, does it matter that there might be a 4, or 3, etc. that are then being ignored?  Maybe this should preference collected values somehow?

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/prior_living_situation_filter.rb
eanders: I'm not clear on what the goal is, can you explain a bit more?
It does seem like you can limit enrollments to those with a particular prior living situation and rental subsidy type, if that's all you mean.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/strmu_longevity_filter.rb
```diff
@@ -0,0 +1,59 @@
...
+        'How many households received STRMU assistance more than twice during the previous five years?' => [],
+        'How many households received STRMU assistance during the last five consecutive years?' => [],
+      }
+
+      rows.each do |client_id, entry_dates|
+        # what years are covered by prior enrollments
+        entry_years = entry_dates.map(&:year).uniq.select
+        bucket = nil
+        if entry_years.size >= 5
+          bucket = 'How many households received STRMU assistance during the last five consecutive years?'
```
eanders, line 42: Do we care that this says consecutive, but that entry dates might not be?  
Does each client get a new enrollment every year?  
What happens if a client has a single enrollment that lasts 5 years?

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,186 @@
...
+      report.hopwa_caper_enrollments.distinct.pluck(:report_household_id).in_groups_of(100, false) do |household_ids|
+        enrollments = report.hopwa_caper_enrollments.where(report_household_id: household_ids).order(:id)
+        update_hopwa_eligibility(enrollments)
+      end
+
+      true
+    end
+
+    # ensure consistent values for individuals (can vary based on enrollment entry date)
+    def ensure_uniform_client_attrs(enrollment_rows)
```
eanders, line 139: This is a great pattern.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/income_benefit_level_filter.rb
ttoomey: oh good catch! I'll just filter to pct_api <= 4

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/prior_living_situation_filter.rb
ttoomey: Oh sure, so the question is "A permanent housing situation for formerly homeless persons" which doesn't correspond a prior-living-situation code. But maybe we can infer it with a combination of code 435 (Rental by client, with ongoing housing subsidy) and certain values of RentalSubsidyType (3.917.A). 

Perhaps we just can't answer it though and it should be left as null

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
```diff
@@ -0,0 +1,186 @@
...
+    def self.short_name
+      'HOPWA CAPER'
+    end
+
+    def self.file_prefix
+      "v1.0 #{short_name} #{fiscal_year}"
+    end
+
+    def self.default_project_type_codes
+      HudUtility2024.path_project_type_codes
```
eanders: Pretty sure this should be some other set of project types, maybe PH?

eanders: Or maybe
 ```
HudUtility2024.spm_project_type_codes
```

File: drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb [hunk 31]
```diff
@@ -0,0 +1,186 @@
...
+        user_id: @report.user_id,
+        enforce_one_year_range: false,
+      ).update(@report.options)
+      @filter.funder_ids = HudUtility2024.funder_components.fetch('HUD: HOPWA')
+
+      @filter.project_ids = @report.project_ids
+      @filter.apply(scope)
+    end
+
+    def build_hopwa_caper_models
```
eanders, line 102: I haven't been able to find the `def run!` method that is being called.  Whatever is called from the `RunReportJob` DJ should set the `state` column of the report instance to `Started` and the `started_at` timestamp when the report starts.  (I'm not seeing the running report reflected correctly in the UI, it just says `Queued` but it _is_ running.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: oh thanks, I think that's left-over from when I copied this from the path report. I don't think the method is called though, I'll just take it out

eanders: That gets used if you have never run a HOPWA CAPER to set the default project types that should be included.  It was also limiting which project types were being included, but that might have been an odd side effect.  You can try removing it, but if you do, we'll need to test the scenario where no HOPWA CAPERs have been run before.

app/controllers/document_exports_controller_base.rb
```diff
@@ -115,6 +115,7 @@ def download
       'UserDirectoryReport::DocumentExports::CasUserDirectoryExcelExport',
       'UserDirectoryReport::DocumentExports::WarehouseUserDirectoryExcelExport',
       'TxClientReports::AttachmentThreeReportExports::AttachmentThreeReportExcelExport',
+      'HopwaCaper::DocumentExports::HopwaCaperExport',
```
ttoomey, line 118: allows pdf export to run

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/enrollment_filters/strmu_longevity_filter.rb
ttoomey: Good questions

> Do we care that this says consecutive, but that entry dates might not be?

since we are only looking at the enrollments for last five years, if there are five unique years then they're likely contiguous; I'm sure there are cases where this doesn't hold up though. I'll make it more percise

> Does each client get a new enrollment every year? What happens if a client has a single enrollment that lasts 5 years?

It sounded like strmu enrollments tend to be short-lived. Perhaps we shouldn't assume that though and account for the exit date

eanders: If we're looking at 5 years of data and counting 5 distinct years, that feels fine for "contiguous for 5 years" I was thinking we were looking at the 15 year look back and just counting that they had 5, which wasn't feeling right.

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
eanders: This is working slightly better now, it shows "Started" when it starts, and has the start time, but when it completes it's filling the `started_at` timestamp with a new value making it look like it only took a few seconds to run.  `completed_at` is getting correctly filled, but I think it's updating both columns and should only update `completed_at` when it completes.

app/controllers/document_exports_controller_base.rb
eanders: Works, thanks!

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2024/generator.rb
ttoomey: oh yeah, looks like it resets the started_at on the first question. I adjusted it so that prepare_report marks it as started instead of in report.start(question...)

ttoomey:
I made some adjustments to the demo & prior living filters. Should give more accurate results for the questions about new vs prev years by checking all enrollments 856cb4068a23f71b6f5da7750d57bf9a38a05dc5

eanders (approved):
I think this looks good.

eanders (approved):

gigxz (approved):

eanders:
Waiting on tests to pass, but then let's merge

app/models/hud_reports/report_instance.rb
```diff
@@ -108,26 +108,25 @@ def related_job
     # FIXME: maybe a single question column on report_instance to track if this is a single
     # question run or all questions.... Need better start/complete logic
     def start(question, tables)
-      universe(question).update(status: 'Started', metadata: { tables: Array(tables) })
-      start_report if build_for_questions.count == remaining_questions.count
```
ttoomey, line 112: The report is now marked as started when the report job calls `@generator.prepare_report`. This is intended to be a simplification that lets the job's run-time include time spent in prepare_report (building hopwa caper enrollments). I think this is okay assuming we're always running questions via the job.

eanders: @ttoomey I'm confused, does this mean `start` doesn't call `start_report` ? Where is `start_report` getting called now?

ttoomey: @eanders Yes, report.start only starts the question. The report is marked as started before that: RunReportJob calls generator.prepare_report which calls report.start_report. Happy to huddle about this if it would help

eanders: That's what I was looking for.  Thanks.

ttoomey:
Will merge this soon,  I don't think the CI test failures (in spec/models/health/qualifying_activity_spec.rb) are related

eanders:
Thanks.  Yes, I just pushed fixes to 135 for the failing tests

PR #4653, opened 2024-08-28
Don't assume DisabilityResponse for HOPWA programs at annual

gigxz:

## Description

Fix my incorrect spec for https://github.com/open-path/Green-River/issues/6463. I had incorrectly said that we can assume all clients at HOPWA programs have DisabilityResponse:1 for Disability 8. That turns out not to be true, because households can be enrolled in HOPWA programs with members that do not have HIV/AIDS.

This updates the forms so that Disability 8 is asked at Annual, and W4/W6 are only ever asked if client response "Yes" to Disability 8.

* For HOPWA-funded programs...
   - [x] On Intake/Update/Exit, ask W4/W6 only if client responds "Yes" to HIV/AIDS
   - [x] On Annual, ask  W4/W6 only if client responds "Yes" to HIV/AIDS
* For non-HOPWA-funded programs...
   - [x] Never ask W4/W6

HOPWA questions on Annual:
Screenshot 2024-08-28 at 11 37 48 AM

HOPWA questions on Intake:
Screenshot 2024-08-28 at 11 38 45 AM

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/lib/form_data/default/fragments/disability.json
+++ b/drivers/hmis/lib/form_data/default/fragments/disability.json
@@ -297,7 +297,24 @@
       ]
     },
     {
-      "fragment": "#hopwa_disability"
+      "type": "GROUP",
+      "link_id": "hopwa_section",
+      "text": "HOPWA Questions",
+      "enable_behavior": "ALL",
+      "enable_when": [
+        {
+          "_comment": "q_4_08_2 question may come from disability fragment OR hopwa_disability_for_annual fragment",
+          "question": "q_4_08_2",
+          "operator": "EQUAL",
+          "answer_code": "YES"
+        }
+      ],
+      "item": [
+        {
+          "fragment": "#hopwa_disability",
+          "_comment": "the hopwa_disability fragment also has a rule ensuring that it is only included for HOPWA-funded projects. For non-HOPWA programs, this whole hopwa_section group is excluded (because groups are excluded if all children are excluded)."
+        }
+      ]
     }
   ]
 }
--- a/drivers/hmis/lib/form_data/default/fragments/disability.json
+++ b/drivers/hmis/lib/form_data/default/fragments/disability.json
@@ -303,7 +303,6 @@
       "enable_behavior": "ALL",
       "enable_when": [
         {
-          "_comment": "q_4_08_2 question may come from disability fragment OR hopwa_disability_for_annual fragment",
           "question": "q_4_08_2",
           "operator": "EQUAL",
           "answer_code": "YES"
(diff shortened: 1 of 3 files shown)


drivers/hmis/lib/form_data/default/fragments/disability.json
```diff
@@ -297,7 +297,23 @@
...
+        {
+          "question": "q_4_08_2",
+          "operator": "EQUAL",
+          "answer_code": "YES"
+        }
+      ],
+      "item": [
+        {
+          "fragment": "#hopwa_disability",
+          "_comment": "the hopwa_disability fragment also has a rule ensuring that it is only included for HOPWA-funded projects. For non-HOPWA programs, this whole hopwa_section group is excluded (because groups are excluded if all children are excluded)."
```
martha: I'm confused by this comment since I don't see the `HUD: HOPWA` rule in `hopwa_disability.json`, only `hopwa_disability_for_annual.json`, is that intentional?

martha: oh, I forgot the rule should be coming from `hud_assessment_form_rules_2024.rb` because it is hud-specified, and we don't want users to be able to override it.

In that case I wonder whether the parallel rule applied in `hopwa_disability_for_annual.json` should also be put in that file?

gigxz: Yeah, it could be applied in the ruby file instead of the JSON file. I guess it's a new pattern so I'm not sure where it makes sense to live.

The HudAssessmentFormRules2024 class is mostly for ensuring that the necessary Data Elements are present on HUD Assessments with the right Rules, at the correct stages. It's already validating that W4 and W6 must be present on HUD Annuals (with the HOPWA rule), since you added the annual stage to the `stages` configuration. So, we wouldn't really "get" any additional validation by adding a rule for "hopwa_disability_for_annual". That was my thought process, but I could be convinced.

martha: Yeah, I agree we shouldn't add the rule in two places. Maybe my suggestion is just to update the comment here, for example "the hopwa disability fragment also has a rule ensuring..." -> "the hopwa_disability fragment also has a rule (defined in code, not json) ensuring..." or something similar? But I also could be convinced, maybe what you have is fine and clear enough.

martha (approved):

drivers/hmis/lib/form_data/default/fragments/disability.json
gigxz: Alright I updated the comments, I think it should be clearer now https://github.com/greenriver/hmis-warehouse/pull/4653/commits/b44cdcd3a92961d4c88d72a59c14f90cd5551b06

PR #4654, opened 2024-08-28
Support multiple visibility conditions in HTML External Forms

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6614

Add additional functionality that is needed for PIT:
- [x] Bug: Stop assuming that `link_id` and `custom_field_key` always match
- [x] Support multiple `enable_when` conditions
- [x] Support resolving a `pick_list_reference` when it refers to a HUD enum (eg `NoYesMissing`)
- [x] Support items where `mapping` has `record_type` and/or `field_name` (will be needed because switching to using FormProcessor to process submissions that include related records https://github.com/greenriver/hmis-warehouse/pull/4657)

QA:
- [x] test prevention screening as local HTML form (note: a couple changes need to be made for prevention screening to work, because it has 2 conditionals where it references a dependent `question` by custom_field_key instead of by link_id)
- [x] test PIT as local HTML form

## Type of change
- [x] Bug fix

### DIFF
--- a/config/brakeman.ignore
+++ b/config/brakeman.ignore
@@ -225,7 +225,7 @@
       "check_name": "RegexDoS",
       "message": "Model attribute used in regular expression",
       "file": "app/models/health/qualifying_activity.rb",
-      "line": 68,
+      "line": 70,
       "link": "https://brakemanscanner.org/docs/warning_types/denial_of_service/",
       "code": "/#{[Health::QualifyingActivityV1, Health::QualifyingActivityV2, Health::QualifyingActivityV3].map do\n (version::ATTRIBUTE_SUFFIX + \"$\")\n end.join(\"|\")}/",
       "render_path": null,
@@ -239,29 +239,6 @@
       "cwe_id": [
         20,
         185
-      ],
-      "note": ""
-    },
-    {
-      "warning_type": "File Access",
-      "warning_code": 16,
-      "fingerprint": "13a7f90845cffdc12900858d9d2d0e853dcd03c123ae71ebfe37e725f6d6453a",
-      "check_name": "FileAccess",
-      "message": "Parameter value used in file name",
-      "file": "drivers/client_access_control/app/controllers/client_access_control/history_controller.rb",
-      "line": 119,
-      "link": "https://brakemanscanner.org/docs/warning_types/file_access/",
-      "code": "File.open(Rails.root.join(\"tmp\", \"service_history_pdf_#{::GrdaWarehouse::Hud::Client.destination.find(params[:client_id].to_i).id}.pdf\"), \"wb\")",
-      "render_path": null,
-      "location": {
-        "type": "method",
-        "class": "ClientAccessControl::HistoryController",
-        "method": "pdf"
-      },
-      "user_input": "params[:client_id]",
-      "confidence": "Weak",
-      "cwe_id": [
-        22
       ],
       "note": ""
     },
@@ -663,7 +640,7 @@
       "check_name": "Execute",
       "message": "Possible command injection",
       "file": "drivers/hmis/app/models/hmis/form/definition.rb",
-      "line": 350,
+      "line": 359,
       "link": "https://brakemanscanner.org/docs/warning_types/command_injection/",
       "code": "`No Definition found for System form #{role}`",
       "render_path": null,
@@ -977,6 +954,28 @@
     {
       "warning_type": "Cross-Site Scripting",
       "warning_code": 2,
+      "fingerprint": "4f50bd02cffae24567d5d511f692ceeae1ac8183f4092cb176a63afabf3b18df",
+      "check_name": "CrossSiteScripting",
+      "message": "Unescaped model attribute",
+      "file": "drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form.haml",
+      "line": 4,
+      "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
+      "code": "HmisExternalApis::ExternalForms::FormGenerator.new(self, @form_definition).render_node(node)",
+      "render_path": null,
+      "location": {
+        "type": "template",
+        "template": "hmis_external_apis/external_forms/form"
+      },
+      "user_input": null,
+      "confidence": "High",
+      "cwe_id": [
+        79
+      ],
+      "note": ""
+    },
+    {
+      "warning_type": "Cross-Site Scripting",
+      "warning_code": 2,
       "fingerprint": "51ec304bfb996e9d40ff69207ead86771b77a085b4a1bee4454fa4821f70fa70",
       "check_name": "CrossSiteScripting",
       "message": "Unescaped model attribute",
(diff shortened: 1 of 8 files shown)


config/brakeman.ignore
```diff
@@ -974,6 +951,28 @@
...
+      "message": "Unescaped model attribute",
+      "file": "drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form.haml",
+      "line": 4,
+      "link": "https://brakemanscanner.org/docs/warning_types/cross_site_scripting",
+      "code": "HmisExternalApis::ExternalForms::FormGenerator.new(self, @form_definition).render_node(node)",
+      "render_path": null,
+      "location": {
+        "type": "template",
+        "template": "hmis_external_apis/external_forms/form"
+      },
```
gigxz, line 968: This was already being ignored. The change I made was to pass the `@form_definition` record to the FormGenerator. I'm not really wrapping my head around how this could be vulnerable to XSS, since nothing is coming from the browser (Except in dev mode), we are just generating this HTML using a view. So I think it is a false positive, but I would appreciate a look.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -111,29 +111,59 @@ $(function () {
...
 });
 
-window.addDependentGroup = function (inputName, condValue, targetSelector) {
-  var target = $(targetSelector);
-  var show = function () {
+// conditions: [{input_name: 'name', input_value: 'value'}, ...]
+// targetSelector: selector for the item that is conditionally shown
+// enableBehavior: 'ANY' or 'ALL' conditions must be met to show the target selector
+window.addDependentGroup = function (conditions, targetSelector, enableBehavior = 'ANY') {
+  const target = $(targetSelector); // the item with enable_when on it
```
gigxz: Change back to `var` for highest compatibility with ios strict mode, and others

ttoomey: I see this is rewrite using es5 features; Fair enough, I was trying to avoid es5 when I wrote this initially. Maybe this breaks ie11 (and other older browsers) but that's probably okay as I never got compatibility working with the s3 uploader. I think we should still be cautious about using anything too new here.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -111,29 +111,59 @@ $(function () {
...
+    });
+
+    const meetsCondition = enableBehavior === 'ALL' ? evaluations.every(Boolean) : evaluations.some(Boolean)
+    if (meetsCondition) {
+      show();
     } else {
-      value === condValue ? show() : hide();
+      hide();
     }
+  }
```
ttoomey: * I think using `const` is probably safe but I'd be nervous about using destructuring assignment
* My original code here could have used simplification, I think we can improve this a bit, see below

```js
  const onDependentItemChanged = function () {
    const evaluations = conditions.map(function (condition) {
      const input_name = condition.input_name;
      const input_value = condition.input_value;
      const $el = $('[name="' + input_name + '"]');

      if ($el.is(':radio')) {
        return $('[name="' + input_name + '"]:checked').val() === input_value;
      }

      if ($el.is(':checkbox')) {
        return $el.is(':checked') && $el.val() === input_value;
      }

      return $el.val() === input_value;
    });

    const meetsCondition = enableBehavior === 'ALL' ? evaluations.every(Boolean) : evaluations.some(Boolean);
    meetsCondition ? show() : hide();
  }
```

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -111,29 +111,59 @@ $(function () {
...
+
+  // add change listener to all dependent fields
+  const dependentItemSelectors = conditions.map(c => `[name="${c.input_name}"]`)
+  dependentItemSelectors.forEach(function (name) {
+    $(name).on('change', onDependentItemChanged);
   });
```
ttoomey: not sure about compatibility with fat arrow and template literal. Maybe safer to avoid?
```suggestion
conditions.forEach(function (c) {
  const nameSelector = '[name="' + c.input_name + '"]';
  $(nameSelector).on('change', onDependentItemChanged);
});
```

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
```diff
@@ -111,14 +113,12 @@ def render_display_node(node)
...
+      options = resolve_pick_list(node)
+      raise "missing options in #{node.inspect}" unless options.present?
 
-      options = node['pick_list_options'].map do |option|
-        { label: option['label'], value: option['code'] }
-      end
       render_form_group(node: node) do
         case node['component']
-        when 'DROPDOWN'
+        when 'DROPDOWN', nil
```
ttoomey, line 121: Do we need to have a default? (it's fine if it is, but it does add extra complexity to support it everywhere we deal with `component`)

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
```diff
@@ -154,21 +154,65 @@ def render_form_group(node:, &block)
...
+      pick_list_options = node['pick_list_options']
+
+      if pick_list_reference
+        found_enum = "Types::HmisSchema::Enums::Hud::#{pick_list_reference}".safe_constantize
+        found_enum ||= "Types::HmisSchema::Enums::#{pick_list_reference}".safe_constantize
+        raise "Unable to resolve pick list reference: #{pick_list_reference}" unless found_enum
+
+        found_enum.values.map do |k, v|
+          { value: k.to_s, label: v.description&.gsub(/^\([0-9A-Za-z]+\) /, '') || k.to_s }
+        end
```
ttoomey: There's quite a bit going on here. Is this mirrored in some other part of our form logic? Maybe a comment or something could be helpful

ttoomey (approved):
looks good, thanks!

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: Totally, my comment was a note-to-self to switch back to `var`. So I think we are in agreement

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: Great, thanks for the suggestion! I wasn't sure about the checkbox logic because it wasn't used anywhere in the prevention screening as far as I could tell, and not in PIT either. But I will test it out with this approach.

gigxz: Yeah, I don't think checkbox conditional would even be supported yet because we only support conditionals that use `answer_code`. But I'll keep this code, and we can adjust as needed if/when we add support for conditionals on booleans.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: yep, thanks!

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
gigxz: Yep we need a default, `component` is not required. It looks like all the other types already had a default component behavior, it was only "choice" that didn't.

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
gigxz: Definitely, this mirrors HMIS frontend code. I added comments referencing where this logic is in the frontend: https://github.com/greenriver/hmis-warehouse/pull/4654/commits/e383da36582239a0998fcd3f17f49dbdfacd5cdf

PR #4649, opened 2024-08-27
Add test cases for processing "upgraded" assessment submissions

gigxz:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

Requires frontend PR: https://github.com/greenriver/hmis-frontend/pull/901

* Regression test for bug fixed by https://github.com/greenriver/hmis-frontend/pull/901. The tests were asserting that the correct form definition was used for _rendering_ the DynamicForm/DynamicView, but they were NOT testing the submission process. This PR remediates that.
* Also stop resolving "upgraded definition for editing" for WIP assessments, since they don't need it

## Type of change
- [x] Tests

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4657, opened 2024-08-29
Process FormSubmissions to Clients/Enrollments

martha:

## Description

https://github.com/open-path/Green-River/issues/6546

This PR
- uses the FormProcessor to process external form submissions into CDEs on _review_ of the submission, rather than on consuming the submission
- adds capability to create Client/Enrollment related records when processing the submission
- fixes a few existing bugs with external forms, and adds regression tests

How to test: I tested locally using the external forms UI. I manually seeded the db with an external form that has a data structure like this:

```
{"captcha_score": "1.0", "Client.lastName": "External", "Client.firstName": "Martha", "form_definition_id": "185", "Enrollment.relationshipToHoH": "SELF_HEAD_OF_HOUSEHOLD", "marthas_external_form_question_1": "1234", "marthas_external_form_question_2": "5678"}
```

To match, I also added a form definition like this:

<details>
<summary>Form definition for test</summary>

```
{
  "item": [
    {
      "item": [
        {
          "text": "Question 1",
          "type": "STRING",
          "link_id": "marthas_external_form_question_1",
          "mapping": {
            "custom_field_key": "marthas_external_form_question_1"
          },
          "disabled_display": "HIDDEN"
        },
        {
          "text": "Question 2",
          "type": "STRING",
          "link_id": "marthas_external_form_question_2",
          "mapping": {
            "custom_field_key": "marthas_external_form_question_2"
          },
          "disabled_display": "HIDDEN"
        },
        {
          "text": "First name",
          "type": "STRING",
          "link_id": "first_name",
          "mapping": {
            "field_name": "firstName",
            "record_type": "CLIENT"
          },
          "disabled_display": "HIDDEN"
        },
        {
          "text": "Last name",
          "type": "STRING",
          "link_id": "last_name",
          "mapping": {
            "field_name": "lastName",
            "record_type": "CLIENT"
          },
          "disabled_display": "HIDDEN"
        },
        {
          "text": "HoH",
          "type": "STRING",
          "link_id": "relationship",
          "mapping": {
            "field_name": "relationshipToHoH",
            "record_type": "ENROLLMENT"
          },
          "disabled_display": "HIDDEN"
        }
      ],
      "text": "Group",
      "type": "GROUP",
      "link_id": "group",
      "disabled_display": "HIDDEN"
    }
  ]
}
```

</details>

## Type of change
- [x] Bug fix
- [ ] Dependency update


db/warehouse/migrate/20240829152828_add_to_form_submission.rb
```diff
@@ -0,0 +1,14 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class AddToFormSubmission < ActiveRecord::Migration[7.0]
+  def change
+    add_column :hmis_external_form_submissions, :cleaned_values, :jsonb, null: true
+
+    safety_assured do
+      add_reference :hmis_external_form_submissions, :enrollment, index: true, null: true
```
martha, line 9: I added this with add_reference but without explicitly creating a foreign key because I couldn't get it to work correctly with this [truncation](https://github.com/greenriver/hmis-warehouse/blob/ad0735ad768cb9e9018f6088e23f2ca23036bba0/app/models/grda_warehouse/utility.rb#L4). I remember running into this before (with user staff assignment) but the resolution was a bit different because the relationship spanned databases, so just wanted to flag it and check that it looks okay.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,71 @@ module Mutations
   class UpdateExternalFormSubmission < CleanBaseMutation
     argument :input, Types::HmisSchema::ExternalFormSubmissionInput, required: true
     argument :id, ID, required: true
+    argument :project_id, ID, required: false # not required for backwards compatibility
 
-    field :external_form_submission, Types::HmisSchema::ExternalFormSubmission, null: false
+    field :external_form_submission, Types::HmisSchema::ExternalFormSubmission, null: true
```
martha, line 13: Changed to nullable because null may be returned if there are errors. I think this change could be problematic since it's not backwards compatible from the frontend's perspective, but maybe it's ok if the external form feature is not actively in use yet?

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,71 @@ module Mutations
...
+        # Only if there are Client and/or Enrollment fields in the form definition, initialize an enrollment
+        # (which will in turn initialize a Client, inside the form processor).
+        if definition.link_id_item_hash.values.find { |item| ['ENROLLMENT', 'CLIENT'].include?(item.mapping.record_type) }
+          project = Hmis::Hud::Project.find(project_id)
+          record.build_enrollment(project: project, data_source: project.data_source, entry_date: record.created_at)
+        end
+
+        form_processor = record.form_processor || record.build_form_processor(definition: definition)
+
+        form_processor.values = record.cleaned_values
```
martha: This part is a bit funny - at the least I'll add comments to try to explain what is funny about it, but perhaps we want to rework it a bit.

`record.cleaned_values` is the raw data from the form submission minus any irrelevant keys that aren't present on the form definition. It's created [here](https://github.com/greenriver/hmis-warehouse/pull/4657/files#diff-34c13c8335974559751928b36cbca0a3363274e0103ed5062da19bf038a918a7R44-R46) and saved to a new column on the external form submission record. e.g. `{"first_name": "martha"}`

`input.hud_values` is transformed like `{"Client.firstName": "martha"}` and I have it coming from the frontend [here](https://github.com/greenriver/hmis-frontend/blob/fccfab5cc2e5ac955ffee60beb31632e1724684d/src/modules/form/util/formUtil.ts#L1306-L1315), because I didn't want to duplicate that logic from ruby code, but I do think it's a little odd that the frontend has to pass this at the review step since there's nothing inherently about it that comes from the user interaction at this stage. Let's discuss more...

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,71 @@ module Mutations
...
+        errors.add_ar_errors(record.errors&.errors)
+        return { errors: errors }
       end
 
+      if record.status_changed? && record.status == 'reviewed' && !record.spam
+        definition = record.definition
+
+        # Only if there are Client and/or Enrollment fields in the form definition, initialize an enrollment
+        # (which will in turn initialize a Client, inside the form processor).
+        if definition.link_id_item_hash.values.find { |item| ['ENROLLMENT', 'CLIENT'].include?(item.mapping.record_type) }
```
martha: Are there any additional permissions that we want to check for in this case, like permission to create enrollments or clients? Or is `can_manage_external_form_submissions` enough?

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,71 @@ module Mutations
...
+
+        # Collect validations and warnings from AR Validator classes
+        record_validations = form_processor.collect_record_validations(user: current_user)
+        errors.push(*record_validations)
+
+        errors.deduplicate!
+        return { errors: errors } if errors.any?
+
+        if record.enrollment&.new_record?
+          record.enrollment.client.save!
```
martha: This throws an error if the form values have any client info but lack relationship to HoH. But the user can't do anything about that validation at this stage -- it would be a problem with the form logic.
- do we expect the PIT form to have a question for relationship to HoH?
- if not, does the html form support hidden fields? maybe we could have it as a hidden field inferred from other values, such as the # of household members, which I believe will only be collected for the first HH member?

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -20,10 +20,14 @@ class HmisSchema::ExternalFormSubmission < Types::BaseObject
...
     field :definition, Forms::FormDefinition, null: false
-    custom_data_elements_field
+    field :values, GraphQL::Types::JSON, null: true
 
     def definition
       load_ar_association(object, :definition)
     end
+
+    def values
+      object.cleaned_values
```
martha: This will be nil if the submission was created before the migration introduced in this PR. I debated returning `|| object.raw_data` here, but it would only case an error later on when the form processor encounters an unexpected link ID

martha: For backwards compat, we could also consider displaying the CDEs here if they already exist, but this seems like more complexity than it's worth to me.

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -34,12 +35,16 @@ def self.from_raw_data(raw_data, object_key:, last_modified:, form_definition:,
...
         raw_data: raw_data,
+        cleaned_values: clean(raw_data),
       }
       submission.save!
-      submission.process_custom_data_elements!(form_definition: form_definition)
       submission
     end
 
+    protected def clean(raw_data)
+      raw_data.select { |link_id, _| definition.link_id_item_hash.key?(link_id) }
```
martha: I think this still needs tweaking, since the values in `raw_data` are saved by custom field key, not by link ID.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
gigxz: Yeah, I think it would be good to move that transformation logic to the backend eventually, but probably as a separate chunk of work.

With the change from https://github.com/greenriver/hmis-warehouse/pull/4654, the submitted raw values should have the same format as "hud_values". There is no "values", though, which might cause a problem?

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
gigxz: Hmm. I think PIT should have a question for relationship, but not sure if it would be required. 

Another approach would be to default to `relationship_to_hoh: 99` in the EnrollmentProcessor, if the form values don't contain a relationship to hoh. We do something similar for VeteranStatus in the ClientProcessor, but there we still expect the key to be present. I think that could be a reasonable approach here

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -3865,13 +3865,13 @@ enum ExpelledReason {
 }
 
 type ExternalFormSubmission {
-  customDataElements: [CustomDataElement!]!
```
gigxz: We should still resolve these for backward compatibility

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -445,7 +445,7 @@ def self.open_hoh_enrollments_for_project(project, user:)
     def self.external_form_types_for_project(project)
       return [] unless project.present?
 
-      Hmis::Form::Instance.for_project(project).
+      Hmis::Form::Instance.for_project_through_entities(project).
```
gigxz, line 450: add a note about performance impact?

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -229,10 +229,13 @@ def outgoing_referral_postings(**args)
     end
 
     def external_form_submissions(**args)
-      instances = Hmis::Form::Instance.with_role(:EXTERNAL_FORM).active.where(entity: object)
+      # Not optimized performance, but reduces duplication of project_match logic
+      instances = Hmis::Form::Instance.with_role(:EXTERNAL_FORM).active
+      identifiers = instances.filter { |instance| instance.project_match(object) }.map(&:definition_identifier)
```
gigxz: We could use `for_project_through_entities` here right?
`instances.for_project_through_entities(self).select(:definition_identifier)`

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -399,10 +399,17 @@ def form_definition(id:)
 
     field :external_form_definition, Types::Forms::FormDefinition, null: true do
       argument :identifier, String, required: true
+      argument :id, ID, required: false # not required for backwards compat
```
gigxz: Feels like this would be simpler as a new query `external_form_definition_by_id`, and we can deprecate and eventually remove the old query. This is a little funky as-is because the `identifier` is "required" even though it's not necessary when looking up by ID. Although, I wonder whether we need a special query for looking up by id, vs resolving the definition on the submission https://github.com/greenriver/hmis-frontend/pull/909/files#r1742449259

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
gigxz: Hrm, yeah. The form processor will [raise an error](https://github.com/greenriver/hmis-warehouse/blob/6141755f2fd4f0b7f9cc084ca1ce6aa5c9350d6d/drivers/hmis/app/models/hmis/form/form_processor.rb#L78) if it encounters any keys that don't match `mapping`s in the Form Definition, and that's basically the logic we would need to replicate here.  I'm hesitant about re-implementing that logic. Let's discuss

gigxz: I also thought you and Theron had decided the cleaned_values wasn't strictly necessary, and it would be safe to use the raw values for processing and display. 
For processing, we would just need to drop known extraneous keys `captcha_score` ,`form_definition_id`,and `form_content_digest` from the raw_values prior to processing.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: decision: remove hud_values from schema, get it from the record's raw_values; remove cleaned_values from db (just use raw_values), don't call validation step that normally uses values, since the form is already submitted

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Discussed - tentatively, assume it's present?

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
martha: This is easier now that we are getting rid of `cleaned_values`, no need to migrate

martha: For displaying the form in dynamic view, always use the raw values, regardless of whether it's been processed. Add comment sharing example of changing client name in HMIS

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+
+        response, result = post_graphql(input) { mutation }
+        expect(response.status).to eq(200), result
+        expect(result.dig('data', 'updateExternalFormSubmission', 'externalFormSubmission', 'status')).to eq('reviewed')
+      end.to change(Hmis::Hud::CustomDataElement, :count).by(1).
+        and not_change(Hmis::Hud::Enrollment, :count).
+        and not_change(Hmis::Hud::Client, :count)
+    end
+
+    context 'when the submission has client record info' do
```
gigxz: Since the rawValues can basically contain anything, I think it's important to test unexpected input. Here are a few ideas:

- when the raw values have an attribute that is not present in the form definition (expect error)
- when the values have Client attributes with invalid values, for example `"Client.veteranStatus": "foo"` (expect error)
- when the form has fields for Client but not for Enrollment (should still create an enrollment and succeed, I think? may be the case for some PIT submissions, lets discuss)

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -8,6 +8,7 @@ module HmisExternalApis::ExternalForms
   class FormSubmission < ::HmisExternalApis::HmisExternalApisBase
     self.table_name = 'hmis_external_form_submissions'
     belongs_to :definition, class_name: 'Hmis::Form::Definition'
+    belongs_to :enrollment, class_name: 'Hmis::Hud::Enrollment', optional: true
```
gigxz, line 12:
```suggestion
    # Enrollment that was generated as a result of processing this form submission. Only applicable for certain external forms, like the PIT.
    belongs_to :enrollment, class_name: 'Hmis::Hud::Enrollment', optional: true
```

drivers/hmis_external_apis/spec/factories/external_forms/form_definitions.rb
```diff
@@ -27,5 +27,14 @@
...
     end
+    transient do
+      append_items { nil } # Items to append to FormDefinition content
+    end
+    after(:create) do |instance, evaluator|
+      if evaluator.append_items
+        instance.definition['item'][0]['item'].push(*Array.wrap(evaluator.append_items))
+        instance.save!
+      end
+    end
```
gigxz: This should already be inherited by the `hmis_form_definition` factory, which is listed as its `:parent`. Was that not working?

(Might also make sense to move this factory into `drivers/hmis/spec/factories/hmis/form/definitions.rb`, if that helps)

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+    create_access_control(hmis_user, p1, with_permission: [:can_manage_external_form_submissions, :can_view_project])
+  end
+
+  before(:each) do
+    hmis_login(user)
+  end
+
+  context 'when reviewing a submission' do
+    let!(:definition) { create(:hmis_external_form_definition) }
+    let!(:submission) { create(:hmis_external_form_submission, definition: definition, submitted_at: Date.current) }
```
gigxz: Wondering why pass `submitted_at: Date.current`? (The factory default is `Time.current`, and its a datetime field)

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
+      end
+    end
+
+    protected
+
+    EXTRANEOUS_KEYS = ['captcha_score', 'form_definition_id', 'form_content_digest'].freeze
+
+    def _resolve(id:, project_id:, input:)
       record = HmisExternalApis::ExternalForms::FormSubmission.find(id)
       raise 'Access denied' unless allowed?(permissions: [:can_manage_external_form_submissions])
```
gigxz: It looks like `can_manage_external_form_submissions` is currently applied globally, which is not ideal. That means same person who can process Prevention Screening submissions can also process PIT submissions, technically, even if they can't view that project. That is tangential to your PR, though, and should be another ticket. It might be tricky because FormSubmissions are not really tied to a single project.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
gigxz: Yes, we should ensure that the use has permission to view the project, and to edit enrollments in the project.
Basically this:
https://github.com/greenriver/hmis-warehouse/blob/271a749b704e4b200e7fbb1c35f662ed2fcff278/drivers/hmis/app/graphql/mutations/bulk_assign_service.rb#L14-L16

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
       end
 
+      if record.status_changed? && record.status == 'reviewed' && !record.spam
+        definition = record.definition
+
+        # Only if there are Client and/or Enrollment fields in the form definition, initialize an enrollment
+        # (which will in turn initialize a Client, inside the form processor).
+        if definition.link_id_item_hash.values.find { |item| ['ENROLLMENT', 'CLIENT'].include?(item.mapping.record_type) }
+          project = Hmis::Hud::Project.find(project_id)
+          record.build_enrollment(project: project, data_source: project.data_source, entry_date: record.created_at)
```
gigxz: This could technically create an Enrollment in a non-HMIS data source. It would be fixed by my other comment which introduces `viewable_by` scope, which filters down the projects based on the current_users's HMIS data source ID.

But, just to explain, here's what could happen:
* Say there is a project with id `1`, but it exists in a separate data source  (not the HMIS)
* Bad actor hits this mutation with `project_id: 1`, by tampering with the js 
* `record.build_enrollment(project: project, data_source: project.data_source, ...)` will create an Enrollment in the non-HMIS data source

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
+          # Assume that required values on Client and Enrollment (e.g. relationship to HoH) are present
+          # and correctly mapped in raw_values. The form processor record validator will fail otherwise.
```
gigxz: See comment https://github.com/greenriver/hmis-warehouse/pull/4667#discussion_r1744270493

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
+        end
+
+        form_processor = record.form_processor || record.build_form_processor(definition: definition)
```
gigxz: In what situation would the record already have a form processor? Are we supporting re-processing submissions? (That might be a good idea, actually, just didn't think it was part of the ux)

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
+
+        # Run to create CDEs, and client/enrollment if applicable
+        form_processor.run!(user: current_user)
+
+        # Collect validations and warnings from AR Validator classes
+        record_validations = form_processor.collect_record_validations(user: current_user)
+        errors.push(*record_validations)
+
+        errors.deduplicate!
+        return { errors: errors } if errors.any?
```
gigxz: I'm not sure if we want this. The user can't do anything to fix the error, it basically means the form is in purgatory until we fix whatever is wrong with it from the backend, or it's deleted.

It looks like `collect_record_validations ` will also run the custom validators (like this [hmis_validate](https://github.com/greenriver/hmis-warehouse/blob/271a749b704e4b200e7fbb1c35f662ed2fcff278/drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb#L92-L111)), which are validations that are not strictly needed by the database. I think that's too strict for this scenario.. for example if the submission is resulting in 2 HoHs in one hh, we would completely block them from processing it. I _think_ we want to be more lenient, and allow the submissions to be processed and cleaned up after the fact.

Of course there is another class of validations that will cause `.save!` below to raise an exception. I think it's best to just let those exceptions raise, so that we're notified via Sentry. If we want to error out with a more user-friendly/specific message, we can:
```
  error_out(client.errors.full_messages) unless client.valid?
  error_out(enrollment.errors.full_messages) unless enrollment.valid?
  # proceed to save...
   
    def error_out(msg)
      # error out with user-facing error message
      raise HmisErrors::ApiError.new(msg, display_message: 'Unable to process form submission due to invalid values. Contact your administrator if the problem persists.')
    end
```

That way we can track things in sentry (and mute them if needed), and the user gets something beyond just "an error occurred".

Thoughts? happy to huddle on it!

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -20,10 +20,19 @@ class HmisSchema::ExternalFormSubmission < Types::BaseObject
...
     def definition
       load_ar_association(object, :definition)
     end
+
+    def values
+      # Unlike CustomAssessments, which display on the frontend using the created CDEs once they've been submitted,
+      # ExternalFormSubmissions always display using the values from raw_data. Example: If the form processor created
+      # a client, and then a user manually updates the client's name in HMIS, the form review UI should still display
+      # the client name that was originally submitted with the form.
+      object.raw_data
```
gigxz: We probably also want to exclude the "extraneous keys" here. Maybe it makes sense to have a function on the `HmisExternalApis::ExternalForms::FormSubmission`, so that list of keys also lives closer to home:

```
# keys that are added to raw_values upon submission, but are not part of the form values
EXTRANEOUS_KEYS = ['captcha_score', 'form_definition_id', 'form_content_digest'].freeze

def form_values
   raw_data.reject { |key, _| EXTRANEOUS_KEYS.include?(key) }
end
```

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,24 +8,72 @@ module Mutations
...
-        errors = record.errors
-        record = nil
+      errors = HmisErrors::Errors.new
+
+      unless record.valid?
+        errors.add_ar_errors(record.errors&.errors)
+        return { errors: errors }
       end
 
+      if record.status_changed? && record.status == 'reviewed' && !record.spam
```
gigxz, line 29: We will need to refactor some of this out into shared code when we add a mutation for "bulk review". Thinking ahead to that, would it be possible to refactor out most of this processing code to live on the FormSubmission record?

```ruby
if record.status_changed? && record.status == 'reviewed' && !record.spam
   record.run_form_processor(project: project) #<== processes and build related records, but doesnt save

   if record.enrollment&.new_record?
          record.enrollment.client.save!
          record.enrollment.save_new_enrollment!
    end
   record.form_processor.save! # add has_one to FormSubmission
end
```

drivers/hmis_external_apis/spec/factories/external_forms/form_definitions.rb
martha: Ah, yeah, thanks for the catch. I think i filched that while this was still in PR https://github.com/greenriver/hmis-warehouse/pull/4649/files but yes, it should be removed.

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
martha: These can be removed

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Created https://github.com/open-path/Green-River/issues/6628

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Ok great I will add the permission check. Also, should we check this permission earlier (maybe from the frontend)? I'll add some thoughts to https://github.com/open-path/Green-River/issues/6628

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Outdated, see https://github.com/greenriver/hmis-warehouse/pull/4667#discussion_r1744270493

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: 😥 Thanks! With `viewable_by` scope added, there is no additional change needed, right?

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Good point, I was copying from the submit assessment mutation.

Right now though, the UX doesn't prevent this:
- Review a form submission
- Open the reviewed submission again, uncheck "Mark reviewed" and hit "Save"
- Open the un-reviewed submission again, check "Mark reviewed" and hit "Save"

I'm seeing this leads to a deadlock error so I can investigate more
Screenshot 2024-09-04 at 4 35 02 PM

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: This sounds good to me, so I've updated to use your suggestions. Maybe it would be good to pair on testing this? I think the test case I've come up with will break once we add logic for automatically setting the relationship to HoH, per your comment [here](https://github.com/greenriver/hmis-warehouse/pull/4667#pullrequestreview-2281004233).

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: I like the idea of the refactor, and took a stab at this here https://github.com/greenriver/hmis-warehouse/pull/4657/commits/a9621b661414974afa3002db1fddce6f88c4bf7f

Let me know what you think about the `definition.creates_client_or_enrollment` helper. I'm not sure if making the check twice will be too slow, especially when we do this in bulk - perhaps it should be passed in as an arg?

(Relates to https://github.com/open-path/Green-River/issues/6628)

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: This should be covered by the test case here: https://github.com/greenriver/hmis-warehouse/pull/4657/files#diff-360e572a5a39d6a50ef9f0365163ca539293ed1cff0617f610d7d316b4d8043cR252-R256

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Hm, the deadlock is not specific to the re-review case, it turns out. I suspect maybe introduced here: https://github.com/greenriver/hmis-warehouse/pull/4657#discussion_r1744714593, I'll keep investigating. Edit: I think this is solved with moving the permission check outside of the transaction https://github.com/greenriver/hmis-warehouse/commit/a0d1f9feacad670c4a8d0249c64dae099a49e978)

anyway, I also added test coverage for the re-reviewing situation. https://github.com/greenriver/hmis-warehouse/pull/4657/commits/429c464f06bd9542eee80eefa08d20504264cbac

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -50,37 +53,65 @@ def self.from_raw_data(raw_data, object_key:, last_modified:, form_definition:,
...
-          DateCreated: now,
-          DateUpdated: now,
-        }.merge(HmisExternalApis::ExternalForms::FormSubmission.cde_value_fields(cded, value))
+    def run_form_processor(project, current_user)
+      values_to_process = form_values.clone
+
+      # Only if there are Client and/or Enrollment fields in the form definition, initialize an enrollment
+      # (which will in turn initialize a Client, inside the form processor).
+      if definition.creates_client_or_enrollment?
+        household_id = form_values['Enrollment.householdId']
```
gigxz, line 80: This logic looks good to me. The nested if-else is a little hard to follow, though, and we should also add some validation for the incoming values. Suggestion (untested!):

```ruby
household_id = form_values['Enrollment.householdId']
# drop household id if it doesn't match expected format
household_id = nil unless household_id.starts_with?('HH') && household_id.size > 20
# drop household id if it has a conflict in this data source in another project (just to be safe)
household_id = nil if Hmis::Hud::Enrollment.where(household_id: household_id, data_source: project.data_source).where.not(project: project).exists?
# if no household id, generate a new one
household_id ||= Hmis::Hud::Enrollment.generate_household_id

relationship_to_hoh = form_values['Enrollment.relationshipToHoH']
# todo: drop if form value for relationshipToHoH was an invalid value
# set default relationship_to_hoh (1 for new household, 99 for existing household)
relationship_to_hoh ||= project.enrollments.where(household_id: household_id).exists? ? 99 : 1

File: drivers/hmis/app/graphql/mutations/update_external_form_submission.rb (hunk 19 above)
Comment (gigxz) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb]: Ahmm, okay. I don't think we should allow "un-review".. we don't allow anything like that for form submission (going from submitted back to WIP), and I think the desired behavior is unclear (eg should the client be deleted?). I didn't realize the UI permitted that currently. Do you have a use-case in mind for keeping it?

If you're good with removing it, we can do the work to disable the checkbox in a separate issue, but here it probably makes sense to add a check in `UpdateExternalFormSubmission` to ensure that a reviewed submission can't get marked as unreviewed.

File: drivers/hmis/app/graphql/mutations/update_external_form_submission.rb (hunk 22 above)
Comment (gigxz) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb:29]: Yep I think it looks good. I'm on the fence on whether it should be `creates_client_or_enrollment` or `updates_client_or_enrollment`... the presence of the "mapping" doesn't tell us whether the form is used for creation or updating (And sometimes the same form is used for both). The link_id_item_hash is on an instance variable so we should only need to traverse the definition once

File: drivers/hmis/app/graphql/mutations/update_external_form_submission.rb (hunk 19 above)
Comment (gigxz) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb]: This would be a quick solution to add on FormSubmission, which the mutation will handle returning:
```
validate :validate_status_change
private def validate_status_change
  return unless status_changed? && status.present? && status_was.present?

  errors.add(:status, :invalid, full_message: 'Cannot change status from Reviewed to New') if status_was == 'reviewed' && status == 'new'
end
```

File: drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb (hunk 23 above)
Comment (martha) [reply drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb:80]: Thanks, I like the restructure. I expanded on it a little bit to ensure that invalid household ID also nulls out the relationship. Also, in both cases, the values need to be dropped from the hash passed to the form processor, otherwise the form processor will override what we pass when creating the enrollment.

File: drivers/hmis/app/graphql/mutations/update_external_form_submission.rb (hunk 19 above)
Comment (martha) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb]: Ok perfect. I agree, there is no use case for it. It was just leftover from when 'reviewing' a form submission didn't really do anything aside from changing the attribute on the record.

Comment (martha) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb]: I added a bullet to disable the checkbox to this frontend ticket: https://github.com/open-path/Green-River/issues/6598

**Comment by martha** [comment]
Still outstanding:

From ticket:
> trigger warehouse processing to generate Warehouse Client for each

I also noticed that it's quite slow when I run it locally. I think the `warehouse_identify_duplicate_clients` is causing the slowness. I think I remember something about delayed jobs running concurrently in dev, but I can't find it now.

File: drivers/hmis/app/graphql/mutations/update_external_form_submission.rb (hunk 17 above)
Comment (gigxz) [reply drivers/hmis/app/graphql/mutations/update_external_form_submission.rb]: nope looks good now!

**Comment by gigxz** [comment]
@martha It should already be queuing the warehouse processing job when the client is saved, because of this hook called on `after_create`: https://github.com/greenriver/hmis-warehouse/blob/0d8b24fda445fb964111865ca1f32bbc3f606545/drivers/hmis/app/models/hmis/hud/client.rb#L382-L386

File: db/reporting_structure.sql [hunk 24]
```diff
@@ -22,50 +22,50 @@ SET row_security = off;
```
gigxz: @martha I think these to sql file changes should be backed out, and the warehouse_structure.sql file should be added, since that's the db that the migration affects

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -8,28 +8,60 @@ module Mutations
...
     argument :input, Types::HmisSchema::ExternalFormSubmissionInput, required: true
     argument :id, ID, required: true
+    argument :project_id, ID, required: false # not required for backwards compatibility
 
-    field :external_form_submission, Types::HmisSchema::ExternalFormSubmission, null: false
+    field :external_form_submission, Types::HmisSchema::ExternalFormSubmission, null: true
     field :errors, [Types::HmisSchema::ValidationError], null: false, resolver: Resolvers::ValidationErrors
 
-    def resolve(id:, input:)
+    def resolve(id:, project_id:, input:)
```
gigxz: since it's optional, this needs to be:
```suggestion
    def resolve(id:, project_id: nil, input:)
```

gigxz: Although it looks the mutation will error out if project is missing, so it's not _actually_ backwards compatible. It looks like it wouldn't be too difficult to make it truly backward compatible.. I think the project could be an optional arg on the processing function, like this(?):

```
# Run a FormProcessor against the values, to build related records (such as CustomDataElements, Enrollment, etc). It does not save the related records.
# Note: project is only needed if the form creates Enrollments. If the form only generates CDEs, it doesn't need a project. (it can retrieve the data source ID from current_user)
def run_form_processor(current_user, project: nil)
```

db/reporting_structure.sql
martha: Ok, thanks! To be honest, I often don't feel sure which .sql file diffs to include in my PRs..
Here I've backed out the changes from `db/structure.sql` and `db/reporting_structure.sql`, and kept the full changes that my migration generates against `db/warehouse_structure.sql`, although not all of those changes are related to this migration.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -283,7 +283,7 @@ def normalize_single_custom_field_value(cded, value)
       case value
       when nil, ''
         nil
-      when true, false
+      when true, false, 0, 1
         value
```
gigxz: Since this is a normalization function, probably best to return `true` and `false` as the boolean values, so that caller knows exactly what to expect. (Even though rails will save 0/1 correctly into boolean fields). Maybe:
```
when 0, "0", false
   false
when 1, "1", true
   true
```

db/reporting_structure.sql
gigxz: The rule of thumb is to only commit the sql file changes if your code contains a migration that affects it. So if you added a migration to `db/warehouse/migrate`, then you should commit `db/warehouse_structure.sql`, etc. It's expected that there are other changes in the sql file in addition to the change from your migration, since we all have slightly different setups.

drivers/hmis/spec/requests/hmis/external_form_submissions_spec.rb
```diff
@@ -57,20 +56,66 @@
...
+      }
+      expect(result.dig('data', 'project', 'externalFormSubmissions', 'nodes')).to contain_exactly(expected)
+    end
+  end
+
+  context 'when there are several external forms' do
+    let!(:definition_2) do
+      fd = create(:hmis_external_form_definition)
+      HmisExternalApis::PublishExternalFormsJob.new.perform(fd.id)
+      fd.reload
```
gigxz: Does the form need to be run through this job before testing? I think the factory should generate the CDEDs for it, so this might not be necessary anymore. (I see it was present in the older test case above).

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -0,0 +1,300 @@
...
+
+      let!(:submission) do
+        data = {
+          'Client.firstName': 'Oranges',
+          'captcha_score': '1.', # also test that extraneous fields get filtered out
+          'form_definition_id': definition.id,
+          'form_content_digest': 'something random',
+        }.stringify_keys
+        create(:hmis_external_form_submission, raw_data: data, definition: definition)
+      end
```
gigxz: I think this submission should be inside the `when the submission specifies client info only` context?

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -0,0 +1,300 @@
...
+
+      it 'should raise an error' do
+        expect do
+          expect_gql_error(post_graphql(input) { mutation })
+        end.to not_change(Hmis::Hud::CustomDataElement, :count)
+      end
+    end
+
+    context 'when the form definition accepts client/enrollment information' do
```
gigxz, line 95: nice test coverage, thanks!

gigxz (approved):

PR #4658, opened 2024-08-30
EKS Prep

yakloinsteak:
## Description

Changes needed to support Delayed Jobs on EKS.
- DJ metrics recorded and exposed
- Script to check if database is available and migrations have run.
- cronjob in EKS support

https://github.com/greenriver/argocd-manifests/pull/1032

## Type of change
- [x] Dependency update (new gems)


app/jobs/base_job.rb
```diff
@@ -8,36 +8,45 @@ class BaseJob < ApplicationJob
...
 
-      msg = "Started dir is `#{STARTING_PATH}`\nCurrent dir is `#{expected_path}`\nExiting in order to let systemd restart me in the correct directory. Delayed::Job"
-      notify_on_restart(msg)
+    def after(job)
+      after_handler(job)
+    end
 
-      unlock_job!(job.id)
+    def before_handler(job)
+      self.start_time = Time.now
```
eanders: It's probably fine since you're using it all in the same file and just getting the time diff, but we tend to prefer `Time.current` over `Time.now` to incorporate the timezone.

app/jobs/base_job.rb
```diff
@@ -52,39 +61,4 @@ def before(job)
...
-
-  def unlock_job!(job_id)
-    a_t = Delayed::Job.arel_table
-    job_object = Delayed::Job.where(a_t[:handler].matches("%job_id: #{job_id}%").or(a_t[:id].eq(job_id))).first
-    return unless job_object
-
-    msg = "Restarting stale delayed job: #{job_object.locked_by}"
-    notify_on_restart(msg)
-
-    job_object.update(locked_by: nil, locked_at: nil)
```
eanders, line 88: Is this all code related to the bare-metal deployments?  It looks like we moved this logic to the `WorkerStatus` class for use with ECS, just want to confirm.

config/deploy/docker/lib/cronjob.rb
```diff
@@ -0,0 +1,152 @@
...
+  attr_accessor :capacity_type
+  attr_accessor :description
+  attr_accessor :jobname
+  attr_accessor :schedule_expression
+  attr_accessor :k8s_image
+  attr_accessor :k8s_service_account_name
+
+  MAX_NAME_LENGTH = 52
+
+  def initialize(description:, schedule_expression:, command:, capacity_type: 'spot')
```
eanders, line 14: I haven't found it yet, so just noting, `capacity_provider` should probably not be spot if the task is not interruptible.

eanders (approved):
@yakloinsteak I think this looks great.  Thank you.  Can we target `release-134` so it doesn't go out to production immediately?

app/jobs/base_job.rb
yakloinsteak: I believe this was all related to bare-metal deployments. Are you agreeing that it's dead code? That's what I was thinking it was at least.

eanders: Yes, I was agreeing with that assessment, just wanted to make sure that's why you were removing it.

config/deploy/docker/lib/cronjob.rb
yakloinsteak: I'm just making the default the cheaper option. The logic in `config/schedule.rb` is explicit about all of the entries I think. interruptible jobs are put on spot capacity. non-interruptible on on-demand.

yakloinsteak:
May I merge this today, @eanders ? You might want to peek at https://github.com/greenriver/hmis-warehouse/blob/tb/dj-metrics/lib/rds_sql_server/sql_server_base.rb where I've memoized some things to reduce AWS API hits for all of the subclasses of that class. I believe it will help with boot time. On that topic, I was wondering if we should combine some of the rake tasks in the deploy_task.sh script. I think it would speed things up to _not_ load the rails stack over and over.

eanders (approved):
Looks good to me, let's give it a try.

PR #4661, opened 2024-08-31
6615-postgresql-config-improvement

ttoomey:

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6615)

Testing the waters on configuration improvement. We should monitor and iterate.
* There should be no user-facing impact of this change. 
* the query planner is more likely to use our indexes, hopefully resulting in improved query time
* I don't have access to RDS metrics to see the impact of changing working mem. Guessing that the default 4mb is probably too low
* could use more memory per connect in exchange for faster sorting operations and less disk io

## Type of change
- [x] performance optimization

- [x If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

### DIFF
--- a/config/database.yml
+++ b/config/database.yml
@@ -20,6 +20,12 @@ default_app_db: &default_app_db
   <% if ENV['DATABASE_HOST'] %>
   host: <%= ENV['DATABASE_HOST'] %>
   <% end %>
+  variables:
+    # our dbs our on SSDs so random access is almost as cheap as seq_page_cost
+    random_page_cost: 1.1
+    # can improve performance for queries that involve large sorts or hash operations but increasing this value too
+    # much could lead to excessive memory usage
+    work_mem: 8MB
 
 default_warehouse_db: &default_warehouse_db
   adapter: <%= ENV['WAREHOUSE_DATABASE_ADAPTER'] %>
--- a/config/database.yml
+++ b/config/database.yml
@@ -21,7 +21,7 @@ default_app_db: &default_app_db
   host: <%= ENV['DATABASE_HOST'] %>
   <% end %>
   variables:
-    # our dbs our on SSDs so random access is almost as cheap as seq_page_cost
+    # our dbs use SSDs so random access is almost as cheap as seq_page_cost
     random_page_cost: 1.1
     # can improve performance for queries that involve large sorts or hash operations but increasing this value too
     # much could lead to excessive memory usage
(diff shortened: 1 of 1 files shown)


config/database.yml
```diff
@@ -4,7 +4,16 @@
 # https://github.com/rails/rails/issues/38924#issuecomment-612881746
 # i have to unroll some parts so that it's simple enough for rails to figure out that I want to use multi-db
 
+global_pg_db_config: &global_pg_db_config
+  variables:
+    # our dbs use SSDs so random access is almost as cheap as seq_page_cost
+    random_page_cost: 1.1
+    # can improve performance for queries that involve large sorts or hash operations but increasing this value too
+    # much could lead to excessive memory usage
+    work_mem: 8MB
```
eanders, line 21: Let's check with @yakloinsteak on this one.  I think it has the potential to massively increase memory usage depending on number of connections allowed.

ttoomey: Agreed, I also don't have a clear indication that this is a problem; the relevant monitoring (Temp bytes / files) isn't available to me. But if I had to guess, I'd say the default of 4MB is likely to be too low

yakloinsteak: The first value sounds like a really good idea to test. Might be worth having a little more explanation and possibly set seq_page_cost as well (even if you're just setting it to the default). Mostly to help out those that aren't familiar with the settings. This could be a huge win. I always presumed that RDS set sane values assuming SSDs, but @ttoomey is spot on. Feels like something we would want across the board in all our postgres databases. Thanks for bringing this up, Theron!

Hard to say about the safety of the second one, but I'm definitely on board with increasing it. It's at the default of 4kB right now, so this is a pretty big jump. I'd probably increase it less on the first round. It's kind of a function of the number of connections and the amount of RAM available and what kinds of queries are being run. Maybe start with 8kB to double it? Or maybe 250kB.

yakloinsteak (changes requested):
I approve after lowering the `work_mem` value.

config/database.yml
yakloinsteak: Oh, I misread the numbers in the database. It's 4096kB, not 4kB. 8MB might actually be okay.

yakloinsteak (approved):

config/database.yml
ttoomey: Yeah I also assumed RDS did something sensible with PG configuration but we seem to have the defaults. It makes me think there's probably other opportunities for tuning but we can start with these vars and see what happens.

We're only doubling the default size from 4 to 8MB but still I'd suggest we try this on QA first which should have an isolated RDS. It should give us some idea of the impact on memory use from this config change.

ttoomey:
@eanders - what do you think about merging this so we can get it up for QA next week?

PR #4662, opened 2024-09-02
6429 upgrade to postgres 16

ttoomey:

## Description

[GH Issue](https://github.com/open-path/Green-River/issues/6429)

Upgrade is something like this:

Make sure the new db image is built
```bash
host $ docker compose build db
```

Migrate your local application database (or ignore this and just start a new db from scratch)

If you aren't using a custom data volume, (DB_DATA_VOLUME), you can just run this
```bash
host $ bash docker/pg16/upgrade.sh
```

If you're using a custom data volume, make a dir for pg 16 and set appropriate env vars  
```bash
host $ mkdir -p /Users/jdoe/pg-16-data
host $ export DB_DATA_VOLUME_PREVIOUS=/Users/jdoe/pg-13-data
host $ export DB_DATA_VOLUME=/Users/jdoe/pg-16-data
host $ DB_DATA_VOLUME_PREVIOUS=/Users/jdoe/pg-13-data DB_DATA_VOLUME=/Users/jdoe/pg-16-data  bash docker/pg16/upgrade.sh
```

Remove old dbdata_pg13 and dbdate docker volumes once you're sure you won't need them again

## Type of change
- [x] Dependency update


docker/docker-compose.yml
```diff
@@ -169,37 +169,34 @@ services:
       timeout: 3s
       retries: 3
     volumes:
-      - dbdata:/var/lib/postgresql/data
+      - dbdata_pg13:/var/lib/postgresql/data
```
eanders: I wonder if we should update this to match the previous volume setup:
`- ${DB_DATA_VOLUME:-dbdata_pg16}:/var/lib/postgresql/data`
It would need an ENV, and I think we'd want a different name for the ENV.  Maybe `DB_DATA_VOLUME_PREVIOUS`.  That way we won't accidentally use the same volume, maybe?

eanders: Or, maybe my memory is shaky on this.  Do we _intend_ to use the same volume?  If so, then it should mirror what's in `db`.

ttoomey: The intention is use a new volume for pg16 and leave the current data volume untouched.

Good point about the DB_DATA_VOLUME. I wonder if we can just remove that var and suggest people use a `docker-compose.override.yaml` file for customizations rather than an env variable.

eanders: Hmm, maybe, though I think there was some issue with the way volumes/arrays are overlayed that made that difficult.

ttoomey: Yes I think you would need to re-declare the whole volumes array in the override, which isn't ideal but should work. Is it your impression folks are actually using the `$DB_DATA_VOLUME`? Perhaps the override is an okay work-around if this var isn't commonly used

eanders: I'm using it, I think others are too.  I use it so my db is stored on my local machine instead of a docker volume.

ttoomey: got it, I'll use the env vars as you suggest and add that to the migration instructions

docker/docker-compose.yml
```diff
@@ -169,17 +169,16 @@ services:
...
+      - ${DB_DATA_VOLUME_PREVIOUS:-dbdata_pg13}:/var/lib/postgresql/data
     ports:
       - 5432
     expose:
       - 5432
 
-  # This should no longer need to be overridden for ARM
-  # processors
+  # This should no longer need to be overridden for ARM processors
   db:
```
yakloinsteak, line 208: I think an `image` value will force a build and allow switching branches. Maybe "image: warehousepg:16.0".  My initial boot on this branch created and initialized the pg16 volume with pg13.

docker/pg16/upgrade.sh
```diff
@@ -0,0 +1,33 @@
+#!/bin/env bash
+
+set -ex
+set -oi pipefail
+
+# If you need to start over for some reason, destroy the db container and run:
+# docker volume rm hmis-warehouse_dbdata_pg16
+
+docker-compose stop db_previous db
```
yakloinsteak: Shouldn't everybody be using the plugin and running `docker compose` without the dash?

docker/docker-compose.yml
```diff
@@ -150,10 +150,10 @@ services:
       yarn:
         condition: service_started
 
-  # To transition from a previous major version to pg13, run ./docker/pg13/upgrade.db
+  # To transition from a previous major version to pg16, run ./docker/pg16/upgrade.sh
   db_previous:
     build:
-      context: docker/pg12
+      context: docker/pg13
```
yakloinsteak, line 184: Maybe we don't want to change anything about the previous service (image or volume name) in this PR? I had to fiddle with volumes and the image to be able to boot this.

docker/docker-compose.yml
yakloinsteak, line 208: When I've done this sort of thing before, I've often named both the services and volumes with the postgres version to make sure they don't get mismatched. I don't think one has to do that. Just wanted to mention it.

yakloinsteak (commented):
I was able to copy the database successfully, but I had to do a few things manually.

docker/docker-compose.yml
ttoomey: I think changing the service name would require updating the host in our rails config (and maybe local superset config too)

docker/docker-compose.yml
ttoomey: Maybe defining an image on db_previous will fix this. Otherwise I guess we'd leave db_previous as pg12 and add a new db_13 service

docker/pg16/upgrade.sh
eanders: For those of us on colima, there's an extra step to get `docker compose` to work without the dash.  That's [documented here](https://github.com/greenriver/hmis-warehouse/pull/4670), so I think we're good dropping it.

docker/docker-compose.yml
yakloinsteak: I think with the  `container_name` set you might not need to update anything. If there are references to just the service name and that's working, that sounds fragile and I'd recommend using the container_name value instead.

ttoomey: Sorry if I'm not following but I imagine folks use hostname in the config (`DATABASE_HOST=db`. I guess we could use the hostname property and rename the service

Would you propose we use something like this?
```yaml
services:
  pg16:
    container_name: hmis-warehouse-db
    image: warehousepg:16
    hostname: db
    volumes:
      - ${DB_DATA_VOLUME:-dbdata_pg16}:/var/lib/postgresql/data
    # ...
```

Or if we just use the image property, does that address the concern around mismatched containers?

yakloinsteak: I would make sure that the hostname that people are using is unique to this app. Anybody that's using `db` or `pg16` or anything like that will have problems if they run two apps at the same time on the same docker network. My personal preference is to leave out that hostname, keep the service name pg16 and be done with it. I don't know if it matters that much. Maybe I'm the only person running more than one app at the same time.

ttoomey:
papertrail: I'll rerun this and merge to 136

PR #4666, opened 2024-09-03
FY2024 LSA

eanders:

## Description

This is some scaffold for the FY2024 LSA
1. Notes that locally this may need to be run with `aws-vault` in the README
2. FY2024 sample code
3. Scaffold for running FY2024 LSA

## Type of change


dtgreiner (approved):
Looks good to me!

ttoomey:
@eanders - this is mostly duplicate from last year right? Are there any new pieces I should review more closely?

eanders:
Good question @ttoomey !  Yes, this is 85% a copy from last year, 5% new sample code, 10% adjustments to:
1. How we detect if a report has failed
2. How we expose/identify missing PDDE data
3. How we interact with RDS

If you can focus on the following files, that would be really helpful:
```
app/assets/javascripts/filter_projects.js
app/models/hud_reports/report_instance.rb
lib/rds_sql_server/rds.rb
```

lib/rds_sql_server/rds.rb
```diff
@@ -34,11 +34,17 @@ def self.timeout
     @timeout || 50_000_000
   end
 
+  def self.rds_available?
+    new.client.present?
+  end
+
   def initialize
     self.client = Aws::RDS::Client.new
+  rescue RuntimeError
```
ttoomey, line 43: We want RuntimeError specifically?

eanders:
Some screenshots of the missing data changes are here:
https://github.com/open-path/Green-River/issues/956#issuecomment-2338848961

lib/rds_sql_server/rds.rb
```diff
@@ -34,11 +34,17 @@ def self.timeout
...
+    self.client = nil
   end
 
-  define_method(:sqlservers) { _list.select { |server| server.engine.match(/sqlserver/) } }
+  define_method(:sqlservers) { _list&.select { |server| server.engine.match(/sqlserver/) } }
```
ttoomey, line 47: I wonder why this uses define_method in this class. The methods don't appear to be dynamic
```suggestion
  def sqlservers
    _list&.select { |server| server.engine.match(/sqlserver/) }
  end
```

app/assets/javascripts/filter_projects.js
```diff
@@ -74,21 +75,86 @@ App.StimulusApp.register('filter-projects', class extends Stimulus.Controller {
...
+      return
+    }
+    $(this.missingItemsTarget).attr('href', this.buildMissingDataLink(html));
+  }
+
+  checkMissingData(html) {
+    if (! this.hasMissingItemsTarget) {
+      return false
+    }
+    return $.ajax({ async: false, type: 'GET', url: this.buildMissingDataLink(html, true) }).done().responseJSON
```
ttoomey, line 122: This should work. Can we avoid using async:false since it locks the page? I guess then we'd need a loading indicator. Also maybe we could use some error handling. Looking at the whole thing, perhaps it could use a bit of general TLC; could be an issue for the technical backlog

app/models/hud_reports/report_instance.rb
```diff
@@ -79,6 +90,19 @@ def failures
       report_cells.where.not(error_messages: nil).pluck(:question, :cell_name, :status, :error_messages)
     end
 
+    private def job_failed?
+      related_job.present? && related_job.failed?
+    end
+
+    def related_job
+      # See if we can find a related job (this is really overloading the jobs_for_class scope, but should work)
+      dj = Delayed::Job.jobs_for_class('RunReportJob').jobs_for_class(id.to_s)
```
ttoomey, line 99: Does this condition work? If we're tyring to find either class, it seems like it should be
```suggestion
      dj = Delayed::Job.jobs_for_class(['RunReportJob', id.to_s])
```

app/models/hud_reports/report_instance.rb
```diff
@@ -79,6 +90,19 @@ def failures
...
+      return dj&.first if dj.count == 1
+
+      # If we didn't find an obvious match, just return nothing
+      nil
```
ttoomey: maybe golf it down:
```suggestion

      # If we didn't find an obvious match, just return nothing
      dj.order(:id).first unless dj.many?
```

app/models/filters/hud_filter_base.rb
```diff
@@ -42,7 +42,7 @@ def effective_project_ids
           pluck(p_t[:id])
       end
 
-      if coc_codes.present?
+      if coc_codes&.reject(&:blank?).present?
```
ttoomey: maybe
```suggestion
      if coc_codes&.any?(&:present?)
```

drivers/hud_lsa/app/models/hud_lsa/filters/lsa_filter.rb
```diff
@@ -17,14 +17,16 @@ def coc_codes
...
+      filters[:coc_code] = filters[:coc_code].first if filters[:coc_code].is_a?(Array)
       super
 
       filters = filters.to_h.with_indifferent_access
       pit_date = filters.dig(:on)&.to_date
       return unless pit_date.present?
 
       self.start = pit_date
-      self.end = pit_date + 1.day
+      self.end = pit_date
```
ttoomey, line 31: was this a bug fix or spec change?

drivers/hud_lsa/app/models/hud_lsa/filters/lsa_filter.rb
```diff
@@ -17,14 +17,16 @@ def coc_codes
 
     # If a single day for the PIT has been selected, set the start and end dates from the 'on' date
     def update(filters)
+      # CoC Codes get treated in some odd ways, try to standardize, LSA & HIC can only run on one CoC at a time
+      filters[:coc_code] = filters[:coc_code].first if filters[:coc_code].is_a?(Array)
```
ttoomey: should this raise if there are many coc codes?

drivers/hud_lsa/app/models/hud_lsa/generators/fy2024/rds_concern.rb
```diff
@@ -0,0 +1,73 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HudLsa::Generators::Fy2024::RdsConcern
```
ttoomey, line 7: this is fine, although it seems like RDS shouldn't change much between years and could be shared.

ttoomey (approved):
looks good, I focused on the few files you flagged for attention

lib/rds_sql_server/rds.rb
eanders: @yakloinsteak, this is something you wrote (years ago).  Any chance you remember why you went with `define_method` instead of `def`?  I feel like maybe it had to do with the dynamic loading and the lack of always having an RDS connection to rely on, so maybe it got unhappy at compilation time if they were static method definitions?

app/assets/javascripts/filter_projects.js
eanders: This could definitely use a visit from someone with a solid modern JS background.

app/models/hud_reports/report_instance.rb
eanders: That's an intentional AND.  I'm mis-using the `jobs_for_class` method which really just does a `like` on the json `handler` column.  Basically, we want to ensure we get a row that is working on the right class AND has the right ID.

app/models/hud_reports/report_instance.rb
eanders: Not sure the order is necessary since we explicitly only return it if there is one, but I like the simplification, thanks!

drivers/hud_lsa/app/models/hud_lsa/filters/lsa_filter.rb
eanders: Yep, that's the **big** HIC spec change this year.

lib/rds_sql_server/rds.rb
yakloinsteak: `_list` might have been a variable from above that method in some previous iteration, but I've also been known to use `define_method` for one-line methods to save space. I don't think we need it, and what @ttoomey sounds fine to me.

eanders: Thanks!

lib/rds_sql_server/rds.rb
yakloinsteak: That is what's thrown when there are AWS creds missing. Maybe SqlServerBase could become smarter about when to instantiate this class. Or the handling of errors maybe belongs there.

app/models/hud_reports/report_instance.rb
ttoomey: oh yeah, I have the habit of always using `first` and `order` together since it ensures behavior is deterministic.

In this case there are two queries executed on this line (many and first). So you could have a super tiny edge case with `many` returning false, a new job is added, and then `first` returns one of those two records. So always returning the same record is helpful for consistency

PR #4667, opened 2024-09-04
Support linked households for external form submissions

gigxz:

## Description
https://github.com/open-path/Green-River/issues/6572

## Questions / To Dos
* Should have a better way to indicate the "individual or family" item? the expectation is hard-coded.
* TBD exactly how the js "householdId" will be used to inform the actual `Enrollment.HouseholdID` during processing. We could use it directly or create a consistent hash of it (md5). We should definitely check before creation that the household id does not already exist in a different project.
* js `Date.now` compat 
* Update copy on alert and modals (design sesh thurs)

## How to test
* Set up [this form](https://github.com/open-path/Green-River/issues/6544#issuecomment-2329739980) locally as an EXTERNAL_FORM
* Add an external_form_object_key
* Add `AppConfigProperty` with `key: "external_forms/presign_url"` and `value: "/hmis_external_api/external_forms/presign"`
* Comment out captcha code from JS
* Test at `https://hmis-warehouse.dev.test/hmis_external_api/external_forms/<your key>`

## Screenshots

Screenshot 2024-09-04 at 2 14 12 PM

==>(if "Family" selected)

Screenshot 2024-09-04 at 2 15 15 PM

==>(if "Add another household member" clicked)

Screenshot 2024-09-04 at 2 15 29 PM

## Type of change


drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
```diff
@@ -35,10 +35,6 @@
         max-width: 540px;
         margin-bottom: 4rem;
       }
-      .dependent-form-group > .form-group {
-        padding-left : 1.25rem;
-        border-left: 5px solid #eee;
-      }
```
gigxz, line 41: This was indentation with a gray vertical bar for conditional items. This is not practical for PIT because there are items that are conditional on value from a question earlier in the form, that should not be indented

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_household_warning.haml
```diff
@@ -0,0 +1,7 @@
+-# household warning is hidden by default
+%div{id: 'householdWarning', class: 'alert alert-info', style: 'display:none;'}
+  %p
+    %span.font-weight-bold= 'Warning:'
+    You are adding a member to an existing household. If you are trying to add a new member that is NOT part of the previous household, click the below button to refresh.
```
gigxz: *Copy needs updating

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form.haml
```diff
@@ -1,5 +1,27 @@
...
+-# { "link_id": "individual_or_family",
+-#   "type": "CHOICE",
+-#   "pick_list_options": [{"code": "Individual"}, {"code": "Family"}]
+-# }
+-#
+-# If this link_id is NOT present, none of the household-specific JS will be loaded.
+-# If this link id IS present, the form will generate `householdId` for each submission, and link future submissions to previous submissions as specified by the user.
+- individual_or_family_node = @form_definition.link_id_item_hash['individual_or_family']
+- if individual_or_family_node
+  - raise 'individual_or_family item doesnt match expectation' unless individual_or_family_node['type'] == 'CHOICE' && individual_or_family_node['pick_list_options'].find { |o| o['code'] == 'Family' }
```
gigxz, line 19: 🚨 so far, I built this with a hard-coded expectation that there is a very specific "individual or family" question, which determines whether the household js is loaded or not. If "family" is selected, the user will have the option to add another household member after submission.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -36,6 +47,14 @@ $(function () {
       formData[item.name] = item.value;
     });
 
+    if (window.householdId) {
+      // If form collects household information, add householdId and relationshipToHoH to submission
+      formData['Enrollment.householdId'] = window.householdId;
+      if (!formData['Enrollment.relationshipToHoH']) {
+        formData['Enrollment.relationshipToHoH'] = window.isFamilyMember ? 'DATA_NOT_COLLECTED' : 'SELF_HEAD_OF_HOUSEHOLD';
+      }
+    }
```
gigxz, line 59: @martha let's discuss. The form as-written does not ask a question about Relationship To HoH, so I added this. 

However I think it would be safer to also implement this logic in the backend. That way, if someones browser is behaving in a way we don't expect, and householdId is not present for whatever reason, they could still be processed into valid Enrollments (albeit all with exclusively single-member households).

The backend  logic would be something like:
* If HouseholdID present on the submission and is valid (doesnt exist in another project),
  * If relationshipToHoH is also present on the submission, use that
  * Else set relationshipToHoH to 1 if householdid not found, 99 otherwise
* Else,
  * Generate a new HouseholdID, set RelationshipToHoH to 1

martha: Thanks @gigxz , makes sense. I've taken a stab at that logic here: https://github.com/greenriver/hmis-warehouse/pull/4657/commits/10c12c6b5724c3788c9b8052263c3e0bdc63ec91

Do we need to add support for a hidden form item in order to submit it as a form value? If not, we probably need to change the behavior of the form processor so it doesn't raise when it encountered an unexpected value based on the form definition. https://github.com/greenriver/hmis-warehouse/blob/bdaf04c01458bb256af19ce3fbfd7b9e397a5f1e/drivers/hmis/app/models/hmis/form/form_processor.rb#L78

ttoomey:
note, for testing you can render with
"https://hmis-warehouse.dev.test//hmis_external_api/external_forms/<external_object_key>:

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -10,11 +11,21 @@ $(function () {
...
   }
 
   $('.reload-button').on('click', function () {
-    window.location.reload();
+    window.location = window.location.pathname; // drop hh_id param
+  });
+
+  $('#addAnotherHouseholdMemberButton').on('click', function () {
+    window.location = window.location.pathname + '?hh_id=' + window.householdId;
   });
```
ttoomey: these don't seem to work for me since it doesn't include the origin, at least locally. These seem to work though
```js
  $('.reload-button').on('click', function () {
    const url = new URL(window.location.href);
    window.location.href = url.origin + url.pathname;
  });

  $('#addAnotherHouseholdMemberButton').on('click', function () {
    const url = new URL(window.location.href);
    url.searchParams.set('hh_id', window.householdId);
    window.location.href = url.toString();
  });
```

ttoomey (commented):
Looks like it should work. Some of this may break our current external form, we should test that out

Originally the form rendering code was intended to be generic and support rendering relatively simple forms. We're departing from that here; as form-specific behaviors are added, this could become hard to maintain. It should get us through the PIT this year though. 

Perhaps we can find another solution down the road.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -118,11 +137,69 @@ $(function () {
...
 });
 
+window.addHouseholdTypeListener = function (individualOrFamilySelector) {
+  // Generates a "household id" that is unlikely to have collisions among form submissions,
+  // without using external library (uuid) or modern browser features (crypto)
+  var generateHouseholdId = function () {
+    var current = Date.now().toString(); // OK? https://caniuse.com/mdn-javascript_builtins_date_now
```
martha: can use Date.getTime()?

martha (approved):

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: Weird, it works for me (and martha) locally. What behavior are you seeing?
I will update to your suggestion

ttoomey: I didn't look into it too closely but it seemed to be dropping host, sending the browser to something like:
http:///hmis_external_api/external_forms/tt-pit

I can try to replicate if it needs investigation

gigxz:
@ttoomey 

> Some of this may break our current external form, we should test that out

Is there something specific you see that you think is breaking it, or is this just a general note? I have tested the prevention screening form plenty, but I may have missed something.


> Originally the form rendering code was intended to be generic and support rendering relatively simple forms. We're departing from that here; as form-specific behaviors are added, this could become hard to maintain. It should get us through the PIT this year though.

Yes. I think we'll have to see what our next needs are for external forms. For a potential future CE access point form, I don't _think_ the household piece will make sense, but the Client/Enrollment processing for HoH likely will be needed. For future PITs, the household part should be reusable, but might need tweaking.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: No that's fine, just curious. I assume it is because you have a different local dev setup

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: that does seem to have better coverage, thanks

PR #4675, opened 2024-09-06
Add geolocation component to external form

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6481

Tested using an external form that includes the following item:
```
{
  "text": "Where are you located?",
  "type": "GEOLOCATION",
  "link_id": "geolocation",
  "mapping": {
    "field_name": "coordinates",
    "record_type": "GEOLOCATION"
  }
},
```

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/form/record_type.rb
```diff
@@ -68,6 +68,11 @@ def self.all
           owner_type: 'Hmis::Hud::YouthEducationStatus',
           processor_name: 'YouthEducationStatus',
         ),
+        new(
+          id: 'GEOLOCATION',
+          owner_type: 'Hmis::Hud::Enrollment',
+          processor_name: 'Geolocation',
```
martha, line 74: Processor doesn't exist yet - will be part of https://github.com/open-path/Green-River/issues/6541

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_generator.rb
```diff
@@ -128,6 +131,12 @@ def render_choice_node(node)
       end
     end
 
+    def render_geolocation_node(node)
+      render_form_group(node: node) do
+        render_form_geolocation(label: node['text'], required: node['required'], name: node_name(node))
```
martha, line 147: The way I have it here, unlike the other `render` input methods, `render_form_geolocation` doesn't generate a new HTML id using `next_html_id` and pass it to the partial; instead, inside the partial, the html ID for the hidden input is generated and the label applies to that.

Maybe this is too topsy turvy, so I want to revisit it and match the existing pattern

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
```diff
@@ -0,0 +1,101 @@
...
+        marker = L.marker([lat, lon]).addTo(map);
+        input.val(JSON.stringify({ "latitude": lat, "longitude": lon }));
+      }
+
+      var onError = function(error) {
+        getLocationButton.prop('disabled', false);
+        mapContainer.hide();
+        clearButton.hide();
+        helperText.text("Location unavailable.", error.message);
+        input.val('unavailable');
```
martha: This is a first pass at the desired "required" behavior that we discussed yesterday:
- clicking the button is required
- but the user is allowed to clear and still submit
- user can also submit if geolocation attempt resulted in an error, or geolocation isn't available in their browser

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
```diff
@@ -0,0 +1,108 @@
+:ruby
```
gigxz, line 1: woohoo 🥳

drivers/hmis_external_apis/public/schemas/form_definition.json
```diff
@@ -957,6 +973,19 @@
...
     },
+    "geolocationFormNode": {
+      "allOf": [
+        {
+          "$ref": "#/$defs/genericInputFormNode"
+        },
+        {
+          "type": "object",
+          "properties": {
+            "component": "COORDINATES"
```
gigxz: Is this left over from a previous iteration? I don't see a component type for this (and don't think we need one?)

martha: You're right, it's not needed! I was getting the hang of the json form schema by just copying other elements, but this one doesn't need its node type, it can use the generic one.

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
martha: To do: Test on device what happens if the user clicks the button but refuses the browser permission to access their location. Form should still be submittable.

drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
```diff
@@ -12,6 +12,8 @@
       %script{src: "https://www.google.com/recaptcha/api.js?render=#{page_config.recaptcha_key}"}
 
     %link(rel="stylesheet" href="https://cdn.jsdelivr.net/npm/bootstrap@4.6.2/dist/css/bootstrap.min.css" integrity="sha384-xOolHFLEh07PJGoPkLv1IbcEPTNtaed2xpHsD9ESMhqIYd0nLMwNLD69Npy4HI+N" crossorigin="anonymous")
+    %link(rel="stylesheet" href="https://unpkg.com/leaflet@1.9.4/dist/leaflet.css" integrity="sha256-p4NxAoJBhIIN+hmNHrzRCf9tD/miZyoHS5obTRR9BMY=" crossorigin="")
+    %script(src="https://unpkg.com/leaflet@1.9.4/dist/leaflet.js" integrity="sha256-20nQCchB9co0qIjJZRGuk2/Z9VM+kNiyxNV1lvTlZBo=" crossorigin="")
```
gigxz: It would be preferable to _only_ include the leaflet js if this is for a form that collects geolocation (aka, only if the geolocation partial is included). Looks doable, I think: https://stackoverflow.com/a/10226010

drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
```diff
@@ -70,6 +72,9 @@
       fieldset > .form-group:has(> .form-check) {
           margin-bottom: 0px;
       }
+      .leaflet-container {
+        height: 500px;
```
gigxz: Looking on mobile using dev tools, I would  suggest smaller (300px) - but we can see how it looks on a real device

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
```diff
@@ -0,0 +1,108 @@
...
+      var getLocationButton = $(getLocationButtonId);
+      var clearButton = $(clearButtonId);
+      var mapContainer = $(mapContainerId);
+      var input = $(inputId);
+
+      helperText.hide();
+      mapContainer.hide();
+      clearButton.hide();
+
+      var map = L.map(#{map_id.to_json.html_safe});
```
gigxz: I think we may want to add some more options https://leafletjs.com/reference.html#map-option but I haven't reviewed them all. I guess my first thought it that it doesn't need to be draggable, since you can't move the pin–– and it might be annoying if you are trying to scroll down the page, but you accidentally drag the map instead.
I tried `{ dragging: false, boxZoom: false, tapHold: false, touchZoom: false}`.
We can also wait on this until we have it on a device and can get some feedback!

Another thought, re: mobile ux, would be that it _might_ be nice to provide an option to hide the map once you're finished collecting geolocation. But again, we can see how it feels on device.

gigxz (changes requested):
Looks great!! requested change is to avoid including leaflet on the page for forms that doesn't need it

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
gigxz: Thanks Martha. Sorry we didn;t get a chance to sync up on that- my takeaway was that it would also be _fine_ to have the geolocation be optional, because they could rely on training to make sure folks use it. I think that is probably the safer option–– safer as in, we less chance of someone being unable to submit forms because of device issues. Thoughts?

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
martha: Yeah that makes sense, I agree it will be better UX to disable most interaction with the map so users can scroll down the page in peace. I looked through the other options and disabled a few more. It's now still possible to zoom using the zoom in/out buttons or when focused on the element using keyboard nav. But it won't zoom or pan with normal click/tap interactions. Let's test it out on devices for sure

gigxz (approved):
📍 👍

drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
martha: Thanks yes good point 🙏

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
gigxz: perfect, thanks!

drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
martha: Sounds good. I think either behavior is supported right now. If the item has `required` set to true, then the behavior is what I described in my earlier comment. Otherwise geolocation is fully optional.

gigxz: great 👍

PR #4680, opened 2024-09-10
Process external form geolocation into CLH

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6541

This PR adds form processing for geolocation from external form submissions. This involves
- Two new columns on the `clh_locations` table
  - a reference to `hmis_form_processors`
  - a new `located_at` column to store more granular timestamp info, since the existing `located_on` stores date only
- New `GeolocationProcessor` for processing the CLH
- A new `before_create` hook on the CLH for adjusting the polymorphic source so it always points to a GRDA record, not an HMIS record
- Spec tests

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -55,6 +55,7 @@ class Hmis::Hud::Client < Hmis::Hud::Base
   has_many :hmis_services, through: :enrollments # All services (HUD and Custom)
   has_many :services, through: :enrollments # HUD Services only
   has_many :custom_services, through: :enrollments # Custom Services only
+  has_many :client_location_histories, class_name: 'ClientLocationHistory::Location'
```
eanders: This should be added in the driver:
```
drivers/client_location_history/extensions/grda_warehouse/hmis/client_extension.rb
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -59,6 +59,7 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
   has_many :current_living_situations, **hmis_enrollment_relation('CurrentLivingSituation'), inverse_of: :enrollment, dependent: :destroy
   has_many :employment_educations, **hmis_enrollment_relation('EmploymentEducation'), inverse_of: :enrollment, dependent: :destroy
   has_many :youth_education_statuses, **hmis_enrollment_relation('YouthEducationStatus'), inverse_of: :enrollment, dependent: :destroy
+  has_many :enrollment_location_histories, class_name: 'ClientLocationHistory::Location', as: :source
```
eanders: This should be added in the driver:
```
drivers/client_location_history/extensions/grda_warehouse/hmis/enrollment_extension.rb
```

drivers/hmis/app/models/hmis/hud/client.rb
martha: Thank you @eanders is there any documentation you recommend about Rails drivers? I think there is some railsy magic going on (or rather, trying to go on, and not working correctly on my attempts) that I should understand better.

martha: https://github.com/degica/rails_drivers ?

eanders: That's the one!
https://github.com/degica/rails_drivers

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -0,0 +1,46 @@
...
+        @processor.send(factory_name).assign_attributes(attribute_name => value)
+      end
+    end
+
+    def assign_metadata
+      clh = @processor.send(factory_name, create: false)
+      return if clh&.destroyed?
+
+      clh&.assign_attributes(
+        source: @processor.enrollment_factory,
```
gigxz, line 50: untested, comment for linking to other issue..
```
@processor.enrollment_factory.as_warehouse_unpersisted # clh_locations should reference grdawarehouse enrollment

...

def as_warehouse_unpersisted
   raise 'bad' if persisted?
   GrdaWarehouse::Hud::Enrollment.new(self.attributes)
end
```

martha: Thanks @gigxz this is an interesting idea, thought I haven't gotten it working. Both records get saved as separate enrollments with separate IDs, and the CLH is attached to the second one (not the one that gets related to the submission).

martha: Here's one idea that maybe works but feels pretty hacky to me: https://github.com/greenriver/hmis-warehouse/pull/4680/files#diff-81d5acd9511ddd9bf95cd056ce5b5252916e30476d6fdb1b3315c85e7d9a4616R15-R21

drivers/hmis/app/models/hmis/hud/client.rb
martha: Thanks! This ended up working instead: `drivers/client_location_history/extensions/hmis/hud/client_extension.rb`

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
```diff
@@ -41,8 +41,8 @@ def resolve(id:, input:, project_id: nil)
             error_out(record.enrollment.client.errors.full_messages) unless record.enrollment.client.valid?
             error_out(record.enrollment.errors.full_messages) unless record.enrollment.valid?
 
-            record.enrollment.client.save!
             record.enrollment.save_new_enrollment!
```
martha: Save the enrollment first; otherwise the enrollment gets saved alongside the client as a related record, but then `save_new_enrollment` raises because it's not a new enrollment. (The reason to use `save_new_enrollment` is to respect the project's auto enter config)

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -0,0 +1,46 @@
...
+        latitude = attribute_value[:latitude]
+        longitude = attribute_value[:longitude]
+        not_collected_reason = attribute_value[:notCollectedReason]
+        raise ArgumentError, 'Geolocation coordinates in unexpected format' unless (latitude && longitude) || not_collected_reason
+
+        return @processor.send(factory_name).destroy if not_collected_reason
+
+        @processor.send(factory_name).assign_attributes(lat: latitude, lon: longitude)
+      else
+        @processor.send(factory_name).assign_attributes(attribute_name => value)
```
gigxz: Do we ever hit this code? If we don't expect any other fields, maybe we should raise an error.

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
gigxz: Oh well, that makes sense. I don't mind the after_save solution..

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -0,0 +1,46 @@
...
+      end
+    end
+
+    def assign_metadata
+      clh = @processor.send(factory_name, create: false)
+      return if clh&.destroyed?
+
+      clh&.assign_attributes(
+        source: @processor.enrollment_factory,
+        located_on: @processor.enrollment_factory.entry_date, # entry_date holds the submission time of the form
```
gigxz: Hmm that's true for PIT, but won't be true for, say, collecting geolocation on a current living situation. (In that case, it would be `@processor.current_living_situation.information_date` aka `@processor.owner.information_date`.

Maybe we can solve that by having the CLS form explicitly set the located_on using a form item (`mapping: {record_type:GEOLOCATION, field_name: located_on}`), and have the processor only set it to `@processor.enrollment_factory.entry_date` as a default (if `@hud_values['Geolocation.locatedOn']` doesn't exist..?)

Or- we could just cross this bridge when we implement https://github.com/open-path/Green-River/issues/5726. Just wanted to point it out.

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -0,0 +1,46 @@
...
+        processed_at: Time.now,
```
gigxz:
```suggestion
        processed_at: Time.current,
```

preferred over Time.now because of timezone issues

drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
```diff
@@ -72,6 +72,9 @@
       fieldset > .form-group:has(> .form-check) {
           margin-bottom: 0px;
       }
+      .hide {
```
gigxz: We import the bootstrap css, which should have a class for this. Did you try `hidden` or `d-none`?

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
martha: I'm in favor of kicking the can. I'll add a comment, though

drivers/hmis_external_apis/app/views/layouts/hmis_external_apis/external_forms.haml
martha: It seems like d-none works but sets `!important`, preventing the jquery `.hide()`/`.show()` calls from functioning correctly. edit: Fixed by toggling the class using jquery instead.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: hrm, you are right to 👀 , something is still wonky here. This does not work when auto enter is turned on in the project; the client has to already exist when save_new_enrollment is called.

martha: Ah, so it's specifically an issue when geolocation is collected. Since location is related to both client and enrollment, when client is saved first, location is saved, and enrollment is persisted too. Then the 'Unexpected: save_new_enrollment called on a persisted enrollment' error is thrown when we try to save the new enrollment on the next line.

Ideas for workarounds:
- Does `save_new_enrollment!` _need_ to raise if the enrollment is already persisted? I feel a bit hesitant to change this when it may have far-reaching consequences, but grepping in the code it does not actually look that widely used. 
- On the other hand: Setting the location's `source` to the HMIS enrollment before it has been persisted is already proving tricky for us, so maybe this is one more prompting in the direction of post-processing location after everybody else has already been persisted (instead of keeping the hack in `after_save`)

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
martha: Edit: Per discussion here https://github.com/greenriver/hmis-warehouse/pull/4696#discussion_r1757167999, seems like it would be helpful to update this so that the CLH record holds the timestamp of the form submission, for the sake of marker display on the map. What I've proposed here is to stuff that into the `processed_at` field, even though it feels a bit unintuitive to me since I would expect `processed_at` to hold the processing (admin review) timestamp and not the submission timestamp.

I guess another approach we could take is to migrate the located_on col to be a datetime instead of a date. Or, to add a new col located_at which would hold the full timestamp. @gigxz what do you think?

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -0,0 +1,53 @@
...
+
+    def information_date(date)
+      clh = @processor.send(factory_name, create: false)
+      return if clh&.destroyed?
+
+      # TODO(#5726) This logic may need to be updated for CLS.
+      # Maybe unintuitive, but for PIT, we store the timestamp of the form submission in the processed_at field because
+      # it is a granular timestamp and not a date like located_on.
+      clh&.assign_attributes(processed_at: date, located_on: date)
+    end
```
gigxz: Using the `information_date` callback feels hard to reason about here. It's current purpose is to set HUD InformationDate fields on related records for HUD Assessments. Maybe we could simplify this by moving the logic to the assign_metadata method, which explicit cases for the situations we support:



```
      owner = @processor.owner_factory
      case owner
      when HmisExternalApis::ExternalForms::FormSubmission
        clh&.assign_attributes(
          # special case for processed_at to equal submission tiem, if we want ot keep that approach
          processed_at: owner.submitted_at, located_on: owner.submitted_at.to_date,
        )
      when Hmis::Hud::CurrentLivingSituation
        clh&.assign_attributes(
          processed_at: Time.current, located_on: owner.InformationDate,
        )
      when Hmis::Hud::CustomAssessment
        clh&.assign_attributes(
          processed_at: Time.current, located_on: owner.AssessmentDate,
        )
      end
```

or, simplified:
```
      owner = @processor.owner_factory
      located_on = case owner
      when HmisExternalApis::ExternalForms::FormSubmission
        owner.submitted_at.to_date
      when Hmis::Hud::CurrentLivingSituation
        owner.InformationDate
      when Hmis::Hud::CustomAssessment
        owner.AssessmentDate
      else
        raise 'unable to determine located_on date for client location'
      end

      clh&.assign_attributes(
        source: @processor.enrollment_factory,
        collected_by: @processor.enrollment_factory.project.name,
        located_on: located_on,
        processed_at: Time.current,
      )
```

I think I'm in favor of adding another column like `located_at` to store the timestamp when the location was collected, rather than overloading the `processed_at` field which currently has a different meaning. lets discuss at standup.

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -106,7 +106,7 @@ def run_form_processor(current_user, project: nil)
         build_enrollment(
           project: project,
           data_source: project.data_source,
-          entry_date: created_at,
+          entry_date: submitted_at, # Enrollment entry date is set to the date of form submission
```
gigxz: I'm not sure about that, was that spec'd somewhere?
It doesn't feel right that if a form was submitted and reviewed 1 week later, than the Entry Date would be backdated. I would assume it's most intuitive for the entry date to be equal to the date when the form was reviewed.

drivers/client_location_history/app/models/client_location_history/location.rb
```diff
@@ -10,7 +10,17 @@ class Location < GrdaWarehouseBase
     belongs_to :source, polymorphic: true, optional: true
     belongs_to :client, class_name: 'GrdaWarehouse::Hud::Client'
     belongs_to :place, class_name: 'GrdaWarehouse::Place', primary_key: [:lat, :lon], foreign_key: [:lat, :lon], optional: true
-    belongs_to :enrollment, class_name: 'GrdaWarehouse::Enrollment', optional: true
+    # this relation isn't used; use polymorphic `source` above
+    belongs_to :enrollment, class_name: 'GrdaWarehouse::Hud::Enrollment', optional: true
+    after_save :fix_source
+
+    def fix_source
```
gigxz: maybe `ensure_grda_warehouse_source`

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
eanders: I think I'd be inclined to add a new column.  We could coalesce `located_at` with `located_on` for display purposes.

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
martha: Discussed on a call, we'll use `before_create` instead so it doesn't cause an extra roundtrip to the DB.

drivers/hmis/app/graphql/mutations/update_external_form_submission.rb
martha: Discussed on a call, we will keep `save_new_enrollment` as is, since in general we only want to call this method on a brand new enrollment. We'll separately call similar code `save_and_auto_enter` from the external forms mutation keeping the impact localized.

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
martha: I'll add that coalescence on the locations map PR https://github.com/greenriver/hmis-warehouse/pull/4696.

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
martha: Thanks, perfect!

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
martha: Fixed 👍

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -284,16 +284,27 @@ def self.generate_enrollment_id
...
+    if should_auto_enter?
+      save_and_auto_enter!
     else
-      # Otherwise, save as WIP
       save_in_progress!
     end
   end
 
+  def should_auto_enter?
+    Hmis::ProjectAutoEnterConfig.detect_best_config_for_project(project)
```
gigxz: Methods with `?` by convention should return a boolean (can tack on `.present?`)

gigxz (approved):
looks good!

eanders (approved):
Thanks @martha, this is excellent.

PR #4682, opened 2024-09-10
Error on unauthorized object; adjust ClientAccessLoader logic for bad data

gigxz:

## Description

* Fix a bug that would arise when there were undeleted enrollments at deleted projects. The Client.viewable_by scope joins the client with _projects_, so the ClientAccessLoader is updated to do the same. (This mismatch was causing the authorized check to fail)
* Raise an error when authorized check fails on an object. 

https://github.com/open-path/Green-River/issues/6641

Bug reproduction:

```rails
client = Hmis::Hud::Client.hmis.find { |c| c.enrollments.count == 1 }
deleted_project = Hmis::Hud::Project.only_deleted.last
en = client.enrollments.first
en.update!(project: deleted_project)
# search for the client in HMIS by name
# OLD BEHAVIOR: an exception will be raised because they are unauthorized
# NEW BEHAVIOR: no error, client is returned. they appear to have no enrollments in hmis
```

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/spec/requests/hmis/client_search_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/client_search_spec.rb
@@ -284,7 +284,7 @@ RSpec.describe Hmis::GraphqlController, type: :request do
         # client has 1 non-deleted Enrollment that is at a deleted Project (specific regression scenario)
         enrollment = create(:hmis_hud_enrollment, client: client, data_source: ds1)
         enrollment.project.delete
-        expect(enrollment.reload.project).to be_nil # confirm setup
+        raise unless enrollment.reload.project.nil? # confirm setup
       end
 
       it 'should return the client' do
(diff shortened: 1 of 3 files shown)


eanders (approved):

drivers/hmis/spec/requests/hmis/client_search_spec.rb
```diff
@@ -262,6 +262,40 @@ def perform_search(input = { text_search: 'Darlene Ranger' })
...
+
+      expect_gql_error perform_search, message: /failed authorization check/
+    end
+
+    context 'when client has an Enrollment at a deleted project (regression #6641)' do
+      before(:each) do
+        # client has 1 non-deleted Enrollment that is at a deleted Project (specific regression scenario)
+        enrollment = create(:hmis_hud_enrollment, client: client, data_source: ds1)
+        enrollment.project.delete
+        expect(enrollment.reload.project).to be_nil # confirm setup
```
ttoomey: Micro-nit, feel free to ignore. I wonder if we should try and limit assertions/expectations to only within expect blocks. If the need is to verify our setup then perhaps just throwing an exception would be clearer?

```suggestion
        raise unless enrollment.reload.project.nil? # confirm setup
```

ttoomey (approved):
Looks great! Thank you so much tracking this down 🙌

PR #4684, opened 2024-09-10
Raise on empty client search in HMIS

gigxz:

## Description

I noticed this while working on tests for https://github.com/greenriver/hmis-warehouse/pull/4682: the clientSearch query allows receiving an empty input object, which will resolve all viewable clients. The frontend doesn't expose a way to do it, but the api allowed it.

This PR removes the ability to search without any search parameters. 

Note: `ClientSearchUtil::NameSearch` already forbids _text search_ where the search term is < 3 chars. In that case, it will just return empty.

## Type of change
- [x] Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
@@ -39,6 +39,8 @@ module Types
     end
 
     def client_search(input:, **args)
+      raise 'Invalid search. At least 1 search param is required.' unless input.to_h.excluding(:projects, :organizations).values.compact_blank.any?
+
       # if the search should also sort by rank
       sorted = args[:sort_order] == :best_match
       search_scope = Hmis::Hud::Client.client_search(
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -39,6 +39,8 @@ def organizations(**args)
     end
 
     def client_search(input:, **args)
+      raise 'Invalid search. At least 1 search param is required.' unless input.to_h.excluding(:projects, :organizations).values.compact_blank.any?
```
eanders, line 42: If this gets bubbled up to the UI, maybe we should make the language a bit less techy?   "Your search cannot be completed, at least one search term is required."

eanders (approved):

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: `raise '..'` don't get bubbled up to the UI in production. In prod, users will see "an error occurred, it is being investigated, yadayada" (but we get the real message in sentry). I'm `raise`-ing here because it is an unexpected input-- the frontend application does not expose a way to pass no search params.

eanders: Perfect then.  Thanks.

PR #4691, opened 2024-09-11
Fake enrollment data generator

gigxz:

## Description

https://github.com/open-path/Green-River/issues/6101

Job to generate fake HMIS Enrollment data for local development and staging environments. It generates households of varying sizes. The default behavior is to generate open enrollments for existing clients.

Records it generates:
- Clients (optionally)
- Enrollments (Household size varies from 1-4, except for NBN shelters which are set to 1)
- Exits (optionally)
- Disabilities (only for intake stage)
- Bed Night Services (for NBN projects, 1-100 per Enrollment)
- Intake and Exit Assessments

Usage:
  Hmis::CreateFakeEnrollmentsJob.perform_now(num_households: 10)
  Hmis::CreateFakeEnrollmentsJob.perform_now(num_households: 10, generate_clients: true, exited: true, project_ids: [72])

#### Future improvements
* generate clients <18y
* cohesive households (eg all household members have same last name)
* generate other intake assessment records (only does disability currently)
  * IncomeBenefit
  * HealthAndDv
  * EmploymentEducation
  * YouthEducationStatus
* generate records for exit assessment
* generate records for annual assessment
* generate other HUD service types (only bed night currently)
* generate Current Living Situation

## Type of change


gigxz:
to add:
-trigger warehouse processing
-test performance on larger sets

martha (approved):
This looks good, thank you @gigxz. I ran it a couple of times in my local environment, with small n, trying different combinations of arguments (with and without generating clients, exits, etc.) Everything worked fine. Seems really useful!

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
```diff
@@ -0,0 +1,311 @@
...
+      export_id: 'FAKED'
+    )
+      raise "won't create fake enrollments in production" if Rails.env.production?
+
+      setup_notifier('CreateFakeEnrollmentsJob')
+      @data_source = if data_source_id
+        ::GrdaWarehouse::DataSource.hmis.find(data_source_id)
+      else
+        ::GrdaWarehouse::DataSource.hmis.sole # 'sole' will raise if there are >1 HMIS data sources
+      end
```
ttoomey, line 50: This will only ever be run manually, right? Maybe just pass the data source in directly to keep it simple?

gigxz: Yes but it's meant to be able to run on the long running queue, eg `perform_later(data_source_id: x...)` - so I think for that it's better to pass the ID? Although I'm not sure, maybe it doesn't matter

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
```diff
@@ -0,0 +1,311 @@
...
+            exited: exited,
+            projects: project_scope,
+          )
+        end
+        remaining -= n
+      end
+
+      # queue client deduplication and service history processing
+      Hmis::Hud::Client.last.send(:warehouse_identify_duplicate_clients) if generate_clients && !Rails.env.development?
+      Hmis::Hud::Enrollment.last.send(:queue_service_history_processing!)
```
ttoomey: Looks like the existing interface to enqueue these jobs don't support this well, specifically:
* we have call send to invoke protected methods
* these methods are more appropriately class scoped, since they don't just impact the client/enrollment instance on which they are called

This is scope creep but It might be worth making a small adjustment to allow these to be called as public class methods on Enrollment and Client
```ruby
      Hmis::Hud::Client.warehouse_identify_duplicate_clients if generate_clients && !Rails.env.development?
      Hmis::Hud::Enrollment.queue_service_history_processing!
```

```ruby
# Class method
class Hmis::Hud::Client < Hmis::Hud::Base

  def self.warehouse_identify_duplicate_clients
    # ...
  end

  # Instance method that calls the class method
  def warehouse_identify_duplicate_clients
    self.class.warehouse_identify_duplicate_clients
  end
end
```

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
```diff
@@ -0,0 +1,311 @@
...
+          personal_id: enrollment.personal_id,
+          date_created: to_datetime(dt),
+          date_updated: to_datetime(dt),
+          **hud_attributes,
+        )
+      end
+    end
+
+    def build_fake_exit(enrollment)
+      exit_date = [enrollment.EntryDate + rand(30..1000), today].min
```
ttoomey, line 234: If entry date + rand(1000) is greater than today, it seems like this could make exits in the future. Perhaps:
```ruby
exit_date = [exit_date, today].min
```

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
```diff
@@ -0,0 +1,311 @@
...
+      client_pool.sample(hh_size).each_with_index.map do |client, idx|
+        enrollment = build_fake_enrollment(
+          client: client,
+          project: project,
+          is_hoh: idx == 0,
+        )
+        enrollment.HouseholdID = hh_id if hh_id
+        hh_id ||= enrollment.HouseholdID
+        enrollment
+      end.flatten
```
ttoomey: it's harmless but probably don't need flatten
```suggestion
```

drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
```diff
@@ -43,7 +43,7 @@ class MigrateAssessmentsJob < BaseJob
     #  - DateCreated = earliest creation date of related records
     #  - DateUpdated = latest update date of related records
     #  - UserID = UserID from the related record that was most recently updated
-    def perform(data_source_id:, project_ids: nil, clobber: false, delete_dangling_records: false, preferred_source_hash: nil, generate_empty_intakes: false)
+    def perform(data_source_id:, project_ids: nil, clobber: false, delete_dangling_records: false, preferred_source_hash: nil, generate_empty_intakes: false, enrollments: nil)
```
ttoomey: suggest adding some documentation for these params. Also, is there a case for passing both project_ids and enrollment_ids? Seems like they might conflict. Perhaps should raise an ArgumentError if both are supplied
```suggestion
    # - enrollments = active record enrollment scope, only create records for enrollments in this scope; default is all enrollments at project_ids
    # - project_ids = array of integers, project pks, only create records for enrollments at these projects in the same data source
    def perform(data_source_id:, project_ids: nil, clobber: false, delete_dangling_records: false, preferred_source_hash: nil, generate_empty_intakes: false, enrollments: nil)
```

ttoomey (approved):
Looks great! This will be super helpful, thanks!

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
gigxz: Ah thanks, yeah that was a bit hacky. I updated it to use class methods as you suggested. I also removed the the dev check, because it's fine to run `warehouse_identify_duplicate_clients` in dev, it just seems to run in place instead of queuing (in development)

drivers/hmis/app/jobs/hmis/create_fake_enrollments_job.rb
gigxz: this already takes the minimum of exit date and today

drivers/hmis/app/jobs/hmis/migrate_assessments_job.rb
gigxz: Added argument error and param descriptions, thanks!

PR #4679, opened 2024-09-10
🍄 PIT External Forms (Deploy Branch)

gigxz:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

Deploy branch for PIT changes. All changes have been reviewed in individual PRs.

Frontend: https://github.com/greenriver/hmis-frontend/pull/916

PR #4696, opened 2024-09-12
Create project locations map

martha:

## Description
GH issue: https://github.com/open-path/Green-River/issues/6580

This PR,
- Adds a Map to the Warehouse Project page showing all locations associated with enrollments in this project
- Adds a new permission to control access to the map
- Adds a date filter similar to the client map

Screenshot 2024-09-12 at 2 20 02 PM

To do:
- [ ] Discuss comments/questions inline
- [x] Test with 1000s of markers

## Type of change
- [ ] Bug fix
- [ ] Dependency update


app/models/grda_warehouse/hud/project.rb
```diff
@@ -74,6 +73,8 @@ class Project < Base
     # project_coc has been deleted
     belongs_to :project_cocs_with_deleted, class_name: 'GrdaWarehouse::Hud::WithDeleted::ProjectCoc', primary_key: [:ProjectID, :data_source_id], foreign_key: [:ProjectID, :data_source_id], optional: true
 
+    include RailsDrivers::Extensions
```
martha, line 77: I moved this here to avoid the following error. I haven't run the full test suite yet so I don't feel confident this change isn't having any other negative impacts elsewhere.

`
Cannot have a has_many :through association 'GrdaWarehouse::Hud::Project#enrollment_location_histories' which goes through 'GrdaWarehouse::Hud::Project#enrollments' before the through association is defined. (ActiveRecord::HasManyThroughOrderError)
`

app/controllers/projects_controller.rb
```diff
@@ -62,6 +63,46 @@ def destroy
     }
   end
 
+  private def set_location_params
+    return unless RailsDrivers.loaded.include?(:client_location_history)
+
+    @locations = @project.enrollment_location_histories.where(located_on: location_filter.range)
+    @markers = @locations.map { |l| l.as_marker_with_name(current_user) }
```
martha: What should go on the marker? Right now it shows:
- Name
- Seen on - this is less useful for PIT since it's a date column, not a timestamp.
  - Proposal: update CLH date processing to set `processed_at` to the timestamp of form submission? https://github.com/greenriver/hmis-warehouse/pull/4680/files#diff-2a90bfd47b853e15be95aa8a3449e1f74df7264d603a1fd070503b563ea48de7R42-R43 I'm not sure how generic this solution is, but our existing solution is also not generic, and it seems like it would work fine for PIT since we already store the timestamp of processing in `created_at`/`updated_at`.
- Collected by - this will always be the project, so it's less useful to show from the project page
  - Proposal: Remove

app/views/projects/show.haml
```diff
@@ -121,6 +121,10 @@
   %p The following import overrides are in effect for this project.
   .mb-4.card
     = render 'hmis_csv_importer/import_overrides/table', overrides: overrides, data_source: @project.data_source, editable: false, show_associated_project: false
+
+- if RailsDrivers.loaded.include?(:client_location_history) && can_view_project_locations?
```
martha: To discuss: `can_view_project_locations?` permission. I realized it's the first time I have tried to work with granular permissions in the warehouse and it seems it's set up a bit differently. Are there separate Collections and User Groups in WH, or does WH use the same ones HMIS uses?

app/controllers/projects_controller.rb
martha: Screenshot 2024-09-12 at 2 13 12 PM

app/controllers/projects_controller.rb
```diff
@@ -62,6 +63,45 @@ def destroy
...
+    @bounds = ClientLocationHistory::Location.bounds(@locations)
+    @options = {
+      bounds: @bounds,
+      cluster: true,
+      marker_color: '#72A0C1',
```
martha: Should we keep this hardcoded hex value for consistency? Should we put it in some common location? Matches with:
- `drivers/client_location_history/app/controllers/client_location_history/clients_controller.rb`
- `drivers/client_location_history/app/controllers/client_location_history/warehouse_reports/client_location_history_controller.rb`

app/assets/stylesheets/application/modules/_projects.scss
```diff
@@ -0,0 +1,4 @@
+.c-filters {
+  // Override the filter card's usual large bottom margin, so it looks proportional when rendered as part of the Project view
+  margin-bottom: 1em;
+}
```
eanders: Pretty sure this will override that everywhere, even though it's in the `projects` module.  (We load all CSS on all pages pretty much.)  Probably better to just add `.mb-1` to the filter div on the projects page to achieve this in one place.

app/controllers/projects_controller.rb
eanders: That all sounds good to me.

app/controllers/projects_controller.rb
```diff
@@ -10,6 +10,7 @@ class ProjectsController < ApplicationController
   before_action :require_can_edit_projects!, only: [:edit, :update]
   before_action :set_project, only: [:show, :update, :edit, :destroy]
   before_action :set_census_params, only: [:show]
+  before_action :set_location_params, only: [:show]
```
eanders: Would we consider making this a sub-page of the Project page?  I think we have rough plans to move the client list off the bottom already, and this page seems to be doing a TON of different stuff right now.

app/controllers/projects_controller.rb
eanders: Probably worth moving it to a constant or class method on `ClientLocationHistory::Location` for now.  At least that will centralize it, and we could potentially move it to the DB if we decide we need to.  Maybe?
```
marker_color: ClientLocationHistory::Location::MARKER_COLOR
```

app/models/grda_warehouse/hud/project.rb
eanders: That makes sense.  Sometimes we drop it at the bottom of the file, not sure if that's necessary here, or desirable.  Can you add a comment on line 75 to indicate it needs to come after `has_many :enrollments` so we don't accidentally move it back up?

app/views/projects/show.haml
eanders: The warehouse has separate collections and user groups.  This will check globally if the user can see any project locations.  It's probably worth checking if the user can `can_view_project_locations?` for the given project.  It's Friday, so I'm probably missing it, but I'm not sure we have a method for that yet.  Essentially we need to call:
```
GrdaWarehouse::Hud::Project.viewable_by(current_user, permission: :can_view_project_locations).where(id: @project.id).exists?
```

eanders: If you want to turn that into a method on Project:
```
def can?(user, permission: :can_view_projects)
  self.class.viewable_by(user, permission: permission).where(id: id).exists?
end
```

I wouldn't be opposed.  Then the check could be `@project.can?(current_user, permission: :can_view_project_locations)` which feels a bit cleaner.

martha: Sounds good. I wonder if it's worth trying to more closely match the hmis pattern for this, which I think lives here in this `#{permission}_for?` helper: https://github.com/greenriver/hmis-warehouse/blob/e28ca6f6c13632c733079bcb515ca4904539d36a/drivers/hmis/app/models/hmis/user.rb#L134-L137

martha: But maybe that'd just result in a lot more code duplication, like duplicating the `EntityAccessLoaderFactory` https://github.com/greenriver/hmis-warehouse/blob/e28ca6f6c13632c733079bcb515ca4904539d36a/drivers/hmis/app/models/hmis/entity_access_loader_factory.rb#L24

eanders: Hmm, I'm not opposed to a more generic solution, but I'm liking the simplicity of a `can?` method. With `permission_for?` we'd need to define similar methods on each entity (I think that's kinda-sorta what EntityAccessLoaderFactory is going.  An alternative route, that's probably more inline with where we've been going is to make a class that arbitrates this.  So you could call something like:

`PermissionProvider.new(user, entity).can?(permission)`

Then we could encapsulate the calculations for what's possible for each entity in there.   That potentially allows us to do some batch calculations as well.

I'm not sure I have the clock cycles to devote to thinking through that whole thing right now, but if you think that's a better approach, we could spend some time on it in the near future.

app/controllers/projects_controller.rb
gigxz: I think that would be fine @eanders but what is the design pattern for that? I wasn't sure how that should look since the Project dashboard doesn't have tabs.

I am seeing two patterns:
* `Project Contacts.......View/Edit` as a table row (link to a new nested page)
* Buttons on the left that go to a new nested page, like "Configure Synthetic Assessments" and "Edit Project"


The first one is probably better, out of the two.. `Client Locations ..... View`

eanders: Yep, I think the first is the right approach.  In some future iteration we can improve that page and sub-navigation overall.

gigxz: Okay lets go with that @martha - sorry for steering you towards using the existing page, should have checked with warehouse team first!

app/models/role.rb
```diff
@@ -554,6 +554,12 @@ def self.permissions_with_descriptions
         category: 'Client Extras',
         sub_category: 'General Client Access',
       },
+      can_view_project_locations: {
+        description: 'Access to the map view of project locations',
+        administrative: false,
```
martha, line 559: The metadata (category, subcategory, administrative) I added here just matches `can_view_client_locations`

app/views/projects/show.haml
```diff
@@ -49,6 +49,10 @@
           %tr
             %th Project Contacts
             %td= link_to 'View/Edit', project_contacts_path(@project)
+        - if RailsDrivers.loaded.include?(:client_location_history)
+          %tr
+            %th Client Locations
```
martha, line 54: Maybe "Location Map" to match language of the location page's title?

drivers/client_location_history/app/models/client_location_history/location.rb
```diff
@@ -26,32 +28,31 @@ def as_point
...
 
-    def label
-      [
-        "Seen on: #{located_on}",
-        "by #{collected_by}",
-      ]
-    end
-
-    def as_marker
+    def as_marker(user = nil, label_attributes = [:seen_on, :collected_by])
```
martha, line 31: This is a minor refactor where, instead of accumulating more `as_marker_with_...` methods, I wanted to consolidate into 1 method that can return a differently formatted label depending on the options passed.

drivers/client_location_history/config/routes.rb
```diff
@@ -9,6 +9,9 @@
     resources :clients, only: [:none] do
       get :map, on: :member
     end
+    resources :projects, only: [:none] do
```
martha, line 12: This is a blind copy/paste, I don't really know what `only: [:none]` is doing.

drivers/client_location_history/app/controllers/client_location_history/projects_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+      @bounds = ClientLocationHistory::Location.bounds(@locations)
+      @options = {
+        bounds: @bounds,
+        cluster: true,
+        marker_color: ClientLocationHistory::Location::MARKER_COLOR,
+      }
+    end
+
+    private def set_project
+      @project = project_scope.includes(:enrollment_location_histories).find(params[:id].to_i)
```
martha, line 25: I tested this out locally with a user who shouldn't have permission
- Doesn't have any permissions on this project: works fine, they are unable to view
- Has view permissions on this project, but not `can_view_project_locations` - can see even though should not. in the debugger I can see that `using_acls?` is false for this user, but I don't know how to turn that on.

app/views/projects/show.haml
martha: It seems like maybe this is no longer necessary, with the location map moved to its own page? However, see https://github.com/greenriver/hmis-warehouse/pull/4696#discussion_r1763250524

eanders: I agree, if it's on its own page, this should require access to the project and access to the new permission.  Ideally, we'd correlate those two to make sure you have access to _this_ project's locations (for ACL users) but probably fine to require each of them for now.

drivers/client_location_history/config/routes.rb
eanders: `only: [:none]` is there because we wanted to reuse the resources defined in the main routes file, but not stomp on them (none could be anything that isn't in the main routes file.)  If you don't have sub pages defined, you wouldn't need this.  I think you want something like:
```
resources :projects, only: [:none] do
  get :map, on: :member
end
```

drivers/client_location_history/app/controllers/client_location_history/projects_controller.rb
eanders: In the case where you aren't using ACLs, this should return data if you have access to the project through any group and any role with `can_view_project_locations`.  If you are using ACLs (set on the user edit screen), then you should need an access control that has a role with `can_view_project_locations` AND a collection containing the project.

martha: Ok, thanks, this was a very helpful introduction to the warehouse legacy permissions! I tested and think everything looks good now. I re-added the `can?` permission on project that we discussed earlier, in order to granularly determine whether the link to the project map should be shown.

Tested:

Not using ACLs
- User who has access to the project through any group, and has a role granting both `can_view_projects` and `can_view_project_locations`
  - Can see link on project page
  - Can see map
- User who has access to the project but has a role only granting `can_view_projects`, not `can_view_project_locations`
  - Cannot see link on project page
  - If they try to navigate there directly, they get "Sorry you are not authorized to do that"

Using ACLs
- User who has `can_view_project` but not `can_view_project_locations` on this project, even though they may have `can_view_project_locations` on another project:
  - Can't see the link on the project page
  - If they navigate directly to the route, they get `ActiveRecord::RecordNotFound`
- User who has `can_view_project` and `can_view_project_locations`:
  - Can see link on project page
  - Can see map
- User who has `can_view_project_locations` and not `can_view_project` (This should not happen)
  - Cannot see project page at all
  - If they navigate directly to the route, they get `ActiveRecord::RecordNotFound`

eanders: That sounds perfect.  Thanks @martha !

eanders (approved):

gigxz (approved):
nice!

gigxz:
@martha maybe we should target this against release-134 instead of pit-external-forms. That branch was meant to isolate out any External Forms behavior that might affect Prevention Screening (and review). This seems relatively isolated and could be QA'd independently on 134

martha:
Hrm @gigxz sorry for the mess, it looks like I merged `pit-external-forms` into this branch a couple times during development, so it's a bit of a pain to tease out which commits are part of this isolated change and could be retargeted to 134. I'd be happy to do that though if you think it's worth while. Maybe the simplest thing is to open a new PR with the squashed diff?

gigxz:
Oh yeah, no worries. I think it's worth merging to 134 so we can proceed with QA for it. (The staging external forms env is still pending, and that PIT branch is already heavy).

Yes I think a new PR, you should be able to get the diff into a single commit with: `git checkout release-34; git checkout martha/6580-pit-map app db/migrate drivers/client_location_history` (then migrate & commit to a new branch)

gigxz:
Oh wait, sorry. I just realized that the `clh_locations` table changes are on pit-external-forms, so this should go into pit-external-forms too, since it expects `located_on` to exist.
https://github.com/greenriver/hmis-warehouse/pull/4679/files

gigxz (approved):

drivers/client_location_history/app/controllers/client_location_history/projects_controller.rb
martha: Closing the circle here for posterity: There was a bug with the case `User who has access to the project but has a role only granting can_view_projects, not can_view_project_locations`. I was fooled by local caching when I tested it and assumed it was working. See PR fix: https://github.com/greenriver/hmis-warehouse/pull/4787

PR #4697, opened 2024-09-12
Handle duplicate Bed Night assignment gracefully

gigxz:

## Description

Issue: https://github.com/open-path/Green-River/issues/6493

If someone tries to assign a bed night to an Enrollment that already has a bed night on that date, we can safely ignore it because the desired state (bed night on X date) is already present. This makes handling of these 2 scenarios more graceful:
1. Apollo retried the request, or
2. Provider was looking at a "stale" screen that indicated that the client was unassigned, when indeed they were already assigned (by another user or by the same user in another window). We can safely ignore and return success in that case (without creating the dup).

**Repro instructions**
* Open bed nights page in two windows for the same project
* In one window, click assign to a client
* Wait
* In the other window, click assign to the same client
  * Previously result: unexplained error
  * New result: appears successful

## Type of change
- [x] Bug fix


drivers/hmis/app/graphql/mutations/bulk_assign_service.rb
```diff
@@ -15,7 +15,7 @@ def resolve(input:)
       access_denied! unless project
       access_denied! unless current_permission?(permission: :can_edit_enrollments, entity: project)
 
-      clients = Hmis::Hud::Client.viewable_by(current_user).where(id: input.client_ids).preload(:enrollments)
+      clients = Hmis::Hud::Client.viewable_by(current_user).where(id: input.client_ids)
```
gigxz, line 18: this should be unecessary since we use load_open_enrollment_for_client

ttoomey (approved):
Looks good, thanks!

PR #4706, opened 2024-09-16
Fix conditional validation on external forms

gigxz:

## Description

Don't validate any fields that are hidden due to conditional visibility. I previously tried to fix this by relying on disabled attribute, but it doesn't catch cases where an input is hidden because it belongs to a group that is hidden.

Bug reproduction:
* Open the PIT in external forms view locally
* Select "interview"
* Enter a couple fields and submit (Expect validation error)
* Switch to "observational"
* Enter all visibly required fields and submit
  * EXPECTATION: successful submit
  * REALITY: fails to submit without visible error, because some of the interview-only fields were required

## Type of change
- [x] Bug fix


martha (approved):
Great, thanks for the fix 🙏

PR #4713, opened 2024-09-18
Add external forms bulk review mutation

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6598
- Add a bulk external forms review mutation and spec tests
- Make external form "reviewed" checkbox disabled if the form has already been reviewed
- ~Fix bug in Geolocation Processor to make it more flexible to bad/nil input~ Edit: Separated this out into https://github.com/greenriver/hmis-warehouse/pull/4716
- Move spec factory for form definition that updates enrollment/client into a shared location

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/lib/form_data/static/external_form_submission_review.json
```diff
@@ -15,7 +15,16 @@
       "text": "Mark reviewed",
       "mapping": {
         "field_name": "reviewed"
-      }
+      },
+      "enable_when": [
+        {
+          "local_constant": "$reviewed",
```
martha, line 21: Still wrapping my head around the order of ops here (if the local constant is even the right approach) so that everything is backwards compatible. If we merge this PR first and the frontend changes after, the reviewed checkbox will be disabled for every submission while that local constant is not yet present.

Maybe we need to merge 3 separate PRs:
1. to hmis-frontend release branch, adding this local constant to the context https://github.com/greenriver/hmis-frontend/pull/932
2. merge this PR to hmis-warehouse pit-external-forms branch
3. to hmis-frontend pit-external-forms branch, with schema updates and UX for bulk review https://github.com/greenriver/hmis-frontend/pull/933

gigxz: This does seem to work, but it also just feels like the "static form" paradigm just doesn't align super neatly with what we want to do. We already have some custom log for setting the [initial values](https://github.com/greenriver/hmis-frontend/blob/6d7f14fe7a862dfa30671b30b7b2e3505d3260bc/src/modules/projects/components/tables/ProjectExternalSubmissionsTable.tsx#L124-L130) on the form, which is somewhat non-standard, and now we also want to hide or disabling the "reviewed" option based on whether it is has been reviewed.

Imo we should only use a "static form" when it simplifies development. I don't _think_ that's really the case here, so maybe we should just switch the frontend to use a custom form. I think that would simplify things and make it easier to maintain. (It would also let us improve the UI around the "mark reviewed" vs "mark spam" later on, for example making them buttons..)

What do you think?


replacing static form dialog with something like:
```
<CommonDialog>
  <DialogTitle></DialogTitle>
  <DialogContent>
    <DynamicView />
    <TextInput label='Reviewer Notes' multiline minRows={2} />
    <LabeledCheckbox label='Mark as Reviewed' />
    <LabeledCheckbox label='Mark as Spam' />
  </DialogContent>
  <DialogActions></DialogActions>
</CommonDialog>
```

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
```diff
@@ -0,0 +1,55 @@
...
+
+    def resolve(external_submission_ids:)
+      submissions = HmisExternalApis::ExternalForms::FormSubmission.where(id: external_submission_ids)
+      definitions = Hmis::Form::Definition.where(id: submissions.pluck(:definition_id))
+
+      # Submissions could come from different definitions (versions), but should be all the same form identifier
+      identifiers = definitions.pluck(:identifier).uniq
+      error_out('Cannot bulk process submissions from multiple form identifiers: ' + identifiers.inspect) unless identifiers.one?
+
+      project = Hmis::Form::Instance.where(definition: definitions).uniq.sole.entity # Expect only 1 instance/project
```
gigxz: `project = submissions.first.parent_project # can check first because we know they all have the same form identifier`
maybe? I ran into an issue because I had other deactivated rules. This approach ensures that the logic matches

drivers/hmis/lib/form_data/static/external_form_submission_review.json
martha: Sure I think that makes sense. In that case, maybe we should also get rid of the static form dialog's `beforeFormComponent` argument, which is only used here.

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
```diff
@@ -0,0 +1,55 @@
...
+        submissions.each do |record|
+          record.status = 'reviewed'
+          record.run_form_processor(current_user, project: project)
+
+          if record.enrollment&.new_record?
+            error_out(record.enrollment.client.errors.full_messages) unless record.enrollment.client.valid?
+            error_out(record.enrollment.errors.full_messages) unless record.enrollment.valid?
+
+            record.enrollment.client.save!
+            project.should_auto_enter? ? record.enrollment.save_and_auto_enter! : record.enrollment.save_in_progress!
```
gigxz: It doesn't look like `project.should_auto_enter?` caches itself at all, so we could pull it out of the loop to avoid running `detect_best_config_for_project` n times

drivers/hmis/lib/form_data/static/external_form_submission_review.json
martha: In terms of timing, should we make this change in the frontend against the release branch (not the pit branch)? That way we can also deprecate the external form static form role enum on the release branch, and then later, either as part of the pit branch or on another release, we can remove it entirely.

gigxz: That sounds reasonable to me!

gigxz (approved):

PR #4718, opened 2024-09-18
Enforce non-null on some columns

eanders:

## Description

This enforces non-null-ness of the following columns at the database level:
```
  [
      ['Enrollment', 'PersonalID'],
      ['Enrollment', 'EnrollmentID'],
      ['Enrollment', 'EntryDate'],
      ['Enrollment', 'data_source_id'],
      ['Exit', 'ExitID'],
      ['Exit', 'ExitDate'],
      ['Client', 'PersonalID'],
      ['Client', 'data_source_id'],
    ]
```

## Type of change


dtgreiner (approved):

gigxz (commented):
Yay! Just 2 questions:

* Should be deployed separately, so it can be more easily rolled back independently if we encounter any issues? If there are other changes in release-134 that are not reversible, it could be a pain. If you feel confident putting it in 134, I'm also OK with that.
* When we checked that these fields are all present in environments (I think we checked with a job?), did we check records with `DateDeleted` too? If you are relying on my word for the HMIS envs, I might want to double check that.

gigxz (approved):

eanders:
@gigxz 
1. I like the idea of pulling this in as the only change in `release-135` and then maybe deploying it off cycle.
2. We didn't thoroughly check null values for these, but my understanding is that implementing it this way (two migrations with checks and validations) should fail gracefully if it hits any.  We'd have to go fix them and then re-run the migrations.
3. There are still some test failures I need to fix before we can deploy this.

gigxz:
ok, sounds good to me! let me know if you need a hand with any hmis tests

PR #4720, opened 2024-09-18
Resolve key/value summary fields on form submission type

gigxz:

## Description

issue: https://github.com/open-path/Green-River/issues/6622
frontend pr: https://github.com/greenriver/hmis-frontend/pull/936

This supports showing summary fields on the table view (see screenshots on frontend pr). This is not the nicest approach, but it works for our current need which is to show some identifying details during PIT review. There is no option for specifying the sort order, they are always resolved alphabetically.

## Type of change

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
@@ -40,22 +40,26 @@ module Types
 
     # "Summary" fields are a subset of the form's fields that are displayed on the external forms review table.
     # Includes Client first/last name plus any CDEDs where show_in_summary is true.
+    Field = Struct.new(:id, :key, :value, keyword_init: true)
     def summary_fields
       cdeds_by_key = load_ar_association(definition, :custom_data_element_definitions).
         select(&:show_in_summary).
         index_by(&:key).stringify_keys
 
-      cded_keys = cdeds_by_key.keys
+      cded_keys = cdeds_by_key.keys.to_set
 
       object.form_values.map do |key, value|
+        value = sanitized_value(value)
+        next unless value
+
         case key.to_s
         when 'Client.firstName'
-          OpenStruct.new(id: 'first', key: 'First Name', value: value)
+          Field.new(id: 'first', key: 'First Name', value: value)
         when 'Client.lastName'
-          OpenStruct.new(id: 'last', key: 'Last Name', value: value)
-        when *cded_keys
+          Field.new(id: 'last', key: 'Last Name', value: value)
+        when cded_keys
           cded = cdeds_by_key[key]
-          OpenStruct.new(id: cded.id, key: cded.label, value: value)
+          Field.new(id: cded.id, key: cded.label, value: value)
         end
       end.compact.sort_by(&:key)
     end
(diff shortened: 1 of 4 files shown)


drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -34,5 +37,34 @@ def values
...
+        when 'Client.firstName'
+          OpenStruct.new(id: 'first', key: 'First Name', value: value)
+        when 'Client.lastName'
+          OpenStruct.new(id: 'last', key: 'Last Name', value: value)
+        when *cded_keys
+          cded = cdeds_by_key[key]
+          OpenStruct.new(id: cded.id, key: cded.label, value: value)
+        end
+      end.compact.sort_by(&:key)
+    end
```
ttoomey: This is fine. There's probably so few cdeds here that it doesn't matter but some small optimizations could be made:
- a single struct definition should be more memory efficient than calling OpenStruct in a loop
- use a set for cded_keys should make the loop a little faster 

```suggestion
    Field = Struct.new(:id, :key, :value, keyword_init: true)
    def summary_fields
      cdeds_by_key = load_ar_association(definition, :custom_data_element_definitions).
        select(&:show_in_summary).
        index_by(&:key).stringify_keys

      cded_keys = cdeds_by_key.keys.to_set

      object.form_values.map do |key, value|
        case key.to_s
        when 'Client.firstName'
          Field.new(id: 'first', key: 'First Name', value: value)
        when 'Client.lastName'
          Field.new(id: 'last', key: 'Last Name', value: value)
        when cded_keys
          cded = cdeds_by_key[key]
          Field.new(id: cded.id, key: cded.label, value: value)
        end
      end.compact.sort_by(&:key)
    end
```

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
```diff
@@ -34,5 +37,34 @@ def values
...
+    # "Summary" fields are a subset of the form's fields that are displayed on the external forms review table.
+    # Includes Client first/last name plus any CDEDs where show_in_summary is true.
+    def summary_fields
+      cdeds_by_key = load_ar_association(definition, :custom_data_element_definitions).
+        select(&:show_in_summary).
+        index_by(&:key).stringify_keys
+
+      cded_keys = cdeds_by_key.keys
+
+      object.form_values.map do |key, value|
```
ttoomey, line 51: value is raw and could be anything here right? Maybe sanitize it a little?
```ruby
  value = sanitized_value(value)
  next unless value

#...

def sanitized_value(value)
  value&.to_s&.strip&.truncate(100)&.gsub(/[[:cntrl:]]/, '').presence
end
```

ttoomey (approved):
looks good!

martha (approved):

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
gigxz: Thanks, updated!

drivers/hmis/app/graphql/types/hmis_schema/external_form_submission.rb
gigxz: Thanks that;'s good to know about Open Structs, updated to match your suggestion

PR #4730, opened 2024-09-20
Add mutation for refreshing external forms

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6621
This PR adds a mutation to refresh external form submissions.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
```diff
@@ -0,0 +1,20 @@
...
+module Mutations
+  class RefreshExternalSubmissions < CleanBaseMutation
+    field :success, Boolean, null: false
+
+    def resolve
+      handlers = ['HmisExternalApis::ConsumeExternalFormSubmissionsJob']
+      return if Delayed::Job.queued?(handlers) || Delayed::Job.running?(handlers)
+
+      queue = ENV.fetch('DJ_SHORT_QUEUE_NAME', :short_running)
+      HmisExternalApis::ConsumeExternalFormSubmissionsJob.delay(priority: 12, queue: queue)
```
martha: What should priority be here? I just copied this from the service history queuer but maybe this should be higher priority?

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
```diff
@@ -0,0 +1,20 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Mutations
```
martha, line 6: Should this indeed be a mutation (because it is, sorta kinda, mutating data on the backend) or is it more like a query?

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
```diff
@@ -0,0 +1,20 @@
...
+    field :success, Boolean, null: false
+
+    def resolve
+      handlers = ['HmisExternalApis::ConsumeExternalFormSubmissionsJob']
+      return if Delayed::Job.queued?(handlers) || Delayed::Job.running?(handlers)
+
+      queue = ENV.fetch('DJ_SHORT_QUEUE_NAME', :short_running)
+      HmisExternalApis::ConsumeExternalFormSubmissionsJob.delay(priority: 12, queue: queue)
+
+      { success: true }
```
martha, line 17: I think the ux we’re proposing is that a user hits ‘refresh’, the button spins for a second, then that query resolves, but no new records are added to the table, because a delayed job was kicked off. Then, a few seconds/minutes later, the user would need to refresh the page to actually see the new records.

I'm wondering whether this query should *also* accept a form definition ID arg and return the latest results? That way at least the user would be able to hit refresh over and over and eventually see some new data. Maybe I misunderstood the plan for the UI of this feature, though - let's discuss more.

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
martha: Ah hm maybe priority should be nil/not passed, if priority is an alternative to queue as described here: https://github.com/collectiveidea/delayed_job?tab=readme-ov-file#parameters

gigxz: I think it should be high priority (`priority: 0`) since it was a user-initiated action

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
gigxz: Yeah I think mutation since it's not returning any data. It is modifying data on the server by queuing up the fetch job.

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
gigxz: Another option would be for the frontend to invalidate the cached submissions when this mutation returns successfully, to trigger a re-fetch of the table. That way, if the job _did_ get queued and executed immediately, they would see those results. It's still not ideal but might be an easy way to improve the experience slightly.

gigxz: Sounds like with the long-running queue it would take a few minutes for it to get queued, so it probably _won't_ be ready immediately. That said, if they click the button over and over, eventually the cache-clearing will yield results. So I think we should go with that.

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
```diff
@@ -0,0 +1,20 @@
...
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Mutations
+  class RefreshExternalSubmissions < CleanBaseMutation
+    field :success, Boolean, null: false
+
+    def resolve
+      handlers = ['HmisExternalApis::ConsumeExternalFormSubmissionsJob']
+      return if Delayed::Job.queued?(handlers) || Delayed::Job.running?(handlers)
```
gigxz: This should probably return ` { success: true }`?

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
martha: It looks like we also have negative numbers for even-higher priority https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/config/initializers/delayed_job.rb#L9-L11

app/models/aws_s3.rb
```diff
@@ -87,7 +87,7 @@ def fetch_key_list(prefix: '')
     @bucket.objects(prefix: prefix).sort_by(&:last_modified).map(&:key)
   end
 
-  def list_objects(max_keys = 1_000, prefix: '')
+  def list_objects(max_keys = 10_000, prefix: '')
```
gigxz: I don't think we should change the default for everyone, that might have some wide-ranging impacts that we are unaware of. We can pass the `max_keys` argument to list_objects from ConsumeExternalFormSubmissionsJob instead.

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
```diff
@@ -0,0 +1,20 @@
...
+module Mutations
+  class RefreshExternalSubmissions < CleanBaseMutation
+    field :success, Boolean, null: false
+
+    def resolve
+      handlers = ['HmisExternalApis::ConsumeExternalFormSubmissionsJob']
+      return { success: true } if Delayed::Job.queued?(handlers) || Delayed::Job.running?(handlers)
+
+      queue = ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
+      HmisExternalApis::ConsumeExternalFormSubmissionsJob.delay(priority: 0, queue: queue)
```
martha: Still to do: Modify the s3 client with the behavior we agreed on (https://github.com/open-path/Green-River/issues/6621#issuecomment-2368996651). I'd like to do this in a separate PR if that's ok.

gigxz (approved):
LGTM after isolating the max_keys change the the external forms job

app/models/aws_s3.rb
martha: Oh, thanks good catch. It looks like the s3 api won't allow a max keys of 10,000 anyway.

gigxz: ahh k good call to revert

gigxz (approved):

drivers/hmis/app/graphql/mutations/refresh_external_submissions.rb
martha: Edit: This is split out into a separate ticket https://github.com/open-path/Green-River/issues/6727

PR #4734, opened 2024-09-23
Cohorts Selectable Project Groups

dtgreiner:

## Description

1. Add paper trail to Cohorts.
2. Fix bug where a user has access to edit a cohort but does not have access to view the selected project group tied to the cohor automation. The cohort edit page now shows the selected project group but disables the selection in this scenario. Previously it would clear out the selected project group on save if the editing user did not have access to view the selected project group.

## Type of change
- [X] Bug fix

### DIFF
--- a/app/controllers/cohorts_controller.rb
+++ b/app/controllers/cohorts_controller.rb
@@ -164,8 +164,8 @@ class CohortsController < ApplicationController
     cohort_options = cohort_options.except(:name) if @cohort.system_cohort
 
     # checks to see if user can see the project group OR if the project group is on the cohort pre-save.
-    user_can_view_new_project_group = cohort_options['project_group_id'].blank? || GrdaWarehouse::ProjectGroup.viewable_by(current_user).exists?(cohort_options['project_group_id'])
-    project_group_on_cohort_pre_save = @cohort.project_group_id == cohort_options['project_group_id'].to_i
+    user_can_view_new_project_group = cohort_options[:project_group_id].blank? || GrdaWarehouse::ProjectGroup.viewable_by(current_user).exists?(cohort_options[:project_group_id])
+    project_group_on_cohort_pre_save = @cohort.project_group_id == cohort_options[:project_group_id].to_i
     cohort_options = cohort_options.except(:project_group_id) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
 
     @cohort.update(cohort_options)
--- a/app/controllers/cohorts_controller.rb
+++ b/app/controllers/cohorts_controller.rb
@@ -166,7 +166,7 @@ class CohortsController < ApplicationController
     # checks to see if user can see the project group OR if the project group is on the cohort pre-save.
     user_can_view_new_project_group = cohort_options['project_group_id'].blank? || GrdaWarehouse::ProjectGroup.viewable_by(current_user).exists?(cohort_options['project_group_id'])
     project_group_on_cohort_pre_save = @cohort.project_group_id == cohort_options['project_group_id'].to_i
-    cohort_options = cohort_options.except(:project_group) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
+    cohort_options = cohort_options.except(:project_group_id) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
 
     @cohort.update(cohort_options)
 
--- a/app/controllers/cohorts_controller.rb
+++ b/app/controllers/cohorts_controller.rb
@@ -163,6 +163,11 @@ class CohortsController < ApplicationController
     # END_ACL
     cohort_options = cohort_options.except(:name) if @cohort.system_cohort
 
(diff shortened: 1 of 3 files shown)


app/views/cohorts/edit.haml
```diff
@@ -37,7 +37,7 @@
       = f.input :viewer_ids, collection: @users, as: :select_two, selected: @viewer_ids, input_html: {multiple: :multiple, style: 'width: 100%;'}, label: 'Grant read-only-level access to users', hint: 'Specified user will have access to view, but not change clients on this cohort.'
       %h3 Automation
       .well
-        = f.input :project_group_id, collection: GrdaWarehouse::ProjectGroup.options_for_select(user: current_user), as: :select_two, include_blank: 'Manually controlled (no project group)', hint: 'If you select a project group, this cohort will always contain clients with an open enrollment at a project contained in the group.  You will not be able to add and remove clients manually.'
+        = f.input :project_group_id, disabled: !@cohort.selected_project_group_viewable_by(current_user), collection: @cohort.project_group_options_for_select(current_user), as: :select_two, include_blank: 'Manually controlled (no project group)', hint: 'If you select a project group, this cohort will always contain clients with an open enrollment at a project contained in the group.  You will not be able to add and remove clients manually.'
```
eanders: @dtgreiner in your digging, do you know if the controller checks visibility of the project group on update?  I'm tempted to leave this editable, but ensure that the user can always safely submit the existing project group.  That way they can change the project group if necessary, but it won't remove it if they don't have access.  I'd expect something in the controller that checks to see if they can see the project group OR if the project group is on the cohort pre-save.

dtgreiner: Nothing popping out about project_groups pre-save but it looks like adding it in would only be a couple lines to update. If I'm reading this right, we want the following:

1. show the current selection even if the current user doesn't have access to view the group
2. leave the selection as editable
3. validate selection pre-save on the controller to ensure the new selection is either Viewable by the current user OR existed as the project group prior to save

This would let the user change the project group or remove it even if they don't have access to view the current selection, and it would prevent overwriting the group inadvertently like this scenario.

dtgreiner: @eanders I just pushed a new commit that should address this

eanders: Thank you, and yes, that's the spec I was shooting for.

app/controllers/cohorts_controller.rb
```diff
@@ -163,6 +163,11 @@ def update
     # END_ACL
     cohort_options = cohort_options.except(:name) if @cohort.system_cohort
 
+    # checks to see if user can see the project group OR if the project group is on the cohort pre-save.
+    user_can_view_new_project_group = cohort_options['project_group_id'].blank? || GrdaWarehouse::ProjectGroup.viewable_by(current_user).exists?(cohort_options['project_group_id'])
+    project_group_on_cohort_pre_save = @cohort.project_group_id == cohort_options['project_group_id'].to_i
+    cohort_options = cohort_options.except(:project_group) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
```
eanders: Should this be?
```suggestion
    cohort_options = cohort_options.except(:project_group_id) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
```

app/controllers/cohorts_controller.rb
```diff
@@ -166,7 +166,7 @@ def update
     # checks to see if user can see the project group OR if the project group is on the cohort pre-save.
     user_can_view_new_project_group = cohort_options['project_group_id'].blank? || GrdaWarehouse::ProjectGroup.viewable_by(current_user).exists?(cohort_options['project_group_id'])
     project_group_on_cohort_pre_save = @cohort.project_group_id == cohort_options['project_group_id'].to_i
-    cohort_options = cohort_options.except(:project_group) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
+    cohort_options = cohort_options.except(:project_group_id) unless user_can_view_new_project_group || project_group_on_cohort_pre_save
```
eanders, line 169: sorry, this is probably fine, but should that be a string version to match the above two lines?
```
cohort_options['project_group_id']
```
It's probably fine as a symbol if this is a parameter, but if that's hash, it would need to be `with_indifferent_access` for the symbol and string to both work.
```
cohort_options.except('project_group_id')
```

dtgreiner: both the symbol and string appear to be working. I'll push another change to set it as a string to stay consistent.

eanders: Thank you!

dtgreiner: looking more, the cohort_params are being referenced elsewhere as symbols. I'll update all of these calls to match the rest of the references as a symbol instead

eanders (approved):
Thanks @dtgreiner !

PR #4737, opened 2024-09-24
Fix identify duplicates performance issue

gigxz:

## Description

Quick fix for https://github.com/open-path/Green-River/issues/6740 by removing expensive code from the initialize fn.

Introduced in https://github.com/greenriver/hmis-warehouse/pull/4553/files#r1773347288

A better solution would probably be to changes this class to a Job, and queue it normally. Right now maybe only HMIS code queues it on the long running queue.

Add `private` in front of methods that are not used outside the class

## Type of change
- [x] Bug fix

### DIFF
--- a/app/models/grda_warehouse/tasks/identify_duplicates.rb
+++ b/app/models/grda_warehouse/tasks/identify_duplicates.rb
@@ -12,9 +12,6 @@ module GrdaWarehouse::Tasks
     def initialize(run_post_processing: true)
       setup_notifier('IdentifyDuplicates')
       @run_post_processing = run_post_processing
-      super()
-      build_source_lookups
-      build_destination_lookups
     end
 
     def run!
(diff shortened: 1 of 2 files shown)


app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -12,9 +12,6 @@ class IdentifyDuplicates
     def initialize(run_post_processing: true)
       setup_notifier('IdentifyDuplicates')
       @run_post_processing = run_post_processing
-      super()
```
gigxz, line 15: Unless there is some magic I'm unaware of, there is no superclass so this is pointless. Am I missing something?

ttoomey (approved):
Looks good, thank you! So sorry for the trouble, I should have looked more carefully at how this was being invoked.

gigxz:
No worries! I don't like the non-standard queueing we do, I think it would probably be better if this was a BaseJob, so it's more obvious how it's used. If that seems like a good idea to others, we can make the change in a future release. (I think it might have a wider impact, impacting how these methods are invoked across the codebase)

gigxz:
Ah, turns out we already have a 7-year-old unused RunIdentifyDuplicatesJob... I'll make a PR against 135 to bring that back into commission

gigxz:
looks like there is a test failure, working on that

eanders (approved):
Thanks @gigxz !

PR #4743, opened 2024-09-24
Process client age range from external submission

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6731

This PR adds processing for the age range field from PIT submissions. If a precise DOB is provided, that should always take precedence.

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -50,6 +50,13 @@ def process(field, value)
       when 'veteran_status'
         # Veteran status is non-nullable. It should be saved as 99 even if hidden. (It's hidden for minors)
         { attribute_name => attribute_value || 99 }
+      when 'age_range'
+        # if precise DOB was already processed, don't overwrite it with vaguer age range
+        client.dob.present? ? {} : process_age_range(value)
```
gigxz: This check assumes that `dob` was processed before `age_range`, which might not be the case. I think instead we want to look for dob on `@hud_values`.

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -152,6 +159,19 @@ def self.input_to_multi_fields(input_field, enum_map, not_collected_key, none_fi
...
 
+    private def process_age_range(value)
+      return { dob_data_quality: 99 } if value == "Doesn't know / Prefers not to answer"
+
+      dob_range = HudUtility2024.age_range[value]
+      raise ArgumentError, "Unknown value for age range: #{value}" unless dob_range
+
+      {
+        # set to a date in the middle of the range
+        dob: Date.current - ((dob_range.begin + dob_range.end) / 2).round.years,
```
gigxz: I think it might be a good idea to set this to the start of the year.
For a provider viewing the data in HMIS, `1990-01-01 (Quality: approximate)` makes more sense than `1990-09-25 (Quality: approximate)` eg its a little more obvious that it is indeed approximate and just based on age.

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -152,6 +159,19 @@ def self.input_to_multi_fields(input_field, enum_map, not_collected_key, none_fi
       name_attributes.merge(client_attributes)
     end
 
+    private def process_age_range(value)
+      return { dob_data_quality: 99 } if value == "Doesn't know / Prefers not to answer"
```
gigxz: Hrmmm, a predicament. Don't like hard-coding this since it comes from a customizable form, but we don't want to ignore unrecognized input either (since it could lead to silent data loss).

Here is an idea for some logic that might be more future-proof. What do you think? (I'm not necessarily proposing this specific case / syntax, just the logic itself)

```ruby
case value
# value matches a known age range, so process it onto dob with low DQ
when *HudUtility2024.age_range.keys
   # process onto dob
   { dob: approximate_dob(value), dob_data_quality: 2 }
# value matches a known missing data reason, so store that. (not currently expected but future-proofing for desired pick lists)
when *HudUtility2024.dob_data_quality_options.values_at(8,9,99)
   { dob_data_quality:  HudUtility2024.dob_data_quality(value, true) }
# value string-matches a missing data reason (PIT form is an example), so don't raise and store 99
when /doesn't know|prefers not to answer|not collected/i # i'm sure this regex isnt right
   { dob_data_quality: 99 }
# this might be an age range that we don't support processing, so we raise
else
  raise "unexpected value"
end
```

martha: The regex you put looks fine actually :)

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
martha: Sounds good, in that case I can remove the case added for 'dob'

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -303,6 +303,100 @@
...
+        it 'should process the age range onto DOB with low data quality' do
+          expect do
+            response, result = post_graphql(input) { mutation }
+            expect(response.status).to eq(200), result.inspect
+            expect(result.dig('data', 'updateExternalFormSubmission', 'externalFormSubmission', 'status')).to eq('reviewed')
+          end.to change(Hmis::Hud::Client, :count).by(1).
+            and change(Hmis::Hud::Enrollment, :count).by(1)
+
+          submission.reload
+          expect(submission.enrollment.client.dob).to be_between(Date.today - 24.years, Date.today - 18.years)
```
gigxz: To reduce test flakiness, can define a variable `let(:today) { Date.current }` and reference it throughout the tests. Then there aren't failures at midnight. Also reminder on `current` instead of `today`

lib/util/hud_utility_2024.rb
```diff
@@ -762,4 +762,23 @@ def assessment_name_by_data_collection_stage
       6 => 'Post-Exit Assessment',
     }.freeze
   end
+
+  # Utility for defining age range logic in one place.
+  # These can overlap, not all are used in every form dropdown/filter
+  def age_range
```
gigxz, line 768: Flagging this to @eanders for awareness. We couldn't find an existing mapping like this. These values were defined in an HMIS GQL Enum that we used for the Age Range table filter on client/enrollment tables in HMIS. Now (some of them) are also being used for PIT age categories, and this PR introduces logic for processing age range responses into low-data-quality DOB values.

drivers/hmis/app/graphql/types/hmis_schema/enums/age_range.rb
```diff
@@ -11,16 +11,20 @@ class HmisSchema::Enums::AgeRange < Types::BaseEnum
...
+      { code: 'Ages25to34', description: '25-34' },
+      { code: 'Ages35to44', description: '35-44' },
+      { code: 'Ages45to54', description: '45-54' },
+      { code: 'Ages55to61', description: '55-61' },
+      { code: 'Ages55to64', description: '55-64' },
+      { code: 'Age62Plus', description: '62+' },
+      { code: 'Age65Plus', description: '65+' },
+    ].map do |option|
+      value option[:code], option[:description], value: HudUtility2024.age_range[option[:description]]
+    end
```
gigxz, line 28: This enum could be generated based on the map values, but I think it's better to keep it so as not to change the graphql api unnecessarily. So looks good 👍

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -50,6 +50,9 @@ def process(field, value)
       when 'veteran_status'
         # Veteran status is non-nullable. It should be saved as 99 even if hidden. (It's hidden for minors)
         { attribute_name => attribute_value || 99 }
+      when 'age_range'
+        # Prioritize exact DOB if it is provided
+        process_age_range(value) unless @hud_values.key?('Client.dob') && @hud_values['Client.dob'].present?
```
gigxz: I don't think the first `key?` condition is needed. This works fine: `{}['key'].present?`

gigxz (approved):
looks good to me!

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
martha: Thanks, sorry

lib/util/hud_utility_2024.rb
eanders: Thanks, I saw this.  I think this is probably fine.  Eventually we should build this out to support/share code with:
https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/app/models/concerns/filter/filter_scopes.rb#L84-L103
and
https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/app/models/filters/filter_base.rb#L907-L1008

Those are the two that I know we've been trying to standardize on.  There are some additional ranges specific to HUD reports here
https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/app/models/concerns/hud_reports/ages.rb#L51-L64
and some I don't have any memory of here:
https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/app/models/concerns/known_categories/age.rb#L10-L33

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
gigxz: all good, thanks!

PR #4749, opened 2024-09-25
Raise user-facing message on unit availability error

gigxz:

## Description

Sentry: https://green-river.sentry.io/issues/5704537232/?project=4503977346662400&referrer=issue-stream&statsPeriod=90d&stream_index=0

Fixes an issue where if the unit becomes unavailable between when a user selects it and when the user submits the Enrollment form, they would see a generic "an error occurred" message. With this, they will see the appropriate message about the unit no longer being available.

The big change here is error handling in FormProcessor. If we hit an ApiError we assume it should be user-facing, since we often use ApiError that way. It doesn't feel super clean, but I do think we want to allow some hook for the Form Processors themselves to raise user-facing messages at runtime. (Or, maybe it would be better if the processors can add Validation Errors during execution? That felt like a bigger lift, but might be worthwhile).

### How to reproduce
* Mimic prod-user-facing behavior is by setting `dev_or_test` to `false` in the `GraphqlController`.
* Seed AC forms
* Go to a Project
* Add some units
* Go to => Enrollments => Add Enrollment
* Manually comment out the `if` in assign_unit so an error gets raised (this is the easiest way to see the behavior, can also repro it as a race condition)
* Enroll a client, choosing a unit to assign them to. You will see an error toast.

## Type of change
- [x] Bug fix


gigxz:
marking as draft again. I want to try making this a validation error instead. part of the point is we shouldn't get sentry alerts for it, since it's a race/stale-data condition

PR #4757, opened 2024-09-26
Dg/va_history_dashboard-6646

dtgreiner:

## Description

Update th history page for the VA dashboard to only show projects available to the current user. The revoked chack on the files page was also remove to allow access to the homeless history. The files here are being limited based on visibility to the user.

_Note: This is using a project visibility method new to release-135._

## Type of change

### DIFF
--- a/drivers/client_access_control/app/models/client_access_control/client_history_month.rb
+++ b/drivers/client_access_control/app/models/client_access_control/client_history_month.rb
@@ -48,16 +48,19 @@ module ClientAccessControl
     end
 
     def available_projects(month:, year:, client:, user:)
-      @available_projects ||= ::GrdaWarehouse::Hud::Project.
-        joins(service_history_enrollments: :enrollment).
-        merge(::GrdaWarehouse::Hud::Enrollment.visible_to(user)).
-        merge(
-          ::GrdaWarehouse::ServiceHistoryEnrollment.open_between(
-            start_date: date_range_for(month: month, year: year).first,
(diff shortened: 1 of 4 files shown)


drivers/client_access_control/app/models/client_access_control/client_history_month.rb
```diff
@@ -281,7 +286,10 @@ def weeks_data(month:, year:, client:, user:)
             days: week,
           }
           projects = {}
-          enrollments(month: month, year: year, client: client, week: week, user: user).each do |she|
+          enrollments = enrollments(month: month, year: year, client: client, week: week, user: user)
+          # Limit the visible projects to just those available to the user for when the VA dashboard is being used.
+          enrollments = enrollments.select { |e| e.project.can?(user) } if ::GrdaWarehouse::Config.get(:client_dashboard).to_s == 'va'
```
eanders: Can you confirm the permission that `Project.can?` uses by default?  I think you want to pass some version of "can the user see this enrollment"  If you actually have an enrollment, it would probably be better to just use the enrollment arbiter.
I'm actually confused about why there needs to be a limit in either of these situations.  `available_projects` already merges `::GrdaWarehouse::Hud::Enrollment.visible_to(user)` which _should_ correctly apply the limits.  I'd be surprised if the call to `enrollments(..)` on line 289 didn't also enforce that somewhere in the chain.

dtgreiner: @eanders This should be addressed in the new commit

PR #4768, opened 2024-09-27
Process disability from external forms

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6733

This PR enables processing disability responses from the PIT external form onto Disability records. The form structure I'm working with in dev looks like this:

<details>
<summary>Disability questions</summary>

```
{
    "item": [
      {
        "text": "Substance use disorder",
        "type": "BOOLEAN",
        "link_id": "substance_use_disorder",
        "mapping": {
          "field_name": "substanceUseDisorder",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      },
      {
        "text": "Chronic health condition",
        "type": "BOOLEAN",
        "link_id": "chronic_health_condition",
        "mapping": {
          "field_name": "chronicHealthCondition",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      },
      {
        "text": "Mental health disorder",
        "type": "BOOLEAN",
        "link_id": "mental_health_disorder",
        "mapping": {
          "field_name": "mentalHealthDisorder",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      },
      {
        "text": "Physical disability",
        "type": "BOOLEAN",
        "link_id": "physical_disability",
        "mapping": {
          "field_name": "physicalDisability",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      },
      {
        "text": "Developmental disability",
        "type": "BOOLEAN",
        "link_id": "developmental_disability",
        "mapping": {
          "field_name": "developmentalDisability",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      },
      {
        "text": "HIV/AIDS",
        "type": "BOOLEAN",
        "link_id": "hiv_aids",
        "mapping": {
          "field_name": "hivAids",
          "record_type": "DISABILITY_GROUP"
        },
        "component": "CHECKBOX",
        "disabled_display": "HIDDEN"
      }
    ],
    "text": "Do you have any of the following conditions?",
    "type": "GROUP",
    "link_id": "do_you_have_any_of_the_following_conditi",
    "component": "INPUT_GROUP",
    "enable_when": [
      {
        "operator": "EQUAL",
        "question": "survey_type",
        "answer_code": "In-Person Interaction"
      }
    ],
    "enable_behavior": "ALL",
    "disabled_display": "HIDDEN"
  },
  {
    "text": "Do any of those conditions keep you from actively maintaining employment or living independently?",
    "type": "CHOICE",
    "link_id": "do_any_of_those_conditions_keep_you",
    "mapping": {
      "field_name": "disablingCondition",
      "record_type": "ENROLLMENT"
    },
    "enable_when": [
      {
        "operator": "EQUAL",
        "question": "substance_use_disorder",
        "answer_code": "1"
      },
      {
        "operator": "EQUAL",
        "question": "chronic_health_condition",
        "answer_code": "1"
      },
      {
        "operator": "EQUAL",
        "question": "mental_health_disorder",
        "answer_code": "1"
      },
      {
        "operator": "EQUAL",
        "question": "physical_disability",
        "answer_code": "1"
      },
      {
        "operator": "EQUAL",
        "question": "developmental_disability",
        "answer_code": "1"
      },
      {
        "operator": "EQUAL",
        "question": "hiv_aids",
        "answer_code": "1"
      }
    ],
    "warn_if_empty": false,
    "enable_behavior": "ANY",
    "disabled_display": "HIDDEN",
    "pick_list_options": [
      {
        "code": "NO",
        "label": "No"
      },
      {
        "code": "YES",
        "label": "Yes"
      },
      {
        "code": "DATA_NOT_COLLECTED",
        "label": "Doesn't know / Prefers not to answer"
      }
    ]
  }
],
```
</details>

To dos:
- [x] Indefinite and Impairs (on the disability record) doesn't get set. We should try to make the form processor collect it automatically as 99 if the response is yes. [Done but outstanding question about this]
- [x] Substance use disorder collects as 1 (true) but in the processor this maps to "Alcohol use disorder"
- [x] Overall disabling condition (on the enrollment record) should get set for hiv/aids and also developmental disability. We discussed doing this in a post-processing one off script if it is difficult to do in the form processor. However, I think it might be worth doing in the form processor if we expect this to come up again, which it seems like it reasonably might?

Questions:
- [x] The Disability records are getting processed with no data collection stage, is this correct? Should it be 1 (intake)?
- [x] Is there anywhere in the ui that the client or enrollment's disability records are all shown? Or is it (so far) only shown on the assessment where it was collected?

## Type of change
- [ ] Bug fix
- [ ] Dependency update


drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -222,15 +227,13 @@ def client_factory(create: true) # rubocop:disable Lint/UnusedMethodArgument
     end
   end
 
-  # Common HUD Assessment-related attributes
+  # Common HUD Assessment-related attributes. These always have enrollments so enrollment_factory *should* exist
   def common_attributes
     data_collection_stage = owner.data_collection_stage if owner.respond_to?(:data_collection_stage)
-    personal_id = owner.personal_id if owner.respond_to?(:personal_id)
-    information_date = owner.information_date if owner.respond_to?(:information_date)
+    personal_id = enrollment_factory(create: false)&.personal_id
```
martha, line 233: I tested the removal of `information_date` here with an intake. It looks ok to me, the disability records still collect information date from [here](https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/drivers/hmis/app/models/hmis/hud/processors/base.rb#L37). I think @gigxz you are right that this is legacy. But if you are nervous to remove it I can revert the change.

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
```diff
@@ -9,7 +9,11 @@ class DisabilityGroupProcessor < Base
     def process(field, value)
       if field_mapping.key?(field)
         disability_type, disability_field, enum_type = field_mapping[field]
-        disability_value = attribute_value_for_enum(enum_type, value)
+        disability_value = if enum_type.values.values.map(&:value).map(&:to_s).include?(value)
```
martha: Not sure what you will think of this, but this is the hacky workaround to the fact that the processor expects these values to come from enums, but in the new form they are booleans.

but, as mentioned above, it doesn't work for substance abuse disorder where 1 means "alcohol use disorder (only)" and not simply "yes".

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -221,7 +221,7 @@ window.addDependentGroup = function (conditions, targetSelector, enableBehavior
   // When *any* dependent item changes, this function will check all the conditions, and show/hide the target item accordingly.
   var onDependentItemChanged = function() {
     var evaluations = conditions.map(function (condition) {
-      var $el = $('[name="' + condition.input_name + '"]')
+      var $el = $('[name="' + condition.input_name + '"]:visible')
```
martha, line 224: I _think_ this is a bug due to the hidden input here: https://github.com/greenriver/hmis-warehouse/blob/f039ec719577065e8d3c7c0987b0ee3e3a131c40/drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_checkbox.haml#L1 

I'm not sure if there were any existing places in our external forms where we have enableWhen conditions based on checkboxes.

drivers/hmis/spec/requests/hmis/update_external_form_submission_spec.rb
```diff
@@ -458,6 +458,29 @@
...
+          expect do
+            response, result = post_graphql(input) { mutation }
+            expect(response.status).to eq(200), result.inspect
+            expect(result.dig('data', 'updateExternalFormSubmission', 'externalFormSubmission', 'status')).to eq('reviewed')
+          end.to change(Hmis::Hud::Client, :count).by(1).
+            and change(Hmis::Hud::Enrollment, :count).by(1).
+            and change(Hmis::Hud::Disability, :count).by(1)
+
+          submission.reload
+          expect(submission.enrollment.disabilities.sole.disability_type).to eq(6)
```
gigxz: this should probably assert on `disability_response` too?

drivers/hmis/app/models/hmis/form/form_processor.rb
gigxz: this looks like a good clean up to me, thanks

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
gigxz: Part of me thinks we should just make `attribute_value_for_enum` more forgiving (like below), so it can flexibly allow "real" values alongside enum values. Then we don't need to special-case it in each form processor any time we have an external form doing something like this. 
I _think_ this would be appropriate, because the form processor is already forgiving in the way it ingests CDEDs (eg bool can be `0`,`'0'`,`false`), so we have the concept already.
What do you think? 

```
attribute_value_for_enum(Types::HmisSchema::Enums::Hud::NoYesReasonsForMissingData, 'YES') => 1
attribute_value_for_enum(Types::HmisSchema::Enums::Hud::NoYesReasonsForMissingData, '1') => 1
attribute_value_for_enum(Types::HmisSchema::Enums::Hud::NoYesReasonsForMissingData, 1) => 1
```

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -222,15 +227,13 @@ def client_factory(create: true) # rubocop:disable Lint/UnusedMethodArgument
...
     {
       data_collection_stage: data_collection_stage,
```
gigxz, line 235: @martha to your other note, Data Collection Stage is required for Disability records, so we need to set it somehow. maybe form submissions can just have a method for dcs that returns 1 (intake).

gigxz:
>  Is there anywhere in the ui that the client or enrollment's disability records are all shown? Or is it (so far) only shown on the assessment where it was collected?

They are shown in the warehouse on the client dashboard

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
martha: @gigxz i think it sounds sensible to me that `attribute_value_for_enum` should be flexible, but I'm not sure exactly how it should be implemented; how would that method know whether an int or a string should be returned, other than by checking the values of the enum type provided to see if they are integers? I've taken a stab at that here but I think it's kind of ugly.

https://github.com/greenriver/hmis-warehouse/pull/4768/commits/47ea7fe9373e822d01c9d2ee2a8a90b5153bec3e

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
```diff
@@ -10,7 +10,19 @@ def process(field, value)
       if field_mapping.key?(field)
         disability_type, disability_field, enum_type = field_mapping[field]
         disability_value = attribute_value_for_enum(enum_type, value)
-        @processor.send(disability_type).assign_attributes(disability_field => disability_value)
+
+        attributes = { disability_field => disability_value }
+
+        if Hmis::Hud::Disability.positive_responses.include?(disability_value.to_i)
```
martha: Collecting `indefinite_and_impairs` like this is a change for the intake form's behavior. I'd be fine with reverting it and just collecting all of them as nil. It looks like the warehouse also does some processing around this to return a sensible value even if the database info is incomplete? https://github.com/greenriver/hmis-warehouse/blob/10de915680000021ecdd89df692f4cf3f22a6dd8/app/models/grda_warehouse/hud/disability.rb#L132-L139

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
martha: This also still wouldn't quite work for substance abuse disorder. Maybe the right path forward is to make 2 new accepted field mappings `alcoholUseDisorder` and `drugUseDisorder` that collect 2 different booleans, and then collect that onto the same disability record?

gigxz: A cleaner approach might be to create an alternate method like `value_for` but that accepts a key or value (eg `'1',1,'YES'`, and call that instead of value_for.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -106,7 +106,7 @@ def attribute_value_for_enum(enum_type, value)
     elsif is_array
       value.map { |val| enum_type&.value_for(val) || val }
     else
-      enum_type&.value_for(value) || value
+      enum_type&.flexible_value_for(value) || value
```
gigxz: I think to reduce impact, we can drop this now since we agreed on using the simpler radio buttons instead.. I'm sorry for all the back and forth @martha

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
gigxz: Yeah I don't think this is necessary, we already have the logic encoded in our frontend disability component for HMIS forms. For PIT it's ok to be null, since it's not used for pit count reporting.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz: Yes just confirming there were no usages before, thanks!

drivers/hmis_external_apis/app/models/hmis_external_apis/external_forms/form_submission.rb
```diff
@@ -76,6 +76,10 @@ def form_values
       raw_data.reject { |key, _| EXTRANEOUS_KEYS.include?(key) }
     end
 
+    def data_collection_stage
+      1 # use Intake for external forms
+    end
```
gigxz: Add a note about usage, since it's not obvious where this is called
```suggestion
    # Consider external form submissions to have "Project Entry" data collection stage, because they may generate new Enrollments. This method is used by the form processor to set DataCollectionStage on related records that are produced, such as Disability records.
    def data_collection_stage
      1
    end
```

drivers/hmis/app/models/hmis/hud/processors/disability_group_processor.rb
martha: Would you like to keep the special case around hiv/aids and developmental disability (not collected for pit but in the generic case) or trust the value coming from frontend now that we know we will ask that question (though not require it)?

gigxz: That logic makes me a little nervous, because it's the DisabilityGroup setting an attribute on a _different_ record (the Enrollment). Afaik we don't have anything like that. 2 thoughts:
1. Is this logic that should actually belong on the Enrollment Processor, since it's about setting an Enrollment field? e.g. EnrollmentProcessor checking for presence of `hud_values[DisabilityGroup.developmentalDisability]` or whatnot and setting Enrollment.disablingCondition based on those
2. Doesn't this logic introduce a dependency between the order of factories? I don't think we iterate through fields in a consistent order. Sometimes it might process Disability fields first, other times it might process Enrollment first.


I'm sure we could find reasonable answers to those issues, but they feel somewhat hairy. I don't think we have a compelling reason to add this complexity/effort. Instead, it could just be a point of training for the users to know when to select "yes" for the disabling condition question. (And on the HMIS side, we already made the choice to encapsulate this logic in `DisabilityTable.tsx`, for better or worse, the responsibility of generating a correct DisabilityCondition already kinda lives with the form itself).

martha: Ok, I'm happy with that! I'll remove this. I added a note to our tchc meeting notes about the training point you mentioned. I don't know if it is exactly the right place, but just didn't want to forget it entirely.

gigxz: Perfect, thank you

gigxz (approved):
lgtm,  thanks for your work on this!

PR #4769, opened 2024-09-29
individual rspec test performance improvement

ttoomey:

## Description

The goal is to reduce the time required to run an individual test. On my local env, this was around 25 seconds. With these changes, the time goes to around 3 seconds.

changes:
* enable spring and update binstubs. Shaves off almost all the startup time for running a test or opening the console
* add logic to selectively load fixtures/form defs. This shaves off considerable time when running an individual test
* explicitly disable spring in CI since it just adds overhead there
* add spring-watcher-listen (fs events rather than polling to detect change). This works for me but maybe not for everyone?

example:
```ruby
#untracked/simple_spec_2.rb 
require 'rails_helper'

RSpec.describe 'hello world test', type: :model do
  it "is okay" do
    puts 'testing here'
    expect(1).to eq(1)
  end
end
```
```
/app $ time rspec untracked/simple_spec_2.rb 
Running via Spring preloader in process 30502
testing here
.

Finished in 1.82 seconds (files took 0.45802 seconds to load)
1 example, 0 failures

real    0m 2.75s
user    0m 0.22s
sys     0m 0.03s
```

## Type of change
* developer efficiency 


config/spring.rb
```diff
@@ -1,6 +0,0 @@
-Spring.watch(
```
ttoomey, line 1: I don't think we need this if we are using fs events(?)

spec/rails_helper.rb
```diff
@@ -70,28 +70,37 @@
...
+    if example_file_paths.grep(%r{/drivers/hmis/}).any? # rubocop:disable Style/RegexpLiteral
+      ::HmisUtil::JsonForms.seed_all if ENV['ENABLE_HMIS_API'] == 'true'
     end
 
-    GrdaWarehouse::Utility.clear!
-    Delayed::Job.delete_all
-    GrdaWarehouse::WarehouseReports::ReportDefinition.maintain_report_definitions
+    # I'm not sure that we need these on every test run
+    # GrdaWarehouse::Utility.clear!
+    # Delayed::Job.delete_all
```
eanders: If the tests are passing without them, then we don't.  There were some tests, historically, that didn't cleanup well, so this was intended to catch that.  But if we've reached a stables state, it would be great to remove them.

spec/rails_helper.rb
```diff
@@ -70,28 +70,37 @@
...
+
+    example_file_paths = RSpec.world.filtered_examples.values.flatten.map { |e| e.metadata[:file_path] }.uniq
+    # load fixpoints if we're running tests that may need them
+    if example_file_paths.grep(%r{/drivers/(hud_path_report|hud_spm_report|hud_data_quality_report)/}).any? # rubocop:disable Style/RegexpLiteral
+      Dir.glob('{drivers,spec}/**/fixpoints/*.yml').each do |filename|
+        FileUtils.rm(filename)
+      end
+      Dir.glob('{drivers,spec}/**/fixpoints/*.sql').each do |filename|
+        # The 2024 test kit is HUGE, so we'll use the preprocessed version
+        # This means when the test kit changes, we need to update both the CSVs and the SQL fixpoint
```
eanders, line 87:
```suggestion
        # The 2024 test kit is HUGE, so we'll use the preprocessed version
        # This means when the test kit changes, we need to update both the CSVs and the SQL fixpoint
        # To rebuild, push a commit that contains `[gh:rebuild_fixpoints]` in the commit message.
```

eanders (approved):
Thanks @ttoomey !  This looks great.  Do you have a sense of how much time overall this shaves off?  It looks like this is most useful when running a handful of tests locally, which is also a big win.

gigxz (approved):

ttoomey:
> Do you have a sense of how much time overall this shaves off? It looks like this is most useful when running a handful of tests locally, which is also a big win.

It shouldn't impact overall CI run time. It's really just to allow rapid iteration on a single test. Once spring is started, the time to run one test is about 3 seconds (in my env).

martha:
I rely pretty heavily on the Rubymine graphical debugger for running rspec tests, and this has broken something for me, but after spending about 1.5hr trying to debug it, I think I'm in over my head. I'll leave it for now and try to come back to it later. :(

ttoomey:
> I rely pretty heavily on the Rubymine graphical debugger for running rspec tests, and this has broken something for me, but after spending about 1.5hr trying to debug it, I think I'm in over my head. I'll leave it for now and try to come back to it later. :(

Sorry @martha! Happy to pair on this if it would be helpful

PR #4771, opened 2024-09-30
Add protection for time analysis on login attempts to ennumerate valid login usernames

dtgreiner:

## Description

Enforce a minimum 2 second login time to protect against identifying valid usernames using login time analysis. 

## Type of change

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -6,10 +6,29 @@
 
 class Users::SessionsController < Devise::SessionsController
   include AuthenticatesWithTwoFactor
+  # Start the time log before other methods are called in the authentication stack so we can get the server begin time for the login attempt
+  prepend_before_action :begin_time_log, only: :create
   prepend_before_action(
     :authenticate_with_two_factor,
     if: -> { action_name == 'create' && two_factor_enabled? },
   )
+
+  # Minimum required login processing time for ALL login attempts (seconds)
+  MIN_REQ_LOGIN_TIME = 2
+
+  def begin_time_log
+    # Timestamp for tracking login time to help ensure that the application response time is consistent for valid/invalid usernames.
+    # This helps prevent using login method time to enumerate valid vs invalid usernames
+    @session_create_timestamp = Time.now
+  end
+
+  def end_time_log
+    # Wait a semi-random length of time to return after a login attempt to prevent using login time analysis to enumerate valid usernames.
+    # We want to make sure valid and invalid username attempts all take the same minimum amount of time to process and then add a random salt.
+    elapsed = Time.now - @session_create_timestamp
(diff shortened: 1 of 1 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -6,18 +6,41 @@
...
     if: -> { action_name == 'create' && two_factor_enabled? },
   )
 
+  # Minimum required login processing time for ALL login attempts (seconds)
+  MIN_REQ_LOGIN_TIME = 2
+
+  def begin_time_log
+    # Timestamp for tracking login time to help ensure that the application response time is consistent for valid/invalid usernames.
+    # This helps prevent using login method time to enumerate valid vs invalid usernames
+    @session_create_timestamp = Time.now
```
ttoomey: It probably doesn't matter here but I think it's better to always use Time.current instead of Time.now ([ref](https://thoughtbot.com/blog/its-about-time-zones#do-use))

ttoomey (approved):
Looks good!

Fwiw, [there's also an "around_action"](https://guides.rubyonrails.org/action_controller_overview.html#after-action-and-around-action-callbacks) callback that could work here, although what you have is totally fine.

eanders (approved):
Let's fix the `Time.current` note Theron had.  I think this is a fine approach, but wouldn't it be easier to just introduce a random wait on every login request?  Something between 0 and 1.5 seconds?  This probably ensures overall faster responses, so maybe that's better.

ttoomey:
>  wouldn't it be easier to just introduce a random wait on every login request? Something between 0 and 1.5 seconds?

If I'm understanding the problem correctly, I think Dave's approach should mask the timing issue as long as our auth is consistently less than 2 secs. If there's just a random wait time it seems like an attacker could still get useful timing information (although it would require a larger sample)

That reminds me, we might also want to check that wait_time is positive. Sleep on a negative number raises an exception and it looks like it might occur in the unlikely event that `elapsed > 2.5`

app/controllers/users/sessions_controller.rb
```diff
@@ -6,18 +6,41 @@
...
+  end
+
+  def end_time_log
+    # Wait a semi-random length of time to return after a login attempt to prevent using login time analysis to enumerate valid usernames.
+    # We want to make sure valid and invalid username attempts all take the same minimum amount of time to process and then add a random salt.
+    elapsed = Time.now - @session_create_timestamp
+    wait_time = MIN_REQ_LOGIN_TIME - elapsed + rand(0.5..1)
+    sleep(wait_time) if elapsed < MIN_REQ_LOGIN_TIME
```
eanders: @ttoomey like this?
```suggestion
    sleep(wait_time) if wait_time.positive? && elapsed < MIN_REQ_LOGIN_TIME
```

dtgreiner:
The `Time.now` -> `Time.current` change and the `wait_time.positive?` change have been added to the PR

PR #3602, opened 2023-10-03
186114231 enrollment performance improvement

ttoomey:
## Description
https://www.pivotaltracker.com/story/show/186114231

Improves the responsiveness of the api endpoints for services and enrollments (households) on projects with large numbers of records.

* Rework hmis_households view to use a union. Previous view was quite slow for a large record set; I couldn't track down the exact cause. Benchmark: count on 10,000 records went from 2,123ms to 140ms locally
* Replace pluck with a correlated subquery in project.households_including_wip
* changed household visible_by to use correlated subquery
* adjust project.hmis_services to work around enormously slow pg queries
* adjust hmis_services view: add missing data_source_id to join to service types, order view to help with query performance
* adjust household filter to be more performant, replacing join to enrollment with subqueries

## Type of change
- [x] Performance fix

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [ ] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4779, opened 2024-10-01
Fix double-layer bug with case management permission

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6758

- Add guard in the graphql permission checker to make sure the data source ID on the object being returned matches the context
- Add viewable scope to Staff Assignments and use it in the user and household staff assignment queries
- Add spec test

## Type of change
- [x] Bug fix
- [ ] Dependency update


drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
```diff
@@ -8,6 +8,10 @@ def self.current_permission_for_context?(context, permission:, entity:)
     current_user = context[:current_user]
     return false unless current_user&.present?
 
+    if entity.data_source_id
```
martha, line 11: Will these entities always have `data_source_id`? I wasn't sure, but we can remove the guard if they always will. I was thinking since this is used on the base graphql object it won't always be present.

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -103,8 +103,8 @@ def enrollment_access_summaries(filters: nil)
     end
 
     def staff_assignments
-      # n+1, not performant for queries on colletions
-      object.staff_assignments.order(created_at: :desc, id: :desc)
+      # n+1, not performant for queries on collections
+      object.staff_assignments.viewable_by(current_user).order(created_at: :desc, id: :desc)
```
martha, line 107: Just for posterity here's the link to the conversation when this bug slipped in: https://github.com/greenriver/hmis-warehouse/pull/4578/files#r1697269513 😅

drivers/hmis/app/models/hmis/staff_assignment.rb
```diff
@@ -11,4 +11,9 @@ class Hmis::StaffAssignment < Hmis::HmisBase
   belongs_to :user, class_name: 'Hmis::User'
   belongs_to :household, class_name: 'Hmis::Hud::Household', primary_key: [:data_source_id, :HouseholdID], foreign_key: [:data_source_id, :household_id], inverse_of: :staff_assignments
   belongs_to :staff_assignment_relationship, foreign_key: :hmis_staff_assignment_relationship_id, class_name: 'Hmis::StaffAssignmentRelationship'
+
+  scope :viewable_by, ->(user) do
+    # Special case this, rather than using EnrollmentRelated concern, because we have to join through Household
+    joins(household: :enrollments).merge(Hmis::Hud::Enrollment.viewable_by(user))
```
gigxz: Theron made some perf improvements to the Household.viewable_by scope [here](https://github.com/greenriver/hmis-warehouse/pull/3602/files#diff-e9d936441ab6d0722484ec8a11511424d896354709c10ad718fd658cf317a955) by using a correlated subquery. 
Since this loads on the home page for some users, I think we should look at perf. Is this faster?:
```
joins(:household).merge(Hmis::Hud::Household.viewable_by(user)
```

drivers/hmis/app/models/hmis/staff_assignment_relationship.rb
```diff
@@ -5,6 +5,9 @@
 ###
 
 class Hmis::StaffAssignmentRelationship < Hmis::HmisBase
+  # Note that this model isn't currently tied to a data source, which means that if we have multiple HMISes
+  # within the same installation, they will all see the same options for StaffAssignmentRelationship.
+  # We could change this if the need arises.
```
gigxz, line 10: Good thought. I have been adding bullet points to https://github.com/open-path/Green-River/issues/6612 as I notice things, recently. The big one is the config tables (https://github.com/open-path/Green-River/issues/6691) and I think this falls into that category. Added a bullet point there

drivers/hmis/app/models/hmis/staff_assignment.rb
martha: They are comparable in speed, but the original version had a bug which your update fixes. I also think your update is clearer. Thanks!

They are not too fast, 6 seconds on average for me locally. I'd like to check it out on QA.

Bug was: When joining to enrollments, the dashboard table would show n rows per household where n is the number of enrollments in the household (but all rows would display the HoH's name and link to the HoH's enrollment).

<details>
<summary>timing details</summary>

```
joins(:household).merge(Hmis::Hud::Household.viewable_by(user))
```
- Completed 200 OK in 6143ms (Views: 0.7ms | ActiveRecord: 5920.9ms | Allocations: 259556)
- Completed 200 OK in 6312ms (Views: 0.6ms | ActiveRecord: 6124.2ms | Allocations: 213006)
- Completed 200 OK in 5948ms (Views: 1.0ms | ActiveRecord: 5794.0ms | Allocations: 94312)

```
joins(household: :enrollments).merge(Hmis::Hud::Enrollment.viewable_by(user))
```
- Completed 200 OK in 5938ms (Views: 2.3ms | ActiveRecord: 5720.6ms | Allocations: 247556)
- Completed 200 OK in 5983ms (Views: 8.4ms | ActiveRecord: 5763.1ms | Allocations: 128526)
- Completed 200 OK in 6422ms (Views: 3.9ms | ActiveRecord: 6231.3ms | Allocations: 128522)
</details>

gigxz: okay good catch. sounds good re: testing perf on qa.

drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
gigxz: Usually but not always, I don't think. I think this check needs to go after (or in) the `entity_access_loader_factory` which is responsible for basically translating the "entity" into the correct access Loader (Project, Organization, Client, or Data Source). _Most_ (or all) of the time, the entity that we pass does have a data_source_id. But there is nowhere that enforces that.

This will throw a `NoMethodError` if there is anywhere that we pass an entity without data_source_id. So at minimum lets roll that back or add a check for `respond_to?(:data_source_id)` so we don't get a new bug.


For more detail on the entity loader..it supports passing any record that has a project association (resolve_through_project), so it SHOULD work on a record associated to project, even if the record doesn't have data_source_id. I'm not sure if we have a usage like that right now, but I don't think we should prevent it. https://github.com/greenriver/hmis-warehouse/blob/7690e9f42e5083fb676468211443eb7acee1eb41/drivers/hmis/app/models/hmis/entity_access_loader_factory.rb#L70-L80 


So lets look at this more closely, I'd be happy to pair on it!

martha: Thanks, good point, and sorry for merging this prematurely! I opened https://github.com/greenriver/hmis-warehouse/pull/4788 adding `respond_to?(:data_source_id)` to prevent this bug. but let me take a minute to grok the larger change of moving this into the entity access loader factory...

gigxz: thanks! ok lets discuss when you're ready!

PR #4787, opened 2024-10-03
Fix Project Location map link permissions

martha:

## Description

PIT QA fix for: https://github.com/open-path/Green-River/issues/6481#issuecomment-2389756617

**Bug**: For a user using legacy permissions (not ACLs), you can still see the "Client Locations" link in the Warehouse Project page, even if you don't have the permission to view project locations.

The issue is that the `Project`'s `viewable_by_entity` [scope](https://github.com/greenriver/hmis-warehouse/blob/10de915680000021ecdd89df692f4cf3f22a6dd8/app/models/grda_warehouse/hud/project.rb#L298-L323) accepts a permission arg, but only uses the permission if the user is using ACLs. If they are using legacy access controls then it just checks viewability.

**Fix:** Per @gigxz 's suggestion I've simplified the permission check so it's not ACL-compatible, since this view on the whole is not ACL-ready (it is also using `can_edit_projects?` and `can_delete_projects?` without checking against the specific entity).

(I thought I [checked this case](https://github.com/greenriver/hmis-warehouse/pull/4696#discussion_r1763441566) before but was likely fooled by local caching.)

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/app/models/grda_warehouse/hud/project.rb
+++ b/app/models/grda_warehouse/hud/project.rb
@@ -292,7 +292,8 @@ module GrdaWarehouse::Hud
     #   within the context of reporting confidential_scope_limiter is almost always non_confidential
     #   within the client dashboard context, confidential_scope_limiter is :all, which includes confidential projects
     #   names of confidential projects are obfuscated unless the user can_view_confidential_project_names
-    # @param permission [Symbol] a permission to determine the scope for which the projects are viewable
+    # @param permission [Symbol] a permission to determine the scope for which the projects are viewable.
+    #   Caution, this permission is NOT checked if the user is using legacy permissions (not on ACLs).
     scope :viewable_by, ->(user, confidential_scope_limiter: :non_confidential, permission: :can_view_projects) do
       query = viewable_by_entity(user, permission: permission)
       # If a user can't report on confidential projects, exclude them entirely
@@ -327,10 +328,6 @@ module GrdaWarehouse::Hud
         )
       end
       # END_ACL
-    end
-
-    def can?(user, permission: :can_view_projects)
-      self.class.viewable_by(user, permission: permission).where(id: id).exists?
     end
 
     scope :editable_by, ->(user) do
--- a/app/views/projects/show.haml
+++ b/app/views/projects/show.haml
@@ -49,7 +49,7 @@
           %tr
             %th Project Contacts
             %td= link_to 'View/Edit', project_contacts_path(@project)
-        - if @project.can?(current_user, permission: :can_view_project_locations) && RailsDrivers.loaded.include?(:client_location_history)
+        - if can_view_project_locations? && RailsDrivers.loaded.include?(:client_location_history)
           %tr
             %th Client Locations
             %td= link_to 'View', map_client_location_history_project_path(@project)


eanders:
@martha I'd like to dig into why this isn't ACL ready yet.  `Project.viewable_by` defers to `Project.viewable_by_entity`, which _is_ setup to handle ACLs.   Was it just that it's not setup to handle **multiple** permissions for a single entity?

app/models/grda_warehouse/hud/project.rb
```diff
@@ -329,10 +330,6 @@ def serves_children?
       # END_ACL
     end
 
-    def can?(user, permission: :can_view_projects)
-      self.class.viewable_by(user, permission: permission).where(id: id).exists?
```
gigxz, line 333: @eanders the underlying problem Martha identified was that  `viewable_by_entity(user, permission: x)` completely ignores the permission arg for non-ACL users.
https://github.com/greenriver/hmis-warehouse/blob/fe54601cc6d5a7f5a36a294e1a2347450f788c63/app/models/grda_warehouse/hud/project.rb#L304-L330

Is that a feature or a bug? I think it is misleading naming. I would expect it to check the permission in both cases.

app/views/projects/show.haml
```diff
@@ -49,7 +49,7 @@
           %tr
```
gigxz, line 49: @eanders am I correct in reading that this file is not yet "ACL-ready"? On line 108 below it appears to globally check `can_edit_projects?`, rather than checking if the user can edit _this_ project. (Maybe the correct check happens when the action is performed, but when deciding whether to show the button?). Same goes for the project delete button.

gigxz: This question is unrelated to the bug. It's just something I noticed.

eanders: The view is probably not fully updated, but the controller should be.  As in you shouldn't be able to get to the page unless you have access setup correctly ACL or no.

gigxz: Okay thats what I thought. Thanks.

gigxz (approved):

app/models/grda_warehouse/hud/project.rb
martha: @eanders @gigxz basically this PR prioritizes non-ACL user ux over ACL user ux. Is that what we want to do?

Behavior on release-135:
- non-ACL users always see the "Client Locations" button if they can view the project, even without `can_view_project_locations` (incorrect)
- ACL users see the "Client Locations" button only if they have `can_view_project_locations` on this project (correct)

Behavior in this PR:
- non-ACL users only see the "Client Locations" button if they have some role granting them `can_view_project_locations` (correct)
- ACL users see the "Client Locations" button if they have `can_view_project_locations` anywhere, even if not in this project (incorrect, but we propose to plan to update this when we change the rest of this view to be ACL-compatible)

@eanders let me know if you think this makes sense to merge, or needs more work/thought?

gigxz: I agree with your assessment. I propose prioritizing the non-ACL user for `release-135`, since the majority of users are still role-based. And, addressing the underlying issue with viewable_by_entity in a future PR. That future PR can also address the other ACL issues in this view (e.g. always showing "edit" and "delete" without acl logic). IMO that should be a separate issue because it's for this entire view, not just the client location map link.

But interested to hear what Elliot thinks, lets wait on that.

eanders: Sorry, I probably won't have time to weigh in much for another couple hours.

app/models/grda_warehouse/hud/project.rb
```diff
@@ -292,7 +292,8 @@ def serves_children?
     #   within the context of reporting confidential_scope_limiter is almost always non_confidential
     #   within the client dashboard context, confidential_scope_limiter is :all, which includes confidential projects
     #   names of confidential projects are obfuscated unless the user can_view_confidential_project_names
-    # @param permission [Symbol] a permission to determine the scope for which the projects are viewable
+    # @param permission [Symbol] a permission to determine the scope for which the projects are viewable.
+    #   Caution, this permission is NOT checked if the user is using legacy permissions (not on ACLs).
```
martha, line 296: todo @martha - add comment - need to add the check for permissions somewhere else, because in role based, it's not related to the specific entity.

app/models/grda_warehouse/hud/project.rb
eanders:
```
# NOTE: at the moment, this will require an extra query per project for non-ACL users
def can?(user, permission:)
      return false if ! user.using_acls? && ! user.send("#{permission}?")

      self.class.viewable_by(user, permission: permission).where(id: id).exists?
    end
```

gigxz: Relevant HMIS pattern:
https://github.com/greenriver/hmis-warehouse/blob/bf9f28be3bfa1edc6e61188f1c08d39654066d56/drivers/hmis/app/models/hmis/user.rb#L74-L84

The `EntityAccessLoaderFactory` is a fancy thing Theron made that will resolve the relevant loader and entity fo the thing you're checking against. So if you want to check if someone has X permission for an Enrollment or a Service, it will return the relevant Project and ProjectAccessLoader. There's a lot of complex code there, and this was written in part to serve the performance needs of the GraphQL API, so I don't know if we need this level of complexity in the warehouse. 


In HMIS world, the view logic would be:
```
- if current_user.can_view_project_locations_for(@project)
  
```

gigxz: https://github.com/open-path/Green-River/issues/6785

PR #4788, opened 2024-10-03
Update GraphQL permission checker's data source guard

martha:

## Description

GH issue: https://github.com/open-path/Green-River/issues/6758
Bugfix followup from: https://github.com/greenriver/hmis-warehouse/pull/4779#discussion_r1786304300

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
+++ b/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
@@ -29,8 +29,6 @@ module GraphqlPermissionChecker
 
     # Return false because even if the user has permission to view this record in another hmis data source,
     # they should not be able to resolve it in the context of the hmis they are currently using (hmis_data_source_id).
-    # The viewable_by scope is our primary defense against this; this object-level authorization is a secondary guard.
-    # But caution because it's not applied 100% consistently; for example, organizations don't have object-level authorization
     return false unless current_user.hmis_data_source_id == data_source_id
 
     context.dataloader.with(Sources::UserEntityAccessSource, loader).load([subject, permission])
--- a/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
+++ b/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
@@ -8,7 +8,7 @@ module GraphqlPermissionChecker
     current_user = context[:current_user]
     return false unless current_user&.present?
 
-    if entity.data_source_id
+    if entity.respond_to?(:data_source_id)
       return false unless current_user.hmis_data_source_id == entity.data_source_id
     end
 
--- a/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
+++ b/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
@@ -8,10 +8,6 @@ module GraphqlPermissionChecker
     current_user = context[:current_user]
     return false unless current_user&.present?
 
-    if entity.respond_to?(:data_source_id)
-      return false unless current_user.hmis_data_source_id == entity.data_source_id
-    end
-
     # Just return false if we don't have this permission at all for anything
     return false unless current_user.send("#{permission}?")
 
@@ -22,7 +18,14 @@ module GraphqlPermissionChecker
         context.dataloader.with(Sources::ActiveRecordAssociation, association).load(record)
       end
     end
-    raise "Missing loader for #{entity.class.name}##{entity.id}" unless loader
+    raise "Missing loader for #{entity.class.name}##{entity.id}" unless loader && subject
+
+    data_source_id = if subject.is_a? GrdaWarehouse::DataSource
+      subject.id
+    else
+      subject.data_source_id
+    end
+    return false unless current_user.hmis_data_source_id == data_source_id
 
     context.dataloader.with(Sources::UserEntityAccessSource, loader).load([subject, permission])
   end
--- a/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
+++ b/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
@@ -20,11 +20,17 @@ module GraphqlPermissionChecker
     end
     raise "Missing loader for #{entity.class.name}##{entity.id}" unless loader && subject
 
+    # subject is expected to be one of Client, Project, Organization, or Data Source.
     data_source_id = if subject.is_a? GrdaWarehouse::DataSource
       subject.id
     else
       subject.data_source_id
     end
+
+    # Return false because even if the user has permission to view this record in another hmis data source,
+    # they should not be able to resolve it in the context of the hmis they are currently using (hmis_data_source_id).
+    # The viewable_by scope is our primary defense against this; this object-level authorization is a secondary guard.
+    # But caution because it's not applied 100% consistently; for example, organizations don't have object-level authorization
     return false unless current_user.hmis_data_source_id == data_source_id
 
     context.dataloader.with(Sources::UserEntityAccessSource, loader).load([subject, permission])
--- a/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
@@ -32,7 +32,7 @@ module Types
 
     def organizations(**args)
       # todo @martha - don't understand why this is failing the test but not manifesting in qa-hmis-2
-      resolve_organizations(Hmis::Hud::Organization.viewable_by(current_user), **args)
+      resolve_organizations(Hmis::Hud::Organization.all, **args)
     end
 
     clients_field :client_search, 'Search for clients' do |field|
--- a/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
@@ -31,7 +31,6 @@ module Types
     organizations_field :organizations
 
     def organizations(**args)
-      # todo @martha - don't understand why this is failing the test but not manifesting in qa-hmis-2
       resolve_organizations(Hmis::Hud::Organization.all, **args)
     end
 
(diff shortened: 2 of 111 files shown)


gigxz (approved):

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -31,7 +31,8 @@ def projects(**args)
     organizations_field :organizations
 
     def organizations(**args)
-      resolve_organizations(Hmis::Hud::Organization.all, **args)
+      # todo @martha - don't understand why this is failing the test but not manifesting in qa-hmis-2
```
martha: @gigxz sorry for the delay. I got swept down an unrelated rabbit hole.

- after we paired, your recommendation was to "add tests for checking against data source and organization" - which I still have a todo for [here](https://github.com/greenriver/hmis-warehouse/pull/4788/files#diff-d700a2b25aff90f1650d8cabf6ae2a496773aa4b668e12dc579c0e340a430413R283) and will add comments about below
- while I was working on trying to find where I would put those tests / if they already exist, I noticed that here in the organizations query, we weren't limiting scope to viewable_by the current user. This is bug, right? I reproduced it in an rspec test below, fixed the bug, and the test passes.
- BUT, it's a mystery to me why this bug isn't manifesting on qa-hmis-2. I visted the [organizations page](https://qa-hmis-2.openpath.host/projects) there and I don't see all the organizations from qa-hmis. Maybe this is not really a bug and I'm missing something?

drivers/hmis/spec/models/hmis/hud/organization_access_spec.rb
```diff
@@ -61,5 +72,10 @@
       viewable_orgs = Hmis::Hud::Organization.viewable_by(user_with_p2_access)
       expect(viewable_orgs).to contain_exactly(o2)
     end
+
+    it 'includes only organizations associated with the data source ID of the current user' do
```
martha, line 76: This isn't reproducing the bug, I just added it while I was playing around trying to understand all the pieces, and I figure we can commit it if it's helpful. Happy to revert if not.

drivers/hmis/spec/requests/hmis/organization_spec.rb
```diff
@@ -75,5 +75,19 @@
       orgs = result.dig('data', 'organizations', 'nodes')
       expect(orgs.length).to eq(0)
     end
+
+    context 'when there are multiple HMIS data sources' do
+      let!(:ds2) { create :hmis_data_source }
+      let!(:o5) { create :hmis_hud_organization, data_source: ds2, user: u1, OrganizationName: 'Pineapples' }
+      let!(:ds2_access_control) { create_access_control(hmis_user, ds2) }
+
+      it 'does not return organizations from other data source' do
```
martha, line 84: This is the spec test that reproduces the maybe-bug I wrote about above https://github.com/greenriver/hmis-warehouse/pull/4788/files#r1786621775

drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
```diff
@@ -22,7 +18,14 @@ def self.current_permission_for_context?(context, permission:, entity:)
...
     end
-    raise "Missing loader for #{entity.class.name}##{entity.id}" unless loader
+    raise "Missing loader for #{entity.class.name}##{entity.id}" unless loader && subject
+
+    data_source_id = if subject.is_a? GrdaWarehouse::DataSource
+      subject.id
+    else
+      subject.data_source_id
+    end
+    return false unless current_user.hmis_data_source_id == data_source_id
```
gigxz, line 32: Can you add some comments here about (1) `subject` is expected to be one of data source, project, organization, and (2) return false because even if the user has permission to view this record in another hmis data source, they should not be able to resolve it in the context of the hmis they are currently using (hmis_data_source_id)

This code is tricky so any comments will surely help us in the future :)

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Yeah good question, this code jumps through some hoops. It's because the viewable_by scope is already applied in the `HasOrganizations` concern. I think we use that pattern in several places, so I think we can leave it as is. The `user` args defaults to `current_user`:

https://github.com/greenriver/hmis-warehouse/blob/f550cdd12e2ad040c59c774056850bc55ce4d0de/drivers/hmis/app/graphql/types/hmis_schema/has_organizations.rb#L36-L41

drivers/hmis/spec/models/hmis/hud/organization_access_spec.rb
gigxz: I think that's a good test case to have. Doesn't look like it's covered by other tests. Thanks!

drivers/hmis/spec/models/hmis/hud/organization_access_spec.rb
```diff
@@ -61,5 +72,10 @@
...
+      viewable_orgs = Hmis::Hud::Organization.viewable_by(user_with_ds1_and_ds2_access)
+      expect(viewable_orgs).to contain_exactly(o1, o2, o3)
```
gigxz, line 78: maybe just to drive the point home, we could add:
```ruby
# change hmis_data_source_id to ds2 to ensure it can access o4
user_with_ds1_and_ds2_access.hmis_data_source_id = ds2.id
viewable_orgs = Hmis::Hud::Organization.viewable_by(user_with_ds1_and_ds2_access)
expect(viewable_orgs).to contain_exactly(o4)

```

drivers/hmis/spec/requests/hmis/organization_spec.rb
```diff
@@ -75,5 +75,19 @@
...
+        # expect(Hmis::Hud::Organization).to receive(:viewable_by).and_return(Hmis::Hud::Organization.all)
```
martha: Ok, here's a thread to talk about the original todo you left me with, which is testing the graphql permission checker against organization and data source. 

This line is commented out because, if I add it, I would expect the graphql query to fail. Instead, what happens is that the query succeeds and returns 5 orgs, including the one in ds2. 

I think, maybe, it is because when the graphql permission checker's `current_permission_for_context?` is called on a client, enrollment, or project, it is passed with a specific permission, whereas when it's called on base_field, it is passed without a specific permission. Inside base_field's `authorized?` method, you can see that it is only checking `current_permission_for_context?` if `permissions.any?`. I haven't fully followed what this means, but I think it implies that the recent changes we made in `current_permission_for_context` only impact entities where the access loader is project, client, or enrollment.

drivers/hmis/spec/requests/hmis/organization_spec.rb
gigxz: Does this reproduce it? I can't tell from the assertion, since its just saying there are 4 nodes.

drivers/hmis/spec/requests/hmis/organization_spec.rb
gigxz: Okay.
Let's huddle, but here are some initial notes:
* there is no permission for org visibility, it's purely based on access control "collections". If you wanted to test organization-level permissions that would be either `can_edit_organization` (evaluated by SubmitForm mutation) and `can_delete_organization` (evaluated by DeleteOrganization)
* there is no object-level authorization on the organization type
* field-level authorization is different from object-level authorization. I'm not totally tracking on yoru read of

martha: Sum up from huddle: This was just my misunderstanding of `base_field`, which is for field level permissions (as opposed to object level permissions).

drivers/hmis/spec/requests/hmis/organization_spec.rb
martha: In fact it didn't, because the model factories were set up wrong. It turns out that 
```
let!(:o5) { create :hmis_hud_organization, data_source: ds2, user: u1 }
```
results in an o5 that is still in ds1.

```
let!(:o5) { create :hmis_hud_organization, data_source: ds2 }
```
is correct. No bug here!

gigxz: :| right, because u1 overwrites the data source id

drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb
```diff
@@ -22,7 +18,20 @@ def self.current_permission_for_context?(context, permission:, entity:)
...
+    data_source_id = if subject.is_a? GrdaWarehouse::DataSource
+      subject.id
+    else
+      subject.data_source_id
+    end
+
+    # Return false because even if the user has permission to view this record in another hmis data source,
+    # they should not be able to resolve it in the context of the hmis they are currently using (hmis_data_source_id).
+    # The viewable_by scope is our primary defense against this; this object-level authorization is a secondary guard.
+    # But caution because it's not applied 100% consistently; for example, organizations don't have object-level authorization
```
gigxz: So this doesn't feel like the totally right place for this warning, since this class `GraphqlPermissionChecker` _happens_ to be used for object-level authorization, but its not ONLY used for object-level authorization. Maybe you could put this comment on the `self.authorized?` implementations?


It would also be great to have this in the README. Would you mind adding an "faq" or section about this? Here is a rough description of what we discussed earlier:

**Authorization considerations when writing GraphQL queries and mutations**
* Most HMIS records have a scope called `viewable_by`. Always use this scope when looking up a record. Most HMIS records' viewable_by scopes also perform a filter on `current_user.hmis_data_source_id`, ensuring that objects from other HMIS data sources are not included.
* When checking permissions, 90% of the time we want to check against a specific entity. (Eg can the user edit THIS enrollment, not just can the user edit any enrollments). The preferred way to do this is using `current_permission?(permission:, entity:)` because it uses the graphql data loader and also ensures that the entity belongs to the correct data source.
* For nodes that are resolved from lots of places in the graph (and have PII), add object-level authorization to the object type (`def self.authorized?`). These are a secondary check to ensure the presence of the minimal permissions needed to resolve the object.

gigxz (approved):
🙌

gigxz:
Thanks!  merging

PR #4791, opened 2024-10-04
Flash form errors on override page & add optional description field to describe Why the override is being added

dtgreiner:

## Description

Add a description field to the import override table to describe why the override is being added & fix issue with override submission allowing submission without all required fields entered

## Type of change


drivers/hmis_csv_importer/app/models/hmis_csv_importer/import_override.rb
```diff
@@ -125,6 +130,10 @@ def describe_when
     "#{replaces_column} is #{replaces_value}" if replaces_value.present?
   end
 
+  def describe_why
+    description == ':NULL:' ? nil : description
```
eanders: Why does the description need the `:NULL:`?  It shouldn't affect the processing at all, I think this method can just go away and the table can just call `override.description`

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_overrides/new.haml
```diff
@@ -10,6 +10,7 @@
       .well
         .form-inputs
           = f.input :file_name, collection: HmisCsvImporter::ImportOverride.available_classes, label: 'File', as: :select_two, include_blank: false
+          = f.input :description, label: 'Description', hint: 'Why is this override being created?', required: false
```
eanders, line 13: required is false by default.
```suggestion
          = f.input :description, label: 'Description', hint: 'Why is this override being created?'
```

drivers/hmis_csv_importer/app/controllers/hmis_csv_importer/import_overrides_controller.rb
```diff
@@ -16,7 +16,8 @@ def index
   end
 
   def create
-    @override = import_override_source.create!(permitted_params.merge(data_source_id: @data_source.id))
+    @override = import_override_source.create(permitted_params.merge(data_source_id: @data_source.id))
+    flash[:alert] = @override.errors.full_messages.join(', ') unless @override.valid?
```
eanders, line 20: This is fine, but I _thought_ `respond_with` usually handled the errors natively.  Maybe we're not using `simple_form_for`?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/import_override.rb
dtgreiner: I can adjust this one, I had it set this way to mimic the way the other fields were being displayed.

eanders: Yeah, I don't think we need to yet.  Thanks!

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_overrides/new.haml
dtgreiner: This driver appears to be using the opposite logic and defaulting to true. The `required: false` tag is being applied to the optional fields in the views.

eanders: weird.  Thanks

drivers/hmis_csv_importer/app/controllers/hmis_csv_importer/import_overrides_controller.rb
dtgreiner: It is using `simple_forms_for`, and  I remember fighting with this one for a while. It looks like it should be showing the errors at the field, but I wasn't able to determine where along the line the form was going against that. Happy to dig more into it to see if anything pops out this time.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/import_override.rb
dtgreiner: This has been updated - https://github.com/greenriver/hmis-warehouse/pull/4791/commits/bcb01769041fb8718afddcb7488859d6a81f4c8d

drivers/hmis_csv_importer/app/controllers/hmis_csv_importer/import_overrides_controller.rb
eanders: That's fine, this should get very little use, let's push it over the finish line and see what turns up in QA.

eanders (approved):

PR #4792, opened 2024-10-04
6210 hmis-supplemental-data (configurable client external data tabs)

ttoomey:
## _Merging this PR_

## Description
[6210](https://github.com/open-path/Green-River/issues/6210)

### features
* add supplemental data sets
* adds supplemental data sets support collections
* can_view_supplemental_client_data permission

#### misc changes
* render_paginated_list helper
* validate   bucket on remote credential s3 bucket config

## Type of change
New feature, Code clean-up

### Create data set
* Assume you have a Data Source, "Data Source A"; this data source has source client / enrollment data
* as an admin, go to the Supplemental Data Set list page: 
  * visit "Data -> Data Sources" and choose a "Data Source A"
  * click the "Supplemental Data Sets" button at the bottom of the button list
* create a new data set, let's say "Data Set A"
  * Add a field configuration JSON, you can copy the example in the form hint

#### Configure permission
* Setup a collection:
  * Visit "Warehouse Admin -> Access -> Collections" 
  * create a new collection and choose "Supplemental Data Sets" as data set type
  * add "Data Set A" to the new collection
* Create an Access Control
  * visit  "Warehouse Admin -> Access -> Access Controls"
  * create a new access control
  * add the collection you just created and a role/group for your account
* Assign permission
  * Visit  "Warehouse Admin -> Access -> Roles & Permissions"
  * Search for "supplemental client"
  * Enable the "Can view supplemental client data" on the roles you added to the access control
 
#### View the data in the client dashboard
* visit the client dashboard for a client record who's source client /enrollment records are in "Data Source A"
* You should see the a tab for each data set at the top of the dashboard, "Data Set A" in this example
* Clicking on "Data Set A" should show a table matching the field configuration of the given data set

#### Populate the client data
A direct upload feature allows data processing to be tested without needing s3
* create a csv sheet with the appropriate structure, see blow
  * For a client data set, each row should include a personal_id for a source client in "Data Source A"
  * for enrollment-based data sets, each row should have an enrollment_id in "Data Source A"
* upload the data set
  * Go back to the  Supplemental Data Set list page for your "Data Source A".
  * Select "upload CSV" from the actions menu for "Data Set A" (right-most column of the table)
  * use the form to upload a csv file for immediate processing
* Once uploaded, visit the client dashboard. Uploaded values should be visible for clients 
example csv for a "Data Set A":
```csv
personal_id,primary_language
dbb6d274b88db0b2bb6ffe4828f5067c, English
```

#### duplicate rows
if multiple rows have a value for the same field:
* for single-valued fields, only the value from the first row is saved
* for multi-valued fields, the values from all rows are saved in row order

#### multi-valued row formatting
one row may contain multiple values fro the same field delimited with '|'


app/views/admin/collections/_supplemental_data_sets.haml
```diff
@@ -0,0 +1,20 @@
+%h2 Data Sets Included (#{@collection.supplemental_data_sets.count})
```
ttoomey, line 1: copy pasted this view from cohorts

app/views/admin/collections/entities/_supplemental_data_sets.haml
```diff
@@ -0,0 +1,18 @@
+%table.table.datatable.border.collections
```
ttoomey, line 1: copy pasted view from cohorts

drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
```diff
@@ -0,0 +1,81 @@
...
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# a "tab" on the client dashboard for viewing a data set.
+# This controller is for both client and enrollment-based data sets
+module HmisSupplemental
+  class ClientDataSetsController < ApplicationController
+    include ClientController
```
ttoomey, line 11: This controller gets flagged by brakeman for having user-controlled execution. It seems to be because of this mix-in

drivers/hmis_supplemental/app/controllers/hmis_supplemental/data_sets_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    protected
+
+    def load_data_set
+      data_set_scope.find(params[:id])
+    end
+
+    def data_set_scope
+      HmisSupplemental::DataSet.order(:id)
```
ttoomey: assume admins can see all the data sets

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
```diff
@@ -0,0 +1,81 @@
...
+             foreign_key: 'data_set_id',
+             dependent: :restrict_with_exception
+    validates :name, presence: true
+    validates :object_key, presence: true
+    has_many :group_viewable_entities,
+             class_name: 'GrdaWarehouse::GroupViewableEntity',
+             as: :entity,
+             dependent: :destroy
+
+    scope :viewable_by, ->(user) do
```
ttoomey, line 28: something like this may already exist. Happy to dry up if it's there's another implementation

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/index.haml
```diff
@@ -0,0 +1,11 @@
...
+  .o-page__title
+    - content_for :title, "HMIS Supplemental Data Sets"
+    %h1= content_for :title
+
+  .d-flex.mb-4
+    .ml-auto.text-right
+      = link_to url_for(action: :new), class: 'btn btn-primary mb-2' do
+        %span.icon-plus
+        Add Data Set
+  = render_paginated_list(scope: @data_sets, item_name: 'data set', list_partial: 'list')
```
ttoomey: DRYed up this common pattern

app/models/collection.rb
```diff
@@ -152,6 +154,7 @@ def entity_types
       projects: 'Projects',
       reports: 'Reports',
       cohorts: 'Cohorts',
+      supplemental_data_sets: 'Supplemental Data Sources',
```
eanders: I think everywhere else you are calling it `Supplemental Data Sets` is there a reason this is different?

app/models/collection.rb
```diff
@@ -128,6 +129,7 @@ def self.available_collection_types
       'Project Groups',
       'Reports',
       'Cohorts',
+      'Supplemental Data Sets',
```
eanders, line 132: Probably not worth doing right now, but looking at the number of places you had to touch in this one file, I think Collections could use a bit of yak shaving.  If we had a single constant hash that defined a bunch of bits, like:
```
{
 supplemental_data_sets: {
  title: 'Supplemental Data Sets',
  source_class: 'HmisSupplemental::DataSet',
  name_method: :name,

},
...
}
```
I think we could add new types of collections by just adding the variation to the constant.
Do you think that's worth doing @ttoomey ?

app/models/role.rb
```diff
@@ -560,6 +560,12 @@ def self.permissions_with_descriptions
         category: 'Client Extras',
         sub_category: 'Privacy',
       },
+      can_view_supplemental_client_data: {
+        description: 'Access to external supplemental client data',
```
eanders:
```suggestion
        description: 'Access to assigned external supplemental client data',
```

app/views/clients/_default_tab_navigation.haml
```diff
@@ -104,6 +104,18 @@
...
+    supplemental_data_sets = HmisSupplemental::DataSet.viewable_by(current_user).where(data_source_id: @client.source_clients.select(:data_source_id))
+    supplemental_data_sets.each do |data_set|
+      key = hmis_supplemental_data_set_client_data_set_path(id: @client.id, data_set_id: data_set.id)
+      hmis.merge!({
+        key => {
+          title: data_set.name,
+          permission: :can_view_supplemental_client_data?,
+        }
+      })
+    end
```
eanders: Since this is coming from a driver, to be good citizens, we need to wrap calls in the main app with something like `if RailsDrivers.loaded.include?(:hmis_supplemental)`.

I didn't call it out in the other files, but if the code is referencing a driver, we should be making sure it _can_ be run without the driver.

app/models/collection.rb
ttoomey: Yes, I was thinking the same. This could definitely use some TLC and the approach you have above could work well. I would do that in a different issue though; perhaps adding tests (if they don't exist already).
https://github.com/open-path/Green-River/issues/6801

app/models/collection.rb
ttoomey: Good catch, I used the wrong word. Will fix

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
```diff
@@ -0,0 +1,22 @@
...
+    .alert.alert-danger{style: 'display: block;'}
+      %p Failed to save
+      %ul.error_list
+        - data_set.errors.full_messages.uniq.each do |msg|
+          %li= msg
+
+  .form-inputs
+    = f.input :name, required: true
+    = f.input :owner_type, collection: ['enrollment', 'client'], required: true
+    = f.input :data_source_id, collection: data_sources.order(:name).map {|d| [d.name, d.id]}, required:true
```
eanders:
```suggestion
    = f.input :owner_type, collection: ['enrollment', 'client'], required: true, as: :select_two
    = f.input :data_source_id, collection: data_sources.order(:name).map {|d| [d.name, d.id]}, required:true, as: :select_two
```

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
```diff
@@ -0,0 +1,22 @@
...
+    = f.input :field_config, as: :text, label: 'Field Configuration JSON', hint: 'example: `[{"key":"widget","type":"float","label":"The Widget","multiValued":false}]`', input_html: {rows: 10}
+    = f.fields_for(:remote_credential) do |sf|
+      %fieldset
+        %legend S3 Configuration
+        = sf.input :bucket, as: :string, label: 'S3 bucket'
```
eanders: Pretty sure we need `region` as well, see `app/views/hmis_import_configs/_form.haml` for an example.

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
eanders, line 1: I wonder if this should be accessed via the data source similar to how we do "Custom Imports."  Potentially this could just be another type of "custom import."  What do you think about moving the configuration there @ttoomey ?  

Screenshot 2024-10-07 at 1 12 24 PM

drivers/hmis_supplemental/app/controllers/hmis_supplemental/data_set_uploads_controller.rb
```diff
@@ -0,0 +1,41 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# Admin data sets
+module HmisSupplemental
+  class DataSetUploadsController < ApplicationController
+    before_action :require_can_edit_users!
```
eanders: Let's use this permission set for authorization:
`app/controllers/data_sources/custom_imports_controller.rb`

drivers/hmis_supplemental/app/controllers/hmis_supplemental/data_sets_controller.rb
eanders: Moving this to the data source should solve this too (they should only be able to see data sets for data sources they can see.)

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
```diff
@@ -0,0 +1,131 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class HmisSupplemental::ImportJob
```
eanders: Should this inherit from `BaseJob`?

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
```diff
@@ -0,0 +1,131 @@
...
+      if values.empty?
+        log('Could not map custom data elements from CSV')
+        return
+      end
+
+      values = deduplicate_rows(values)
+      data_set.transaction do
+        # Delete and recreate. This could be an upsert
+        data_set.field_values.delete_all(:delete_all)
+        HmisSupplemental::FieldValue.import!(values, validate: false)
```
eanders, line 31: Do you think we should add some configuration to determine if this is a delta or a full-refresh?  I'm fine with assuming full-refresh for now, but we might want to make that an explicit choice (via argument).

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
```diff
@@ -0,0 +1,131 @@
...
+
+  def process_row(row)
+    data_set.fields.map do |field|
+      data = field.row_value_data(row)
+      next if data.nil?
+
+      owner_key = field.row_owner_key(row)
+
+      if owner_key.nil?
+        log("could not find owner for in CSV, line #{row[:row_number]}")
```
eanders:
```suggestion
        log("could not find owner in CSV, line #{row[:row_number]}")
```

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
```diff
@@ -0,0 +1,131 @@
...
+        result[:data] = Array(result[:data]) + Array(row[:data])
+      else
+        results[row_key] = row
+      end
+    end
+
+    results.values
+  end
+
+  def process_row(row)
```
eanders: maybe?
```suggestion
  def row_as_hash(row)
```

I was expecting `process_row` to save it somehow, probably just my own assumptions though.

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
```diff
@@ -0,0 +1,131 @@
...
+    csv = CSV.parse(csv_string, headers: true)
+    results = csv.map.with_index(1) do |row, row_number|
+      new_row = {}
+      # normalize headers
+      csv.headers.each do |header|
+        new_row[header.downcase] = row[header]&.strip.presence
+      end
+      next if new_row.empty?
+
+      new_row[:row_number] = row_number
```
eanders, line 126: Is it intentionally that `row_number` continues to increment even if we throw out a row?  (I think that's fine as long as the row number is being used to refer to the original CSV and not the cleaned data set, which I think is what's happening.)

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
eanders: We've implemented this in each model that is represented by a collection, but I believe the logic is different (by design) for each.

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
```diff
@@ -0,0 +1,81 @@
...
+      # not supporting legacy role base permissions
+      return none unless user.using_acls?
+
+      # first ask the WH db for all collections containing this resource type
+      gve_scope = GrdaWarehouse::GroupViewableEntity.where(entity_type: sti_name)
+      collection_ids = gve_scope.distinct.pluck(:collection_id)
+
+      # include global system collection
+      collection_ids += [Collection.system_collection(:data_sources)&.id].compact
```
eanders: This doesn't make sense to me.  `DataSet`s are explicitly assigned in collections.  The "system" collection of **All Data Sets** should take care of this.  This would be granting anyone who has access to **All Data Sources** (which are a representation of collections HMIS projects) access to all data sets.  
```suggestion
```

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
eanders: Since we only have one permission, it's probably not necessary, but this might want to take a permission argument for future expansion.

```suggestion
    scope :viewable_by, ->(user, permission: :can_view_supplemental_client_data) do
```

drivers/hmis_supplemental/app/models/hmis_supplemental/field.rb
```diff
@@ -0,0 +1,114 @@
...
+      else
+        raise "unknown type: #{type}"
+      end
+    end
+
+    def row_owner_key(row)
+      owner_id = row_owner_id(row)
+      return nil unless owner_id
+
+      "#{owner_type}/#{row_owner_id(row)}"
```
eanders, line 58: do we need to include `data_source_id` here?  neither PersonalID nor EnrollmentID are guaranteed to be unique across data sources.

drivers/hmis_supplemental/app/views/hmis_supplemental/data_set_uploads/_breadcrumbs.haml
```diff
@@ -0,0 +1,3 @@
+= content_for :crumbs do
+  = link_to hmis_supplemental_data_sets_path do
+    &laquo; Supplemental Data Sources
```
eanders:
```suggestion
    &laquo; Supplemental Data Sets
```

drivers/hmis_supplemental/app/controllers/hmis_supplemental/data_sets_controller.rb
```diff
@@ -0,0 +1,87 @@
...
+        flash[:notice] = "#{@data_set.name} was successfully updated."
+        redirect_to action: :index
+      else
+        render :edit
+      end
+    end
+
+    def destroy
+      @data_set = load_data_set
+      @data_set.destroy!
```
eanders, line 53: Should we offer a means of destroying the underlying fields/data as well?  Maybe deleting a data set should remove the data?  It will be removed from the view, but potentially hangs around indefinitely.

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/index.haml
eanders: Thanks!

lib/tasks/grda_warehouse.rake
```diff
@@ -282,6 +282,13 @@ namespace :grda_warehouse do
   desc 'Full import routine'
   task daily: [:environment, 'log:info_to_stdout'] do
     Importing::RunDailyImportsJob.new.perform
+
+    HmisSupplemental::DataSet.order(:id).each do |data_set|
+      HmisSupplemental::ImportJob.new.perform(data_set_id: data_set.id)
+    rescue StandardError => e
+      puts e.message
+      Sentry.capture_exception(e)
+    end
```
eanders: This should probably go in the `hourly` rake task and only execute in the set hour.  See `MaReports::CsgEngage::Report.run_if_ready` for a decent example.  It might mean adding an `hour` column to the data set, or in the `additional_headers` column for the credential.  That way we can configure which hour per installation, and it isn't dependent on a, multi-hour process to finish successfully before it runs.

Also, this should probably throw the job in the DJ queue instead of running inline in the rake task.

drivers/hmis_supplemental/app/controllers/hmis_supplemental/data_sets_controller.rb
ttoomey: right now we prevent it from being destroyed if there are values (restrict_with_exception) but maybe that's too troublesome. I'll change it to delete_all

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
ttoomey: Feels like support for incremental updates could come later? Maybe we will need to support it if the data sets are large. I'm unsure as to the scale

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
ttoomey: yeah, the intent is to report what lines in the CSV contain errors

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
ttoomey: Thanks -  I'll take it out. I guess my thinking was that a data set is 'owned' by a data source, so having access to 'all data sources' would give you access to every data set (there is similar logic in policy provider related to projects and clients). But maybe the relationship between data set and data source isn't really ownership so this doesn't make sense.

But it sounds like it would be correct include the system data set group here:
`collection_ids += [Collection.system_collection(:supplemental_data_sets)&.id].compact`

drivers/hmis_supplemental/app/models/hmis_supplemental/field.rb
ttoomey: We could include it but the data source is on data_set already so it's not strictly necessary. The values don't really make sense without the field config which is on data set. FieldValue.for_owner(client/enrollment) joins to dataset to filter them by data source

lib/tasks/grda_warehouse.rake
ttoomey: Does the hour need to be configurable? Looks like we're using hard-coded hours

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
ttoomey: not sure, maybe the permission used to determine visibility should stay as internal implementation; not something that the caller should know about.

drivers/hmis_supplemental/app/jobs/hmis_supplemental/import_job.rb
eanders: We can kick this can for sure.

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
eanders: I guess I see these scopes more as `accessible_by_within_context(user, context)` where context is a permission, but let's leave it as is for now, perhaps the policy provider will make it more clear what scopes we'll need.

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
```diff
@@ -0,0 +1,24 @@
...
+          %li= msg
+
+  .form-inputs
+    = f.input :name, required: true
+    = f.input :owner_type, collection: ['enrollment', 'client'], required: true, as: :radio_buttons
+    = f.input :field_config, as: :text, label: 'Field Configuration JSON', hint: 'example: `[{"key":"widget","type":"float","label":"The Widget","multiValued":false}]`', input_html: {rows: 10}
+    = f.fields_for(:remote_credential) do |sf|
+      %fieldset
+        %legend S3 Configuration
+        = sf.input :region, as: :string, label: 'S3 region'
```
eanders: This should be a collection
```suggestion
        = sf.input :region, label: 'S3 region', collection: GrdaWarehouse::RecurringHmisExport.available_s3_regions, as: :select_two
```
Hmm, that list should probably be moved to `AwsS3` though.

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
eanders: To find all collections that might include that data set, yes, the system collection for `:supplemental_data_sets` should include it, but won't that collection be included in `gve_scope.distinct.pluck(:collection_id)`?

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
ttoomey: sure I'll change it and move that constant to awss3.

drivers/hmis_supplemental/app/models/hmis_supplemental/data_set.rb
ttoomey: oh right! 🤦 I'll take it out

drivers/hmis_supplemental/app/views/hmis_supplemental/data_sets/_form.haml
```diff
@@ -0,0 +1,24 @@
...
+        = sf.input :region, label: 'S3 region', collection: GrdaWarehouse::RecurringHmisExport.available_s3_regions, as: :select_tw
```
eanders: I think this should have been:
```suggestion
        = sf.input :bucket, as: :string, label: 'S3 Bucket'
        = sf.input :region, label: 'S3 region', collection: GrdaWarehouse::RecurringHmisExport.available_s3_regions, as: :select_two
```

ttoomey: thanks!

drivers/hmis_supplemental/app/models/hmis_supplemental/field.rb
```diff
@@ -0,0 +1,114 @@
...
+      STRING_TYPE = 'string'.freeze,
+      ID_TYPE = 'id'.freeze,
+      INT_TYPE = 'int'.freeze,
+      FLOAT_TYPE = 'float'.freeze,
+      BOOLEAN_TYPE = 'boolean'.freeze,
+      DATE_TYPE = 'date'.freeze,
+    ].freeze
+    CLIENT_OWNER = 'client'.freeze
+    CLIENT_ID_FIELD = 'personal_id'.freeze
+    ENROLLMENT_OWNER = 'owner'.freeze
```
eanders:
```suggestion
    ENROLLMENT_OWNER = 'enrollment'.freeze
```

ttoomey: Good spot- thanks!

eanders (approved):
Looks good to me, thanks @ttoomey !
I'm going to let the tests pass again, then merge.

PR #4793, opened 2024-10-04
Fix form version bugs in Client Details and Case Notes

martha:

## Description
GH issue: https://github.com/open-path/Green-River/issues/6779

Regarding 2 in the linked ticket: I've proposed here to ~restrict the Instance's Definitions relationship to `{ where(status: 'published') }`. I don't yet feel confident that that will not have any other unintended knock-on effects and indeed it seems to be causing a test failure that I'm still investigating.~ Edit: there are for sure other places in the code where we need the definition regardless of its status. For example, when checking if the user has permission to delete a form instance, we check the role from the definition. I've proposed instead prioritizing the published definition in this relation.
Regarding 3: I've proposed to use the form definition ID attached to the case note's form processor. I'm eager for feedback on this idea.
Regarding 1: Interestingly, I never got to the bottom of why a draft form definitions was getting returned in the buggy [`client_detail_forms` query](https://github.com/greenriver/hmis-warehouse/blob/10de915680000021ecdd89df692f4cf3f22a6dd8/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L486-L490). Even when I ssh'ed in to training and ran the code from that query in a rails console, I could only get it to return published versions. However, I think this probably will be fixed by the fix to 2 above, so I haven't spent too much time investigating.

## Type of change
- [x] Bug fix
- [ ] Dependency update

### DIFF
--- a/drivers/hmis/app/graphql/mutations/submit_form.rb
+++ b/drivers/hmis/app/graphql/mutations/submit_form.rb
@@ -25,11 +25,11 @@ module Mutations
       # Look up form definition
       definition = Hmis::Form::Definition.find_by(id: input.form_definition_id)
       raise HmisErrors::ApiError, 'Form Definition not found' unless definition.present?
-      raise HmisErrors::ApiError, 'FormDefinition status is invalid' unless definition.valid_status_for_submit?
+      raise HmisErrors::ApiError, "FormDefinition #{definition.id} status #{definition.status} is invalid" unless definition.valid_status_for_submit?
 
       # Determine record class
       klass = definition.owner_class
-      raise HmisErrors::ApiError, 'Form Definition not configured' unless klass.present?
+      raise HmisErrors::ApiError, "Form Definition #{definition.id} not configured" unless klass.present?
 
       # Find or create record
       if input.record_id.present?
@@ -44,7 +44,6 @@ module Mutations
       raise HmisErrors::ApiError, 'Record not found' unless record.present?
 
       # Check permission
-      allowed = nil
       perms_to_check = definition.record_editing_permissions
       if definition.allowed_proc.present?
         allowed = definition.allowed_proc.call(entity_for_permissions, current_user)
--- a/drivers/hmis/app/graphql/types/hmis_schema/custom_case_note.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/custom_case_note.rb
@@ -17,6 +17,7 @@ module Types
     field :client, HmisSchema::Client, null: false
     field :content, String, null: false
     field :information_date, GraphQL::Types::ISO8601Date, null: true
+    field :form_definition_id, ID, null: false
     custom_data_elements_field
 
     def enrollment
@@ -25,6 +26,10 @@ module Types
 
     def client
       load_ar_association(object, :client)
+    end
+
+    def form_definition_id
+      load_ar_association(object, :form_processor)&.definition_id
     end
   end
 end
--- a/drivers/hmis/app/models/hmis/form/instance.rb
+++ b/drivers/hmis/app/models/hmis/form/instance.rb
@@ -10,7 +10,9 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
   self.table_name = :hmis_form_instances
 
   belongs_to :entity, polymorphic: true, optional: true
-  belongs_to :definition, -> { where(status: 'published') }, foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
+  belongs_to :definition,
+             -> { order(Arel.sql("CASE WHEN status = 'published' THEN 0 WHEN status = 'draft' THEN 1 ELSE 2 END")) },
+             foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
 
   belongs_to :custom_service_category, optional: true, class_name: 'Hmis::Hud::CustomServiceCategory'
   belongs_to :custom_service_type, optional: true, class_name: 'Hmis::Hud::CustomServiceType'
@@ -39,7 +41,7 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
                      )
                    end
 
-  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role)) }
+  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role).and(fd_t[:status].eq(:published))) }
 
   # Find instances that are for a specific Project
   scope :for_project, ->(project_id) { for_projects.where(entity_id: project_id) }
--- a/drivers/hmis/app/models/hmis/form/instance.rb
+++ b/drivers/hmis/app/models/hmis/form/instance.rb
@@ -41,7 +41,9 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
                      )
                    end
 
-  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role).and(fd_t[:status].eq(:published))) }
+  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role)) }
+
+  scope :published, -> { joins(:definition).merge(Hmis::Form::Definition.published) }
 
   # Find instances that are for a specific Project
   scope :for_project, ->(project_id) { for_projects.where(entity_id: project_id) }
--- a/drivers/hmis/app/models/hmis/form/instance.rb
+++ b/drivers/hmis/app/models/hmis/form/instance.rb
@@ -11,16 +11,14 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
 
   belongs_to :entity, polymorphic: true, optional: true
 
-  # This relationship could be clearer as a has_many relationship, since the instance may have many definition records.
-  # However, there are many usages of this relation from before we implemented form versioning,
-  # so to reduce impact, for now we have just modified the scope to prioritize a published definition if it exists.
+  # This belongs_to relationship is a bit confusing since now form identifiers can have multiple versions.
+  # We should aim to gradually replace usages of :definition with the has_many :definitions relationship below,
+  # so that we're explicit about which statuses of definition (draft, published, retired) we accept in a given situation.
   belongs_to :definition,
              -> { order(Arel.sql("CASE WHEN status = 'published' THEN 0 WHEN status = 'draft' THEN 1 ELSE 2 END")) },
              foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
 
-  # Use published_definition when we want to filter out Instances on all non-published forms
-  belongs_to :published_definition, -> { published },
-             foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
+  has_many :definitions, primary_key: :definition_identifier, foreign_key: :identifier, class_name: 'Hmis::Form::Definition'
 
   belongs_to :custom_service_category, optional: true, class_name: 'Hmis::Hud::CustomServiceCategory'
   belongs_to :custom_service_type, optional: true, class_name: 'Hmis::Hud::CustomServiceType'
--- a/drivers/hmis/app/models/hmis/form/instance.rb
+++ b/drivers/hmis/app/models/hmis/form/instance.rb
@@ -10,7 +10,7 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
   self.table_name = :hmis_form_instances
 
   belongs_to :entity, polymorphic: true, optional: true
-  belongs_to :definition, foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
+  belongs_to :definition, -> { where(status: 'published') }, foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
 
   belongs_to :custom_service_category, optional: true, class_name: 'Hmis::Hud::CustomServiceCategory'
   belongs_to :custom_service_type, optional: true, class_name: 'Hmis::Hud::CustomServiceType'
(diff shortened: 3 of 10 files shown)


drivers/hmis/app/models/hmis/form/instance.rb
```diff
@@ -39,7 +41,7 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
                      )
                    end
 
-  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role)) }
+  scope :with_role, ->(role) { joins(:definition).where(fd_t[:role].in(role).and(fd_t[:status].eq(:published))) }
```
martha: I'm not sure this is the right place for this change; we can also move it onto the client detail forms query if we are worried about broader impact. But I think if we don't put it here we might see this bug crop up again (where the scope returns multiple duplicate Instances because of the join to Definition)

drivers/hmis/app/graphql/mutations/submit_form.rb
```diff
@@ -44,7 +44,6 @@ def _resolve(input:, record_lock_version: nil)
       raise HmisErrors::ApiError, 'Record not found' unless record.present?
 
       # Check permission
-      allowed = nil
```
gigxz, line 47: why is this removed?

martha: oh, because my IDE was telling me "Local variable 'allowed' is not used after assignment". but I am happy to re-add it.

gigxz: ohh k.. no that's true

martha:
I retargeted this at 137 since it's new work, but let me know if you think it should stay targeted at 136 since it's a client bugfix

drivers/hmis/app/models/hmis/form/instance.rb
gigxz: Yeah this makes me nervous. There are 21 usages of this scope, so we potentially impacting a lot of behavior. I would feel more comfortable auditing them each and making changes one-by-one. I can already see there are a few places where we use this scope and want it to return instances regardless of whether the form has a published version or not. (For example `form_rules` on query_type, and `assessment_names_for_project` which is supposed to include inactive/retired since its just used for filtering existing assessments).

So, I think I would prefer to add a scope `scope :published, ->(role) { joins(:definition).merge(Hmis::Form::Definition.published) }` and call it from the client detail forms query.

It would be wise to audit other places we use the instance though.. I can see a few places we might want to add the published scope, like possibly `external_form_types_for_project`. Lets split that work out from this bug fix, though.

drivers/hmis/app/models/hmis/form/instance.rb
```diff
@@ -10,7 +10,9 @@ class Hmis::Form::Instance < ::GrdaWarehouseBase
   self.table_name = :hmis_form_instances
 
   belongs_to :entity, polymorphic: true, optional: true
-  belongs_to :definition, foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
+  belongs_to :definition,
+             -> { order(Arel.sql("CASE WHEN status = 'published' THEN 0 WHEN status = 'draft' THEN 1 ELSE 2 END")) },
+             foreign_key: :definition_identifier, primary_key: :identifier, class_name: 'Hmis::Form::Definition'
```
gigxz, line 19: Is there any reason we shouldn't just change this to a `has_many`, since the instance may have multiple definition records? That feels clearer to me. I am only finding 2 usages of this:
* DeleteFormRule (can just use `definitions.first`)
* UpdateFormRule (deprecated mutation, but same deal)

drivers/hmis/app/graphql/types/hmis_schema/custom_case_note.rb
```diff
@@ -17,6 +17,7 @@ class HmisSchema::CustomCaseNote < Types::BaseObject
     field :client, HmisSchema::Client, null: false
     field :content, String, null: false
     field :information_date, GraphQL::Types::ISO8601Date, null: true
+    field :form_definition_id, ID, null: false
```
gigxz, line 20: I think we will want to add this field to all types of records that have Form Processors tied to them, to support https://github.com/open-path/Green-River/issues/6113. lets discuss that issue again next week

drivers/hmis/app/models/hmis/form/instance.rb
martha: Hm, yeah I agree the `has_many` relationship makes a lot more sense. But I think there might be a few more usages we need to check, like in `Project.available_service_types`, scopes internal to `Instance` and also `Definition`, `enabled_features` on client. For each of those places where we are joining from Instance to Definition, if we change from `belongs_to` to `has_many`, we have to make sure that the scope doesn't then contain duplicate Instances, correct?

martha: I'll give it a go for sure, but I'm also wondering if now is the time to separate out the two unrelated bugfixes in this PR so the case notes form definition ID fix isn't blocked by this.

drivers/hmis/app/models/hmis/form/instance.rb
martha: Okay, that sounds good to me. I can make a followup ticket.

martha: Created https://github.com/open-path/Green-River/issues/6793

drivers/hmis/app/models/hmis/form/instance.rb
gigxz: Ahh okay, yes. I didn't think about the places where we join instances on the `:definition` relation. I think your approach is probably best (and least disruptive), then.

martha: Thanks 🙏 I'll add some comments

drivers/hmis/app/models/hmis/form/instance.rb
gigxz: Thanks!

drivers/hmis/app/models/hmis/form/instance.rb
gigxz: And, I'm sorry, I feel like I had this but then I lost it. How does this specific change affect the Client Details or Case Notes bug?

martha: Yes so, without this change, the `definition` on the Instance points to the first definition with this identifier, which isn't necessarily the published one (could be draft or, more likely, retired). `HmisSchema::OccurrencePointForm` then gets returned to the frontend with a `definition` that is not the currently published version, and that's the definition used for pathway editing.

You can see the bug in action by reverting the change locally and then running `client_detail_forms_spec`.

gigxz: Okay I see, thanks.  I think the desired state is that `project { occurrencePointForms }` and `clientDetailForms` should _only_ resolve forms that are published. (There is no use-case for "retired" on those, from what I can think up right now).

So this does seem to accomplish that, because you added the `published` scope, and updated this `definition` relation to _prefer_ the publish form, which results in OccurrencePointForm's `load_ar_association(object, :definition)` loading the correct form. Am I following that correctly?

I wonder if we can make things a bit clearer & safer in the OccurrencePointForm object resolver. We know that we only ever want to resolve a published FormDefinition here, so we can enforce that:

```ruby
definition = load_ar_association(object, :definitions).select(&:published?) # uses has_many relation
raise 'bad' unless definition

# or

definition = load_ar_association(object, :published_definition) # uses belongs_to :published_definition, -> { published }
raise 'bad' unless definition

# or, with your sorted relation

definition = load_ar_association(object, :definition) # this is sorted to prefer published definitions
raise 'bad' unless definition.published?
```


Does that make sense?

martha: @gigxz yes your summary of the bug/fix matches my understanding, and yes what you wrote does make sense. I think I prefer one of your first two suggestions, even though either would add some complexity with an extra relation on the Instance model. An admin hypothetically could add an Instance on a Client Detail form that is still a Draft, and then option 3 would cause an uncaught exception, whereas option 1 or 2 would simply exclude that form from the list, I think.

gigxz: All 3 could potentially raise if we forget to filter to published instances only, I updated my example. I guess I lean towards adding the `has_many :definitions` relation and going that route, since it is descriptive of the actual db relationship. I don't see that as extra complexity, since it is the more accurate relation. We may want to use that one going forward for other uses. (In addition to your change in the `belongs_to :definition` which I think is wise. That relation feels problematic in general, but at least its better if its preferring the published one..).

Here I was thinking we crossed all our Ts with the versioned form definition support 😅

gigxz: Now I see you added published_definition -also happy to go with that!

martha: Ah yeah it's a good point about the has many relationship. We should create it now and gradually transition to using it instead of the `belongs_to :definition`.

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
```diff
@@ -14,7 +14,7 @@ class Hmis::Hud::CustomServiceCategory < Hmis::Hud::Base
   belongs_to :user, **hmis_relation(:UserID, 'User'), optional: true
   has_many :service_types, class_name: 'Hmis::Hud::CustomServiceType'
   has_many :form_instances, class_name: 'Hmis::Form::Instance'
-  has_many :definitions, through: :form_instances
+  has_many :definitions, through: :form_instances, source: :definition # todo @martha - `definitions` is probably clearer, but `definition` would be the one that doesn't change behavior
```
gigxz: Ah interesting, I'm curious which test caught this?

martha: [drivers/hmis/spec/requests/hmis/service_category_spec.rb](https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/drivers/hmis/spec/requests/hmis/service_category_spec.rb#L73-L80)

```
Hmis::GraphqlController when the user has access should successfully create a service category
     Failure/Error: expect(response.status).to eq(200), result.inspect
       {"errors"=>[{"message"=>"Ambiguous source reflection for through association. Please specify a :source directive on your declaration like:\n\n  class Hmis::Hud::CustomServiceCategory < ActiveRecord::Base\n    has_many :definitions, {:through=>:form_instances, :source=>\"definition\"}\n  end", "backtrace"=>["/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/reflection.rb:930:in `source_reflection_name'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/reflection.rb:822:in `source_reflection'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/reflection.rb:963:in `check_validity!'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/associations/association.rb:41:in `initialize'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/associations/has_many_through_association.rb:10:in `initialize'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/associations.rb:304:in `new'", "/bundle/ruby/3.1.0/gems/activerecord-7.0.8.4/lib/active_record/associations.rb:304:in `association'", "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:24:in `block (2 levels) in fetch'", "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:23:in `each'", "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:23:in `block in fetch'", "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:22:in `each'", "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:22:in `fetch'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/dataloader/source.rb:136:in `run_pending_keys'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/dataloader.rb:271:in `each'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/dataloader.rb:271:in `block in spawn_source_fiber'", "/bundle/ruby/3.1.0/gems/graphql-2.3.1/lib/graphql/dataloader.rb:233:in `block in spawn_fiber'"]}], "data"=>{}}
     # ./drivers/hmis/spec/requests/hmis/service_category_spec.rb:75:in `block (3 levels) in <top (required)>'
     # /bundle/ruby/3.1.0/gems/webmock-3.23.0/lib/webmock/rspec.rb:39:in `block (2 levels) in <main>'
```

[drivers/hmis/spec/requests/hmis/project_services_spec.rb](https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/drivers/hmis/spec/requests/hmis/project_services_spec.rb#L112-L127) - both tests in here also fail with a similar error.

gigxz: Oh I see, thanks. Its just from loading the model. I looked back at [when this relation was added](https://github.com/greenriver/hmis-warehouse/commit/06ec5c2b275fe415474a9c197ed35ccbe07ab701#diff-0694f7e51774a198ad58178bd4c2d8e780a4bdf3d7a25f7a9bc18ed810c36482). Its used to resolve the list of forms that are used for a given service type/category, here: https://github.com/greenriver/hmis-warehouse/blob/652e046dbd37ed7f54602c7db023ccd988ce218b/drivers/hmis/app/graphql/types/hmis_schema/service_type.rb#L42-L48


I think we probably want to add a similar `select(&:published?)` to that at the end, so we're not repeating the form definition for each version that it has (And not including draft/retired forms). (Maybe there is a case to show retired ones, but I'm not super concerned about what is being resolved there since it's only informational for the config tool service page, which needs TLC anyway..)

martha: Oh, makes sense, thanks for digging into it!

gigxz (approved):
thanks @martha!!

PR #4798, opened 2024-10-04
Add viewable_by scope to ProjectConfig

martha:
## _Merging this PR_

## Description
I encountered [this error](https://greenriver.slack.com/archives/C0610D9HSSH/p1728058451516149) while QAing the recent change to the [GraphQL object-level authorization](https://github.com/open-path/Green-River/issues/6758). The problem is that `ProjectConfigs` don't have a `viewable_by` scope. Previously they would have just all been returned, but since we added the data source ID check to the GraphQL permission checker, now it is surfacing as an uncaught error.

I know this isn't a complete fix because project configs are not associated to data source IDs directly and they should be (https://github.com/open-path/Green-River/issues/6691).

To reproduce, visit https://qa-hmis-2.openpath.host/admin/project-configs

## Type of change
- Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/project_config.rb
+++ b/drivers/hmis/app/models/hmis/project_config.rb
@@ -28,7 +28,7 @@ class Hmis::ProjectConfig < Hmis::HmisBase
   scope :viewable_by, ->(user) do
     # Special case this, rather than using ProjectRelated concern, because project isn't a direct relationship.
     # Project config can apply to a project, an organization, or a project type.
-    # Right now we have no way to gate viewability for ProjectConfigs defined by project type - TODO(#6691)
+    # Right now we have no way to gate viewability by data source for ProjectConfigs defined by project type - TODO(#6691)
     project_ids = Hmis::Hud::Project.viewable_by(user).pluck(:id)
     organization_ids = Hmis::Hud::Organization.viewable_by(user).pluck(:id)
     where(project_id: project_ids).
--- a/drivers/hmis/app/models/hmis/project_config.rb
+++ b/drivers/hmis/app/models/hmis/project_config.rb
@@ -25,6 +25,17 @@ class Hmis::ProjectConfig < Hmis::HmisBase
 
   validate :validate_config_options_json
 
+  scope :viewable_by, ->(user) do
+    # Special case this, rather than using ProjectRelated concern, because project isn't a direct relationship.
+    # Project config can apply to a project, an organization, or a project type.
+    # Right now we have no way to gate viewability for ProjectConfigs defined by project type - TODO(#6691)
+    project_ids = Hmis::Hud::Project.viewable_by(user).pluck(:id)
+    organization_ids = Hmis::Hud::Organization.viewable_by(user).pluck(:id)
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/models/hmis/project_config.rb
```diff
@@ -25,6 +25,17 @@ def exactly_one_of_project_org_type
 
   validate :validate_config_options_json
 
+  scope :viewable_by, ->(user) do
+    # Special case this, rather than using ProjectRelated concern, because project isn't a direct relationship.
+    # Project config can apply to a project, an organization, or a project type.
+    # Right now we have no way to gate viewability by data source for ProjectConfigs defined by project type - TODO(#6691)
+    project_ids = Hmis::Hud::Project.viewable_by(user).pluck(:id)
```
martha, line 32: I suspect there is a better more efficient way to structure this query. My rails chops weren't quite up to the task yet so I'd be glad for pointers!

drivers/hmis/spec/requests/hmis/project_config_spec.rb
```diff
@@ -47,28 +47,51 @@
...
-      mutation_input = { config_type: 'AUTO_ENTER', project_type: 'ES_ENTRY_EXIT' }
-      response, result = post_graphql(input: mutation_input) { create_project_config }
-      expect(response.status).to eq(200), result.inspect
-      project_config_id = result.dig('data', 'createProjectConfig', 'projectConfig', 'id')
-      expect(project_config_id).not_to be_nil
-      project_config = Hmis::ProjectConfig.find(project_config_id)
-      expect(project_config.class.name).to eq('Hmis::ProjectAutoEnterConfig')
-      expect(project_config.project_type).to eq(0)
-    end
+  it 'should successfully create a new auto enter config' do
```
martha, line 66: These first 2 tests are substantively unchanged. I just restructured them by making use of `remove_permissions`. This way we can share `access_control` at the top level across all 3 tests. The 3rd test beginning on line 83 is the new test case added in this PR.

drivers/hmis/app/models/hmis/project_config.rb
gigxz: This seems alright as an intermediate fix for the error. As you point out, it doesn't give the ability to distinguish project configs by project type. We will need https://github.com/open-path/Green-River/issues/6691 for that.

I also don't know that we need to limit project config access at such a granular level. (Eg allow a setup where a user can set project configs at project X but not project Y in the same DS). But, maybe- I could see that being a future request for "site managers" etc to manage their own auto-exit rules.

If we do want to keep things global, once we do add `data_source_id`, we could revert to a simpler check:
```
scope :viewable_by, ->(user) do
  return none unless current_user.can_configure_data_collection?

  where(data_source_id: user.hmis_data_source_id)
end
```

gigxz:
I think its fine to put this in release-137. If we need to test project configs in the mean time, we can use other qa environments. (Since this bug only affects qa where we have 2 HMIS's)

gigxz (approved):

PR #4799, opened 2024-10-05
Authorization policy improvements

ttoomey:
## _Merging this PR_

### Improve policy for legacy (role-based) permissions
* add missing check to ensure client/project is in access groups (same pattern as for collections)
* remove legacy-auth-performed. The legacy policy no longer assumes the controller will check access
* attempt to fix legacy client visibility spec. Previously it checked client names but those are not visible due to PII perm

### Project Page
* [migrate project page to policy pattern (6785)](https://github.com/open-path/Green-River/issues/6785)

### ROI Authorizations
* per discussion with Elliot: create ROI authorizations table which is populated from client ROI fields. Prepare the way for clients with multiple ROI releases
* add ROI logic to client policy

### Other
* refactor/simplify auth policy structure for greater maintainability
* break the previous 1:1 parity between role and policy permissions
* added new controller authorization pattern, see example in ProjectController

## Type of change
Fix / enhancement / cleanup

  - [x] ensured this does not introduce (new) N+1s


app/models/access_group.rb
```diff
@@ -69,6 +69,12 @@ class AccessGroup < ApplicationRecord
     )
   end
 
+  # all access groups that include any coc_codes
+  scope :for_coc_codes, ->(coc_codes) do
+    quoted = SqlHelper.quote_sql_array(coc_codes, type: :varchar)
+    where("#{quoted_table_name}.coc_codes = #{quoted}")
```
eanders: Should this be `=` or something more like overlap? The comment says includes any, but I _think_ this is looking for equivalence.

app/models/grda_warehouse/auth_policies/legacy_user_role_policy.rb
```diff
@@ -6,16 +6,24 @@
...
 
   Role.permissions.each do |permission|
     method_name = :"#{permission}?"
     define_method method_name do
-      !!user.public_send(method_name)
+      # check if the user has permission on any role
+      return false unless user.public_send(method_name)
+
+      # check if the user in in any of the access groups
+      user.access_groups.where(id: @access_group_ids).exists?
```
eanders: As discussed, this probably needs to account for the various ROI setups.  Proposal is to generate a table of Client ROI that we can more easily union with the project-based permissions so we can accommodate the ROI setup.

app/models/grda_warehouse/project_access_group_member.rb
```diff
@@ -0,0 +1,13 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# DB view
```
eanders: @ttoomey can you add a bit more description here about what this view is doing, and maybe a link to the view migration? (Though I realize that might go away eventually)

app/models/access_group.rb
ttoomey: fixed, thanks

app/models/collection.rb
```diff
@@ -87,8 +87,9 @@ class Collection < ApplicationRecord
 
   # all collections that include any coc_codes
   scope :for_coc_codes, ->(coc_codes) do
-    q_codes = coc_codes.map { |code| connection.quote(code) }
-    where("#{quoted_table_name}.coc_codes ?| array[#{q_codes.join(',')}]")
+    quoted = SqlHelper.quote_sql_array(coc_codes, type: :varchar)
```
ttoomey, line 90: use array quote helper for consistency

app/controllers/application_controller.rb
```diff
@@ -41,6 +49,18 @@ class ApplicationController < ActionController::Base
 
   prepend_before_action :skip_timeout
 
+  # raise NotAuthorizedError which we can rescue from. This stops flow on a failed authorization check
```
ttoomey, line 52: moved these auth methods from ControllerAuthorization concern due to conditional inclusion

db/warehouse/migrate/20241016010729_create_client_roi_authorizations.rb
```diff
@@ -0,0 +1,11 @@
+class CreateClientRoiAuthorizations < ActiveRecord::Migration[7.0]
+  def change
+    create_table :client_roi_authorizations do |t|
+      t.references :destination_client, null: false, index: { unique: true }
```
ttoomey, line 4: the unique index is handy for upserts but will need to change when multiple rois are added

lib/tasks/grda_warehouse.rake
```diff
@@ -338,6 +338,15 @@ namespace :grda_warehouse do
       Rails.logger.error(e.message)
     end
 
+    if DateTime.current.hour == 20
```
ttoomey, line 341: run once a day

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
```diff
@@ -97,10 +101,11 @@
         expect(response).to have_http_status(200)
       end
       it 'can see window source client, but not non-window source client' do
-        get client_path(both_destination_client)
+        get rollup_client_path(both_destination_client, partial: :demographics)
```
ttoomey, line 104: does this seem reasonable? We check the demographics rollup since I don't know that the main page has a list of source clients. The client names are not always visible due to permissions so check ids in a data attr.

docs/code_patterns_and_conventions.md
```diff
@@ -40,8 +40,30 @@ When creating new JavaScript assets, use esbuild in `app/javascripts` rather tha
 When authorizing an action on an individual record, for example a Client, use a resource policy.
 
 ```ruby
-policy = user.policies.for_client(client)
-not_authorized! unless policy.show?
+policy = user.policy_for(client, type: client)
+not_authorized! unless policy.can_view?
```
ttoomey, line 44: Design question, which seems better?
```ruby
policy.can_view?
# or
policy.view?
```

Other policy patterns I've used favor the latter, which is more succinct. However it is a further departure from our permission naming convention. Perhaps that's desirable though since the policy is no longer 1-1 with role; breaking the naming convention makes the lack of parity obvious.

app/controllers/projects_controller.rb
```diff
@@ -4,15 +4,18 @@
...
 
-class ProjectsController < ApplicationController
-  before_action :require_can_view_projects!
-  before_action :require_can_delete_projects_or_data_sources!, only: [:destroy]
-  before_action :require_can_edit_projects!, only: [:edit, :update]
+class ProjectsController < ApplicationControllerV2
+  include ArelHelper
+
   before_action :set_project, only: [:show, :update, :edit, :destroy]
+  authorize_with { project_policy.can_view? }
```
eanders, line 11: I think this is fine since we don't have an `index` page for projects currently, but `project_policy.can_view?` isn't the same as "am I allowed to view any projects?"  Do we maybe need a special case in `permission_granted_by_role?(permission)` when the resource id is not present?

app/models/grda_warehouse/auth_policies/client_policy.rb
```diff
@@ -0,0 +1,142 @@
...
+  #   - if the user has `can_view_client_enrollments_with_roi`, we grant `can_view_clients`
+  # - ROI does not confer additional permissions. Additional permissions are identical to clients without an ROI, such as via direct assignment
+  def permission_granted_through_roi?(permission)
+    if roi_authorizations.any? && permission_granted_by_role?(permission)
+      return true if roi_authorizations.any? { |a| a.matches_coc_codes?(user.coc_codes) }
+    end
+    false
+  end
+
+  memoize def client_collection_ids
```
eanders:
```suggestion
  # Collection IDs that contain this client through an enrollment at a project
  memoize def client_collection_ids
```

app/models/grda_warehouse/auth_policies/client_policy.rb
```diff
@@ -0,0 +1,142 @@
...
+      joins(:group_viewable_entities, :clients).
+      where(gve_t[:collection_id].not_eq(nil)).
+      where(c_t[:id].eq(client_id)).
+      pluck(gve_t[:collection_id])
+
+    collection_ids += system_collection_ids(:data_sources)
+    collection_ids.uniq.sort
+  end
+
+  memoize def client_access_group_ids
```
eanders:
```suggestion
  # Access Group IDs that contain a client through an enrollment at a project
  memoize def client_access_group_ids
```

app/models/grda_warehouse/auth_policies/client_policy.rb
```diff
@@ -0,0 +1,142 @@
...
+      permission_granted_by_role?(permission)
+    end
+    memoize method_name
+  end
+
+  # can the user see the full client dash page and additional details?
+  memoize def can_view?
+    return true if permission_granted_through_roi?(:can_view_client_enrollments_with_roi)
+
+    permission_granted_by_role?(:can_view_clients)
```
eanders: For performance, I wonder if the order of these should be flipped.  Maybe checking ROI is faster since that tends to be a smaller dataset and fewer installations.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -203,6 +203,10 @@ class Client < Base
     has_many :cas_reports, class_name: 'GrdaWarehouse::CasReport', inverse_of: :client
 
     has_many :chronics, class_name: 'GrdaWarehouse::Chronic', inverse_of: :client
+    has_many :roi_authorizations,
```
eanders, line 201: We should make sure these are included in the `hmis_dependent_items` method so that they get moved when a client is merged and not lost when the destination is deleted.

app/controllers/projects_controller.rb
ttoomey: It's a good question with this pattern. My feeling is that can_index  doesn't fit well on a resource policy since it's a permission on a collection, not a single resource.

Often there is another entity that is a natural fit for a collection permission. For example on an enrollment index page, you'd probably have a project so you would check project_policy.can_view_enrollments

In this case, with projects#index, we just want to know if the user can just see the index page itself. I would say there's an implicit global 'site' or 'warehouse' singleton.resource you could ask. I might add a new global WarehousePolicy for those permissions. Then the checks would be something like this:
```ruby
  authorize_with { global_policy.can_view_projects? }, only: :index # equiv. to require_can_view_projects!
  authorize_with { project_policy.can_view? }, only: [:show, :update, :edit, :destroy]

  # or maybe if one global policy is too much, you could have a policy for each, for example a ProjectDashboardPolicy. Not sure...
  authorize_with { project_dashboard_policy.can_index? }

```

app/models/grda_warehouse/auth_policies/data_source_policy.rb
```diff
@@ -0,0 +1,78 @@
...
+  memoize def data_source
+    resource_from_arg(resource, GrdaWarehouse::DataSource)
+  end
+
+  memoize def data_source_id
+    id_from_arg(resource, GrdaWarehouse::DataSource)
+  end
+
+  def permission_granted_by_role?(permission)
+    if user.using_acls?
```
eanders: This looks like duplicate `permission_granted_by_role?` methods.  I think the method on line 25 is correct since it moves the permission check outside of the `if` statement for performance.

app/controllers/projects_controller.rb
eanders: Yeah, I keep trying to conflate the two.  Let's leave it as is for now, as a scope on whatever model is appropriate, but I think I like the idea of a warehouse authorization (or global authorization singleton) for these in the future.

app/models/grda_warehouse/auth_policies/project_policy.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    if user.using_acls?
+      user.access_controls.joins(:role).
+        where(collection_id: project_collection_ids).
+        merge(Role.where(permission => true)).any?
+    else
+      # check if the user is in any of the access groups
+      user.access_groups.where(id: project_access_group_ids).exists?
+    end
+  end
```
eanders: Might be best to leave this where it is until we've developed more of these policies, but I think this could be re-written and moved to the base class.  If there were `related_collection_ids` and `related_access_group_ids` methods this could be golfed to:
```suggestion
  def permission_granted_by_role?(permission)
    # early return unless the user has this permission on any role. For acl users, this is performance optimization
    return false unless user.public_send("#{permission}?")

    if user.using_acls?
      user.access_controls.joins(:role).
        where(collection_id: related_collection_ids).
        merge(Role.where(permission => true)).any?
    else
      # check if the user is in any of the access groups
      user.access_groups.where(id: related_access_group_ids).exists?
    end
  end
```

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
```diff
@@ -0,0 +1,101 @@
...
+      end
+    end
+
+    protected
+
+    # we only consider building ROI authorizations for destination clients, excluding data sources that do not obey consent
+    def destination_client_scope
+      GrdaWarehouse::Hud::Client.destination.
+        joins(:data_source).
+        merge(GrdaWarehouse::DataSource.obeys_consent)
```
eanders: This doesn't make sense, the `obeys_scope` only applies to source data sources, which would never have a destination client.  Destination clients will only exist in the "warehouse" data source.

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
```diff
@@ -0,0 +1,101 @@
...
+        client.consent_expires_on
+      when 'Indefinite'
+        nil
+      else
+        raise "unknown release duration \"#{duration}\""
+      end
+    end
+
+    def roi_coc_codes(client)
+      client.consented_coc_codes&.uniq&.sort&.presence
```
eanders: should this reject blank?

app/views/clients/_demographic_row.haml
```diff
@@ -1,5 +1,5 @@
 - pii = client.pii_provider(user: current_user)
-%tr
+%tr{'data-test-client-id' => client.id}
```
eanders, line 3: Was this for testing?  If not, maybe we should remove `test` from the name? 
I think including the source client id in the view is a great idea.

db/warehouse/migrate/20241016010729_create_client_roi_authorizations.rb
eanders: I was going to suggest we add an `roi_type` column and make the unique index on `destination_client_id, roi_type` and make `roi_type` something like `warehouse_overall`, but thinking this through, we may end up with historic records as well (one ROI covering 2022, one covering 2023, revocation in 2024, etc.)  Maybe this should just be the "current ROI state" for each client.  Removing the index is easy later on, but I worry a bit about baking that in for the import now and then having to tear it out and refactor the import significantly later.

docs/code_patterns_and_conventions.md
eanders: No strong preference from me.  I like how `prolicy.can_view?` reads, but I think we could all make the leap as well.

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
```diff
@@ -11,6 +11,10 @@
 RSpec.describe ClientAccessControl::ClientsController, type: :request do
   include_context 'visibility test context'
 
+  def doc_has_client?(doc, warehouse_client)
+    doc.css("[data-test-client-id='#{warehouse_client.source_id}']").any?
```
eanders, line 15: oh, I see, `test` is for testing....  I guess I still think we should just call it `data-client-id`.

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
eanders: Yeah, I think this is a good solution, thanks!

spec/controllers/controller_authorization_spec.rb
```diff
@@ -0,0 +1,62 @@
...
+    get :authorized_action
+    expect(response).to have_http_status(:success)
+  end
+
+  it 'raises when authorization is not performed' do
+    expect { get :unauthorized_action }.to raise_error(AuthorizationNotPerformedError)
+  end
+
+  it 'does not have access to legacy authorization methods' do
+    expect(@controller.respond_to?(:require_can_view_projects)).to be false
```
eanders: Should this be:
```suggestion
    expect(@controller.respond_to?(:require_can_view_projects!)).to be false
```

app/models/grda_warehouse/auth_policies/client_policy.rb
ttoomey: ROI should be faster; usually putting faster checks first is optimal but if ROI permission is rare, leading with the most common condition makes sense

app/models/grda_warehouse/auth_policies/data_source_policy.rb
ttoomey: thanks! I missed the rubocop alert on that dupe

app/models/grda_warehouse/auth_policies/project_policy.rb
ttoomey: Yes, extracting the common pieces would be good once we have a few more in place. Perhaps there's a mixin for policies that share this structure.

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
ttoomey: yeah, naming it data-test-client-id is supposed to let folks know it's only for the test, not actual functionality. I'll add a comment to the haml view

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
ttoomey: thanks for clarifying! I've adjusted this and the roi check on the client policy. Let's huddle about this to confirm the logic

app/models/grda_warehouse/auth_policies/data_source_policy.rb
```diff
@@ -0,0 +1,39 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class GrdaWarehouse::AuthPolicies::DataSourcePolicy < GrdaWarehouse::AuthPolicies::BasePolicy
```
ttoomey, line 7: note, the data source policy is not actually used yet

app/models/grda_warehouse/project_group.rb
```diff
@@ -16,7 +16,8 @@ class ProjectGroup < GrdaWarehouseBase
     attr_accessor :editor_ids
 
     validates_presence_of :name
-    after_create :maintain_system_group
+    attr_accessor :skip_maintain_system_group
+    after_create :maintain_system_group, unless: :skip_maintain_system_group
```
ttoomey, line 21: maintain_system_group  seems to take 30 seconds

app/models/grda_warehouse/auth_policies/project_policy.rb
ttoomey: I've extracted the common pieces out into a policy context which has legacy and acl variants

db/warehouse/migrate/20241016010729_create_client_roi_authorizations.rb
ttoomey: The importer just does an upsert/delete so there shouldn't be too much to tear out. I guess it's not clear to me that we will even want a dedicated batch importer like this if the roi authorizations replace the client roi fields.

app/models/grda_warehouse/auth_policies/project_pii_policy.rb
```diff
@@ -0,0 +1,37 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# FIXME: in a report context we may only have a project id to determine PII. Using just the project id is efficient but
+# the # permissions not are entirely accurate in every case. We are not including permissions granted by all the
+# client's enrollments or direct client-record visibility
```
eanders: Would the fix for this be to convert usage to a ProjectPolicy or ClientPolicy instead?  Basically batch fetch the actual resources, index on `id` and then act on the appropriate resource instead of the id?

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -0,0 +1,75 @@
...
+  memoize def role_permissions
+    clients = GrdaWarehouse::Hud::Client.where(id: client_id)
+    project_ids = GrdaWarehouse::Hud::Project.joins(:clients).merge(clients).distinct.pluck(:id)
+
+    results = Set.new
+    project_ids.each do |project_id|
+      results.merge(context.project_role_permissions(project_id))
+    end
+    results.merge(context.direct_client_role_permissions(client_id))
+  end
```
eanders: Can you add some documentation here, something like:
* `project_ids` are the projects for which the client has an enrollment
* `results` is a set of permissions the user has for either the project or the client which would grant them access to this client within a permission context
(does that capture it at all?)
I think the method name `role_permissions` is still confusing to me since those are not **role-based** permissions, but just permissions you you have within an appropriate context, using Access Controls when appropriate and roles when appropriate.

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -0,0 +1,75 @@
...
+
+  # An ROI confers some level of visibility to the client under the following circumstances:
+  # - the source client must be in a data source with `obeys_consent=true`
+  # - the ROI has a valid status (not revoked). ROI fields are stored on the destination client record (for now)
+  # - if the ROI is restricted to certain COCs then the user's COCs must match
+  # - the user has a role granting permission on source client project as follows:
+  #   - if the user has `can_search_client_with_roi`, we grant `can_search_own_clients` and `can_search_all_clients`
+  #   - if the user has `can_view_client_enrollments_with_roi`, we grant `can_view_clients`
+  # - ROI does not confer additional permissions. Additional permissions are identical to clients without an ROI, such as via direct assignment
+  def roi_authorized?
```
eanders, line 69:
```suggestion
  # NOTE: this will query `destination_client.roi_authorizations` which will cause additional queries, if you are calling this in a loop, you'll want to `preload(destination_client: :roi_authorizations)`
  def roi_authorized?
```

app/models/grda_warehouse/auth_policies/user_acl_context.rb
```diff
@@ -0,0 +1,85 @@
...
+    collection_ids = GrdaWarehouse::ProjectCollectionMember.
+      where(project_id: project_id).
+      pluck(:collection_id)
+
+    # collections for the projects via coc_codes
+    coc_codes = GrdaWarehouse::Hud::ProjectCoc.
+      joins(:project).
+      where(p_t[:id].eq(project_id)).
+      pluck(:coc_code)
+    collection_ids += Collection.for_coc_codes(coc_codes).pluck(:id) if coc_codes.any?
```
eanders: Do you think it's worth trying to write the union for `ProjectCollectionMember` that would include access granted by coc_code?  I'm concerned about performance if we add it, but also that we may forget this extra check/query if we don't.

app/models/grda_warehouse/auth_policies/user_legacy_context.rb
```diff
@@ -0,0 +1,89 @@
...
+
+require 'memery'
+
+# cross-policy memoized utils for Legacy user-role-based permissions
+class GrdaWarehouse::AuthPolicies::UserLegacyContext
+  include Memery
+  attr_accessor :user
+
+  def initialize(user)
+    @user = user
```
eanders, line 15: Should this raise if `user.using_acls?`?

app/models/grda_warehouse/project_group.rb
eanders: Maybe a comment that this is only skipped in tests? 
We definitely need to maintain the system groups in non-test environments.
Oddly `maintain_system_group` doesn't seem to call the collection equivalent.  Pretty sure that should be:
```
private def maintain_system_group
  AccessGroup.delayed_system_group_maintenance(group: :project_groups)
  Collection.delayed_system_group_maintenance(group: :project_groups)
end
```

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
```diff
@@ -0,0 +1,100 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# Build ROI values from the client. We change ROI authorization to the canonical record in the future, instead of
```
eanders:
```suggestion
# Build ROI values from the client. If and when we change ROI authorization to the canonical record in the future, instead of
```

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
```diff
@@ -0,0 +1,100 @@
...
+# generating it from the client. At that point, this task can be removed
+module GrdaWarehouse::Tasks
+  class GenerateClientRoiAuthorizationsTask
+    def self.perform(...)
+      new.perform(...)
+    end
+
+    # @param client_ids only rebuild
```
eanders: Should this be something more like:
`# @param client_ids [Array<Integer>, nil] client ids of the clients who will have their ROI authorizations rebuilt, pass nil to rebuild all clients`

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
```diff
@@ -0,0 +1,100 @@
...
+        expires_at: roi_expiry_date(destination_client),
+        # maybe add file, other fields
+      }
+    end
+
+    def roi_expiry_date(client)
+      duration = GrdaWarehouse::Hud::Client.release_duration
+      case GrdaWarehouse::Hud::Client.release_duration
+      when 'One Year', 'Two Years'
+        return nil unless client.consent_form_signed_on
```
eanders: This is interesting because it means the client didn't sign the consent, but we determined it was valid, and therefore we are granting an indefinite consent.  This would probably arise if the client's consent was uploaded before the release duration was set or if the configuration wasn't correctly enforcing collection of a signature date.
I don't have a better solution, but it might be worth calling out here.   

What would you think about returning `Date.current + 1.years` or `Date.current + 2.years`?  

That might be misleading since it'll be extended every day...  Maybe we should count from the release form `created_at` date if we have one in this situation....

app/models/role.rb
```diff
@@ -192,6 +192,12 @@ def self.title_for_column(column)
     column.to_s.humanize
   end
 
+  def permissions
```
eanders: Is this returning an array of "enabled" permission names?  Maybe this should be named "active_permissions" or "enabled_permissions."

drivers/core_demographics_report/app/models/core_demographics_report/details_column.rb
```diff
@@ -5,7 +5,7 @@ def value(row)
       raw_value = row[index]
 
       project_id = row[project_id_index]
-      policy = user.policies.for_project(project_id)
+      policy = user.policy_for(project_id, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy)
```
eanders, line 8: Is the reason we can't use the client policy here is because we don't have client objects correct?  I wonder if we should pivot, load all the related clients and index them by id so we can use the appropriate policy.  Potentially that would require storing tens of thousands of client objects in memory, so maybe we can't. 

Looking back through, I think using the client policy may also pull down all of the project ids related to the client, which would be additional queries.    I also think this isn't wrong, it just ignores ROI correct?  So if a client is included here but you can't view their client name based on the project enrollment, but you _could_ if we checked ROI, it will be hidden.  If that's the only side effect, I think this is probably fine as is.

spec/models/auth_policies/project_pii_policy.rb
```diff
@@ -0,0 +1,141 @@
...
+        create(:access_control, role: role, collection: collection, user_group: user_group)
+      end
+      include_examples 'pii permission checks', true
+    end
+
+    context 'without any access' do
+      include_examples 'pii permission checks', false
+    end
+
+    context 'with multiple roles' do
```
eanders: I wonder if we should have a set of tests that confirms access to PII via one project, but denial to another when you have two access controls.  So similar to below, but with different projects in collection 1 and collection 2 (and different policies as a result.)

spec/models/auth_policies/project_policy_spec.rb
```diff
@@ -0,0 +1,242 @@
...
+        project.save!
+        group
+      end
+
+      before { access_group.add_viewable(project_group) }
+      include_examples 'permission checks', true
+    end
+
+    context 'with CoC code access' do
+      let(:coc_code) { 'authtest1' }
```
eanders: It's probably fine to include a non-valid CoC Code, but at various points we've enforced CoC-Code validity, which might cause this test to fail unnecessarily in the future.  To be sure it won't get cleaned up through CoC code validation, you can use `XX-500` or `XX-501`.

spec/models/auth_policies/source_client_policy_spec.rb
```diff
@@ -0,0 +1,133 @@
...
+      before(:each) do
+        create(:warehouse_client, source: client, data_source: data_source)
+        create(:client_roi_authorization, destination_client: client.destination_client)
+        role.update!(can_search_clients_with_roi: true, can_view_client_enrollments_with_roi: true)
+      end
+
+      it 'grants view and search permissions through ROI' do
+        expect(policy.can_view?).to be true
+        # expect(policy.can_search?).to be true
+      end
```
eanders: Should we clean these up?  I think we decided to implement the search permission only through a scope at this point.

app/models/grda_warehouse/auth_policies/project_pii_policy.rb
ttoomey: Per our discussion, we can probably stick with this policy for now rather than trying to make the client policy efficient in a bulk context. 

However if we did want to support bulk permissions, we could adjust the query methods in the legacy and acl contexts to operate efficiently arrays of ids. The queries would be modified to return results grouped by id. Something like this: 
```ruby
# before
context.direct_client_role_permissions(client_id) 
  => Set([:can_do_foo, :can_do_bar])
# after
context.direct_client_role_permissions(client_ids) 
  => {1 => Set([:can_do_foo]), 2 => Set([:can_do_bar, :can_do foo])}
```

That would let us efficiently warm policies for many resources at once. Then we'd need to change the caller warm the policies. Not totally sure on the api but maybe something like this:
```ruby
# warm polices for all resources
user.prepare_policies(policy_class, resource_ids)

# the call this as normal
resource_ids.each do |id|
   policy = user.policy_for(id, batch)
   # ... 
end
```

app/models/grda_warehouse/auth_policies/source_client_policy.rb
ttoomey: will do, I'll rename role_permissions to role_permissions to be more clear

app/models/grda_warehouse/auth_policies/source_client_policy.rb
ttoomey: yes, there are other methods were this might be the case too. Ideally we wouldn't require the caller to preload the required associations but could provide a mechanism to warm the polices or otherwise batch load

app/models/grda_warehouse/auth_policies/user_acl_context.rb
ttoomey: per our discussion, I don't think it's possible to use one query given the database boundary between GVEs and collections/access groups

app/models/grda_warehouse/project_group.rb
ttoomey: thanks. Looks like we are missing collections from `Cohort@#maintain_system_group` too, I'll add it

app/models/grda_warehouse/tasks/generate_client_roi_authorizations_task.rb
ttoomey: per discussion, changing this to exclude ROI's that are missing dates if we are using validity periods

app/models/role.rb
ttoomey: sure, I'll rename it to make it more clear. Perhaps 'granted_permissions'

app/models/grda_warehouse/cohort.rb
```diff
@@ -719,6 +719,7 @@ def refresh_time_dependant_client_data(cohort_client_ids: nil)
 
     private def maintain_system_group
       AccessGroup.delayed_system_group_maintenance(group: :cohorts)
+      Collection.delayed_system_group_maintenance(group: :cohorts)
```
ttoomey, line 838: @eanders added to match project group method

spec/models/auth_policies/destination_client_policy_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+
+  shared_examples 'pii permission checks without access' do
+    context 'standard PII permissions' do
+      it 'denies all PII permissions when user lacks access' do
+        expect(policy.can_view_name?).to be false
+        expect(policy.can_view_photo?).to be false
+        expect(policy.can_view_full_dob?).to be false
+        expect(policy.can_view_full_ssn?).to be false
+        expect(policy.can_view_hiv_status?).to be false
+      end
```
eanders, line 46: Thanks @ttoomey, I think this is much more readable.

eanders (approved):

eanders:
Hmm, looks like we have some failing tests, not sure if those are related, can you check @ttoomey ?

ttoomey:
> Hmm, looks like we have some failing tests, not sure if those are related, can you check @ttoomey ?

I'll look into it, I couldn't replicate those failures locally so it's probably test leakage. I'll track it down

gigxz (approved):

spec/models/grda_warehouse/hud/client_spec.rb
```diff
@@ -35,11 +35,12 @@
     let!(:client) { create :grda_warehouse_hud_client }
 
     before(:all) do
-      @paper_trail_was = PaperTrail.enabled?
       PaperTrail.enabled = true
+      PaperTrail.request.enabled = true
```
ttoomey, line 39: These tests started failing. The fix was to set PaperTrail.request.enabled. I'm still unsure as to the underlying cause

ttoomey:
* fixed tests finally (remaining fail is unrelated rubocop check due to issue in release-141)

@eanders should I merge this to 141 still or wait until 142?

eanders:
141 should be fine.
merging

PR #4801, opened 2024-10-07
6237-fix-xss-in-alert

ttoomey:
## _Merging this PR_

## Description
[GH issue](https://github.com/open-path/Green-River/issues/6237)
Sanitize flash alert message. I'm guessing at what HTML elements we might need in the message.

## Type of change
 Bug fix


eanders (approved):
Thanks.  I was trying to think of additional tags we might get from a flash message.  Maybe `i` and maybe `img` but I doubt it.  I think we only use those in "alerts" that are persistent.
I think we should go with this.

PR #4805, opened 2024-10-08
Require disability at annual for hopwa hud compliance

gigxz:
## _Merging this PR_

## Description
https://github.com/open-path/Green-River/issues/6611

Make a change to require the `disability` fragment to be collected at Annual for all projects and installations. Depending on the configuration, this means the annual will either include a single "Disabling Condition" question, OR, the full "disability table" (physical, mental, etc).

**This fixes a HUD Compliance issue that was affecting some installations, where HOPWA data elements W4/W6 were not being collected at annual**. This approach also lets us remove some complexity we had added for W4 and W6. 

The QA instructions are [on the ticket](https://github.com/open-path/Green-River/issues/6611#issuecomment-2400603466). I tested each scenario locally by seeding the forms with `CLIENT=x rails driver:hmis:seed_definitions` and then viewing the HUD assessments for each project/funder category.

## Type of change
Config change


drivers/hmis/lib/form_data/allegheny/assessments/base_annual.json
```diff
@@ -1,19 +0,0 @@
...
-      "fragment": "#income_and_sources"
-    },
-    {
-      "fragment": "#non_cash_benefits"
-    },
-    {
-      "fragment": "#health_insurance"
-    },
-    {
-      "fragment": "#disability"
```
gigxz, line 16: We no longer need this override, because the point of it was to include disability at annual. Now we do that by default.

drivers/hmis/lib/form_data/springfield/assessments/base_annual.json
```diff
@@ -1,19 +0,0 @@
...
-      "fragment": "#income_and_sources"
-    },
-    {
-      "fragment": "#non_cash_benefits"
-    },
-    {
-      "fragment": "#health_insurance"
-    },
-    {
-      "fragment": "#disability"
```
gigxz, line 16: We no longer need this override, because the point of it was to include disability at annual. Now we do that by default.

martha (approved):
Looks good to me. Thank you for the thorough qa instructions.

PR #4809, opened 2024-10-08
Go through warehouse client for chronicity

martha:
## _Merging this PR_

## Description
GH issue https://github.com/open-path/Green-River/issues/6812
This PR fixes the bug outlined/confirmed in the issue.

## Type of change
Bug fix

  - [ ] ensured this does not introduce N+1s - Yes, no new N+1s are introduced. `hud_chronic` has an existing n+1 problem documented in comments, which is ok in the HMIS context since we only fetch it for 1 client at a time on the enrollment dashboard (?)

### DIFF
--- a/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
@@ -177,8 +177,8 @@ RSpec.describe Hmis::GraphqlController, type: :request do
   end
 
   context 'with a client who is chronically homeless per HUD definition' do
-    let!(:client) { create :hmis_hud_client_with_warehouse_client, data_source: ds1 }
-    let!(:enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: client, DisablingCondition: 1, MonthsHomelessThisTime: 13 }
+    let!(:c1) { create :hmis_hud_client_with_warehouse_client, data_source: ds1 }
+    let!(:e1) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, DisablingCondition: 1, MonthsHomelessThisTime: 13 }
 
     it 'should return chronic status correctly' do
       response, result = post_graphql(id: c1.id) { query }
--- a/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
@@ -185,8 +185,8 @@ RSpec.describe Hmis::GraphqlController, type: :request do
              project: p1,
              client: c1,
              DisablingCondition: 1,
-             MonthsHomelessThisTime: 13,
-             DateToStreetESSH: 13.months.ago
+             MonthsHomelessPastThreeYears: 112, # see MonthsHomelessPastThreeYears enum
+             TimesHomelessPastThreeYears: 4
     end
     let!(:disability) { create :hmis_disability, client: c1, enrollment: e1 }
 
--- a/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
@@ -174,11 +174,13 @@ RSpec.describe Hmis::GraphqlController, type: :request do
       'emailAddresses' => [{ 'value' => 'email@e.mail', 'use' => 'home', 'system' => 'email' }],
     )
     expect(result.dig('data', 'client', 'image')).not_to be_nil
+    expect(result.dig('data', 'client', 'hudChronic')).to eq(false)
   end
 
   context 'with a client who is chronically homeless per HUD definition' do
     let!(:c1) { create :hmis_hud_client_with_warehouse_client, data_source: ds1 }
     let!(:e1) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, DisablingCondition: 1, MonthsHomelessThisTime: 13 }
+    let!(:disability) { create :hmis_disability, client: c1, enrollment: e1 }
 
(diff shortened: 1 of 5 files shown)


drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -175,6 +176,55 @@
     expect(result.dig('data', 'client', 'image')).not_to be_nil
   end
 
+  context 'with a client who is chronically homeless per HUD definition' do
```
martha, line 180: @gigxz I'd like to pair with you on this to get a better understanding... I'm copying here from the hud chronic spec (https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/spec/models/grda_warehouse/hud_chronic_spec.rb#L4). I don't think we want to be invoking all these GrdaWarehouse factories for a hmis request spec test, but I don't know how to properly set up the models so that the warehouse client is considered chronic.

I am also stuck on doing this in my local (not spec) environment. I created a local client using the PIT and set their disability and months homeless values such that I would expect them to be chronic, but it's not the case.

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -175,6 +176,55 @@
...
+    let!(:warehouse_ds) { create :destination_data_source }
+    let!(:client) { create :grda_warehouse_hud_client, data_source_id: warehouse_ds.id }
```
gigxz: @martha we have a factory `hmis_hud_client_with_warehouse_client` that _should_ help with this setup. I believe it will generate the source client, warehouse client, and the destination client.

I think it just needs some tweaks, because it looks like it generates the destination client in the same ds as the source, which isn't right.  A potential modification for https://github.com/greenriver/hmis-warehouse/blob/7f1164546483039af1c9c479f1df81e10de817d5/drivers/hmis/spec/factories/hmis/hud/clients.rb#L60-L71

is:
```ruby
  # Client that is in the Warehouse destination data source (not the HMIS data source)
  factory :destination_client, parent: :hmis_hud_base_client do
    data_source { association :destination_data_source }
  end

  # WarehouseClient that links the source HMIS client to the destination client
  factory :hmis_warehouse_client, class: 'Hmis::WarehouseClient' do
    data_source { association :hmis_data_source }
    destination { association :destination_client }
    source { association :hmis_hud_base_client, data_source: data_source }
    sequence(:id_in_source, 100)
  end

  # HMIS Source Client that is linked to a destination client (Via WarehouseClient)
  factory :hmis_hud_client_with_warehouse_client, parent: :hmis_hud_base_client do
    after(:create) do |client|
      create(:hmis_warehouse_client, data_source: client.data_source, source: client)
    end
  end
```

If that works, I thiiiink you should be able to keep the setup simple here:
```
let!(:client) {  create :hmis_hud_client_with_warehouse_client, data_source: ds1 }
let!(:enrollment) {  create :hmis_hud_enrollment, data_source: ds1, project: p1, client: client }
```

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -175,6 +176,55 @@
...
+    let!(:project) do
```
gigxz: I don't think this is needed, the enrollment will create its own project if you don't pass one. Or you could use p1 which already exists.

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -175,6 +176,55 @@
...
+      create(
+        :warehouse_client,
+        destination: client,
+        source: source_client,
+        data_source_id: source_client.data_source_id,
+      )
+    end
+    let!(:source_enrollment) do
+      create(
+        :hud_enrollment,
```
gigxz: I think using the hmis factories is preferable

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -175,6 +176,55 @@
...
+    end
+
+    before(:each) do
+      Rails.cache.delete('chronically_disabled_clients')
+    end
+
+    it 'should return chronic status correctly' do
+      response, result = post_graphql(id: c1.id) { query }
+      expect(response.status).to eq 200
+      expect(result.dig('data', 'client', 'hudChronic')).to eq(true)
```
gigxz, line 196: This probably wants a companion case where the chronicity status is false, just to make sure its not always true

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
gigxz: Ok! happy to pair. I may have answered the first part of your question here https://github.com/greenriver/hmis-warehouse/pull/4809/files#r1792473983, I agree we should avoid the grda factories.

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
martha: Thanks! This is interesting, but I think that the factories need to be further modified so that the two clients have the same Personal Id? I will keep poking to see if I can find out how to do that in factory bot.

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
martha: Oof, okay, I got there in the end! My local PIT project had the wrong type.

gigxz: ohh tricky

gigxz (approved):
looks good, thanks!!

PR #4813, opened 2024-10-09
Validate form definition identifier

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6778#issuecomment-2400687969

This PR adds form definition identifier validation.

I've added the requirement that it should contain only alphanumeric characters, dashes, and underscores, and must start with a letter. But, please do let me know if you think it should be looser (allow any character to start? We have some examples of this in the qa db) or stricter (only underscores, no dashes, or vice versa?)

Screenshot 2024-10-09 at 5 43 27 PM

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/form/definition.rb
+++ b/drivers/hmis/app/models/hmis/form/definition.rb
@@ -40,6 +40,8 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   # convenience attr for passing graphql args
   attr_accessor :filter_context
 
+  validates :identifier, format: { with: /\A[a-zA-Z][a-zA-Z0-9_-]*\z/, message: 'must contain only alphanumeric characters, underscores, and dashes, and must start with a letter' }
+
   # --- Relations by id ----
   has_many :form_processors, dependent: :restrict_with_exception
   has_many :external_form_submissions, class_name: 'HmisExternalApis::ExternalForms::FormSubmission', dependent: :restrict_with_exception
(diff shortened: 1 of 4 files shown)


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -40,6 +40,8 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
   # convenience attr for passing graphql args
   attr_accessor :filter_context
 
+  validates :identifier, format: { with: /\A[a-zA-Z][a-zA-Z0-9_-]*\z/, message: 'must contain only alphanumeric characters, underscores, and dashes, and must start with a letter' }
```
ttoomey, line 43: This looks good but maybe we can restrict this further to make our lives easier:
* do we need to support dashes? Dashes are invalid in an identifier (variable name)
* perhaps we don't need mixed-case? Keeping everything lowercase is easier to handle

We can work around these of course with transformations or quoting. If we already have definitions with loosey-goosey identifiers it's probably not worth it to fix. But if not, I think it would be better if the identifiers were 1:1 with the reporting table names.

ttoomey (approved):
looks good, thanks!

drivers/hmis/app/models/hmis/form/definition.rb
martha: It looks like we do already have many (mostly?) loosey goosey identifiers, unfortunately.

(in client prod,
```
Hmis::Form::Definition
  .where("identifier ~* ?", '([A-Z]|-)') # capitals or dashes
  .pluck(:identifier)
  .uniq.count
=> 71
Hmis::Form::Definition.pluck(:identifier).uniq.count
=> 71
```
)

gigxz (approved):

PR #4814, opened 2024-10-10
Add published scope to form instance with_role usages

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6793

I searched for usages of the `FormInstance.with_role` scope to catch more places we need to add `.published` in order to return the correct set of forms. It's a followup to [this bugfix](https://github.com/greenriver/hmis-warehouse/pull/4793/files#diff-88f6bf8063dd5dfb242c415dbb41b8bd2d36fdb621d555ef684f15adf99a10c4R504). I didn't find exactly the 21 usages mentioned in the ticket, but here is a catalog of usages with my decisions and testing instructions:

`Instance.with_role` usages:
- [QueryType form_rules](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L451-L459) should continue to return all instances
- [QueryType client_detail_forms](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L501-L505) - this is the one that's fixed in the PR linked above.
- [PickListOption external_form_types_for_project](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/forms/pick_list_option.rb#L451-L460) updated in this PR
  - To test, go to the config tool and create a new external form type in draft. Add a rule enabling it in a project. Navigate to the project.
  - If it was the only external form rule for this project, then External Forms should not be available in the project left nav
  - If there is exactly one other, then navigating to External Forms for this project should auto-select that other one (and not show you a dropdown where you can select this one)
  - If there are multiple others, this one should not be present in the dropdown
- [PickListOption assessment_names_for_project](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/forms/pick_list_option.rb#L504-L519) - used for filtering existing assessments, so no change needed
- [HmisSchema::Project external_form_submissions](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/project.rb#L233-L247) - no change needed, since this is just used for getting the definition identifiers to look for. If the form is in draft, there wouldn't be any submissions, but if there were, it wouldn't be a bug to include them.
- [Project.receiving_referrals and receives_referrals](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/models/hmis/hud/project.rb#L137) - updated in this PR
  - To test, go to the admin UI and create a new Referral form type in draft. Add a rule enabling it in a project.
  - Go to a project that can make outgoing referrals and click New Referral. The project should not appear in the dropdown of available projects receiving referrals.
- [project.data_collection_features](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/models/hmis/hud/project.rb#L224) - updated in this PR, but see inline comment; let's discuss
  - To test, create a draft data collection feature form (for example, CLS or CE event) and add a rule to enable it in a project. Navigate to an enrollment dashboard in that project, you should not see the data collection feature in the left-hand nav.
- [project.occurrence_point_form_instances](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/models/hmis/hud/project.rb#L278) - updated in this PR
  - To test, create a draft Occurrence Point Form and add a rule enabling it in a project.
  - Visit an enrollment dashboard in that project. On release-137 you will see an error thrown, "Unable to load definition for instance". This is because we added the `published` scope [here in the `OccurrencePointForm` schema type](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb#L29-L30), but didn't yet add it to the query for form instances. This branch fixes the issue. You should not see an error, and should also not see the draft occurrence point form.
  - We should either push this whole PR to release-137 or factor this specific change out into a separate PR so that it goes out with 137.

`Definition.with_role` usages:
- [QueryType static_form_definition](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L241-L248) - to discuss. Should we change this to use the published scope on Definition?
- [QueryType external_form_definition](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L412-L419) - I just removed the whole query, since it was deprecated on release-135
- [QueryType form_identifiers](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L439-L449) - `latest_versions` scope is already applied so this won't return duplicates. It should continue to return unpublished forms.
- [CreateFormDefinition mutation](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/mutations/create_form_definition.rb#L18) - no change needed since we are checking for duplicate identifiers, which should be unique regardless of publication status of the form
- [Definition for_project](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/models/hmis/form/definition.rb#L261-L263) and [definition_for_role](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/models/hmis/form/definition.rb#L346-L349) already includes published scope

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/project.rb
+++ b/drivers/hmis/app/models/hmis/hud/project.rb
@@ -134,7 +134,7 @@ class Hmis::Hud::Project < Hmis::Hud::Base
 
   scope :receiving_referrals, -> do
     # Find all active instances that enable the Referral functionality
-    instance_scope = Hmis::Form::Instance.active.with_role(:REFERRAL)
+    instance_scope = Hmis::Form::Instance.active.with_role(:REFERRAL).published
     # Find open projects that have an instance that match the criteria, which indicates that the
     # project accepts referrals.
 
@@ -180,7 +180,7 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   end
 
   def receives_referrals?
-    Hmis::Form::Instance.active.with_role(:REFERRAL).any? { |instance| instance.project_match(self) }
+    Hmis::Form::Instance.active.published.with_role(:REFERRAL).any? { |instance| instance.project_match(self) }
   end
 
   def active
@@ -221,7 +221,9 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   def data_collection_features
     # Create OpenStruct for each enabled feature
     Hmis::Form::Definition::DATA_COLLECTION_FEATURE_ROLES.map do |role|
-      base_scope = Hmis::Form::Instance.with_role(role)
+      # To discuss: We don't currently support retiring forms, but if we do, this would break in a way that we might not catch right away.
+      # What's the most graceful way to future proof this? e.g. published_or_retired scope; reuse valid_status_for_submit?
+      base_scope = Hmis::Form::Instance.with_role(role).published
       # Service instances must specify a service type or category.
       base_scope = base_scope.for_services if role == :SERVICE
 
@@ -275,7 +277,7 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   # Occurrence Point Form Instances that are enabled for this project (e.g. Move In Date form)
   def occurrence_point_form_instances
     # All instances for Occurrence Point forms
-    base_scope = Hmis::Form::Instance.with_role(:OCCURRENCE_POINT).active
+    base_scope = Hmis::Form::Instance.with_role(:OCCURRENCE_POINT).active.published
 
     # All possible form identifiers used for Occurrence Point collection
     occurrence_point_identifiers = base_scope.pluck(:definition_identifier).uniq
--- a/drivers/hmis/app/models/hmis/hud/project.rb
+++ b/drivers/hmis/app/models/hmis/hud/project.rb
@@ -221,8 +221,10 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   def data_collection_features
     # Create OpenStruct for each enabled feature
     Hmis::Form::Definition::DATA_COLLECTION_FEATURE_ROLES.map do |role|
-      # To discuss: We don't currently support retiring forms, but if we do, this would break in a way that we might not catch right away.
-      # What's the most graceful way to future proof this? e.g. published_or_retired scope; reuse valid_status_for_submit?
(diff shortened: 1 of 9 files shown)


drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -221,7 +221,9 @@ def to_pick_list_option
   def data_collection_features
     # Create OpenStruct for each enabled feature
     Hmis::Form::Definition::DATA_COLLECTION_FEATURE_ROLES.map do |role|
-      base_scope = Hmis::Form::Instance.with_role(role)
+      # To discuss: We don't currently support retiring forms, but if we do, this would break in a way that we might not catch right away.
+      # What's the most graceful way to future proof this? e.g. published_or_retired scope; reuse valid_status_for_submit?
```
gigxz: Good point, yes, I think we should include both published and retired to maintain the current intended behavior, and add a comment about why. "valid_status_for_submit" probably isn't the clearest since this isn't a submission.

https://github.com/open-path/Green-River/issues/6159 outlines switching this to always show a feature if there is any data for it, in addition to checking the form rules. May make sense to just update the comment to reference that ticket as well.

gigxz (approved):
excellent!! thank you! 
looks good to me, only suggestion is to include exclusively retired forms in the data collection feature query, as you pointed out.

gigxz:
And thank you for writing out each usage so clearly, it was super helpful. For the question about static forms, they by definition shouldn't have multiple versions since they aren't managed in the config tool. So I don't think we need any change there.

> [QueryType static_form_definition](https://github.com/greenriver/hmis-warehouse/blob/release-137/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L241-L248) - to discuss. Should we change this to use the published scope on Definition?

PR #4815, opened 2024-10-10
Households are not getting counted in APR

dtgreiner:
## _Merging this PR_

## Description
Update the HH Move In Date calculation so it doesn't leave out HH members without a move in date and instead correctly calculates and uses the HH move in date based on the entrry date & HoH move in date.

## Type of change
Bug fix

### DIFF
--- a/app/models/concerns/hud_reports/households.rb
+++ b/app/models/concerns/hud_reports/households.rb
@@ -119,12 +119,27 @@ module HudReports::Households
       # If the move-in-date is valid, just use it
       return move_in_date if move_in_date.present? && move_in_date >= she.first_date_in_program
 
-      # If the client moved in before the entry date, and the HoH was present on the move-in date, use the
-      # entry date as the move-in date.
+      # Get HoH for further calculations
       household_members = households[hh_id]
       hoh = household_members.detect { |hm| hm[:relationship_to_hoh] == 1 }
-      return nil unless hoh.present?
-      return she.first_date_in_program if hoh[:move_in_date].present? && hoh[:entry_date] <= (move_in_date || hoh[:move_in_date])
+      # HoH does not exist - cannot do further calculations
+      return nil unless hoh.present? && hoh[:move_in_date].present?
+
+      # [Handling Housing Move-In Dates] - https://files.hudexchange.info/resources/documents/HMIS-Standard-Reporting-Terminology-Glossary-2024.pdf
+
+      # Heads of household with [housing move-in dates] prior to their [project start dates] should have the [housing move-in dates] disregarded entirely.
+      return nil unless hoh[:entry_date] <= hoh[:move_in_date]
+
+      # When a household member was already in the household when they became housed (individual’s [project
+      # start date] <= head of household’s [housing move-in date]), the head of household’s [housing move-in date]
+      # should be used as the individual’s [housing move-in date]. If the household member exited before the
+      # household moved into housing, they do not inherit this [housing move-in date].
+      return hoh[:move_in_date] if she.first_date_in_program <= hoh[:move_in_date]
+
+      # When a household member joins the household after they are already housed (individual’s [project start
+      # date] > head of household’s [housing move-in date]), the individual’s [project start date] should be used as
+      # the individual’s [housing move-in date].
+      return she.first_date_in_program if she.first_date_in_program > hoh[:move_in_date]
 
       # Otherwise this move-in is completely invalid
       nil
--- a/app/models/concerns/hud_reports/households.rb
+++ b/app/models/concerns/hud_reports/households.rb
@@ -117,7 +117,7 @@ module HudReports::Households
     private def calculate_move_in_date(hh_id, she)
       move_in_date = she.move_in_date
       # If the move-in-date is valid, just use it
-      return move_in_date if move_in_date.present? && move_in_date >= she.first_date_in_program
+      return move_in_date if move_in_date.present? && move_in_date >= she.entry_date
 
       # Get HoH for further calculations
       household_members = households[hh_id]
@@ -134,12 +134,12 @@ module HudReports::Households
       # start date] <= head of household’s [housing move-in date]), the head of household’s [housing move-in date]
       # should be used as the individual’s [housing move-in date]. If the household member exited before the
       # household moved into housing, they do not inherit this [housing move-in date].
-      return hoh[:move_in_date] if (she.first_date_in_program..she.exit_date).include?(hoh[:move_in_date])
+      return hoh[:move_in_date] if (she.entry_date..she.exit_date).include?(hoh[:move_in_date])
 
       # When a household member joins the household after they are already housed (individual’s [project start
       # date] > head of household’s [housing move-in date]), the individual’s [project start date] should be used as
       # the individual’s [housing move-in date].
-      return she.first_date_in_program if she.first_date_in_program > hoh[:move_in_date]
+      return she.entry_date if she.entry_date > hoh[:move_in_date]
 
       # Otherwise this move-in is completely invalid
       nil
--- a/app/models/concerns/hud_reports/households.rb
+++ b/app/models/concerns/hud_reports/households.rb
@@ -122,6 +122,7 @@ module HudReports::Households
       # Get HoH for further calculations
       household_members = households[hh_id]
       hoh = household_members.detect { |hm| hm[:relationship_to_hoh] == 1 }
+
       # HoH does not exist or does not have a move-in date - cannot do further calculations
       return nil unless hoh.present? && hoh[:move_in_date].present?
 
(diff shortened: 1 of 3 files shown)


app/models/concerns/hud_reports/households.rb
```diff
@@ -115,18 +115,16 @@ def hoh_age(household_id, date)
...
-      return move_in_date if move_in_date >= she.first_date_in_program
+      return move_in_date if move_in_date.present? && move_in_date >= she.first_date_in_program
 
       # If the client moved in before the entry date, and the HoH was present on the move-in date, use the
       # entry date as the move-in date.
       household_members = households[hh_id]
       hoh = household_members.detect { |hm| hm[:relationship_to_hoh] == 1 }
       return nil unless hoh.present?
-      return she.first_date_in_program if hoh[:entry_date] <= move_in_date
+      return she.first_date_in_program if hoh[:move_in_date].present? && hoh[:entry_date] <= (move_in_date || hoh[:move_in_date])
```
eanders: I think this isn't quite right. Maybe:
```suggestion
      # Household was moved in when client arrived
      return she.first_date_in_program if hoh[:move_in_date].present? && hoh[:move_in_date] <= she.first_date_in_program
```
We might also want to short circuit if the HoH move-in date is < the HoH entry date, but that's more of a data quality issue.

app/models/grda_warehouse/tasks/service_history/enrollment.rb
```diff
@@ -60,7 +60,11 @@ def self.ensure_there_are_no_extra_enrollments_in_service_history(client_ids)
     end
 
     def invalidate_source_data!
-      update(processed_as: nil)
+      if self.HouseholdID.present?
```
eanders, line 64:
```suggestion
      # Invalidate all enrollments for this household if this enrollment is being invalidated
      if self.HouseholdID.present?
```

app/models/concerns/hud_reports/households.rb
dtgreiner: rerunning the test kits with this change and will add it in if they pass

app/models/grda_warehouse/tasks/service_history/enrollment.rb
dtgreiner: will include this in the next commit

app/models/concerns/hud_reports/households.rb
dtgreiner: I think I've got something working for this. It's passing the test kit for the organization that was failing earlier. Running a full test kit to verify it doesn't break elsewhere and if we're good, I'll get it into the PR.

dtgreiner: @eanders a new commit for this has just been pushed to the PR

app/models/concerns/hud_reports/households.rb
```diff
@@ -115,18 +115,31 @@ def hoh_age(household_id, date)
...
+      return move_in_date if move_in_date.present? && move_in_date >= she.first_date_in_program
 
-      # If the client moved in before the entry date, and the HoH was present on the move-in date, use the
-      # entry date as the move-in date.
+      # Get HoH for further calculations
       household_members = households[hh_id]
       hoh = household_members.detect { |hm| hm[:relationship_to_hoh] == 1 }
-      return nil unless hoh.present?
-      return she.first_date_in_program if hoh[:entry_date] <= move_in_date
+      # HoH does not exist - cannot do further calculations
```
eanders:
```suggestion
      # HoH does not exist or does not have a move-in date - cannot do further calculations
```

app/models/concerns/hud_reports/households.rb
```diff
@@ -115,18 +115,31 @@ def hoh_age(household_id, date)
...
+      # [Handling Housing Move-In Dates] - https://files.hudexchange.info/resources/documents/HMIS-Standard-Reporting-Terminology-Glossary-2024.pdf
+
+      # Heads of household with [housing move-in dates] prior to their [project start dates] should have the [housing move-in dates] disregarded entirely.
+      return nil unless hoh[:entry_date] <= hoh[:move_in_date]
+
+      # When a household member was already in the household when they became housed (individual’s [project
+      # start date] <= head of household’s [housing move-in date]), the head of household’s [housing move-in date]
+      # should be used as the individual’s [housing move-in date]. If the household member exited before the
+      # household moved into housing, they do not inherit this [housing move-in date].
+      return hoh[:move_in_date] if she.first_date_in_program <= hoh[:move_in_date]
```
eanders: This looks like they _are_ inheriting from the HoH (136 would imply otherwise).  I think this is inline with 133-135.

eanders (approved):
I think the logic looks correct, but I'm not convinced the comments are quite right to describe what's going on.  Can you take a quick look at those?

app/models/concerns/hud_reports/households.rb
dtgreiner: you're right, I believe it is saying that the HoH move in date is inherited only if the HoH Move in date is between the first date in the program and the exit date.

app/models/concerns/hud_reports/households.rb
```diff
@@ -115,18 +115,31 @@ def hoh_age(household_id, date)
...
+      # When a household member was already in the household when they became housed (individual’s [project
+      # start date] <= head of household’s [housing move-in date]), the head of household’s [housing move-in date]
+      # should be used as the individual’s [housing move-in date]. If the household member exited before the
+      # household moved into housing, they do not inherit this [housing move-in date].
+      return hoh[:move_in_date] if (she.first_date_in_program..she.exit_date).include?(hoh[:move_in_date])
+
+      # When a household member joins the household after they are already housed (individual’s [project start
+      # date] > head of household’s [housing move-in date]), the individual’s [project start date] should be used as
+      # the individual’s [housing move-in date].
+      return she.first_date_in_program if she.first_date_in_program > hoh[:move_in_date]
```
eanders: Thanks @dtgreiner this looks great.  One tiny tweak since you have `she.exit_date` let's go ahead and use the aliases for both dates:
```suggestion
      return hoh[:move_in_date] if (she.entry_date..she.exit_date).include?(hoh[:move_in_date])

      # When a household member joins the household after they are already housed (individual’s [project start
      # date] > head of household’s [housing move-in date]), the individual’s [project start date] should be used as
      # the individual’s [housing move-in date].
      return she.entry_date if she.entry_date > hoh[:move_in_date]
```

dtgreiner: Adding this in and will have the commit pushed to the PR momentarily

eanders:
@dtgreiner there seems to be a mis-match with the database shape and the fixtures.  I guess maybe we need to push a commit to re-generate the test kit fixtures?

dtgreiner:
> @dtgreiner there seems to be a mis-match with the database shape and the fixtures. I guess maybe we need to push a commit to re-generate the test kit fixtures?

I'll get that queued up momentarily

app/models/grda_warehouse/tasks/service_history/enrollment.rb
```diff
@@ -60,7 +60,12 @@ def self.ensure_there_are_no_extra_enrollments_in_service_history(client_ids)
     end
 
     def invalidate_source_data!
```
gigxz, line 62: @dtgreiner @eanders I'm trying to see if households are invalidated when an enrollment is invalidate. Elliot pointed me to this change, but afaict this method `invalidate_source_data!` is not called anywhere in the codebase. Am I missing something?

For HMIS Enrollments, and it appears for WH Enrollments too, [these invalidate_processing! methods are used](https://github.com/greenriver/hmis-warehouse/blob/3b88c3a659a940fdf4cd7c2916792d36b7a0a43c/app/models/concerns/hud_concerns/enrollment.rb#L45-L55). I think those methods need to be updated to invalidate all household members.

eanders: @gigxz `invalidate_processing!` is called in `ClientCleanup` and `ProjectCleanup`.  These are run nightly and in situations where changes are made to either the projects or clients that require rebuilds, they are used to tell `app/models/grda_warehouse/tasks/service_history/enrollment.rb` that the enrollments need to be rebuilt.  I think you are correct about `invalidate_source_data!`.  That was added to fix this issue, but I'm unable to figure out where it is called. @dtgreiner was that method maybe supposed to replace `invalidate_processing!`?

dtgreiner: @eanders If I'm recalling correctly, we identified the `invalidate_source_data!` method as the one supposed to be run when the service history source data changes, but it does look like `invalidate_processing!` is doing this. The `invalidate_source_data!` method appears to have been around since 2020, but looking into it more it doesn't seem to ever have been referenced. It may be worth a call if you have time to dig into this a little more.

PR #4817, opened 2024-10-10
Allow bulk reviewed submissions to succeed even when others fail

martha:
## _Merging this PR_

## Description

GH issue: https://github.com/open-path/Green-River/issues/6744

Previous behavior: If any submission failed bulk review, all the submissions in that batch would fail (keeping their 'new' status), and the user would not know which submission(s) caused the failure.

New behavior: If any submission fails in bulk review, the other submissions still succeed. The user gets an error message telling them which submissions they might want to consider investigating and marking as spam.

How to test: I ran the following in my local shell,
```
s = HmisExternalApis::ExternalForms::FormSubmission.where(status: 'new').last
s.raw_data = {"invalid key"=>"bad value"}
s.save!
```
This causes the submission to fail review in the External Form Submission UI.

## Type of change
Existing feature improvement

### DIFF
--- a/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
+++ b/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
@@ -64,6 +64,12 @@ module Mutations
         base_message = "Bulk review failed on #{failed_to_review.size} of #{submissions.size} records."
         error_message = base_message + "\n" + failed_to_review.map { |id, message| "\tSubmission #{id}: #{message}" }.join("\n")
         display_message = base_message + ' This may indicate that the following submissions are spam: ' + failed_to_review.keys.join(', ')
+
+        # Raise, instead of returning a ValidationError, so we will get notified in Sentry, since either a bug with
+        # our implementation or an influx of spam would be interesting for us to get notified about.
+        # If this gets extremely noisy due to spam, we might be tempted to turn off the Sentry error and return a
+        # ValidationError instead. If we do that, we should take care to *keep* Sentry errors on for invalid
+        # client/enrollment errors (that would indicate an implementation problem and not bad/spammy data).
         raise HmisErrors::ApiError.new(error_message, display_message: display_message)
       end
 
--- a/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
+++ b/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
@@ -28,14 +28,14 @@ module Mutations
       failed_to_review = {}
 
       submissions.each do |record|
-        raise ArgumentError, 'Already reviewed' if record.enrollment_id.present? || record.status == 'reviewed'
+        next if record.enrollment_id.present? || record.status == 'reviewed'
 
         record.status = 'reviewed'
         record.run_form_processor(current_user, project: project)
 
         if record.enrollment&.new_record?
-          raise ArgumentError, record.enrollment.client.errors.full_messages unless record.enrollment.client.valid?
-          raise ArgumentError, record.enrollment.errors.full_messages unless record.enrollment.valid?
+          raise record.enrollment.client.errors.full_messages unless record.enrollment.client.valid?
+          raise record.enrollment.errors.full_messages unless record.enrollment.valid?
 
           record.enrollment.client.save!
           should_auto_enter ? record.enrollment.save_and_auto_enter! : record.enrollment.save_in_progress!
@@ -43,14 +43,14 @@ module Mutations
 
         record.form_processor.save!
         record.save!
-      rescue StandardError => e
+      rescue RuntimeError => e
         failed_to_review[record.id] = e.message
       end
 
       if failed_to_review.any?
         base_message = "Bulk review failed on #{failed_to_review.size} of #{submissions.size} records."
         error_message = base_message + "\n" + failed_to_review.map { |id, message| "\tSubmission #{id}: #{message}" }.join("\n")
-        display_message = base_message + ' This may indicate that the following submissions are spam, or have already been reviewed: ' + failed_to_review.keys.join(', ')
+        display_message = base_message + ' This may indicate that the following submissions are spam: ' + failed_to_review.keys.join(', ')
         raise HmisErrors::ApiError.new(error_message, display_message: display_message)
       end
 
--- a/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
+++ b/drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
@@ -43,14 +43,14 @@ module Mutations
 
         record.form_processor.save!
         record.save!
-      rescue StandardError => e # does it make sense to rescue all StandardError here or be specific?
+      rescue StandardError => e
         failed_to_review[record.id] = e.message
       end
 
       if failed_to_review.any?
         base_message = "Bulk review failed on #{failed_to_review.size} of #{submissions.size} records."
         error_message = base_message + "\n" + failed_to_review.map { |id, message| "\tSubmission #{id}: #{message}" }.join("\n")
-        display_message = base_message + ' This may indicate that the following submissions are spam: ' + failed_to_review.keys.join(', ')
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
```diff
@@ -22,36 +22,39 @@ def resolve(external_submission_ids:)
...
+          record.enrollment.client.save!
+          should_auto_enter ? record.enrollment.save_and_auto_enter! : record.enrollment.save_in_progress!
         end
-      end
 
-      { success: true }
-    end
+        record.form_processor.save!
+        record.save!
+      rescue StandardError => e
```
gigxz: Rescuing `StandardError` can cause unexpected behavior, because I most ruby errors are subclasses of StandardError. So basically _any_ execution error that happens in the block above will be swallowed.

A safer option is to just skip and count up failures as you go:

```ruby
submissions.each do |record|
    next if already_processed

    if !record.valid?
       failed_to_review << record
       next
    end

end
```

drivers/hmis/spec/requests/hmis/bulk_review_external_submissions_spec.rb
```diff
@@ -66,6 +66,24 @@ def perform_mutation(ids: [s1.id, s2.id])
...
   end
 
+  context 'when some submissions fail but others are fine' do
+    let!(:s3) { create(:hmis_external_form_submission, status: 'reviewed', definition: definition, raw_data: { 'your_name' => 'Barry' }) }
+    let!(:s4) { create(:hmis_external_form_submission, definition: definition, raw_data: { 'some_invalid_key' => 'Bad data!' }) }
+
+    it 'succeeds with the non-problematic submissions' do
+      expect do
+        expect_gql_error perform_mutation(ids: [s1.id, s2.id, s3.id, s4.id]), message: /Bulk review failed on 2 of 4 records./
+        s1.reload
```
gigxz: A little briefer `[s1, s2, s3, s4].map(&:reload)`

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
```diff
@@ -22,36 +22,39 @@ def resolve(external_submission_ids:)
...
+        failed_to_review[record.id] = e.message
+      end
 
-    protected
+      if failed_to_review.any?
+        base_message = "Bulk review failed on #{failed_to_review.size} of #{submissions.size} records."
+        error_message = base_message + "\n" + failed_to_review.map { |id, message| "\tSubmission #{id}: #{message}" }.join("\n")
+        display_message = base_message + ' This may indicate that the following submissions are spam, or have already been reviewed: ' + failed_to_review.keys.join(', ')
+        raise HmisErrors::ApiError.new(error_message, display_message: display_message)
```
gigxz, line 73: The message looks good to me!

gigxz: Since we're raising instead of returning a ValidationError, we will get notified in Sentry. I think that's probably the right call since it _could_ point to a problem with our implementation.

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
martha: Ok, sounds good! Confirming behavior change - it's fine if a submission has already been reviewed (we should just silently pass on rather than returning it as a failure?) I like this change, just want to make sure it's also what you intended.

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
martha: Yes, this was my thought too. Also because we're curious to see how much spam we will actually get, if any. If it gets really noisy for reasons we can't control then we can turn off the sentry alert by returning a `ValidationError` instead.

drivers/hmis/app/graphql/mutations/bulk_review_external_submissions.rb
gigxz: Yeah I think that's good, it mirrors how bulk services works now, too. It will help in situations where there are multiple users reviewing at the same time.

martha: Hm so, I think we still need some sort of try/catch here because the failure won't always be due to validation errors on the record or its related record. The reason I originally caught `StandardError` is bc that's what's [caught and then re-thrown by the FormProcessor](https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/drivers/hmis/app/models/hmis/form/form_processor.rb#L80-L84). We could catch `RuntimeError` specifically, or modify [unknown_field_error](https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/drivers/hmis/app/models/hmis/form/form_processor.rb#L51-L53) to return a different, more specific type of error.

martha: I guess my proposal is to catch `RuntimeError` and, since we need the try/catch anyway, to reduce code cruft by raising if the client or enrollment were invalid also (instead of your suggestion of `if valid?`, but I could go either way on this).  https://github.com/greenriver/hmis-warehouse/pull/4817/commits/055e3f4e4dec4a1754dccaa8792ec662443f6016

As an aside, while I was working on this, I couldn't find/remember a good test case for invalid client or enrollment. I tried some obvious ones (client DOB is 200 years ago, etc.) before realizing that [client validations are only run in specific contexts](https://github.com/greenriver/hmis-warehouse/blob/stable/drivers/hmis/app/models/hmis/hud/client.rb#L82), and external form submission is not one of them. I'm assuming that is intentional, because we want them to be able to create the client/enrollment and then fix DQ issues later.

gigxz: Oh, I didn't realize you were trying to catch errors from the FormProcessor run! method, too, but that makes sense. I guess I would suggest making the `rescue` block as small as possible to make that clearer, and to reduce the risk of swallowing errors that don't come from the form processor.

```
submissions.each do |record|
    next if already_processed

    begin
        run_form_processor!
    rescue StandardError => e
         # if there was an unexpected error running the form processor, it could mean the form response was malformed. catch this error so that it doesn't cause the whole bulk batch to fail.
          failed_to_review << e
    end

    if something_isnt_valid
       failed_to_review << record
       next
    end

end
```

And I agree it would be even better to subclass StandardError inside the form processor like you suggested, so we could `rescue FormProcessingError`. But I think that might involve a little too much code churn.. "unknown field error" isnt the only type of error, I think we would also wanna re-raise other exceptions as FormProcessingErrors [here](https://github.com/greenriver/hmis-warehouse/blob/5eed1631d1ae87951eb01bb1b4d4e7d064142b27/drivers/hmis/app/models/hmis/form/form_processor.rb#L80-L84).


In my example I didn't mean to say anything special about the validity check, sorry about that confusion. I agree that it makes sense to check the validity of the client/enrollment in order to prevent an unhandled exception on save!. One way to have and invalid client is to have a null PersonalID or EnrollmentID, but I'm not sure if there's even a way to make it happen with bad input.

martha: Ok, great, thank you for clarifying! I think this all makes sense.

> I'm not sure if there's even a way to make it happen with bad input.

I guess this makes me wonder whether we should just allow `.save!` to raise an unhandled exception? If it can't be due to bad data, it has to be because of a bug we introduced, so I think of it as in a different class of error -- and we want to get notified via sentry right away. On the other hand, I guess there's no harm in allowing the rest of the bulk submissions to succeed... as long as we never do [what I suggested in this comment](https://github.com/greenriver/hmis-warehouse/pull/4817#discussion_r1797320177) and turn off sentry alerts when it gets too noisy. I'll add some comments to this effect

gigxz: If theres no way to set it off from bad input, I agree it would be fine to let it raise. I didn't spend much time looking into it, though.

PR #4818, opened 2024-10-10
Fix for LSA filters

eanders:
## _Merging this PR_

## Description
* Passes filters to the json end point as a json object
* Passes all form fields instead of a select few
* Re-uses report and filter logic from the report to bring check in-sync with what is being run

## Type of change

### DIFF
--- a/app/assets/javascripts/filter_projects.js
+++ b/app/assets/javascripts/filter_projects.js
@@ -108,13 +108,27 @@ App.StimulusApp.register('filter-projects', class extends Stimulus.Controller {
     return data
   }
 
+  rawFormFiltersInput() {
+    const inputs = {}
+    $(this.element).find(':input').each((i, el) => {
+      // Hack to convert nasty rails param format into a functional JSON object
+      if(el.name.startsWith('filter[')) {
+        let name = el.name.replace('filter[', '').replace('][]', '').replace(/\]$/, '')
+        let val = $(el).val();
+        inputs[name] = val;
+      }
+    })
+    return { filter: inputs }
+  }
+
   checkMissingData() {
     if (! this.hasMissingItemsTarget) {
       return false
     }
+    const form_data = this.rawFormFiltersInput()
     return $.ajax({
-      async: false, type: 'POST', url: $(this.missingItemsTarget).attr('formaction') + '.json', data: { filter: this.formData() } }).done().responseJSON
-  }
+      async: false, type: 'POST', url: $(this.missingItemsTarget).attr('formaction') + '.json', data: JSON.stringify(form_data), contentType: "application/json" }).done().responseJSON
+    }
 
   addMissingDataWarning() {
     if (!this.hasMissingItemsTarget) {
--- a/app/assets/javascripts/filter_projects.js
+++ b/app/assets/javascripts/filter_projects.js
@@ -109,16 +109,17 @@ App.StimulusApp.register('filter-projects', class extends Stimulus.Controller {
   }
 
   rawFormFiltersInput() {
-    const inputs = {}
+    let inputs = {}
     $(this.element).find(':input').each((i, el) => {
       // Hack to convert nasty rails param format into a functional JSON object
       if(el.name.startsWith('filter[')) {
-        let name = el.name.replace('filter[', '').replace('][]', '').replace(/\]$/, '')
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/filter_projects.js
```diff
@@ -108,13 +108,27 @@ App.StimulusApp.register('filter-projects', class extends Stimulus.Controller {
...
 
+  rawFormFiltersInput() {
+    const inputs = {}
+    $(this.element).find(':input').each((i, el) => {
+      // Hack to convert nasty rails param format into a functional JSON object
+      if(el.name.startsWith('filter[')) {
+        let name = el.name.replace('filter[', '').replace('][]', '').replace(/\]$/, '')
+        let val = $(el).val();
+        inputs[name] = val;
+      }
```
ttoomey: This is functional but it makes me a little uncomfortable that it will break if we ever introduce another level of nesting in any of our report filters.
For example:
```js
let key = 'filter[client][name]';
key.replace('filter[', '').replace('][]', '').replace(/\]$/, '');
=> 'client][name'
```

We could support deeply nested objects generically but that's more some complexity than we might need. Perhaps we can just alert to sentry if we encounter a field we can't process? 

___Note:___ it's not clear to me that we actually have client side js sentry in the WH. I'm assuming we do and I'm just missing it

```suggestion
      // not tested
      if(el.name.startsWith('filter[')) {
        const matches = el.name.match(/^filter\[([^\]]+)\](\[\])?$/); // matches one level of nesting
        const fieldName = matches ? matches[1] : null;

        if (fieldName) {
          inputs[fieldName] = $(el).val();
        } else {
           // deeply nested fields are not supported
           Sentry.captureMessage(`field format not supported: "${el.name}"`);
        }
      }
```

ttoomey (approved):
Looks good! I left a note around unsupported filter fields

app/assets/javascripts/filter_projects.js
eanders: Pretty sure we _don't_ have client side sentry in the warehouse.

ttoomey: Got it, thanks. Feels like we should have that [tracking that here](https://github.com/open-path/Green-River/issues/6837)

app/assets/javascripts/filter_projects.js
```diff
@@ -108,13 +108,28 @@ App.StimulusApp.register('filter-projects', class extends Stimulus.Controller {
...
+      // Hack to convert nasty rails param format into a functional JSON object
+      if(el.name.startsWith('filter[')) {
+        // remove any trailing arrays, this is a rails-ism and generates weird empty hashes
+        let name_parts = el.name.replaceAll('][]', '').replaceAll("]", "").split("[")
+        const input = name_parts.reduceRight((nestedObj, key) => ({ [key]: nestedObj }), $(el).val());
+        inputs = $.extend(true, {}, inputs, input);
+      }
+    })
+    return inputs
+  }
```
ttoomey: Looks good to me, very elegant!

This is fine as is. It might benefit from a little more documentation and some small tweaks to name_parts (use let rather than const since we aren't reassigning, use camaleCase). 

```suggestion
/**
 * Transform rails-formatted input field names into a deeply nested object for submission
 * 
 * Example inputs:
 * <input name="filter[project_ids][]" value="123">
 * <input name="filter[data_source_ids][]" value>
 * <input name="filter[widget]" value="100">
 * 
 * Example output:
 * {
 *   filter: {
 *     project_ids: ["123"],
 *     data_source_ids: [], // empty arrays get lost with unless we submit as JSON
 *     widgets: "100"
 *   }
 * }
 */
  rawFormFiltersInput() {
    let inputs = {};
    $(this.element).find(':input').each((i, el) => {
      if(el.name.startsWith('filter[')) {
        // remove any trailing arrays, this is a rails-ism and generates weird empty hashes
        const nameParts = el.name.replaceAll('][]', '').replaceAll("]", "").split("[");
        const input = nameParts.reduceRight((nestedObj, key) => ({ [key]: nestedObj }), $(el).val());
        inputs = $.extend(true, {}, inputs, input); // use jquery for deep merge
      }
    })
    return inputs;
  }
```

ttoomey (approved):
Looks good, thanks!

PR #4823, opened 2024-10-11
CA and THRIVE QA fixes

eanders:
## _Merging this PR_

## Description
This enforces that QualifyingActivities occur on or after the completion of the associated assessments.

## Type of change
Hotfix

### DIFF
--- a/drivers/health_comprehensive_assessment/app/models/health_comprehensive_assessment/assessment.rb
+++ b/drivers/health_comprehensive_assessment/app/models/health_comprehensive_assessment/assessment.rb
@@ -52,6 +52,16 @@ module HealthComprehensiveAssessment
       completed_on.present?
     end
 
+    # A QualifyingActivity that is the same or newer than the date created of this CA
+    def ca_completed_qa?
+      return false unless completed_on.present?
+
+      qa = patient.current_qa_factory&.ca_completed_qa
+      return false unless qa.present? && qa.date_of_activity.present?
+
+      qa.date_of_activity >= completed_on
+    end
+
     def active?
       completed_on && completed_on >= 1.years.ago
     end
--- a/drivers/health_thrive_assessment/app/models/health_thrive_assessment/assessment.rb
+++ b/drivers/health_thrive_assessment/app/models/health_thrive_assessment/assessment.rb
@@ -43,6 +43,16 @@ module HealthThriveAssessment
       completed_on.present?
     end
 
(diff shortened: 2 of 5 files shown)


drivers/health_comprehensive_assessment/app/models/health_comprehensive_assessment/assessment.rb
```diff
@@ -52,6 +52,16 @@ def completed?
...
     end
 
+    # A QualifyingActivity that is the same or newer than the date created of this CA
+    def ca_completed_qa?
+      return false unless completed_on.present?
+
+      qa = patient.current_qa_factory&.ca_completed_qa
+      return false unless qa.present? && qa.date_of_activity.present?
+
+      qa.date_of_activity >= completed_on
```
ttoomey, line 62: This is fine as-is. But is `present?` necessary because we might have empty strings or something? If not, maybe skipping the presence check and relying on falsy nils is less boilerplate and more clear:
```suggestion
      return false unless completed_on

      date_of_activity = patient.current_qa_factory&.ca_completed_qa&.date_of_activity
      return false unless date_of_activity

      date_of_activity >= completed_on
```

drivers/health_thrive_assessment/app/models/health_thrive_assessment/assessment.rb
```diff
@@ -43,6 +43,16 @@ def completed?
       completed_on.present?
     end
 
+    # A QualifyingActivity that is the same or newer than the date created of this CA
+    def hrsn_screening_qa?
```
ttoomey, line 47: see note above on `ca_completed_qa?`; do we need to call `.present?`

ttoomey (approved):
looks good

drivers/health_comprehensive_assessment/app/models/health_comprehensive_assessment/assessment.rb
eanders: I almost did that, but went with `.present?` because I don't know the code that's further up the chain, and it's fairly complex and potentially distributed.  We've been bit before with exceptions from assuming a particular return type, adding `.present?` is a relatively cheap way of avoiding the exception until we have strong typing.

ttoomey: makes sense, sounds like we have reason for extra defensive programming

PR #4824, opened 2024-10-14
Fix assessment-related test failures on 136

martha:
## _Merging this PR_

## Description
Fixes test failures: https://github.com/greenriver/hmis-warehouse/actions/runs/11309250113/job/31452914339?pr=4762 which were introduced in release-136.

I don't know exactly what change introduced the test failures, but the problem was that it was relying on the pre-existence of a form definition that didn't exist in the test db. Instead, that definition should be created as part of the test.

## Type of change
Test bugfix


eanders (approved):
Thanks @martha!  I wonder if it has to do with whose SQL files were committed to the repo.  This definitely looks better to me!

PR #4827, opened 2024-10-14
Add system test for form behavior

martha:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

## Description
[//]: # (Summarize changes and include links related issue. List any new dependencies)
GH issue: https://github.com/open-path/Green-River/issues/6776
This PR adds system tests for the form behaviors listed in the linked ticket.

I left some todos for more tests to add, which I'd propose to add as a part of this ticket; let me know if you can think of more cases to add to the list.

I've also left some todos, especially in the bounds section, for form behavior. These, I'd propose to move out into their own ticket(s).

## Type of change
[//]: # (e.g., Bug fix, New feature, Code clean-up, Dependency update)
New system test

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [n/a] If adding a new endpoint / exposing data in a new way, I have:
  - [ ] ensured the API can't leak data from other data sources
  - [ ] ensured this does not introduce N+1s
  - [ ] ensured permissions and visibility checks are performed in the right places
- [n/a] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [n/a] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #4829, opened 2024-10-15
Add system test for bulk services

martha:
## _Merging this PR_
- Note that this PR targets [martha/6776-form-sys-test](https://github.com/greenriver/hmis-warehouse/pull/4827/files) rather than release-138. This was for my convenience only, since it was helpful to be able to copy/paste/modify my work on my previous PR. That PR should be merged first and then this one's base branch can be auto-updated. If this is inconvenient, I'm happy to take the time to tease them out.

## Description
GH issue: https://github.com/open-path/Green-River/issues/6736

This PR adds a spec test testing the behavior of the bulk assign services ux. It also tests behavior for clients with active alerts.
- I remember confusion that sometimes the alert doesn't show up. Is it possible that the current ux, where the client alert is only shown when you enroll a client singly (but not when you bulk enroll), is causing confusion?
- I had some fun trying to chase down timezone/date issues that seem to be specific to the cuprite/capybara environment... or are they? Let's discuss.
  - Working on it at 8pm, it failed to create an enrollment since the entry date can't be in the future
  - capybara chrome is in utc, so [browser's `new Date()`](https://github.com/greenriver/hmis-frontend/blob/ad4c822e79b6b6878fb33d11643ff26b023c23e2/src/modules/services/components/bulk/BulkServicesPage.tsx#L32) is 10/15, but in backend, [`.future?` is after 10/14](https://github.com/greenriver/hmis-warehouse/blob/stable/drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb#L71)
  - I haven't fully reasoned through yet: is this really specific to capybara, or has it just never arisen yet bc we don't have any clients in future timezones relative to our servers?
  - Edit: Created https://github.com/open-path/Green-River/issues/6856

## Type of change
New system test


gigxz:
> I remember confusion that sometimes the alert doesn't show up. Is it possible that the current ux, where the client alert is only shown when you enroll a client singly (but not when you bulk enroll), is causing confusion?

Yes that was it - https://github.com/open-path/Green-River/issues/6735#issuecomment-2394570410

drivers/hmis/spec/system/hmis/bulk_services_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+
+        # But you can still click to continue and add the service
+        click_button 'Add Bed Night'
+        expect(target_row.find('button', text: 'Assigned')).not_to be_nil
+        c1.reload
+        expect(c1.services.sole.type_provided).to eq(bednight_service_type.hud_type_provided)
+        expect(c1.services.sole.project).to eq(p1)
+      end
+
+      it 'does not block you from enrolling in bulk' do
```
gigxz, line 112: Link https://github.com/open-path/Green-River/issues/6796

drivers/hmis/spec/system/hmis/bulk_services_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+  describe 'bed nights' do
+    it 'correctly searches, enrolls, and assigns service to clients' do
+      find('input[type="checkbox"][aria-label="select all"]', visible: :all).trigger(:click)
+      click_button 'Enroll (3) + Assign (3)'
+      assert_text 'Assigned' # wait for it to process
+
+      # Find the indices of the two columns we want to check
+      header_cells = all('thead th')
+      last_bed_night_date_index = header_cells.find_index { |cell| cell.text == 'Last Bed Night Date' }
+      assign_bed_night_index = header_cells.find_index { |cell| cell.text == "Assign Bed Night for #{today.strftime('%m/%d/%Y')}" }
```
gigxz: Ah I see, so this is explicitly testing that the Bed Night tool defaults to selecting the current date. I think we should also test that the user can change the assignment date.

drivers/hmis/spec/system/hmis/bulk_services_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        # Check the "Last Bed Night Date" column
+        last_bed_night_date = row.all('td')[last_bed_night_date_index].text
+        expect(last_bed_night_date).to eq("Today (#{today.strftime('%m/%d/%Y')})")
+
+        # Check the "Assign Bed Night for mm/dd/yyyy" column
+        assign_button = row.all('td')[assign_bed_night_index].find('button')
+        expect(assign_button.text).to eq('Assigned')
+      end
+
+      services = Hmis::Hud::Service.joins(enrollment: [:project, :client]).all
```
gigxz: This feel unnecessarily broad / prone to interference, if there is another test that didn't clean up services. Can we just say `p1.services`?

gigxz: _And_ maybe wrapping the above `click_button` in an assertion to confirm that the block increases `(Hmis::Hud::Service, :count).by(3)` , so we feel confident that there were no services generated outside of the project.

drivers/hmis/spec/system/hmis/bulk_services_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+        expect(c1.services.sole.type_provided).to eq(bednight_service_type.hud_type_provided)
+        expect(c1.services.sole.project).to eq(p1)
+      end
+
+      it 'does not block you from enrolling in bulk' do
+        # TODO - maybe this was what was causing confusion?
+        find('input[type="checkbox"][aria-label="select all"]', visible: :all).trigger(:click)
+        click_button 'Enroll (3) + Assign (3)' # Does not pop up the alert when processing in bulk
+        assert_text 'Assigned' # wait for it to process
+        expect(Hmis::Hud::Service.count).to eq(3)
```
gigxz: Prefer wrapping `click_button` and `assert_text` in an expectation block with a `change` assertion, to emphasize that this action created 3 new services that didn't exist before.

gigxz (approved):
great, thanks!

PR #4830, opened 2024-10-15
Ability to link Warehouse account to specific LMS account

dtgreiner:
## _Merging this PR_

## Description

Add a talent_lms_email field to the user record. This field can be used to link talent lms accounts with the local data we have on record. 

## Type of change
New feature

### DIFF
--- a/app/models/talentlms/facade.rb
+++ b/app/models/talentlms/facade.rb
@@ -60,7 +60,7 @@ module Talentlms
       email_address = lms_email(user)
 
       # If we have a local login record, we have a talent account id. Check here first for the talent record
-      result ||= talentlms_user_record_by_id(login.lms_user_id) if login.present?
+      result = talentlms_user_record_by_id(login.lms_user_id) if login.present?
       # The local login record does not exist OR Talent does not have an account for this id.
       # Check for a Talent account with the assocaited email address
       result ||= talentlms_user_record_by_email(email_address)
--- a/app/models/talentlms/facade.rb
+++ b/app/models/talentlms/facade.rb
@@ -10,18 +10,89 @@ module Talentlms
       @api = Config.first
     end
 
+    def talentlms_user_record_by_id(id)
+      result = @api.post('users', { id: id })
+      result
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
+
+    def talentlms_user_record_by_email(email)
+      result = @api.post('users', { email: email })
+      result
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
+
+    def talentlms_user_record_by_username(username)
+      result = @api.post('users', { username: username })
+      result
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
+
+    def lms_username(user)
+      "#{ENV['RAILS_ENV']}_#{user.id + Integer(ENV.fetch('DEV_OFFSET', 0))}"
+    end
+
+    def lms_email(user)
+      email_address = "#{lms_username(user)}@#{ENV['FQDN']}"
+      email_address = user.talent_lms_email if user.talent_lms_email.present?
+      email_address
+    end
+
     # Login user to TalentLMS
     #
     # @@param user [User] the user
     # @return [String] URL to redirect user to
     def login(user)
       login = Login.find_by(user: user)
-      if login.nil?
-        result = create_account(user)
-      else
-        result = @api.post('userlogin', { login: login.login, password: login.password })
-      end
+
+      # We may have a local talent record for this user, but we need to run it
+      # through the sync process to ensure that the email address that we have
+      # on record matches the email address that TalentLMS has for this user.
+      result = sync_lms_account(user, login)
       result['login_key']
+    end
+
+    def sync_lms_account(user, login)
+      username = lms_username(user)
+      email_address = lms_email(user)
+
+      # If we have a local login record, we have a talent account id. Check here first for the talent record
+      result ||= talentlms_user_record_by_id(login.lms_user_id) if login.present?
+      # The local login record does not exist OR Talent does not have an account for this id.
+      # Check for a Talent account with the assocaited email address
+      result ||= talentlms_user_record_by_email(email_address)
+      # Talent does not have a record associated with this email address, check the default username. If we generated an
+      # account for this user prior to allowing emails to be set, we should be able to find it this way.
+      result ||= talentlms_user_record_by_username(username)
+      # Talent does not have a record associated with this user, create an account for them.
+      result ||= create_account(user)
+
+      # Update the talent_lms_email on the local user if it differs from the email address in TalentLMS
+      if result.present? && result['email'] != user.talent_lms_email
+        user.talent_lms_email = result['email']
+        user.save!
+      end
+
+      # If we want to update the email in Talent to match the local account instead of updating the local email
+      # address to match what we are seeing in Talent. The line below will update the account in TalentLMS.
+      # Leaving this here in case we want to use this behavior.
+      ## result = @api.post('edituser', { user_id: result['id'], email: email_address }) if result.present? && result['email'] != email_address
+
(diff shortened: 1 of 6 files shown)


app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,111 @@ def initialize
...
+    end
+
+    def talentlms_user_record_by_username(username)
+      result = @api.post('users', { username: username })
+      result
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
+
+    def lms_username(user)
```
eanders, line 40: Can you add a brief comment describing what this is doing and specifically what the dev_offset is for?

app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,111 @@ def initialize
...
+      "#{ENV['RAILS_ENV']}_#{user.id + Integer(ENV.fetch('DEV_OFFSET', 0))}"
+    end
+
+    def lms_email(user)
+      email_address = "#{lms_username(user)}@#{ENV['FQDN']}"
+      email_address = user.talent_lms_email if user.talent_lms_email.present?
+      email_address
+    end
```
eanders:
```suggestion
    def lms_email(user)
      return user.talent_lms_email if user.talent_lms_email.present?
      
      "#{lms_username(user)}@#{ENV['FQDN']}"
    end
```

app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,111 @@ def initialize
...
+      result = sync_lms_account(user, login)
       result['login_key']
     end
 
+    def sync_lms_account(user, login)
+      username = lms_username(user)
+      email_address = lms_email(user)
+
+      # If we have a local login record, we have a talent account id. Check here first for the talent record
+      result ||= talentlms_user_record_by_id(login.lms_user_id) if login.present?
```
eanders: result will always be undefined here, so no need for the `||`:
```suggestion
      result = talentlms_user_record_by_id(login.lms_user_id) if login.present?
```

app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,111 @@ def initialize
...
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
+
+    def talentlms_user_record_by_username(username)
+      result = @api.post('users', { username: username })
+      result
+    rescue RuntimeError => e
+      raise e unless e.message.include?('The requested user does not exist')
+    end
```
eanders: I think the idea with naming methods in the facade was that they would be `lms_*` rather than specific to the LMS of the facade, that should mean you could rely on the names of the methods (or the API) to not change between facades.  Acknowledging it means rewriting code further down, can we do:
   ```suggestion
    def lms_find_user_by_id(id)
      result = @api.post('users', { id: id })
      result
    rescue RuntimeError => e
      raise e unless e.message.include?('The requested user does not exist')
    end

    def lms_find_user_by_email(email)
      result = @api.post('users', { email: email })
      result
    rescue RuntimeError => e
      raise e unless e.message.include?('The requested user does not exist')
    end

    def lms_find_user_by_username(username)
      result = @api.post('users', { username: username })
      result
    rescue RuntimeError => e
      raise e unless e.message.include?('The requested user does not exist')
    end
```

Comment (eanders) [reply app/models/talentlms/facade.rb]: alternatively, maybe these should be private and then I don't care as much.

File: app/models/talentlms/facade.rb [hunk 5]
```diff
@@ -10,42 +10,111 @@ def initialize
...
+      # account for this user prior to allowing emails to be set, we should be able to find it this way.
+      result ||= talentlms_user_record_by_username(username)
+      # Talent does not have a record associated with this user, create an account for them.
+      result ||= create_account(user)
+
+      # Update the talent_lms_email on the local user if it differs from the email address in TalentLMS
+      if result.present? && result['email'] != user.talent_lms_email
+        user.talent_lms_email = result['email']
+        user.save!
+      end
```
eanders: Nothing wrong with this, but maybe?
```suggestion
      user.update!(talent_lms_email: result['email']) if result.try(:[], 'email') != user.talent_lms_email
```

app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,111 @@ def initialize
...
+
+      # If we want to update the email in Talent to match the local account instead of updating the local email
+      # address to match what we are seeing in Talent. The line below will update the account in TalentLMS.
+      # Leaving this here in case we want to use this behavior.
+      ## result = @api.post('edituser', { user_id: result['id'], email: email_address }) if result.present? && result['email'] != email_address
```
eanders, line 92: This is a good question.  So if we have an email and Talent has an email, but they don't match, we preference our local email right now?  Would we ever be able to identify the user in talent if they had an account in Talent that wasn't created by us and our local email didn't match?

eanders (changes requested):
@dtgreiner this looks great! Just some minor tweaks and a question or two.

app/models/talentlms/facade.rb
dtgreiner: added in https://github.com/greenriver/hmis-warehouse/pull/4830/commits/8325cf3128615c0819c2a15ea6edafd7b64bc8ed

app/models/talentlms/facade.rb
dtgreiner: If we have an email and Talent has an email and they don't match, but the accounts were matched (by matching the `Talentlms::Login.lms_user_id` record to the TalentLMS account id or the `Talentlms::Login.login` record to the TalentLMS account username. The current setup will update the `User.talent_lms_email` record to what we are getting from the TalentLMS account email. We could push an edit to the TalentLMS account so it uses our stored email, but I defaulted to this way to keep any changes within our environment.

Regarding identifying a user in talent if we didn't create it and our email didn't match, that can be done but would require being able to set/edit what we have stored for the username (`Talentlms::Login.login`) or the TalemtLMS account id (`Talentlms::Login.lms_user_id`)

app/models/talentlms/facade.rb
dtgreiner: Updated with this commit https://github.com/greenriver/hmis-warehouse/pull/4830/commits/b4cced473718c2a9835eb18fe8cb27462b8f5821

app/models/talentlms/facade.rb
dtgreiner: Updated with this commit https://github.com/greenriver/hmis-warehouse/pull/4830/commits/8b874ba5c8f3716b7b723e6957881784d4471187

app/models/talentlms/facade.rb
```diff
@@ -10,42 +10,116 @@ def initialize
...
+    end
+
+    # Login username to TalentLMS
+    #
+    # ENV['DEV_OFFSET'] can be set to an integer to prevent username collisions between development environments
+    #
+    # @@param user [User] the user
+    # @return [String] username to be used for generated users in TalentLMS
+    def lms_username(user)
+      "#{ENV['RAILS_ENV']}_#{user.id + Integer(ENV.fetch('DEV_OFFSET', 0))}"
```
eanders: Is the integer useful in production environments?  If not, maybe just:
```suggestion
      username = "#{ENV['RAILS_ENV']}_#{user.id}"
      return username unless Rails.env.development?
      
      "#{username}_#{ENV.fetch('DEV_OFFSET', 0)}"
```

I think what you have will increment user ids rather than adding an increment at the end, which might be the intention, but seems odd, it would mean your user_id 1 is my user_id 2 potentially.

app/models/talentlms/facade.rb
eanders: Thanks for the explanation.  I think the behavior you have sounds correct for our scenarios.

app/models/talentlms/facade.rb
dtgreiner: This makes sense and definitely feels cleaner. I was using the existing logic, just pulled into its own method to prevent duplication. I'll push this change up.

eanders (approved):

eanders:
Thanks @dtgreiner !

PR #4834, opened 2024-10-16
6625 db index improvement

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/6625)

* migration to remove unused indexes from import/loader tables. Note, decided to remove all unused indexes here, regardless of size, to keep things simple
* adjust pg index stats to include fields for future unused-index removal. Also only collect stats on user defined indexes (pg_stat_user_indexes)

### Benefits
This should free up a substantial amount of space on our larger warehouse databases.

### Risks
I believe dropping indexes should not cause disruption however we should test this on QA; of particular concern would be if there is a spike in IOPs from the migration. I don't think this will occur but we should check.

I've checked against a good representative pg index stats to select only those that appear unused. However it's possible an active index is removed in-error. This could cause severe performance problems since the importer/loader tables are very large. It also follows that rebuilding one of these indexes could be an expensive task if we need to rollback.

It's worth noting that it may be possible to disable rather than remove these indexes, allowing us to simulate removal while retaining the ability to recover quickly. But it's not clear to me how well this is supported in postgres or what side effects there might be (corruption?). If it's worth pursuing, we could try `update pg_index set indisvalid = false where indexrelid = [unused idx]` [[ref](https://www.postgresql.org/docs/current/catalog-pg-index.html)]

Another way to hedge risk would be to remove only a few indexes at a time. But it might take quite a while to make our way through them all with our current release cadence.

## Type of change
Data cleanup, cost-savings

- [x I have run the code that is being changed under ideal conditions, and it doesn't fail


app/models/dba/unused_warehouse_index_migration_helper.rb
```diff
@@ -0,0 +1,67 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# This code is probably only used once to identify indexes that are safe for removal.
+require 'csv'
+module Dba
+  class UnusedWarehouseIndexMigrationHelper
```
ttoomey: maybe we don't want this in our code base. Including to show my work

app/models/app_resource_monitor/postgres_inspector.rb
```diff
@@ -116,23 +116,34 @@ def query_variables
...
+      psai.idx_tup_fetch::text AS tuples_fetched,
+      pg_relation_size(ipg.oid)::text AS index_size,
+      pg_get_indexdef(x.indexrelid) AS index_definition,
+      0 <> ALL (x.indkey) AS has_expression_index_column,
+      x.indisunique AS is_unique_index,
+      EXISTS (
+        SELECT 1
+        FROM pg_catalog.pg_constraint con
+        WHERE con.conindid = x.indexrelid
+      ) AS enforces_constraint
```
ttoomey, line 132: This could use another set of eyes. The idea is that these can assist with identifying unused indexes that are not used to enforce constraints

db/warehouse/migrate/20241010005805_prune_warehouse_indexes_phase_1.rb
```diff
@@ -0,0 +1,542 @@
...
+      JOIN pg_class t ON t.oid = i.indrelid
+      JOIN pg_class idx ON idx.oid = i.indexrelid
+      WHERE idx.relname = $1
+    SQL
+    result = GrdaWarehouseBase.connection.select_value(sql, 'SQL', [index_name])
+    raise "expected index #{index_name} to be on #{table_name} but was #{result}" unless result == table_name
+  end
+
+  # it seems that the index names aren't consistent between prod and dev. Attempt to remove the index by name only
+  def remove_index_safely(table:, index_name:, columns:)
```
yakloinsteak, line 42: Another layer of safety could be to double check here that there are zero scans (and maybe idx_tup_read = 0
  and idx_tup_fetch = 0). Those two columns might always be zero if there are zero scans, so it might be redundant.

app/models/app_resource_monitor/postgres_inspector.rb
```diff
@@ -116,23 +116,34 @@ def query_variables
   INDEX_STATS_SQL = <<~SQL.freeze
     SELECT
       %{db_name} AS database,
-      t.tablename,
+      c.relname AS tablename,
```
yakloinsteak, line 119: https://github.com/greenriver/hmis-warehouse/blob/stable/app/models/dba/database_bloat.rb?plain=1#L61
and related code in that file might be unified with this.

yakloinsteak (approved):
This is awesome! Thanks for doing this. Maybe we should talk briefly about the DatabaseBloat class I wrote. Automating the re-index part of it might be a relatively safe next step (vs using the repack and vacuum full features)

db/warehouse/migrate/20241010005805_prune_warehouse_indexes_phase_1.rb
```diff
@@ -0,0 +1,542 @@
...
+  #           SELECT
+  #             1
+  #           FROM
+  #             pg_catalog.pg_constraint con
+  #           WHERE
+  #             con.conindid = s.indexrelid
+  #         );
+  #  2) using our custom app inspector data, restrict the above set to only those that are reported as unused across a
+  #     a reasonable sample of our installations. See Dba::UnusedWarehouseIndexMigrationHelper
+  INDEX_TEXT = <<~TXT.freeze
```
eanders: I think this should be moved into `Dba::UnusedWarehouseIndexMigrationHelper` since future installations may need these removed as well.  I guess they won't be in the structure file so new installations won't get them, so maybe that's ok, but once we collapse migrations again, we'll lose this list.

ttoomey: Sure, I can move them.
I assume we want them to fade out eventually but if you think these indexes might get recreated somehow we can retain them for reference. Ideally we would track down whatever is creating these indexes and fix it

I'll add a todo-or-die to UnusedWarehouseIndexMigrationHelper so we remember to take out the garbage.

ttoomey:
> Maybe we should talk briefly about the DatabaseBloat class I wrote. Automating the re-index part of it might be a relatively safe next step (vs using the repack and vacuum full features)

Thanks @yakloinsteak - I put a short meeting on the calendar for next week when Elliot is back

eanders (approved):
Thanks @ttoomey !  Appreciate the move to the helper, I'm not sure we need them in the future, but I'd hate to lose them if we do.

PR #4835, opened 2024-10-16
Add system test for HUD form components

martha:
## _Merging this PR_
- ~Note that this PR targets [martha/6736-bulk-service-systest](https://github.com/greenriver/hmis-warehouse/pull/4829) rather than release-138. This was for my convenience only, since it was helpful to be able to copy/paste/modify my work on my previous PR. That PR should be merged first and then this one's base branch can be auto-updated. If this is inconvenient, I'm happy to take the time to tease them out.~

## Description
GH issue: https://github.com/open-path/Green-River/issues/6806

This PR adds system tests for the HUD components listed in the ticket.

~I think it also catches and fixes a bug in HOPWA question collection; see inline comment.~ (Removed to separate PR)

## Type of change
- Bug fix
- System test


drivers/hmis/app/graphql/types/hmis_schema/disability_group.rb
```diff
@@ -35,6 +35,7 @@ class HmisSchema::DisabilityGroup < Types::BaseObject
     field :t_cell_source, HmisSchema::Enums::Hud::TCellSourceViralLoadSource, null: true
     field :viral_load_available, HmisSchema::Enums::Hud::ViralLoadAvailable, null: true
     field :viral_load, Integer, null: true
+    field :viral_load_source, HmisSchema::Enums::Hud::TCellSourceViralLoadSource, null: true
```
martha: Bug, I think: Viral load source is not correctly collected. Even when "Medical report" is selected and then the intake is saved, the disability record's viral load source field is still set to 0.

Fix implemented here: Add the viral load source field to the disability object schema, so that the processor knows what enum in to use when we call `attribute_value_for_enum(graphql_enum(field), value)`.

Outstanding question: If we confirm this is a bug, we probably need to backfill data based on the raw form values stored in the processor. Let's create a separate ticket for that work to track it separately?

gigxz: Yeesh good find! Created https://github.com/open-path/Green-River/issues/6864

martha: Sounds good, I created #4840 and reverted the change here to prevent merge conflict. This PR's build will fail until that one is merged into release-138, I think.

martha: (interestingly the test failures I'm seeing on CI are not the same as the failure I would expect based on the above, so I'm retriggering it to see if that's something to investigate)

martha: On the test rerun I'm seeing the failures I expect to see. We should still make sure everything goes green once the other PRs are merged, but I think there is nothing else to investigate here

gigxz:
@martha it looks like the last build had a failure https://github.com/greenriver/hmis-warehouse/actions/runs/11389945825/job/31690479946

martha:
@gigxz 

> looks like the last build had a failure https://github.com/greenriver/hmis-warehouse/actions/runs/11389945825/job/31690479946

I think this system test failure was fixed in https://github.com/greenriver/hmis-warehouse/pull/4840. The most recent system test step succeeded: https://github.com/greenriver/hmis-warehouse/actions/runs/11505527598/job/32027451359?pr=4835

drivers/hmis/spec/system/hmis/form_hud_elements_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+    end
+  end
+
+  describe 'disabilities' do
+    context 'with project type of ES and no funders' do
+      let!(:p1) { create :hmis_hud_project, data_source: ds1, organization: o1, user: u1, funders: [] }
+
+      it 'renders limited disability component, only collecting overall disabling condition' do
+        assert_text 'Disabling Condition'
+        assert_no_text 'The following questions help determine if there are additional housing services or benefits available for the client.'
```
gigxz: I think we mainly want to test that the disability-specific questions are not asked (like physical disability, substance use, etc). This text doesn't really matter / may change or be removed over time, which shouldn't impact this test. 

```suggestion
        assert_no_text 'Physical Disability'
```

drivers/hmis/spec/system/hmis/form_hud_elements_spec.rb
```diff
@@ -0,0 +1,335 @@
...
+        mui_table_select('Yes', row: 'Physical Disability', column: 'Disabling Condition')
+        expect(mui_table_value_for(row: 'Overall Disabling Condition', column: 'Status')).to eq('Yes')
+
+        expect do
+          save_ignoring_warnings
+        end.to change(e1, :disabling_condition).to(1)
+        expect(e1.disabilities.where(disability_type: 5).sole.indefinite_and_impairs).to eq(1)
+      end
+
+      it 'updates overall Disabling Condition when you select an always-disabling condition' do
```
gigxz: So this is going to be kind of annoying, but, I would feel more comfortable if this test asserts on a few more things, for the sake of being confident in HUD compliance:

* [x] When you mark Developmental Disability as `yes`, the overall Disabling Condition becomes Yes **and it is disabled**. (Its important per data dictionary that the user can't override it to No)
  * [x] Then, when you clear or change Developmental Disability, overall Disabling Condition is no longer disabled and its cleared. (It's important that the field is interactable)
* [x] If you open the assessment an enrollment that already has a value for Enrollment.DisablingCondition it should appear, and then if you add developmental disability it should auto change to Yes
* [x] For the cases where you set just Physical Disability, adding an assertion to make sure Disabling Condition field is interactable (or add a step to set its value)
* [x] Add a test case that fills in all the disability values and submits

drivers/hmis/spec/system/hmis/form_hud_elements_spec.rb
```diff
@@ -0,0 +1,335 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+require 'rails_helper'
+require_relative '../../requests/hmis/login_and_permissions'
+require_relative '../../support/hmis_base_setup'
+
+RSpec.feature 'Hmis Form behavior for HUD elements', type: :system do
```
gigxz: Since this test pertains to HUD Assessments, suggest to change the file name to `hud_assessment_data_elements_spec`? I think of "form" as more generic to any type of form, whereas "assessment" is a type of form. So it is a little more specific and consistent with how we refer to assessments elsewhere.

gigxz (commented):
This is awesome, thank you!!!
I left a few small suggestions and request to add a few more cases for Disability.

PR #4836, opened 2024-10-16
Add can_view_units permission

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6605

This PR adds two new permissions
- `can_view_units` as noted in the ticket
- `can_manage_units` which is meant to be the same as current `can_manage_inventory`. I saw the todo to update this name and thought now would be a good time to do it, and figured this was the most straightforward path to rename, since either way we have to add a migration so that `can_manage_units` will be true for all roles where `can_manage_inventory` is currently true. I'm not sure about this though, let's discuss.

## Type of change
New permission

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - on the ticket https://github.com/open-path/Green-River/issues/6605#issuecomment-2418057759


drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -178,6 +182,8 @@ def unit_types
 
     # TODO use dataloader
     def units(**args)
+      raise HmisErrors::ApiError, 'Access denied' unless current_permission?(entity: object, permission: :can_view_units)
+
       resolve_units(**args)
     end
```
martha, line 189: I wasn't sure whether to add any permission gating on `has_units`. Searching in the frontend code, this doesn't seem to be used, but maybe I am missing it.

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -178,6 +182,8 @@ def unit_types
 
     # TODO use dataloader
     def units(**args)
+      raise HmisErrors::ApiError, 'Access denied' unless current_permission?(entity: object, permission: :can_view_units)
```
martha: Let's discuss whether to raise or just return empty?
- This is an attribute on the project object, so we don't want to raise and prevent someone from returning other project attrs that they do have permission to see
- However, from the frontend schema right now it is only used in purpose-built queries for the units page, so I figured that raising was the correct behavior

gigxz: I think return empty. That's our default for "field level authorization", and this feels similar. Also think about the "viewable_by" scopes which we use when resolving most fields like this. (In that case if the user has no view access, it's just empty).
I can imagine a future page where maybe the Project Overview page has a summary of unit availability stats. We would want to be able to just make a single query for that may include units, and not worry about that throwing an error if the user lacks access to view it.

drivers/hmis/app/graphql/types/hmis_schema/project.rb
gigxz: Hmm not sure if/where it's used, but I think it can remain as-is

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -162,6 +164,8 @@ def affiliated_projects
 
     # Build OpenStructs to resolve as UnitTypeCapacity
     def unit_types
+      return {} unless current_permission?(entity: object, permission: :can_view_units)
```
gigxz: If there are values, this method returns an array. I guess it's not causing problems, but I think it's clearer to return the same type `[]` instead of an empty object.

gigxz (approved):

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -8602,12 +8602,13 @@ type ProjectAccess {
   canManageDeniedReferrals: Boolean!
   canManageExternalFormSubmissions: Boolean!
   canManageIncomingReferrals: Boolean!
-  canManageInventory: Boolean!
```
gigxz, line 8605: This will be a breaking change: if the frontend is on an older version that is trying to fetch canManageInventory, and then we deploy this, that request will raise an exception. My suggestion on the frontend PR was to stop resolving this field, so that we can safely remove it in a future release. Sorry for the confusion!
As of yet the only documentation we have around this is in the Runbook, but maybe it should be moved to github somewhere and expanded upon.

gigxz (approved):

drivers/hmis/app/graphql/schema.graphql
martha: Thanks Gig for cleaning this up - I was rushing and didn't think it through. Thanks for the fix!

martha: @gigxz let's look at this [build fail](https://github.com/greenriver/hmis-frontend/actions/runs/11470351871) introduced here together tomorrow. I spent some time today puzzling over [PermissionFilter here](https://github.com/greenriver/hmis-frontend/blob/release-139/src/modules/permissions/PermissionsFilters.tsx), but wasn't quite able to figure it out.

It is the same build fail I encountered [a couple commits ago](https://github.com/greenriver/hmis-frontend/pull/963/commits/346d6f2d75f159766c4c6eac66c73f53548d0a7e), which I attempted to fix in my next commit by just [ripping out `canManageInventory` entirely](https://github.com/greenriver/hmis-frontend/pull/963/commits/8846cc862513a979075cfddd007e5fbb84708cb2). You rightly pointed out that that's not backwards compatible, and needs to be reverted, but the build fail reappears.

We could just `ts-ignore` it - maybe that's the best solution?

martha: paper-trail - opened https://github.com/greenriver/hmis-frontend/pull/970

PR #4837, opened 2024-10-16
Upload Client Headshot without callbacks; set user_id correctly

gigxz:
## _Merging this PR_

## Description

Issue: https://github.com/open-path/Green-River/issues/6863

1. Fix a bug where client headshot upload would fail in HMIS for migrated-in clients with bad (or missing) `UserID` values. [Bug repro steps](https://github.com/open-path/Green-River/issues/6863#issuecomment-2417986739).
2. As part of this fix, I changed the approach for uploading client headshots so it no longer happens in an `after_save` hook. Instead, the File is built directly in the ClientProcessor.
3. Fix file preview issue in dev that stemmed from a [change](https://api.rubyonrails.org/v7.1/classes/ActiveStorage/Blob.html#method-i-url) the the blob api 

For QA, are 3 places in HMIS where client files can be uploaded:
- When creating a new client (`/client/new`)
- Direct upload on client dashboard (`/client/1/overview`)
- Direct upload a file (`/client/1/files`) - this doesn't make it a "headshot" on the HMIS profile, even if is tagged as "client headshot". It should be unrelated to this code, but should test/qa anyway.

## Type of change


martha (approved):
Looks good to me!

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -369,6 +349,23 @@ def full_name
...
+
+    # Note: this builds a GrdaWarehouse::ClientFile where the associated client is a Hmis::Hud::Client,
+    # (not a GrdaWarehouse::Hud::Client), which may lead to unexpected behavior. Skip callbacks to avoid issues.
+    file = client_files.build(
+      user_id: current_user.id,
+      name: 'Client Headshot',
+      visible_in_window: false,
+    )
+    file.callbacks_skipped = true
+    file.tag_list.add('Client Headshot') # Special string in ClientFileBase, don't change
```
eanders: Let's make this a class method in `ClientFileBase`
```
def self.headshot_tag_name
  'Client Headshot'
end
```

then this become:
```suggestion
    file.tag_list.add(ClientFileBase.headshot_tag_name)
```

That should keep us in sync in the future.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1212,6 +1212,22 @@ def process_record(record:, hud_values:, user:, definition:, save: true)
...
+      let!(:file) { File.open('drivers/hmis/spec/fixtures/files/client_photo_00001.jpg') }
+      let(:blob) { ActiveStorage::Blob.create_and_upload!(io: file, filename: 'client_photo_00001.jpg', content_type: 'image/jpeg') }
+      let(:hud_values) { complete_hud_values.merge('imageBlobId' => blob.id.to_s) }
+
+      it 'creates a ClientFile tagged as a client headshot' do
+        client = Hmis::Hud::Client.new(data_source: ds1, user: u1)
+        process_record(record: client, hud_values: hud_values, user: hmis_user, definition: definition)
+
+        file = client.client_files.first
+        expect(file.name).to eq('Client Headshot')
```
eanders: this one too

eanders (approved):
@gigxz I think this looks good, just the one comment to try and prevent future drift.

drivers/hmis/app/models/hmis/hud/client.rb
eanders: I think the suggestion is slightly off since `ClientFileBase` is a module

gigxz: Thanks yeah this became `file.tag_list.add(GrdaWarehouse::ClientFile.headshot_tag_name)` on https://github.com/greenriver/hmis-warehouse/pull/4837/commits/127b8adf6b76cea328c4fa8976dc042ab7986e39

PR #4842, opened 2024-10-18
6625 pg hero

ttoomey:
## _Merging this PR_

## Description
Add pghero gem for db monitoring. Not enabled on prod instances. Accessible to users with can_manage_config permission

To test, log-in to account with required perms, visit '/pghero'. There's intentionally no link to it, it's intended for engineers.

Details 
* Enabled historic query and space tracking, to enable, add`pg_stat_statements` pg extension per setup guide. [ref guide](https://github.com/ankane/pghero/blob/master/guides/Rails.md#historical-query-stats)

Future
* pghero may be able to read RDS stats out of cloudwatch if we provide 'aws_db_instance_identifier'

## Type of change
Monitoring/infra


yakloinsteak (approved):
We usually have the `pg_stat_statements` extension installed in case that's useful to you. Feel free to create a DevOps issue if you find the app needs more AWS permissions to get what it needs.

ttoomey:
I went ahead and enabled historic tracking. It should detect pg_stat_statements and run appropriately, collecting stats hourly

lib/tasks/grda_warehouse.rake
```diff
@@ -342,6 +342,8 @@ namespace :grda_warehouse do
     AppResourceMonitor::CollectStatsJob.perform_later if stats_collector.should_enqueue?
 
     BuildTranslationCacheJob.perform_later
+
+    PgheroCollectStatsJob.perform_later(clean: DateTime.current.hour == 5) if !Rails.env.production? && PgHero.query_stats_enabled?
```
ttoomey, line 346: collecting stats on the hour seem okay? I don't know how frequently we should run this

yakloinsteak: Do you know the volume of data involved? If it's storing information about all tables or all query shapes, it might be big and slow. I imagine it won't be intensive on the database, so hourly is probably fine I'd guess.

config/pghero.yml
```diff
@@ -0,0 +1,44 @@
+<% require File.expand_path('../lib/util/rails_db_uri_extractor', __FILE__) %>
+databases:
+  warehouse:
+    # try an extract the uri from the rails db config
```
eanders:
```suggestion
    # try to extract the uri from the rails db config
```

config/pghero.yml
```diff
@@ -0,0 +1,44 @@
...
+# explain: true / false / analyze
+
+# Statement timeout for explain
+explain_timeout_sec: 10
+
+# Visualize URL for explain
+# visualize_url: https://...
+
+# Time zone (defaults to app time zone)
+# time_zone: "Pacific Time (US & Canada)"
```
eanders, line 29: Should we pull this from `ENV['TIMEZONE']`?

eanders (approved):
Thanks @ttoomey, looks like it works as designed and should prove useful!

lib/tasks/grda_warehouse.rake
ttoomey: Good point, I believe the max number of queries tracked by pg_stat_statements is 5000 by default (`SHOW pg_stat_statements.max`). Of course, the sql statements themselves could be quite long. My intuition is this shouldn't be a performance issue. We could keep an eye on staging after this is deployed and tune or disable stats collection if it seems to be a problem

PR #4847, opened 2024-10-18
Ability to disable client auto-deduplication

eanders:
## _Merging this PR_

## Description
Provides the ability to completely disable client de-duplication even if PII is an exact match.

## Type of change
* New feature


dtgreiner (approved):

app/views/admin/configs/_client_calculations.haml
```diff
@@ -1,6 +1,7 @@
 %section.o-section-card
   %h3 Client Calculations
   .c-card.c-card--flush.c-card--padded.flex-wrap
+    = f.input :enable_auto_deduplication, label: 'Automatically de-duplicate and merge clients', hint: 'If clients share 2 of 3 of the following, the warehouse will join the records for reporting purposees.  Name, DOB, valid SSN.'
```
ttoomey: Maybe a little clearer:
```suggestion
    = f.input :enable_auto_deduplication, label: 'Automatically de-duplicate and merge clients', hint: 'Records will be joined for reporting purposes when clients share at least 2 of these 3 identifiers:  Name, DOB, valid SSN.'
```

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -7,123 +7,196 @@
...
-    expect(check_for_obvious_match(client_in_source)).to be client_in_destination.id
-  end
-
-  describe 'obvious match processing with a split' do
-    let!(:split) { create :grda_warehouse_client_split_history, split_from: client_in_source.id, split_into: client_in_destination.id }
-
-    it 'does not return an obvious match if it was split' do
-      expect(check_for_obvious_match(client_in_source)).to be_nil
+  describe 'When matching is enabled' do
+    before(:all) do
```
ttoomey: Can this be in a before each rather than all? That way db changes will occur within a transaction and not impact other tests. Also probably cleanup cache on the way out?

```suggestion
    after(:each) { @config.invalidate_cache }

    before(:each) do    
```

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -7,123 +7,196 @@
...
+
+        it 'sees two destination clients after split' do
+          expect(GrdaWarehouse::Hud::Client.destination.count).to eq(2)
+        end
+
+        it 'does not re-merge the split clients' do
+          GrdaWarehouse::Tasks::IdentifyDuplicates.new(run_post_processing: false).match_existing!
+
+          expect(GrdaWarehouse::Hud::Client.destination.count).to eq(2)
+        end
```
ttoomey: This is fine; it looks like we have the pattern in these tests of processing in a before block. Maybe it's too much trouble but perhaps it would be more complete to move the processing in an expect block to verify the values are actually changing before and after the split?

```suggestion
        def split_client
          destination_client = client_in_source.destination_client
          split_client_id = destination_client.source_clients.last.id
          destination_client.split(
            [split_client_id],
            destination_client.id,
            destination_client.id,
            user,
          )
        end
        
        let(:destination_scope) { GrdaWarehouse::Hud::Client.destination }

        it 'sees two destination clients after split' do
          expect { split_client }.to change(destination_scope, :count).from(1).to(2)
        end
        
        it 'does not re-merge the split clients' do
          expect do
            split_client
            GrdaWarehouse::Tasks::IdentifyDuplicates.new(run_post_processing: false).match_existing!
          end.to change(destination_scope, :count).from(1).to(2)
        end
```

ttoomey (approved):
looks good, thanks for updating the tests!

PR #4852, opened 2024-10-21
Transfer Move-in Date when HoH changes

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/6857

* When changing the HoH in HMIS, transfer the Move-in Date and Move-in Address from the old HoH onto the new HoH.
* If the new HoH entered after the household moved in, their Entry Date gets set as their Move-in Date (AAQ is pending)
* Also, clear out any existing Move-in Dates for non-HoH members, because they shouldn't be present.

I moved the HoH-change logic into a new class, rather than keeping it in the mutation, because it's something we may want to run in a bulk operation in the future. (For example, fix all households where HoH is a child, for example).

## Type of change
Bug fix / HUD compliance


drivers/hmis/app/graphql/mutations/update_relationship_to_ho_h.rb
```diff
@@ -13,82 +13,24 @@ class UpdateRelationshipToHoH < BaseMutation
...
-      # Set new relationship value
-      enrollment.relationship_to_ho_h = relationship_to_ho_h
-      # Set user HUD that most recently touched the record
-      enrollment.user_id = hmis_user.user_id
-
-      # Return if there are any AR errors (not expected)
-      unless enrollment.valid?
-        errors.add_ar_errors(enrollment.errors&.errors)
-        return { errors: errors }
-      end
```
gigxz, line 77: I don't think this was necessary, so updated to just raise if something fails on `save!`. These validations won't be user-fixable anyway most likely

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -0,0 +1,113 @@
...
+  # The HohChangeHandler class is responsible for handling changes to the Head of Household (HoH) in a household enrollment.
+  # It provides methods to perform the change, apply the changes to the database, and validate the changes.
+  #
+  # This logic is split out from the UpdateRelationshipToHoH mutation in case we need to use it in cleanup routines,
+  # for example bulk assigning HoH to the oldest adult in a household.
+  #
+  # Parameters:
+  # - new_hoh_enrollment: The Enrollment that is becoming the Head of Household
+  # - hud_user_id: The "UserID" of the HUD user making the change.
+  class HohChangeHandler
```
gigxz, line 12: Is this the right location/naming for this? We have some other non-model classes in `drivers/hmis/app/models` too, like the enrollment reminder generator.

drivers/hmis/app/graphql/mutations/update_relationship_to_ho_h.rb
```diff
@@ -13,82 +13,24 @@ class UpdateRelationshipToHoH < BaseMutation
...
-        "#{new_hoh_name} will be the Head of Household."
-      end
-    end
-
     def resolve(enrollment_id:, enrollment_lock_version: nil, relationship_to_ho_h:, confirmed: false)
       enrollment = Hmis::Hud::Enrollment.viewable_by(current_user).find_by(id: enrollment_id)
-      raise 'Not found' unless enrollment.present?
-      raise 'Access denied' unless current_user.permissions_for?(enrollment, :can_edit_enrollments)
+      access_denied! unless enrollment.present?
+      access_denied! unless current_user.permissions_for?(enrollment, :can_edit_enrollments)
```
ttoomey: present is not needed here, this will be nil if the record is not found. Maybe golf down?
```suggestion
      access_denied! unless enrollment && current_user.permissions_for?(enrollment, :can_edit_enrollments)
```

drivers/hmis/app/models/hmis/hoh_change_handler.rb
ttoomey: This is fine with me; I don't mind mixing PORO classes together with AR models.

drivers/hmis/spec/requests/hmis/update_relationship_to_ho_h_spec.rb
```diff
@@ -64,6 +64,112 @@
...
+      }
+    end
+
+    def perform_mutation
+      response, result = post_graphql(input: input) { mutation }
+      expect(response.status).to eq(200), result.inspect
+      [hoh, hhm, hhm2].map(&:reload)
+    end
+
+    context 'when new HoH enetered before move-in' do
```
ttoomey:
```suggestion
    context 'when new HoH entered before move-in' do
```

drivers/hmis/spec/requests/hmis/update_relationship_to_ho_h_spec.rb
```diff
@@ -64,6 +64,112 @@
...
+      it 'should transfer move-in date to new HoH' do
+        perform_mutation
+
+        # old HoH is cleared
+        expect(hoh.relationship_to_ho_h).to eq(99)
+        expect(hoh.move_in_date).to be_nil
+        # new HoH inherited move-in date
+        expect(hhm.relationship_to_ho_h).to eq(1)
+        expect(hhm.move_in_date).to eq(hoh_move_in_date)
+      end
```
ttoomey: this is fine but maybe use expect blocks to verify the values are changing in perform_mutation (not tested)
```ruby
      it 'should transfer move-in date to new HoH' do
        expect { perform_mutation }.to change { hoh.relationship_to_ho_h }.from(1).to(99)
          .and change { hoh.move_in_date }.from(hoh_move_in_date).to(nil)
          .and change { hhm.relationship_to_ho_h }.from(99).to(1)
          .and change { hhm.move_in_date }.from(nil).to(hoh_move_in_date)
      end
```

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -0,0 +1,113 @@
...
+
+      # HoH shouldn't be a child, unless all members are children
+      new_hoh_age = new_hoh_enrollment.client.age
+      add_warning(self.class.child_hoh_message) if new_hoh_age.present? && new_hoh_age < 18 && household_enrollments.find(&:adult?)
+
+      validation_errors
+    end
+
+    def apply_changes!
+      Hmis::Hud::Enrollment.transaction do
```
ttoomey: It's debatable but I wonder if it would be better to move the transaction to the caller. In a batch context you'd probably want one tx wrapping multiple calls to apply_changes.

Also I suggest adding a lock here for safety. Probably easiest to lock the project since this spans all enrollments. Maybe something like this in the mutation
```ruby
  hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user_id: 
  hmis_user.user_id)
  # use lock to avoid duplicate request collisions
    validations = hoh_changer.validate(include_warnings: !confirmed)
    return { errors: validations } if validations.any?
    hoh_changer.apply_changes!
  end
```

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -0,0 +1,113 @@
...
+            hhm.relationship_to_ho_h = 99
+            # Move-in Address(es) from old HoH should transfer to new HoH. We only expect 1, but it's OK if there are more.
+            hhm.move_in_addresses.each { |addr| addr.update!(enrollment: new_hoh_enrollment) }
+          end
+
+          # Clear Move-in Date on non-HoH members
+          hhm.move_in_date = nil
+
+          if hhm.changed?
+            hhm.user_id = hud_user_id # set user who lasted touched the record
```
ttoomey:
```suggestion
            hhm.user_id = hud_user_id # set user who last touched the record
```

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -0,0 +1,113 @@
...
+    attr_accessor :new_hoh_enrollment, :household_enrollments, :old_hoh_enrollments, :new_hoh_move_in_date, :old_hoh_move_in_date, :validation_errors, :hud_user_id
+
+    def initialize(new_hoh_enrollment:, hud_user_id:)
+      self.new_hoh_enrollment = new_hoh_enrollment
+      self.hud_user_id = hud_user_id
+
+      # All Enrollments in the household (including new and old HoHs)
+      self.household_enrollments = new_hoh_enrollment.household_members.preload(:client)
+      # Previous HoHs (should only be one, but we handle multiple for data cleanup)
+      self.old_hoh_enrollments = household_enrollments.sort_by(&:DateCreated).select(&:head_of_household?)
```
ttoomey: I think there should be a tie breaker, using id. Also I prefer `filter` rather than `select` with arrays to disambiguate from active_record's `select` method
```suggestion
      self.old_hoh_enrollments = household_enrollments.
        filter(&:head_of_household?).
        sort_by { |e| [e.DateCreated, e.id] }
```

ttoomey (approved):
Looks great! Left a few notes

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: updated, thanks!

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: I pushed a change like this https://github.com/greenriver/hmis-warehouse/pull/4852/commits/caf357b7b70a76fab78441e025785110cc0e8e2c :
```ruby
        Hmis::Hud::Enrollment.transaction do
          # Lock to avoid duplicate request collisions
          enrollment.household_members.lock!

          hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user_id: hmis_user.user_id)
          validations = hoh_changer.validate(include_warnings: !confirmed)
          return { errors: validations } if validations.any?

          hoh_changer.apply_changes!
        end
```
Thoughts? We only really need to lock the household members, not the entire project, which seems potentially disruptive. Of course there is the chance that an additional member could be added to the household, which this would miss, but that seems worth it to avoid locking the whole project imo. What do you think?



For my own reference, how I tested the behavior manually:
```ruby
# Process 1
Hmis::Hud::Enrollment.transaction do
  enrollment = Hmis::Hud::Enrollment.find(60298)
  enrollment.household_members.lock!

  enrollment.household_members.each do |member|
    member.move_in_date = Date.current
    member.save!
  end
  sleep 20
end

# Process 2 (waits and then fails with StaleObjectError)
Hmis::Hud::Enrollment.transaction do
  enrollment = Hmis::Hud::Enrollment.find(60298)
  enrollment.update!(move_in_date: 1.year.ago)
end

# Process 3 (doesn't wait, not part of household)
Hmis::Hud::Enrollment.transaction do
  enrollment = Hmis::Hud::Enrollment.find(51128)
  enrollment.update!(move_in_date: 1.year.ago)
end
```

ttoomey: Thanks Gig

I would caution against locking more than one record (enrollment.household_members). The reason being that it might pose a risk of creating db deadlocks which can lead to app reliability issues. If you only lock one record, you avoid this issue completely. Note, if household_members is ordered by id, this becomes much less prone to deadlocks (but still a possibility).

I guess I'm not sure that locking the project would be disruptive in this case. The default lock is 'select for update' meaning that, while attempts to lock or update the project will block, reads are non-blocking. It seems like this operation should be quick so I don't know that this be much of an issue in practice but I might be missing something. I guess if there was some pathological case of a household containing thousands of enrollments this would be an issue.

Generally my thought is it would be nice to have some way to synchronize operations on an entire household and the project seems most natural for this, assuming we can keep our operations performant.

Happy to huddle about this to discuss further

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -0,0 +1,114 @@
...
+
+      # All Enrollments in the household (including new and old HoHs)
+      self.household_enrollments = new_hoh_enrollment.household_members.preload(:client)
+      # Previous HoHs (should only be one, but we handle multiple for data cleanup)
+      self.old_hoh_enrollments = household_enrollments.
+        filter(&:head_of_household?).
+        sort_by { |en| [en.DateCreated, en.id] }
+
+      self.old_hoh_move_in_date = old_hoh_enrollments.map(&:move_in_date).compact.first
+      self.new_hoh_move_in_date = [old_hoh_move_in_date, new_hoh_enrollment.entry_date].max if old_hoh_move_in_date
```
eanders: Since this is the move-in date for the household, I think you want the min, and might possibly have to move the new HoH entry date back so that they are present for the entire household enrollment.  That's what we're supposed to do for reporting, but there has been acknowledgment that it doesn't reflect reality.  Relatedly, annual assessments are on the anniversary of the HoH entry date, so if you change that, previous annual assessments will have been corrected outside of the window and future assessments will be on a different cadence.

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: @ttoomey thanks for the feedback and explanation. Makes sense that it could deadlock, and that project locking shouldn't be disruptive provided that this operation is relatively fast (unless household size is huge like you say).
I adjusted  it to use `with_lock` on the project.

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: responded on issue https://github.com/open-path/Green-River/issues/6857#issuecomment-2437677775

eanders (approved):

gigxz:
Still need to update this to NOT transfer in the case when new HoH entered after MID

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: OK, https://github.com/greenriver/hmis-warehouse/pull/4852/commits/e38d819dac5dbca42638a7d5ba0b147bb0b9d54a should reflect the change discussed at standup. We do NOT transfer Move-in Date in the scenario where the new HoH entered after move-in. This ends up in a "bad data" scenario, because the non-HOh still has a move-in date. Plan to adjust pending HUD guidance, so left a bunch of TODOs.

gigxz:
Ok this is ready to merge imo @eanders . I would be comfortable with it going into release-139

PR #4854, opened 2024-10-23
6866-supplemental-data-fixes

ttoomey:
## _Merging this PR_

## Description
[GH issue](https://github.com/open-path/Green-River/issues/6866)

Note, I opened this against 138 in case we want to get in in earlier than 139

* fix bug where all source clients were visible, should be restricted to match data_set
* add toggle to control nightly syncs

## Type of change

### DIFF
--- a/drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
+++ b/drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
@@ -47,8 +47,8 @@ module HmisSupplemental
 
     def source_enrollments
       @client.source_enrollments.
-        where(data_source_id: @data_set.data_source_id).
         visible_to(current_user).
+        where(data_source_id: @data_set.data_source_id).
         order(entry_date: :desc, id: :desc).
         preload(:project)
     end
@@ -58,9 +58,10 @@ module HmisSupplemental
     end
 
     def client_groups
+      # note: order is important here, source_visible_to appears to clobber the data source condition
       clients = @client.source_clients.
-        where(data_source_id: @data_set.data_source_id).
         source_visible_to(current_user).
+        where(data_source_id: @data_set.data_source_id).
         order(:id)
       clients.map do |client|
         {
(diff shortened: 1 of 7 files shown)


drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
```diff
@@ -58,9 +58,10 @@ def load_authorized_data_set
     end
 
     def client_groups
+      # note: order is important here, source_visible_to appears to clobber the data source condition
```
eanders, line 61: If `source_visible_to` clobbers the data source id, don't we need to check that `@data_set.data_source_id` is in `source_visible_to(current_user).select(:data_source_id)`?  Is there a possibility you could get here without having that check performed on `@data_set` first?

eanders: if you use `where(c_t[:data_source_id].eq(@data_set.data_source_id))` does the clobber still happen?  I think we use the arel versions pretty heavily in the `source_visible_to` scope, so maybe there's an odd interaction with the active record version. (Not sure that's a better solution, but it might mean the order isn't important).

ttoomey: > if you use where(c_t[:data_source_id].eq(@data_set.data_source_id)) does the clobber still happen? 

Good idea but unfortunately I do see the same order-dependent behavior using arel. Glass-half full is that AR is consistent at least.

> If source_visible_to clobbers the data source id, don't we need to check that @data_set.data_source_id is in source_visible_to(current_user).select(:data_source_id)?

Good point, we do check that data set is visible to user but not the data source. It does seem possible that a user could see the data set but not the data source. I guess the data set permission doesn't imply permission to see the client data?

I'll update `DataSet.viewable_by` to reflect this assumption; checking for `can_view_supplemental_client_data for` on the data source as well. So then the permission to see the data set tabs are:
- user must be using acls
- user needs `can_view_supplemental_client_data` on a collection that includes the data set
- user needs `can_view_supplemental_client_data` on the associated data_source as well

eanders (approved):

drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
eanders: Let's give that a go and see if the data source is also necessary long-term.  Feels safer.

PR #4856, opened 2024-10-23
Disallow editing HUD service types

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6831

This PR
- Prevents creating a new service type in a HUD category
- Prevents deleting a HUD service type
- Prevents updating a HUD service type
- Adds a new picklist for custom service categories only, so that the modal for creating a new service type can use that picklist. (The "All service categories" picklist is still used by Form Rules, so a separate one is needed here.)
  - Adds a new scope `non_hud` on the Category model in support of the new picklist

## Type of change
Bug fix (feature functionality cleanup?)

  - [x] ensured this does not introduce N+1s - I tried to prevent introducing n+1s to the new picklist; it does perform 2 queries, though, and I'd be glad for feedback to improve it.


drivers/hmis/app/graphql/mutations/create_service_type.rb
```diff
@@ -15,6 +15,11 @@ class CreateServiceType < CleanBaseMutation
     def resolve(input:)
       access_denied! unless current_user.can_configure_data_collection?
 
+      category = Hmis::Hud::CustomServiceCategory.find(input.service_category_id)
+
+      # Can't add a custom service to a HUD service category
+      access_denied! if category.service_types.any?(&:hud_service?)
```
martha: Let's discuss - should this be `any` or `all`?

drivers/hmis/app/graphql/mutations/delete_service_type.rb
```diff
@@ -15,8 +15,15 @@ def resolve(id:)
 
       service_type = Hmis::Hud::CustomServiceType.find(id)
 
-      # TODO: Eventually this should be a user-facing ValidationError returned in the {errors:} object
-      raise 'Cannot delete a service type that has services' if service_type.custom_services.exists?
+      # Can't delete HUD service type
+      access_denied! if service_type.hud_service?
```
martha: To confirm - is `access_denied` the right error to throw in these mutations? I used it bc the UI will prevent these actions, so the spirit feels correct: if the user tries to do this, they are attempting something they are not allowed to do. However, it could be confusing because it's not a permissions issue - no user has access to do this.

gigxz: I think it's usually preferable to have a descriptive error message, if only for code readability. We also get the benefit of grouping separately from "access denied" in Sentry, which we really only use for cases where the user lacks permission for the operation/object.

I suppose I can see the argument for this being a denial of "access" because we're just saying '**nobody** has access to delete HUD service types' implicitly in the code. OTOH, it doesn't really feel like an "access" check if there are NO users that can do it. Does that make sense?

In general I'd say it's great to include a descriptive message if you have something to say. In this case, for example, a descriptive error msg would remove the need for having a comment above it.

drivers/hmis/app/graphql/mutations/update_service_type.rb
```diff
@@ -17,6 +17,11 @@ def resolve(id:, name:, supports_bulk_assignment:)
       access_denied! unless current_user.can_configure_data_collection?
 
       service_type = Hmis::Hud::CustomServiceType.find(id)
+
+      # Can't update HUD services. This prevents users from unknowingly renaming/reusing services for another purpose
+      # while the service continues to collect HUD records for the original type
+      access_denied! if service_type.hud_service?
```
gigxz: same here, prefer descriptive error message instead of comment

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
```diff
@@ -18,6 +18,15 @@ class Hmis::Hud::CustomServiceCategory < Hmis::Hud::Base
 
   validates_presence_of :name, allow_blank: false
 
+  scope :non_hud, -> do
+    hud_service_category_ids = Hmis::Hud::CustomServiceCategory.
+      left_joins(:service_types).
+      where.not(service_types: { hud_record_type: nil }).
+      distinct
```
gigxz: I think you're missing a `pluck(:id)`?
Apparently Rails is happy with `where(id: scope)` but it doesn't look like that was the intention.

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
```diff
@@ -18,6 +18,15 @@ class Hmis::Hud::CustomServiceCategory < Hmis::Hud::Base
 
   validates_presence_of :name, allow_blank: false
 
+  scope :non_hud, -> do
```
gigxz, line 21: Despite the left join, this will exclude empty categories because of the `IS NOT NULL` clause:
```
"SELECT \"CustomServiceCategories\".* FROM \"CustomServiceCategories\" LEFT OUTER JOIN \"CustomServiceTypes\" \"service_types\" ON \"service_types\".\"DateDeleted\" IS NULL AND \"service_types\".\"custom_service_category_id\" = \"CustomServiceCategories\".\"id\" WHERE \"CustomServiceCategories\".\"DateDeleted\" IS NULL AND \"service_types\".\"hud_record_type\" IS NOT NULL"
```

Maybe a better way would be to pick any custom service category that is empty or has any service with a null record_type. That would also make this match the "all" logic in the mutation (see other comment). In other words, exclude all the HUD-only service categories. Ex:
```
irb(main):032:0> Hmis::Hud::CustomServiceCategory.left_joins(:service_types).where(service_types: { hud_record_type: nil }).to_sql
=> "SELECT \"CustomServiceCategories\".* FROM \"CustomServiceCategories\" LEFT OUTER JOIN \"CustomServiceTypes\" \"service_types\" ON \"service_types\".\"DateDeleted\" IS NULL AND \"service_types\".\"custom_service_category_id\" = \"CustomServiceCategories\".\"id\" WHERE \"CustomServiceCategories\".\"DateDeleted\" IS NULL AND \"service_types\".\"hud_record_type\" IS NULL"

```

drivers/hmis/app/graphql/mutations/create_service_type.rb
gigxz: If I had to pick I would say "All", so we can continue supporting adding custom service types to the existing categories that have both HUD and custom (like "Check-in" that you found).

Although, I think it's simpler to just remove this check. It feels a little sticky that this logic and the picklist logic need to match (because we don't want the picklist to present an option that the mutation will reject). Nothing "bad" happens if there is a custom service type in the same category as HUD service type. - and as you pointed out we already have one like that.

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
gigxz: To make this suggestion a little clearer: it would support frontend users adding new customer service types to the "Check In" category example (which has both bed night and day shelter check-in). But it would not support adding customer services to the "stock" HUD service categories that are set up initially (that happens here https://github.com/greenriver/hmis-warehouse/blob/f06264d238d94eccc73ba50b6fe6b498c44efbdd/drivers/hmis/lib/hmis_util/service_types.rb#L17-L21)

drivers/hmis/app/graphql/mutations/create_service_type.rb
martha: Ok, makes sense. I will remove the check.

I was thinking that the "bad" thing that happens when there's a custom service type in the same category as a HUD service type, is that that category is now displayed in the frontend as "Custom" instead of "HUD", because of the logic [here](https://github.com/greenriver/hmis-warehouse/blob/341367e3474166cd329b3b263ac7dd1212f26d83/drivers/hmis/app/graphql/types/hmis_schema/service_category.rb#L28), which is a bit misleading.

However, it looks to me like that can safely be removed now that we are removing the separate Category interface in the [next pr](https://github.com/greenriver/hmis-frontend/pull/969)!

Edit: I proposed that deprecation [here](https://github.com/greenriver/hmis-warehouse/pull/4859/commits/0f6720451bb910a65c255b1165c344c5a6093b42) and [here](https://github.com/greenriver/hmis-frontend/pull/969/commits/bd92cb2c7bd636204b28a6572d1c35d881394e4c), we can discuss on those PRs.

drivers/hmis/app/graphql/mutations/delete_service_type.rb
martha: Ok great, makes sense, thanks!

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
martha: Thanks yes 🤦‍♀️

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
martha: Okay, thank you for the added detail. Let me know if you would like to re-add the check on the mutation.

martha: How did the "Check In" example get the way it is? Did it exist as a seeded category and then have a custom type added to it? The ui changes we are introducing in this PR would prevent more examples like that from cropping up, right? Or did we add it manually and configure it that way on purpose?

gigxz: It was set up manually in the rails console. I probably renamed the existing bed night category. Yes it was on purpose to have "bed night" and "check-in" in the same group, but I think it's fine to have the UI disallow that going forward. All the custom categories have been set up in the console since we don't have a UI for it yet.

gigxz (approved):

drivers/hmis/app/models/hmis/hud/custom_service_category.rb
martha: Thanks! Just to clarify,

> All the custom categories have been set up in the console since we don't have a UI for it yet

We don't have a UI for creating a custom category, but we have an existing UI for adding a custom type to a HUD category, thereby converting that category to a "Custom" category.
- Visit the Services admin UI, click New Service Type, and create a new custom service type
- Pick a HUD service category in the dropdown
- In the Categories tab, you can see that that category changed its "Type" column from "HUD Services" to "Custom Services." (In QA you can see that I just did that this morning with the "Bed Night" category when I added the new type "Custom bed night")

This is what I was asking about when I asked how the "Check In" example got the way it is -- did we do it intentionally on the console, or did the client do it (possibly unwittingly) using the UI following the steps I described above. Sounds like it was intentional on the console. 👍

PR #4857, opened 2024-10-23
Track consent revoked by and display it on the file card

dtgreiner:
## _Merging this PR_

## Description

Track consent revoked by and display the data on the file card the file has been revoked.

## Type of change
New feature


app/controllers/clients/files_controller.rb
```diff
@@ -249,6 +252,11 @@ def batch_params
       params.require(:batch_download).permit(:file_ids)
     end
 
+    def set_user
+      # binding.pry
```
eanders:
```suggestion
```

app/models/grda_warehouse/client_file.rb
```diff
@@ -233,15 +234,21 @@ class ClientFile < GrdaWarehouse::File
     ####################
     # Callbacks
     ####################
-    after_create_commit :notify_users, if: ->(m) { m.should_run_callbacks? }
-    before_save :adjust_consent_date, if: ->(m) { m.should_run_callbacks? }
-    after_save :note_changes_in_consent, if: ->(m) { m.should_run_callbacks? }
-    after_commit :set_client_consent, on: [:create, :update], if: ->(m) { m.should_run_callbacks? }
+    after_create_commit :notify_users, if: ->(m) { m.should_run_callbacks? } # rubocop:disable Style/SymbolProc
+    before_update :adjust_revoked_by, if: ->(m) { m.should_run_callbacks? && consent_revoked_at_changed? }
```
eanders: Is there some reason we can't just call this from the controller?  There are a bunch of callbacks, but the controller should know the user that made the change, so should be able to set `consent_revoked_by_user_id: current_user.id` when creating or updating the record.  (There is significant complexity here, so I may be missing something.)

While we're in here, it might be worth adding `has_paper_trail`.  I was going to suggest we pull the user out of the versions table, but it appears these aren't being logged.

eanders (changes requested):
Happy to discuss if you already looked at these ideas, but dropped a few questions inline.

app/models/grda_warehouse/client_file.rb
dtgreiner: It was mentioned in the ticket that this may be a good way to go. At the time, it felt like capturing it at a deeper level may good if we are working with files in multiple controllers, but I see what you are saying and it does feel less complex to just have the controller handle it. I'll adjust the PR.

dtgreiner:
@eanders I just pushed an update to pull the logic out of the callbacks and into the controllers. It should be good for another look whenever you have the chance. Thanks!

Edit - missed adding `paper_trail` here. Working on that now.

app/controllers/clients/files_controller.rb
```diff
@@ -107,7 +107,10 @@ def update
         attrs[:effective_date] = attrs[:consent_form_signed_on]
         attrs[:consent_form_confirmed] = true if GrdaWarehouse::Config.get(:auto_confirm_consent)
       end
-      @file.update(attrs)
+      @file.assign_attributes(attrs)
+      revoked_by_id = current_user.id if @file.consent_revoked_at.present?
+      @file.consent_revoked_by_user_id = revoked_by_id if @file.consent_revoked_at_changed?
```
eanders:
```suggestion
      if @file.consent_revoked_at_changed?
        @file.consent_revoked_by_user_id = if @file.consent_revoked_at.present?
          current_user.id
         else
           nil
         end
       end
```

eanders: The above can go in a method on `ClientFile` so you can call `@file.set_revocation_info(current_user)`

app/controllers/clients/files_controller.rb
```diff
@@ -230,6 +233,7 @@ def file_params
           :effective_date,
           :expiration_date,
           :consent_revoked_at,
+          :consent_revoked_by_id,
```
eanders: Don't think you need this since we aren't passing it as a parameter.

eanders (approved):
Thanks @dtgreiner, this is way cleaner.

app/models/grda_warehouse/file.rb
```diff
@@ -7,6 +7,7 @@
 module GrdaWarehouse
   class File < GrdaWarehouseBase
     acts_as_paranoid
+    has_paper_trail
```
gigxz, line 10: This caused a test regression on HMIS, because duplicate versions are being created for HMIS files. https://github.com/greenriver/hmis-warehouse/actions/runs/11525170831/job/32086897713

The problem is that `Hmis::File` [also calls](https://github.com/greenriver/hmis-warehouse/blob/6ec65c770a43d29291ecbed7918979fa27e5384a/drivers/hmis/app/models/hmis/file.rb#L10-L16) `has_paper_trail` (and attaches some metadata). 

Paper Trail has an unreleased (breaking) change that will actually cause this to fail https://github.com/paper-trail-gem/paper_trail/blob/master/CHANGELOG.md#breaking-changes / https://github.com/paper-trail-gem/paper_trail/pull/1479. They acknowledge that this can create duplicates. Glad we caught it with the test first!

I don't have much context on this PR, so not sure what the best course of action is. This class is pretty lightweight, so one approach is for Hmis::File to just not subclass it. I did that here https://github.com/greenriver/hmis-warehouse/pull/4872 and the tests pass, but it needed some weirdness to get `type` set.

eanders: Good catch @gigxz !  I believe our intent was actually to add `has_paper_trail` to ClientFile.  Not sure we need it in the base class.  I'll give that a try and see if tests still fail.

PR #4859, opened 2024-10-23
Support adding new service category

martha:
## _Merging this PR_

~Note, for now this targets [martha/6831-hud-svc-name](https://github.com/greenriver/hmis-warehouse/tree/martha/6831-hud-svc-name) but should retarget the release branch once that is merged.~

## Description

GH issue: https://github.com/open-path/Green-River/issues/5737

In this PR
- Add support in the Create and Update service type mutations to accept a category name as an alternative to category ID. If provided, the category name will be saved as a new category
- Remove unused mutations related to service categories
- Update the service type schema object to expose the whole category record, not just its name

## Type of change
New feature


drivers/hmis/app/graphql/mutations/create_service_category.rb
```diff
@@ -1,32 +0,0 @@
-#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
martha, line 1: Ok, it still takes me minutes not seconds to reason through backwards compatibility, but I think this deletion (and deletion of the other mutations for `delete_` and `rename_`) is fine, since it's part of the frontend schema but not used. So, this version of the backend should be compatible with both the old and new version of the frontend.

drivers/hmis/app/graphql/types/hmis_schema/service_type.rb
```diff
@@ -25,10 +25,12 @@ class HmisSchema::ServiceType < Types::BaseObject
     field :hud_record_type, HmisSchema::Enums::Hud::RecordType, null: true
     field :hud_type_provided, HmisSchema::Enums::ServiceTypeProvided, null: true
     field :category, String, null: false
+    field :category_record, HmisSchema::ServiceCategory, null: false
     field :form_definitions, [Forms::FormDefinition], null: false, description: 'Definitions that are specified for this service type'
 
     # object is a Hmis::Hud::CustomServiceType
 
+    # TODO(#5737) - remove this in favor of category_record, then rename category_record to category ?
```
martha: Let's discuss - this is to add support for matching the service category in the frontend to a pick list option which is keyed by category ID https://github.com/greenriver/hmis-frontend/pull/969/files#diff-817dbd1069fa5529fc67ea3079fdcb4a54eee9430fb15dc14fa9b279dc8d5b63R82.

gigxz: maybe it would be less disruptive to just add a field `field :category_id, ID, null: false`, instead of resolving a new object which duplicates the category name?

Another option would be to have the picklist just be names (not IDs), though it looks like the database doesn't validate the uniqueness of category names, so we would need the in order to confidently use names only.

drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb
```diff
@@ -8,14 +8,26 @@ module Types
...
     argument :supports_bulk_assignment, Boolean, required: false
 
     def to_params
-      result = to_h.except(:service_category_id)
-      result[:custom_service_category_id] = service_category_id
-      result
+      to_h.except(:service_category_id, :service_category_name)
+    end
+
+    def get_or_create_service_category(user_id, data_source_id)
```
gigxz: I would avoid using "create" in the method name here, because the method doesn't actually `create` (persist) the custom service category. it just builds it an returns it.  Maybe `find_or_initialize_service_category`, which mimics the language of the Rails built-in find_or_initialize_by.

drivers/hmis/app/graphql/types/hmis_schema/service_type_input.rb
```diff
@@ -8,14 +8,26 @@ module Types
...
+      # will be nil if neither is provided; that's fine for update (but not for create, which validates presence)
+      if service_category_id.present?
+        Hmis::Hud::CustomServiceCategory.find(service_category_id)
+      elsif service_category_name.present?
+        Hmis::Hud::CustomServiceCategory.new(
```
gigxz: To avoid creating duplicate service categories, lets use an existing service of that name if it exists. Something like:
```
Hmis::Hud::CustomServiceCategory.where(data_source_id: data_source_id, name: service_category_name).first_or_initialize(user_id: user_id)
```

drivers/hmis/app/graphql/types/hmis_schema/service_type.rb
martha: Yes, the first option is what I had implemented initially to minimize disruption. However, I thought that since this will need to be a 2-release feature for backwards compat anyway (search TODO(#5737) for other follow-up changes), I thought the disruption isn't so bad and it may be worth it to better match our usual pattern.

However, if you are also worried about another kind of disruption (likelihood of introducing mistakes/bugs?) I'm happy to update it to take that approach.

gigxz: I'm not worried about bugs as much as the effort involved to make these changes. If I understand correctly the change would need to span 3 releases:
* add `categoryRecord`, start fetching it from the frontend  (this release)
* change existing `category` type to be of type `HmisSchema::ServiceCategory` and update frontend to fetch _that_ instead; deprecate `categoryRecord` (next release)
* remove deprecated `categoryRecord` (next-next release)

Spanning 3 releases for a naming change just feels like a lot of effort relative to the benefit. If we remove the `"then rename category_record to category"` step, then its only these 2 releases, which seems OK since you already have other API deprecations to clean up. That said I have no preference for `categoryRecord { name, id }` vs  `category, categoryId`, so would be totally fine with leaving as-proposed. (With the suggestion that we don't re-name it again later).

drivers/hmis/app/graphql/mutations/create_service_category.rb
gigxz: That's right!

gigxz (approved):
lgtm! I like how easy it feels to add a new service type to a new category.

drivers/hmis/app/graphql/types/hmis_schema/service_type.rb
martha: Yes, good point that 3 releases seems a bit much. I'm proposing to rename it to `serviceCategory` and then keep that name in the long term, avoiding the 3rd release. I guess it's not perfect, but it does seem to match the pattern in a couple places elsewhere (like `projectType` for example seems similar-spirited to me?)

gigxz: that works, thanks!

PR #4860, opened 2024-10-23
improve auth error handling

ttoomey:
## _Merging this PR_

## Description
Discovered the WH `not_authorized!` method was not stopping control flow which could result in security issues. 

[related sentry issue](https://green-river.sentry.io/issues/5730469002/events/a1699e0b96294fa4948e210ace77ab50/)

* Changed implementation so it throws an error to stop flow
* Made HMIS auth error generic so it can be used on the WH too

example of the problem:
```ruby
def update
  not_authorized!
  # we should not get here
  # record.update(params)
```

## Type of change
Safety fix


eanders (approved):
@ttoomey this seems to work, but I've been unable to reproduce the original issue, were you?

ttoomey:
> this seems to work, but I've been unable to reproduce the original issue, were you?

I'm not sure that this change will fix the sentry issue; I flagged it since it seems like it could. The main issue I was addressing was that not_authorized! doesn't raise, which leads to issues like the above.

I think that raising is always what we want, I can't imagine a case where you'd want to continue after calling not_authorized.

If this doesn't fix the issue on the client controller, maybe we can try and move the 'before_actions' around so the auth check occurs first.

eanders:
Oh, I thought the goal with not authorized wasn't to raise, but to redirect to the user's root URL, which would then expose that they were trying to access an unauthorized path.

gigxz:
This looks good to me. We definitely _do_ want it to raise in the `hmis` and `hmis_external_apis` controllers, so this looks good to me. (Afaict it was already raising in those places)

ttoomey:
> Oh, I thought the goal with not authorized wasn't to raise, but to redirect to the user's root URL, which would then expose that they were trying to access an unauthorized path.

So you do still get that behavior with this change: not_authorized! raises a NotAuthorizedError which we then rescue explicitly in the controller. That rescue redirects to  the users root, so the user-facing behavior should be the same.

The important difference is that the flow is stopped. Continuing to run code after we know the action is not authorized could lead to security issues (or double-renders). For example this endpoint will redirect but also still destroy the user
```
def destroy
  not_authorized! unless user.is_admin?
  user.destroy
end 
```

Also, from a convention standpoint, the bang on the end of the method suggests to me that it should raise (much like active record save! or destroy!)

eanders:
Thanks for the explanation @ttoomey !

PR #4861, opened 2024-10-24
fix for sentry crash in hud apr

ttoomey:
## _Merging this PR_

## Description
[Should fix sentry issue](https://green-river.sentry.io/issues/5999913151/events/c642f101a49745cf830ad6a897d147d6/)

Somewhat untested. Needs second opinion. Could also try and track down occurrence of nulls in time_to_move_in

## Type of change
Bug fix

### DIFF
--- a/app/models/concerns/hud_reports/start_to_move_in_question.rb
+++ b/app/models/concerns/hud_reports/start_to_move_in_question.rb
@@ -36,7 +36,7 @@ module HudReports::StartToMoveInQuestion
             when :average
               value = 0
               scope = relevant_members.where(col_cond).where(a_t[:move_in_date].between(@report.start_date..@report.end_date))
-              stay_lengths = scope.pluck(a_t[:time_to_move_in])
+              stay_lengths = scope.pluck(a_t[:time_to_move_in]).compact
               value = (stay_lengths.sum(0.0) / stay_lengths.count).round if stay_lengths.any? # using round since this is an average number of days
               row.append_cell_value(value: value)
             else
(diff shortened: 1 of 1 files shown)


app/models/concerns/hud_reports/start_to_move_in_question.rb
```diff
@@ -36,7 +36,7 @@ def start_to_move_in_question(question:, members:, populations: sub_populations)
             when :average
               value = 0
               scope = relevant_members.where(col_cond).where(a_t[:move_in_date].between(@report.start_date..@report.end_date))
-              stay_lengths = scope.pluck(a_t[:time_to_move_in])
+              stay_lengths = scope.pluck(a_t[:time_to_move_in]).compact
```
eanders: I don't think we want to compact these since we count them to get averages on the next line.  Probably we need to coalesce with 0, which I thought was what `.sum(0.0)` was doing, guess not.  How about?
```suggestion
              stay_lengths = scope.pluck(a_t[:time_to_move_in]).map { |len| len.precence || 0 }
```

eanders: alternatively if we have ArelHelper already:
```suggestion
              stay_lengths = scope.pluck(cl(a_t[:time_to_move_in], 0))
```

ttoomey: Thanks! sounds like it's okay to treat nulls as 0 in this case

ttoomey:
```suggestion
              # treat null values as having 0 time in this case
              stay_lengths = scope.pluck(a_t[:time_to_move_in]).map { |len| len || 0 }
```

eanders (approved):
Looks good pending tests passing.  Thank you!

PR #4864, opened 2024-10-25
Patterns for cleanup of Client.rb

eanders:
## _Merging this PR_

## Description
* This attempts to define some patterns for compartmentalizing some of the code in Client.rb
* Defines some PORO to handle enrollment calculations and to support the enrollment view on the client dashboard
* Moves more CAS related code into the CAS concern

## Type of change
* Code clean-up


app/models/client_history/calculator.rb
```diff
@@ -0,0 +1,56 @@
...
+
+class ClientHistory::Calculator
+  attr_accessor :client
+  def initialize(client:)
+    self.client = client
+  end
+
+  # If we haven't been in a literally homeless project type (ES, SH, SO) in the last 30 days, this is a new episode
+  # You aren't currently housed in PH, and you've had at least a week of being housed in the last 90 days
+  def new_episode?(enrollments:, enrollment:)
```
eanders: Investigate if this can be combined with `enrollment.new_episode?`

eanders: Add a test for `new_episode?`

eanders:
@gigxz, @ttoomey and I spoke about this, but it would be great to get your feedback.  This is a first pass at removing some low hanging fruit from the client model with the intent of slowly working through the model to decrease the overall size and compartmentalize the logic.

app/models/client_history/calculator.rb
gigxz: `(enrollments:, enrollment:)` is hard to wrap my head around, maybe we could change at least the internal API to have more descriptive parameter names?

app/models/client_history/enrollment_view.rb
```diff
@@ -0,0 +1,67 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class ClientHistory::EnrollmentView
```
gigxz, line 7: Maybe a comment for what should go in this class? Eg is it for helpers specifically for the `clients/enrollment_table` view?

gigxz:
I think this looks great @eanders, I like that direction.

My only thoughts would be:
* Should we take this time to add comments to the PORO classes (/public methods on them) so we know what they are for? (I just have no context on these)
* Do we have enough test coverage?

app/models/client_history/calculator.rb
eanders: @gigxz how about using `residential_enrollments:, enrollment:`? (as it is now?)

eanders:
@gigxz turns out we do have test coverage, and I'd broken some with the last change.  I'm good merging when you are.

gigxz (approved):

PR #4867, opened 2024-10-25
Show data collection features when data is present

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6159

- Modify the Data Collection Feature property on Projects to return a feature if there is legacy data for it, even if no form instance exists.
- Add a similar Data Collection Feature to Enrollments.

Still to do
- [x] Test out Referrals case - has this borked any behavior?
- [x] Fix broken test
- [x] Add more spec tests

## Type of change
New feature

- [some, could add more] I have added spec tests (or not applicable)


drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -281,6 +281,65 @@ def self.generate_enrollment_id
...
+
+      # If the best instance doesn't match this enrollment, return, UNLESS there is existing data.
+      # (Example: data is collected for HoH. If HoH changes, you still see past data from the original HoH's enrollment)
+      next if !has_any_data && best_instance.present? && !best_instance.project_and_enrollment_match(project: project, enrollment: self)
+
+      OpenStruct.new(
+        role: role.to_s,
+        id: [id, role, best_instance&.id].join(':'), # Unique ID for Apollo caching
+        legacy: has_any_data && !best_instance,
+        data_collected_about: best_instance&.data_collected_about || 'ALL_CLIENTS', # Doesn't really matter for legacy
```
martha, line 333: Maybe it doesn't make sense to return this field at all for enrollment features?

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -281,6 +281,65 @@ def self.generate_enrollment_id
...
+
+      # Choose the "best" instance, i.e. the one that would actually be selected when recording new data.
+      # We need to do this so that we can accurately set "data collected about" based on the most applicable form.
+      best_instance = instance_scope.detect_best_instance_for_project(project: project)
+
+      has_any_data = case role
+      when :CURRENT_LIVING_SITUATION
+        current_living_situations.exists?
+      when :SERVICE
+        services.exists?
```
gigxz:
```suggestion
        custom_services.exists? || services.exists?
```

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -221,32 +222,42 @@ def to_pick_list_option
...
-        base_scope.inactive,
-      ].lazy.map { |scope| scope.order(updated_at: :desc).detect_best_instance_for_project(project: self) }.detect(&:present?)
-      next unless best_instance
+      best_instance = instance_scope.detect_best_instance_for_project(project: self)
+
+      has_any_data = case role
+      when :CURRENT_LIVING_SITUATION
+        current_living_situations.exists?
+      when :SERVICE
+        services.exists?
```
gigxz: include `custom_services`

drivers/hmis/spec/models/hmis/hud/project_spec.rb
```diff
@@ -82,11 +83,22 @@ def selected_instances
       expect(selected_instances.size).to eq(0)
     end
```
gigxz, line 85: This is pretty niche/intensive-to-QA behavior that involves both the frontend and backend, so I think it's a good candidate for a system test.

For example:
* for user who has access to edit [CLS], open the enrollment dashboard for an enrollment that has [CLS] but for which no [CLS form] is enabled in the project. (Or, enabled for different client type). User should be able to see and edit old [CLS], but not add a new one.
  * (and conversely if no CLS data exists, they should not see the tab)
  * include case where CLS is "migrated-in" (no FormProcessor) and case where CLS has a FormProcessor that references a definition
* for user who has access to edit [CLS], open the project dashboard for a project that has any [CLS] but for which no [CLS form] is enabled in the project. User should be able to see the [CLS] tab.
  * (and conversely if no CLS data exists, they should not see the tab)

martha: Whew, thank you for suggesting this! System test was a great idea, it turned up a couple of interesting bugs, some of which I'd like to reason through together - I'll add some more inline comments, and we can discuss tomorrow.

drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb
```diff
@@ -10,6 +10,7 @@ module Types
   class HmisSchema::CurrentLivingSituation < Types::BaseObject
     include Types::HmisSchema::HasHudMetadata
     include Types::HmisSchema::HasCustomDataElements
+    include Types::HmisSchema::HasFormDefinitionId
```
martha: This is one of the bugs turned up by the system test. We had added this functionality to support [case notes](https://github.com/greenriver/hmis-frontend/pull/951) a few weeks ago, but we didn't update it for any other data collection features.

I'd like to reason thru together: Where else do we need to make an update like this?
- Current Living Situation - updated here
- Service - ?
- CE Event - ?
- CE Assessment - ?
- Case note - was already updated
- Referral - client specific, no change needed?
- Referral request - client specific, no change needed?
- External form - no change needed, these already display using the definition from the submission

drivers/hmis/spec/system/hmis/data_collection_features_spec.rb
```diff
@@ -0,0 +1,196 @@
...
+        visit "/projects/#{p1.id}/overview"
+        expect(side_nav_elements).to include('Current Living Situations')
+
+        click_link 'Current Living Situations'
+        table_row = find('tbody').find_all('tr').sole.text
+        expect(table_row).to include(spouse.first_name)
+        expect(table_row).to include('Safe Haven')
+
+        click_link spouse.first_name
+        assert_current_path(/^\/client\/[^\/]+\/enrollments\/[^\/]+\/current-living-situations$/)
```
martha: This might be making a mountain of a molehill, but: I think 
`assert_current_path("/client/#{hoh.id}/enrollments/#{hoh_enrollment.id}/current-living-situations")` 
would be nicer, but the IDs don't match up. In reality the URL looks like 
- `/client/==M3ZRc2cxbVg5Z3J4dW9wejZ4NDByQT09/enrollments/==RzdzUG5mTm1STGM0UHcvb3IvODZwZz09/...`

instead of 
- `/client/200/enrollments/68686/...`

I've noticed this before in QA/prod envs too, and always figured it makes sense for "something something application security," but today I realized 
1) I don't really understand the mechanism by which it's making our application more secure? (we still do expose the true IDs directly in the client profile/enrollment overviews). So it must be for some other reason that I'll smack my forehead once I learn it!
2) Also I don't know where in the application we do this (and where we specify the behavior is turned off for local envs)
3) Pertinent to this line of code, is there any way to get from the project's object ID to the ID that will be used in the URL?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -193,6 +193,13 @@ def record_form_definition(role:, project_id: nil, id: nil)
         Hmis::Form::Definition.find_definition_for_role(role, project: project)
       end
 
+      # If the frontend is making this query, it's trying to display some data using a form definition,
+      # so we try to avoid returning nil.
+      # Even if we didn't find a good match (e.g., because the form is retired or instance is inactive),
+      # we return some default "best guess", enabling the application to display the data somehow instead of erroring.
+      # TODO(#6763) Use "system_managed" flag when it exists
+      record ||= Hmis::Form::Definition.where(role: role).first
```
martha: I think there's more finicky logic we could apply here. Like: Excluding draft forms, ranking any published form above any retired form (but still returning a retired form if it's the only one available), etc. I've avoided it so far since it seems like #6763 is coming down the pipes soon and should present a more comprehensive solution, do you think that's the right call?

Now that I'm thinking about it more, though, I wonder if that last bit about the retired forms _is_ important to bring in now, since `.first` will choose chronologically?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -235,7 +242,12 @@ def service_form_definition(service_type_id:, project_id:)
       service_type = Hmis::Hud::CustomServiceType.find_by(id: service_type_id)
       raise HmisErrors::ApiError, 'Service type not found' unless service_type.present?
 
-      Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+      definition = Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+
+      # Similar to record_form_definition above, we always want to return a definition if we possibly can
+      # TODO @MARTHA - this identifier could be different across environments, what is the best way to find the most basic service to return?
+      # service, hud-service-with-note, ...
+      definition || Hmis::Form::Definition.with_role(:SERVICE).where(identifier: 'service-date-only').first
```
martha: I'm sure `identifier` is not the right query here, since I noticed that across different environments (qa/stg/prd) the "basic" identifiers like "service-date-only" or "hud-service-with-note" don't consistently exist from env to env. Let's discuss what the best way is. Maybe this just becomes simple with #6763 and should be skipped until then.

drivers/hmis/spec/models/hmis/hud/project_spec.rb
```diff
@@ -128,13 +140,6 @@ def selected_instances
 
       expect(selected_instances).to contain_exactly(inst_for_project)
     end
-
-    it 'if all are inactive, includes the most specific inactive' do
```
martha, line 132: This test can be fully removed, it's no longer relevant, since if all instances are inactive and there's no legacy data, it should exclude from the data collection features list.

drivers/hmis/spec/models/hmis/hud/project_spec.rb
```diff
@@ -166,9 +171,9 @@ def selected_instance
     end
 
     it 'chooses instance specified by type (type > category)' do
+      expected = create(:hmis_form_instance, role: role, entity: nil, custom_service_type: cst)
```
martha: _This_ test on the other hand needs to be updated, it's currently failing with an existing, non-manifesting?, "bug".
- Check out `release-140` and run this test, it succeeds
- But if you swap lines 170 and 171 so that `expected` is not the last instance to be created, now it fails
- Why: Data collection features uses `Instance`'s [`detect_best_instance_for_project` method](https://github.com/greenriver/hmis-warehouse/blob/release-140/drivers/hmis/app/models/hmis/form/instance.rb) to choose the best instance. It is not applying any logic to prioritize type over category for service rules, as far as I can tell. So this [updated_at sort](https://github.com/greenriver/hmis-warehouse/blob/42ceab1b683f195a9ba6e69e6045720b4c64c853/drivers/hmis/app/models/hmis/hud/project.rb#L242) takes precedence, and since in the existing version of the test the `expected` is the last to be created, it has been falsely succeeding.
- Impact: I don't think this has any actual impact outside of the world of spec tests, because when you actually go to create a service you use the [`service_form_definition` query](https://github.com/greenriver/hmis-warehouse/blob/42ceab1b683f195a9ba6e69e6045720b4c64c853/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L238) which is correctly using `find_definition_for_service_type`.

Maybe the simple fix is to have the data collection features also use `find_definition_for_service_type` specifically for the SERVICE feature only?

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -221,32 +222,42 @@ def to_pick_list_option
...
+        current_living_situations.exists?
+      when :SERVICE
+        custom_services.exists? || services.exists?
+      when :CE_EVENT
+        false # Only resolved on enrollments. Would need to update this logic if we resolve CE events on projects
+      when :CE_ASSESSMENT
+        false # Only resolved on enrollments. Would need to update this logic if we resolve CE assessments on projects
+      when :CASE_NOTE
+        false # Only resolved on enrollments. Would need to update this logic if we resolve case notes on projects
+      when :REFERRAL # todo @Martha - test this out in frontend
```
martha: Testing how this impacts Referrals
- Appearance of "Referrals" in the project side nav is gated on [permission](https://github.com/greenriver/hmis-frontend/blob/64b8bafb3ef821d9518ef7a5b34cecc803b1b9d3/src/modules/projects/hooks/useProjectDashboardNavItems.tsx#L86-L95), not presence in data collection features list. We don't plan to change that behavior in this PR, per https://github.com/open-path/Green-River/issues/6159#issuecomment-2441679112
- `DataCollectionFeatureRole.Referral` and `.ReferralRequest` are not used, as far as I could tell in my code searches?
- I think I still don't really understand why they are considered data collection features at all, per my comment here https://github.com/greenriver/hmis-warehouse/pull/4867/files#diff-3b6ffa3383300231271de6203a0c29f562c97a5889cebdd37572baa5f7d3e76fR81-R82

Cases I've tested:
- Submitting a referral still works
- Accepting a referral still works
- Referral request - haven't fully tested, what is the best way to circumvent the external API?

Overall - I'd like a 2nd opinion, but I don't think this PR will impact anything in referrals land, since the referral data collection features are not used.

drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb
martha: This is basically testing behavior that hasn't been implemented yet and is not in scope https://github.com/open-path/Green-River/issues/6113

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Throw a `.published` scope here. All the system managed forms are published

drivers/hmis/spec/system/hmis/data_collection_features_spec.rb
martha: turn this off for testing for convenience

drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb
gigxz: Thanks for listing out, I will updated https://github.com/open-path/Green-River/issues/6113 which will capture this behavior change

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Use "service" identifier and add a todo for https://github.com/open-path/Green-River/issues/6763

drivers/hmis/spec/models/hmis/hud/project_spec.rb
martha: Remove irrelevant test. There's no such thing as a "best" instance for services in this context where we don't have a service type

drivers/hmis/app/models/hmis/hud/project.rb
martha: Discussed and decided: yes, it won't impact referrals, but keep consistency. No need to test referral request scenario.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: @gigxz using the system service form doesn't quite work nicely for custom services. I don't feel that confident about the workaround I've added here: https://github.com/greenriver/hmis-warehouse/pull/4867/files#diff-de4ce2db2cf589590502110af670f97dd57d9aae64e79a4615a3ece0644acbe3R16-R18. It would break if the user is editing a custom service and trying to change a filled-in value to nil.

martha:
System test failure step is due to https://github.com/open-path/Green-River/issues/6856#issuecomment-2451110332 👻

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -281,6 +281,68 @@ def self.generate_enrollment_id
     generate_uuid
   end
 
+  # Data Collection Features that are enabled for this enrollment, thru its project (e.g. Current Living Situation)
+  #
+  # Is it enabled?
+  #   If ANY data exists for it in this enrollment, even if no active instances exist, then yes.
+  #   (This supports both inactive instances and migrated-in data, as well as context changes such as change in HoH)
```
gigxz: I think the "inactive instances" part could be clarified. Maybe something like:


```
This allows users to see data in the following "legacy" scenarios:
(1) data was previously collected in HMIS, but the rule (Instance) has since been deactivated
(2) data was previously collected for this Enrollment, but is no longer collected because of some change in context (e.g. they are no longer HoH, or the project funder/attributes changed)
(3) data was migrated-in, but has no rule (Instance) to enable it
```

+ maybe some clarification that data in those categories is considered "legacy" b/c it's not collected going forward

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -281,6 +281,68 @@ def self.generate_enrollment_id
...
+      when :CE_EVENT
+        events.exists?
+      when :CE_ASSESSMENT
+        assessments.exists?
+      when :CASE_NOTE
+        custom_case_notes.exists?
+      when :REFERRAL
+        false # Referrals are resolved only on projects, not enrollments
+      when :REFERRAL_REQUEST
+        false # Referrals are resolved only on projects, not enrollments
```
gigxz: This is a little confusing to me. It seems like we DO resolve them on Enrollments if we find a `best_instance`, since they are included here.  If they _shouldn't_ be resolve on Enrollments (which I agree with), then we should exclude them from the list (`Hmis::Form::Definition::DATA_COLLECTION_FEATURE_ROLES.excluding(:REFERRAL, :REFERRAL_REQUEST, :EXTERNAL_FORM`).

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -281,6 +281,68 @@ def self.generate_enrollment_id
...
+        false # Not resolved on enrollments
+      else
+        raise "Unexpected data collection feature role: #{role}"
+      end
+
+      next unless best_instance || has_any_data
+
+      matches_enrollment = best_instance&.project_and_enrollment_match(project: project, enrollment: self)
+
+      # If an instance applies to this project but doesn't match this enrollment, return, UNLESS there is existing data.
```
gigxz: I think this will have incorrect behavior in the following scenario:
* Rule 1: "collect CURRENT_LIVING_SITUATION using form X for project 1 for HoH"
* Rule 2 "collect CURRENT_LIVING_SITUATION using form Y for ES projects for all clients"
* Enrollment is a non-HoH in Project 1, with is ES

Rule 1 is `best_instance` because it is more specific, but it ends up being thrown out here because the it doesn't match on DataCollectedAbout

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Yeah that's true. I think by "resolved" I meant to say "resolved by the frontend" but I should've said "*used* by the frontend" since it's resolved either way.

I'll make the change you're suggesting.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: https://github.com/greenriver/hmis-warehouse/blob/fce7cb27f276f99970c219e555b4567d16b58ca4/drivers/hmis/app/models/hmis/enrollment_assessment_eligibility_list.rb#L77-L87

You can use code similar to this to find the "best instance" while taking into account both the Project and the Enrollment. Find all the published definitions for this role, and rank them by their `project_and_enrollment_match` to find the one that's "best" for this specific Enrollment context. (For services, we only care if _any_ applicable form exists, because none are really "best" as discussed. For non-Service features, we only care which is "best" insofar as it determines what the "data_collected_about" value is.).

It seems like we may not be respecting the DataCollectedAbout value when choosing _which_  form to use for non-assessment data collection (record_form_definition query), which is a separate issue/bug. But that shouldn't stop us from implementing the right logic here.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Maybe- let's review how we're using this in the frontend now (I think, not at all) - and whether we want the frontend to use it in at ideal world.

martha: Sounds good. I checked the [frontend PR](https://github.com/greenriver/hmis-frontend/pull/972/files) to review every usage. (That PR replaces usages of `enrollment.project.dataCollectionFeatures` with the new `enrollment.dataCollectionFeatures` so I think it should be comprehensive)
- EnrollmentRoute access wrapper - `dataCollectedAbout` is not used
- EnrollmentCaseNotes, EnrollmentQuickActions, EnrollmentCeAssessmentsPage, EnrollmentCeEventsPage, EnrollmentCurrentLivingSituationsPage - all of these don't use `dataCollectedAbout`, they only check if a feature is `legacy`. (Maybe a new custom hook is in order?)
- EnrollmentDashboard - `dataCollectedAbout` is not used

Do we want the frontend to use it in an ideal world?
- Perhaps not, bc a data collection feature should only be returned on an enrollment at all if "data collected about" is true for this enrollment. At the frontend side, we don't need to ask "who should this data be collected about in general?" when we already have the answer to "is this data collected about this individual?"
- Otoh, would we ever want to add some instructional text that helps users understand the customizability, like understanding why a tab/form appears when it does? Something like "Collect case notes here for user XX because this project is type YY and the user is the HoH." I have a hard time seeing us adding this because I think the target audience of this instructional text probably wouldn't be the same person who has permission to configure data collection?
- Is there something else I'm missing around complex situations where multiple instances apply to an enrollment? My instinct is still saying that all that complexity should live on the backend only.

martha: But one nice thing about leaving it as-is is that we can use the same type `DataCollectionFeatureFieldsFragment` for both project and enrollment

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Thanks for pointing out the bug!

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Hm, by the same logic, should we exclude `CE_EVENT`, `CE_ASSESSMENT` and `CASE_NOTE` from being resolved on Project? https://github.com/greenriver/hmis-warehouse/blob/martha/6159-data-collection/drivers/hmis/app/models/hmis/hud/project.rb#L242-L247 (and maybe referral and referral request as well, though we already talked about these and decided to include them?)

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -227,4 +227,26 @@
...
+      # Rule 2 is less specific for the project, but applies to the non-HOH enrollment in question
+      let!(:rule2) { create :hmis_form_instance, entity: nil, project_type: p1.project_type, role: 'CURRENT_LIVING_SITUATION', definition_identifier: 'custom_cls' }
+
+      let!(:hoh) { create :hmis_hud_client, first_name: 'Annie', data_source: ds1 }
+      let!(:spouse) { create :hmis_hud_client, first_name: 'Jessie', data_source: ds1 }
+      let!(:hoh_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: hoh, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 1 }
+      let!(:spouse_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: spouse, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 3 }
+
+      it 'should still return this feature for the enrollment' do
+        expect(spouse_enrollment.data_collection_features.pluck(:role)).to include('CURRENT_LIVING_SITUATION')
```
gigxz: We can get more specific here, to ensure the correct rule was selected:
```
expect(spouse_enrollment.data_collection_features.first).to have_attributes('instance' => rule2, 'data_collected_about' => 'ALL_CLIENTS')

expect(hoh_enrollment.data_collection_features.first).to have_attributes('instance' => rule1, 'data_collected_about' => 'HOH')

```
(maybe as 2 different test cases)

drivers/hmis/spec/system/hmis/data_collection_features_spec.rb
```diff
@@ -0,0 +1,197 @@
...
+
+  before(:each) do
+    sign_in(hmis_user)
+  end
+
+  def side_nav_elements
+    find_all('a[id^="side-nav-"]').map(&:text)
+  end
+
+  context 'when no CLS is enabled in the project' do
```
gigxz, line 26: These tests look great btw, thanks!

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -227,4 +227,26 @@
...
+      # Rule 1 is more specific for this project, but doesn't apply to the specific enrollment in question
+      let!(:rule1) { create :hmis_form_instance, entity: p1, data_collected_about: 'HOH', role: 'CURRENT_LIVING_SITUATION', definition_identifier: 'custom_cls' }
+      # Rule 2 is less specific for the project, but applies to the non-HOH enrollment in question
+      let!(:rule2) { create :hmis_form_instance, entity: nil, project_type: p1.project_type, role: 'CURRENT_LIVING_SITUATION', definition_identifier: 'custom_cls' }
+
+      let!(:hoh) { create :hmis_hud_client, first_name: 'Annie', data_source: ds1 }
+      let!(:spouse) { create :hmis_hud_client, first_name: 'Jessie', data_source: ds1 }
+      let!(:hoh_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: hoh, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 1 }
+      let!(:spouse_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, client: spouse, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 3 }
```
gigxz: Just so you know, you don't _need_ to create the clients explicitly. If there aren't specific attributes that you need on the client, you can allow the enrollment factory to create them:
```suggestion
      let!(:hoh_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 1 }
      let!(:spouse_enrollment) { create :hmis_hud_enrollment, data_source: ds1, project: p1, entry_date: 1.month.ago, household_id: 'household1', relationship_to_hoh: 3 }
```

Not a necessary change here, just pointing it out. In this case it is probably convenient that each client is assigned a distinct `first_name` so we can easily tell them apart when debugging.

gigxz: Oops - I thought I was adding that comment to the system test (hence the debugging comment). But same goes here.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: DataCollectionFeatures.dataCollectedAbout used here:
https://github.com/greenriver/hmis-frontend/blob/ccf0f5de4a293ca3f301eaba78290e981a8fd92e/src/modules/enrollment/components/pages/EnrollmentDashboard.tsx#L46-L60
(which calls [this util fn](https://github.com/greenriver/hmis-frontend/blob/ccf0f5de4a293ca3f301eaba78290e981a8fd92e/src/modules/hmis/hmisUtil.ts#L567-L588) to evaluate DataCollectedAbout against the Enrollment).

So based on that, my understanding is that DataCollectedAbout from the project feature object IS currently taken into account when determining whether to show a given `navItem`. (That happens [here](https://github.com/greenriver/hmis-frontend/blob/ccf0f5de4a293ca3f301eaba78290e981a8fd92e/src/modules/enrollment/hooks/useEnrollmentDashboardNavItems.tsx#L44-L46)). In other words, if the `project.dataCollectionFeatures` includes `CURRENT_LIVING_SITUATION` with dca `HOH`, then the CLS tab won't show up on a non-HoH Enrollment dashboard. So it's working as expected (the old behavior, I mean).

I agree that the _new_ approach is to evaluate the "data collected about" logic on the backend, since we are now resolving data collection features on the Enrollment itself. As a result, we choose to just exclude the feature if it's not enabled for this context. So we should be able to pull out most of that logic from the frontend.

I agree we should still resolve `data_collected_about` on the Enrollment feature, for future explanatory text and to use the same type for both.

gigxz: > by the same logic, should we exclude CE_EVENT, CE_ASSESSMENT and CASE_NOTE from being resolved on Project

Ah yeah I see your point. But I have no problem keeping them. It's accurate information, and we may want to show it in the future (for example in a project "configuration summary" that explains which data elements are collected https://github.com/open-path/Green-River/issues/5999 ... though that may need to take a different shape) or we may want a Project Case Notes page later, etc.

martha: Ok, yes I agree with your understanding of both the old behavior and the new approach. I'm good with the decisions to keep resolving `data_collected_about` here and keep resolving all features on the Project. 👍

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
martha: Cool, yeah I think here in the unit test it makes sense to do this. I'll keep the system test as is for convenience in debugging.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -235,7 +241,12 @@ def service_form_definition(service_type_id:, project_id:)
       service_type = Hmis::Hud::CustomServiceType.find_by(id: service_type_id)
       raise HmisErrors::ApiError, 'Service type not found' unless service_type.present?
 
-      Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+      definition = Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+
+      # Similar to record_form_definition above, we always want to return a definition if we possibly can, so use the
+      # HUD default form. For non-HUD services, the only field that would show up is Date Provided.
+      # TODO(#6763) Use "system_managed" flag when it exists
+      definition || Hmis::Form::Definition.with_role(:SERVICE).where(identifier: 'service').first
```
gigxz: I have thought this through a little more @martha and maybe come to a different conclusion. Given that we don't have a "default" form that works for all Services, I think we should just revert this to the old behavior where it returns empty (and subsequently errors) if no form is found for a CustomService type. 

```suggestion
      # Similar to record_form_definition above, we always want to return a definition if we possibly can, so use the
      # default HUD Service form for HUD service types. For Custom service types, return empty because we can't determine which form to use.
      # TODO(#6763) Use "system_managed" flag when it exists
      definition || Hmis::Form::Definition.with_role(:SERVICE).where(identifier: 'service').first if service_type.hud_service?
```



[This is the frontend code](https://github.com/greenriver/hmis-frontend/blob/ccf0f5de4a293ca3f301eaba78290e981a8fd92e/src/modules/services/hooks/useServiceDialog.tsx#L114-L118) that hits this query. There are 2 cases:
1) User is creating a new service, and they selected a type in the dropdown (`AVAILABLE_SERVICE_TYPES`). In this case we expect the form to exist, and its OK to error if it doesn't.
2) User is editing an existing Custom Service. It should have a form processor definition_id, because CustomServices are generated in HMIS, or they are migrated-in during which we'd add a FormProcessor that points to the correct form.
3) User is editing an existing HUD Service. Default to the HUD form as suggested.


(1) should be functional already, but I added https://github.com/open-path/Green-River/issues/6939 as an idea to simplify it.
(2) won't really be real until https://github.com/open-path/Green-River/issues/6113 is implemented, but I think it's OK to retain current behavior in the mean time (where it expects the service to still be enabled).
(3) is achieved by the diff

What do you think?

drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
```diff
@@ -13,6 +13,10 @@ def process(field, value)
       # service is a Hmis::Hud::Service or Hmis::Hud::CustomService
       service = @processor.service_factory
 
+      # Get around the fact that we use the system service form for editing legacy custom services, which don't accept
+      # fields like other_type_provided and moving_on_other_type
+      return if service.instance_of?(Hmis::Hud::CustomService) && attribute_value.nil?
```
gigxz: Suggest removing per my other comment. This feels like it could swallow bugs

gigxz (commented):
Looks great! One remaining comment on service otherwise lgtm

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: https://github.com/greenriver/hmis-warehouse/pull/4867#discussion_r1829480896

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: I tested this out locally with a disabled service. It doesn't throw an error in the frontend, just hides the form like this:
Screenshot 2024-11-05 at 12 26 24 PM
(Normally a form would go here where you can update the date/values.)

I think it's a good solution, thanks Gig!

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -235,7 +240,11 @@ def service_form_definition(service_type_id:, project_id:)
       service_type = Hmis::Hud::CustomServiceType.find_by(id: service_type_id)
       raise HmisErrors::ApiError, 'Service type not found' unless service_type.present?
 
-      Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+      definition = Hmis::Form::Definition.find_definition_for_service_type(service_type, project: project)
+
+      # Similar to record_form_definition above, we always want to return a definition if we possibly can, so use the
+      # default HUD Service form for HUD service types. For Custom service types, return empty because we can't determine which form to use.
+      definition || (service_type.hud_service? ? Hmis::Form::Definition.with_role(:SERVICE).where(managed_in_version_control: true).first : nil)
```
gigxz: I think the `||` with a ternary is a but more complicated than it needs to be. What about this (untested)?
```suggestion
      definition ||= Hmis::Form::Definition.with_role(:SERVICE).where(managed_in_version_control: true).first if service_type.hud_service?
      definition
```

gigxz (approved):
Phew, nice job working through all that. I think this looks good and is ready for QA.  🙌 

One minor code complexity comment below >

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Yup this seems to work well with local testing, but I added a new spec too so I'll wait for the tests to pass before merging. Thanks for the suggestion!

martha:
I think the latest test failure is unrelated to this PR:

<details>

```
Failures:

  1) Curl CA Certificate Works does throw an error when curling over https
     Failure/Error:
       expect do
         Curl::Easy.perform('https://services.etosoftware.com/') do |http|
           # Without defining a CA Cert we throw an error
           # http.cacert = '/etc/ssl/certs/ca-certificates.crt'
           http.verbose = true
         end
       end.to raise_error(Curl::Err::SSLCaertBadFile)

       expected Curl::Err::SSLCaertBadFile but nothing was raised
     # ./spec/models/curl_spec.rb:17:in `block (4 levels) in <main>'
     # /usr/local/bundle/gems/vcr-6.2.0/lib/vcr.rb:274:in `turned_off'
     # ./spec/models/curl_spec.rb:16:in `block (3 levels) in <main>'
     # /usr/local/bundle/gems/webmock-3.23.0/lib/webmock/rspec.rb:39:in `block (2 levels) in <main>'

Finished in 21 minutes 40 seconds (files took 16.03 seconds to load)
2673 examples, 1 failure, 10 pending

Failed examples:

rspec ./spec/models/curl_spec.rb:15 # Curl CA Certificate Works does throw an error when curling over https
```
</details>

PR #4869, opened 2024-10-28
Add `managed_in_version_control` flag to forms, and support duplicating them

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/6763
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/973

The background here is that currently, the form builder lets you make new draft of system forms (say, the default Client form), and that causes problems when we try to deploy because the seed script doesn't expect it. The approach we went with was to disallow editing such forms in the config tool. (For now)

* Add `managed_in_version_control` flag to forms that are updated on deploy from the JsonForms seed script. This flag tells us that the form should NOT be edited in the Form Builder. (Very open to naming suggestions!! 👋  I felt that "system" was too generic and sometimes used to refer to forms that are required) 
* Add a mutation `CreateDuplicateFormDefinition` that allows admins to duplicate existing forms. This provides a path forward for **super admins** to create duplicates of forms that are "managed in version control" -- particularly helpful for QA.

## Type of change
New feature / bug fix

## Testing Instructions
* Run `rails driver:hmis:seed_definitions` to set the flag (tested with different clients)
* Pull down frontend branch https://github.com/greenriver/hmis-frontend/pull/973 with this, you should see:
  * system-managed forms are flagged and sorted to the end
  * when you open a system-managed form, there is an alert saying contact support to edit, and there is a Duplicate Form button
  * the Duplicate button works and creates a new form
  * if there were custom fields in the base form, the new form collects onto a new cded (not the same one) after publish. I tested this by temporarily turning on the "duplicate" button for a non-system form that I knew had CDEs on it.


drivers/hmis/app/graphql/mutations/create_duplicate_form_definition.rb
```diff
@@ -0,0 +1,67 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Mutations
+  class CreateDuplicateFormDefinition < CleanBaseMutation
```
martha, line 8: I think just "DuplicateFormDefinition" would be fine and more succinct, but don't have a strong feeling either way

drivers/hmis/app/graphql/mutations/create_duplicate_form_definition.rb
```diff
@@ -0,0 +1,67 @@
...
+        where(identifier: identifier).
+        published_or_retired.
+        order(version: :desc).first
+
+      raise 'not found' unless definition_to_duplicate
+
+      access_denied! unless current_user.can_manage_forms_for_role?(definition_to_duplicate.role)
+
+      # Drop all custom_field_key mappings on definition structure
+      cleaned_definition_json = remove_custom_field_mappings(definition_to_duplicate)
```
martha: This is actually in-place-modifying the `definition_to_duplicate.definition` attribute, and then it just doesn't save changes on `definition_to_duplicate` to the db, right?

Maybe code-wise it would be clearer/safer to clone this definition first before `walk_definition_nodes`? Maybe this takes up more space in memory though?

martha (approved):
Cool thank you! 👏

martha:
> Run rails driver:hmis:seed_definitions to set the flag

Aside- this step fails with `ActiveRecord::RecordInvalid: Validation failed: Identifier has already been taken` if there's an existing draft on a system form. There shouldn't be in client envs, but I had quite a lot of those in my local db, which I just cleared out manually.

I had so many because I've been in the habit of creating drafts whenever I want to examine the json of a form, whether or not I need to actually make changes to it. Maybe it could be worthwhile for developer experience to add a super-admin button for viewing the json of a published form? If you agree that it would be useful, I'd be happy to make that change.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -412,6 +414,10 @@ def valid_status_for_submit?
     published? || retired?
   end
 
+  def managed_in_version_control?
+    managed_in_version_control
+  end
```
ttoomey: surprised this is necessary, active record should give you this method automatically

ttoomey (approved):
Looks great, thanks!

gigxz:
>  Maybe it could be worthwhile for developer experience to add a super-admin button for viewing the json of a published form? If you agree that it would be useful, I'd be happy to make that change.

Yes I've been wanting that too, that would be great

drivers/hmis/app/graphql/mutations/create_duplicate_form_definition.rb
gigxz: Yeah fair point, I'll throw in a `dup`!

drivers/hmis/app/graphql/mutations/create_duplicate_form_definition.rb
gigxz: thanks- agree but keeping as is

drivers/hmis/app/models/hmis/form/definition.rb
gigxz: 🤦  that's right, removed

gigxz:
@martha sorry for the delay on getting this one in. It's retargeted to 141 and ready to merge, I'll just wait for tests to pass again.

PR #4871, opened 2024-10-28
Remove can_manage_inventory permission

martha:
## _Merging this PR_

## Description

GH issue: https://github.com/open-path/Green-River/issues/6605

This follows up on the changes to unit management permissions deployed in release-138. It fully removes the `can_manage_inventory` permission/column.

## Type of change
Code clean-up

### DIFF
--- a/db/migrate/20241016202657_add_can_view_units_permission.rb
+++ b/db/migrate/20241016202657_add_can_view_units_permission.rb
@@ -2,6 +2,7 @@ class AddCanViewUnitsPermission < ActiveRecord::Migration[7.0]
   def up
     ::Hmis::Role.ensure_permissions_exist
     ::Hmis::Role.reset_column_information
+    ::Hmis::Role.where(can_manage_inventory: true).update_all(can_view_units: true, can_manage_units: true) if column_exists?(:hmis_roles, :can_manage_inventory)
   end
 
   def down
--- a/db/migrate/20241016202657_add_can_view_units_permission.rb
+++ b/db/migrate/20241016202657_add_can_view_units_permission.rb
@@ -2,7 +2,6 @@ class AddCanViewUnitsPermission < ActiveRecord::Migration[7.0]
   def up
     ::Hmis::Role.ensure_permissions_exist
     ::Hmis::Role.reset_column_information
-    ::Hmis::Role.where(can_manage_inventory: true).update_all(can_view_units: true, can_manage_units: true)
   end
 
   def down
--- a/db/migrate/20241028193644_remove_can_manage_inventory_permission.rb
+++ b/db/migrate/20241028193644_remove_can_manage_inventory_permission.rb
@@ -1,0 +1,13 @@
+class RemoveCanManageInventoryPermission < ActiveRecord::Migration[7.0]
+  def up
+    safety_assured do
+      remove_column :hmis_roles, :can_manage_inventory
+    end
+  end
+
+  def down
+    # This won't really restore the column if it hasn't been added back in the rails code.
+    ::Hmis::Role.ensure_permissions_exist
+    ::Hmis::Role.reset_column_information
+  end
+end
--- a/drivers/hmis/app/models/hmis/role.rb
+++ b/drivers/hmis/app/models/hmis/role.rb
@@ -155,13 +155,6 @@ class Hmis::Role < ::ApplicationRecord
         category: 'Project Access',
         sub_category: 'Management',
       },
-      can_manage_inventory: { # TODO: deprecated, remove in next release in favor of can_manage_units below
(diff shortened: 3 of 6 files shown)


db/migrate/20241016202657_add_can_view_units_permission.rb
```diff
@@ -2,7 +2,6 @@ class AddCanViewUnitsPermission < ActiveRecord::Migration[7.0]
   def up
     ::Hmis::Role.ensure_permissions_exist
     ::Hmis::Role.reset_column_information
-    ::Hmis::Role.where(can_manage_inventory: true).update_all(can_view_units: true, can_manage_units: true)
```
martha, line 5: I'm suggesting to remove this since we're removing the column `can_manage_inventory`. I remember problems in the past with data in migrations, and the controversy over whether they should be used for schema updates only even though it's sometimes convenient to also use them for data changes, so it just seems safer to me to remove it.

db/migrate/20241028193644_remove_can_manage_inventory_permission.rb
```diff
@@ -0,0 +1,13 @@
+class RemoveCanManageInventoryPermission < ActiveRecord::Migration[7.0]
+  def up
+    safety_assured do
```
martha, line 3: I think that we don't have to worry about this warning from StrongMigrations since we weren't reading this attribute anywhere in the previous version of the code?

```
Caused by:
StrongMigrations::UnsafeMigration:
=== Dangerous operation detected #strong_migrations ===

Active Record caches attributes, which causes problems
when removing columns. Be sure to ignore the column:

class HmisRole < ApplicationRecord
  self.ignored_columns += ["can_manage_inventory"]
end

Deploy the code, then wrap this step in a safety_assured { ... } block.

class RemoveCanManageInventoryPermission < ActiveRecord::Migration[7.0]
  def change
    safety_assured { remove_column :hmis_roles, :can_manage_inventory }
  end
end
```

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -155,13 +155,6 @@ def self.permissions_with_descriptions
         category: 'Project Access',
         sub_category: 'Management',
       },
-      can_manage_inventory: { # TODO: deprecated, remove in next release in favor of can_manage_units below
```
martha, line 158: After this the 2 remaining occurrences of `can_manage_inventory` in the code are in `db/structures/primary.sql` and `drivers/datalab_testkit/spec/fixpoints/warehouse_app_test.sql`. These both look like generated files - is there a code gen step that I should run to get rid of it here, or is it fine to leave?

gigxz: Huh I've never noticed those files. It looks like maybe they get updated when we collapse migrations, which we do like every 6 months (latest update was https://github.com/greenriver/hmis-warehouse/pull/4583). So I wouldn't worry about that.

db/migrate/20241016202657_add_can_view_units_permission.rb
gigxz: That seems fine as long as we are sure that the migration has already been applied everywhere. In this case, it hasn't -- the migration went into release-139 https://github.com/greenriver/hmis-warehouse/pull/4836/files
So we will want to target this PR against 140! And of course be sure that release-139 is deployed independently, before we deploy 140.

db/migrate/20241028193644_remove_can_manage_inventory_permission.rb
gigxz: Yeah that seems fine to me

db/migrate/20241016202657_add_can_view_units_permission.rb
eanders: Alternatively, you can ask if the column exists and only run the migration in that case, that should be safer since we may deploy to different places at different times.

```
::Hmis::Role.where(can_manage_inventory: true).update_all(can_view_units: true, can_manage_units: true) if column_exists?(:hmis_roles, :can_manage_inventory)
```

drivers/hmis/app/models/hmis/role.rb
eanders: the files in `db/structures/*.sql` are, as @gigxz noted, the files we generate as part of the migration collapses.  They shouldn't be a problem, but new installations will use them to bootstrap, so running those and then the subsequent migrations needs to work.  In this case, the column will probably be added and then removed.

gigxz (approved):
looks good. by the way the test failure will be fixed with https://github.com/greenriver/hmis-warehouse/pull/4874

db/migrate/20241016202657_add_can_view_units_permission.rb
martha: Thanks, good idea!

eanders (approved):

PR #4873, opened 2024-10-29
6718-remove-hmis-records

ttoomey:
## _Merging this PR_

## Description
[GH issue](https://github.com/open-path/Green-River/issues/6718)

* Delete soft-deleted hmis client and related records
* Ignore project records
* Delete related records that would crash bulk deletes due to fk integrity

## Type of change
Feature / housekeeping


app/jobs/purge_soft_deleted_records_job.rb
```diff
@@ -0,0 +1,112 @@
...
+            check_max_deleted(dependent_scope.size)
+            dependent_scope.delete_all unless @dry_run
+          end
+        end
+
+        check_max_deleted(batch.size)
+        batch.delete_all unless @dry_run
+      end
+    end
+  end
```
eanders: I'm probably missing something here, but couldn't we skip the batching and just do something like the following?
```
num_deleted = scope.limit(@max_deleted).delete_all
check_max_deleted(num_deleted)
```

As long as you loop over the dependent items first/client last, the client records should be safe to delete once everything else is gone.

eanders (approved):
@ttoomey thanks for putting this together.  I think it's fine as is, but may be doing more work than necessary.

app/jobs/purge_soft_deleted_records_job.rb
ttoomey: My thought was that batches result in smaller individual transactions that should be less load on the db (smaller memory peak usage but a longer run-time).  It also makes the dependent joins a little easier. I could be convinced otherwise though

app/jobs/purge_soft_deleted_records_job.rb
```diff
@@ -0,0 +1,112 @@
+###
+# Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# hard-delete HMIS old soft-deleted records
```
gigxz: I don't think "HMIS" is a clear descriptor here, since this isn't focused on "HMIS" as in the data source / app, nor is it limited to only HUD-HMIS-spec records (since we include custom tables).

Maybe we could just say it cleans up "source data". In addition to cleaning up "destination Clients"? Thats seems more in line with the language we use elsewhere, but I could be wrong.

gigxz (approved):
looks good

app/jobs/purge_soft_deleted_records_job.rb
ttoomey: good point, thanks. That comment is stale. I've improved the documentation of the class and params

app/jobs/purge_soft_deleted_records_job.rb
ttoomey: I kicked up the batch size to 5k, we can see how it runs on QA

PR #4877, opened 2024-10-29
Show occurrence point forms with legacy data

martha:
## _Merging this PR_
- ~I targeted this against [martha/6159-data-collection](https://github.com/greenriver/hmis-warehouse/tree/martha/6159-data-collection) (the other PR relevant to this ticket), but it should target the release branch once that is merged~

## Description
GH issue: https://github.com/open-path/Green-River/issues/6972

- Move Occurrence Points applicable to an enrollment, which were previously resolved on Projects, to be resolved on Enrollments.
- Add logic to the occurrence point resolver so that it returns if there is any legacy data present, even if there is no matching form instance.
  - It only searches for legacy data if there is no match
  - It works for both record data and CDEs (where the owner is an Enrollment)
- Add spec test

## Type of change

  - [ ] ensured this does not introduce N+1s - if I understand correctly, this does introduce n+1, but I'm not sure it's worth the time to avoid since it isn't used in batch right now.
  - [x] ensured permissions and visibility checks are performed in the right places - yes, based on the way Projects are currently configured it doesn't seem like a permission check is needed on occurrence point forms since they are config not pii


drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -340,6 +340,49 @@ def data_collection_features
...
+          custom_field_key = item.mapping&.custom_field_key
+
+          next unless record_type == 'ENROLLMENT' || custom_field_key
+
+          if record_type && field_name
+            next unless respond_to?(field_name)
+
+            has_any_data = send(field_name).present?
+          elsif custom_field_key
+            # For simplicity, for now, just look for CDEDs where the owner is an Enrollment or a Client
```
martha: Add comment to make more clear - this has only been tested to work in specific cases. Drop the client thing for now

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -8553,7 +8558,7 @@ type Project {
   """
   Forms for individual data elements that are collected at occurrence for this Project (e.g. Move-In Date)
   """
-  occurrencePointForms: [OccurrencePointForm!]!
+  occurrencePointForms: [OccurrencePointForm!]! @deprecated(reason: "Occurrence points should now be resolved on Enrollments")
```
martha: Probably don't actually deprecate this

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -343,6 +343,42 @@ def data_collection_features
     end.compact
   end
 
+  def occurrence_point_forms
+    project.occurrence_point_form_instances.map do |instance|
+      matches = instance.project_and_enrollment_match(project: project, enrollment: self).present?
+      has_any_data = false
+
+      unless matches # Only do this computation if the instance doesn't match this enrollment
```
gigxz: To reduce indentation, we could bump up the `next`:

```ruby
project.occurrence_point_form_instances.map do |instance|
  matches = instance.project_and_enrollment_match(project: project, enrollment: self).present?

  # This instance is enabled for this project _and_ matches this Enrollment (for "Data Collected About"), so we know it should be shown for this Enrollment. Proceed
  next instance if matches

  # This instance is enabled for this project BUT doesn't match this Enrollment. Determine if it has any data, in which case we will show it anyway.
  ... has_any_data computation logic

```

I think this logic is a little different from what we discussed, though. My understanding of where we landed was that the "occurrence point form" should _always_ be shown if it has any data (specifically for HUD elements -- MID/DOE/PATH status), even if that form is not enabled in the given project.

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -227,4 +227,122 @@
...
+
+  describe 'occurrence point form instances' do
+    let(:role) { :OCCURRENCE_POINT }
+    let!(:project) { create(:hmis_hud_project, data_source: ds1) }
+    let!(:hoh_enrollment) { create(:hmis_hud_enrollment, project: project, data_source: ds1, household_id: 'household1', relationship_to_hoh: 1) }
+    let!(:spouse_enrollment) { create(:hmis_hud_enrollment, project: project, data_source: ds1, household_id: 'household1', relationship_to_hoh: 3) }
+
+    context 'no relevant instance exists' do
+      it 'does not return the form when no instance exists' do
+        expect(hoh_enrollment.occurrence_point_forms).to be_empty
```
gigxz: In this test, is there any OCCURRENCE_POINT form definition in the database? I'm not totally understanding asserting about "the form" when there is no form (as far as I can see in the fixtures)

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -227,4 +227,122 @@
...
+                'record_type': 'ENROLLMENT',
+              },
+              'repeats': false,
+              'component': 'ADDRESS',
+            },
+          ],
+        }
+      end
+      let!(:definition) { create :hmis_form_definition, role: role, definition: definition_json }
+      let!(:instance) { create(:hmis_form_instance, role: role, entity: project, active: true, definition: definition) }
```
gigxz, line 337: add test case for what happens for an enrollment in a different project that has a MoveInDate

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -227,4 +227,122 @@
...
+              :hmis_hud_enrollment,
+              project: project,
+              data_source: ds1,
+              household_id: 'household1',
+              relationship_to_hoh: 3,
+              move_in_date: 3.weeks.ago,
+            )
+          end
+
+          let!(:address) { create(:hmis_move_in_address, data_source: ds1, enrollment: spouse_enrollment) }
```
gigxz: Is it necessary to have the address? It should work if just the move_in_date is present

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Thanks, I like this suggestion. I didn't know about the `next instance if matches` syntax, that's pretty 👍 

> should always be shown if it has any data, even if that form is not enabled in the given project.

That's my understanding too. Is this not accomplishing that?
- If the instance matches, show the form
- otherwise, if it has any data (regardless of instance match), show the form
- otherwise don't show

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
martha: Ok, I moved the fixtures creating the definition up to the top level. Thanks!

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: > That's my understanding too. Is this not accomplishing that?

I don't think so. `project.occurrence_point_form_instances` only returns instances that are "enabled" for this project (it uses detect_best_instance_for_project internally). So say, for example, an Enrollment has a Move-in Date value on it but the Move-in Date form is not enabled for that project type. Move-in date would (incorrectly) not be shown on that enrollment, because we only consider Occurrence Point forms that are **enabled** for the project. Does that track?

martha: Ah I see, yep that tracks. Thanks. I'll take another look at it today and create a test where there is legacy data in a non-enabled form.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,42 @@ def data_collection_features
     end.compact
   end
 
+  def occurrence_point_forms
```
gigxz, line 339: I think we will have duplication issue with this. Say for example there is a Form Definition for Move-in Date that has 4 rules (one enabling it for PH, one for RRH, etc.). We're looking at an Enrollment that has a `MoveInDate` _value_ but is not in a project where the Move-in Date form is _enabled_. Since we're iterating through all `instances` below, I think it will result in the Move In Date Form Definition being resolved 4 times on the Enrollment. 
Is that true?

Maybe we should tackle this more similarly to data_collection_features logic, aka check each latest-version of OCCURRENCE POINT definitions (by identifier) to see (1) is it enabled for this context, and if not (2) is there legacy data for it. Sorry not a fully formed suggestion, but lets make sure we are avoiding returning duplicates of the same form definition identifer

martha: Thanks @gigxz good point. I tinkered with it and came up with something that I think is along the lines of what you had in mind. I added a spec test to illustrate the scenario you're describing.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,57 @@ def data_collection_features
...
+    # All instances for Occurrence Point forms, including inactive/retired, so we can check for legacy data.
+    # Draft forms should never have legacy data, but let's include them just in case.
```
martha: Let's discuss this. Let me know if you think we should explicitly exclude draft-only forms here.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,57 @@ def data_collection_features
...
+    base_scope = Hmis::Form::Instance.with_role(:OCCURRENCE_POINT)
+
+    # All possible form identifiers used for Occurrence Point collection
+    occurrence_point_identifiers = base_scope.pluck(:definition_identifier).uniq
+
+    # Get the latest version of each form definition so we can check for legacy data
+    # todo @martha - edge case with drafts
```
martha: There is an edge case here with draft forms that I haven't really decided is worth going down the rabbit hole for. What do you think?
- Suppose definition identifier 'foo' has two versions. Definition 0 is published and has question 'baz'. Definition 1, in draft, has the 'baz' question removed. 
- `latest_versions` returns definition 1.
- This wouldn't impact the `best_instance` chosen on line 354, since that is chosen by identifier.
- Suppose no `best_instance` is found for this enrollment, now we look for `has_any_data`.
- Suppose the enrollment has legacy data for 'baz'. We would not catch that, since `has_any_data` is walking definition nodes on definition 1, which lacks the 'baz' question.

Hm 🦆 As I type this out I'm realizing this is not a specific problem with draft forms. It could happen anytime a previous version of a form collected different data than a later version. It just seems especially egregious when the latest version is a draft. 

I don't think it seems at all worthwhile to iterate over *all* previous form definitions (let me know if you agree). 

But maybe for draft forms specifically, we should walk definition nodes on the published version, if one exists.

martha: Ok, I made that change. Let me know how you think it looks!

martha:
Should I retarget to 142? -- this work is pretty independent of the already-merged work on data collection features, but if it's targeting a different release then I'll open a separate ticket in order to track QA.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Hmm, I'm not understanding the scenario where we would want to show a draft form in the application. Can you explain if there is a scenario you are thinking of? In my mind that is a (not-appropriately-documented) boundary in terms of the config tool spec, that drafts should _never_ be used to actually render forms/data. Only `published`/`retired` forms should do that. In terms of the form builder, we have no messaging indicating that drafts could be presented in-app -- it should feel safe to mess around with drafts and be confident that they won't be used anywhere. We should document that if you're on the same page.

My suggestion, with that in mind, would be to adjust the scope below to `published_or_retired.latest_versions` (assuming that achieves the desired effect of dropping drafts but keeping the latest published/retired- I didn't check the sql statement, it may need tweaking).

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,58 @@ def data_collection_features
...
+  def occurrence_point_forms
+    # All instances for Occurrence Point forms, including inactive/retired, so we can check for legacy data.
+    # Draft forms should never have legacy data, but let's include them just in case.
+    base_scope = Hmis::Form::Instance.with_role(:OCCURRENCE_POINT)
+
+    # All possible form identifiers used for Occurrence Point collection
+    occurrence_point_identifiers = base_scope.pluck(:definition_identifier).uniq
+
+    # Get the latest version of each form definition so we can check for legacy data
+    definitions = Hmis::Form::Definition.where(identifier: occurrence_point_identifiers).latest_versions
```
gigxz: This feels a bit roundabout- the `role` column is actually on the _definitions_ table, not the instances table (`with_role` does a join) so we are doing an unnecessary join instead of just saying `definition.where(role: : OCCURRENCE_POINT)`.

Can we simplify to something like this?


```ruby
definitions = Hmis::Form::Definition.where(role: :OCCURRENCE_POINT).published_or_retired.latest_versions

definitions.each do |definition|
    best_instance = definition.instances.active.detect_best_instance_for_enrollment(enrollment: self)
end
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,58 @@ def data_collection_features
...
+
+    definitions.map do |definition|
+      # Choose the most specific instance for each definition identifier
+      instance_scope = base_scope.where(definition_identifier: definition.identifier).order(updated_at: :desc)
+      best_instance = instance_scope.active.published.detect_best_instance_for_enrollment(enrollment: self)
```
gigxz: I'm not following why we need `published` which does another join back to the definition table- can you explain that one? (My suggestion above excludes it, but maybe there is something I'm missing)

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,58 @@ def data_collection_features
...
+            has_any_data = cded && Hmis::Hud::CustomDataElement.where(data_element_definition: cded, owner: self).any?
+          end
+          break if has_any_data # if we found data, we don't need to keep iterating through the items
+        end
+      end
+
+      # If we found legacy data, then return an instance so we can display it. Prioritize active/published
+      if has_any_data
+        best_instance ||= instance_scope.active.published.first # prioritize active/published instances
+        best_instance ||= instance_scope.first # but if none exist, just return the latest instance
```
gigxz: Hmm, maybe it would make more sense to use an OpenStruct like we do for data collection features. I don't see much of a benefit to using an instance as the underlying structure for the `HmisSchema::OccurrencePointForm` type. For legacy data where there is _no_ instance for the form, using an OpenStruct would allow us to resolve it (just like we do with data collection features). We could also add a `legacy` flag. That feels a bit more consistent with data collection features, which now have similar logic. What do you think?

```
OpenStruct.new(
  id: definition.id,
  legacy: has_any_data && !best_instance,
  definition: definition,
  data_collected_about: best_instance&.data_collected_about || 'ALL_CLIENTS',
)
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,58 @@ def data_collection_features
...
+      end
+
+      best_instance
```
gigxz: It looks like we will not be able to resolve this "instance" unless it points to a `definition_identifier` that has a `published` version:
https://github.com/greenriver/hmis-warehouse/blob/3cf565aaab342aa002802b1ff4b796154c623de3/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb#L29

That could be a potential conflict. If retired forms are permitted here, they should be resolvable

gigxz: This point is moot if we go with an OpenStruct since we won't need to re-load the definition https://github.com/greenriver/hmis-warehouse/pull/4877/files#r1842203473

gigxz:
@martha yes lets move to 142, thanks!

Sorry that the spec here wasn't clear in describing the desired behavior for all the various edge cases of form status/form instance combinations, that got tricky.

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Looks good! Left a comment about drafts https://github.com/greenriver/hmis-warehouse/pull/4877#discussion_r1842134995

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: I didn't have a scenario in mind other than "if there is legacy data we should try to display it with whatever form is available." I think you are right that draft forms should not be called upon for that purpose, though.

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: thanks, yeah I was just arriving at the same thing -- it was roundabout because I started with the logic [here](https://github.com/greenriver/hmis-warehouse/blob/dddd4abdd451afd00a8462da75475d97510075dc/drivers/hmis/app/models/hmis/hud/project.rb#L281-L291), but that logic was incorrect given the new spec, because it would exclude published definitions with legacy data that had no instance in the system (even inactive). I'm not sure if I can think of an actual scenario that would lead to that, but maybe migrated-in data?

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: 👍 it's not needed. (I added it here when I removed the `published` scope from the instance base scope.)

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Ok sounds good, I think an OpenStruct is a good way to go both for consistency and for the resolution logic issue you mentioned below. I'd like to add graphql spec tests for this change.

martha: Is there any reason to resolve the `legacy` flag on the frontend? It's a bit different than Data Collection Features, where we need to know the legacy status so we can hide the button for creating a new record.

spec/rails_helper.rb
```diff
@@ -93,11 +93,6 @@
       GrdaWarehouse::WarehouseReports::ReportDefinition.maintain_report_definitions
     end
 
-    # load hmis forms if we're testing the driver
```
martha: I'm proposing to remove this, since I found it confusing that the database was seeded with forms when I ran a single test locally, but that same seeded data wasn't present when the whole test suite ran on CI.

Is this breaking anyone else's workflow? cc @ttoomey?

ttoomey: Happy to remove it entirely if CI passes without it; I don't think it's breaking anyone's workflow. I'd made this conditional as a performance optimization (so iterating on a single spec was faster) but much better if we can just remove it.

Thanks!

martha: Hrm ok I'm seeing that this breaks many tests on ci.
- https://github.com/greenriver/hmis-warehouse/actions/runs/11845216767/job/33010195933?pr=4877
- https://github.com/greenriver/hmis-warehouse/actions/runs/11845216238/job/33010191920?pr=4877
- https://github.com/greenriver/hmis-warehouse/actions/runs/11845216238/job/33010192112?pr=4877

I haven't really grokked why some tests are being run with a command that includes `drivers/hmis` on ci, and many of these tests depend on seeded data being present; while others (that are also in the hmis driver) seem to be run with a broader command that doesn't match the pattern, and these tests rely on seeded data _not_ being present. I think it is not ideal, but would like to split that out into a separate technical debt ticket, if others agree. For now I can just comment these lines out when running the tests individually locally.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -502,8 +502,17 @@ def project_can_accept_referral(destination_project_id:, source_enrollment_id:)
 
     field :client_detail_forms, [Types::HmisSchema::OccurrencePointForm], null: false, description: 'Custom forms for collecting and/or displaying custom details for a Client (outside of the Client demographics form)'
     def client_detail_forms
-      # No authorization required, this just resolving application configuration
-      Hmis::Form::Instance.active.with_role(:CLIENT_DETAIL).published.sort_by_option(:form_title)
+      # No authorization required, this just resolves application configuration
+      Hmis::Form::Instance.active.with_role(:CLIENT_DETAIL).published.sort_by_option(:form_title).
```
martha, line 506: This is to resolve test failures here: https://github.com/greenriver/hmis-warehouse/actions/runs/11846175768/job/33013264262?pr=4877

This definitely begins to feel like scope creep, so if you prefer, we can leave it as-is and have the `HmisSchema::OccurrencePointForm` provide a default value (as it did before). I think I don't have a strong preference, although I guess my feeling is that it's more graceful for the schema OccurrencePointForm to always be backed by an OpenStruct, as opposed to sometimes an OpenStruct and sometimes an Instance?

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: I was thinking that we had discussed a visual treatment indicating that something is "legacy." But we de-scoped that and don't have designs for it. So you're right there is no need to resolve it now.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Oh I forgot about this one. Yeah I agree a graphql type should always be backed by the same object type (Open Struct, now)

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,50 @@ def data_collection_features
...
+          next unless item.mapping.present?
+
+          record_type = item.mapping&.record_type
+          field_name = item.mapping&.field_name&.underscore
+          custom_field_key = item.mapping&.custom_field_key
+
+          next unless record_type == 'ENROLLMENT' || custom_field_key
+
+          if record_type && field_name
+            has_any_data = respond_to?(field_name) && send(field_name).present?
```
gigxz:
```suggestion
           # Example: if this item collects `move_in_date` and the Enrollment has a Move-in Date value, then we want to show this form on the Enrollment Dashboard (even though it isn't "enabled" via an instance)
            has_any_data = respond_to?(field_name) && send(field_name).present?
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,50 @@ def data_collection_features
...
+        definition.walk_definition_nodes(as_open_struct: true) do |item|
+          next unless item.mapping.present?
+
+          record_type = item.mapping&.record_type
+          field_name = item.mapping&.field_name&.underscore
+          custom_field_key = item.mapping&.custom_field_key
+
+          next unless record_type == 'ENROLLMENT' || custom_field_key
+
+          if record_type && field_name
```
gigxz: nit, a bit more readable (to me) would be:
```ruby
has_any_data = if x
 # something
elsif y
 # something
end
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -336,6 +336,50 @@ def data_collection_features
...
+            has_any_data = respond_to?(field_name) && send(field_name).present?
+          elsif custom_field_key
+            # For simplicity, for now, just look for CDEDs where the owner is an Enrollment
+            owner_class_name = self.class.sti_name
+            cded = Hmis::Hud::CustomDataElementDefinition.for_type(owner_class_name).find_by(key: custom_field_key)
+            has_any_data = cded && Hmis::Hud::CustomDataElement.where(data_element_definition: cded, owner: self).any?
```
gigxz: I think we can pull these lookups out so we're not hitting the db so much. For example:

```ruby
# enrollment cdeds that this enrollment has CDE record(s) for:
cdeds_this_enrollment_has = custom_data_element_definitions.pluck(:key).to_set

...

# in the definition/item loops, no need to hit the db:
has_any_data = cdeds_this_enrollment_has.include?(custom_field_key)

```

The `custom_data_element_definitions` relation comes from this concern:
https://github.com/greenriver/hmis-warehouse/blob/e566d8a7f0bf834f548b2dc5086b002210c5af98/drivers/hmis/app/models/hmis/hud/concerns/form_submittable.rb#L16-L17

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
```diff
@@ -14,22 +14,17 @@ class HmisSchema::OccurrencePointForm < Types::BaseObject
...
+    # object is an OpenStruct, see Hmis::Hud::Enrollment occurrence_point_forms
 
     def id(parent:)
       # Include project id (if present) so that instance is not cached for use across projects.
-      [object.id, parent&.id].compact.join(':')
-    end
-
-    def data_collected_about
-      object.data_collected_about || 'ALL_CLIENTS'
+      [object[:id], parent&.id].compact.join(':')
```
gigxz: So the ID was previously  `<instance id>:<project id>` is now `<definition id>:<project id>` is that right?

Maybe it would be a little clearer if we just choose the `id` here within the Type, instead of each place where we create the OpenStruct. That way it's easier to tell what the `id` actually contains. For example:


When creating Open Structs:
```ruby
        OpenStruct.new(
          # id: instance.definition.id, <= removed this
          legacy: false,
          definition: instance.definition,
          data_collected_about: instance.data_collected_about || 'ALL_CLIENTS',
        )
```

And when resolving type:
```ruby
    def id(parent:)
      # Include project id (if present) so that instance is not cached for use across projects.
      [object.definition.id, parent&.id].compact.join(':')
    end
```

Does that make sense?

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Yes for migrated-in data. Hopefully that wouldn't happen since our setup script https://github.com/open-path/Green-River/issues/6771 _should_ make sure that form instances exist for these forms (move-in date, date of engagement, etc) but if for some reason those instances didn't get created, we would hit that case.

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
martha: Yes that makes sense to me. The parent can be either an enrollment or a project now, so I updated to account for that. Thanks for your patience with this PR.

gigxz: Ok I see. thank you for your flexibility with new changes and requirements popping up!

martha:
Okay, @gigxz despite the vulnerability scan build fail, the test suite on this one is finally passing, and I think it's ready to merge if you agree 🙏

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
```diff
@@ -14,23 +14,25 @@ class HmisSchema::OccurrencePointForm < Types::BaseObject
...
-      object.data_collected_about || 'ALL_CLIENTS'
+      [object.definition.id, parent_project(parent)&.id].compact.join(':')
     end
 
     def definition(parent:)
-      definition = load_ar_association(object, :definitions).select(&:published?).first
-      raise "Unable to load definition for instance: #{object.id}" unless definition.present?
-
-      definition.filter_context = { project: parent } if parent.is_a?(Hmis::Hud::Project)
+      definition = object[:definition]
```
gigxz: nit: can access with`object.definition` since it's an Open Struct

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
```diff
@@ -14,23 +14,25 @@ class HmisSchema::OccurrencePointForm < Types::BaseObject
...
+      definition.filter_context = { project: parent_project(parent) }
```
gigxz, line 26: This will result in  attribute `.filter_context = { project: nil }` when `parent` is not passed, whereas before the filter_context would just not get set at all. Looking at the usage in `drivers/hmis/app/graphql/types/forms/form_definition.rb` I think that should be fine

gigxz (approved):
looks good to me, thanks @martha!

PR #4889, opened 2024-10-31
hmis data model erd

ttoomey:
## _Merging this PR_

## Description
It seems helpful to have a diagram of the core HMIS data model to refer to

[preview (updated)](https://github.com/greenriver/hmis-warehouse/blob/28125d8aa8e9392ecb8f0b44d3d7ca2c28c3df75/docs/hud_data_model.md)

## Type of change
Documentation

### DIFF
--- a/docs/hud_data_model.md
+++ b/docs/hud_data_model.md
@@ -28,24 +28,24 @@ erDiagram
     Project ||--o{ Affiliation : "may be affiliated with"
     Project ||--o{ HMISParticipation : "has"
     Project ||--o{ CEParticipation : "has"
-
+    
     Client ||--o{ Enrollment : "has"
-    Enrollment ||--|| Project : "enrolled in"
+    Enrollment ||--|| Project : "has"
     Enrollment }|..|| CoC : "served in"
-    Enrollment ||--o{ Exit : "may have"
+    Enrollment ||--|o Exit : "may have"
     Enrollment ||--o{ IncomeBenefits : "has"
     Enrollment ||--o{ HealthAndDV : "has"
     Enrollment ||--o{ EmploymentEducation : "has"
     Enrollment ||--o{ Disabilities : "has"
-    Enrollment ||--o{ Services : "receives"
+    Enrollment ||--o{ Services : "has"
     Enrollment ||--o{ CurrentLivingSituation : "has"
     Enrollment ||--o{ Assessment : "has"
-    Assessment ||--o{ AssessmentQuestions : "contains"
-    Assessment ||--o{ AssessmentResults : "produces"
+    Assessment ||--o{ AssessmentQuestions : "has"
+    Assessment ||--o{ AssessmentResults : "has"
     Enrollment ||--o{ Event : "has"
     Enrollment ||--o{ YouthEducationStatus : "has"
 
-
+   
 
     Export {
         string ExportID PK
--- a/docs/hud_data_model.md
+++ b/docs/hud_data_model.md
@@ -23,23 +23,18 @@ erDiagram
   CustomServiceType ||--|{ CustomService : "has"
   CustomServiceCategory ||--|{ CustomService : "has"
 
-
-
   CustomAssessment }|--|| Enrollment : "has"
   CustomCaseNote }|--|| Enrollment : "has"
   CustomClientName }|--|| Client : "has"
   CustomClientAddress }|--|| Client : "has"
   CustomClientContactPoint }|--|| Client : "has"
 
-
   CoC }|..|| Enrollment : "served in"
   Project ||--|| Enrollment : "enrolled in"
 
-
   Enrollment }|..|| Household: "in"
 
-
-  Enrollment ||--o| Exit : "may have"
+  Enrollment ||--o| Exit : "has"
   Enrollment ||--o{ IncomeBenefits : "has"
   Enrollment ||--o{ HealthAndDV : "has"
   Enrollment ||--o{ EmploymentEducation : "has"
@@ -52,7 +47,6 @@ erDiagram
   Enrollment ||--o{ Event : "has"
   Enrollment ||--o{ YouthEducationStatus : "has"
   Enrollment }o--|| Client : "has"
-
 
   Client {
       string PersonalID PK
--- a/docs/hud_data_model.md
+++ b/docs/hud_data_model.md
@@ -12,89 +12,107 @@ VERSION 1.4
 * Each Enrollment is tied to exactly one Project and one Client
 * Each Enrollment can have multiple associated services, assessments, and other records
 
-## ERD
+## Enrollment ERD
 
-Note "CoC" is not a defined table; is included in the diagram as an "implied" entity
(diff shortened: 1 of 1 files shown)


eanders (approved):
Thanks @ttoomey !  
This seems like a really good starting point.  Do we want to include any of the custom tables we've added, or are we thinking this should be minimal for now?
One other thought, this seems to indicate the starting point is the client, but maybe we want the starting point to be organization or project?  It probably doesn't matter much since the relationships are the important feature, and arguably the client IS the starting point....

ttoomey:
> Do we want to include any of the custom tables we've added, or are we thinking this should be minimal for now?

I was considering adding our custom data but was concerned that the diagram might become too hard to read. Perhaps we can add another diagram for our custom tables with just subset of the hud tables.

> One other thought, this seems to indicate the starting point is the client, but maybe we want the starting point to be organization or project?

Because the Client element is at the top? I'm not an expert in mermaid but I don't know how much control we have over the layout with this chart type. I think we might be at the mercy of the layout algorithm

eanders:
> > Do we want to include any of the custom tables we've added, or are we thinking this should be minimal for now?
> 
> I was considering adding our custom data but was concerned that the diagram might become too hard to read. Perhaps we can add another diagram for our custom tables with just subset of the hud tables.

Let's hold off on this, and revisit later for additional ERDs.
> 
> > One other thought, this seems to indicate the starting point is the client, but maybe we want the starting point to be organization or project?
> 
> Because the Client element is at the top? I'm not an expert in mermaid but I don't know how much control we have over the layout with this chart type. I think we might be at the mercy of the layout algorithm

That's what I was getting at.  I think you're probably right that we don't control the layout.

docs/hud_data_model.md
```diff
@@ -0,0 +1,100 @@
...
+        string ExportID PK
+        string SourceType
+        string SourceID
+        datetime ExportDate
+        string HashStatus
+    }
+
+    Organization {
+        string OrganizationID PK
+        string CoCCode FK
```
gigxz: This isn't quite right, there is no CoCCode on the Organization.

The CoCCode is on the ProjectCoC, which has a relationship to the Project (Which you already have)

docs/hud_data_model.md
```diff
@@ -0,0 +1,100 @@
...
+
+Note "CoC" is not a defined table; is included in the diagram as an "implied" entity
+
+```mermaid
+erDiagram
+    CoC ||..o{ Organization : "operates in"
+    CoC ||..o{ ProjectCoC : "contains"
+    Export ||--o{ Organization : contains
+    Organization ||--o{ Project : operates
+    Project ||--o{ ProjectCoC : "has locations in"
```
gigxz: "operates in" feels more in line with how we refer to a project that serves multiple cocs. The project doesn't necessarily need to have a physical location there. (Think of street outreach for example that crosses a coc boundary)

docs/hud_data_model.md
```diff
@@ -0,0 +1,100 @@
...
+    Project ||--o{ Funder : "funded by"
+    Project ||--o{ Inventory : "has"
+    Project ||--o{ Affiliation : "may be affiliated with"
```
gigxz: The affiliation should always link back to another project (eg project X is affiliated with project Y). But I don't think that is super important.

docs/hud_data_model.md
```diff
@@ -0,0 +1,100 @@
...
+
+## ERD
+
+Note "CoC" is not a defined table; is included in the diagram as an "implied" entity
+
+```mermaid
+erDiagram
+    CoC ||..o{ Organization : "operates in"
+    CoC ||..o{ ProjectCoC : "contains"
+    Export ||--o{ Organization : contains
```
gigxz: I don't understand this one, it's not part of the HUD spec to my knowledge. ALL the HUD CSV files (tables) have an ExportID that references this record (not just  `Organization`).

docs/hud_data_model.md
```diff
@@ -0,0 +1,100 @@
...
+        string PersonalID PK
+        string FirstName
+        string LastName
+        string SSN
+        date DOB
+        string VeteranStatus
+    }
+
+    Enrollment {
+        string EnrollmentID PK
```
gigxz: Add HouseholdID here? that is an important one, and I think we would consider it a FK?

gigxz:
Looks great @ttoomey, thanks!

I agree it's unfortunate that we can't adjust the positioning at all.

My only idea would be to put the  Project Descriptor relationships in a separate diagram, which would keep this one focused on Enrollment (example below from sandbox). That may also make it more feasible to add custom Enrollment-related elements to this (Eg CustomAssessment) coded with a different color. The Project data is not as confusing as the Enrollment relationships, and it has less churn. One caveat being the CoC relationship which is relevant to explain, as you did.

Just an idea - take or leave it, I don't feel super strongly.

Screenshot 2024-11-04 at 11 00 32 AM


Screenshot 2024-11-04 at 11 01 36 AM

docs/hud_data_model.md
ttoomey: oops, good catch

docs/hud_data_model.md
ttoomey: added the back reference

docs/hud_data_model.md
ttoomey: This was an error, thanks for catching

ttoomey:
> I agree it's unfortunate that we can't adjust the positioning at all.

Actually I discovered that I can adjust the layout by fiddling the order of the relationships. This is kind of tedious but I think I managed to achieve a better layout

> My only idea would be to put the Project Descriptor relationships in a separate diagram, which would keep this one focused on Enrollment (example below from sandbox). That may also make it more feasible to add custom Enrollment-related elements to this (Eg CustomAssessment) coded with a different color. 

That's a great idea, I've split the diagram as you suggest. I'm not sure we can color code the entities though

docs/hud_data_model.md
```diff
@@ -0,0 +1,118 @@
...
+
+
+  CoC }|..|| Enrollment : "served in"
+  Project ||--|| Enrollment : "enrolled in"
+
+
+  Enrollment }|..|| Household: "in"
+
+
+  Enrollment ||--o| Exit : "may have"
```
eanders: Arguably many of these are "may have".  I think I understand the reasoning for calling out Exit differently, but I wonder if we should indicate the optional nature across the board (based on HUD spec) or if we should just change this to "has".

Project has similar optional relationships.

eanders:
Thanks @ttoomey !  I think the new separated ERDs and organization is working well.

gigxz (approved):

docs/hud_data_model.md
ttoomey: fair, I changed it to "has"

Perhaps these labels should be more descriptive? Something like:
* Exit "terminates" enrollment
* Enrollment "reports" or "records" Assessments, CLS, Income, etc

or we could take out the labels entirely if they aren't useful (`Enrollment ||--o| Exit : ""`)

PR #4893, opened 2024-11-01
Add form item level permissions

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6843
- add "editor_user_ids" to the form item JSON schema
- respect it in the form processor by ignoring values passed where the user doesn't have edit access. See inline comment for the reason for this approach.

<details>
<summary>Assessment JSON</summary>

```
{
  "item": [
    {
      "item": [
        {
          "text": "Assessment Date",
          "type": "DATE",
          "initial": [
            {
              "initial_behavior": "IF_EMPTY",
              "value_local_constant": "today"
            }
          ],
          "link_id": "linkid_date",
          "mapping": {
            "field_name": "assessmentDate"
          },
          "required": true,
          "warn_if_empty": false,
          "assessment_date": true,
          "disabled_display": "HIDDEN"
        },
        {
          "text": "How many? This can be edited by everybody",
          "type": "INTEGER",
          "link_id": "how_many",
          "mapping": {
            "custom_field_key": "marthas_custom_assessment_how_many"
          },
          "required": false,
          "read_only": false,
          "warn_if_empty": true,
          "disabled_display": "HIDDEN"
        },
        {
          "text": "What is the answer? This can only be edited by select users",
          "type": "STRING",
          "link_id": "what_is_the_answer_this_can_only_be_edit",
          "mapping": {
            "custom_field_key": "marthas_custom_assessment_what_is_the_answer_this_can_only_be_edit"
          },
          "editor_user_ids": ["3"],
          "disabled_display": "HIDDEN"
        }
      ],
      "type": "GROUP",
      "link_id": "q1",
      "disabled_display": "HIDDEN"
    }
  ]
}
```
</details>

## Type of change
New feature

- [ ] I have provided testing instructions in this PR or the related issue (or not applicable) - not yet, I will add once this is merged to QA and I can provide an example assessment


drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -73,6 +73,15 @@ def run!(user:)
         processor = container_processor(container)
         raise unknown_field_error(definition) unless processor
 
+        # TODO @MARTHA couple things:
+        # - This doesn't work because the frontend includes the value, even if nil
+        #   - We can't just check for nil values, because that would miss users trying to turn an existing value back to nil
+        #   - We probably don't want to take the time to go into every form processor and do this individually, but there is no other way to know whether the attribute has already been filled out or not
+        #   - Alternatively can we get the frontend to stop sending these values, or is that too risky a change?
+        # - with mapped_form_items I'm trying to be generic to other future use cases where we need to know something about the item's attributes when processing the form. Should we somehow try to combine mapped_form_items with the other maps or is that too risky a change?
+        item = mapped_form_items["#{container}:#{field}"]
```
martha: @gigxz I would like to check in about this because I think this is a non-straightforward aspect of the Approach 2 which we settled on. Since we liked Approach 2 for simplicity and ease, before proceeding I want to make sure this doesn't tip us towards a different approach.

Problem:
- In `hud_values_by_container` passed from the frontend, all fields are included, even when the value is nil. So the "raise" condition I've added here incorrectly fails when a user submits a form with non-editable-by-them items on it, even if they haven't actually modified those items
- I think we can't just check for non-nil values, because that would miss turning an existing value back to nil
- In order to compare against the existing saved value, I think we would need to modify every form processor - there is no other way to know whether the attribute has already been saved or not, right?

Possible solutions:
- Approach 2a: modify the behavior of each individual form processor to check if a value has already been saved and if so, if it's different from the value coming from the form submission. This seems more time consuming than is worthwhile for something that is intended to be a temporary solution.
- Approach 2b: modify the behavior of [transformSubmitValues](https://github.com/greenriver/hmis-frontend/blob/release-140/src/modules/form/util/formUtil.ts#L1139) on the frontend so that it doesn't even include keys for items on which the user doesn't have edit permission. (`transformSubmitValues` would need to accept a new param for the current user's ID.)

martha: Approach 2c: Backend can just ignore the value if the user doesn't have access.

Take care to save the original raw state in the hud values to the database though.

gigxz: Discussed: lets just have the backend processor ignore the value if it's a field that the user cannot edit

drivers/hmis/spec/models/hmis/form/form_processor_editor_user_ids_spec.rb
```diff
@@ -0,0 +1,80 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
martha, line 1: This would've gone in `form_processor_spec.rb` but that file is getting so enormous, so I'm proposing to separate it out here.

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -591,6 +596,22 @@ def collect_record_validations(user: nil, household_members: nil)
     end
   end
 
+  # @return <Hash{container_name:field_name=> FormItem }>
+  private def mapped_form_items
```
martha, line 601: I'm trying to present a sort of flexible solution here so that, in case there is ever another future use case where we want to know something about the form item while processing the submitted value, we can reuse this.

I briefly considered trying to combine this with `mapped_custom_form_fields` and `mapped_record_form_fields` so we don't have to construct so many maps: both for the sake of improving code readability and reducing iterations thru the form definition. That could look, for example, like removing those 2, checking against this single map of definition items, and determining by looking at each item whether it's custom or record, and processing accordingly. But I figured that that level of disruption to this battle-tested code is probably not in scope. Let me know what you think.

drivers/hmis/app/models/hmis/form/form_processor.rb
```diff
@@ -73,6 +73,11 @@ def run!(user:)
         processor = container_processor(container)
         raise unknown_field_error(definition) unless processor
 
+        # Check item-level permissions (editor_user_ids) here. The frontend passes all values even if nil, so rather
+        # than raising an exception, just skip processing if we see a value this user doesn't have permission to edit.
+        item = mapped_form_items["#{container}:#{field}"]
+        next if item&.editor_user_ids.present? && !item.editor_user_ids.include?(user.id)
```
ttoomey: I think this could be simplified a bit
```suggestion
        next unless item&.editor_user_ids&.include?(user.id)
```

ttoomey (approved):
this looks good to me, thanks!

drivers/hmis/app/models/hmis/form/form_processor.rb
martha: It can be simplified a bit, but I think we do still need to check the presence of the array separately, since if the array is not present then we should continue processing. (Everyone can edit items that don't specify `editor_user_ids`)

ttoomey: oh right! Thanks

PR #4896, opened 2024-11-04
Fix bug with display text on Prior Living Situation component

martha:
## _Merging this PR_

## Description
I noticed this bug out of the corner of my eye while I was working on system tests (https://github.com/open-path/Green-River/issues/6806). It has since been on my list to investigate/confirm there is a bug and make a bugfix.

**To reproduce:** 
- Open an assessment that renders the more complex version of the Prior Living Situation HUD component (such as an intake in non-SO, ES, or SH projects). [QA example](https://qa-hmis.openpath.host/client/==RTE1YkNaQ0V1SHg5UFZ6NTFDd0NBdz09/enrollments/==OU1LRlRiRmY3YmcrZG1XTTk3UURmQT09/assessments/new/==V2pCMG1TdWNJZGFhOW1FdTFzK0RZQT09)
- Scroll to Prior Living Situation and select any Institutional Situation as the PLS
- For Length of Stay, select "90 days or more" or "One year or longer"
- You see text show up that says "Client does not meet the HUD definition of chronic at entry." This is the text that is only supposed to show up on the Read-Only view of this assessment, like when you've already submitted the intake. On the editable version of the assessment, the [more detailed text](https://github.com/greenriver/hmis-warehouse/blob/release-141/drivers/hmis/lib/form_data/default/fragments/prior_living_situation_b.json#L91) should appear.

**This PR:** removes the "readonly" flag added (probably erroneously?) to that form item, which was what was causing the readonly text to always appear even on editable forms. (You can see elsewhere in the json file that other similar display elements don't have this flag)

**Impact:** This is just a mistake in the instructional text we were showing to users. They should see the more detailed version. But it doesn't impact data collection in any way.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/lib/form_data/default/fragments/prior_living_situation_b.json
+++ b/drivers/hmis/lib/form_data/default/fragments/prior_living_situation_b.json
@@ -87,7 +87,6 @@
         {
           "type": "DISPLAY",
           "link_id": "q_3_917B_A",
-          "read_only": true,
           "text": "Client stayed 90+ days in an institutional setting. This is considered a \"break\" according to the HUD definition of chronic homelessness. Stopping data collection for 3.917 Prior Living Situation.",
           "readonly_text": "Client does not meet the HUD definition of chronic at entry.",
           "enable_behavior": "ALL",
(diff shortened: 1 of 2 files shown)


gigxz (approved):
thanks!

PR #4745, opened 2024-09-24
Update Open HMIS link to support multiple

gigxz:
[//]: # 'remove this if PR is for a release-* branch'
# _Please squash merge this PR_

## Description

https://github.com/open-path/Green-River/issues/6750

Addresses issue on QA HMIS where there are multiple HMIS's. This PR fixes two problems:
1. They show the same link text so are indistinguishable, and
2. Both links show up even for users that only have access to 1 HMIS

Screenshot 2024-10-16 at 8 30 51 PM

How to test: create two HMIS data sources locally.

There should be no change to existing systems (aside from QA HMIS).

## Type of change
- [x] New feature (adds functionality)

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] My code includes comments and/or descriptive variable names to help other engineers understand the intent (or not applicable)
- [x] My code follows the style guidelines of this project (rubocop)
- [x] I have updated the documentation (or not applicable)
- [x] If it's not obvious how to test this change, I have provided testing instructions in this PR or the related issue

PR #4906, opened 2024-11-07
Fix missing Open HMIS link for some users

gigxz:
## _Merging this PR_

## Description
This is a follow-up to https://github.com/greenriver/hmis-warehouse/pull/4745 which had the wrong logic. It was only showing "Open HMIS" link from the Warehouse if the user had any Access Control with a group that contained an entire Data Source. It _should_ show the link to a user that has access to ANY projects or orgs within the DS, which is what this change does.

## Type of change
Bug fix

## How to test
* impersonate a user that only has access to one project or org in hmis. they should see the Open HMIS link.

- n/a If adding a new endpoint / exposing data in a new way, I have:

### DIFF
--- a/drivers/hmis/app/models/hmis/group_viewable_entity.rb
+++ b/drivers/hmis/app/models/hmis/group_viewable_entity.rb
@@ -53,5 +53,14 @@ module Hmis
     scope :includes_entities, ->(entities) do
       where(id: Array(entities).flat_map { |entity| includes_entity(entity).pluck(:id) })
     end
+
+    scope :includes_any_entity_in_data_source, ->(data_source) do
+      project_ids = data_source.projects.select(:id)
+      organization_ids = data_source.organizations.select(:id)
+
+      where(entity: data_source).
+        or(where(entity_type: Hmis::Hud::Project.sti_name, entity_id: project_ids)).
+        or(where(entity_type: Hmis::Hud::Organization.sti_name, entity_id: organization_ids))
+    end
   end
 end
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -181,7 +181,7 @@ class User < ApplicationRecord
   end
 
   def can_access_hmis_data_source?(data_source_id)
-    as_hmis_user&.data_source_ids&.include?(data_source_id)
+    as_hmis_user&.can_access_hmis_data_source?(data_source_id)
   end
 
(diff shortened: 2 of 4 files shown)


drivers/hmis/app/models/hmis/group_viewable_entity.rb
```diff
@@ -53,5 +53,14 @@ class GroupViewableEntity < GrdaWarehouseBase
...
       where(id: Array(entities).flat_map { |entity| includes_entity(entity).pluck(:id) })
     end
+
+    scope :includes_any_entity_in_data_source, ->(data_source) do
+      project_ids = data_source.projects.select(:id)
+      organization_ids = data_source.organizations.select(:id)
+
+      where(entity: data_source).
+        or(where(entity_type: Hmis::Hud::Project.sti_name, entity_id: project_ids)).
+        or(where(entity_type: Hmis::Hud::Organization.sti_name, entity_id: organization_ids))
```
eanders, line 63: I _think_ `GrdaWarehouse::ProjectCollectionMember` might make this simpler.  @ttoomey any chance you can take a quick look?

gigxz: I don't think we have the equivalent for HMIS, though it looks helpful. `GrdaWarehouse::ProjectCollectionMember` looks like it only has relationships to `Collection` which is the warehouse table

eanders: makes sense.

gigxz:
Test failed on `rspec ./spec/models/curl_spec.rb:15 # Curl CA Certificate Works does throw an error when curling over https`, I haven't seen that one before

eanders (approved):

eanders:
> Test failed on `rspec ./spec/models/curl_spec.rb:15 # Curl CA Certificate Works does throw an error when curling over https`, I haven't seen that one before

This is a test that was a canary for a bad certificate or bad implementation of CURL, sounds like it's been fixed, we should review, but not as part of this PR.

drivers/hmis/app/models/hmis/group_viewable_entity.rb
ttoomey: Hmm, project collection member is collection_id, project_id so not directly applicable for this query. Similar shape though

PR #4909, opened 2024-11-07
One-time migration task 20241108

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/6912

One-off migration task. This will need to be run before it's deployed, but figure it should go into into the codebase anyway, and get some review.

Sorry it's a bit of a long mess. See https://github.com/open-path/Green-River/issues/6912#issuecomment-2463110334 for more details.

## Type of change
Data migration

### DIFF
--- a/drivers/hmis_external_apis/lib/tasks/import.rake
+++ b/drivers/hmis_external_apis/lib/tasks/import.rake
@@ -125,6 +125,7 @@ namespace :import do
 
     dry_run = args.dry_run == 'true'
     dir = args.dir
+    project_id = args.project_id
 
     Rails.logger.info "Dry run? #{dry_run}"
 
@@ -166,7 +167,7 @@ namespace :import do
 
     data_source = GrdaWarehouse::DataSource.hmis.first
     system_hud_user = Hmis::Hud::User.system_user(data_source_id: data_source.id)
-    project = Hmis::Hud::Project.hmis.find(args.project_id)
+    project = Hmis::Hud::Project.hmis.find(project_id)
 
     # Read Export file. Each row is a "Case" which represents a HoH Enrollment. It gives us the Entry Date.
     export_file = HmisExternalApis::TcHmis::Importers::Loaders::XlsxFile.new(filename: "#{dir}/HCM-EXPORT.xlsx", sheet_number: 0, header_row_number: 1)
@@ -221,7 +222,10 @@ namespace :import do
     end
 
     # Find Clients. No special care taken to situation where there are multiple clients with the same MCI ID, just pick one.
-    mci_id_to_client = HmisExternalApis::ExternalId.mci_ids.where(value: mci_ids).preload(:source).index_by(&:value)
+    mci_id_to_client = HmisExternalApis::ExternalId.mci_ids.
+      where(value: mci_ids).
+      order(:id).
+      preload(:source).index_by(&:value)
 
     # Build Enrollments
     enrollments = []
--- a/drivers/hmis_external_apis/lib/tasks/import.rake
+++ b/drivers/hmis_external_apis/lib/tasks/import.rake
@@ -312,7 +312,7 @@ namespace :import do
     Rails.logger.info "Skipped Case IDs: #{skipped_case_ids}"
     Rails.logger.info "Skipped MCI IDs: #{skipped_mci_ids}"
     Rails.logger.info "Skipped Case Notes Case IDs (CONTANT_ID): #{skiped_case_notes_ids}"
-    Rails.logger.info "Built #{enrollments.count} Enrollments and #{case_notes.count} Case Notes"
+    Rails.logger.info "Built #{enrollments.count} Enrollments and #{case_notes.count} Case Notes for Project #{project.ProjectName}"
     next if dry_run
 
     # Import Enrollment and Case Note records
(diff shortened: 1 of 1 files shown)


drivers/hmis_external_apis/lib/tasks/import.rake
```diff
@@ -115,4 +115,220 @@ namespace :import do
...
+      25 => 'Shelter',
+      26 => 'Hotel',
+    }
+
+    data_source = GrdaWarehouse::DataSource.hmis.first
+    system_hud_user = Hmis::Hud::User.system_user(data_source_id: data_source.id)
+    project = Hmis::Hud::Project.hmis.find(args.project_id)
+
+    # Read Export file. Each row is a "Case" which represents a HoH Enrollment. It gives us the Entry Date.
+    export_file = HmisExternalApis::TcHmis::Importers::Loaders::XlsxFile.new(filename: "#{dir}/HCM-EXPORT.xlsx", sheet_number: 0, header_row_number: 1)
```
gigxz, line 173: Reused this TcHmis XlsxFile class and it was helpful. Maybe we should move it to a generic utility class -- I'm not sure which `driver` that would go in? Or, I could just move it to another folder within `HmisExternalApis` (out of TcHmis) with the intention to use for any future custom imports that use xlsx.

drivers/hmis_external_apis/lib/tasks/import.rake
```diff
@@ -115,4 +115,220 @@ namespace :import do
...
+      raise "Referral ID #{referral_id} linked to multiple households. unexpected." if referral_id_to_hh_id.key?(referral_id) && referral_id_to_hh_id[referral_id] != household_id
+
+      referral_id_to_hh_id[referral_id] = household_id
+
+      referral_id_to_household[referral_id] ||= []
+      referral_id_to_household[referral_id] << { mci_id: mci_id, relation_to_id: relation_to_id }
+    end
+
+    # Find Clients. No special care taken to situation where there are multiple clients with the same MCI ID, just pick one.
+    mci_id_to_client = HmisExternalApis::ExternalId.mci_ids.where(value: mci_ids).preload(:source).index_by(&:value)
```
ttoomey: perhaps add an order so if there are duplicates, at least it's deterministic 
```suggestion
    mci_id_to_client = HmisExternalApis::ExternalId.mci_ids.where(value: mci_ids).order(:id).preload(:source).index_by(&:value)
```

drivers/hmis_external_apis/lib/tasks/import.rake
```diff
@@ -115,4 +115,220 @@ namespace :import do
...
+    # Read Export file. Each row is a "Case" which represents a HoH Enrollment. It gives us the Entry Date.
+    export_file = HmisExternalApis::TcHmis::Importers::Loaders::XlsxFile.new(filename: "#{dir}/HCM-EXPORT.xlsx", sheet_number: 0, header_row_number: 1)
+
+    skipped_case_ids = []
+    referral_id_to_case_id = {}
+    referral_id_to_entry_date = {}
+    export_file.each do |row|
+      referral_id = row.field_value('ReferralID')
+      case_id = row.field_value('CASE_ID')
+      entry_date_time = row.field_value('OPEN_DT')
```
ttoomey, line 181: should there be format validation or constraints on this date?

drivers/hmis_external_apis/lib/tasks/import.rake
```diff
@@ -115,4 +115,220 @@ namespace :import do
...
+
+      household = referral_id_to_household[referral_id]
+      if !household
+        Rails.logger.info "Household not found for referral ID #{referral_id}. Skipping."
+        skipped_case_ids << case_id
+        next
+      end
+
+      household_id = Hmis::Hud::Base.generate_uuid
+      household.each do |member|
```
ttoomey, line 248: should we check for duplicate household members? (one client in the same household multiple times)

drivers/hmis_external_apis/lib/tasks/import.rake
```diff
@@ -115,4 +115,220 @@ namespace :import do
...
+      case_notes << Hmis::Hud::CustomCaseNote.new(
+        CustomCaseNoteID: Hmis::Hud::Base.generate_uuid,
+        EnrollmentID: enrollment.EnrollmentID,
+        PersonalID: enrollment.PersonalID,
+        data_source_id: enrollment.data_source_id,
+        content: note_content,
+        UserID: user_id,
+        DateCreated: date_created,
+        DateUpdated: date_updated,
+        information_date: information_date.to_date,
```
ttoomey, line 310: should these all be '.to_date' ?

drivers/hmis_external_apis/lib/tasks/import.rake
ttoomey: 👍 to making it generic. Maybe HmisExternalApis::Importers::XlsxFile

ttoomey (approved):
Looks good!

drivers/hmis_external_apis/lib/tasks/import.rake
gigxz: No, only Information Date is a date. The timsetamps are datetimes (which are in the file). Double-checked that these are being set correctly

drivers/hmis_external_apis/lib/tasks/import.rake
gigxz: I dont think so, that's an acceptable data quality issue. Similarly I am not checking for multiple HoHs or no HoHs. Those are things they can clean up, it is a small enough data set.

drivers/hmis_external_apis/lib/tasks/import.rake
gigxz: added, thanks

PR #4910, opened 2024-11-07
Add file to CDE

martha:
## _Merging this PR_
- **Should go into release-145**, because frontend could still be at release-143 when backend points at release-144, and [this change](https://github.com/greenriver/hmis-frontend/pull/987) which is needed for compatibility will go into release-144

## Description
GH issue: https://github.com/open-path/Green-River/issues/6208

This PR
- Adds a new column to CustomDataElement table, `value_file_id`, alongside the existing `value_string`, `value_json`, etc.
- Adds the field to the model and graphql type
- Enables saving a CustomDataElement whose value refers to a file record
  - Filters out such files from the Client Files page
- Prevents "save and finish later" on assessments which save file records (See [inline comment thread](https://github.com/greenriver/hmis-warehouse/pull/4910#discussion_r1866789333) for reasoning)

To test:
- Checkout https://github.com/greenriver/hmis-frontend/pull/982. In the description of that PR I pasted the assessment json definition I've been using locally. (Also, that PR makes some minor updates to the Form Builder enabling the File question type to be configurable, so you could try that out too.)

## Type of change
New feature

  - [ ] ensured permissions and visibility checks are performed in the right places - to confirm: files uploaded on an assessment should be visible to anyone who has permissions to view the assessment?


drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +290,28 @@ def normalize_single_custom_field_value(cded, value)
       else
         raise "unexpected value \"#{value}\""
       end
+    when 'file'
```
martha, line 304: Overall, is this the right place for this logic to go? I was wondering if there's something cleverer I could do with the [FileProcessor](https://github.com/greenriver/hmis-warehouse/blob/release-142/drivers/hmis/app/models/hmis/hud/processors/file_processor.rb), but I wasn't sure how to adapt that processor, which is geared towards processing individual attributes onto a File record, to make it compatible with this new usage, which is geared towards processing a Custom Data Element that _points_ at a File record.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +290,28 @@ def normalize_single_custom_field_value(cded, value)
...
+        file.enrollment = @processor.send(:enrollment_factory)
+        file.client_file.attach(blob)
+        file.user = @processor.current_user
+        file.updated_by = @processor.current_user
+        file
+      when Object
+        # If the value is an Object, it should be a File record in our db
+        raise "unexpected value \"#{value}\"" unless value.key?('fileBlobId') && value.key?('id')
+
+        Hmis::File.find(value['id'])
```
martha: Should the File's `updated_by_user_id` change to point at a user who submits the assessment but doesn’t change the file? Rn the behavior is to update the CDE updated date without changing the file itself doesn’t change.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +290,28 @@ def normalize_single_custom_field_value(cded, value)
...
       end
+    when 'file'
+      case value
+      when String
+        # If the value is a string, it should be a blob ID for a file that's just been uploaded to ActiveStorage,
+        # but doesn't yet have a File record in our database.
+        blob = ActiveStorage::Blob.find_by(id: value)
+        file = Hmis::File.new
+        file.name = blob.filename
+        file.client = @processor.send(:client_factory)
```
martha: This makes files added in an Assessment show up under Client Files, which may not be desirable.

drivers/hmis/app/models/hmis/hud/custom_data_element.rb
```diff
@@ -25,12 +25,15 @@ class Hmis::Hud::CustomDataElement < Hmis::Hud::Base
...
     :value_string,
     :value_text,
+    :value_file,
   ].freeze
 
   belongs_to :owner, polymorphic: true, optional: false
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
   belongs_to :user, **hmis_relation(:UserID, 'User'), optional: true, inverse_of: :custom_data_elements
   belongs_to :data_element_definition, class_name: 'Hmis::Hud::CustomDataElementDefinition', optional: false
+  belongs_to :value_file, class_name: 'Hmis::File', optional: true, autosave: true, foreign_key: 'value_file_id'
```
martha: This adds a relationship going from the CDE to the file, but I haven't configured anything in the opposite direction, so right now you can't easily get from the file back to the CDE. Would that be desirable?

drivers/hmis/app/models/hmis/hud/processors/base.rb
martha, line 304: Just to note, saving without submitting is currently not really working. I'll start a thread to ask about this on the frontend PR.

gigxz: This is inside of the normalization function, but this seems to really go above and beyond normalizing to actually process the blob id into a file object.

Maybe we can call this separately after normalization occurs, something like:
```ruby
    # Normalize the input value
    value = normalize_custom_field_value(cded, value)
    # Transform file blob id into (new or existing?) File. We will save the reference to this File on the CDE
    value = process_blob_id_to_file(value) if cded.field_type == :file
```

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +290,28 @@ def normalize_single_custom_field_value(cded, value)
       else
         raise "unexpected value \"#{value}\""
       end
+    when 'file'
+      case value
+      when String
+        # If the value is a string, it should be a blob ID for a file that's just been uploaded to ActiveStorage,
+        # but doesn't yet have a File record in our database.
+        blob = ActiveStorage::Blob.find_by(id: value)
```
gigxz: Should we check to see whether this blob id is already persisted? (Maybe that's not expected, but it still seems like something we should handle gracefully)

And can we add a specific error for the case where the `ActiveStorage::Blob` is not found?

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: Yeah let's take another look at how we should structure this. We don't want them to show up there or in the warehouse.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +290,28 @@ def normalize_single_custom_field_value(cded, value)
...
+        file = Hmis::File.new
+        file.name = blob.filename
+        file.client = @processor.send(:client_factory)
+        file.enrollment = @processor.send(:enrollment_factory)
```
gigxz: What happens if there is no enrollment returned from the factory? Can we validate that? I don't think we want to store any files without enrollments. (`enrollment_id` is optional on `files`)

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: No, I don't think the updated_by_user_id on the file should be updated when the assessment is re-submitted

drivers/hmis/app/models/hmis/hud/processors/base.rb
martha: ok, I added a filter in the Client schema object, let me know if you think this is the right place to put it.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -177,6 +177,15 @@ def file(id:)
       Hmis::File.viewable_by(current_user).find_by(id: id)
     end
 
+    # don't paginate since we don't expect many records passed (todo @martha - discuss, maybe not ideal)
+    field :file_blobs, [Uploads::DirectUploadType], null: true do
+      argument :ids, [ID], required: true
+    end
+    def file_blobs(ids:)
+      # todo @martha PERMISSION LEAK
+      ActiveStorage::Blob.find(ids)
```
martha: @gigxz Here is a sketch of the approach we discussed last time for how we would display previously saved-but-not-submitted assessment files.

There are some bugs to unravel with this approach from the frontend, but before doing so, I'd like to discuss what I see as (potentially) a fatal flaw to the approach, which is: at the point when we have uploaded a DirectUpload blob, but have not yet saved a file record to the DB, it's difficult [not possible?] to track down what client/enrollment this file is associated with, and by extension, difficult [impossible?] to appropriately gate permissions for any endpoint where we directly expose such blobs.

I'm not sure, maybe there is a simple way around this that I'm just not understanding? But to me it is starting to look like we are required to save the file as a file record on save-without-submit, and then return it on subsequent fetches of the assessment. That approach is messy too, because it requires circumventing a bunch of form processor stuff that normally only runs on submit. But I can't think of a cleaner way to get the behavior we want without a permission leak.

martha: Decision from pairing: only support WIP saving for non file assessments. New API field `supports_save_in_progress`. Make it also return false if assessment has already been submitted

gigxz: @martha great point, thank you for catching that! I agree this would expose a potential leak. I also agree that adding a processing step to "save in progress" with a partial save (saving only Files) is a can of worms.

I think our best course of action right now may be to disallow saving files as "in progress" as discussed

drivers/hmis/app/graphql/types/forms/form_definition.rb
```diff
@@ -118,6 +119,10 @@ def updated_by
       load_last_user_from_versions(object)
     end
 
+    def supports_save_in_progress
```
martha, line 122: @gigxz contrary to our discussion, I ended up putting this here on the Form Definition instead of on the Assessment, since the Assessment may not exist yet at the point where we make the determination about whether to show the button. ([relevant frontend code](https://github.com/greenriver/hmis-frontend/pull/982/files#diff-daa760be4e03c0307f7d08234bd1dfdf4f24ad26040b196b043da24237def3ebR72-R76))

drivers/hmis/app/models/hmis/file.rb
```diff
@@ -34,6 +34,12 @@ def data_source
   belongs_to :client, class_name: '::Hmis::Hud::Client'
   belongs_to :user, class_name: 'Hmis::User', optional: true
   belongs_to :updated_by, class_name: 'Hmis::User', optional: true
+  has_one :custom_data_element, class_name: 'Hmis::Hud::CustomDataElement', foreign_key: 'value_file_id', inverse_of: :value_file
+
+  def self.find_by_blob_id(blob_id)
+    attachment = ActiveStorage::Attachment.find_by(blob_id: blob_id)
```
martha: To do pending https://github.com/greenriver/hmis-warehouse/pull/4980

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -151,32 +151,43 @@ def process_custom_field(field, value)
...
       else
         cde_attributes[value_field_name] = value
         cde_attributes.merge!(attrs)
       end
-    # If value(s) haven't changed, just update the User and timestamps
-    elsif existing_values.map(&:value) == Array.wrap(value)
-      cde_attributes = existing_values.map { |cde| { id: cde.id, user: @processor.hud_user } }
-    # Else create new custom field value(s), and delete any existing ones.
     else
+      submitted = Array.wrap(value)
```
martha, line 174: Currently, when any CDE value changes, we delete and replace all of them. See [regression test](https://github.com/greenriver/hmis-warehouse/pull/4910/files#diff-1d09bfc99a44479d26529c3d5ef39deb9f1e365bcb470c9984dd77923fef66daR1804-R1822) farther down on this PR. I think this is a bug, but until now it wouldn't have had any impact except that such CDEs would have incorrect creation timestamps. The reason to fix it now is to prevent deletion and re-creation of File records.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -229,7 +229,8 @@ def custom_case_notes(...)
     end
 
     def files(**args)
-      resolve_files(**args)
+      # Exclude files that have been uploaded by custom assessments
+      resolve_files(object.files.left_joins(:custom_data_element).where(custom_data_element: { id: nil }), **args)
```
eanders: This isn't really a problem, and will probably be fine going forward for the `id` column, but generally we write these as:
```suggestion
      resolve_files(object.files.left_joins(:custom_data_element).merge(Hmis::Hud::CustomDataElement.where(id: nil)), **args)
```

That way we aren't confused by the rails-ism where the table name is probably pluralized, but the hash key isn't.  Basically it just makes the model tell you the scope to merge in.

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -413,6 +413,15 @@ def valid_status_for_submit?
     published? || retired?
   end
 
+  def supports_save_in_progress?
+    walk_definition_nodes(as_open_struct: true) do |item|
+      # Saving assessments with File item types isn't supported
+      return false if item.type == 'FILE'
```
eanders: is there any benefit to calling `.upcase` here to ensure the casing is the same, or are we confident that `.type` will always be upcased?

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -271,13 +271,15 @@ def check_cdeds(document, role)
...
-        when 'GROUP', 'OBJECT', 'FILE', 'IMAGE'
+        when 'GROUP', 'OBJECT'
           # We don't expect these types to have custom field mappings. If they do, raise an error
           raise "Item #{link_id} has type #{item_type}, so it should not have a custom_field_key"
         when 'DISPLAY'
           # DISPLAY types should really be in the above category too,
           # but we have existing cases that store an autofill value
           next
+        when 'FILE', 'IMAGE'
+          raise_bad_type_match(link_id, item_type, cded_key, cded_type) unless cded_type == 'file'
```
eanders, line 282: same question here with `.downcase`

eanders (approved):

drivers/hmis/app/models/hmis/form/definition.rb
martha: I think we are confident that it will always be uppercase, because otherwise it would be an invalid form definition ([schema here](https://github.com/greenriver/hmis-warehouse/blob/681544cdafe35c9ec96ccc0e706a14cb371382a3/drivers/hmis_external_apis/public/schemas/form_definition.json#L4)).

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: I think we are confident it will always be lowercase, because `cded.field_type` is generated based on which of the `value_` columns is non-nil. https://github.com/greenriver/hmis-warehouse/blob/681544cdafe35c9ec96ccc0e706a14cb371382a3/drivers/hmis/app/models/hmis/hud/custom_data_element.rb#L62

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -413,6 +413,15 @@ def valid_status_for_submit?
     published? || retired?
   end
 
+  def supports_save_in_progress?
```
gigxz, line 416: Makes sense that this is on the definition!
I think this could be slightly optimized by returning immediately if this isn't an assessment definition.
The file search can also be a _little_ more succinct by using link_id_item_hash:
```
    # Only Assessments can be saved as WIP
    return false unless ASSESSMENT_FORM_ROLES.include?(role.to_sym)
    # If the assessment contains a File collection item, it cannot be saved as WIP (# link issue nr?)
    return false if link_id_item_hash.values.find? { |item| item.type == 'FILE' }

    true
```

gigxz: Oh and perhaps including `IMAGE` type as well

drivers/hmis/app/models/hmis/hud/custom_data_element.rb
```diff
@@ -30,12 +30,15 @@ class Hmis::Hud::CustomDataElement < Hmis::Hud::Base
     :value_json,
     :value_string,
     :value_text,
+    :value_file,
```
gigxz: I believe this should be `value_file_id` since the array contains the column names. Looking at the usage, we don't need to load the file every time-- for example when checking `equal_for_merge?` its sufficient to compare the file ids. (Not that you would ever have two referencing the same file..)

drivers/hmis/app/models/hmis/form/definition_validator.rb
gigxz: I agree no need to downcase, but don't think the referenced code is applicable. On line 271 we assign `cded_type = cded.field_type` which is what's being referenced here. It looks like we maybe do still need a rails validation constraint on the `CustomDataElementDefinitions.field_type` column, the values are expected to match these:
https://github.com/greenriver/hmis-warehouse/blob/0c5bd1c35388d1f618ceea8ce23ed9e961c7886e/drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb#L14-L22

(Where you already added `file`).

I do worry that we are breaking the following assumption:
* The `CustomDataElement` has an attribute `value_[x]` where "x" matches the `CustomDataElementDefinition.field_type`.

That assumption is broken because the field name is `value_file_id` whereas the field_type is `file`. I think that could cause problems, but maybe it's masked because `value_file` is still an assignable attribute?

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +301,18 @@ def normalize_single_custom_field_value(cded, value)
       else
         raise "unexpected value \"#{value}\""
       end
+    when 'file'
+      case value
+      when String
+        value
```
gigxz, line 308: do we have any expectations about this string? it would be helpful to add a note explaining what we expect it to be.

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -290,6 +301,18 @@ def normalize_single_custom_field_value(cded, value)
...
+      when Object
+        # If the value is an Object, it should be a File record in our db
+        raise "unexpected value \"#{value}\"" unless value.key?('fileBlobId') && value.key?('id')
+
+        Hmis::File.find(value['id'])
```
gigxz: I think it's a little confusing for this lookup to be in the normalization function instead of in the `process_blob_id_to_file` method, can we move it down?
If we are expecting an object maybe we should just normalize it into the string that we want to process, so returning `value['id']` here? I'm still wrapping my head around why its an object, but will keep digging into that

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -151,32 +151,43 @@ def process_custom_field(field, value)
...
+
+      # Helper lambda to check if an existing value was also in the submitted values
+      is_in_submitted = ->(cde) { submitted.include?(cde.value) }
+
+      new_values = submitted - existing_cdes.map(&:value)
+      unchanged = existing_cdes.filter(&is_in_submitted)
+      to_remove = existing_cdes.reject(&is_in_submitted)
+
       cde_attributes = [
+        # If value(s) haven't changed, just update the User and timestamps
```
gigxz: so minor, but since this is complicated code,
```suggestion
        # For value(s) that haven't changed, just update the User and timestamps
```

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: Ahh yep, thank you for addressing that!
How will it handle situations where there are repeated values that exist or are submitted? Like `[{value_string: "foo"}, {value_string: "foo"}]` (I'm thinking on an open-choice field, perhaps).
I don't think it's really necessary to preserve both values if they are identical, but it shouldn't explode or do something weird- maybe worth having a test case.

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: Thanks Gig. I can add a validation. I think you are right about the assumption breaking, but I can't think of a good way around it. I played around with it a bit, but ran up against the fact that in the form processing code, when the value is saved, we can't assign to `value_file_id` because we don't have a file ID yet, since the file record hasn't yet been saved.

drivers/hmis/app/models/hmis/hud/processors/base.rb
martha: Thanks, yeah I agree it is confusing. After reading your feedback and taking another look, I think it would be best not to pass around the whole file object. I'm almost there with a tweaked approach - coming soon ...

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: Hm ok I'm proposing to remove this validation for now since it causes test failures. (https://github.com/greenriver/hmis-warehouse/actions/runs/12188033326/job/34000177154) This PR is already way over estimate so I'd like to descope fixing the tests.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -1801,6 +1801,45 @@ def add_cde_item_to_definition
...
+
+          expect(record.custom_data_elements.size).to eq(2)
+          expect(record.custom_data_elements.map(&:value_string)).to contain_exactly('foo', 'foo')
+
+          hud_values = complete_hud_values.merge(
+            cded.key => ['foo', 'foo', 'foo', 'bar'],
+          )
+          process_record(record: record, hud_values: hud_values, user: hmis_user, definition: definition)
+
+          # expect(record.custom_data_elements.size).to eq(4)
```
martha: The commented-out line fails, since `custom_data_elements.size` is equal to 3. The last `foo` value doesn't get properly added because it's considered an existing value.

I _think_ the most graceful way to deal with this is to just prevent submission of duplicate values. That way we don't have to keep track of value counters or anything like that. Maybe in `normalize_custom_field_value`?

drivers/hmis/app/models/hmis/hud/processors/base.rb
martha: I added a test case. It doesn't explode but it does do something weird - see comment below https://github.com/greenriver/hmis-warehouse/pull/4910/files#r1876005116

drivers/hmis/app/models/hmis/hud/processors/base.rb
martha: Ok, I think this ready for re-review. 
This won't work with regular blob ID, only a signed blob ID; so I think it would need to go out in the next release after https://github.com/greenriver/hmis-frontend/pull/987. But I'd like to think this through a little more together.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
martha: I implemented this, let me know what you think

drivers/hmis/app/models/hmis/hud/processors/base.rb
```diff
@@ -300,4 +319,44 @@ def safe_parse_date(string)
...
+    end
+  end
+
+  private def process_id_to_file(value)
+    return value unless value
+
+    if value.to_i.to_s == value.to_s
+      # If value is a string representation of an integer, then it should be the ID of an existing File record.
+      # This is a file we've seen and saved before.
+      Hmis::File.find(value)
```
martha, line 341: Access control check here

martha: is this custom data element already associated with this custom assessment (the owner of this processor)? throw an error if not

drivers/hmis/app/models/hmis/hud/processors/base.rb
gigxz: Thanks @martha this looks good! Yes as discussed let's re-target this to 145

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -229,7 +229,8 @@ def custom_case_notes(...)
     end
 
     def files(**args)
-      resolve_files(**args)
+      # Exclude files that have been uploaded by custom assessments
+      resolve_files(object.files.left_joins(:custom_data_element).merge(Hmis::Hud::CustomDataElement.where(id: nil)), **args)
```
gigxz: This is equivalent to `.where.missing(:custom_data_element)`! learned that one recently, I think it's neat

```rails
> Hmis::File.where.missing(:custom_data_element).count
  Hmis::File Count (3.0ms)  SELECT COUNT(*) FROM "files" LEFT OUTER JOIN "CustomDataElements" "custom_data_element" ON "custom_data_element"."DateDeleted" IS NULL AND "custom_data_element"."value_file_id" = "files"."id" WHERE "files"."type" = $1 AND "files"."deleted_at" IS NULL AND "custom_data_element"."id" IS NULL  [["type", "Hmis::File"]]

> Hmis::File.left_joins(:custom_data_element).merge(Hmis::Hud::CustomDataElement.where(id: nil)).count
  Hmis::File Count (3.3ms)  SELECT COUNT(*) FROM "files" LEFT OUTER JOIN "CustomDataElements" ON "CustomDataElements"."DateDeleted" IS NULL AND "CustomDataElements"."value_file_id" = "files"."id" WHERE "files"."type" = $1 AND "files"."deleted_at" IS NULL AND "CustomDataElements"."DateDeleted" IS NULL AND "CustomDataElements"."id" IS NULL   [["type", "Hmis::File"]]
```

gigxz (approved):
nice work!!

PR #4912, opened 2024-11-08
6730 inventory and scrub pii

ttoomey:
## _Merging this PR_

## Description

* Mechanism to track PII in columns, see `HasPiiAttributes`
* Inventory and track current PII in warehouse and reporting tables
* Task to scrub client PII
* Task to scrub all PII

See [ADR-0002](docs/adr/0002-pii-management-strategy.md)

[Ref issue](https://github.com/open-path/Green-River/issues/6730)

### Testing the scrubbers

Note this task must be invoked manually.

```ruby
# logger messes with the progress bar display
Rails.logger.level=1
# scrub all clients in data-source 8, use fake values if possible
GrdaWarehouse::Tasks::ScrubPii::ScrubClientPiiTask.perform(custom_scrubber: :fake, progress: true, data_source_ids: [8])
# scrub all PII in all tables
GrdaWarehouse::Tasks::ScrubPii::ScrubAllPiiTask.perform
```

## Type of change
Feature


app/models/grda_warehouse/tasks/scrub_pii/base_strategy.rb
```diff
@@ -0,0 +1,75 @@
...
+        ProjectID: enrollment.ProjectID,
+        EntryDate: enrollment.EntryDate,
+        # fields to overwrite
+        LastPermanentStreet: nil,
+        LastPermanentCity: nil,
+        LastPermanentState: nil,
+        LastPermanentZIP: nil,
+        AddressDataQuality: 99,
+        last_locality: nil,
+        last_zipcode: nil,
```
ttoomey: probably all deprecated fields?

app/models/grda_warehouse/tasks/scrub_pii/scrub_pii_task.rb
```diff
@@ -0,0 +1,198 @@
...
+          arel[:label].matches('%phone%'),
+        ].reduce(&:or)
+
+        scope = Hmis::Hud::CustomDataElement.
+          where(owner_type: model.sti_name, owner_id: ids).
+          joins(:data_element_definition).
+          merge(Hmis::Hud::CustomDataElementDefinition.where(conditions))
+
+        delete_versions(Hmis::Hud::CustomDataElement, scope.pluck(:id))
+        scope.delete_all
```
ttoomey: we could also use the strategy to replace the CDEs rather than deleting them

eanders (commented):
@ttoomey this looks like a solid start, but I wonder if the scrubber needs to be less specific.  Something that could take a class, scope, strategy, and set of columns.

```
klass = HudApr::Fy2020::AprClient
scope = HudApr::Fy2020::AprClient.where(report_instance_id: 10)
scrub(klass, scope: scope, strategy: :fake, columns: { first_name: :name, last_name: :name, ssn: :identifier, dob: :date })
```

One of the tasks we need to do is document everywhere we have PII that may need to be centralized and scrubbed, I think the pattern you've setup for clients is great, but probably needs an increased scope.

ttoomey:
notes from conversation with @eanders :
* Let's build an additional task that can clean reports
* Report records may be orphaned (not associated client pk) so we cannot rely on joining from the client
* These tasks should be able to descend though the graph from an entry point (client, report record) and execute a strategy. Eventually strategies other than scrubbing might be needed; in particular we might want to use the task to assist with migration to a centralized PII store, such as by building or reporting on the relationships for migration.

app/models/grda_warehouse/tasks/scrub_pii/scrub_all_pii_task.rb
```diff
@@ -0,0 +1,94 @@
...
+        TextMessage::TopicSubscriber,
+        GrdaWarehouse::HealthEmergency::UploadedTest,
+        GrdaWarehouse::Contact::Base,
+        GrdaWarehouse::Hmis::Staff,
+        GrdaWarehouse::HmisClient,
+        # warehouse.ClientUnencrypted, # no model?
+
+        # hud records
+        GrdaWarehouse::Hud::Client,
+        GrdaWarehouse::Hud::User,
```
gigxz, line 44: This feels a bit out of place. All the other records are scrubbing client data, whereas this is scrubbing user data. We are not clearing out the `users` table, so why bother with the hud User. I dont _think_ it would be a problem, just seems different from the other pii here

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
```diff
@@ -0,0 +1,115 @@
...
+module GrdaWarehouse::Tasks::ScrubPii
+  # * Selectively scrub personally identifiable information (PII) for HMIS Clients
+  # * Delete client-related custom hmis data
+  # * Delete versions
+  class ScrubClientPiiTask
+    def self.perform(...)
+      new.perform(...)
+    end
+
+    def perform(client_ids: nil, data_source_ids: nil, custom_scrubber: nil, progress: false)
```
gigxz, line 17: Can you say more about the intended use-case for this task?
The original ask was around cleaning up an environment containing PII, without the need to really retain anything except client IDs. This all looks good to me for that purpose.

Since this accepts `client_ids`, should I be reviewing this as if we were to use it to remove PII for an individual client upon request?(Or if certain length of time has passed since enrollment, etc). These are use-cases we have discussed in the past, but haven't been totally spec'd out afaik. If that's the intended use for this then I may have some different feedback, for example clearing the content of the notes rather than fully deleting all the notes.

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
```diff
@@ -0,0 +1,115 @@
...
+      ].each do |model|
+        scope = model.with_deleted.where(data_source_id: data_source_id, PersonalID: personal_ids)
+        delete_versions([model], scope.map(&:id))
+        scope.delete_all
+      end
+
+      # delete custom data elements from tables that may contain sensitive info
+      [
+        Hmis::Hud::CustomAssessment,
+        # other models?
```
gigxz: any model with `include ::Hmis::Hud::Concerns::FormSubmittable` could conceivable have a custom data element on it from a form. Looking at CustomAssessment, Service, Enrollment, Client, CurrentLivingSituation would be most likely to have pii

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
```diff
@@ -0,0 +1,115 @@
...
+      delete_versions([GrdaWarehouse::Hud::Client, Hmis::Hud::Client], client_ids)
+
+      clients.group_by(&:data_source_id).map do |data_source_id, ds_clients|
+        personal_ids = ds_clients.map(&:PersonalID)
+        delete_hmis_client_records(personal_ids, data_source_id)
+        delete_enrollment_related_records(personal_ids, data_source_id)
+      end
+    end
+
+    def delete_custom_data_elements(model, ids)
```
gigxz: This could have a clearer name to indicate that it's not wholesale deleting ALL the custom data elements on the model. `delete_custom_data_elements_with_pii`?

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
gigxz: And vice versa, for the _first_ use case (environment cleanup) it would be fine to just drop _all_ the custom data elements, notes, etc.- we're retaining more than we really need for that use case.

app/models/grda_warehouse/tasks/scrub_pii/scrub_all_pii_task.rb
ttoomey: Good question. I would consider staff data as PII as well but maybe that's incorrect. I see the distinction from users is that lower-case `users` is authentication/auth information (configuration not data) so sanitizing it would break functionality. It's a good point that, to be consistent, we should define PII fields on the users table as well. I'll add that

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
ttoomey: > Since this accepts client_ids, should I be reviewing this as if we were to use it to remove PII for an individual client upon request?

Yes, that's the intention of this task. It's trying to selectively delete PII from certain client records; the other task is intended to comprehensively scrub all PII from the entire database. I'll add some better inline comments to make that clear.

> I may have some different feedback, for example clearing the content of the notes rather than fully deleting all the notes.

The original spec had "Free-form text (delete) - case notes, custom fields?"  which I took as delete. I think generally sanitizing fields rather than deleting is better so I'm happy to make that change. Two points:
* Some fields probably can't be nulled out since a case-node or cde with a null value could break the app. Safe to replace with a static or faker-value
* Perhaps we shouldn't try to select the subset of sensitive fields from CDE's since we might miss some. We could instead sanitize all CDEs. Maybe we could check-in on this briefly

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
ttoomey: Good suggestion, I'll rename it. Perhaps we can replace this in the future with a more formal formal pii tracking field, such as  `Hmis::Hud::CustomDataElement.where(pii_level: ..2)`

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
ttoomey: Great, I'll add those. Thanks!

app/models/concerns/has_pii_attributes.rb
```diff
@@ -0,0 +1,113 @@
...
+    age: { level: 3 },
+    geo_locality: { level: 3 },    # city/town is less sensitive than street
+    geo_admin_1: { level: 3 },     # state/province
+    geo_admin_2: { level: 3 },     # county/district
+
+    # Level 4 - Contextual Identifiers
+    # These fields may contain embedded PII in unstructured form
+    free_text: { level: 4 }, # case notes - sensitivity varies by content
+    url: { level: 4 }, # depends on content/context
+    attached_file: { level: 4 },
```
eanders, line 65: Thanks for catching all these, especially the file.  I was thinking about how we might handle HMIS CSV exports, or prior LSA runs.  These may have attached files with client names, so great to see that this should cover those scenarios.

app/models/concerns/hmis_structure/client.rb
```diff
@@ -7,13 +7,20 @@
...
     self.ignored_columns += [:search_name_full, :search_name_last]
     self.hud_key = :PersonalID
     self.additional_upsert_columns = [:demographic_dirty]
     acts_as_paranoid(column: :DateDeleted) unless included_modules.include?(Paranoia)
+
+    pii_attr :SSN
+    pii_attr :FirstName
+    pii_attr :MiddleName
+    pii_attr :LastName
+    pii_attr :DOB
```
eanders: @ttoomey we've aliased these to underscore format like `:first_name` which can be used in most contexts, do you know if those would be a problem here?  I'm leaning toward what you did (use the same shape as the database column name) but it might be nice to use the aliased versions if that would work.

ttoomey: it should be fine to convert these to snake case, either should work. I think it's probably better to use snake case generally, I can change that

app/models/concerns/has_pii_attributes.rb
ttoomey: well we support defining it but it's not yet handled by the scrubber. The scrubber doesn't deal with level 3/4 fields right now although we should add that.

app/models/grda_warehouse/tasks/scrub_pii/scrub_all_pii_task.rb
eanders: I _think_ we want that here, but definitely understand that this is not Client PII, but someone else's PII.  Similar to client contacts, or case managers.  It is an interesting question if we were scrubbing PII for a single client, would we also scrub the associated `User` records?

app/models/pii/scrubber/dob_scrubber.rb
```diff
@@ -0,0 +1,71 @@
...
+    #   new_dob = scramble_dob(today - current, fuzz_years: fuzz_years)
+    #   (today - new_dob).years
+    # end
+
+    # Scrambles a date of birth while preserving approximate age bracket of the original
+    # For example, with a 5-year bracket, someone aged 32 will get a DOB corresponding to age 30-35.
+    def scramble_dob(current, fuzz_years: 5)
+      return nil unless current
+
+      age_at_scrub = ((today - current) / 365.25).floor
```
eanders: maybe a bit DRYer?
```suggestion
      age_at_scrub = age_in_years(current)
```

app/models/pii/scrubber/replacement_pii.rb
```diff
@@ -0,0 +1,48 @@
...
+        Faker::IdNumber.invalid.gsub(/^\d{3}/, '999')
+      when :first_name
+        Faker::Name.first_name
+      when :last_name
+        Faker::Name.last_name
+      when :middle_name
+        Faker::Name.middle_name
+      when :full_name
+        Faker::Name.name
+      end
```
eanders, line 46: This doesn't encompass all of the keys in `STATIC_TYPE_VALUES`.  I haven't found where we're calling `fake_value` but should this extend to cover the whole set of keys or raise on unknown key?

eanders: Or maybe it just return `nil` which is fine for the other keys

app/models/pii/scrubber/scrub_model_pii.rb
```diff
@@ -0,0 +1,114 @@
...
+
+          # scrub remaining sensitive fields
+          basic_scrubber.perform(pii_fields.filter { |f| f.sensitive? && !f.scrubbed? })
+
+          pii_attrs = pii_fields.
+            filter(&:scrubbed?).
+            to_h { |f| [f.name, f.scrubbed_value] }
+
+          # include non-nullable attrs to ensure upsert works in postgres
+          required_attrs = non_nullable_cols.to_h do |column|
```
eanders, line 59: Why the `to_h` here?  should it be after the `end`?

eanders (approved):
@ttoomey, thanks for the great tests.  I dropped a few questions inline, but generally, this looks good to me.

app/models/pii/scrubber/replacement_pii.rb
ttoomey: This is called by the FakeScrubber. Scrubbers are stacked together so if you scrub with `custom_scrubber: :fake` it fills in what it can with fake values using this method. The other fields are handled either by the dob scrubber or the basic scrubber (catch-all).

So this method doesn't have to be comprehensive but we could add in more fields if we find it useful. I didn't include emails, phones, and addresses in `fake_value` since I want to make sure they are easily distinguishable from real values.

app/models/pii/scrubber/scrub_model_pii.rb
ttoomey: this is a variant of array.to_h that takes a block returning [key,value]
```
array.to_h { |item| [key, value] }
=> { key => value}
```

Which accomplishes the same thing as `array.map { |i| [k,v] }.to_h` but without an intermediate array.

app/models/pii/scrubber/dob_scrubber.rb
ttoomey: good spot, thanks!

ttoomey:
@gigxz lmk if you have any questions on this PR

app/models/grda_warehouse/tasks/scrub_pii/scrub_all_pii_task.rb
gigxz: Ok, I see yes this right for scrubbing *all* PII. 

> It is an interesting question if we were scrubbing PII for a single client, would we also scrub the associated User records?

I would say no, because there are other Clients/records that are referencing that User. We don't want to lose data about other clients.

app/models/grda_warehouse/tasks/scrub_pii/scrub_client_pii_task.rb
gigxz: As discussed last time, this is a foundation for scrubbing PII, we will need to revisit the specifics if/when we introduce any features for scrubbing individual client data. As discussed there are several potential use-cases that may have different requirements (eg "anonymize me, but keep my data intact for continued service" vs "remove my data entirely, but keep my data intact for reporting" ).

gigxz (approved):
Looks great @ttoomey ✨

app/models/pii/scrubber/scrub_model_pii.rb
eanders: Amazing.  Thanks.

PR #4913, opened 2024-11-10
Client Supplemental data bugfix

ttoomey:
## _Merging this PR_

## Description

Looks like the previous bugfix wasn't correctly applied. Fix and add tests to prove it. We were still seeing the behavior of too many source clients shown on a the client data source

## Type of change
Bug fix

## testing
visit [QA client page](https://qa-warehouse.openpath.host/hmis_supplemental/data_sets/1/client_data_sets/19574). Should only one set of values for a given client in the data source 

### DIFF
--- a/drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
+++ b/drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
@@ -39,9 +39,10 @@ module HmisSupplemental
     protected
 
     def source_clients
+      # order is important here, source_visible_to appears to clobber the data source condition
       @client.source_clients.
-        where(data_source_id: @data_set.data_source_id).
         source_visible_to(current_user).
+        where(data_source_id: @data_set.data_source_id).
         order(:id)
     end
 
@@ -55,20 +56,6 @@ module HmisSupplemental
 
     def load_authorized_data_set
       data_set_scope.find(params[:data_set_id])
-    end
-
-    def client_groups
-      # note: order is important here, source_visible_to appears to clobber the data source condition
-      clients = @client.source_clients.
-        source_visible_to(current_user).
-        where(data_source_id: @data_set.data_source_id).
-        order(:id)
-      clients.map do |client|
-        {
-          title: client.data_source.name,
-          values: @data_set.field_values.for_owner(@client).index_by(&:field_key),
-        }
-      end
     end
 
     def data_set_scope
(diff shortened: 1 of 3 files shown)


drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
```diff
@@ -39,9 +39,10 @@ def show
     protected
 
     def source_clients
+      # order is important here, source_visible_to appears to clobber the data source condition
       @client.source_clients.
-        where(data_source_id: @data_set.data_source_id).
         source_visible_to(current_user).
+        where(data_source_id: @data_set.data_source_id).
```
ttoomey, line 45: this was the bug

drivers/hmis_supplemental/app/controllers/hmis_supplemental/client_data_sets_controller.rb
```diff
@@ -57,20 +58,6 @@ def load_authorized_data_set
       data_set_scope.find(params[:data_set_id])
     end
 
-    def client_groups
```
ttoomey, line 60: unused code I forgot to remove 🤦

drivers/hmis_supplemental/spec/requests/client_data_set_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
...
+        expect(assigns(:groups).map { |h| h[:title] }).to include(
+          a_string_including(project1.name),
+          a_string_including(project2.name),
+        )
+        expect(assigns(:groups).size).to eq(2)
+      end
+    end
+  end
+
+  context 'without access' do
```
eanders: Isn't the check we need that when there is supplemental data on both the source data sources (`data_source` and `data_source_other`) but the user is only given access to one of them, that the `data_source_other` data isn't visible?  I might be missing something here, but I'd expect something like
```
expect(assigns(:groups).map { |h| h[:title] }).to_not include(
           a_string_including(project_other.name),
         )
```

ttoomey: The bug addressed in this PR was that the page was showing for all visible source clients (not just the one client that matching the data set's data source). So the bug wasn't really an issue related to data set visibility and it should be covered by the example above.

> Isn't the check we need that when there is supplemental data on both the source data sources (data_source and data_source_other)

So if you only have access to one data set, you would only see the enrollments for the source client in that data set? I'll adjust the test for that, probably a good idea

eanders (approved):
Thanks @ttoomey, looks good.

PR #4915, opened 2024-11-11
ADR for consideration

ttoomey:
## _Merging this PR_

## Description
Proposal to adopt ADRs. Perhaps this is too heavy for  us? Does it seem like it's worth trying?

## Type of change
Documentation / process change


docs/adr/README.md
```diff
@@ -0,0 +1,73 @@
...
+Our ADR process began in 2024. Not all historical decisions may be documented as ADRs, but we maintain ADRs for all significant architectural decisions going forward.
+
+## Repository Structure
+
+```
+docs/
+  adr/
+    README.md             <- This file
+    0001-first-decision.md
+    0002-second-decision.md
```
eanders, line 62: I propose we keep the ADRs in a subdirectory so we can easily find the template and readme.
Additionally, depending on how long these hang around before being ratified, or how many we have, it might make sense to include `proposed` and `superseded` directories.

```suggestion
    accepted/0001-first-decision.md
    accepted/0002-second-decision.md
    proposed/0003-third-decision.md
    superseded/0000-no-longer-active.md
```

I also wonder if there is a sequential numbering scheme is going to be hard eventually.  Probably good to keep them in chronological order, but if we end up with dozens it might be hard to find one about a particular topic.  Perhaps we can cross that bridge when we get there.

docs/adr/README.md
```diff
@@ -0,0 +1,73 @@
...
+```
+
+## Contributing
+
+1. Copy `template.md` to create your new ADR
+2. Fill in the template
+3. Create a pull request
+4. Seek review from team
+5. Update based on feedback
+6. Merge once approved
```
eanders, line 75: Should we offer some guidance about the timing of an ADR and the associated work?  I think the expectation would be the ADR would be completed first as part of the planning process, but work isn't always that sequential, so maybe calling out that the ADR should be completed before or as a part of any code adjustments.

eanders (approved):
Thanks @ttoomey.  

This feels like a relatively light way to increase our documentation of important decisions.  

I'd like to get feedback from @dtgreiner and @martha as well, I'll add them as reviewers.

docs/adr/README.md
martha: We could add it to the code review checklist?
` - [ ] I have updated the documentation and/or added an ADR (or not applicable)`

docs/adr/README.md
martha: My 2c would be to keep a flat directory structure. My previous team's approach was to just keep all ADRs in the same directory, updating the "Status" section with "Superseded" (or "Aging Poorly" or "Needs Thought", etc.) when needed. 

One of the great things about ADRs is that they are so light touch to maintain, because they document a decision at a point in time, rather than promising to document the current state of the system comprehensively; and I think a flatter directory structure helps keep the maintenance burden lighter.

martha (approved):
Oh, this is great, we used ADRs in a codebase I previously worked on and I really liked them as a documentation method. Thanks Theron!

docs/adr/README.md
ttoomey: Thanks for your insight @martha! Discussed with Elliot yesterday and it sounds like a flat structure will work for now, we can revisit as needed

docs/adr/README.md
ttoomey: > I have updated the documentation and/or added an ADR (or not applicable)

I like the idea of calling out ADRs in the PR but not sure if we should encourage people to update them outside of a formal decision making process. I guess I would expect that ADRs are relatively static once they are accepted (aside from correcting typos or something). If we change our minds on something, we would create a new ADR that supersedes the previous one. Maybe that's too heavy for us?

Perhaps:
```
```

gigxz (approved):

dtgreiner (approved):

PR #4922, opened 2024-11-12
Add missing timestamps

eanders:
## _Merging this PR_

## Description
* Adds `DateCreated` and `DateUpdated` values to new clients added through authoritative data sources in the warehouse
* Migration to find any client who doesn't have a `DateCreated` and set `DateCreated` and optionally `DateUpdated` based on the associated warehouse client record.

## Type of change
* Bug fix
* 


app/controllers/clients_controller.rb
```diff
@@ -30,6 +30,10 @@ class ClientsController < ApplicationController
 
   def create
     clean_params = client_create_params
+    # Enforce DateCreated and DateUpdated
+    clean_params[:DateCreated] = Time.current
+    clean_params[:DateUpdated] = Time.current
```
gigxz, line 35: These should be getting set automatically from `save` below anyway, right?

db/warehouse/migrate/20241112181349_set_missing_timestamps.rb
```diff
@@ -0,0 +1,39 @@
+class SetMissingTimestamps < ActiveRecord::Migration[7.0]
+  def up
+    # Destination Clients
+    import_batch = []
```
gigxz: This should be defined inside the `find_in_batches` loop? With the current setup it looks like clients will be re-imported repeatedly with each batch.

gigxz (approved):
Logic looks good to me

app/controllers/clients_controller.rb
eanders: That was the root of the issue (they aren't saved automatically on the warehouse side).

db/warehouse/migrate/20241112181349_set_missing_timestamps.rb
eanders: Good catch, thank you!

PR #4924, opened 2024-11-12
Add created_by and updated_by col to CustomAssessments

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6904
This PR:
- Adds a migration to the `CustomAssessments` table, adding the created_by_user_id and updated_by_user_id cols, and updates the model with the new relationships
- Modifies the `SubmitAssessment` mutation to update both values, and updates the spec test
- Modifies the schema `Assessment` object to use the values, rather than loading from versions
- Adds a one-time script to populate these columns from the versions table. On my local db, which has 1300 versioned assessments (most of which only have 1 version), this takes about 12 seconds.

## Type of change


db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
```diff
@@ -0,0 +1,14 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class AddCreatedByToCustomAssessment < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
+      # These refer to the users table in the app db (not warehouse), so fk relationship is not made explicitly
+      add_reference :CustomAssessments, :created_by_user, null: true, type: :integer
+      add_reference :CustomAssessments, :updated_by_user, null: true, type: :integer
```
ttoomey: are you sure you want an integer here? Usually pks are `bigint` (which is the default)

```suggestion
      add_reference :CustomAssessments, :created_by_user, null: true
      add_reference :CustomAssessments, :updated_by_user, null: true
```

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
```diff
@@ -136,6 +136,16 @@ def enrollment
...
     end
 
+    def load_last_user_from_versions(object)
+      # Don't need to load from versions since assessments have updated_by as an attribute on the record
+      object.updated_by
+    end
+
+    def load_created_by_user_from_versions(object)
+      # Don't need to load from versions since assessments have created_by as an attribute on the record
+      object.created_by
```
ttoomey: should use a data loader to avoid n+1?

```suggestion
    def load_last_user_from_versions(object)
      # Don't need to load from versions since assessments have updated_by as an attribute on the record
      load_ar_association(object, :updated_by)
    end

    def load_created_by_user_from_versions(object)
      # Don't need to load from versions since assessments have created_by as an attribute on the record
      load_ar_association(object, :created_by)
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,25 @@
+desc 'One time data migration to populate created_by and updated_by on Custom Assessments'
+# rails driver:hmis:migrate_assessments_20241111
+task migrate_assessments_20241111: [:environment] do
+  assessments = Hmis::Hud::CustomAssessment.joins(:versions).distinct
```
ttoomey: The join to versions is to exclude assessments that have no versions? I'd be tempted to just iterate through all assessments since most of them should have versions anyway. My intuition is that this join could be a bit expensive since versions is a huge table on prod.

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,25 @@
...
+  total_records = assessments.count
+
+  assessments.find_each.with_index(1) do |custom_assessment, index|
+    versions = custom_assessment.versions.order(created_at: :desc)
+    update_version = versions.first
+    create_version = versions.find_by(event: "create")
```
ttoomey: Looks like there are three queries here. I suspect preloading the versions would be more efficient as it should result in many fewer queries. However it will potentially load more version records than are strictly needed

```ruby
  assessments = Hmis::Hud::CustomAssessment.preload(:versions)
  total_records = assessments.count
  assessments.preload(:versions).find_each.with_index(1) do |custom_assessment, index|
    versions = custom_assessment.versions.to_a # use array methods since this is preloaded
    next if versions.blank?
    
    update_version = versions.max_by(&:created_at)
    create_version = versions.detect { |v| v.event == 'create' }
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,25 @@
...
+
+    if !create_version || !update_version
```
ttoomey: Looks like this will skip unless both create and update_version are present; is that right? Perhaps we should still update updated_by_user_id even if created_by_user_id is null?

```suggestion
    if create_version.nil? && update_version.nil?
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,25 @@
...
+      puts "Check versions of #{custom_assessment.id}. It has versions but they look incorrect. Skipping."
+      next
+    end
+
+    # use `update_column` to bypass paper trail and timestamp updates
+    custom_assessment.update_column(:created_by_user_id, create_version.clean_true_user_id || create_version.clean_user_id) if create_version
+    custom_assessment.update_column(:updated_by_user_id, update_version.clean_true_user_id || update_version.clean_user_id) if update_version
```
ttoomey: looks like there are two updates for the same record here; I think update_columns lets you do this with one update
```suggestion
    attrs = {}
    attrs[:created_by_user_id] = create_version.clean_true_user_id || create_version*.clean_user_id if create_version
    attrs[:updated_by_user_id = update_version.clean_true_user_id || update_version.clean_user_id if update_version
    custom_assessment.update_columns(**attrs) if attrs.values.any?
```

further optimization could be had using active-record-import to upsert the values in batches but that would be more complex and maybe isn't needed for a one-time task.

drivers/hmis/app/graphql/mutations/submit_assessment.rb
```diff
@@ -53,6 +53,9 @@ def resolve(input:, assessment_lock_version: nil)
 
       has_already_been_submitted = assessment.persisted? && !assessment.in_progress?
 
+      assessment.updated_by = current_user
+      assessment.created_by = current_user unless has_already_been_submitted
```
ttoomey: In the script to backfill the users from the versions we are using the "true" user if it exists. I believe that means that we use the original user if there was an impersonation. But we don't use the impersonating user here.

Does that inconsistency matter?

ttoomey (approved):
Looks good, thanks!

I left some suggestions around performance optimizations. Maybe optimization isn't needed. Sounds like you were processing around 100 records/sec locally. On a larger record set that could become significant (15min for 100k records). But if we are only dealing with a couple thousand records, it's probably fine as-is

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
martha: I'm not sure, but I used `integer` because the pk of the `users` table is an `integer`:
https://github.com/greenriver/hmis-warehouse/blob/release-142/db/structure.sql#L3275

Maybe we intend to update this to `bigint`? Or maybe `users` is a table that we don't expect to ever grow that big?

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
martha: Thanks, yes!

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: I thought that the majority of assessments won't have versions, since they are migrated-in. However, I agree with your below suggestion to preload versions and once we do that I don't think we need to separately do the join here. Thanks!

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: I'm not 100% confident, but that was my intended logic. My thinking was just that if we have versions for this object but no version marked "create," then some funny business is going on and we should investigate.

Now that I'm looking at it again, `update_version` will never be nil since we stop processing if versions is an empty arr. So we can just log something useful if `create_version` is nil and continue processing otherwise.

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: Cool, thanks!

drivers/hmis/app/graphql/mutations/submit_assessment.rb
martha: Hm thanks for pointing that out, yes, I should fix that. My understanding is that we want the new `created_by` and `updated_by` attributes to match whatever the current behavior is on `load_last_user_from_versions` and `load_created_by_user_from_versions`, which appears to load the true user. https://github.com/greenriver/hmis-warehouse/blob/release-142/drivers/hmis/app/graphql/sources/paper_trail_versions.rb#L26-L39

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
ttoomey: Thanks, my understanding is that we should avoid using integers for references. I think we use them now because many of the tables were migrated from sql server. I'd recommend using bigints and we can plan to transition the pks on those legacy tables in the future. Does that sound right @eanders ?

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
ttoomey: I wonder about assessments that we created as part of the onboarding data migration. I could be wrong but those might not have a 'create' event since I think we disable papertrail during the import. The assessments could have a follow-up update though

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
ttoomey: oh I'm noticing there might be an issue here. If the last version didn't have a user_id, we would could miss the user who last updated it
```ruby
    update_version = versions.filter do |version| 
      (version.clean_true_user_id || version.clean_user_id)
    end.max_by(&:created_at)
```

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
eanders: If it allows `bigints` and the app works, then yes, the plan is to convert everything, so setting up now as `bigint` is one fewer data piece to migrate.

martha: Ok, perfect, thank you both!

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: Oh, that sounds right to me. Ok, I'll remove the unneeded log.

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: Yes good point. I'm proposing to introduce this change earlier in the script so we only consider versions with a user for both updated_by and created_by, if that sounds good to you.

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -14,6 +14,10 @@ def hmis_user
     Hmis::Hud::User.from_user(current_user)
   end
 
+  def true_hmis_user
+    Hmis::Hud::User.from_user(true_user)
+  end
```
ttoomey: maybe check if impersonating to avoid an extra query

```suggestion
  def impersonating?
    current_user != true_user
  end
  
  def true_hmis_user
    impersonating? ? Hmis::Hud::User.from_user(true_user) : hmis_user
  end
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,75 @@
...
+  puts "Now going back through to update #{total_records} assessments with their `created_by_hud_user`"
+
+  assessments.find_each.with_index(1) do |assessment, index|
+    create_version = assessment.versions.detect { |v| v.event == 'create' && (v.clean_true_user_id.present? || v.clean_user_id.present?) }
+    if create_version
+      hud_user = app_user_ids_to_hud_users[create_version.clean_true_user_id || create_version.clean_user_id]
+      hud_user.reload
+
+      assessment.update_column!(:created_by_hud_user_id, hud_user.id)
+    end
```
ttoomey: I'm probably missing something but rather than look at the version again, can't we just use assessment.created_by_user_id to map to hud_user_id?
```suggestion
    hud_user_id = app_user_ids_to_hud_users[assessment.created_by_hud_user]&.id
    assessment.update_column!(:created_by_hud_user_id, hud_user_id)
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,75 @@
...
+
+  # explicitly do not support multiple data sources
+  raise 'unexpectedly found multiple HMIS data sources' if data_source_ids.size > 1
+
+  # find (and if needed, save) a HUD user for each app user we encountered
+  app_user_ids_to_hud_users = {}
+  user_ids.each do |user_id|
+    user = Hmis::User.find(user_id)
+    user.hmis_data_source_id = data_source_ids.first
+    hud_user = Hmis::Hud::User.from_user(user)
```
ttoomey: seems okay to just save these here since there shouldn't too many user records; then we don't need the `Hmis::Hud::User.import(users_to_import)` logic below.

```suggestion
    hud_user = Hmis::Hud::User.from_user(user)
    hud_user.save!
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,75 @@
...
+
+  assessments = Hmis::Hud::CustomAssessment.where(created_by_hud_user: nil).preload(:versions)
+  total_records = assessments.count
+  puts "Now going back through to update #{total_records} assessments with their `created_by_hud_user`"
+
+  assessments.find_each.with_index(1) do |assessment, index|
+    create_version = assessment.versions.detect { |v| v.event == 'create' && (v.clean_true_user_id.present? || v.clean_user_id.present?) }
+    if create_version
+      hud_user = app_user_ids_to_hud_users[create_version.clean_true_user_id || create_version.clean_user_id]
+      hud_user.reload
```
ttoomey: I could be missing something but I think this will raise an exception if the hud user record wasn't persisted

ttoomey (commented):
Thanks Martha, it's looking good! I left a few suggestions; we can connect tomorrow if you have any questions

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: Yes! But wait, I think it has to be assessment.created_by_user, rather than assessment.created_by_hud_user. Sorry for the confusing naming.

This poked at a nascent thought I was having yesterday, which is that I think it'd be better if this script were idempotent, which it isn't right now thanks to line 59: `assessments = Hmis::Hud::CustomAssessment.where(created_by_hud_user: nil)`. I'll make an update.

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
martha: It turns out calling `save!` is not needed since `Hmis::Hud::User.from_user(user)` calls `first_or_create`, not `first_or_initialize`.

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
```diff
@@ -0,0 +1,17 @@
...
+
+class AddCreatedByToCustomAssessment < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
+      # These refer to the users table in the app db (not warehouse), so fk relationship is not made explicitly
+      add_reference :CustomAssessments, :created_by_user, null: true
+      add_reference :CustomAssessments, :updated_by_user, null: true
+      # This refers to the User table in the warehouse db
+      add_reference :CustomAssessments, :created_by_hud_user, foreign_key: { to_table: :User }, null: true
+      # no need to add updated_by_hud_user, since UserID column exists
```
gigxz: I may have missed something in the PR discussion, but why do we need `created_by_hud_user`?

TBH I would be more inclined to eventually drop the UserID column entirely, and just generate it on HMIS CSV exports (once we start exporting custom assessments...).  I don't see the benefit in keeping both references to `users` and `User` tables. (_Maybe_ there is an argument for having that reference for migrated-in data, to store a reference to some associated "user" that doesn't exist in the application).

martha: Ah thanks for raising this @gigxz. Theron and I discussed it on a huddle but I forgot to update the PR description to describe what we talked about.

My understanding is that the `users` table is currently difficult to import into and join within Superset, since it's in the app db, but the `User` table is already there and easily joinable, since it's in the warehouse db. The long term vision is that we don't want to keep the 2 separate dbs, so everything will be easily imported into Superset, and we would then drop the `created_by_hud_user` column in favor of `created_by_user`. 

Based on your comment, I wonder if `created_by_hud_user` is the one that should prevail, in which case `created_by_user` could be removed now?

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -14,6 +14,14 @@ def hmis_user
     Hmis::Hud::User.from_user(current_user)
   end
 
+  def impersonating?
+    current_user != true_user
+  end
```
gigxz: https://github.com/greenriver/hmis-warehouse/blob/997ee79b0c3659f5c503239009e3e71ef6887a5b/drivers/hmis/app/controllers/hmis/base_controller.rb#L74-L76
maybe this should be moved to a shared concern instead of defined twice

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -14,6 +14,14 @@ def hmis_user
...
+  
+  def true_hmis_user
+    impersonating? ? Hmis::Hud::User.from_user(true_user) : hmis_user
+  end
```
gigxz: I'm confused by this one. We already have references to `true_hmis_user` which I thought was defined by the 
[pretender](https://github.com/ankane/pretender) gem as the application user id (of the user who is impersonating someone else). See other places we refer to this like [in the base controller](https://github.com/greenriver/hmis-warehouse/blob/997ee79b0c3659f5c503239009e3e71ef6887a5b/drivers/hmis/app/controllers/hmis/base_controller.rb#L74-L76). Now here we are defining the same variable for use in queries/mutations, but this time it's a HUD `User` instead of an application user (`users`).

Am I grasping this? Maybe let's do a huddle on this.

drivers/hmis/app/graphql/mutations/submit_assessment.rb
```diff
@@ -87,7 +87,16 @@ def resolve(input:, assessment_lock_version: nil)
       assessment.assign_attributes(
         user_id: hmis_user.user_id,
         assessment_date: assessment.form_processor.find_assessment_date_from_values,
+        # Use true_user rather than current_user in case of impersonation, matching load_last_user_from_versions.
+        # See comments on CustomAssessments about user, created_by, created_by_hud_user, and updated_by
+        updated_by: true_user,
```
gigxz: Again I probably missed some convo here @martha but I'm not sure if this is the desired behavior. Previously we intentionally set `UserID` based on the "current user" even if they were not the true user, because of some feedback we had gotten about data correction workflows. The true user will always be available in the audit history. It seems odd to me that the UserID and updated_by_user can be different, now.

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
```diff
@@ -136,6 +136,16 @@ def enrollment
       load_ar_association(object, :enrollment)
     end
 
+    def load_last_user_from_versions(object)
```
gigxz: I think it should work to just say `def user` to override what's in the concern?

drivers/hmis/app/graphql/types/hmis_schema/assessment.rb
```diff
@@ -136,6 +136,16 @@ def enrollment
...
+      # Don't need to load from versions since assessments have updated_by as an attribute on the record
+      load_ar_association(object, :updated_by)
+    end
+
+    def load_created_by_user_from_versions(object)
```
gigxz: Same can we just say `def created_by`

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
martha: Probably for clarity should have been named `true_hud_user` (but maybe pending further discussion can be completely removed)

drivers/hmis/app/graphql/mutations/submit_assessment.rb
martha: Makes sense. The thought process on my part was probably something like,
1. add `created_by` and `updated_by`. Per the ticket, these refer to the application user and not the HUD user. Use the "true" user because this is being populated from paper trail in the first place, so make it match that behavior going forward.
2. add `created_by_hud_user` based on changing spec, per my comment [here](https://github.com/greenriver/hmis-warehouse/pull/4924/files#r1852242341). But try to make it match the `created_by` and `updated_by`, so invent some kind of concept of a "true hud user". I think this is the leap that you're saying doesn't really make sense.

Let's discuss more when we call later, just trying to articulate the history :)

Possible the solution we will land on could be something like
- remove `created_by` and `updated_by`, continue loading these from paper trail, even though they were the original asks from the ticket they're no longer really needed(?)
- for `created_by_hud_user`, populate this from paper trail, but going forward, populate it similarly to `UserID` (aka don't worry about the "true" user and just use the HUD user.)

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
gigxz: Thanks @martha that makes sense. I don't follow on removing `created_by_user` (since we're saying we want to move to those eventually when the db issue is resolved?) but lets discuss.

@ttoomey  I'd like to understand the superset limitation a little better-- is there an issue I can look at? I wonder if we can find a path forward for joining from warehouse db tables to the `users` table at least. There are other warehouse records that reference the users table, that will need to join with `users` to display information in superset. For example:
* staff assignment (reference users directly) - I can see this being asked for in Superset to track caseload over time
* files (reference users directly) - I don't see a clear use-case for reporting on files, but I'm sure there will be one eventually
* user access controls - reporting on access control changes (may or may not want to end up in superset, but tagging https://github.com/open-path/Green-River/issues/404)

drivers/hmis/app/graphql/mutations/submit_assessment.rb
gigxz: Ok I see, good point. Since we're currently loading the user from paper trail, we [do resolve](https://github.com/greenriver/hmis-warehouse/blob/75ca057f8f77da8e9754485fdaf8b1751eda28b0/drivers/hmis/app/graphql/sources/paper_trail_versions.rb#L38) the true user in the api right now.

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -27,7 +27,16 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
...
   belongs_to :client, **hmis_relation(:PersonalID, 'Client')
+
+  # UserID - references the HUD user as specified by the HUD spec (User table in warehouse db)
   belongs_to :user, **hmis_relation(:UserID, 'User'), inverse_of: :assessments, optional: true
+  # [created|updated]_by_user_id - only for HMIS data, references application user (users table in app db)
+  belongs_to :created_by, foreign_key: :created_by_user_id, class_name: 'Hmis::User', optional: true
+  belongs_to :updated_by, foreign_key: :updated_by_user_id, class_name: 'Hmis::User', optional: true
+  # created_by_hud_user - references the HUD user. This is for ease of reporting in Superset,
+  # but eventually when everything is in the same DB, this should be removed in favor of created_by
+  belongs_to :created_by_hud_user, foreign_key: :created_by_hud_user_id, class_name: 'Hmis::Hud::User', optional: true, autosave: true
```
gigxz: I don't think we need autosave here? We are not making changes to the User record

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,70 @@
+desc 'One time data migration to populate created_by and updated_by on Custom Assessments'
+# rails driver:hmis:migrate_assessments_20241111
+task migrate_assessments_20241111: [:environment] do
```
gigxz: This could use a more descriptive name, `migrate_assessment_user_references_` or something? We already have an assessment migration job so I could see it being confusing.

db/warehouse/migrate/20241111212106_add_created_by_to_custom_assessment.rb
martha: Notes from my huddle with @gigxz, to discuss at dev standup Monday:

Background info
- Right now we’re not displaying the UserID anywhere in the application.
- We used to, but now we load the user from paper trail. 
- Adding these cols was intended to just lighten the expensive load of fetching it from paper trail.
- UserID is required by HUD, but not used in reports.- HUD doesn’t even specify if it’s the user who created or last updated. Right now we update UserID anytime we update the record.

Ideal future state
- For most things that want to reference user, use application users. 
  - We own/control that table and it’s not subject to disruptions if someone’s email address changes
- What’s the future of the HUD User table? 
  - Maybe (extreme case) it even goes away, and is just generated at export when we make a CSV? (Though this is a big departure from our posture which is that the db schema matches hud exactly)

2 possible paths for this PR
- Path 1: If we can’t join with the users table. 
    - Then just point it at UserID (w/out worrying about impersonation) and don’t update when the assessment updates
    - Still add the user reference columns for future use
    - In the place where we call the HUD user the “hmis” user, rename that across the whole file
- Path 2 (preferred): How hard would it be to pursue a way to get the users table into superset, w/out merging it into the warehouse database?
    - Get rid of created_by_hud_user
    - Keep created_by_user and updated_by_user
    - Make some decision about impersonation. It probably doesn’t matter that much to customers either way

Side note about impersonation
- Use case: Someone who likes impersonating while doing data correction, so that the “real” person who created the assessment is still the user associated with it

martha: Decision from dev standup 11/25
- add `created_by_hud_user_id` and `updated_by_hud_user_id` both pointing at the HUD user table by PK, and not by data source relation
- don't add any columns related to application user

gigxz: thanks @martha !

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -10,7 +10,7 @@ def true_user
     context[:true_user]
   end
 
-  def hmis_user
+  def hud_user
```
gigxz, line 13: thanks for updating this 🙏

drivers/hmis/app/graphql/mutations/submit_assessment.rb
```diff
@@ -85,9 +85,11 @@ def resolve(input:, assessment_lock_version: nil)
         hud_values: input.hud_values,
       )
       assessment.assign_attributes(
-        user_id: hmis_user.user_id,
+        user_id: hud_user.user_id,
         assessment_date: assessment.form_processor.find_assessment_date_from_values,
+        updated_by_hud_user: hud_user, # see comments on CustomAssessment
       )
+      assessment.assign_attributes(created_by_hud_user: hud_user) unless has_already_been_submitted
```
gigxz: It looks like the approach you took is:
* when assessment is saved as in-progress, DON'T set created_by or last_updated_by
* when assessment is saved against as in-progress, still DON'T set created_by or last_updated_by
* when assessment is submitted, DO set both created_by and last_updated_by
* when assessment is re-submitted, DO set last_updated_by

Is my understanding right?

I would ask to adjust the logic so that created_by and last_updated_by are set when the assessment is first created and each subsequent time that it is saved, regardless of whether the save originates from a "submit" or a "save-as-WIP" action. IMO these fields should record actions on the record (like DateCreated and DateUpdated) so I don't think it makes sense to leave them off for WIP assessments. It also seems odd to have a new assessment be created but not have a created_by user reference (While it's in the WIP state).

So, desired logic:
* when assessment is first saved as in-progress, DO set created_by and last_updated_by
* when assessment is saved against as in-progress, DO set last_updated_by
* when assessment is first submitted, DO last_updated_by. DO set created_by IF it's the first time this assessment is created
* when assessment is re-submitted, DO set last_updated_by


If you adjust the [new_with_deafaults](https://github.com/greenriver/hmis-warehouse/blob/90489af9cb475c86439203e12a2bbcded8272980/drivers/hmis/app/models/hmis/hud/custom_assessment.rb#L180) helper directly to set `created_by_hud_user: user.id`, that should initialize it correctly for both SaveAssessment and SubmitAssessment mutations which use a common find_or_create_assessment for assessment init.

drivers/hmis/app/graphql/mutations/update_relationship_to_ho_h.rb
```diff
@@ -22,7 +22,7 @@ def resolve(enrollment_id:, enrollment_lock_version: nil, relationship_to_ho_h:,
       if relationship_to_ho_h == 1
         # Lock to avoid duplicate request collisions. with_lock also starts a transaction.
         enrollment.project.with_lock do
-          hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user_id: hmis_user.user_id)
+          hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user: hud_user.user_id)
```
gigxz:
```suggestion
          hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user_id: hud_user.user_id)
```

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,52 @@
...
+# rails driver:hmis:migrate_assessment_user_references_20241111
+task migrate_assessment_user_references_20241111: [:environment] do
+  assessments = Hmis::Hud::CustomAssessment.joins(:user).preload(:versions)
+  total_records = assessments.count
+  updated_count = 0
+
+  hud_user_local_cache = {} # Cache to store users in memory, reducing db hits
+
+  def find_or_create_hud_user(user_id, data_source, hud_user_local_cache)
+    hud_user_local_cache[user_id] ||= begin
```
gigxz, line 12: This assumes all users are in the same data source (since it's not part of the hash key), which is fine, because that is true right now and we don't expect to re-run this migration.

Since we're already making that assumption, we can pull out the data_source to the top of this task `data_source = GrdaWarehouse::DataSource.hmis.sole` and  limit the Custom Assessment scope to that data source (there shouldn't be any assessments outside of that, but just to be safe).

That would save us the extra database hits we are doing in the loop when we call `assessment.data_source` to look up the data source. (An alternative is to just throw on a preload for data source, that would be fine too).

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      if update_version
+        user_id = update_version.clean_true_user_id || update_version.clean_user_id
+        hud_user = find_or_create_hud_user(user_id, assessment.data_source, hud_user_local_cache)
+        attrs[:updated_by_hud_user] = hud_user
+      end
+
+      # use `update_columns` to bypass paper trail and timestamp updates
+      assessment.update_columns(**attrs)
```
gigxz, line 45: It won't work to pass a relation to `update_columns` like this, you need to pass the actual column name. So above, `attrs[:updated_by_hud_user_id] = hud_user.id` instead of `attrs[:updated_by_hud_user] = hud_user`

gigxz (commented):
Looks good, comments below!

drivers/hmis/app/graphql/mutations/submit_assessment.rb
martha: Thanks Gig! Good point about the logic on save WIP, I updated that and added a spec test.

drivers/hmis/app/graphql/mutations/submit_household_assessments.rb
```diff
@@ -84,7 +84,7 @@ def resolve(submissions:, confirmed:, validate_only: false)
 
       # Run form processor on each assessment to create/update related records
       assessments.each do |assessment|
-        assessment.assign_attributes(user_id: hmis_user.user_id)
+        assessment.assign_attributes(user_id: hud_user.user_id)
```
gigxz: Is there a reason we're not updating `updated_by_hud_user` when a user submits household assessment(s) from the "Complete Exit/Entry" tab? I think it's probably uncommon that it's a different user from the user who saved them, but it could occur. Elsewhere we seem to be consistent with where we update the `UserID` column, but not here.

drivers/hmis/lib/tasks/one_time_migration_20241111.rake
```diff
@@ -0,0 +1,53 @@
+desc 'One time data migration to populate created_by and updated_by on Custom Assessments'
+# rails driver:hmis:migrate_assessment_user_references_20241111
```
gigxz, line 2: Looks good to me! I successfully ran it locally. It didn't update the majority of my assessments because they are generated so don't have versions, but it did update the ones that were created manually via HMIS, which is expected.

gigxz (approved):
Sorry for the delay @martha - looks great! I tested individual/household submission locally, as well as running the script.

I had one outstanding question about household assessment submission, but LGTM once that is answered/addressed

drivers/hmis/app/graphql/mutations/submit_assessment.rb
gigxz: Ok, thanks!

gigxz:
Oh it looks like the latest build has a test failure, is that related? https://github.com/greenriver/hmis-warehouse/actions/runs/12039482165/job/33567378362

drivers/hmis/app/graphql/mutations/submit_household_assessments.rb
martha: Thanks, good point, I added this.

martha:
Merge conflict fixed. I believe the test failures in [`drivers/hmis/spec/requests/hmis/project_enrollment_search_spec.rb`](https://github.com/greenriver/hmis-warehouse/actions/runs/12123017907/job/33797592660?pr=4924) and [`spec/models/health/qualifying_activity_spec.rb`](https://github.com/greenriver/hmis-warehouse/actions/runs/12123017907/job/33797592934?pr=4924) are unrelated, as I'm also seeing them on release-143.

gigxz (approved):

PR #4933, opened 2024-11-15
fix performance issue on project names in report filters

ttoomey:
## _Merging this PR_

## Description
Performance fix on report filters page; tested on core-demographics page

## Type of change
Bug fix

## Testing instructions
Check that the core demographics report filters are loading correctly. In particular the "Projects" filter

### DIFF
--- a/app/models/concerns/filter/control_sections.rb
+++ b/app/models/concerns/filter/control_sections.rb
@@ -124,37 +124,37 @@ module
             id: 'coc_codes',
             label: 'CoC Codes',
             short_label: 'CoC',
-            value: @filter.chosen_coc_codes,
+            value: -> {  @filter.chosen_coc_codes },
           )
         end
         section.add_control(
           id: 'funding_sources',
-          value: @filter.funder_names,
+          value: -> { @filter.funder_names },
         )
         section.add_control(
           id: 'funding_other_sources',
-          value: @filter.funder_others,
+          value: -> { @filter.funder_others },
         )
         section.add_control(
           id: 'data_sources',
-          value: @filter.data_source_names,
+          value: -> { @filter.data_source_names },
         )
         section.add_control(
           id: 'organizations',
-          value: @filter.organization_names,
+          value: -> { @filter.organization_names },
         )
         section.add_control(
           id: 'projects',
-          value: @filter.project_names,
+          value: -> { @filter.project_names },
         )
         section.add_control(
           id: 'project_groups',
-          value: @filter.project_groups,
+          value: -> { @filter.project_groups },
         )
         if GrdaWarehouse::Cohort.viewable_by(@filter.user).exists?
           section.add_control(
             id: 'cohorts',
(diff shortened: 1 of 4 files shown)


app/models/concerns/filter/control_sections.rb
```diff
@@ -124,37 +124,37 @@ def control_sections
...
         )
         section.add_control(
           id: 'organizations',
-          value: @filter.organization_names,
+          value: -> { @filter.organization_names },
         )
         section.add_control(
           id: 'projects',
-          value: @filter.project_names,
+          value: -> { @filter.project_names },
```
ttoomey, line 148: This was the significant change but generally using lambdas for lazy evals seems to help the "Edit" report modal performance

app/models/filters/filter_base.rb
```diff
@@ -1522,6 +1522,8 @@ def organization_names
     end
 
     def project_names(ids = project_ids)
+      return [] if ids.blank?
```
ttoomey, line 1525: short circuit since if ids are empty, we won't return anything. Avoids unnecessarily calling project_name which was slowing down the index on the demographic report

drivers/core_demographics_report/spec/requests/core_demogphics_report_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+      get core_demographics_report_warehouse_reports_core_index_path
+      expect(response).to be_successful
+      expect(assigns(:report)).to be_an_instance_of(CoreDemographicsReport::Core)
+
+      # check queries once the cache is warm
+      expect do
+        get core_demographics_report_warehouse_reports_core_index_path
+      end.to make_database_queries(
+        count: 10..40,
+        matching: /\A(?!.*SELECT "translations".* FROM "translations")/, # exclude translations
```
ttoomey, line 43: There are so many db queries involving translations; it's probably slowing down our tests

app/models/concerns/filter/control_sections.rb
eanders: Thanks!  That seems like a solid idea and will probably be beneficial across the app.

drivers/core_demographics_report/spec/requests/core_demogphics_report_spec.rb
eanders: I believe, though maybe we should confirm, that in production those get loaded once and then pulled from redis by key, I wonder if we can revisit that for the test suite as well (maybe just using the memory cache.)

eanders (approved):
Thanks @ttoomey, this looks like it may significantly help performance across the app, especially in installations with many projects, and thanks for extending our test suite at the same time.

ttoomey:
Thanks Elliot, just to confirm should I merge this into r141?

eanders:
@ttoomey, yes, let's put this in 141 since the other changes are there as well.

PR #4935, opened 2024-11-15
Demographic Summary Report Index/Detail Counts

dtgreiner:
## _Merging this PR_

## Description

More closely align index and detail calculations counts in the demographic summary report.

Also include a way to run the pages inline in development for easier use.

## Type of change

### DIFF
--- a/drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
@@ -24,11 +24,11 @@ module
       }
     end
 
-    def age_range
-      {
-        'child' => (0..17),
-        'adult' => (18..110),
-      }
+    def age_range_for(category)
+      return (0..17) if category == 'child'
+      return (18..110) if category == 'adult'
+
+      raise "unknown category: #{category}"
     end
 
     def age_detail_hash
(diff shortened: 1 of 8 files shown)


drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
```diff
@@ -24,6 +24,13 @@ def age_categories
       }
     end
 
+    def age_range
```
eanders: can we rename this `adult_and_child_age_ranges`?  There are other uses of `age_range` further down in this file within a block `age_categories.each do |age_range, age_title|` which can make the name of this method confusing.

eanders: alternatively something like:
```
def age_range_for(category)
  return (0..17) if category == 'child'
  return (18..110) if category == 'adult'

  raise "unknown category: #{category}"
end
```

might be more clear what's going on.
I don't _think_ we're ever looping over those age ranges, so returning the range is probably better anyway.

eanders (approved):
Thanks @dtgreiner this appears to enforce consistency well.  If you can make one minor adjustment to the way `age_ranges` works/is named, I think we're good.

drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
dtgreiner: update this method to `age_range_for` and returning the range instead of the hash.

dtgreiner:
@eanders This should be good to go for another check. Thanks!

eanders (approved):
Thanks!

PR #4937, opened 2024-11-16
Security update for yarn package

eanders:
## _Merging this PR_

## Description
Addresses https://github.com/greenriver/hmis-warehouse/security/dependabot/106

## Type of change
* Dependency update

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,8 @@
   },
   "resolutions": {
     "minimatch": "^3.0.5",
-    "ws": "^8.17.1"
+    "ws": "^8.17.1",
+    "cross-spawn": "^7.0.5"
   },
   "scripts": {
     "lint": "eslint app/assets/javascripts",
--- a/package.json
+++ b/package.json
@@ -37,8 +37,7 @@
   },
   "resolutions": {
     "minimatch": "^3.0.5",
-    "ws": "^8.17.1",
-    "cross-spawn": "^7.0.5"
+    "ws": "^8.17.1"
   },
   "scripts": {
     "lint": "eslint app/assets/javascripts",
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,45 +1,45 @@
 module.exports = {
-    'env': {
-        'browser': false,
-        'es6': true
-    },
-    'extends': 'eslint:recommended',
-    'globals': {
-        "Atomics": "readonly",
-        "SharedArrayBuffer": "readonly",
-        "window": true,
-        "alert": true,
-        "console": true,
-        "document": true,
-        "localStorage": true,
-        "setInterval": true,
-        "setTimeout": true,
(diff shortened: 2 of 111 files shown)


package.json
```diff
@@ -37,7 +37,8 @@
   },
   "resolutions": {
     "minimatch": "^3.0.5",
-    "ws": "^8.17.1"
+    "ws": "^8.17.1",
+    "cross-spawn": "^7.0.5"
```
gigxz: So we only have cross-spawn because `eslint` depends on it:
```bash
eb10d31a667a:/app$ yarn why cross-spawn
yarn why v1.22.22
[1/4] Why do we have the module "cross-spawn"...?
[2/4] Initialising dependency graph...
[3/4] Finding dependency...
[4/4] Calculating file sizes...
=> Found "cross-spawn@7.0.3"
info Reasons this module exists
   - "eslint" depends on it
   - Hoisted from "eslint#cross-spawn"
info Disk size without dependencies: "68KB"
info Disk size with unique dependencies: "132KB"
info Disk size with transitive dependencies: "192KB"
info Number of shared dependencies: 3
Done in 0.15s.
```

So if you were to run the below, it would resolve the vulnerability by upgrading eslint from v8 [which is eol](https://eslint.org/version-support/) to v9:
```bash
yarn upgrade eslint # upgrades to latest, which is v9. The latest v8 eslint@v8.57.1 still uses the bad cross-spawn
yarn why cross-spawn # see its at 7.0.6
yarn audit # vulnerability is gone
```
There are several breaking changes with v8=>9 for eslint though, so we would want to review those before taking that approach: https://eslint.org/docs/latest/use/migrate-to-9.0.0
I think that is worth investigating- if we leave eslint at v8, it is inevitable there will be more vulnerability alerts popping up because it uses older packages.

When I run `yarn lint` I see lots of errors, so it seems like we're not actually enforcing the linter, or maybe we're running something different on CI?


Anyway, point is: I see using `resolutions` as a last-ditch solution when you **can't** upgrade the package via it's dependencies. (Say for example if the latest version of eslint _still_ used corss-spawn 7.0.4, then we would have no choice but to upgrade it via the resolution, which is risky since you don't know if that upgrade could cause some issues with how eslint is using it).

The yarn explanation is here: https://classic.yarnpkg.com/en/docs/selective-version-resolutions/#toc-why-would-you-want-to-do-this

gigxz: At the same time, this doesn't feel like a risky change since it's only affecting the linter which we may or may not use. Just wanted to explain that reasoning. I would be fine with merging this and doing a follow-up to upgrade eslint to v9 and make sure we're enforcing it in the way we want to

eanders:
Closing in favor of https://github.com/greenriver/hmis-warehouse/pull/4942, something weird happened here.

PR #4939, opened 2024-11-18
OP Analytic views

eanders:
## _Merging this PR_

## Description
* Removes old `bi_*` views
* Adds views as a "contract" for OP Analytics consumption under the `analytics` schema in the warehouse database.

## Type of change
* New feature

### DIFF
--- a/db/views/analytics_clients_v01.sql
+++ b/db/views/analytics_clients_v01.sql
@@ -1,0 +1,1 @@
+SELECT * FROM "Client" WHERE "DateDeleted" is NULL
--- a/app/models/bi/view_maintainer.rb
+++ b/app/models/bi/view_maintainer.rb
@@ -7,7 +7,10 @@
 module Bi
   class ViewMaintainer
     include ArelHelper
-
+    TodoOrDie('Remove this maintainer class and comment out any related migrations or code that references it after release-144 has been merged into stable', by: '2025-01-10')
+    # Currently included in
+    # db/warehouse/migrate/20241030133448_remove_current_living_situation_verified_by_limit_1.rb
+    # db/warehouse/migrate/20241030133450_remove_current_living_situation_verified_by_limit_3.rb
     HUD_CSV_VERSION = '2020'.freeze
     NAMESPACE = 'bi'.freeze
     PG_ROLE = 'bi'.freeze
--- a/db/views/analytics_affiliations_v01.sql
+++ b/db/views/analytics_affiliations_v01.sql
@@ -1,0 +1,1 @@
+SELECT * FROM "Affiliation" WHERE "DateDeleted" is NULL
--- a/db/views/analytics_assessment_questions_v01.sql
+++ b/db/views/analytics_assessment_questions_v01.sql
@@ -1,0 +1,1 @@
+SELECT * FROM "AssessmentQuestions" WHERE "DateDeleted" is NULL
--- a/db/views/analytics_assessment_results_v01.sql
+++ b/db/views/analytics_assessment_results_v01.sql
@@ -1,0 +1,1 @@
+SELECT * FROM "AssessmentResults" WHERE "DateDeleted" is NULL
--- a/db/views/analytics_assessments_v01.sql
+++ b/db/views/analytics_assessments_v01.sql
@@ -1,0 +1,1 @@
+SELECT * FROM "Assessment" WHERE "DateDeleted" is NULL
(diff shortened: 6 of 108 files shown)


db/views/analytics_clients_v01.sql
```diff
@@ -0,0 +1 @@
+SELECT * FROM "Client" WHERE "DateDeleted" is NULL
```
ttoomey, line 1: Should we be removing app-internal fields from these models? Maybe just export the CamelCase fields? 
```sql
-- Hmis::Hud::Client.column_names.grep(/\A[A-Z]/ ).sort.map { |f| "\"#{f}\""}.join
(",\n")
SELECT id, "AdditionalRaceEthnicity",
"AfghanistanOEF",
-- ...
"WorldWarII",
"YearEnteredService",
"YearSeparated" FROM "Client" WHERE "DateDeleted" is NULL
```


Maybe even fix the casing in our fields?
```sql
-- `Hmis::Hud::Client.column_names.grep(/\A[A-Z]/ ).sort.map { |f| "\"#{f}\" as #{f.underscore}"}.join(",")`
SELECT id,  
"AdditionalRaceEthnicity" as additional_race_ethnicity,
"AfghanistanOEF" as afghanistan_oef,
-- ...
"WorldWarII" as world_war_ii,
"YearEnteredService" as year_entered_service,
"YearSeparated" as year_separated
 FROM "Client" WHERE "DateDeleted" is NULL
```

ttoomey (commented):
This looks great, thanks!

Couple things:
* I assume there was a script that autogenerated this migrations. If so, can you include it in the pr?
* What's the criteria for selecting which tables to export. Is it just anything under ` GrdaWarehouse::Hud::*` 
* Should we take the opportunity to fix the casing on hud tables to be easier to work with in postgres? Maybe a question for the data team on how much a lift it would be form them, maybe not worth the effort
* Should we try to exclude internal app-columns from the views?

db/views/analytics_clients_v01.sql
eanders: I want to discuss, probably with @Jacobihi, if we should limit the columns to those in the HMIS CSVs (essentially.)  At the moment we are using the PG copier to pull the whole table and then do the refactoring in DBT (or at least that's my understanding.)

eanders:
> * I assume there was a script that autogenerated this migrations. If so, can you include it in the pr?
Not much a script, but I did gather up a list of class names and loop over them to generate `rails generate scenic:view 'analytics.Export'` and `"SELECT * FROM #{klass.quoted_table_name}"`  It might be worth writing something more official, but I don't have anything at the moment.

> * What's the criteria for selecting which tables to export. Is it just anything under ` GrdaWarehouse::Hud::*`
These are chosen based on the table list we are currently pulling in https://github.com/greenriver/superset-sync/blob/main/docker/pgs3copy/src/pgs3copy/conf/default.yaml

> * Should we take the opportunity to fix the casing on hud tables to be easier to work with in postgres? Maybe a question for the data team on how much a lift it would be form them, maybe not worth the effort
I think the idea is to keep these views as light as possible since we're already doing all of the heaving lifting in DBT, but if we want to move that logic to the app instead, we can.  The goal here was to pivot the source of the data to a compartmentalized area that we can make more structured eventually, if necessary.

> * Should we try to exclude internal app-columns from the views?
I think there is some utility in this, but also a benefit in having more of the data available for OP Analytics since we don't always know what will be useful upstream (downstream?).

ttoomey (approved):

db/views/analytics_clients_v01.sql
Jacobihi: That is what we're doing currently. I think handling all the transformation in one place is preferable so I think renaming the columns downstream is fine, because at this juncture it would be more work to re-map newly named fields (because we have a pattern of decoding the hud fields and call the column `CamelCase` -> `camel_case_code`

If there's data as shouldn't be exposed then I certainly agree we don't need to copy it, so as long as all the internal ids that make the joins work are present the HMIS csvs can strip to only the CamelCase fields + id.

Jacobihi (approved):
From my standpoint this looks good.

db/views/analytics_clients_v01.sql
eanders: Chatted with @Jacobihi and we're going to leave these as is for V1.  This replicates the same data footprint we currently have but moves it to the analytics schema, which should mean we can pivot to a simpler mechanism to ingest it.  Once that work has been completed we can revisit if we want to decrease the available footprint by adjusting the views.

PR #4942, opened 2024-11-19
Eslint upgrade

eanders:
## _Merging this PR_

## Description
Addresses https://github.com/greenriver/hmis-warehouse/security/dependabot/106

## Type of change
* Dependency update


yarn.lock
```diff
@@ -819,40 +826,41 @@ escape-string-regexp@^4.0.0:
...
+eslint-visitor-keys@^3.4.1, eslint-visitor-keys@^3.4.3:
+  version "3.4.3"
+  resolved "https://registry.yarnpkg.com/eslint-visitor-keys/-/eslint-visitor-keys-3.4.3.tgz#0cd72fe8550e3c2eae156a96a4dddcd1c8ac5800"
+  integrity sha512-wpc+LXeiyiisxPlEkUzU6svyS1frIO3Mgxj1fdy7Pm8Ygzguax2N3Fa/D/ag1WqbOprdI+uY6wMUl8/a2G+iag==
 
 eslint@^8.44.0:
-  version "8.44.0"
-  resolved "https://registry.yarnpkg.com/eslint/-/eslint-8.44.0.tgz#51246e3889b259bbcd1d7d736a0c10add4f0e500"
-  integrity sha512-0wpHoUbDUHgNCyvFB5aXLiQVfK9B0at6gUvzy83k4kAsQ/u769TQDX6iKC+aO4upIHO9WSaA3QoXYQDHbNwf1A==
+  version "8.57.1"
```
gigxz: This is upgrading to the latest v8 which is still EOL. v9 is the current version. Seems OK just pointing out it's not the latest version.
I did review the notes at https://eslint.org/docs/latest/use/migrate-to-9.0.0 and I think we'd be fine to upgrade to v9

gigxz (approved):

eanders:
@gigxz this is ready for a re-review with eslint 9.x and a new config file

eslint.config.js
```diff
@@ -0,0 +1,31 @@
+module.exports = [
```
gigxz, line 3: Is this replacing the existing `.eslintrc.js`, if so lets remove that file?

eanders: It is.  I was only leaving the other one in case we needed to go back, but I suppose, git, so it'll be there.

gigxz: Oh okay. It should've still worked with the old file, but they just made this format be the default.
Why was `'extends': 'eslint:recommended',` removed? That adds all the recommended lint rules

eanders: I couldn't get that working with the new format, or didn't see a way to do that.  Do you want to give it a try?

gigxz: Ok, added per https://eslint.org/docs/latest/use/migrate-to-9.0.0#-eslintrecommended-and-eslintall-no-longer-accepted-in-flat-config but a little different since we can't use `import` here with our setup.

I confirmed the difference in configs with and without the recommended rules by printing config for a file:
```
yarn eslint --print-config "assets/javascripts/cable.js"
```

eanders: Thank you!

PR #4944, opened 2024-11-19
Talent Captive Portal

dtgreiner:
## _Merging this PR_

## Description

Updates for the Talent Captive Portal:
- Add Site Level config for number of courses the user is required to complete and expose this on the TalentLMS admin page
- Send user to captive portal if multiple LMS configs require training (if only one config requires training, send the user directly to that portal)
- Link to each course (if the user has a login for  the corresponding portal) on the captive portal
- Updates to phrasings on the captive portal

## Type of change
New feature

- [X] If adding a new endpoint / exposing data in a new way, I have:
  - [X] ensured the API can't leak data from other data sources
  - [X] ensured this does not introduce N+1s
  - [X] ensured permissions and visibility checks are performed in the right places


app/controllers/admin/talentlms_controller.rb
```diff
@@ -43,6 +48,20 @@ def destroy
       @config = config_scope.find(params[:id].to_i)
     end
 
+    private def set_site_config
+      @site_configs = site_config_scope.where(id: 1).first_or_create
```
eanders, line 52: Did you intend to leave `where(id: 1)` here?

app/views/admin/talentlms/index.haml
```diff
@@ -6,63 +6,97 @@
...
-        %th Default
-        %th
-    %tbody
-      - @configs.each do |config|
-        - config.courses.each do |course|
+.h2 
+  Configuration
+.well
+  = simple_form_for @site_configs, url: update_site_config_admin_talentlms_path(@site_configs), method: :post, remote: true do |f|
+    = f.input :number_lms_courses_required, as: :select_two, collection: { "All": -1, "One of Any": 1 },  input_html: {class: :jRemoteSubmit}, label: 'TalentLMS Courses Completion Requirement'
```
eanders: Can we move the value for `collection` into the config model?  We should call something like `@site_configs.class.available_completion_counts` (that's probably not quite right).  Having the options in the view makes them harder to enforce and to track down later.

eanders (commented):
@dtgreiner this looks good.  Can you give me a tour (and maybe add some screenshots of the changes to the associated issue?)

app/controllers/admin/talentlms_controller.rb
dtgreiner: This is how it's being referenced elsewhere in the app ([example](https://github.com/greenriver/hmis-warehouse/blob/dg/talent-captive-portal_6669/app/controllers/admin/configs_controller.rb#L30), [example](https://github.com/greenriver/hmis-warehouse/blob/dg/talent-captive-portal_6669/app/controllers/admin/de_duplication_controller.rb#L42), [example](https://github.com/greenriver/hmis-warehouse/blob/dg/talent-captive-portal_6669/spec/models/grda_warehouse/system_cohorts/system_cohort_spec.rb#L31)) so I matched to stay consistent. It does feel like removing the `.where(id: 1)` from each of these locations and going with `GrdaWarehouse::Config.first_or_create` would be better. Happy to adjust it in all of these spots if you feel it is worth doing.

app/views/admin/talentlms/index.haml
dtgreiner: Will add this in the next push to the PR

dtgreiner:
### Updated TalentLMS Admin Page
Screenshot 2024-11-20 at 8 45 37 AM
Screenshot 2024-11-20 at 8 45 42 AM

### Updated Captive Portal
Screenshot 2024-11-20 at 8 46 15 AM

app/controllers/admin/talentlms_controller.rb
eanders: Oh, I mis-read that, I was thinking `GrdaWarehouse::Config` was some sort of other Talent related config.  Huh, that's weird, but I think the issue is that you have to have something between the class and `.first_or_create`.  Let's leave this as is.  Thanks.

app/views/admin/talentlms/index.haml
dtgreiner: This has been included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/4944/commits/8921e9abdb3166418ba55263e027c235c8a0cc52).

eanders (approved):
Thanks @dtgreiner !

eanders:
@dtgreiner can you fix the conflicts and then squash merge?  Thanks!

PR #4950, opened 2024-11-20
Fix HMIS theme validation bug and add tests

gigxz:
## _Merging this PR_

## Description
1. Fix validation error that would come up when there was a Warehouse (non-HMIS) theme existing alongside a default HMIS theme.
2. Add tests for `hmis_theme_for_origin` even though it's not used yet ( https://github.com/open-path/Green-River/issues/6774)

How to test:
- create a warehouse theme with no hmis value
- create another theme with an hmis_value but no hmis_origin. previously it was invalid, now it is valid

## Type of change
Bug fix

- n/a If adding a new endpoint / exposing data in a new way, I have:

### DIFF
--- a/app/models/grda_warehouse/theme.rb
+++ b/app/models/grda_warehouse/theme.rb
@@ -36,7 +36,10 @@ module GrdaWarehouse
     belongs_to :remote_credential, class_name: 'GrdaWarehouse::RemoteCredentials::S3', optional: true
 
     # Max 1 theme per HMIS origin
-    validates :hmis_origin, uniqueness: { scope: [:client] }, if: :hmis_theme?
+    validates_uniqueness_of :hmis_origin,
+                            scope: :client,
+                            if: :hmis_theme?, # only run validation if this is an HMIS theme
+                            conditions: -> { where.not(hmis_value: [nil, '']) } # only validate uniqueness against other HMIS themes
 
     def set_theme_defaults
       # Fetch files from S3 if available or use defaults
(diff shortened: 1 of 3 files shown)


app/models/grda_warehouse/theme.rb
```diff
@@ -36,7 +36,10 @@ class Theme < GrdaWarehouseBase
     belongs_to :remote_credential, class_name: 'GrdaWarehouse::RemoteCredentials::S3', optional: true
 
     # Max 1 theme per HMIS origin
-    validates :hmis_origin, uniqueness: { scope: [:client] }, if: :hmis_theme?
+    validates_uniqueness_of :hmis_origin,
+                            scope: :client,
+                            if: :hmis_theme?, # only run validation if this is an HMIS theme
```
martha, line 41: If line 41 wasn't working, could we just remove it to avoid confusion? Since `hmis_theme?` checks `.present?` which is essentially the same thing you are checking here on line 42. But maybe I'm misunderstanding

martha: Also, maybe this is just me, but potentially it would be helpful to add a comment here pointing readers of the code to `GrdaWarehouse::DataSource.hmis.<first>.hmis` (as you pointed me to in Slack). Prior to your hint, I found it a bit hard to discover by searching in the code what this col should actually be set to.

martha (approved):
Looks good, and works for me locally. Cool unit tests ✨

app/models/grda_warehouse/theme.rb
gigxz: It is working, it just didn't work in the way I thought it would. The `if` condition determines _whether_ the validation should run for a given record. (It does not say anything about which records it should compare with for uniqeuness). So we still want it because we only need to validate this for hmis themes. https://api.rubyonrails.org/v7.2.2/classes/ActiveRecord/Validations/ClassMethods.html#method-i-validates_uniqueness_of

eanders (approved):
Thank you @gigxz !

PR #4955, opened 2024-11-22
avoid calling roi_authorized on window check

ttoomey:
performance  tweak, avoid calling roi_authorized


ttoomey:
@eanders looks like this didn't make it into 141 before deploy. Should this get retargeted to 142?

eanders:
I wanted to discuss before merging.  Fine to retarget 142.  Performance is acceptable on 141 for now.

ttoomey:
@eanders made another attempt at improving readability of the window access logic. I also broke the `resource_permissions` method up to make it more approachable.  LMK if it seems better.

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -84,4 +64,45 @@ def roi_authorized?
...
+  # a set of permissions the user has for either the project or the client which would grant them access to this client
+  memoize def resource_permissions
+    results = Set.new
+    add_legacy_data_source_permissions(results)
+    add_project_based_permissions(results)
+    add_direct_client_permissions(results)
+    results
+  end
+
+  # windowed data sources are a deprecated legacy client data sharing mechanic, replaced by ACLs
```
eanders:
```suggestion
  # Window data sources are a deprecated legacy client data sharing mechanic, replaced by a system Collection in when using Access Controls-based permissions
```

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -84,4 +64,45 @@ def roi_authorized?
...
+      joins(:clients).
+      merge(GrdaWarehouse::Hud::Client.where(id: client_id)).
+      distinct.
+      pluck(:id)
+    enrolled_project_ids.each do |project_id|
+      results.merge(context.project_role_permissions(project_id))
+    end
+  end
+
+  # permissions the user has directly on the client (for destination clients with no enrollments)
```
eanders:
```suggestion
  # permissions the user has directly on the client (for destination clients with no enrollments in authoritative data sources)
```

eanders (approved):
@ttoomey this feels much cleaner and more clear to me.  Thank you!
Dropped a couple notes on comments that I think will help us communicate what's going on the to ourselves and others in the future.

PR #4958, opened 2024-11-24
6956 hmis chronic uses ch enrollment

ttoomey:
## Merging this PR

## Description
Replace `client.hud_chronic?` with `enrollment.ch_enrollment` to determine chronic homless status in the HMIS. This should mean:
* better performance
* uses a different implementation of chronic calculation. It's unclear what the functional difference is other than the ch_enrollment seems like newer code.

[GH Issue](https://github.com/open-path/Green-River/issues/6956)

## Testing
The chronic status on the HMIS enrollment dashboard should reflect our new logic. Note that the `can_view_hud_chronic_status` permission is required to access. 

The status is updated by a job every night so there is a lag in the status appearing.

New logic:
* Client is __not__ CH is they have an open PH project (3, 9, 10, 13) with a move in date that has occurred on or before the current date
* Client is CH if they have any open enrollments at project types (0, 1, 4, 8) that meet the criteria of CH (DisablingCondition:1, MonthsHomelessPastThreeYears: 1 year+, TimesHomelessPastThreeYears: 4+)

## Type of change
Bug fix, Performance improvement

### DIFF
--- a/app/models/grda_warehouse/ch_enrollment.rb
+++ b/app/models/grda_warehouse/ch_enrollment.rb
@@ -4,7 +4,13 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
-# persist the HUD calculation for chronic homelessness at the enrollment level
+# ChEnrolment persist the HUD calculation for chronic homelessness at the enrollment level
+#
+# Details:
+# - Rows are calculated on a schedule via GrdaWarehouse::ChEnrollment.maintain!
+# - Each ChEnrollment represents the status of it's associated enrollment based on self-reported data
+# - The status is calculated as-of the current date when the ch_enrollment is generated
+# - The data that is stored will represent CH at Project Start for non-homeless projects (which is actually the same thing as CH at a Point In Time since you don't acquire homeless time being in a non-homeless project. BUT, in homeless projects, you do acquire days for every day you are there. So joining ChEnrollment is the equivalent of CH at a Point In Time for each enrollment.
 module GrdaWarehouse
   class ChEnrollment < GrdaWarehouseBase
     include ArelHelper
@@ -53,7 +59,8 @@ module GrdaWarehouse
               chronically_homeless_at_entry: chronically_homeless_at_start?(enrollment, date: date),
             }
           end
-          import(batch)
+          result = import(batch)
+          raise "failed to import ChEnrollments: #{result.inspect}" if result.failed_instances.present?
         end
     end
 
@@ -72,13 +79,14 @@ module GrdaWarehouse
             chronically_homeless_at_entry: chronically_homeless_at_start?(enrollment, date: date),
           }
         end
-        import!(
+        result = import!(
           batch,
           on_duplicate_key_update: {
             conflict_target: [:id],
             columns: [:processed_as, :chronically_homeless_at_entry],
           },
         )
+        raise "failed to import ChEnrollments: #{result.inspect}" if result.failed_instances.present?
       end
     end
 
--- a/app/models/grda_warehouse/ch_enrollment.rb
+++ b/app/models/grda_warehouse/ch_enrollment.rb
@@ -4,6 +4,7 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# persist the HUD calculation for chronic homelessness at the enrollment level
 module GrdaWarehouse
   class ChEnrollment < GrdaWarehouseBase
     include ArelHelper
--- a/drivers/hmis/app/graphql/types/hmis_schema/client.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/client.rb
@@ -340,12 +340,14 @@ module Types
     def hud_chronic
       return unless current_permission?(permission: :can_view_hud_chronic_status, entity: object)
 
+      # We are intentionally not checking enrollment visibility. The can_view_hud_chronic_status
+      # permission grants visibility into HUD Chronic status across all the client's enrollments
       open_enrollments = object.enrollments.open_excluding_wip.preload(:ch_enrollment, :project)
 
       # Check if there's an open enrollment at a residential project with a move in date that has occurred
       today = Date.current
       client_is_housed = open_enrollments.any? do |enrollment|
-        enrollment.project.project_type.in?(HudUtility2024.residential_project_type_ids) && (enrollment.move_in_date&.<= today)
+        enrollment.project.project_type.in?(HudUtility2024.permanent_housing_project_types) && (enrollment.move_in_date&.<= today)
       end
       return false if client_is_housed
 
--- a/drivers/hmis/app/graphql/types/hmis_schema/client.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/client.rb
@@ -339,9 +339,7 @@ module Types
     def hud_chronic
       return unless current_permission?(permission: :can_view_hud_chronic_status, entity: object)
 
-      # causes n+1 queries
-      enrollments = object.enrollments.where(data_source_id: current_user.hmis_data_source_id)
-      !!object.destination_client&.as_warehouse&.hud_chronic?(scope: enrollments)
+      load_ar_association(object, :enrollments_with_chronic_homelessness).any?
     end
 
     def audit_history(filters: nil)
--- a/drivers/hmis/app/graphql/types/hmis_schema/client.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/client.rb
@@ -82,7 +82,7 @@ module Types
     field :contact_points, [HmisSchema::ClientContactPoint], null: false
     field :phone_numbers, [HmisSchema::ClientContactPoint], null: false
     field :email_addresses, [HmisSchema::ClientContactPoint], null: false
-    field :hud_chronic, Boolean, null: true
+    field :hud_chronic, Boolean, null: true, description: 'Meets the definition for HUD chronically homeless as of today (time of API request)'
 
     field :active_enrollment, Types::HmisSchema::Enrollment, null: true do
       argument :project_id, ID, required: true
@@ -336,10 +336,23 @@ module Types
       load_ar_association(object, :alerts, scope: Hmis::ClientAlert.active).sort_by(&:created_at).reverse
     end
 
+    # not optimized for batch queries, causes n+1 queries
     def hud_chronic
       return unless current_permission?(permission: :can_view_hud_chronic_status, entity: object)
 
-      load_ar_association(object, :enrollments_with_chronic_homelessness).any?
+      open_enrollments = object.enrollments.open_excluding_wip.preload(:ch_enrollment, :project)
+
+      # Check if there's an open enrollment at a residential project with a move in date that has occurred
+      today = Date.current
+      client_is_housed = open_enrollments.any? do |enrollment|
+        enrollment.project.project_type.in?(HudUtility2024.residential_project_type_ids) && (enrollment.move_in_date&.<= today)
+      end
+      return false if client_is_housed
+
+      # Check chronic status at open enrollments for homeless project types
+      open_enrollments.any? do |enrollment|
+        enrollment.project.project_type.in?(HudUtility2024.chronic_project_types) && enrollment.ch_enrollment&.chronically_homeless?
+      end
     end
 
     def audit_history(filters: nil)
--- a/drivers/hmis/app/models/hmis/hud/client.rb
+++ b/drivers/hmis/app/models/hmis/hud/client.rb
@@ -54,6 +54,10 @@ class Hmis::Hud::Client < Hmis::Hud::Base
   has_many :hmis_services, through: :enrollments # All services (HUD and Custom)
   has_many :services, through: :enrollments # HUD Services only
   has_many :custom_services, through: :enrollments # Custom Services only
+  has_many :enrollments_with_chronic_homelessness,
+           -> { where(chronically_homeless_at_entry: true) },
+           through: :enrollments,
+           source: :ch_enrollment
 
   # History of merges into this client
   has_many :merge_histories, class_name: 'Hmis::ClientMergeHistory', primary_key: :id, foreign_key: :retained_client_id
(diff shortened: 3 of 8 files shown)


gigxz:
can ignore test failure from https://github.com/open-path/Green-River/issues/6856

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -339,9 +339,7 @@ def alerts
     def hud_chronic
       return unless current_permission?(permission: :can_view_hud_chronic_status, entity: object)
 
-      # causes n+1 queries
-      enrollments = object.enrollments.where(data_source_id: current_user.hmis_data_source_id)
-      !!object.destination_client&.as_warehouse&.hud_chronic?(scope: enrollments)
+      load_ar_association(object, :enrollments_with_chronic_homelessness).any?
```
gigxz: The frontend tooltip states `Whether this client is considered chronically homeless, as of today. Follows the HUD definition for Chronic at PIT.`

I think to make that continue to be true, we need to:
1) Exclude Exited enrollments. If I understand the ch_enrollment code correctly, this will return `true` if the client has _Ever_ been considered CH at the time of exit. That is not in line with the tooltip description.
2) Return null (or false) if client has no open enrollments? Looking at ch_enrollment, it seems like we are unable to calculate whether the client is chronic "today" if they have no open enrollments. But maybe I'm getting the HUD definition of chronic wrong. @eanders do you know whether a client with no open enrollments should have a HUD chronicity status? It seems like we wouldn't have enough information to calculate that.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -54,6 +54,10 @@ class Hmis::Hud::Client < Hmis::Hud::Base
   has_many :hmis_services, through: :enrollments # All services (HUD and Custom)
   has_many :services, through: :enrollments # HUD Services only
   has_many :custom_services, through: :enrollments # Custom Services only
+  has_many :enrollments_with_chronic_homelessness,
+           -> { where(chronically_homeless_at_entry: true) },
+           through: :enrollments,
+           source: :ch_enrollment
```
gigxz: Is there a description somewhere about what `chronically_homeless_at_entry: true` means?

From looking at the code, I think it means "if the enrollment is exited, client was deemed chronically homeless on the day of exit. If the enrollment is open, client is deemed chronically homeless today." I'm interpreting that from [the update_existing code](https://github.com/greenriver/hmis-warehouse/blob/46e2ed52a4ece023b13be213bd2dcec65b9f81d9/app/models/grda_warehouse/ch_enrollment.rb#L66-L71) and how it chooses date. But the name of the field itself is `chronically_homeless_at_entry`.. I don't see entry date being used in the actual calculation.

gigxz: Maybe we can give the relation a more descriptive name, like `enrollments_with_chronic_today_or_at_exit`  🫣 or, filter to open enrollments only and call this `open_enrollments_hud_chronic_at_point_in_time` or `open_enrollments_with_hud_chronic_today` etc

eanders: @gigxz this calculation should match up with **CH at Project Start* from the glossary for any project where the project type is not 0, 1, 4, or 8.  For those projects (homeless projects) I believe it should align with **CH at a Point In Time**.  The column name is definitely confusing.

gigxz: Can you point me to the code where that braching happens? I'm seeing that the function `chronically_homeless_at_start_steps` receives the same `date` to work from, regardless of project type. I see that the project type affects some internal calculations, like how `total_months_homeless` is calculated, but ultimately we are still calculating whether they were considered chronic at the point-in-time (which is [either Today or Exit Date](https://github.com/greenriver/hmis-warehouse/blob/46e2ed52a4ece023b13be213bd2dcec65b9f81d9/app/models/grda_warehouse/ch_enrollment.rb#L66)).

gigxz: maybe I'm missing something but these two comments appear to be conflicting, and I think the first one may be inaccurate..
```
    # Accept an optional date which will be used for extending the homeless
    # range if the project is a homeless project
```
...
```
    # Optionally accepts a date to use for "CH at a point-in-time" calculation.
```

drivers/hmis/app/graphql/types/hmis_schema/client.rb
ttoomey: Thanks! I should have thought to check the tooltip to get a better understanding of the intention. 
* I'll limit to chronic status on currently open enrollments (open_excluding_wip)
* I'll add that to the field description in graphql
* I'll assume that this should be false if there are no open enrollments

drivers/hmis/app/models/hmis/hud/client.rb
ttoomey: fwiw I'll probably just remove the `enrollments_with_chronic_homelessness` scope and handle it in the hud_chronic field resolver instead

gigxz: Ok. We should still wait to hear from @eanders because if it _does_ work in the way he described (CH at Entry for certain project types, CH at P-I-T for others) then this implementation does NOT match up with the tooltip displayed in the frontend, which says that it's displaying chronicity statues for point-in-time of today.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: Thanks, I _think_ that's correct, based on my reading of the chronicity calculations.

drivers/hmis/app/models/hmis/hud/client.rb
eanders: So, the calculation is:
1. Limited to the associated enrollment
2. aware of the current date (but can be fed a different date)
3. cached/calculated daily - meaning the data that is stored will represent **CH at Project Start** for non-homeless projects (which is actually the same thing as **CH at a Point In Time** since you don't acquire homeless time being in a non-homeless project.  BUT, in homeless projects, you do acquire days for every day you are there.  So joining `ChEnrollment` is the equivalent of **CH at a Point In Time** for each enrollment.  

Does that make sense?

ttoomey: Thanks for explaining, I'll capture the above in code comments. Maybe we can discuss this at dev stand up.

gigxz: Can we huddle on this?

I think I get that what you're saying matches up with the reporting glossary (quote below), but I don't fully understand:
1. how this is reflected in the code
2. how we should be communicating this new value to the user. It no longer feels accurate to present this as answering "is this client experiencing chronic homelessness as of today?".. it's more like, "were they chronic at entry at any of their open non-homeless enrollments, or are they chronic today based on homeless enrollments?"


> Households in street outreach, shelters, and safe havens ([project type] = 0, 1, 4, or 8) may also have a “point-intime” CH status designed to include time spent from [project start date] up to a [point-in-time date] as time spent experiencing homelessness. Clients in projects other than street outreach, shelters, and safe havens may only have a CH status at project start and cannot age into CH by virtue of being in any of those projects. Reporting on point intime CH for these other project types simply uses the client’s CH status at project start.



The original implementation of this field used `hud_chronic?`, which matched [the implementation of](https://github.com/greenriver/hmis-warehouse/blob/a96126ec80ec9f3e1069ce490afc74580cd0407e/app/views/clients/chronic/_chronic_calculator.haml#L20) the Chronic tab on the Warehouse Client dashboard. It was nice that those two matched because you could go over and see the explanation (and validate that they match), provided that you have the same level of access. The Chronic tab presents the data in much the same way we do in HMIS-- answering the question "is this Client considered HUD Chronic on this Date?". It's still not clear to me how the ChEnrollment answers this -- and why we don't use the same implementation on the Warehouse Chronic tab, since it appears to be presenting the same type of information.

Screenshot 2024-11-25 at 12 55 06 PM

drivers/hmis/app/graphql/types/hmis_schema/client.rb
eanders: @gigxz I believe you are correct.  `ChEnrollment` is looking at if the client was chronic at entry, or point-in-time for homeless enrollments, but if the enrollment is exited, it would not contribute to the client being "chronic."  We should exclude exited enrollments from this calculation, and maybe re-word the tooltip to `Whether this client has an open enrollment that meets the HUD definition of Chronic at Project Start or Chronic at a Point-in-Time as of today.`

eanders (approved):

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: maybe we just need to further limit the enrollment scope to `ES (1), SH (8) or SO (4) and not currently housed (other residential types)`

gigxz: discussion:

1) return false immediately if they have an open residential project with a move in date that has occurred, which indicates that they are housed
2) return false immediately if they have no open enrollments at homeless project types
3) limit the scope of enrollments to open enrollments, join to ChEnrollment and look for the `chronically_homeless_at_entry`

drivers/hmis/spec/requests/hmis/lookup_client_spec.rb
```diff
@@ -177,26 +177,6 @@
     expect(result.dig('data', 'client', 'hudChronic')).to eq(false)
   end
 
-  context 'with a client who is chronically homeless per HUD definition' do
```
ttoomey, line 180: moved to it's own spec file

app/models/grda_warehouse/ch_enrollment.rb
```diff
@@ -52,7 +59,8 @@ def self.add_new!
               chronically_homeless_at_entry: chronically_homeless_at_start?(enrollment, date: date),
             }
           end
-          import(batch)
+          result = import(batch)
+          raise "failed to import ChEnrollments: #{result.inspect}" if result.failed_instances.present?
```
ttoomey, line 63: bugfix, raise on import failure

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -336,12 +336,23 @@ def alerts
...
+      # Check if there's an open enrollment at a residential project with a move in date that has occurred
+      today = Date.current
+      client_is_housed = open_enrollments.any? do |enrollment|
+        enrollment.project.project_type.in?(HudUtility2024.residential_project_type_ids) && (enrollment.move_in_date&.<= today)
+      end
+      return false if client_is_housed
+
+      # Check chronic status at open enrollments for homeless project types
+      open_enrollments.any? do |enrollment|
+        enrollment.project.project_type.in?(HudUtility2024.chronic_project_types) && enrollment.ch_enrollment&.chronically_homeless?
```
ttoomey, line 356: maybe chronic_project_types should be homeless_project_types which includes transitional housing

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -336,12 +336,23 @@ def alerts
...
     end
 
+    # not optimized for batch queries, causes n+1 queries
     def hud_chronic
       return unless current_permission?(permission: :can_view_hud_chronic_status, entity: object)
 
-      # causes n+1 queries
-      enrollments = object.enrollments.where(data_source_id: current_user.hmis_data_source_id)
-      !!object.destination_client&.as_warehouse&.hud_chronic?(scope: enrollments)
+      open_enrollments = object.enrollments.open_excluding_wip.preload(:ch_enrollment, :project)
```
gigxz, line 345: Can you add a comment noting that we are intentionally not checking enrollment viewability here? can_view_hud_chronic_status grants permission to view HUD Chronic status across all clients enrollments

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -336,12 +336,23 @@ def alerts
...
+
+      # Check if there's an open enrollment at a residential project with a move in date that has occurred
+      today = Date.current
+      client_is_housed = open_enrollments.any? do |enrollment|
+        enrollment.project.project_type.in?(HudUtility2024.residential_project_type_ids) && (enrollment.move_in_date&.<= today)
```
gigxz: I think this is too broad. I probably incorrectly said to check for "move-in date on residential project" earlier, forgetting that ES and SO and others are considered residential in the reporting world. That always trips me up.

Move-in Date is only collected for Permanent Housing, so I believe we should limit this instead to: `project_type.in?(HudUtility2024.permanent_housing_project_types)`


for reference:
```ruby
> HudUtility2024.residential_project_type_ids
=> [0, 1, 2, 3, 4, 8, 9, 10, 13]
#.. aka
> HudUtility2024.residential_project_type_ids.map { |id| HudUtility2024.project_type(id) }
=>
["Emergency Shelter - Entry Exit",
 "Emergency Shelter - Night-by-Night",
 "Transitional Housing",
 "PH - Permanent Supportive Housing",
 "Street Outreach",
 "Safe Haven",
 "PH - Housing Only",
 "PH - Housing with Services (no disability required for entry)",
 "PH - Rapid Re-Housing"]
```

vs
```ruby
> HudUtility2024.permanent_housing_project_types
=> [3, 9, 10, 13]
> HudUtility2024.project_types_with_move_in_dates
=> [3, 9, 10, 13]
> HudUtility2024.permanent_housing_project_types.map { |id| HudUtility2024.project_type(id) }
=>
["PH - Permanent Supportive Housing",
 "PH - Housing Only",
 "PH - Housing with Services (no disability required for entry)",
 "PH - Rapid Re-Housing"]
```

drivers/hmis/app/graphql/types/hmis_schema/client.rb
gigxz: I don't think so, the reporting glossary calls out these 4 for chronic-at-point in time.
Screenshot 2024-11-25 at 7 31 12 PM

gigxz:
Looks good @ttoomey! One comment about the breadth of checking for move-in date, which I think we should fix for clarity. LGTM after that.

gigxz (approved):
💯

eanders (approved):
@ttoomey I'm good with this, but maybe we should retarget 143?  142 has been mostly QA'd at this point.

PR #4964, opened 2024-11-26
Fix schema type for editor_user_ids form field

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/6843
Follow-up to https://github.com/greenriver/hmis-warehouse/pull/4893

Fix a bug that is blocking the ability to edit forms that have editor_user_id fields. Note that this will cause any existing forms with numeric editor_user_id to fail validation. I think that is desirable (and manageable), it would be more confusing to support both strings and integers.

#### Bug reproduction (release-142)
- create a form
- using JSON editor, add editor_user_id integer values in the editor_user_ids field
- save (succeeds, passed validation)
- create a new draft
- make any form change and try to save -- it fails validation because editor_user_ids are strings. There is no way to resolve it in the UI. That happened because there was a mis-match in types. The JSON schema expected integers, but the GraphQL schema resolved strings (type `[ID]`).

## Type of change


martha (approved):
Thanks, Gig, and sorry for the bug!

gigxz:
all good, thanks for the review!

PR #4969, opened 2024-11-27
Add active date ranges to lms courses

dtgreiner:
## _Merging this PR_

## Description

Add date ranges where TalentLMS courses will be considered active. Only active courses will be included in the required training for individual users.

## Type of change

New feature


app/models/talentlms/course.rb
```diff
@@ -68,5 +85,14 @@ def completed_by?(user)
 
       Talentlms::CompletedTraining.where(login_id: login.id, course_id: id).exists?
     end
+
+    def active_date_range_for_display
+      return start_date if start_date.present? && end_date.present? && start_date == end_date
```
eanders, line 90: This should be impossible based on the validation that ensures `end_date > start date` (also if needed I think you can just use: `start_date.present? && start_date == end_date`

app/controllers/admin/talentlms_courses_controller.rb
```diff
@@ -46,6 +46,8 @@ def course_params
         :courseid,
         :months_to_expiration,
         :default,
+        :start_date,
```
eanders, line 49: I can't comment above, but the courses should probably be sorted in some way that places the active courses first in the table in date order.

eanders (changes requested):
Thanks @dtgreiner.  This looks great, and thanks for adding the validations!
Couple comments inline that I think will make this a bit better.

app/controllers/admin/talentlms_courses_controller.rb
dtgreiner: I'll plan to add some sorting to the course to address this

app/models/talentlms/course.rb
dtgreiner: This should be able to occur. The validation should only be firing if both start and end dates are present and start date is > end date. It should be valid to have it start and end on the same date, but I'll double check the logic to be sure.

eanders: Ah, that makes sense.  Disregard.

dtgreiner:
@eanders I have a different PR for this [here](https://github.com/greenriver/hmis-warehouse/pull/4973). This one had a bad merge when trying to get the latest into the branch and it was easier to recreate.

PR #4970, opened 2024-11-29
7038 user edit history

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/7038)

User edit history page
* fix misalignment with paginated values count and rows shown
* fix performance problems around n+1 and trying to display 500 items on one page
* ensure changes made to the Hmis::User are also shown
* show all changes on the user except for login events
* attempt to summarize changes to be better understood by end-users (invitation sent, etc)
* ensure credentials are clearly redacted in the version history
* displays impersonating users
* add tests

## Testing
As and admin , visit warehouse admin users page, select 'audit' user, select 'edit history' tab.

* login events or failed log in attempts should not be displayed
* change name, email, etc shows detailed changes
* user changes own password is summarized
* user requests a password reset  is summarized
* user changes password from reset form  is summarized
* invitation is sent  is summarized
* account is deactivated
* account is activated
* account is created

## Type of change
Bug fix, New feature, Code clean-up


app/views/admin/edit_histories/show.haml
```diff
@@ -3,26 +3,5 @@
 - content_for :title, title
 %h1= title
 = render 'admin/audits/user_tabs'
-- if @pagy.count.positive?
-  = render 'common/pagination_top', item_name: 'edit'
-  .table-responsive
```
ttoomey, line 8: moved to '_list.haml' partial

app/controllers/admin/edit_histories_controller.rb
```diff
@@ -10,69 +10,12 @@ class EditHistoriesController < ::ApplicationController
...
-      edits = GrPaperTrail::Version.where(
-        pt_a[:item_id].eq(@user_id).and(pt_a[:item_type].eq('User')).
-        or(pt_a[:referenced_user_id].eq(@user_id)),
-      )
-      @versions = edits.where.not(whodunnit: nil).order(created_at: :desc)
-      @pagy, @versions = pagy(@versions, items: 500)
+      @history = UserEditHistory::Versions.new(@user)
     end
 
-    def describe_changes_to(version)
```
ttoomey, line 22: Extracted code from these controllers to the UserEditHistory helpers

app/models/gr_paper_trail/version.rb
```diff
@@ -4,14 +4,94 @@
...
+      return if whodunnit.blank?
+
+      match = WHODUNNIT_IMPERSONATOR_PATTERN.match(whodunnit)
+      match[1] if match
+    end
+
+    protected
+
+    # Extracted this method from the controller. It appears to be a fallback for old versions didn't include changes
+    # * Is this still needed?
```
ttoomey, line 60: How long-ago did we start tracking object changes? Could we replace this with 'contact support to see more' or something rather than trying to compute the changes?

app/models/user_edit_history/user_version_change_summary.rb
```diff
@@ -0,0 +1,169 @@
...
+      return if match_keys && match_keys != changeset.keys.sort
+      return if match && !match.call(version, changeset)
+
+      true
+    end
+  end
+
+  # Holds all the change summary patterns
+  # Note, no need to include a condition for logins as those events are not displayed in the edit-history
+  CHANGE_PATTERNS = [
```
ttoomey, line 76: this summary set could use review

app/models/gr_paper_trail/version.rb
eanders: I could be wrong, but I think this is trying to capture what changed when the record has been deleted.

```
version = self
current = version.reify
```

It's still referencing `version.previous`  which would have a dependency of paper trail.

If we don't need this for inactive users, then I think we can probably remove it.

eanders (approved):
@ttoomey this looks awesome.  Thank you!

app/models/gr_paper_trail/version.rb
ttoomey: Looks like it's pretty old (6668b439a95).

I could see this being needed if the object changes col was missing at one point (you can use paper trail without it) and we wanted to simulate it on old records.

I don't understand why you'd want to build object changes for deleted records... the record was deleted, it's not like you'd represent that as setting all it's attributes to null.

I suspect we could remove it but I also don't really understand the intent, so maybe we should leave it to be safe 🤷

PR #4976, opened 2024-12-03
CI date fixes

ttoomey:
## _Merging this PR_

## Description
* Fixes CI failures due to assumptions in tests about the current time (started failures on 12/2/2024)
* Fixes [tz issue in e2e tests](https://github.com/open-path/Green-River/issues/6856) (probably)

## Type of change
Bug fix


spec/models/health/qualifying_activity_spec.rb
```diff
@@ -25,7 +25,7 @@
     let!(:referral_ds) { create :referral_ds }
 
     before(:each) do
-      travel_to(Date.current - 2.years) # Enrollment durations depend on time if there is no disenrollment date
```
ttoomey, line 28: Looks like the test expects this to match `Health::QualifyingActivityV2::EFFECTIVE_DATE_RANGE` but Date.current has aged out.

Note I do not have much knowledge of what's going on in the Health* space

spec/factories/grda_warehouse/hud/clients.rb
```diff
@@ -1,5 +1,10 @@
 FactoryBot.define do
+  trait :standard_age do
+    DOB { (Date.current - 24.years).to_fs(:db) }
+  end
```
ttoomey: could add other traits to dry up the factories but doing the most minimal thing to get tests passing

eanders (approved):
Thanks @ttoomey !  I think we should keep it minimal for now.  Might be worth a tech debt story to revisit factories at some point in the future.

PR #4977, opened 2024-12-03
Improve enrollment change history

ttoomey:
## _Merging this PR_

## Description
* bug fix for broken use of current_user
* remove problematic rescue clauses
* preload for small n+1 improvement. Fixing would require refactor
* add smoke test

## Type of change
Bugfix, Housekeeping


app/models/grda_warehouse/enrollment_change_history.rb
```diff
@@ -50,6 +50,7 @@ def self.attributes_for_client_on_date client:, date:
         updated_at: Time.now,
         version: 1,
       }
+      # FIXME: how would current_user ever be defined here?
```
ttoomey: I don't understand this use of current_user. Possibly from 8c7e49b6942

spec/models/grda_warehouse/enrollment_change_history_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+    it 'passes a smoke test' do
+      expect {
+        described_class.create_for_clients_on_date!(client_ids: clients.map(&:id), date: today)
+      }.to change(described_class, :count).by(5).and make_database_queries(count: 10..11)
+
+      clients.map(&:reload)
+      clients.each do |client|
+        history = described_class.where(client_id: client.id).sole
+        expect(history.on).to eq(today)
+        # FIXME: not quite sure how to get the job to produce these
```
ttoomey: is it worth it to try and test that the job generates these properties correctly?

app/models/grda_warehouse/enrollment_change_history.rb
eanders: That sure looks like a nasty bug.  I can confirm, basically the rescue with an array is causing no data to be saved.  That should have been `User.setup_system_user` as it is later in the line.

spec/models/grda_warehouse/enrollment_change_history_spec.rb
eanders: It's probably failing because of the `current_user` bug you noted above.

app/models/grda_warehouse/enrollment_change_history.rb
ttoomey: thanks, I'll fix it up and remove the dangerous rescues. We can see if it crashes in practice

app/models/grda_warehouse/enrollment_change_history.rb
```diff
@@ -32,8 +32,9 @@ def self.generate_for_date!(date: Date.current)
         perform_later(date: date.to_s)
     end
 
+    # client_ids: destination client ids
     def self.create_for_clients_on_date! client_ids:, date:
-      clients = GrdaWarehouse::Hud::Client.destination.where(id: client_ids)
+      clients = GrdaWarehouse::Hud::Client.destination.where(id: client_ids).preload(:processed_service_history)
```
ttoomey, line 50: this might help a little but it looks like the job doesn't work in a way where it could easily be preloaded

app/models/grda_warehouse/enrollment_change_history.rb
```diff
@@ -42,17 +43,20 @@ def self.create_for_clients_on_date! client_ids:, date:
...
-        updated_at: Time.now,
+        created_at: now,
+        updated_at: now,
         version: 1,
       }
-      attributes_for_client[:residential] = client.enrollments_for_rollup(en_scope: client.scope_for_residential_enrollments(current_user), user: User.setup_system_user).to_json rescue '[]' # rubocop:disable Style/RescueModifier
-      attributes_for_client[:other] = client.enrollments_for_rollup(en_scope: client.scope_for_other_enrollments(current_user), user: User.setup_system_user).to_json rescue '[]' # rubocop:disable Style/RescueModifier
-      attributes_for_client[:days_homeless] = client.days_homeless rescue 0 # rubocop:disable Style/RescueModifier
+
+      user = User.setup_system_user
```
ttoomey, line 70: If I understand correctly, using the system user means the enrollment history will include all enrollments. Previously this used current_user which suggests the intent to restrict visibility. Does this seem okay?

spec/models/grda_warehouse/enrollment_change_history_spec.rb
```diff
@@ -0,0 +1,57 @@
...
+      source_clients.map(&:destination_client)
+    end
+
+    it 'passes a smoke test' do
+      GrdaWarehouse::Tasks::ServiceHistory::Enrollment.all.each(&:rebuild_service_history!)
+
+      expect {
+        described_class.create_for_clients_on_date!(client_ids: destination_clients.map(&:id), date: today )
+      }.to change(described_class, :count).by(3).
+      and make_database_queries(count: 60)
```
ttoomey: 60 queries for 3 client is not ideal

app/models/grda_warehouse/enrollment_change_history.rb
eanders: Correct, and yes, System User is correct.  Prior to adding the `current_user` that caused it to fail, it wasn't including a user at all.  This is behind an explicit permission and is used to troubleshoot historic problems with data loading.  It allows someone to answer the questions: "they had an enrollment here last month, but now it's gone? or this enrollment had 300 days of service and now only has 10, why?"  We use this log to determine when the change occurred and if it was gradual or sudden.  Once we've identified when, we can dig through import logs or paper trail to figure out the cause if necessary.

spec/models/grda_warehouse/enrollment_change_history_spec.rb
eanders: This should be roughly the same as the residential enrollments roll-up on the client dashboard.  Perhaps we can focus some effort there in the future and streamline both.

app/models/grda_warehouse/enrollment_change_history.rb
ttoomey: thanks, i'll add that to the inline docs

spec/models/grda_warehouse/enrollment_change_history_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+      source_clients.map(&:destination_client)
+    end
+
+    it 'passes a smoke test' do
+      GrdaWarehouse::Tasks::ServiceHistory::Enrollment.all.each(&:rebuild_service_history!)
+
+      expect do
+        described_class.create_for_clients_on_date!(client_ids: destination_clients.map(&:id), date: today)
+      end.to change(described_class, :count).by(3).
+        and make_database_queries(count: 60)
```
eanders: this is fine for now, but I wonder if we should change it to < 70 or something similar so if we tweak the method it uses we catch any increase in queries, but don't really care if it goes down.

eanders (approved):

ttoomey:
@eanders do you think this should go into 143? If this bug-fix seems less important, maybe 144?

eanders:
@ttoomey thanks for checking.  Let's target to 144 so it can have some extra time on QA.

PR #4980, opened 2024-12-03
Accept signed blob ID when uploading images

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7077 ( https://github.com/open-path/Green-River/issues/6208 )
Corresponding frontend PR: https://github.com/greenriver/hmis-frontend/pull/987

This PR updates the (for now, two) places in HMIS where we accept a blob ID: the client headshot uploader, and the file processor. These now accept a signed blob ID, perhaps partially addressing the concern Gig pointed out here https://github.com/open-path/Green-River/issues/6968#issuecomment-2514886196.

For backwards compat, these endpoints also still accept an unsigned blob ID. That should be turned off on the next release.

I split this work out of the other PR since I thought it would be helpful to be able to discuss independently.

QA instructions
- I should be able to upload a client headshot
- I should be able to re-upload a new client headshot for a client that already has one
- I should be able to submit a Client File

## Type of change

- [ ] I have updated the documentation (or not applicable) - documentation to be done in https://github.com/open-path/Green-River/issues/7054


drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -350,7 +350,8 @@ def full_name
   end
 
   def build_client_headshot_file(image_blob_id, current_user)
-    current_image_blob = ActiveStorage::Blob.find_by(id: image_blob_id)
+    # TODO(#7054) - accept either unsigned or signed blob ID for now, for backwards compatibility
```
eanders: What's the lifespan of these?  Could we remove this one week after this is deployed?  
Are we worried someone would have an unsigned url in a bookmark or email?

drivers/hmis/app/models/hmis/hud/processors/file_processor.rb
```diff
@@ -23,7 +23,8 @@ def process(field, value)
       elsif attribute_name == 'file_blob_id'
         return if attribute_value.nil?
 
-        blob = ActiveStorage::Blob.find_by(id: attribute_value)
+        # TODO(#7054) - accept either unsigned or signed blob ID for now, for backwards compatibility
+        blob = ActiveStorage::Blob.find_by(id: attribute_value) || ActiveStorage::Blob.find_signed(attribute_value)
```
eanders: Maybe it doesn't matter, but should we prefer signed over unsigned?
```suggestion
        blob = ActiveStorage::Blob.find_signed(attribute_value) || ActiveStorage::Blob.find_by(id: attribute_value)
```

drivers/hmis/app/models/hmis/hud/client.rb
martha: Yes, I think a week after deploy sounds about right. These aren't URLs so I'm not worried about bookmarks or saved links. We don't support "save and finish later" for assessments that include file uploads. I guess it would cause an error if someone uploaded a file in a form and then left the tab open for weeks without submitting, but that seems very unlikely.

eanders: That should be impossible with the auto logout after 30 minutes.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -350,7 +350,8 @@ def full_name
...
+    current_image_blob = ActiveStorage::Blob.find_by(id: image_blob_id) || ActiveStorage::Blob.find_signed(image_blob_id)
```
eanders: Would it make sense to check if `image_blob_id.to_s.to_i == image_blob_id.to_i` and then `find_by(id:` if true, and find signed if not?  That would prevent requiring two queries if `image_blob_id` is a digest/string.

martha: Yes good idea. It works with `image_blob_id.to_i.to_s`.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -350,7 +350,14 @@ def full_name
   end
 
   def build_client_headshot_file(image_blob_id, current_user)
-    current_image_blob = ActiveStorage::Blob.find_by(id: image_blob_id)
+    # TODO(#6208) - accept either unsigned or signed blob ID for now, for backwards compatibility
+    # binding.pry
```
gigxz: 🧹  binding pry

gigxz (approved):
Looks good to me!

drivers/hmis/app/models/hmis/hud/client.rb
gigxz: +1 to removing this in the subsequent release

drivers/hmis/app/models/hmis/hud/client.rb
martha: Thanks! 🤦‍♀️

app/models/concerns/client_file_base.rb
```diff
@@ -49,7 +50,7 @@ def tags
     end
 
     def file_exists_and_not_too_large
-      errors.add :client_file, full_message: 'No uploaded file found' if (client_file.byte_size || 0) < 100
+      errors.add :client_file, full_message: 'No uploaded file found. The upload may have expired; please try reloading the page and retrying the upload.' if (client_file.byte_size || 0) < 100
```
martha: Sorry about the scope creep. This is in order to give a more graceful ux on form processing in case the signed blob ID is expired. Let me know if you think this change is ok. I played around with just adding to the `file.errors` directly, but it gets cleared out later when `.valid?` is called.
```
f = Hmis::File.last
f.errors.add(:base, :invalid, full_message: 'oh no')
f.errors
=> #<ActiveModel::Errors [#<ActiveModel::Error attribute=base, type=invalid, options={:full_message=>"oh no"}>]>
f.valid?
f.errors
=> #<ActiveModel::Errors []>
```

app/models/concerns/client_file_base.rb
```diff
@@ -16,7 +16,8 @@ module ClientFileBase
 
     has_one_attached :client_file
 
-    validates_presence_of :name
+    # Name is really always required, this conditional just makes the client_file validation take precedence.
+    validates_presence_of :name, if: -> { client_file.attached? }
```
gigxz, line 21: I think maybe the comment is just confusing me here, what do you mean by "client_file validation"

It looks like the `file_exists_and_not_too_large` validation is currently validating that the `client_file` exists,  so we can rely on that and only validate presence of `name` if the file IF at has a `client_file` attachment.. is that right?

gigxz: Adding some screenshots from testing for reference:


This is what I see if I try to submit the File form without actually uploading anything. The first error is coming from the Form Definition validation code, because the "Attachment" field is marked as `required: true`. The second error is coming from here. (The issue of duplicated errors here pre-dates this PR).
Screenshot 2024-12-09 at 9 39 44 AM

This is what I see if I try to upload a blob that has been expired:
(The form required-ness validation isn't present because there was a blob id submitted)

Screenshot 2024-12-09 at 11 23 42 AM

app/models/concerns/client_file_base.rb
gigxz: Adding some screenshots from testing for reference. I only tested on the File form so far, not files on assessments.


This is what I see if I try to submit the File form without actually uploading anything. The first error is coming from the Form Definition validation code, because the "Attachment" field is marked as `required: true`. The second error is coming from here. (The issue of duplicated errors here pre-dates this PR).

Screenshot 2024-12-09 at 9 39 44 AM

This is what I see if I try to upload a blob that has been expired:
(The form required-ness validation isn't present because there was a blob id submitted)

Screenshot 2024-12-09 at 11 23 42 AM

gigxz: I don't know if it's worth adding this message since we added such a long expiration window, and most likely the user will see it because they didn't upload a file. But we should address the duplicate error problem separately.

app/models/concerns/client_file_base.rb
martha: I'm sorry about the confusing comment. Yes, your summary is right, so I'll update the comment to match (if you want to keep this, that is). I added this conditional in order to get rid of "Name can't be blank" in the screenshot below.
Screenshot 2024-12-09 at 11 57 57 AM

app/models/concerns/client_file_base.rb
martha: Ok, makes sense. I am happy to add a ticket to address that, and revert this change here.

app/models/concerns/client_file_base.rb
gigxz: I see, yes it makes sense to me to keep your change. Thanks!

app/models/concerns/client_file_base.rb
gigxz: Ok lets go with that. thanks

martha: Created https://github.com/open-path/Green-River/issues/7075

gigxz (approved):

PR #4982, opened 2024-12-05
7058-input-validation

ttoomey:
## _Merging this PR_

## Description

[GH issue](https://github.com/open-path/Green-River/issues/7058)

* Validate that numeric fields are the correct format.
* Validate bounds on numeric values
* Fix income processor crash on non-numeric values.

See also [client side numeric restriction added in this FE PR](https://github.com/greenriver/hmis-frontend/pull/988)

#### Misc:
Corrected deprecation warnings in rspec test:
```
# wrong
expect(1).to eq(2), details
# right
expect(1).to eq(2), -> { details }
```

## Type of change
Bug fix

## Testing
Update: Since this pr was created, the front-end no longer allows you to enter non-numeric values so this is hard to test. I guess just verify that numeric data is a captured correctly (for example on "Income and Sources")

* ~~Find or create an assessment on an enrollment in a transitional housing project~~
* ~~There should be an "Income and Sources", select yes to "Income form any source"~~
* ~~A table with numeric fields for entering income appears~~
* ~~enter a non-numeric value in one of the fields and submit~~
* ~~Form should not crash~~
* ~~A message should be shown indicating the the value is an invalid currency. Values with commas are not valid~~
* ~~A numeric integer or floating point value should be valid, even if it's negative~~


drivers/hmis/spec/models/hmis/form/validate_form_values_spec.rb
```diff
@@ -32,20 +32,19 @@
         [[], true],
         ['DATA_NOT_COLLECTED', false], # DNC is a valid answer for a required field
         [0, false],
-        ['value', false],
+        ['0', false],
```
ttoomey, line 35: I don't think 'value' should be acceptable as an integer

drivers/hmis/app/models/hmis/hud/processors/income_benefit_processor.rb
```diff
@@ -60,7 +60,14 @@ def income_source_attributes(amount_field, value)
       case @hud_values['IncomeBenefit.incomeFromAnySource']
       when 'YES'
         # If overall income was 1 (yes), then this specific income field must be 1 or 0 (yes or no)
-        bool_attribute_value = amount_attribute_value&.positive? ? 1 : 0
```
ttoomey, line 63: I'm curious as to what is converting these to integers

drivers/hmis/app/models/hmis/hud/processors/income_benefit_processor.rb
```diff
@@ -60,7 +60,13 @@ def income_source_attributes(amount_field, value)
...
+        case amount_attribute_value
+        when Integer, Float, nil
+          bool_attribute_value = amount_attribute_value&.positive? ? 1 : 0
+        else
+          # the client may send non-numeric values
```
ttoomey: I guess this is expected behavior. If not we should probably add
```suggestion
          Sentry.capture_message("Unexpected value \"#{amount_attribute_value}\" received for #{amount_attribute_name}")
```

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -441,6 +441,24 @@ def assessment_date_item
...
+
+  def validate_input_format(item, value)
+    return unless value.present?
+    # special case values?
+    return if value == 'DATA_NOT_COLLECTED'
+    return if value == '_HIDDEN'
+
+    case item.type
+    when 'INTEGER'
+      ['not a valid integer'] unless value.to_s.strip =~ INTEGER_RGX
```
ttoomey: perhaps this is too far? the bug in question only deals with currency fields but it seems like we shouldn't except non-numeric values for integers either

ttoomey:
@gigxz I've made adjustments to include bounds validation per our discussion today

ttoomey:
@gigxz lmk if there's anything I can do to help this along

drivers/hmis/app/models/hmis/form/numeric_input_validator.rb
```diff
@@ -0,0 +1,51 @@
...
+
+  def validate_bounds(item, value)
+    return [] if item.bounds.blank?
+
+    item.bounds.each_with_object([]) do |bound, errors|
+      next unless bound.severity == 'error'
+
+      case bound.type
+      when 'MAX'
+        errors << "must be less than or equal to #{bound.value_number}" if value > bound.value_number
```
gigxz, line 47: `bound.value_number` can be nil, in the case where the bound is against a local constant or another question. Maybe above, adding `next if bound.value_number.nil?`

drivers/hmis/app/models/hmis/hud/processors/income_benefit_processor.rb
gigxz: Agreed, we should add that. Since the frontend input component has been updated, It is not expected for these to be non-numeric.

drivers/hmis/app/models/hmis/hud/processors/income_benefit_processor.rb
gigxz: The frontend is submitting integers

gigxz (commented):
Looks good! Just one request to fix a comparison to nil when checking bounds. Sorry for the delay!

drivers/hmis/app/models/hmis/form/numeric_input_validator.rb
ttoomey: good catch, will fix. thanks!

gigxz (approved):

PR #4984, opened 2024-12-05
Add default Data Collected About value to Occurrence Point Forms

martha:
## _Merging this PR_

## Description
Bugfix arising from https://github.com/open-path/Green-River/issues/6972#issuecomment-2520547404

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
@@ -10,7 +10,7 @@ module Types
   class HmisSchema::OccurrencePointForm < Types::BaseObject
     field :id, ID, null: false, extras: [:parent]
     # Which clients this data should be collected for
-    field :data_collected_about, Types::Forms::Enums::DataCollectedAbout, null: false
+    field :data_collected_about, Types::Forms::Enums::DataCollectedAbout, null: false, default_value: 'ALL_CLIENTS'
     # Form used for Viewing/Creating/Editing records
     field :definition, Types::Forms::FormDefinition, null: false, extras: [:parent]
 
@@ -25,10 +25,6 @@ module Types
       definition = object.definition
       definition.filter_context = { project: parent_project(parent) }
       definition
-    end
-
-    def data_collected_about
-      object.data_collected_about || 'ALL_CLIENTS'
     end
 
     private def parent_project(parent)
--- a/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
@@ -27,6 +27,10 @@ module Types
       definition
     end
 
+    def data_collected_about
+      object.data_collected_about || 'ALL_CLIENTS'
+    end
+
     private def parent_project(parent)
       if parent.is_a?(Hmis::Hud::Project)
         parent


drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
```diff
@@ -27,6 +27,10 @@ def definition(parent:)
       definition
     end
 
+    def data_collected_about
```
martha: @gigxz , I'm still a bit confused about what caused the bug. I wasn't able to reproduce it locally, and also noticed that in the spec test, the instance has a nil value for data collected about. https://github.com/greenriver/hmis-warehouse/blob/release-143/drivers/hmis/spec/requests/hmis/enrollment_occurrence_point_spec.rb#L69

martha: I can keep investigating before merging this PR if you think it is non-urgent.

martha: Enrollment occurrence_point_forms provides a default https://github.com/greenriver/hmis-warehouse/blob/release-143/drivers/hmis/app/models/hmis/hud/enrollment.rb#L360

And so do Client Detail Forms https://github.com/greenriver/hmis-warehouse/blob/release-143/drivers/hmis/app/graphql/types/hmis_schema/query_type.rb#L512

Project occurrence_point_form_instances don't appear to provide a default https://github.com/greenriver/hmis-warehouse/blob/release-143/drivers/hmis/app/models/hmis/hud/project.rb#L297-L309

But this is no longer used from the frontend, and the Sentry error refers to `GetEnrollmentDetails`

gigxz: > Project occurrence_point_form_instances don't appear to provide a default .. But this is no longer used from the frontend, and the Sentry error refers to GetEnrollmentDetails

I see, thanks for that summary. At the time when the issue surfaced, the backend was on release-142 and the frontend was still on release-141. So the frontend was still using the project occurrence point details query in GetEnrollmentDetails. 

So now that the frontend and backends are all on 142, this bug would only be present when resolving `project { occurrencePointForms }`, which we no longer do. It's still part of the API, though. I think its safest to merge this change.

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
```diff
@@ -27,6 +27,10 @@ def definition(parent:)
...
+      object.data_collected_about || 'ALL_CLIENTS'
+    end
+
```
gigxz: There's a more succinct way to set the default:
```
field :data_collected_about, Types::Forms::Enums::DataCollectedAbout, null: false, default_value: 'ALL_CLIENTS'
```

`default_value` is not part of ruby-graphql, we defined it ourselves on BaseField: https://github.com/greenriver/hmis-warehouse/blob/30ba0e4dd283de038b2bafb08c9fd1220fbce175/drivers/hmis/app/graphql/types/base_field.rb#L71-L81

gigxz (approved):

drivers/hmis/app/graphql/types/hmis_schema/occurrence_point_form.rb
martha: Ah, it makes sense that this is related to the backend and frontend being on different releases. I was wondering if it was something like that. I was able to reproduce it and confirm that setting the default value fixes the problem.

martha: Is it ok to keep this PR targeted at release-143, or would you like me to open it against release-142?

gigxz: I think 143 is good, thanks!

PR #4988, opened 2024-12-09
6233 migrate to schema rb

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/6233)

* use ruby to hold our db schema
* the fx gem supports triggers and functions
* the scenic gem (already in use) support views
* replace postgreSQL Rules with triggers/functions
* replace the deprecated rake 'db:structure' tasks with 'db:schema'

## Type of change
Code clean-up


ttoomey:
per discussion with Elliot, I'll break this PR into smaller pieces that are easier to review:
- transition from rules to triggers
- fix the PK types in our dbs

PR #4993, opened 2024-12-09
Resolve Geolocation in GraphQL API, support collection on CLS and Assessments

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/5726
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/992

This change supports processing Geolocation onto CurrentLivingSituation and Custom Assessments. It also supports resolving them for display in an Enrollment Locations Map (map not yet implemented).

See issue for testing instructions, including sample forms to use.

Summary of changes:
* Resolve geolocation on GraphQL API in the following places:
  * on CurrentLivingSituation, so the location can be displayed when viewing a CLS that previously collected it
  * on Assessment, so the location can be displayed when viewing an Assessment that previously collected it
  * on External Form Submission, so the location can be displayed when viewing the submission
  * on Enrollment, so the locations can be displayed in a map (not yet implemented on frontend)
* Add permission `canViewEnrollmentLocationMap` for viewing map (not yet implemented on frontend)
* Make updates to `GeolocationProcessor` to allow for re-processing locations (it was built for processing External Forms, which only get processed once, so there was some additional work needed)
* Move HMIS-specific Location stuff into LocationExtension

## Type of change
New feature

  - [x] ensured this does not introduce N+1s -- N+1 are present but annotated


drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
```diff
@@ -0,0 +1,30 @@
...
+    field :collected_by, String, null: true
+
+    # dates and times - todo remove what we dont need
+    field :located_on, GraphQL::Types::ISO8601Date, null: true
+    field :located_at, GraphQL::Types::ISO8601DateTime, null: true
+    field :processed_at, GraphQL::Types::ISO8601DateTime, null: true
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :updated_at, GraphQL::Types::ISO8601DateTime, null: false
+
+    # todo add project name? via source.project.name
```
gigxz, line 22: x

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -277,6 +277,11 @@ type Assessment {
   enrollment: Enrollment!
   event: Event
   exit: Exit
+
+  """
+  Client Location where the assessment was performed
+  """
+  geolocation: Geolocation
```
gigxz, line 284: @martha I liked your point about naming. Maybe these should all be `clientLocation[s]`?
`location` on it's own feels like it may get confusing down the line with other types of locations (project locations, tagged locations/areas/places, etc?)

clientLocation is also more in-line with name of clh_locations (client history locations) table

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -13,51 +13,81 @@ def schema
...
 
+    def process(field, value)
       attribute_name = ar_attribute_name(field)
-
+      # The only expected attribute name is 'coordinates'
       raise ArgumentError, "Unexpected attribute for Geolocation: #{attribute_name}" unless attribute_name == 'coordinates'
 
-      attribute_value = JSON.parse(value, symbolize_names: true)
+      # if 'Geolocation.coordinates: nil' was submitted, destroy the clh location record
+      return destroy_record if value.nil? || value.empty?
```
ttoomey: maybe just use present?
```suggestion
      return destroy_record unless value.present?
```

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -13,51 +13,81 @@ def schema
...
+
+      attribute_value = clean_coordinate_value(value)
```
ttoomey: exclude empty string values so we don't have to check for that case?
```suggestion
      attribute_value = clean_coordinate_value(value).compact_blank
```

drivers/hmis/app/models/hmis/hud/processors/geolocation_processor.rb
```diff
@@ -13,51 +13,81 @@ def schema
...
       latitude = attribute_value[:latitude]
       longitude = attribute_value[:longitude]
       not_collected_reason = attribute_value[:notCollectedReason]
 
-      raise ArgumentError, 'Geolocation coordinates in unexpected format' unless (latitude && longitude) || not_collected_reason
-
-      return @processor.send(factory_name).destroy if not_collected_reason
+      return destroy_record if not_collected_reason # not collected reason is returned from external PIT form
+      return destroy_record if latitude.nil? && longitude.nil?
```
ttoomey: what if we get just one coord (not likely). But maybe it should early-return unless we have both
```suggestion
      return destroy_record unless latitude && longitude
```

drivers/hmis/app/graphql/types/hmis_schema/geolocation_coordinates.rb
```diff
@@ -0,0 +1,16 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::GeolocationCoordinates < Types::BaseObject
+    field :id, ID, null: false
+    field :latitude, String, null: true, method: :lat
+    field :longitude, String, null: true, method: :lon
```
ttoomey: should these be nullable actually? Or are they doing double-duty as form inputs

drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
```diff
@@ -0,0 +1,46 @@
...
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::Geolocation < Types::BaseObject
+    field :id, ID, null: false
+    field :coordinates, HmisSchema::GeolocationCoordinates, null: true
+    field :collected_by_project_name, String, null: true, method: :collected_by
+    field :located_at, GraphQL::Types::ISO8601DateTime, null: false, description: 'Timestamp when the location was collected'
+    field :source_current_living_situation, HmisSchema::CurrentLivingSituation, null: true, description: 'Associated Current Living Situation record, if this location was collected on a CurrentLivingSituation form'
+    field :source_assessment, HmisSchema::Assessment, null: true, description: 'Associated Assessment record, if this location was collected on an assessment'
```
ttoomey: what's the use-case for walking back to the owner?

ttoomey (approved):
Looks great!

drivers/hmis/app/graphql/types/hmis_schema/geolocation_coordinates.rb
gigxz: No they shouldn't, but the db table has them as nullable. I will update this to make them non-nullable and use a scope instead to filter out any geolocations that are missing lat/lon, since we can't show them anyway.

drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
gigxz: it's part of the location map design. the tooltip shows the "source" record

drivers/hmis/app/graphql/types/hmis_schema/geolocation_coordinates.rb
gigxz: Done with https://github.com/greenriver/hmis-warehouse/pull/4993/commits/9be98fd44e973408ebaf9890747742c551e8341d

drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
gigxz: Updating this to just resolve the name of the form

drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
```diff
@@ -0,0 +1,60 @@
...
+        load_ar_association(form_processor, :definition)&.title || 'Assessment'
+      when HmisExternalApis::ExternalForms::FormSubmission.sti_name
+        load_ar_association(form_processor, :definition)&.title || 'External Form'
+      end
+    end
+
+    # User who collected the location. We don't actually have this, so we just look at who created
+    # the form that collected the location. This could be wrong if the location was added later by someone else.
+    # It would be better to add a column to clh_locations to store the user who collected the location.
+    def collected_by
```
gigxz, line 48: This should be a follow-up. For now I'm just giving the API the right shape, and throwing something in there.
 @eanders what do you think about adding a collected_by_user_id reference to the clh_locations table?

drivers/client_location_history/app/models/client_location_history/location.rb
```diff
@@ -12,16 +12,14 @@ class Location < GrdaWarehouseBase
...
-    before_create :ensure_grda_warehouse_source
 
     MARKER_COLOR = '#72A0C1'.freeze
 
-    private def ensure_grda_warehouse_source
-      # This somewhat hacky solution gets around the fact that during HMIS Form Processing, we haven't yet saved the
-      # Enrollment being generated, so we don't yet have an ID with which to get the Warehouse enrollment.
-      return unless source_type&.starts_with? 'Hmis::Hud::'
+    # Locations where both `lat` and `lon` are present
+    scope :valid, -> { where.not(lat: nil).and(where.not(lon: nil)) }
```
eanders, line 19: I _think_ the following returns an equivalent query and is a little easier to read:
```suggestion
    scope :valid, -> { where.not(lat: nil, lon: nil) }
```


```
irb(main):002:0> ClientLocationHistory::Location.where.not(lat: nil).and(ClientLocationHistory::Location.where.not(lon: nil)).to_sql
=> "SELECT \"clh_locations\".* FROM \"clh_locations\" WHERE \"clh_locations\".\"lat\" IS NOT NULL AND \"clh_locations\".\"lon\" IS NOT NULL"
irb(main):003:0> ClientLocationHistory::Location.where.not(lat: nil, lon: nil).to_sql
=> "SELECT \"clh_locations\".* FROM \"clh_locations\" WHERE NOT (\"clh_locations\".\"lat\" IS NULL AND \"clh_locations\".\"lon\" IS NULL)"
```

drivers/hmis/app/graphql/schema.graphql
eanders: clientLocation (or maybe something like clientPin, or clientCoordiantes) feels right to me too.

drivers/hmis/app/graphql/types/hmis_schema/geolocation.rb
eanders: @gigxz, I'm fine with adding that, but it'll need to be optional since we won't know who collected an imported location, or one generated from an enrollment.

drivers/hmis/app/graphql/schema.graphql
martha: I like clientLocation too!

martha (approved):

drivers/client_location_history/app/models/client_location_history/location.rb
gigxz: We want them both `lat` AND `lon` to be present. I don't think the second query achieves that: `WHERE NOT (\"clh_locations\".\"lat\" IS NULL AND \"clh_locations\".\"lon\" IS NULL)`

drivers/hmis/app/graphql/schema.graphql
gigxz: sticking with geolocation per dev standup (geolocation terminology already used in form api; has searchability benefit; didn't seem like there was a very strong pref)

PR #4995, opened 2024-12-12
Fix stalled method on DataSource; tests to prove it is fixed

eanders:
## _Merging this PR_

## Description
* Fixes calculation to determine if a data source that is setup to auto import is stalled
* Tests to confirm edge cases

## Type of change

### DIFF
--- a/app/models/grda_warehouse/data_source.rb
+++ b/app/models/grda_warehouse/data_source.rb
@@ -499,7 +499,7 @@ class GrdaWarehouse::DataSource < GrdaWarehouseBase
 
   # Returns the date of the most recent fully successful import if the import is stalled, nil if it is not stalled
   # @return [Date, nil] The date the import stalled, or nil if not stalled.
-  def stalled?
+  def stalled_date
     return nil if import_paused
     return nil unless hmis_import_config&.active
 
@@ -507,9 +507,10 @@ class GrdaWarehouse::DataSource < GrdaWarehouseBase
     # fetch the expected number, and confirm they all arrived within a 24 hour window
     most_recent_uploads = uploads.completed.
       # limit look back to 6 months to improve performance, but to potentially highlight missing data
-      where(user_id: User.system_user.id, completed_at: 6.months.ago..DateTime.current).
+      where(user_id: User.system_user.id, completed_at: 6.months.ago..Time.current).
       order(completed_at: :desc, created_at: :desc).
       select(:id, :data_source_id, :user_id, :completed_at).
+      distinct.
       first(hmis_import_config.file_count)
     # We didn't find any uploads in the last 6 months, assume this isn't connected yet
     return nil unless most_recent_uploads.present?
@@ -527,7 +528,7 @@ class GrdaWarehouse::DataSource < GrdaWarehouseBase
     end
 
     # If we processed the expected number of files within a 24 hour period, we're good
-    return nil if min_completion_time > 24.hours.ago && received_files_count == hmis_import_config.file_count
+    return nil if min_completion_time > 24.hours.ago && received_files_count >= hmis_import_config.file_count
 
     # Note the last time we received a file
     min_completion_time.to_date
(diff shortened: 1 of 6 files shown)


app/models/grda_warehouse/data_source.rb
```diff
@@ -497,36 +497,40 @@ def unprocessed_enrollment_count
...
   end
 
-  # Have we received the expected number of files at least once over the past 48 hours?
-  # there is an assumption that within
-  # return the date of the most-recent fully successful import
-  def stalled_since?(date)
-    return nil unless date.present?
+  # Returns the date of the most recent fully successful import if the import is stalled, nil if it is not stalled
+  # @return [Date, nil] The date the import stalled, or nil if not stalled.
+  def stalled?
```
ttoomey: Suggest removing the '?' since it's not returning a boolean
 ```suggestion
   def stall_date
```

File: app/models/grda_warehouse/data_source.rb [hunk 2]
```diff
@@ -497,36 +497,40 @@ def unprocessed_enrollment_count
...
+      # and it came in the last 24 hours, we're good
+      return nil if min_completion_time > 24.hours.ago
 
-    most_recent_upload.completed_at.to_date
+      # if not, return the last time we received a file
+      return min_completion_time.to_date
+    end
+
+    # If we processed the expected number of files within a 24 hour period, we're good
+    return nil if min_completion_time > 24.hours.ago && received_files_count == hmis_import_config.file_count
```
ttoomey: looks like this would appear as stalled if we processed more than the expected number... is that what we want?
```suggestion
    return nil if min_completion_time > 24.hours.ago && received_files_count >= hmis_import_config.file_count
```

app/models/grda_warehouse/data_source.rb
```diff
@@ -497,36 +497,40 @@ def unprocessed_enrollment_count
...
 
     # hmis_import_config.file_count is the expected number of uploads for a given day
     # fetch the expected number, and confirm they all arrived within a 24 hour window
     most_recent_uploads = uploads.completed.
-      # limit look back to 6 months to improve performance
-      where(user_id: User.system_user.id, completed_at: 6.months.ago.to_date..Date.current).
-      order(created_at: :desc).
+      # limit look back to 6 months to improve performance, but to potentially highlight missing data
+      where(user_id: User.system_user.id, completed_at: 6.months.ago..DateTime.current).
+      order(completed_at: :desc, created_at: :desc).
```
ttoomey, line 511: We assume unique files here right? If we expected two different files but just got one of the two files twice, is that a problem? Wondering if there should be count distinct on upload.file

app/models/grda_warehouse/data_source.rb
```diff
@@ -497,36 +497,40 @@ def unprocessed_enrollment_count
...
     return nil unless hmis_import_config&.active
 
     # hmis_import_config.file_count is the expected number of uploads for a given day
     # fetch the expected number, and confirm they all arrived within a 24 hour window
     most_recent_uploads = uploads.completed.
-      # limit look back to 6 months to improve performance
-      where(user_id: User.system_user.id, completed_at: 6.months.ago.to_date..Date.current).
-      order(created_at: :desc).
+      # limit look back to 6 months to improve performance, but to potentially highlight missing data
+      where(user_id: User.system_user.id, completed_at: 6.months.ago..DateTime.current).
```
ttoomey: probably favor Time or DateTime. Also do we need to exclude dates in the future?
```suggestion
      where(user_id: User.system_user.id, completed_at: 6.months.ago..Time.current).
```

ttoomey (approved):
looks good, I left a few suggestions

app/models/grda_warehouse/data_source.rb
eanders: That _should_ be impossible since we're limiting to `first(hmis_import_config.file_count)` in the query, but no harm.

app/models/grda_warehouse/data_source.rb
eanders: Not sure how you could get duplicates given that it is just querying one table, but no harm in a distinct since we're selecting a limited set of columns so it won't try to distinct on a json column.

PR #5002, opened 2024-12-16
Replace CombinePDF with Pdfunite

eanders:
## _Merging this PR_

## Description
* Replaces abandoned CombinePDF with Pdfunite
* Adds pdfunite to the alpine image through `poppler-utils`
* Bumps the bundle version
* Bumps `stringio` version

## Type of change
* Dependency updates


eanders:
Keeping this as a draft until after production testing.

eanders:
@ttoomey this is ready for a review, but preventing merge until we've had a chance to test in production.

app/controllers/concerns/health_careplan.rb
```diff
@@ -62,7 +62,7 @@ def careplan_pdf_coversheet
         wait_until: 'networkidle0',
         print_background: true,
       }
-      CombinePDF.parse(Grover.new(wrap_in_html(coversheet), **grover_options).to_pdf, allow_optional_content: true)
+      Pdfunite.join(Grover.new(wrap_in_html(coversheet), **grover_options).to_pdf) { |obj| obj }
```
ttoomey: I guess the pdf unite gem's block is required? Seems like an awkward API. Maybe we provide a wrapper?
```ruby
class PdfGenerator
  # provides a default block
  def self.merge_pdfs(objects, &block)
    Pdfunite.join(objects, &(block || proc { |pdf| pdf }))
  end
end
```

```suggestion
      PdfGenerator.merge_pdfs(Grover.new(wrap_in_html(coversheet), **grover_options).to_pdf)
```

It would be nice to tidy our Pdf rendering up into one class so we don't have to call grover directly either, but that seems out of scope

ttoomey (approved):
Looks good, thanks!

app/controllers/concerns/health_careplan.rb
eanders: Yeah, that one tripped me up, ended up digging into the source.  Moving it to the PdfGenerator is a great idea, thanks!

eanders: @ttoomey I went with a slightly different approach to try and enforce some consistency in usage (which we actually have.)  Do you think this is ok, or do you prefer your more flexible approach?

drivers/health_comprehensive_assessment/app/models/health_comprehensive_assessment/document_exports/health_ca_pdf_export.rb
```diff
@@ -98,15 +98,15 @@ def perform
           },
         )
 
-        pdf = CombinePDF.new
-        pdf << CombinePDF.parse(PdfGenerator.new.render_pdf(first_page_html, options: first_page_options), allow_optional_content: true)
-        pdf << CombinePDF.parse(PdfGenerator.new.render_pdf(html, options: body_options), allow_optional_content: true)
+        pdf = []
+        pdf << PdfGenerator.merge_inline_pdfs(PdfGenerator.new.render_pdf(first_page_html, options: first_page_options))
+        pdf << PdfGenerator.merge_inline_pdfs(PdfGenerator.new.render_pdf(html, options: body_options))
```
ttoomey: this is fine although  maybe for consistency we should drop a pass-through class method on PdfGenerator for render_pdf so we aren't mixing instance/class methods
```ruby
class PdfGenerator
  def self.render_pdf(...)
    new.render_pdf(...)
  end
end
```

```suggestion
        pdf = []
        
        pdf << PdfGenerator.merge_inline_pdfs(PdfGenerator.render_pdf(first_page_html, options: first_page_options))
        pdf << PdfGenerator.merge_inline_pdfs(PdfGenerator.render_pdf(html, options: body_options))
```

also, could dry this up a bit. Maybe not worth it though
```ruby
pdf = [
  [first_page_html, first_page_options],
  [html, body_options]
].map do |html, options |
   PdfGenerator.merge_inline_pdfs(PdfGenerator.render_pdf(html, options: options))
end
```

ttoomey (approved):
Thanks! This seem good to me. I left another small suggestion but feel free to ignore

PR #5007, opened 2024-12-16
Captive Portal Updates

dtgreiner:
## _Merging this PR_

## Description
Updates for TalentLMS Captive Portal
- Users who are inactive in talentlms are always redirected to the captive portal
- Captive displays if an active account cannot be found for a domain
- Add site config flag to allow using warehouse email for talentlms by default
- Add talentlms::config flag to allow config to do an automatic redirect to the required course (if it is the only course required)
- Always show captive portal if more than one course is required

## Type of change
New feature

### DIFF
--- a/app/views/admin/talentlms/_form.haml
+++ b/app/views/admin/talentlms/_form.haml
@@ -1,3 +1,4 @@
 = f.input :subdomain
 = f.input :api_key, label: 'API Key'
-= f.input :create_new_accounts, label: 'Allow creation of new accounts'
+= f.input :create_new_accounts, label: 'Allow creation of new accounts', hint: 'This will allow accounts to be automatically generated if they do not already exists in the LMS system.'
+= f.input :allow_automatic_redirect_to_course, label: 'Allow automatic course redirects', hint: 'This will allow bypassing the training page when only one course requires training.'
--- a/app/views/admin/talentlms/_form.haml
+++ b/app/views/admin/talentlms/_form.haml
@@ -1,4 +1,4 @@
 = f.input :subdomain
 = f.input :api_key, label: 'API Key'
 = f.input :create_new_accounts, label: 'Allow creation of new accounts', hint: 'This will allow accounts to be automatically generated if they do not already exists in the LMS system.'
-= f.input :allow_automatic_redirect_to_course, label: 'Allow automatic course redirects', hint: 'This will allow bypassing the training page when only one course requires training.'
+= f.input :allow_automatic_redirect_to_course, label: 'Allow automatic course redirects', hint: 'When checked, users who only need to complete one course will be immediately redirected to TalentLMS upon login to the warehouse.  When unchecked or if a user needs to complete more than one course, a summary page will direct users to complete course with links to TalentLMS.'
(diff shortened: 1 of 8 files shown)


db/warehouse/migrate/20241216184819_talent_configuration_flags.rb
```diff
@@ -0,0 +1,6 @@
+class TalentConfigurationFlags < ActiveRecord::Migration[7.0]
+  def change
+    add_column :configs, :default_lms_email_to_warehouse_email, :boolean
+    add_column :talentlms_configs, :allow_automatic_redirect_to_course, :boolean
```
eanders: I think this should default to true so we don't break existing installations right?
```suggestion
    add_column :talentlms_configs, :allow_automatic_redirect_to_course, :boolean, default: true
```

spec/models/talentlms/facade_spec.rb
```diff
@@ -73,6 +73,64 @@
...
+      api_stub_response = lms.get_lms_account_data(config, lms_login)
+      expect(lms.active_user?(nil, nil, api_stub_response)).to be_truthy
+    end
+
+    it 'returns false if user is not active' do
+      stub_so_api_user_is_inactive(config: config)
+      expect(lms.active_user?(config, lms_login)).to be_falsey
+    end
+
+    it 'returns false if user is active in provided api response' do
```
eanders: Is this just missing a not?
```suggestion
    it 'returns false if user is not active in provided api response' do
```

app/views/admin/talentlms/_form.haml
```diff
@@ -1,3 +1,4 @@
 = f.input :subdomain
 = f.input :api_key, label: 'API Key'
-= f.input :create_new_accounts, label: 'Allow creation of new accounts'
\ No newline at end of file
+= f.input :create_new_accounts, label: 'Allow creation of new accounts', hint: 'This will allow accounts to be automatically generated if they do not already exists in the LMS system.'
+= f.input :allow_automatic_redirect_to_course, label: 'Allow automatic course redirects', hint: 'This will allow bypassing the training page when only one course requires training.'
```
eanders:
```suggestion
= f.input :allow_automatic_redirect_to_course, label: 'Allow automatic course redirects', hint: 'When checked, users who only need to complete one course will be immediately redirected to TalentLMS upon login to the warehouse.  When unchecked or if a user needs to complete more than one course, a summary page will direct users to complete course with links to TalentLMS.'
```

eanders (changes requested):
Looks great @dtgreiner.  Dropped a couple ideas inline.

spec/models/talentlms/facade_spec.rb
dtgreiner: yes! thanks for catching that. I'll include it in the next commit

db/warehouse/migrate/20241216184819_talent_configuration_flags.rb
dtgreiner: Yes, defaulting this to true will keep current installations more closely aligned with the current setup. I'll include it in the next commit.

app/views/admin/talentlms/_form.haml
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5007/commits/c4228c3594a3b784977b32b918864b8539524568)

dtgreiner:
@eanders This is ready for another check. Thanks!

eanders (approved):

PR #5013, opened 2024-12-20
CE workflow spike

ttoomey:
## _Merging this PR_

## Description
WIP, Spike for CE
https://github.com/open-path/Green-River/issues/7097
frontend pr: https://github.com/greenriver/hmis-frontend/pull/1089

Adds dependencies: aasm and dentaku gems

There are three modules in this spike:
1) **Coordinated Entry (CE)** - Defines opportunities, referrals, actions that occur during a referral
1a) **Client Match** - Handle eligibility and prioritization for clients into opportunities
2) Workflow Definition (WFD) - Defines the structure and rules of workflows
3) Workflow Execution (WFE) - Handles the runtime execution of workflow instances

### 1) Coordinated Entry (CE)

```mermaid
classDiagram
    Opportunity  --> WFD-Template
    Opportunity  --> Referral
    Referral -->  WFE-Instance
    Referral  -->  ReferralParticipant
    ReferralParticipant --> WFD-Swimlane

    class Opportunity {
        +name: string
        +status: string
        +project: Project
    }
    class Referral {
        +status: string
        +workflow_engine: Engine
        +client: Client
    }
    class ReferralParticipant {
        +user: User
    }
```

- **Opportunity**: Represents an available resource (e.g., housing unit)
- **Referral**: Tracks a client's progress through the workflow
- **ReferralParticipant**: Associates users with specific roles in the referral

#### 1a) CE Client Matching

The `ClientMatch` module supports matching clients to into pools based on configuration in eligibility requirements and priority configurations configuration. Client matches are persisted as 'Candidates'

The `ClientMatch::Engine` has responsibility for executing policy eligibility rules and priority schemes against a set of clients saving the results as match candidates.

```mermaid
classDiagram
    Opportunity  -->  CandidatePool
   
    Candidate --> CandidatePool

    CandidatePool <.. Rule

    Opportunity  <-->  Category

    class CandidatePool {
        +eligibility_requirement: Expression
        +priority_formula: Expression
    }

    class Candidate {
        +client: Client
        +priority
    }
    class Rule  {
        +rule_type: eligibility, prioritization 
        +owner: Proj, Org, Opp, Funder
        +applicability: funder types, project_types
        +expression: string
    }
    
```
- **CandidatePool**: Automatically generated from requirements/priority schemes. Allows the same set of candidates to be shared across opportunities with identical eligibility/priority configuration.
- **OpportunityCategory**: Allows opportunities to be categorized; this restricts the same client from being referred to multiple opportunities of the same type.
- **Candidate**: Tracks how a client is matched to a given policy (and the opportunities that use this policy). 
- **Rule**: Cascading configuration for clients requirements and priority. The "owner" is the entity that manages this configuration. The applicability context provides additional rules that allow configuration to target entity attributes, such as funder type or project type

#### Proposed API
```
# all opportunities the client is currently eligible for based on candidate records and accounting for referrals  
> Hmis::Ce::Opportunity.for_client(client) # not tested
# all candidates eligible for a given opportunity in priority order, accounting for open referrals
> Hmis::Ce::ClientMatch::Candidate.prioritized.for_opportunity(opportunity) # not tested
```

### 2) Workflow Definition
The Workflow Definition (WFD) Defines the structure and rules of workflows. Terminology is inspired by BPMN.

```mermaid
classDiagram
    Template -->  Node
    Template -->  Flow
    Template  -->  Swimlane
    Node  -->  Flow : outflows
    Node  --> Flow : inflows
    Task ..|> Node
    Gateway ..|> Node
    StartEvent ..|> Node
    EndEvent ..|> Node
    Task --> Swimlane

    class Template {
        +name: string
    }
    class Node {
        +type: string
        +name: string
        +trigger_config: json
    }
    class Flow {
        +condition: string
        +position: integer
    }
    class Swimlane {
        +name: string
    }
    class Gateway {
        +gateway_type: string
    }
      class Task {
        +form_definition
    }
```

- **Template**: The top-level container for a workflow definition. To be reused across multiple instances
- **Node**: Base class for all workflow elements
  - **Task**: Work items that need to be completed by users. References form definition and swimlane 
  - **Gateway**: Decision points that control flow branching and joins
  - **StartEvent/EndEvent**: Define workflow boundaries
- **Flow**: Connections between nodes that define the possible paths
- **Swimlane**: Organizational grouping that helps determine task assignments

### 3) Workflow Execution

```mermaid
classDiagram
    Instance --> Step
    Instance --> AuditEvent
    Step --> StepAssignment
    Referral --> Instance
    Step --> Task

    class Instance {
        +template: Template
    }
    class Step {
        +status: string
        +submitted_values: json
    }
    class StepAssignment {
        +user: User
    }
    class AuditEvent {
        +event_type: string
        +event_data: json
    }
```

- **Instance**: A running instance of a workflow template
- **Step**: Tracks the state of a specific task
- **StepAssignment**: Links users to their assigned tasks
- **AuditEvent**: Records all significant workflow events

#### Workflow Engine

The `Engine` class is the core component that manages workflow execution:

### Key Responsibilities

1. **State Management**
   - Tracks current workflow state
   - Manages step transitions
   - Controls task assignments

2. **Flow Control**
   - Evaluates conditions using Dentaku
   - Handles parallel and exclusive paths
   - Processes triggers and messages

3. **Manages Side-Effects**
   - Fires workflow events, triggers, and messages
   - Maintains audit trail

#### Example Flow

```mermaid
sequenceDiagram
    participant User
    participant Engine
    participant MessageHandler
    participant AssignmentHandler

    User->>Engine: start_workflow!
    Engine->>MessageHandler: start workflow triggers
    loop Until End Event
        Engine->>Engine: traverse_nodes
        Engine->>User: create step
        Engine->>AssignmentHandler: get users step
        AssignmentHandler->>Engine: matching users
        User->>Engine: start_step!
        Engine->>MessageHandler: start step triggers
        User->>Engine: complete_step!
        Engine->>MessageHandler: complete step triggers 
    end
    Engine->>MessageHandler: end workflow triggers
```

## Type of change
New feature


db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
```diff
@@ -0,0 +1,55 @@
...
+    create_table :wfd_nodes do |t|
+      t.references :template, null: false, foreign_key: { to_table: :wfd_templates }
+      t.string :type, null: false # For STI
+      t.jsonb :trigger_config # when to send notifications, create ce events, state changes, api calls
+      t.string :name
+      # task nodes have forms
+      t.references :form_definition, foreign_key: { to_table: :hmis_form_definitions }
+      # gateway nodes have types
+      t.string :gateway_type
+
```
gigxz, line 35: I think the node is going to reference the category/role/type of user that this node will be assigned_to. Not sure what other structures we may need to represent those relationships

ttoomey: Yes, I added support for this as follows:
* added a "swimlane" to step (role is a little overloaded so I'm using the workflow-specific term)
* added referral participants to track which users belong to swimlanes for a given referral
* the engine takes an assignmentHandler, currently it just assigns all users in matching the step's swimlane

We probably want some rules around what to do when there are no users for a given swimlane. Maybe a catch-all user or we could make the step as stalled

eanders: It's an interesting question how we want to handle the situation where a step requires a particular actor, but the actor hasn't been assigned.  In CAS we force those assignments on the first step, but given that the workflow here will be more flexible, that may not be possible.  I can see having some administrative permission would be helpful, but being able to move the workflow forward is different from being alerted of the status changes.

drivers/hmis/app/models/hmis/ce/client_match/candidate.rb
```diff
@@ -0,0 +1,41 @@
+# an eligible, prioritized client for a given policy
+module Hmis::Ce::ClientMatch
```
eanders: One take away I have from CAS where we called this ClientOpportunityMatch, is that, we just call them matches.  Can we rename this module to `Match` or is that problematic?  I doubt we'll ever match anything other than a client.

ttoomey: Thanks for the suggestion. We can rename the module; it should be easy at this stage (`Hmis::Ce::Match::Policy, Hmis::Ce::Match::Candidate, etc).

drivers/hmis/app/models/hmis/ce/client_match/sql_expression_translator.rb
```diff
@@ -0,0 +1,145 @@
+require 'dentaku'
+require 'dentaku/visitor/infix'
+
+module Hmis::Ce::ClientMatch
```
ttoomey: This class is a bit complicated and may contains bugs; it's a bit of a proof-of-concept

drivers/hmis/app/models/hmis/ce/client_match/candidate.rb
ttoomey: renamed

ttoomey:
Pushed up support for cascading eligibility/priority config, bug fixes, and tests. Updated docs in PR. Rebased to 150

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
```diff
@@ -0,0 +1,88 @@
...
+    # create_table(:ce_match_pool_requirements) do |t|
+    #   t.references :candidate_pool, null: false, foreign_key: { to_table: :ce_match_policies }, index: false
+    #   t.references :requirement, null: false, foreign_key: { to_table: :ce_match_requirements}, index: false
+    #   t.timestamps
+    # end
+
+    create_table(:ce_match_candidates) do |t|
+      t.references :candidate_pool, null: false, foreign_key: { to_table: :ce_match_candidate_pools }, index: false
+      t.references :client, null: false, foreign_key: { to_table: :Client }
+      t.integer :priority_score, null: true
```
ttoomey, line 37: note, this probably should be multi-valued to support lexicographical priority order. TBD

eanders: I think integers are probably fine within the context of a given pool.  There will be complexity in calculating the priority, but once it is calculated, the order can be stored as an integer.

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
```diff
@@ -0,0 +1,88 @@
...
+      t.timestamps
+      t.index [:candidate_pool_id, :client_id], unique: true, name: 'index_ce_match_candidates_uniq'
+    end
+
+    # Opportunities (vacancies or services within projects)
+    create_table(:ce_opportunities) do |t|
+      t.references :candidate_pool, null: true, foreign_key: { to_table: :ce_match_candidate_pools }
+      t.references :project, null: false # what project provides this opportunity?
```
eanders: Maybe these would be helpful in the DB?
```suggestion
      t.references :project, null: false, comment: 'Project providing this opportunity'
```

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
```diff
@@ -0,0 +1,88 @@
...
+      t.references :swimlane, null: true, foreign_key: { to_table: :wfd_swimlanes }
+      t.timestamps
+    end
+
+    create_table(:ce_referral_notes) do |t|
+      t.references :referral, null: false, foreign_key: { to_table: :ce_referrals }
+      t.references :participant, null: false, foreign_key: { to_table: :ce_referral_participants }
+      t.jsonb :submitted_form_data
+      t.timestamps
+    end
```
eanders: did we decide we needed these at the referral level or only the step level? (I thought step.)

drivers/hmis/app/graphql/mutations/ce_opportunity_create.rb
```diff
@@ -0,0 +1,27 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
eanders: These headers are missing a line:
```
###
#
###
```

We use an exact match when updating it year over year. See `lib/util/code.rb`.

Alternatively you can just leave this off and add it later with `rails code:maintain_copyright`

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeOpportunity < Types::BaseObject
+    field :id, ID, null: false
+    field :name, String, null: false
```
eanders, line 12: I'm guessing these won't have names themselves, or at least not one that is often set by the person creating the opportunity.

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -0,0 +1,29 @@
...
+    def candidates
+      Hmis::Ce::Match::Candidate.
+        for_opportunity(object).
+        order(priority_score: :desc, client_id: :desc).
+        limit(50) # FIXME: add pagination. Just limit to top 50 for now
+    end
+
+    def active_referral
+      object.referrals.order(:id).viewable_by(current_user).active.first
+    end
```
eanders: Is this worth building to support multiple active referrals per opportunity now?  This could just be:
```suggestion
    def active_referrals
      object.referrals.order(:id).viewable_by(current_user).active
    end
```

drivers/hmis/app/models/hmis/ce.rb
```diff
@@ -0,0 +1,10 @@
+module Hmis::Ce
+  def self.table_name_prefix
+    'ce_'
+  end
+
+  def self.enabled?
+    # perhaps this should be an env or db config flag
+    Rails.env.development? || Rails.env.test?
+  end
```
eanders: I would like to see this as a database config.  It should probably be stored on the data source (or related to the data source.)
I also think we should rework `GrdaWarehouse::Config` to be more of a key value store at some point, not sure we want to tackle that change now.

drivers/hmis/app/models/hmis/ce/match/applicability_context.rb
```diff
@@ -0,0 +1,27 @@
+module Hmis::Ce::Match
+  OpportunityApplicability = Struct.new(:owner, :project_types, :project_funders, keyword_init: true) do
```
eanders: Can you add some notes about what this is for and doing?

drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # FIXME: permissions
+    scope :viewable_by, ->(_user) { all }
+
+    # order by descending priority, NULL values last. Use id as a tie-breaker
+    scope :prioritized, -> {
+      order(
+        Arel::Nodes::Descending.new(arel_table[:priority_score]).nulls_last,
+        arel_table[:id].asc,
+      )
```
eanders: I'm probably missing something, but can't this just be:
```suggestion
      order(
        arel_table[:priority_score].desc.nulls_last,
        arel_table[:id].asc,
      )
```

drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    # Which candidates (clients) are eligible for an opportunity
+    # - Filter out candidates that have already been referred to this opportunity.
+    # - Filter out candidates with an active referral to another opportunity with overlapping categories.
+    scope :for_opportunity, ->(opportunity) {
+      return Hmis::Ce::Opportunity.none unless opportunity.candidate_pool
+
+      scope = opportunity.candidate_pool.candidates
+
+      # do we need to allow a referral to be re-started for the same client/opportunity?
```
eanders, line 29: We probably need to allow a re-referral.  (Someone starts with an opportunity, disappears, the referral is closed, someone else get referred, fails, and the original person shows up again with the highest priority)

drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -0,0 +1,41 @@
...
+      scope = scope.where.not(client_id: opportunity.referrals.select(:client_id))
+
+      # clients with active referrals to other opportunities who's categories overlap with this opportunity
+      exclude_client_ids = Hmis::Ce::Referral.active.
+        joins(opportunity: :categories).
+        where.not(opportunity: { id: opportunity.id }). # not this opportunity
+        where(categories: { id: opportunity.categories.select(:id) }). # overlapping categories
+        distinct.pluck(:client_id)
```
eanders, line 37: I _think_ we should implement this as part of the rules/requirements.  Basically, you can apply a rule "Can't be active in a referral of X category" to the opportunity.  Does that make this unnecessary?

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
```diff
@@ -0,0 +1,78 @@
+# update candidate pools to reflect the current opportunities, requirements and priority configuration
+
+module Hmis::Ce::Match
+  class CandidatePoolBuilder
+    # assumes a relatively small number of active opportunities
```
eanders: We should probably define "relatively small" (is it < 10, < 100, or < 1,000?)
Also, maybe a quick note about what might need to be adjusted to handle more.  I expect we'll quickly end up with > 100 opportunities.

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
```diff
@@ -0,0 +1,78 @@
...
+
+    def with_lock(&block)
+      lock_name = self.class.name.demodulize
+      GrdaWarehouseBase.with_advisory_lock(lock_name, timeout_seconds: 30, &block)
+    end
+
+    def _perform
+      grouped = {}
+
+      Hmis::Ce::Match::Rule.order(:owner_type, :id)
```
eanders: What is this doing?

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
```diff
@@ -0,0 +1,78 @@
...
+        on_duplicate_key_update: {
+          conflict_target: [:priority_expression, :requirement_expression],
+          columns: [:configuration_updated_at],
+        },
+      )
+      raise "Failed: #{result.failed_instances}" if result.failed_instances.present?
+
+      # delete pools that haven't been used in a while
+      Hmis::Ce::Match::CandidatePool.
+        where(configuration_updated_at: ...(now - 6.months)).
```
eanders: `6.months` should probably be configurable (eventually) maybe pull it into a constant for now?

drivers/hmis/app/models/hmis/ce/match/client_expression_evaluator.rb
```diff
@@ -0,0 +1,23 @@
...
+    def call(client)
+      calculator = Dentaku::Calculator.new
+
+      # construct client values for the expression.
+      # Note, this isn't lazy so it's less performant that it could be
+      client_values = calculator.dependencies(expression).to_h do |field|
+        [field, field_map.instance_value(client, field)]
+      end
+
+      calculator.evaluate!(expression, **client_values)
```
eanders: An example would be really helpful here.

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
ttoomey: great, that would be a helpful simplification

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
ttoomey: Good question, I remember looking at old mocks and it seemed like visually the notes were presented in a sidebar without a relationship to steps. It would be simpler without a step relationship but we should probably square up with design on this.

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
ttoomey: Good to know. I guess the name would be automatically created from the resource? I'm going to leave it for now until we figure out how to generate the name or if it should be something the user can set optionally

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
ttoomey: Oh sure, I thought that multiple active referrals was forbidden. Is it a known requirement that we need multiple?

drivers/hmis/app/models/hmis/ce.rb
ttoomey: Sure, a generic KV store for HMIS a while ago. The model is `AppConfigProperty`. I'll use that to build Hmis::Ce::Config

drivers/hmis/app/models/hmis/ce/match/applicability_context.rb
ttoomey: for sure, it's just a little single-responsibility service object

drivers/hmis/app/models/hmis/ce/match/candidate.rb
ttoomey: That's better, thanks!

drivers/hmis/app/models/hmis/ce/match/candidate.rb
ttoomey: good to know, maybe this can just exclude active referrals to this opportunity

drivers/hmis/app/models/hmis/ce/match/candidate.rb
ttoomey: We could do that however it might introduce some complexity around the referral lifecycle: after a referral is opened or closed, we'd have to re-evaluate all the matches for the referred client as it would impact their eligibility. I guess that probably wouldn't be that expensive as it's just one client but we'd have to handle the synchronization carefully.

Still, it seems more difficult to implement as a rule since it would require the rule evaluation to be aware of the opportunity's categories. Or we'd have to hard-code the category in the rule which seems potentially fragile. 

My thought with the current approach is that categories could have a simple flag, category.mutually_exclusive to control this behavior. For example we might have a "PH" category that is mutually exclusive.  Maybe we can keep this implementation for now?

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
ttoomey: I'll clarify the comment. This was more of an optimization note; the job is not optimized and has n+1 queries; this is proof-of-concept. Still nothing it's doing is particularly expensive. My intuition is that it could churn through 1000 records. We could see what the run-time is like; it probably could be optimized with minimal effort

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
ttoomey: oops, meant to pull this out of the loop. Thanks for catching! I feel like rubocop should yell at me for this

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
ttoomey: sure, I'll add it to the db-configuration

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
ttoomey: good idea, thanks

drivers/hmis/app/graphql/mutations/ce_opportunity_create.rb
ttoomey: thanks, I corrected it on the mutations

ttoomey:
Thanks for the helpful review @eanders! I think I've addressed everything, some questions remain open (which is expected I think, this is just a spike)

drivers/hmis/app/models/hmis/ce/configuration.rb
```diff
@@ -0,0 +1,41 @@
...
+
+    protected
+
+    # read all configuration values from the db
+    PROPERTIES = [
+      :enabled,
+      :days_to_retain_orphan_candidate_pools,
+    ].freeze
+    def values
+      @values ||= AppConfigProperty.
```
eanders, line 30: This is great.  I thought we had a solution but couldn't remember what it was.  
Only note I have is that CE will probably need to be turned on per-data source as we might have multiple HMISs backed by one warehouse.  Not sure if that should get integrated in the key or value, or somehow else.

drivers/hmis/app/models/hmis/ce/match/applicability_context.rb
```diff
@@ -0,0 +1,33 @@
+# Determines whether a CE rule applies to a given opportunity based on project criteria. This is NOT determining
+# client is eligibility itself; rather it is determine what rule configurations to use for eligibility/priority
```
eanders:
```suggestion
# client is eligibility itself; rather it determines what rule configurations to use for eligibility/priority
```

drivers/hmis/app/models/hmis/ce/match/field_map.rb
```diff
@@ -0,0 +1,47 @@
...
+          instance_value: lambda(&:veteran_status),
+          arel_field: arel.c_t['VeteranStatus'],
+        },
+        current_age: {
+          instance_value: ->(c) { c.age(current_date) },
+          arel_field: age_from('year', current_date, arel.c_t['DOB']),
+        },
+        # additional configuration....
+      }
+    end
```
eanders: Is the eventual idea that these would be loaded from the database?  
Are these something that could be defined in the UI, or would an engineer need to setup the structure for each rule?
As a proof of concept this all makes sense to me, but since these aren't fully self-contained and expressed in something like `json`, it feels like we might need to take another pass at how we structure these fields/rules.  
(Admittedly, I'm guessing at what this is being used for and is representing.)

ttoomey:
I forgot to press submit on my review, I intended to add the comments above yesterday

drivers/hmis/app/models/hmis/ce/configuration.rb
ttoomey: Oh okay, if it's configured per data source I'd be inclined to put the config flag on the data_source table instead. The AppConfigProperty model is helpful for global config but probably better to just the more specific models if possible

drivers/hmis/app/models/hmis/ce/match/field_map.rb
ttoomey: So these aren't rules themselves; rather these are defining the variables that can be used in a rule expression. So if my expression is 
'current_age >= 65 AND veteran_status = 1'

I need to somehow define what 'current_age' and 'veteran_status' are. That's what we're doing here.

My assumption is that the MVP set of variables is relatively small and well defined (what we already have represented in code).

We should discuss if we need data-driven variables that could be managed by the user. I think something like 'current-age' would be hard to represent as user-managed config. But maybe we could have a 'core' set of variables that we define programatically. Then we could define namespaces for custom variables that could be data driven. Then we could use those in an expression like:
'veteran_status = 1 AND custom.my_assessment_score > 90'

drivers/hmis/app/models/hmis/ce/configuration.rb
eanders: In thinking about this some more, I'm not sure we need it per data source.  I think the ability to enable/disable CE is probably more during the build-out than once it's ready.  Once it's ready, we can control access to the features via permissions or project setup.  This is probably fine as is.

ttoomey:
Note, discussed on step notes on design call today. Leaning towards not associating notes with specific steps for mvp.

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
gigxz: Confirmed that notes will be tied to the referral for MVP, not tied to an individual step

gigxz: I do wonder about the complexity of tying the note to a `ce_referral_participants` rather than directly to `users`. 2 thoughts on that:
* There may be some administrative user who isn't a part of any swimlane, but is still able to go in and leave a note (?)
* The user isn't necessarily making that note "in the capacity of" a swimlane, I don't think? Asked another way, if the user is a member of 2 swimlanes (assuming that's possible) how do you decide which ce_referral_participants to link to their note?

drivers/hmis/app/models/hmis/ce/match/candidate.rb
gigxz: I like the approach of category with a mutually_exclusive flag. I agree w keeping this implementation for now.



> We could do that however it might introduce some complexity around the referral lifecycle: after a referral is opened or closed, we'd have to re-evaluate all the matches for the referred client as it would impact their eligibility.

Just to poke at this a little. Eligibility can change over time. For example if an opportunity requires Veteran Status, a client referral is started, and then at some point during the referral that Client is marked as VeteranStatus:false. Similarly the eligibility requirements of the opportunity itself could change.

Do we want/need to deal with that more systemically? It feels non-MVP to me, but I can see it being useful to display as an alert on the referral, eventually. For example:
* "Client is no longer eligible [because they are not a veteran]"
* "Client is no longer eligible [because they are enrolled in a PH program now]"
* "Client is no longer eligible [because they have another active referral in the same mutually exclusive category]"
* "Client is no longer eligible [because the opportunity now requires physical disability]"

@eanders correct me if I'm wrong, I believe in CAS the eligibility is evaluated only at the initial match activation, so its a "snapshot" of the client attributes and oppt requirements at that moment; it is not re-evaluated as the match progresses. Is that right, if so has it posed issues?

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
gigxz: We marked this as "Not needed" in the requirements sheet during one of our requirements sessions. The wording is:

> When an Opportunity is available, a User can select several Clients to "activate" as a Referral for one Opportunity

Unfortunately the note is not very helpful, but I think we somehow determined that the customer that used similar functionality previously, would not need it under the new system(?).

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
eanders: I agree with @gigxz, the notes should probably be tied to the user.  We definitely see people who are operating in 2 swim lanes simultaneously.  (Maybe a CoC level and a housing level for instance.)

If we aren't tying a note to a step, we should probably have some loose coupling to indicate at which step the referral was when the note was left.  Maybe time stamp is enough, as long as we can identify where the referral was at any given time.  I guess I'd vote for tying them to the current step and then rolling them up in the UI for the referral.

drivers/hmis/app/models/hmis/ce/match/candidate.rb
eanders: Yes, CAS eligibility is only evaluate at match start.  This has been deemed a feature as it allows the program to adjust eligibility over time but grandfather existing matches under the eligibility requirements as they were at start.  We _do_ highlight that the eligibility requirements have changed, we just don't currently indicate if that would or would not make the client ineligible.  

We should definitely "save the universe" as we say when a referral is started so that you can identify when requirements have changed.  I don't think we need to expose the change post match start for MVP, just make sure we're storing the data so we could in the future.

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -0,0 +1,66 @@
+# An opportunity is the availability of a resource (housing or other services)
+
+module Hmis::Ce
+  class Opportunity < GrdaWarehouseBase
+    include AASM
+
+    belongs_to :project, class_name: 'Hmis::Hud::Project'
```
gigxz: add `ce_opportunities` relationship to project

drivers/hmis/app/graphql/mutations/ce_referral_create.rb
```diff
@@ -0,0 +1,43 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Mutations
+  class CeReferralCreate < CleanBaseMutation
+    argument :opportunity_id, ID, required: true
+    argument :client_id, ID, required: true
+    field :referral, Types::HmisSchema::CeReferral, null: false
+    argument :input, Types::HmisSchema::CeReferralInput, required: true
```
gigxz: move opportunity_id and client_id into CeReferralInput obj?

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -0,0 +1,151 @@
...
+
+    def start_step!(step, user:)
+      step.start!
+      process_triggers(step.node, 'start_step')
+      log_event('start_step', user: user, step: step)
+    end
+
+    def complete_step!(step, user:, submitted_values:)
+      step.submitted_values = submitted_values
+      step.complete!
```
gigxz, line 48: Is `complete!` implemented? I don't see it but maybe there is rails magic with the statuses?

drivers/hmis/app/graphql/mutations/ce_referral_step_submit.rb
```diff
@@ -0,0 +1,27 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Mutations
+  class CeReferralStepSubmit < CleanBaseMutation
```
gigxz: As discussed there is a need for a "validate" step that dry-runs the submission and notifies the user of the outcome/side-effects.

Either:
* Separate mutation `CeReferralStepValidate` or,
* Use  `CeReferralStepSubmit` pass flag `confirmed:` to match existing form submission pattern. (If `confirmed:false`, then the mutation returns validation warnings about what will happen (if necessary). If `confirmed:true` then the mutation will perform the operation despite validation warnings).

Probably easier for all to use existing pattern (option 2), if it seems workable here.

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -0,0 +1,151 @@
...
+      end
+    end
+
+    def evaluate_condition(expression)
+      # empty expression defaults to true
+      return true if expression.blank?
+
+      calculator = Dentaku::Calculator.new
+      defaults = calculator.dependencies(expression).to_h { |k| [k.to_sym, nil] }
+      calculator.evaluate!(expression, **defaults.merge(all_submitted_values))
```
gigxz: Is this where the `submitted_values` get "processed" (e.g. to determine if the responses constitute a referral rejection)? Can you give an example of what an expression looks like? thanks!

gigxz (commented):
Leaving some more comments while spec'ing out issues for https://github.com/open-path/Green-River/issues/7151.

Overall this looks like an amazing start, if I hadn't said that already. 

Some of my comments are things that can be implemented during those tickets, by you or @martha  (like my comments about resolving different types on the  GQL api), I don't think those need to be addressed in this PR. Just wanted to capture them.

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
ttoomey: Thanks, those are good points. The notes are definitely less well formed in this spike
* I'll replace the participant association with a direct association to user
* I'm not sure "current step" is something that consistently exists. However we should have an audit event paper trail that would allow us to place the note in context. I think adding a optional step association could be something we could phase in pretty easily but suggest waiting until there is a UI that supports associating a note with a step

drivers/hmis/app/graphql/mutations/ce_referral_step_submit.rb
ttoomey: Thanks for the call out. I'd consider making it a query too since it's not really mutating anything but I'm sure it's fine either way

drivers/hmis/app/models/hmis/ce/match/candidate.rb
ttoomey: > We should definitely "save the universe" as we say when a referral is started so that you can identify when requirements have changed. I don't think we need to expose the change post match start for MVP, just make sure we're storing the data so we could in the future.

Snapshotting the entire universe could be tricky. Maybe there's some clever optimizations to be had to keep db space requirements manageable

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
ttoomey: It's magic from the `aasm` gem. Methods correspond to the events in the aasm config for this model:
```ruby
 # step.rb
 aasm column: 'status' do
   # status...
   event :complete do # defines `complete`!
     # ...
   end
 end
```

drivers/hmis/app/models/hmis/ce/match/candidate.rb
eanders: I believe the current version does something like save a json blob of the rules (or rule ids) as they exist when the match starts, and the ids of the clients who match in the priority order.  (Well I guess a bit more.)

https://github.com/greenriver/boston-cas/blob/cd976569563564e38d718646c807499f8352d628/app/controllers/opportunity_matches_controller.rb#L60-L76

drivers/hmis/app/graphql/mutations/ce_referral_create.rb
ttoomey: Let's see what the ui is like before moving these around. Maybe CeReferralInput can match the form fields

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
gigxz: I see, thanks

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -0,0 +1,45 @@
+# A referral of an individual client to an opportunity
+module Hmis::Ce
+  class Referral < GrdaWarehouseBase
+    include AASM
+    belongs_to :opportunity, class_name: 'Hmis::Ce::Opportunity'
+    belongs_to :workflow_instance, class_name: 'Hmis::WorkflowExecution::Instance'
+    has_many :notes, class_name: 'Hmis::Ce::ReferralNote'
+    has_many :participants, class_name: 'Hmis::Ce::ReferralParticipant'
+    belongs_to :client, class_name: 'Hmis::Hud::Client'
```
gigxz, line 18: @ttoomey @eanders 
This data model currently restricts the referral to an individual client, not a household. I think the requirements around this have been a little unclear, I've been meaning to circle back on it. I believe this question is linked to the "manual referral" concept, so I'm putting that into words finally, too.

Can we do a check-in on this at some point next week? I'll put my thoughts below.

From my understanding, we do need:
1) Some way to "manually refer" a **household** from one project to another.
   * UX-wise, this should look very similar to existing referral creation (screenshot below).
   * We have discussed that a possible relationship between the "manual referral" and CE referrals may be: "the [opportunity] may have a requirement specifying that only clients who have been 'manually referred' to this project are eligible"
   * This approach should support many existing (non-housing) workflows, where households can be directly referred from one project to another. In such setups, we expect the Referral Workflow to be a simple one-step workflow (approve or deny). The sending project should have some way to track the outcome of its outgoing referral (like current implementation).
   * This workflow does not necessarily include a housing assessment of any kind.
   * If the receiving project or opportunity has household-related requirements (eg "families with children only", bed room size, etc) then those could be interpreted based on the household, OR you could require the referrer to fill them out in a form (see screenshot at the end)? TBD 🔦
    * Captured in https://github.com/open-path/Green-River/issues/7321
    * **Rough Data Model concept for discussion ?**
      * Source project `Enrollment` has 0-N `OutgoingReferral`s. `OutgoingReferral { source_enrollment_id, destination_project_id, referred_at, resulting_ce_referral_id }`
      * resulting_ce_referral is an optional reference to the `Ce::Referral` that was created when this client's outgoing referral was "activated" into a real ce referral. 
      * `Ce::Referral` has a relationship to 0-N OutgoingReferral that led to it. We use that to understand the household and any data that was sent along with the referral. I say 0-N because the HoH Client could conceivably been referred to the same Project by multiple different sending projects, possibly with different household compositions. Need to dig into that.. when you "activate" the client are you picking one of the referrals..?


2) Some way for a referred individual client to end up in a multi-member household in the receiving project. TBD whether that happens during the course of the enrollment ("add members to this referral"?), or _after_ the destination Enrollment is created (receiving project uses the existing "manage household" workflow).
    * The size and composition of the household may affect which Opportunities they are eligible for. In CAS this works by setting eligibility requirements based on questions that may be asked on a housing assessment (for example "How many members in your household?" "Any children in your household?" "How many bedrooms do you need?"). **Do we want to carry this approach forward, or is there a better way? Are there certain household-related attributes we should elevate to "core" attributes that can be used on housing assessments and/or referral forms, so we enforce some consistency?**. 
    * It's probably simplest to keep the CAS approach, and to just let the receiving project add any household members once they're enrolled. The "previously associated members" should make it easier for them to find those existing household members, provided they have enough access. I just wanted to raise this to reach an explicit decision on it.

---


Screenshot 2025-02-14 at 11 18 32 AM

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
```diff
@@ -0,0 +1,88 @@
...
+
+    # Referrals (instances of clients going through workflow)
+    create_table(:ce_referrals) do |t|
+      t.references :opportunity, null: false, foreign_key: { to_table: :ce_opportunities }
+      t.references :workflow_instance, null: false, foreign_key: { to_table: :wfe_instances }
+      t.string :status, null: false
+      t.references :client, null: false
+      t.references :referred_by
+      t.datetime :completed_at
+      t.timestamps
```
gigxz, line 74: I think we need a nullable reference to the enrollment at the receiving project (IF this referral ended up being accepted, and client enrolled in the receiving project). At some point during the referral the client may become enrolled as a side-effect of this referral, and we should store that reference, so we can display the status of the Enrollment (And likewise on the Enrollment we can display information about the referral that led to the enrollment).

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: I created a fig jam board to review the manual referral idea. Maybe devs can discuss at standup on Monday, and I can bring to design on Tuesday session.

https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-HMIS-%2F-Coordinated-Entry-%2F-FigJam?node-id=2351-3657

gigxz: To summarize more recent conversation with @ttoomey, I think we landed on going in the following direction:
* **Direct/Service Referrals** should generate a `Hmis::Ce::Referral` directly. It could either be in "unstarted" state, or maybe the sending project would fill out an initial "step" containing the data that they're sending over.
  * The receiving project would see the referral appear in their "referral queue" and take action. 
  * The sending project would see the referral in their "outgoing referrals" (at project or enrollment level) and possibly be able to track the status or outcome of that referral.
  * Referrals generated in this way should have a reference to the "Sending project"
  * Referrals generated in this way should have a "snapshot" of the household, as it was the moment the referral was created (each member's `client_id` and `relationship_to_hoh`). We probably do also want to store the `enrollment_id` or `household_id` that was originally sent, so that we can track on the household that it had an "outgoing referral". However it's not necessary/desired that the household composition is updated dynamically on the referral when the "sending" household's composition changes. (We think).
* **Housing Referrals** are created by someone "activating" the top eligible prioritized **client** for a given opportunity.  These are necessarily focused on one Client, however the related client attributes (such as housing assessment responses) likely indicate the household size and composition. (a.k.a. same approach as CAS).
* `Opportunity` records are single-use, but they can be tied to Unit or some other eventual capacity tracking record.

Note: I'm using the terms "housing referral" and "service referral" as a paradigm to think about the different between these 2 inroads to referral creation, but really either one could be used for services or housing (barring types of housing which require certain eligibility/prioritization protocols).


@ttoomey let me know if that tracks with your memory of our conversation! sorry for the belatedness of the summary. I do think this calls for some new columns/tables (i.e. household membership, project reference), but those changes are additive to the current tables, so we could either add them now or wait to add them until we tackle https://github.com/open-path/Green-River/issues/7321.

ttoomey: Hi @gigxz - that's accurate, thank you for summarizing!

> Direct/Service Referrals should generate a Hmis::Ce::Referral directly

Yes, I imagine there would be configuration on opportunity or project that determines if the opportunity requires "direct referral" or "prioritized referral". We probably would skip generating candidate lists for direct-referral opportunities, assuming that direct referrals would never consider eligibility.

>  I do think this calls for some new columns/tables (i.e. household membership, project reference)

Regarding capturing additional enrollment information on the referral, we might want to consider storing some of it in the form step data rather than creating new data structures. That would make it easier for users to make corrections to that household information using existing forms. It also makes those fields directly available for use in conditional expressions within the workflow. I agree that sending project should be a nullable association on referral, set on creation.

> Opportunity records are single-use, but they can be tied to Unit or some other eventual capacity tracking record.

Just to clarify, opportunities are intended to be single use but they are considered 'used' once a referral is completed. When a referral fails the same opportunity is still available for the next eligible candidate to start a new referral (essentially each opportunity may have many referrals but only one completed referral; new referrals cannot be created against an opportunity if it has any active or completed referrals)

ttoomey:
Discussed on standup today, we'd like to merge but maybe needs a second layer of enforcement to ensure this can't be accidentally enabled outside of QA/staging; the visibility scopes are not yet constrained

db/warehouse/migrate/20241207185501_create_ce_referrals.rb
martha: Closing the loop - this is now implemented as `target_enrollment` reference on the referral. `db/warehouse/migrate/20250325205806_add_target_enrollment_to_referral.rb`

db/warehouse_structure.sql
```diff
@@ -278,7 +278,7 @@ CREATE FUNCTION public.service_history_service_insert_trigger() RETURNS trigger
             INSERT INTO service_history_services_2001 VALUES (NEW.*);
          ELSIF  ( NEW.date BETWEEN DATE '2000-01-01' AND DATE '2000-12-31' ) THEN
             INSERT INTO service_history_services_2000 VALUES (NEW.*);
-
```
martha: @ttoomey following up on your comment here https://github.com/greenriver/hmis-warehouse/pull/5244#pullrequestreview-2728215691, maybe the easiest resolution is for you to run a migration in your local branch and push it up here before we merge this into the release branch?

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -0,0 +1,337 @@
+# Helpers to reduce cruft and make the starter pack script more readable
```
martha, line 1: Do we want to commit this starter pack? I'm happy to if it's helpful, but also happy to take it out if we don't want it on the release branch.

I do still use it during development, so maybe it would be helpful to remain.

martha (approved):

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -83,6 +83,16 @@ def self.authorized?(object, ctx)
...
+    field :ce_opportunities, [Types::HmisSchema::CeOpportunity], null: false
+
+    def ce_opportunities
+      raise unless Hmis::Ce.configuration.enabled?
+
+      Hmis::Ce::Opportunity.
+        for_client(object).
+        order(:id).
+        limit(50) # FIXME: add pagination. Just limit to top 50 for now
+    end
```
gigxz: We will probably want this to be `field :eligible_ce_opportunities, Types::HmisSchema::CeOpportunity.page_type, null: false` to support the client dashboard eligible opps page. Not sure what it was intended for but I don't think we have another use for resolving opps on client.

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -208,7 +213,7 @@ def self.source_for(destination_id:, user:)
   end
 
   def ssn_serial
-    self.SSN&.[](-4..-1)
+    self.SSN&.[](-4..)
```
gigxz: Suggest reverting this unrelated change? Is there a bug?

martha: Not a bug but a change to make rubocop happy. See https://github.com/greenriver/hmis-warehouse/actions/runs/14230396384/job/39879679171?pr=5013

I am ok either way if you want to revert it, I just updated it to make the build go green 💚

drivers/hmis/app/graphql/types/hmis_schema/client.rb
martha: Sounds good! I addressed that over here https://github.com/greenriver/hmis-warehouse/pull/5262

drivers/hmis/app/models/hmis/hud/client.rb
martha: Opened https://github.com/greenriver/hmis-warehouse/pull/5267 to separate this out

gigxz: Gotcha thanks!

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
gigxz: I think we can keep it for now, since we still don't have a mechanism in the UI to create most of these records

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
```diff
@@ -0,0 +1,87 @@
...
+class CreateHmisWorkflows < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wfd_templates do |t|
+      t.string :identifier, null: false
+      t.string :name, null: false
+      t.integer :version, null: false
+      t.string :status, null: false # retired, published, draft
+      t.text :description
+      t.references :owner, polymorphic: true # TBD
```
gigxz, line 12: @martha  re: your TODO below about templates being shared across data sources (https://github.com/open-path/Green-River/issues/7502). It's a good point, I don't think we want global wfd_templates shared across all data sources. 

@ttoomey, do you have a recollection on your intention for the `owner` column?

I am tempted to replace it with a non-nullable `data_source_id` to specify which data source this configuration belongs to. We have outstanding work in https://github.com/open-path/Green-River/issues/6612 to add data_source_id to other configuration tables (like form_definitions).


Somewhat relatedly, we also need a way to specify where each workflow template can be used (eg setting a default template for a Project). I am quite sure we will need another table for specifying those relationships (since a single template can be specified for multiple projects), more on that here https://github.com/open-path/Green-River/issues/7522.

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
```diff
@@ -0,0 +1,87 @@
+# frozen_string_literal: true
+
+class CreateHmisWorkflows < ActiveRecord::Migration[7.0]
+  def change
+    create_table :wfd_templates do |t|
```
gigxz, line 5: I am unsure of how we will differentiate between "CE" WorkflowDefinitions vs other kinds of WorkflowDefinitions (if/when we use these general wfd tables to support other types of workflows). Perhaps we need an additional "type" column for the workflow type that we always populate with `ce_referral` for now? @ttoomey thoughts?

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -0,0 +1,99 @@
...
+      referral.accept!
+      referral.opportunity.close!
+      # TBD enroll client, set move-in date, assign to unit if needed
+      # enrollment = referral.target_enrollment
+      # if enrollment.nil?
+      #   referral.target_enrollment ||= referral.project.enrollments.create!(client: referral.client)
+      #   referral.update!(target_enrollment: enrollment)
+      # else
+      #   enrollment.mark-non-wip
+      # end
```
gigxz: I think we can remove this since the `create_enrollment` and `set_move_in_date` message types handle this behavior. Those need to be specified explicitly if they should occur. I think that is correct, because there are some cases where you may never want to enroll the client even if the referral is accepted. (For example, a non-HMIS Participating project that is used to track referrals but does not actually enroll clients).

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
ttoomey: Yes, it seems reasonable to add a type field in anticipation of supporting other workflow applications.

Eventually we a may need some kind of cross-project configuration that sets the default templates for new opportunities. That might obviate the need to select the template when creating the opportunity.

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
ttoomey: I think my intention was that owner was the stand-in for whatever entity manages the template. If we think templates are only ever managed at the data-source level, than replacing owner with data_source makes sense. But if definitions might be managed at the project or org level, we might want to leave it as a polymorphic. For example if we need to support permissions so a user can manage workflows at only a subset of projects.

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
```diff
@@ -0,0 +1,87 @@
...
+    end
+
+    create_table :wfd_nodes do |t|
+      t.references :template, null: false, foreign_key: { to_table: :wfd_templates }
+      t.string :type, null: false # For STI
+      t.jsonb :trigger_config # when to send notifications, create ce events, state changes, api calls
+      t.string :name
+      t.references :swimlane, foreign_key: { to_table: :wfd_swimlanes }
+      # task nodes have forms.
+      t.references :form_definition, foreign_key: { to_table: :hmis_form_definitions }
```
gigxz: I can't find the comment now, but @martha  pointed out somewhere that because this points to a specific `form_definition_id`, it is linked to a particular "version" of the form. That means that if a customer wanted to change the form content for a form that's referenced by a workflow definition, they could publish that change, but it wouldn't go "live" until the relevant record in wfd_nodes was somehow updated to point to the newly published version of the form. (which we dont have a mechanism to do).

That's not typically how we do things - Usually we would specify the form definition `identifier` that should be used, and then when a form is taken we store the `form_definition_id` that was used alongside the data collected (eg on the form_processor, in this case it would make sense to store it on `wfe_steps` alongside `submitted_values`.

Unless anyone has objection to that, I would suggest:
- replace this reference with a form_definition_identifier string
- add form_definition_id reference to the wfe_steps table
- adjust belongs_to relationship on `Hmis::WorkflowDefinition::Task` to be a has_many definitions through identifier (similar to `Hmis::Form::Instance`)
- adjust `CeReferralStep` resolver to load the locally referenced form_definition_id if present, otherwise load the published version of the form

edit found it: https://github.com/open-path/Green-River/issues/7414

gigxz: It might be easier/less churn to make those changes on this branch, so we have less complicated migrations being run in prod!

created https://github.com/open-path/Green-River/issues/7526 just to track it

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
gigxz: Thanks Theron, That makes sense! better to leave it flexible, even if we expect it will only be data sources for now.

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
martha: Sounds good, I removed this in https://github.com/greenriver/hmis-warehouse/pull/5280

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
martha: Opened https://github.com/greenriver/hmis-warehouse/pull/5280

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
martha: It sounds like the followups here are covered by outstanding tickets and there's nothing blocking this PR getting merged, is that right?

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
martha: Opened https://github.com/greenriver/hmis-warehouse/pull/5284

Gemfile
```diff
@@ -245,7 +245,8 @@ gem 'ansi'
 gem 'parallel'
 gem 'todo_or_die'
 
-gem 'business_time'
+gem 'dentaku', require: false
```
gigxz, line 248:
```suggestion
gem 'dentaku'
```

Whats the benefit of `require: false` here @ttoomey ? It wasn't installing for me by default so I was getting errors in the CE workflow, I had to install manually to get it working. Not sure how it will work when deployed.

ttoomey: Strange that is wasn't installing for you. The `require` arg just causes it not to be loaded by default, so you to `require` the gem explicitly.

This saves a little memory on installations that might never need the gem or use CE. But it's probably so marginal that it's not worth the hassle; it should be fine to remove the `require` directive.

gigxz: Ah ok- there must have been something weird going on locally, its working for me now!

gigxz (approved):
wooot 👏 💯  the serious beginnings of a fab coordinated entry tool! nice work @ttoomey and @martha 

(pending tests passing)

db/warehouse_structure.sql
martha: Never mind! After Gig's tip of running `RAILS_ENV=test rails db:reset` locally, I reran migrations and pushed the changes to `warehouse_structure.sql`. The diff looks correct to me now -- it contains only additions related to CE/workflows, no spurious changes. I think we are all good here 👍

PR #5018, opened 2024-12-24
Don't raise on picklist query if args missing

gigxz:
## _Merging this PR_

## Description

These were causing problems in the form builder previews, because we don't have project/client/enrollment IDs on hand to pass to the pick list query. This PR changes the behavior so instead of raising, it just returns an empty option list.

This _may_ make it more likely for bugs to sneak by in the future, if we were to introduce a form that didn't correctly pass "pick list query args" , it could result in empty lists. We'll have to rely on QA and bug reports for those, rather than Sentry errors.

Other alternatives considered:
* Prevent the frontend Form Preview from fetching remote picklists entirely. This feels like it would be a reasonable solution, but has more code churn (with the current form implementation) since we need to pass a new prop through a bunch of components DynamicForm/View=>renderFields=>DynamicViewFields/DynamicFormFields=>DynamicViewField/DynamicField=>usePickList 

Issue: https://github.com/open-path/Green-River/issues/6783
Related issue: https://github.com/open-path/Green-River/issues/7093 ( need in order to work with those forms in the form buidler)
Related (but not dependent) frontend PR: https://github.com/greenriver/hmis-frontend/pull/1009

## How to test

Seed `unit_assignment.json` form from repo and open it in form builder preview. It should appear without throwing an error.

## Type of change


gigxz:
Test failure is unrelated - https://github.com/greenriver/hmis-warehouse/pull/5021

martha (approved):
Makes sense to me. Thank you for thinking it through.

ttoomey (approved):

PR #5020, opened 2024-12-29
Disable pg hero stats

ttoomey:
## _Merging this PR_

## Description
Disable's pghero's broken stats job. [Resolves sentry issue](https://green-river.sentry.io/issues/6113108363/events/1f600bc5f25047a3a18d9acd438f8b24/).

I'm leaving the job in place for now as it seem like could be a useful tool for future db optimization work

## Type of change
Housekeeping 

### DIFF
--- a/lib/tasks/grda_warehouse.rake
+++ b/lib/tasks/grda_warehouse.rake
@@ -355,7 +355,9 @@ namespace :grda_warehouse do
 
     BuildTranslationCacheJob.perform_later
 
-    PgheroCollectStatsJob.perform_later(clean: DateTime.current.hour == 5) if !Rails.env.production? && PgHero.query_stats_enabled?
+    # Disabled pg-hero status job for now. This doesn't have the required permissions
+    # to run in RDS. Note, pg-hero still works without it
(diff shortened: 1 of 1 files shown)


eanders (approved):
@ttoomey this looks good, thank you. 

Is there any way to wrap this in a  "do I have permission?" so we could leave it, or is that too complicated/not worth the effort?

ttoomey:
> Is there any way to wrap this in a "do I have permission?" so we could leave it, or is that too complicated/not worth the effort?

Yeah I did adjust `PgheroCollectStatsJob` to try and handle missing permissions gracefully, I thought it was working but seems like it broke somehow. I didn't want to spend more time to fix it since I'm not sure we are making much use of pghero right now. Maybe we can revisit when we do our next round of performance optimizations

eanders:
> Yeah I did adjust `PgheroCollectStatsJob` to try and handle missing permissions gracefully, I thought it was working but seems like it broke somehow. I didn't want to spend more time to fix it since I'm not sure we are making much use of pghero right now. Maybe we can revisit when we do our next round of performance optimizations

Sounds good.  Let's proceed as is.

PR #5022, opened 2024-12-29
Change handling of legacy occurrence point data

gigxz:
## _Merging this PR_

## Description

Issue: https://github.com/open-path/Green-River/issues/7236

This is walking back a previous assumption which said "you would only have one Occurrence Point form per field you're collecting". So only ONE move-in form, ONE unit assignment form, etc. If you want to customize it it can be done with custom_rule values.

Proposed approach is to only show the enabled forms, PLUS special logic to show Move-in Date, DOE, and PATH status if there is data that would otherwise be hidden.

"Bug" Reproduction:
* Add and publish second form that collects Move-in Date. Don't add any rules to enable it.
* **Previous Behavior** Move-in date is duplicated on all Enrollment Dashboards. It shows up once per form.
* **New Behavior** the new move-in date form doesn't appear unless specified

Screenshot 2024-12-29 at 2 36 30 PM

Scenarios requiring this change:
* Multiple move-in forms (hypothetical)
  * Project type X collects Move-in Date using form A which has a required move-in address on it
  * Project type Y collects Move-in Date using form B which does not have move-in address on it
* Multiple occurrence point forms that collect/display `current_unit` (https://github.com/open-path/Green-River/issues/7093)
  * One form is read-only, used for most projects
  * One form allows anyone to edit, used for certain projects

## Type of change
Bug fix


drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -428,5 +403,49 @@
...
+      let(:legacy_expected_struct) do
+        have_attributes(
+          legacy: true,
+          definition: have_attributes(identifier: 'date_of_engagement'), # default form seeded by JsonForms.seed_all
+          data_collected_about: 'ALL_CLIENTS',
+        )
+      end
+
+      context 'when DateOfEngagement does not exist' do
+        it 'does not return PATH status form' do
```
gigxz, line 439:
```suggestion
        it 'does not return Date of Engagement form' do
```

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -428,5 +403,49 @@
...
+          expect(hoh_enrollment.occurrence_point_forms).to be_empty
+        end
+      end
+      context 'when DateOfEngagement exists' do
+        before(:each) { hoh_enrollment.update!(date_of_engagement: 3.weeks.ago) }
+        it 'returns the default PATH status form' do
```
gigxz, line 445:
```suggestion
        it 'returns the default Date of Engagement status form' do
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -339,49 +339,51 @@ def data_collection_features
...
-        return true if cdeds_this_enrollment_has.include?(custom_field_key)
-      end
+    # Add legacy forms to ensure that HUD Data Elements are not hidden.
+    # In the event that an Enrollment has a MoveInDate, for example, but there is no active form that collects it,
+    # we still need to show it so that user can see the data and perform data correction.
+    [
+      # [<Enrollment field>, <default form identifier that should be used if field has any data>]
+      [:move_in_date, 'move_in_date'],
+      [:date_of_engagement, 'date_of_engagement'],
+      [:date_of_path_status, 'path_status'],
```
gigxz, line 368:
```suggestion
      # [<Enrollment field>, <default form identifier that should be used if field has any data>]
      [:move_in_date, 'move_in_date'], # identifier matches file name of the default form, e.g. drivers/hmis/lib/form_data/default/occurrence_point_forms/move_in_date.json
      [:date_of_engagement, 'date_of_engagement'],
      [:date_of_path_status, 'path_status'],
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -339,49 +339,51 @@ def data_collection_features
     end.compact
   end
 
+  # Occurrence Point Forms that are enabled for this Enrollment.
+  # Returns array of OpenStructs, which are resolved by the HmisSchema::OccurrencePointForm GQL type.
   def occurrence_point_forms
```
ttoomey, line 344: This is fine; it's consistent with how the other models. But the Enrollment class is getting kind of big (550 lines). I wonder if this might be an opportunity to extract this into some kind of single-purpose service class
```ruby
class Enrollment 
  #...
  def occurrence_point_forms
    OccurrencePointFormCollection.new(self).all
  end
end

class OccurrencePointFormCollection
  def initialize(enrollment)
    @enrollment = enrollment
  end

  def all =  active_forms + legacy_forms

  def active_forms
    definitions = Hmis::Form::Definition
      .with_role(:OCCURRENCE_POINT)
      .published_or_retired
      .latest_versions
    # definitions.map ...
  end

  def legacy_forms
    #...
  end
end

**Review by ttoomey — COMMENTED** [review]
looks reasonable to me, thanks!

File: drivers/hmis/app/models/hmis/hud/enrollment.rb [hunk 5]
```diff
@@ -339,49 +339,51 @@ def data_collection_features
...
-
-  private def legacy_occurrence_point_data?(definition, cdeds_this_enrollment_has)
-    definition.walk_definition_nodes(as_open_struct: true) do |item|
-      next unless item.mapping.present?
 
-      record_type = item.mapping&.record_type
-      field_name = item.mapping&.field_name&.underscore
-      custom_field_key = item.mapping&.custom_field_key
-
-      next unless record_type == 'ENROLLMENT' || custom_field_key
```
martha, line 373: Perhaps we don't need to remove all of this logic, but can instead change this (old line 373) to `next unless custom_field_key` and remove lines 375-378? If I'm understanding correctly, occurrence point forms can collect custom data elements. And we've mostly taken the approach of having forms create their own CDEDs instead of sharing them across forms. So, we don't expect the bug impacting move-in date, path status, and date of engagement to impact CDEs.

gigxz: thanks @martha. I am opting to leave CDEs out of it for the sake of reducing complexity

drivers/hmis/app/models/hmis/hud/enrollment.rb
gigxz: Thanks for the suggestion and snippet @ttoomey, I updated this to use a class. Both the Enrollment and Project occurrence point form function logic are moved into it

drivers/hmis/app/models/hmis/form/occurrence_point_form_collection.rb
```diff
@@ -0,0 +1,109 @@
...
+      data_collected_about: data_collected_about || 'ALL_CLIENTS',
+    )
+  end
+
+  # Check if the given FormDefinition collects the given field from the Enrollment.
+  # This is a bit hacky (transforming fieldname to graphql casing) but it works for the known fields (Move-in date, DOE, PATH).
+  def collects_enrollment_field?(definition, field_name)
+    definition.link_id_item_hash.values.find do |item|
+      item.mapping&.record_type == 'ENROLLMENT' && item.mapping&.field_name == field_name.to_s.camelize(:lower)
+    end.present?
```
ttoomey: maybe `any?` is a little clearer
```suggestion
    normalized_field_name = field_name.to_s.camelize(:lower)
    definition.link_id_item_hash.values.any? do |item|
      item.mapping&.record_type == 'ENROLLMENT' && item.mapping&.field_name == normalized_field_name
    end
```

drivers/hmis/app/models/hmis/form/occurrence_point_form_collection.rb
```diff
@@ -0,0 +1,109 @@
...
+#
+# This class is responsible for determining which Occurrence Point forms to display on a given Enrollment in HMIS.
+# The Occurrence Point forms appear in the "Enrollment Details" card on the HMIS Enrollment dashboard.
+#
+# These forms collect data elements onto an Enrollment "at occurrence" (a.k.a. when they occur),
+# as opposed to data elements that are collected at a specific point in time (e.g. at intake, exit).
+###
+class Hmis::Form::OccurrencePointFormCollection
+  # Struct that backs Types::HmisSchema::OccurrencePointForm
+  OccurrencePointForm = Struct.new(:definition, :legacy, :data_collected_about, keyword_init: true)
```
ttoomey, line 18: this looks like a private class, maybe signal that it's only instantiated internally?
```suggestion
  OccurrencePointForm = Struct.new(:definition, :legacy, :data_collected_about, keyword_init: true)
  private_constant :OccurrencePointForm
```

ttoomey (approved):
This looks great, thanks for the restructure!

drivers/hmis/app/models/hmis/form/occurrence_point_form_collection.rb
gigxz: nice thanks!

PR #5025, opened 2024-12-30
Use a job to enforce queing discipline for WarehouseClientsProcessed

eanders:
## _Merging this PR_

## Description
* Replaces `update_cached_counts_no_named_arguments` in `GrdaWarehouse::WarehouseClientsProcessed` with a full-fledged Job class.
* Enforces sequential-ish generating of cached warehouse client data by checking for a running job and re-queuing for a few minutes later if there is one 

To test:
1. Run: `GrdaWarehouse::WarehouseClientsProcessed.update_cached_counts` (where you have more than 5,000 destination clients in the database
2. Fire up 2 or more delayed job workers and watch the DJ queue. 
3. You should see a bunch of jobs created in step 1
4. You should see one job run and the remaining re-queued for a few minutes later, the process should repeat until no jobs remain

## Type of change
* Code clean-up


app/jobs/update_warehouse_clients_caches_job.rb
```diff
@@ -0,0 +1,39 @@
...
+  queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
+  WAIT_MINUTES = 4
+
+  def perform(client_ids: [], include_cas_and_cohorts: false, skip_expensive_calculations: false)
+    # If any for this class are already running, requeue for a few minutes in the future
+    any_jobs_of_type_other_than_self = Delayed::Job.jobs_for_class(self.class.name).
+      running.
+      where.not(Delayed::Job.arel_table[:handler].matches("%#{job_id}%")).
+      exists?
+    return requeue_job if any_jobs_of_type_other_than_self
```
eanders: @ttoomey this is the pattern we've used before, but I suppose this could just take out an advisory lock with the class name and then each run could attempt to open the lock, if it fails, it could re-queue.  Do you have a preference for pattern?

ttoomey: I would use a lock since it both prevents race-conditions and is less code. It's also nice to not have a dependency on DJ implementation internals. I guess I would expect any synchronization code to use a row or advisory lock but maybe there's a reason not to?

app/jobs/update_warehouse_clients_caches_job.rb
```diff
@@ -0,0 +1,39 @@
...
+
+    GrdaWarehouse::WarehouseClientsProcessed.update_cached_counts(client_ids: client_ids, include_cas_and_cohorts: include_cas_and_cohorts, skip_expensive_calculations: skip_expensive_calculations)
+  end
+
+  private def requeue_job
+    # Re-queue this job if there is another already running with the same class name.
+    # There is a potential for a race condition here and multiple may end up running at the same time if they are all picked up
+    # before any have been marked as running.  This is ok, we're just trying to limit it from using the entire DJ workforce
+    a_t = Delayed::Job.arel_table
+    job_object = Delayed::Job.where(a_t[:handler].matches("%job_id: #{job_id}%").or(a_t[:id].eq(job_id))).first
```
ttoomey: at the point where this runs, is job_object the current job, we're just duplicating it for a future run?

app/jobs/update_warehouse_clients_caches_job.rb
```diff
@@ -0,0 +1,39 @@
...
+    any_jobs_of_type_other_than_self = Delayed::Job.jobs_for_class(self.class.name).
+      running.
+      where.not(Delayed::Job.arel_table[:handler].matches("%#{job_id}%")).
+      exists?
+    return requeue_job if any_jobs_of_type_other_than_self
+
+    GrdaWarehouse::WarehouseClientsProcessed.update_cached_counts(client_ids: client_ids, include_cas_and_cohorts: include_cas_and_cohorts, skip_expensive_calculations: skip_expensive_calculations)
+  end
+
+  private def requeue_job
```
ttoomey: It would be nice to just use the built in `reschedule_at` but I'm betting it doesn't work us 
```ruby
def reschedule_at(time, _attempts)
  time + 4.minutes
end
```

ttoomey (approved):
looks like a good cleanup, thanks!

PR #5026, opened 2024-12-30
Move Advisory Locks to DJ for imports

eanders:
## _Merging this PR_

## Description
* This moves an advisory lock out of the importer and up to the calling delayed jobs. 
* It standardizes the delayed job structure for the two importers and enforces one run per data source regardless of which method is used to kick off the import
* Fixes a situation where imports could time out waiting for an advisory lock

## Type of change
* Bug fix
* Code clean-up

## Testing instructions
1. Upload more than one HMIS CSV Zip file to a single data source
2. Confirm the first runs and the subsequent ones are pushed out 15 minutes
3. Upload to another data source while the first is running and confirm it is picked up within a few minutes
4. Test S3 setup - Find a data source that is setup to import from S3 and set it to more than one file.
5. When it runs, the files should be imported sequentially (the first should finish before the second starts.)

### DIFF
--- a/app/jobs/base_job.rb
+++ b/app/jobs/base_job.rb
@@ -61,4 +61,32 @@ class BaseJob < ApplicationJob
       WorkerStatus.new(job).conditional_exit!
     end
   end
+
+  # attempts to requeue this job for a later time
+  # This is somewhat brittle at this time and expects to be operating on
+  # an ActiveJob instance (something like an instance of Importing::HudZip::HmisAutoMigrateJob).
+  # Additionally, this expects the rails job backend to be Delayed::Job
+  def requeue_at(timestamp, message)
+    Rails.logger.info(message) if message.present?
+    new_job = delayed_job.dup
+    new_job.update(
+      locked_at: nil,
+      locked_by: nil,
+      run_at: timestamp,
+      attempts: calculated_attempts,
+    )
+  end
+
+  # Attempt to find the associated delayed job so we can use it
+  def delayed_job
+    job = Delayed::Job.jobs_for_class(job_id).first
+    raise 'Unable to find a related delayed job' unless job.present?
+
+    job
+  end
+
+  # Override as necessary to limit the number of times a job is tried
+  def calculated_attempts
+    0
+  end
 end
--- a/app/jobs/base_job.rb
+++ b/app/jobs/base_job.rb
@@ -80,7 +80,8 @@ class BaseJob < ApplicationJob
   # Attempt to find the associated delayed job so we can use it
   def delayed_job
     job = Delayed::Job.jobs_for_class(job_id).first
-    raise 'Unable to find a related delayed job' unless job.present?
+    # NOTE: job_id will probably be a UUID in the handler of the row
+    raise "Unable to find a related delayed job (ID: #{job_id})" unless job.present?
 
     job
   end
--- a/app/jobs/importing/hud_zip/fetch_and_import_job.rb
+++ b/app/jobs/importing/hud_zip/fetch_and_import_job.rb
@@ -7,18 +7,28 @@
 module Importing::HudZip
   class FetchAndImportJob < BaseJob
     queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
-    WAIT_MINUTES = 4
+    WAIT_MINUTES = 15
 
-    def perform(klass:, options:)
-      safe_klass = known_classes.detect { |m| klass == m }
+    attr_accessor :job
+
+    def initialize(klass:, options:)
+      @klass = klass
+      @options = options
+    end
+
+    def before(job)
+      self.job = job
+    end
+
+    def perform
+      safe_klass = known_classes.detect { |m| @klass == m }
       raise "Unknown import class: #{klass}; You must add it to the list of known classes in FetchAndImportJob" unless safe_klass.present?
 
-      data_source_id = options.with_indifferent_access[:data_source_id]
-      lock_obtained = GrdaWarehouse::DataSource.with_advisory_lock("#{klass}_#{data_source_id}", timeout_seconds: 10) do
-        safe_klass.constantize.new(**options).import!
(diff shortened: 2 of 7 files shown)


app/jobs/importing/hud_zip/hmis_auto_migrate_job.rb
```diff
@@ -11,44 +11,45 @@ class HmisAutoMigrateJob < BaseJob
...
+      requeue_job unless lock_obtained
     end
 
     private def advisory_lock_name
-      "hud_import_auto_migrate_#{@data_source_id}"
+      GrdaWarehouse::DataSource.import_advisory_lock_name(data_source_id)
+    end
+
+    private def data_source_id
+      job.payload_object.instance_variable_get(:@data_source_id)
```
ttoomey: sorry I'm missing something. Why isn't this just @data_source_id?

app/jobs/importing/hud_zip/fetch_and_import_job.rb
```diff
@@ -7,22 +7,65 @@
...
 
-    def perform(klass:, options:)
-      safe_klass = known_classes.detect { |m| klass == m }
-      raise "Unknown import class: #{klass}; You must add it to the whitelist in FetchAndImportJob" unless safe_klass.present?
+    attr_accessor :job
 
-      safe_klass.constantize.new(**options).import!
+    def initialize(klass:, options:)
+      @klass = klass
+      @options = options
```
ttoomey: I am a little confused as to why the params go into initialize instead of perform. I guess it's because we are enqueuing it from `Delayed::Job.enqueue` instead of ActiveJob? If we want to do it this way, should probably include a call to `super()` in initialize.

Also are we enqueuing it this way so we get access to `job` in the before hook?

ttoomey (approved):
Looks good. At some point maybe we should try to find some better patterns for working with DJ to reduce the need for tight coupling in case we need to move away from it

app/jobs/importing/hud_zip/hmis_auto_migrate_job.rb
eanders: nope, I think you are correct, I just blindly moved that from further down.

app/jobs/importing/hud_zip/fetch_and_import_job.rb
eanders: I'm going to see if I can get a better handle on using `perform_later(...)` and move some of the DJ specific logic into the base class or somewhere else centralized.  Thanks for the feedback!

eanders:
@ttoomey I took a stab at cleaning this up a bit more.  There's still some smell around max-attempts, which Active Job doesn't support natively and DelayedJob doesn't store on the record and won't obey the method in the job class unless it is called with `.enqueue`.  
If this looks good, I'll update https://github.com/greenriver/hmis-warehouse/pull/5025 to match.

app/jobs/importing/hud_zip/fetch_and_import_job.rb
```diff
@@ -7,22 +7,46 @@
...
+      delayed_job.update!(attempts: calculated_attempts)
+    end
+
+    def calculated_attempts
+      dj_max_attempts = Delayed::Worker.max_attempts
+      max_attempts = 1
+      return 0 if max_attempts >= dj_max_attempts
+
+      dj_max_attempts - max_attempts
+    end
```
ttoomey: So this sets the job attempts to 1 less than the max. The intention is that it should always re queue?

Also since the return would be constant, maybe it can be written a little more concisely
```suggestion
    def calculated_attempts
      [0, Delayed::Worker.max_attempts - 1].max
    end
```

app/jobs/base_job.rb
```diff
@@ -61,4 +61,32 @@ def before(job)
...
+      locked_by: nil,
+      run_at: timestamp,
+      attempts: calculated_attempts,
+    )
+  end
+
+  # Attempt to find the associated delayed job so we can use it
+  def delayed_job
+    job = Delayed::Job.jobs_for_class(job_id).first
+    raise 'Unable to find a related delayed job' unless job.present?
```
ttoomey: might be good to include the id in case we need to debug
```suggestion
    raise "Unable to find a related delayed job (ID: #{job_id})" unless job.present?
```

app/jobs/importing/hud_zip/fetch_and_import_job.rb
```diff
@@ -7,22 +7,46 @@
...
       ].freeze
     end
+
+    private def advisory_lock_name(data_source_id)
+      GrdaWarehouse::DataSource.import_advisory_lock_name(data_source_id)
+    end
+
+    def enforce_max_attempts
+      delayed_job.update!(attempts: calculated_attempts)
+    end
```
ttoomey, line 42: If the job calls requeue_at when it fails, is this hook still needed? It seems like attempts would already be set to calculated_attempts at that point. Or is DJ updating attempts and this ensures it will run again? Maybe add a comment as to why it's needed

ttoomey (approved):
Looks good, thanks for pulling shared code into the base class. I think I'm following it all but left some small suggestions/questions

app/jobs/importing/hud_zip/fetch_and_import_job.rb
eanders: This doesn't call `requeue_at` when it fails, only when it fails to get the lock.  The `requeue_at` process allows this instance of the delayed job to complete successfully, but creates a new copy of it scheduled for the new time.

`enforce_max_attempts` is only really needed on the initial creation of the job to prevent it from running more than once.

app/jobs/importing/hud_zip/fetch_and_import_job.rb
eanders: The intention here is to limit the number of times Delayed Job attempts to run this job if it fails.  With the way we're queuing it now with `.perform_later` there's no way to indicate how many times the job should be attempted.  To work around that, we just pretend like it's already failed all but one time so the next time it fails DelayedJob/ActiveJob marks it as fully failed. 

Re-queueing is unrelated to failures, and should only happen if the lock is unavailable.

PR #5027, opened 2024-12-31
7129-dest-client-cleanup

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/7129)
Adds rake task to report on integrity issues in warehouse_clients

run as:
```
rails grda_warehouse:clients:validate > warehouse_client_issues.csv
```

## Type of change
Manual data validation

### DIFF
--- a/lib/tasks/grda_warehouse_clients.rake
+++ b/lib/tasks/grda_warehouse_clients.rake
@@ -1,0 +1,63 @@
+# frozen_string_literal: true
+
+# rubocop:disable Naming/MethodParameterName
+
+require 'csv'
+
+namespace :grda_warehouse do
+  namespace :clients do
+    # rails grda_warehouse:clients:validate > warehouse_client_issues.csv
+    desc 'Report on validation issues in WarehouseClient associations, emits issues a csv'
+    task validate: [:environment] do
+      CSV do |csv|
+        csv << ['warehouse_client_id', 'error_message']
+        WarehouseClientValidator.validate do |wc, errors|
+          errors.each do |error|
+            csv << [wc.id, error]
+          end
+        end
+      end
+    end
+  end
+end
+
+class WarehouseClientValidator
+  def self.validate
+    destination_data_source_ids = GrdaWarehouse::DataSource.destination_data_source_ids.to_set
+    warehouse_clients = GrdaWarehouse::WarehouseClient.preload(:source, :destination)
+    destination_client_ids = GrdaWarehouse::WarehouseClient.distinct.pluck(:destination_id).to_set
+    warehouse_clients.find_each(batch_size: 1000) do |wc|
+      errors = validate_warehouse_client(
+        wc,
+        destination_data_source_ids,
+        destination_client_ids,
+      )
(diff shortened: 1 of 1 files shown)


lib/tasks/grda_warehouse_clients.rake
```diff
@@ -0,0 +1,63 @@
...
+    end
+  end
+
+  def self.validate_warehouse_client(wc, destination_data_source_ids, destination_client_ids)
+    errors = []
+
+    # Basic existence and equality checks
+    errors << "Source and destination are the same client (ID: #{wc.source_id})" if wc.source_id == wc.destination_id
+    errors << "Source client (ID: #{wc.source_id}) does not exist" unless wc.source
+    errors << "Destination client (ID: #{wc.destination_id}) does not exist" unless wc.destination
```
ttoomey, line 45: Looks like we have missing source/destinations in QA. Perhaps these are expected?

eanders: Not totally unexpected.  I think our cleanup of WarehouseClient records is minimal.  It might be a useful exercise to include a check against "deleted" clients and log the date they were marked deleted if available.

lib/tasks/grda_warehouse_clients.rake
```diff
@@ -0,0 +1,63 @@
...
+    # Destination-specific checks
+    if wc.destination
+      errors << "Destination client (ID: #{wc.destination_id}) is not in a destination data source" if wc.destination && destination_data_source_ids.exclude?(wc.destination.data_source_id)
+    end
+
+    # Source-specific checks
+    if wc.source
+      errors << "Source client (ID: #{wc.source_id}) is in a destination data source" if destination_data_source_ids.include?(wc.source.data_source_id)
+      errors << "Source client (ID: #{wc.source_id}) does not match warehouse client data source (ID: #{wc.data_source_id})" if wc.data_source_id != wc.source.data_source_id
+      errors << "Source client (ID: #{wc.source_id}) is used as a destination in another warehouse client" if destination_client_ids.include?(wc.source_id)
```
eanders, line 56: I think these are important checks, thanks!

eanders (approved):
Thanks @ttoomey, looks like a solid start that we can learn from.

lib/tasks/grda_warehouse_clients.rake
ttoomey: I think client records have to exist due to the FK, so they must be just marked as deleted. Is there a reason to keep the WarehouseClient record around for a deleted client record? Or maybe we can delete them after some time, say `deleted_at < 1.month.ago` or something?

I could add another task, say `grda_warehouse:clients:prune_invalid` that deletes warehouse clients that seem like they could be safely removed. (if the source or destination client has been marked as deleted OR if the source_id == destination_id).

eanders: I _think_ it is safe to remove the related warehouse client records after some amount of time.  I think we also are removing the deleted client records now after some amount of time (maybe a year?) so maybe that'll happen automatically eventually?

PR #5028, opened 2024-12-31
Infer new RelationshipToHoH when changing HoH

gigxz:
## _Merging this PR_

## Description
This relates to Issue: https://github.com/open-path/Green-River/issues/7127 because we no longer want to set relationships to `99` when changing HoH, since that's considered an invalid value for RelationshipToHoH.

I chose to _keep_ `DATA_NOT_COLLECTED` as an enum value for the Relationship To HoH field, because 99 is a possible value in the CSV export, so we know that we will (and do) have Enrollments with this value. It is preferable to display those as `Data not collected` rather than `Invalid value`. The frontend PR, https://github.com/greenriver/hmis-frontend/pull/1014, takes care of restricting the picklist so that future enrollments going forward cannot be created with a 99 RelationshipToHoH.

This PR makes the following change: **When changing HoH on an existing household**, we can **infer** the old HoH's relationship to the new HoH. Here is an example to illustrate. Say the household consists of **Bob (HoH), Sally (Spouse), Jane (Other relative), and Timmy (Child)**. Some possible scenarios:

* If user changes the HoH to Sally, we can infer that Bob's relationship to Sally is `Spouse`.
* If user changes the HoH to Jane, we can infer that Bob's relationship to Jane is `Other Relative`.
* If user changes the HoH to Timmy (unexpected but possible), we can infer that Bob's relationship to Timmy is `Other Relative` (because opposite of "Child" would be "Parent" which is a relative..)

Related (but not dependent) Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1014
I also added a tiny cleanup routine which we may want to run for some data.

## Testing instructions
Change the HoH on a household in hmis.

## Type of change
Compliance fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
@@ -1,5 +1,5 @@
 ###
-# Copyright 2016 - 2024 Green River Data Analysis, LLC
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
 #
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
--- a/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
@@ -728,6 +728,7 @@ module Types::HmisSchema::Enums::Hud
     value 'SPOUSE_OR_PARTNER', '(3) Spouse or partner', value: 3
     value 'OTHER_RELATIVE', '(4) Other relative', value: 4
     value 'UNRELATED_HOUSEHOLD_MEMBER', '(5) Unrelated household member', value: 5
+    value 'DATA_NOT_COLLECTED', '(99) Data not collected', value: 99
     value 'INVALID', 'Invalid Value', value: -999999
   end
 
--- a/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
@@ -728,7 +728,6 @@ module Types::HmisSchema::Enums::Hud
     value 'SPOUSE_OR_PARTNER', '(3) Spouse or partner', value: 3
     value 'OTHER_RELATIVE', '(4) Other relative', value: 4
     value 'UNRELATED_HOUSEHOLD_MEMBER', '(5) Unrelated household member', value: 5
-    value 'DATA_NOT_COLLECTED', '(99) Data not collected', value: 99
     value 'INVALID', 'Invalid Value', value: -999999
   end
 
--- a/drivers/hmis/app/models/hmis/hoh_change_handler.rb
+++ b/drivers/hmis/app/models/hmis/hoh_change_handler.rb
@@ -67,11 +67,9 @@ module Hmis
         # TODO(#6857) For now, we leave the old MID as-is IF we were unable to transfer it because the new HoH entered after move-in. We plan to adjust this pending guidance from HUD.
         hhm.move_in_date = nil unless hhm.head_of_household? && hhm.move_in_date && !new_hoh_move_in_date
 
-        # Clear RelationshipToHoH on previous HoH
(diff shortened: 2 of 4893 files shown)


drivers/hmis/app/graphql/types/hmis_schema/enums/hud.rb
```diff
@@ -728,7 +728,6 @@ class RelationshipToHoH < Types::BaseEnum
     value 'SPOUSE_OR_PARTNER', '(3) Spouse or partner', value: 3
     value 'OTHER_RELATIVE', '(4) Other relative', value: 4
     value 'UNRELATED_HOUSEHOLD_MEMBER', '(5) Unrelated household member', value: 5
-    value 'DATA_NOT_COLLECTED', '(99) Data not collected', value: 99
     value 'INVALID', 'Invalid Value', value: -999999
```
gigxz: An Enrollment with a missing or 99 value for Relationship To HoH will now resolve and appear as `Invalid` in the application. Is that what we want?

drivers/hmis/app/models/hmis/hoh_change_handler.rb
```diff
@@ -128,5 +128,21 @@ def self.move_in_date_not_transfered_msg(move_in_date)
...
+      when 3 # Spouse
+        3 # Spouse
+      when 4 # Other relative
+        4 # Other relative
+      when 5 # Unrelated household member
+        5 # Unrelated household member
+      when 2 # Child
+        4 # Other relative a.k.a. parent. This is unexpected, because child shouldn't become HoH.
+      else
+        6 # Unrelated household member
```
eanders: Pretty sure this should be a 5
```suggestion
        5 # Unrelated household member
```

drivers/hmis/spec/requests/hmis/update_relationship_to_ho_h_spec.rb
```diff
@@ -175,7 +175,7 @@ def perform_mutation
       errors = result.dig('data', 'updateRelationshipToHoH', 'errors')
       expect(enrollment).to be_present
       expect(errors).to be_empty
-      expect(e1.reload.relationship_to_hoh).to eq(99)
+      expect(e1.reload.relationship_to_hoh).to eq(6)
```
eanders: This should be a 5 as well, I think.
```suggestion
      expect(e1.reload.relationship_to_hoh).to eq(5)
```

drivers/hmis/spec/requests/hmis/update_relationship_to_ho_h_spec.rb
```diff
@@ -86,9 +86,9 @@ def perform_mutation
 
     context 'when new HoH entered before move-in' do
       it 'should transfer move-in date to new HoH' do
-        expect { perform_mutation }.to change { hoh.relationship_to_ho_h }.from(1).to(99).
+        expect { perform_mutation }.to change { hoh.relationship_to_ho_h }.from(1).to(3).
           and change { hoh.move_in_date }.from(hoh_move_in_date).to(nil).
-          and change { hhm.relationship_to_ho_h }.from(2).to(1).
+          and change { hhm.relationship_to_ho_h }.from(3).to(1).
```
eanders, line 91: I'm not seeing a test for 2 -> 4.  Feels like maybe we should have one.  (Changed a child to the HoH, previous HoH should be (4) "Head of household’s other relation member"

eanders (commented):
@gigxz thanks for putting this together.  I think it looks good, but there were a couple `6`s that I don't think would be valid, and it would be nice to have a test for the change from child to HoH.

ttoomey (approved):
nicely handled, thanks!

drivers/hmis/app/models/hmis/hoh_change_handler.rb
gigxz: Oops good catch, thank you!!

drivers/hmis/spec/requests/hmis/update_relationship_to_ho_h_spec.rb
gigxz: Added!

eanders (approved):
Looks great, thanks @gigxz !

PR #5031, opened 2025-01-01
6233 replace postgres rules with functions and triggers

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/6233)

* migration to replace Postgres rules with triggers and functions
* use the fx gem to manage triggers/functions

### Notes:
* This PR is extracted from [4988](https://github.com/greenriver/hmis-warehouse/pull/4988) which was getting to unwieldy. 
* the rules we replace were just there to prevent active record from trying to update/delete into our views

### Rationale
* It's more maintainable to use the fx gem rather than SQL in the schema (much like we manage db views with scenic)
* [Why not use rules](https://wiki.postgresql.org/wiki/Don't_Do_This#Don.27t_use_rules)
* This let's us move to schema.rb in the future

## Type of change
DB internals

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -51,8 +51,8 @@ gem 'autoprefixer-rails', '~> 10.3.3' # pinned until we can update to Bootstrap 
 gem 'kaminari'
 gem 'pagy'
 gem 'with_advisory_lock'
-# gem 'schema_plus_views'
-gem 'scenic'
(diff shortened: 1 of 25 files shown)


eanders (approved):
Thanks @ttoomey, looks like it caught all the rules I'm seeing.

PR #5033, opened 2025-01-01
522-improve-unique-name-generation

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/522)

## Testing
Unique name is used to for fuzzy name matching (double metaphone):
- Identifying potential matches across client records
- In client search by name

test from the console:
```
> GrdaWarehouse::Hud::Client.where(GrdaWarehouse::Hud::Client.metaphone_search(GrdaWarehouse::Hud::Client.arel_table[:id].eq(0), :FirstName, first_name_query_string)).pluck(:first_name
).sort.uniq
```

Improve UniqueName nightly update performance and features
* advisory lock
* batching
* cleanup unused records
* add spec

## Type of change
Enhancement

### DIFF
--- a/app/models/unique_name.rb
+++ b/app/models/unique_name.rb
@@ -36,7 +36,10 @@ class UniqueName < ApplicationRecord
           inserts << UniqueName.new(name: name, double_metaphone: double_metaphone)
         end
       end
-      UniqueName.import(inserts, raise_error: true) if inserts.any?
+      next if inserts.empty?
+
+      result = UniqueName.import(inserts)
+      raise "Failed to import UniqueName: #{result.inspect}" if result.failed_instances.present?
     end
 
     # remove names that are no longer used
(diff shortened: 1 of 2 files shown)


app/models/unique_name.rb
```diff
@@ -5,21 +5,47 @@
...
+
+          name_exists = name.in?(existing_names)
+          existing_names[name] = true
+          next if name_exists
+
+          double_metaphone = Text::Metaphone.double_metaphone(name)
+          inserts << UniqueName.new(name: name, double_metaphone: double_metaphone)
+        end
+      end
+      UniqueName.import(inserts, raise_error: true) if inserts.any?
```
eanders: Any reason not to use:
```suggestion
      UniqueName.import!(inserts) if inserts.any? 
```
(I believe this enforces validation too, but otherwise should be the same).

eanders (approved):
Thanks @ttoomey, and thanks especially for the tests.

app/models/unique_name.rb
ttoomey: oh oops, thanks for catching. I actually intended to raise if the import failed. Fixed

PR #5035, opened 2025-01-03
7135-legacy-policy-bugfix

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/7135)
Source client auth policy fix for legacy users with windowed data sources. Account for behavior of special global permissions 

### Setup
1. In a project with a "windowed" data source:
   - Identify a client in that data source who **has no valid ROI** (Release of Information).  
2. Find a test user meeting the following criteria:
   - The user has **legacy role-based access**.  
   - The user has **no relationship** to the client (e.g., via access groups).  
   - The user initially has **no roles** with `can_view_client_name`, `can_view_clients` or `can_search_all_clients`.

### 1. Name Visibility on Search
1. Assign the user a role with the permission `can_search_all_clients`.  
2. Search for the client using their ID (or destination ID).  
   - Verify that the client search card displays the record with **"Name Redacted"**.  
3. Assign the user a role with the permission `can_view_client_name`.  
4. Search for the client again.  
   - Verify that the client search card now displays the record with the **full name**.  

### 2. Name and SSN Visibility on Dashboard
1. Assign the user roles that grant `can_view_clients` and `can_view_full_ssn`.  
2. Search for the client again.  
   - Click on the client in the search results to view their dashboard.  
3. Verify that the **client's name** and **SSN** are displayed on the dashboard.  

## Type of change
Bug fix

### DIFF
--- a/app/models/grda_warehouse/auth_policies/source_client_policy.rb
+++ b/app/models/grda_warehouse/auth_policies/source_client_policy.rb
@@ -74,6 +74,8 @@ class GrdaWarehouse::AuthPolicies::SourceClientPolicy < GrdaWarehouse::AuthPolic
     results
   end
 
+  BASIC_CLIENT_PII_PERMS = Set.new([:can_view_client_name, :can_view_client_photo, :can_view_full_dob]).freeze
+
   # Window data sources are a deprecated legacy client data sharing mechanic, replaced by a System Collection when using Access Controls-based permissions
   def add_legacy_data_source_permissions(results)
     # is this a user with legacy role-based perms?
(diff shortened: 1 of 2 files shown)


spec/models/auth_policies/source_client_policy_spec.rb
```diff
@@ -118,14 +118,42 @@
...
+              {
+                can_search_all_clients: true,
+                can_view_client_name: true,
+                can_view_full_ssn: true,
+                can_view_full_dob: false,
+              }
+            end
+            it 'has expected PII permissions' do
+              expect(policy.can_view_name?).to be true
+              expect(policy.can_view_full_ssn?).to be false
```
ttoomey, line 154: we expect ssn to not be true because it's not a "basic" permission

eanders:
Should we target 145 so we can test this against the known issue?

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -83,7 +85,19 @@ def add_legacy_data_source_permissions(results)
     # is the client in a window data source?
     return unless context.legacy_window_data_source_ids.include?(client.data_source_id)
 
-    # check roi if the window config requires release (the "can_*_with_roi" permissions are not relevant in this case)
+    # The 'can_view_clients' and 'can_search_all_clients' permissions allow the user's permissions to apply to clients in windowed data sources, regardless of ROI.
+    # For example a user with can_view_clients and can_view_name can view the names of all clients in windowed data sources
```
eanders:
```suggestion
    # The 'can_view_clients' and 'can_search_all_clients' permissions allow certain permissions to apply to clients in window data sources, regardless of ROI.
    # For example a user with can_view_clients and can_view_name can view the names of all clients in window data sources
```

eanders: Maybe this should be written from the alternate perspective.  Something like:
In the legacy role-based context, visibility of client attributes is global if the user can access a client within the "search" or "view" contexts for data sources that are "visible in the window."  This means we do not need to check ROI to determine if the user can see a client's name as long as they have a role that provides that permission AND can access the client through a window data source and `can_view_clients` or `can_search_all_clients` permission.

(That's probably not any more clear, but maybe?)

app/models/grda_warehouse/auth_policies/source_client_policy.rb
```diff
@@ -83,7 +85,19 @@ def add_legacy_data_source_permissions(results)
...
+    if legacy_permissions.include?(:can_view_clients)
+      # all the legacy perms apply to the client
+      results.merge(legacy_permissions)
+      # early return since there's no point in checking ROI
+      return
+    elsif legacy_permissions.include?(:can_search_all_clients)
+      # can_search_all_clients confers a reduced set or permissions
```
eanders: Historically this hasn't conferred fewer permissions, maybe we should make a note here that this is tighter than the historic permissions, but that's ok, since the search results have limited client details.  This is to enforce more strict permissions in the future.

eanders (approved):
Thanks @ttoomey !
If you want to retarget 145 we can give this a try before merging.

ttoomey:
Thanks @eanders  - I rebased to 145 so we can test it out

PR #5040, opened 2025-01-07
Fix bug with Client Assessments rollup where form processor is nil

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7178
Sentry link: https://green-river.sentry.io/issues/6202297210/?alert_rule_id=12523843&alert_type=issue&environment=production&notification_uuid=d4c20e4f-37d5-4368-8c4e-62405c3c0b92&project=4503977346662400&referrer=slack

Bug: Assessment form processors may be nil, in migrated-in data. But the client assessments rollup doesn't flexibly allow for that. It's throwing an error when checking Custom Assessments against CE Assessments to deduplicate them in the view. 

I reproduced the bug and confirmed the fix locally by `destroy`ing a form processor, to make the assessment look like migrated-in data.

Question: I think this is sufficient to avoid the bug, but is this sufficient to maintain the desired functionality of de-duping CE Assessments? Let's say there is a migrated-in Custom Assessment that doesn't have a form processor, but does correspond to a migrated-in CE Assessment? Would that ever happen? 

## Type of change
Bug fix

### DIFF
--- a/app/views/clients/rollup/_assessments.html.haml
+++ b/app/views/clients/rollup/_assessments.html.haml
@@ -12,7 +12,7 @@
   -# remove any CE assessments for which we have a matching HMIS custom assessment.  The custom assessment will give better fidelity
   -# we'll add back the ce assessment details on the show page
   - custom_hmis_assessments.each do |assessment|
-    - ce_assessments.delete_if { |ce| ce.id == assessment.form_processor.ce_assessment_id }
+    - ce_assessments.delete_if { |ce| ce.id == assessment.form_processor&.ce_assessment_id }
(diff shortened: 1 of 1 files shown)


gigxz:
Migrated-in "ce assessments" would need to have a form processor, because thats how they're linked to a definition. I'm not sure why a form processor would be missing actually. Did you identify the reason for that in this particular case? (The assessment migration job should always create them, as far as I understand)

gigxz (approved):

eanders (approved):
Works for me.  Thank you @martha !

gigxz:
As discussed on the issue I think there is a separate underlying bug, but I still think it makes sense to merge this change.

PR #5044, opened 2025-01-09
Fix auto-exit bug when service date is before entry

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7178

- When determining the auto-exit date for NBN shelters, take the max of the last service date and the enrollment entry date. This avoids the AutoExitJob failing due to data quality issues.
- Add an explicit check for validity of the exit record before saving the assessment. This avoids the AutoExitJob silently failing to create the Exit record even thought the Exit Assessment does get saved.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -27,8 +27,9 @@ module Hmis
         project.enrollments.open_excluding_wip.each do |enrollment|
           most_recent_contact = if project.es_nbn? # Night-by-night Emergency Shelter
             # For NBN shelters, the most recent contact is the last bed night.
-            # If the client had no bed nights, use the enrollment (entry date) as the last contact.
-            enrollment.services.bed_nights.where.not(date_provided: nil).order(:date_provided).last || enrollment
+            last_bed_night = enrollment.services.bed_nights.where.not(date_provided: nil).order(:date_provided).last
+            # If the client had no bed nights, or the last bed night is before enrollment entry (invalid), use the enrollment (entry date) as the last contact.
+            [last_bed_night, enrollment].compact.max_by { |entity| contact_date_for_entity(entity) }
           else
             [
               enrollment.services.where.not(date_provided: nil).order(:date_provided).last,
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -81,6 +82,9 @@ def auto_exit(enrollment, most_recent_contact)
         enrollment_id: enrollment.enrollment_id,
       )
       assessment.build_form_processor(exit: exit_record)
+
+      raise ActiveRecord::RecordInvalid, exit_record if exit_record.invalid?
```
martha, line 86: I tested this with a regression test, but removed it, since it was invalidated by the other change:

```
    it 'should not silently fail if exit record is invalid (regression #7178)' do
      e1 = create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, user: u1, entry_date: Date.current - 2.months
      # most recent contact date is before entry date, creating an invalid exit
      create :hmis_hud_service, data_source: ds1, client: c1, enrollment: e1, user: u1, record_type: 200, date_provided: Date.current - 3.months

      expect do
        Hmis::AutoExitJob.perform_now
      end.to raise_error(ActiveRecord::RecordInvalid, /cannot be before entry date/)

      expect(Hmis::Hud::Enrollment.exited).not_to include(e1)
      expect(e1.exit_assessment).to be_nil
    end
```

I didn't see an obvious way to keep the regression test and make the exit record invalid in some other way -- let me know if you can think of one? It didn't seem worth a ton of time to me though.

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
```diff
@@ -54,6 +54,20 @@
       Hmis::AutoExitJob.perform_now
       expect(e1.exit).to have_attributes(auto_exited: be_present, exit_date: e1.entry_date, destination: 30)
     end
+
+    it 'should not fail if enrollment has contact date before entry (regression #7178)' do
+      e1 = create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c1, user: u1, entry_date: Date.current - 2.months
```
gigxz, line 59: just as a random note, it's not necessary to pass `user: u1`. It will be generated using the factory if not passed explicitly. We do have a lot of tests floating around that specify the user, but its not necessary unless we're trying to test something specific about that user.

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: Yeah that seems okay to me.

gigxz (approved):
Thanks!

PR #5045, opened 2025-01-09
Fix bug with auto-exiting HoH when HHMs are still active

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7188 
- Prevent the Auto Exit job from exiting a household member if anyone else in their household shouldn't be exited:
  - If someone else in the household has a more recent "last contact date"
  - If someone else in the household has an incomplete enrollment
- Add spec tests

## Type of change

### DIFF
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -25,19 +25,17 @@ module Hmis
         raise "Auto-exit config unusually low: #{config.length_of_absence_days}" if config.length_of_absence_days < 30
 
         project.enrollments.open_excluding_wip.each do |enrollment|
-          most_recent_contact = if project.es_nbn? # Night-by-night Emergency Shelter
-            # For NBN shelters, the most recent contact is the last bed night.
-            last_bed_night = enrollment.services.bed_nights.where.not(date_provided: nil).order(:date_provided).last
-            # If the client had no bed nights, or the last bed night is before enrollment entry (invalid), use the enrollment (entry date) as the last contact.
-            [last_bed_night, enrollment].compact.max_by { |entity| contact_date_for_entity(entity) }
+          # If this is a HoH of a multimember household, only exit them if they won't be leaving behind a headless household!
+          most_recent_contact = if enrollment.head_of_household? && enrollment.household_members.count > 1
+            # If any household member's enrollment is still in progress, don't exit the HoH
+            next if enrollment.household_members.any?(&:in_progress?)
+
+            # Look at most recent contact across all enrollments in this household when determining whether to exit the HoH
+            enrollment.household_members.
+              map { |hhm| get_most_recent_contact(hhm, project) }.
+              max_by { |entity| contact_date_for_entity(entity) }
           else
-            [
-              enrollment.services.where.not(date_provided: nil).order(:date_provided).last,
-              enrollment.custom_services.order(:date_provided).last,
-              enrollment.current_living_situations.order(:information_date).last,
-              enrollment.custom_assessments.order(:assessment_date).last,
-              enrollment,
-            ].compact.max_by { |entity| contact_date_for_entity(entity) }
+            get_most_recent_contact(enrollment, project)
           end
 
           most_recent_contact_date = contact_date_for_entity(most_recent_contact)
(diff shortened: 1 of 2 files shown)


martha:
I don't think I have a good enough handle on arel to generate the right query, but here is a sql query that showed me the 26 enrollments in my local environment that are affected by this bug:

```ruby
sql = 'select "Enrollment".*
from "Enrollment"
  inner join "Exit"
  on "Exit"."DateDeleted" is null 
  and "Exit"."EnrollmentID" = "Enrollment"."EnrollmentID"
  and "Exit"."PersonalID" = "Enrollment"."PersonalID"
  and "Exit"."data_source_id" = "Enrollment"."data_source_id"
inner join (
  select "Enrollment"."HouseholdID" as household_id, "Exit"."ExitDate" as exit_date
  from "Enrollment" 
    inner join "Exit"
    on "Exit"."DateDeleted" is null 
    and "Exit"."EnrollmentID" = "Enrollment"."EnrollmentID"
    and "Exit"."PersonalID" = "Enrollment"."PersonalID"
    and "Exit"."data_source_id" = "Enrollment"."data_source_id"
  where "Enrollment"."DateDeleted" is null 
  and "Enrollment"."RelationshipToHoH" = 1
  and "Exit"."ExitDate" is not null
) as hoh_exit_date
on "Enrollment"."HouseholdID" = hoh_exit_date.household_id
where "Exit"."ExitDate" is null 
or "Exit"."ExitDate" > hoh_exit_date.exit_date'

Hmis::Hud::Enrollment.find_by_sql(sql).count
```

I did try running this in a client env (tarrant) but got spooked by how long it was taking and killed it. (probably waited ~15 seconds and didn't want to keep hogging the database resources.)

martha:
Treat household as a unit - either auto-exit the whole household or don't auto-exit anybody

martha:
@gigxz I think this is ready for re-review! 🙏

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -24,30 +24,27 @@ def perform
...
-          else
-            [
-              enrollment.services.where.not(date_provided: nil).order(:date_provided).last,
-              enrollment.custom_services.order(:date_provided).last,
-              enrollment.current_living_situations.order(:information_date).last,
-              enrollment.custom_assessments.order(:assessment_date).last,
-              enrollment,
-            ].compact.max_by { |entity| contact_date_for_entity(entity) }
-          end
+        project.households.each do |household|
```
gigxz: Going from iterating  `project.enrollments.open_excluding_wip` to iterating `project.households` will be a performance hit since now we will now perform the `household.enrollments` query once for each household in the project -- including exitedhouseholds.

I haven't tested this, but an improvement may be:
```ruby
# iterate through each active household in the project.
# skip households where any members are WIP, since we shouldn't auto-exit from those
project.households.active.not_in_progress.preload(:enrollments).each do |household|
   ...
```


There is more we could do with preloading, like preloading all the contact records, but recognizing that we didn't do that previously so it may be fine as-is.

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
```diff
@@ -68,6 +68,57 @@
       expect(e1.exit_assessment&.assessment_date).to eq(e1.entry_date)
       expect(e1.exit_assessment&.data_collection_stage).to eq(3)
     end
+
+    let!(:c2) { create :hmis_hud_client, data_source: ds1, user: u1 }
```
gigxz: Id suggest either put this at the top with the rest of a fixtures, or in a `context` that focuses on households

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
```diff
@@ -68,6 +68,57 @@
...
+      hhm_e = create :hmis_hud_enrollment, data_source: ds1, project: p1, client: c2, household_id: household_id, relationship_to_hoh: 2, entry_date: Date.current - 2.months
+
+      # non-hoh has a recent bed night and won't get auto-exited
+      create :hmis_hud_service, data_source: ds1, client: c2, enrollment: hhm_e, user: u1, record_type: 200, date_provided: Date.current - 2.days
+
+      Hmis::AutoExitJob.perform_now
+
+      expect(hhm_e.exit).to be_nil
+      expect(hoh_e.exit).to be_nil # HOH shouldn't get exited
+    end
```
gigxz: (not a change request, just some musings to revisit later....)

I see you followed the pattern of the rest of this file, which sets up fixtures inside the `it` blocks. I'd  love to make some guidelines internally for how to set up rspec tests, because we have soo much variance. I've come to find that things are more understandable and maintainable if fixtures move into the context and the tests can be as short and to-the-point as possible.

Another issue we have is that hosueholds require a lot of setup- it would be nice if we could use factories to set up households in less code.

This is following I think [this thread](https://www.reddit.com/r/rails/comments/1gqlxob/advice_on_setting_up_the_world_in_an_rspec_test/) asks the question nicely.

Example of pattern that we use in many other tests that I have come to prefer:

```

context 'household with a recent contact' do
   # create fixtures here

   it 'should not exit any household members' do
      expect { Hmis::AutoExitJob.perform_now }.not_to change(Hmis::Hud::Exit.count ). and not_change(Hmis::HUd::CustomAssessment.exits.count)
   end
end

context 'household with incomplete enrollments' do
   # ...
end
```

gigxz (approved):
edit::: forgot about my comment about performance improvements, I do suggest those

gigxz (changes requested):
performance improvement https://github.com/greenriver/hmis-warehouse/pull/5045/files#r1917094280

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
martha: Thanks @gigxz this is great 👍

drivers/hmis/spec/jobs/auto_exit_job_spec.rb
martha: Thank you for the feedback and for sharing that thread. I always feel a little hesitant to use contexts because of once reading [this styleguide](https://rspec.rubystyle.guide/#context-cases) early in my rails journey, which indicates
> context blocks should pretty much always have an opposite negative case. It is a code smell if there is a single context (without a matching negative case), and this code needs refactoring, or may have no purpose.

Anyway, that linked guide in general could be a good resource to us when deciding on what patterns we want to centralize around.

gigxz (approved):
lgtm!

PR #5047, opened 2025-01-09
Implement Join Household mutation

martha:
## _Merging this PR_
- Note that the corresponding frontend PR https://github.com/greenriver/hmis-frontend/pull/1022 is targeting the tables deploy branch

## Description
GH issue: https://github.com/open-path/Green-River/issues/5752

Related frontend work: https://github.com/greenriver/hmis-frontend/pull/1022

Summary of changes:
- Add the `JoinHouseholds` mutation
- Add the `hmis_household_events` table and model for storing join and split events
- Update the enrollment validator: When there is an error about a conflicting enrollment, it returns that enrollment's ID in the `data` field of the error
- Add spec tests
- Add system test

Still to do:
- [x] Add system test
- [x] Update unit
- [x] Add household history table

## Type of change


bin/run_hmis_system_tests.sh
```diff
@@ -107,6 +107,12 @@ cd "$ORIGINAL_CWD"
 # skip okta if it's set in our local env
 unset HMIS_OKTA_CLIENT_ID
 unset OKTA_DOMAIN
+# TODO @MARTHA - REVERT
```
martha: Todo

db/warehouse/migrate/20250121204933_add_hmis_household_events_table.rb
```diff
@@ -0,0 +1,29 @@
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+class AddHmisHouseholdEventsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :hmis_household_events do |t|
+      t.string :household_id, index: true # hmis_households is a view, not a table
```
gigxz: Is this the HUD `HouseholdID` string? That's what I would have expected, but am a little confused by your comment about the View. 

I believe this should be NOT NULL as well.

I'm also not totally sure if it's useful to index the HouseholdID alone, since we'd always be looking it up by Data Source ID as well. So maybe an index on `data_source_id,household_id`.

gigxz: Maybe we should just go ahead and use `HouseholdID` casing for this column for consistency. I worry that `household_id` is ambiguous because we may later add a `household` _table_, which then would have a primary key, but we're not referencing that here.

I'd say our convention right now is:
* `EnrollmentID` likely contains the HUD Enrollment "key"
* `enrollment_id` likely contains a primary key reference to the `Enrollment` table

So it may be confusing to add `household_id` that contains a HUD key.

db/warehouse/migrate/20250121204933_add_hmis_household_events_table.rb
```diff
@@ -0,0 +1,29 @@
...
+      t.references :hmis_household_event, null: false, foreign_key: { to_table: :hmis_household_events }, index: { name: 'idx_hmis_household_event_enrollments_on_event_id' }
+      t.references :enrollment, null: false, foreign_key: { to_table: :Enrollment }, index: true # this is the ID of the enrollment record: Enrollment.id not Enrollment.EnrollmentID
+    end
+
+    # Add a composite unique index to prevent duplicate associations
+    add_index :hmis_household_event_enrollments,
+              [:hmis_household_event_id, :enrollment_id],
+              unique: true,
+              name: 'idx_hmis_household_event_enrollments_on_event_and_enrollment'
+  end
```
gigxz: I'm wondering if this complexity is worth it just for fk constraints on the enrollment references.

For other similar tales (client merge histories, client merge audits) we don't have fk constraints.
The primary goal is to record events as they happen, even if the related records (Enrollments) are deleted or changed. Because of that, I think the fk constraints might not be necessary because the historical accuracy of the audit trail is more important than referential integrity.


For example if an Enrollment get's hard-deleted, we probably want the hmis_household_events that it was a part of to remain. We also don't want the soft-deletion of an enrollment to cascade and delete the household events that it was a part of.

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+###
+
+module Mutations
+  class JoinHousehold < BaseMutation
+    argument :receiving_household_id, ID, required: true
+    argument :joining_enrollment_inputs, [Types::HmisSchema::EnrollmentRelationshipInput], required: true
+
+    field :receiving_household, Types::HmisSchema::Household, null: false
+    field :donor_household, Types::HmisSchema::Household, null: true # Will be null if there are no remaining members
+    # todo @martha - possible that there will be multiple households with same ID? mutation should require project_id?
```
gigxz: `HouseholdID` and `data_source_id` are unique together -- there should be no need for project id.

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Mutations
+  class JoinHousehold < BaseMutation
+    argument :receiving_household_id, ID, required: true
+    argument :joining_enrollment_inputs, [Types::HmisSchema::EnrollmentRelationshipInput], required: true
+
+    field :receiving_household, Types::HmisSchema::Household, null: false
+    field :donor_household, Types::HmisSchema::Household, null: true # Will be null if there are no remaining members
```
gigxz, line 13: Do we need to resolve this? From what I can see, the frontend only looks at the receiving household result.

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+    def resolve(receiving_household_id:, joining_enrollment_inputs:)
+      receiving_enrollments = Hmis::Hud::Enrollment.
+        where(household_id: receiving_household_id).
+        viewable_by(current_user).
+        includes(:project, :household)
+      access_denied! if receiving_enrollments.empty?
+
+      receiving_household = receiving_enrollments.first.household
+
+      receiving_project = receiving_enrollments.map(&:project).uniq.sole
```
gigxz: I think this could be simplified by using the household view directly (untested)
```suggestion
      # note: viewable_by takes care of filtering by data source via project with_access
      receiving_household = Hmis::Hud::Household.viewable_by(current_user).find_by(household_id: receiving_household_id)
      access_denied! unless receiving_household

      receiving_enrollments = receiving_household.enrollments
      receiving_project = receiving_household.project
```

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+      map_enrollment_id_to_relationship = joining_enrollment_inputs.map { |e| [e.enrollment_id, e.relationship_to_hoh] }.to_h
+
+      joining_enrollment_ids = map_enrollment_id_to_relationship.keys
+      joining_enrollments = Hmis::Hud::Enrollment.
+        viewable_by(current_user).
+        where(id: joining_enrollment_ids).
+        includes(:household)
+      access_denied! unless joining_enrollments.count == joining_enrollment_inputs.count
+
+      raise 'Cannot merge enrollments from another project' unless joining_enrollments.pluck(:project_pk).uniq.sole == receiving_project.id
```
gigxz: I think we can improve performance slightly by skipping the `viewable_by` scope and just looking for enrollments within the project.
Can also improve perf for large households by not loading the `household` view for each member. We really should only need to load it once.

```suggestion
      joining_enrollments = Hmis::Hud::Enrollment.
        where(project: receiving_project).   # can only join from household in same project
        where(id: joining_enrollment_ids)

      access_denied! unless joining_enrollments.count == joining_enrollment_inputs.count
      raise 'join from multiple households' unless joining_enrollments.map(&:household_id).uniq.size == 1

      donor_household = joining_enrollments.first.household
```

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      donor_household = joining_enrollments.map(&:household).uniq.sole
+      remaining_enrollments = Hmis::Hud::Enrollment.
+        viewable_by(current_user). # Restrict to this data source
+        where(household_id: donor_household.household_id).
+        where.not(id: joining_enrollment_ids)
```
gigxz: I believe this can be written as:
`remaining_enrollments  = donor_household.enrollments - joining_enrollments`
(it will be an array, not a scope)

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+      remaining_hoh = remaining_enrollments.any? { |enrollment| enrollment.relationship_to_hoh == 1 }
+      raise 'This operation would leave behind a household with no HoH, which is not allowed' unless remaining_enrollments.empty? || remaining_hoh
+
+      # If the receiving household is assigned to a unit, re-assign the joining enrollments to the same unit.
+      # This works for the present-tense, but could be improved to better accommodate past data correction.
+      units = receiving_enrollments.map { |enrollment| enrollment.active_unit_occupancy&.unit }.compact.uniq
+
+      receiving_before_state = Hmis::Hud::Enrollment.snapshot_enrollments(receiving_enrollments)
+      donor_before_state = Hmis::Hud::Enrollment.snapshot_enrollments([*joining_enrollments, *remaining_enrollments])
```
gigxz: A bit hard-to-understand exactly what we're snapshotting here. Maybe we could simplify it like below to make it clearer that we are snapshotting _all_ enrollments in each household.
```suggestion
      receiving_before_state = Hmis::Hud::Enrollment.snapshot_enrollments(receiving_household.enrollemnts)
      donor_before_state = Hmis::Hud::Enrollment.snapshot_enrollments(donor_household.enrollments)
```

Or better yet, could the snapshot function be moved to the Household model? Something like:
```
receiving_before_state = receiving_household.snapshot_household_state
donor_before_state = donor_household.snapshot_household_state

# begin transaction..
  # perform changes..
  receiving_after_state = receiving_household.snapshot_household_state
   donor_after_state = donor_household&.snapshot_household_state || []
```

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+      remaining_enrollments = Hmis::Hud::Enrollment.
+        viewable_by(current_user). # Restrict to this data source
+        where(household_id: donor_household.household_id).
+        where.not(id: joining_enrollment_ids)
+      remaining_hoh = remaining_enrollments.any? { |enrollment| enrollment.relationship_to_hoh == 1 }
+      raise 'This operation would leave behind a household with no HoH, which is not allowed' unless remaining_enrollments.empty? || remaining_hoh
+
+      # If the receiving household is assigned to a unit, re-assign the joining enrollments to the same unit.
+      # This works for the present-tense, but could be improved to better accommodate past data correction.
+      units = receiving_enrollments.map { |enrollment| enrollment.active_unit_occupancy&.unit }.compact.uniq
```
gigxz: Can we simplify this to just choose one unit rather than putting each joining member potentially in a different unit? Doesn't seem worth the complexity.

```suggestion
      # If the receiving household is assigned to any unit(s), assign the joining enrollments to the same unit.
      # This works for the present-tense, but could be improved to better accommodate past data correction.
      receiving_unit = receiving_enrollments.joins(:current_unit).first&.unit
```

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      receiving_before_state = Hmis::Hud::Enrollment.snapshot_enrollments(receiving_enrollments)
+      donor_before_state = Hmis::Hud::Enrollment.snapshot_enrollments([*joining_enrollments, *remaining_enrollments])
+
+      Hmis::Hud::Enrollment.transaction do
+        joining_enrollments.each_with_index do |enrollment, index|
+          enrollment.update!(
+            household_id: receiving_household_id,
+            relationship_to_hoh: map_enrollment_id_to_relationship[enrollment.id.to_s],
+          )
```
gigxz: Is it necessary to do two saves per enrollment? (`update!` here and `save!` below). I know `assign_unit` does some validations on household membership, but I _think_ it should work even if the household_id change hasn't been persisted yet.

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        receiving_household.reload
+        donor_household.reload if remaining_enrollments.any?
+
+        joining_event = Hmis::HouseholdEvent.new
+        joining_event.user = current_user
+        joining_event.household = receiving_household
+        joining_event.event_type = Hmis::HouseholdEvent::JOIN
+        joining_event.event_details = {
+          'donorHouseholdId': donor_household.household_id,
```
gigxz: should prefer snake-case for storing json data `donor_household_id`, that's the casing we use elsewhere for json data

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -284,6 +284,15 @@ def self.generate_enrollment_id
     generate_uuid
   end
 
+  def self.snapshot_enrollments(enrollments)
+    enrollments.map do |enrollment|
+      {
+        'enrollmentId': enrollment.id,
+        'relationshipToHoh': enrollment.relationship_to_hoh,
```
gigxz: we should snake case here as well. Suggested in other comment to move this to snapshot the Household if possible, since the household event table is snapshotting households

app/models/grda_warehouse/utility.rb
```diff
@@ -200,6 +200,7 @@ def self.modifier(model)
       GrdaWarehouse::Hud::Client,
       GrdaWarehouse::Hud::Project,
       GrdaWarehouse::Hud::User,
+      GrdaWarehouse::Hud::Enrollment,
```
gigxz: was this added in order to cascade deletion of hmis_household_events when deleting an enrollment? 
it's not needed for the spec test to pass locally for me

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
```diff
@@ -34,23 +34,24 @@ def self.already_enrolled_full_message
...
+  def self.find_conflicting_enrollment(enrollment)
     conflict_scope = Hmis::Hud::Enrollment.
       where(personal_id: enrollment.personal_id, data_source_id: enrollment.data_source_id).
       with_conflicting_dates(project: enrollment.project, range: enrollment.entry_date...enrollment.exit_date)
 
     if enrollment.persisted?
       # If the entry date is being changed on an EXISTING enrollment, and it overlaps with another one, it should be a warning
       conflict_scope = conflict_scope.where.not(id: enrollment.id)
-      return :warning if conflict_scope.any?
+      return :warning, conflict_scope.first if conflict_scope.any?
```
gigxz: This should be selected deterministically, or return all of them. Since it's unlikely to have multiple overlapping enrollments, it's probably OK to pick one -- maybe by earliest entry date, so it's consistent with the below.

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
```diff
@@ -34,23 +34,24 @@ def self.already_enrolled_full_message
...
     else
-      min_conflict_date = conflict_scope.minimum(:entry_date)
-      if min_conflict_date
-        # If the entry date is being set on a NEW enrollment, and the entry date is on or after the entry date of any conflicting enrollments, it should be an error.
-        return :error if enrollment.entry_date >= min_conflict_date
-
-        # if the entry date is being set on a NEW enrollment, and the entry date is before the entry date of any conflicting enrollments, it should be a warning
-        return :warning if enrollment.entry_date < min_conflict_date
+      # If the entry date is being set on a NEW enrollment,
+      conflicting_enrollment = conflict_scope.order(entry_date: :asc).first # pick the conflicting enrollment with the earliest entry date
```
gigxz: nit:
```suggestion
      conflicting_enrollment = conflict_scope.min_by(&:entry_date)
```

drivers/hmis/app/models/hmis/hud/validators/enrollment_validator.rb
```diff
@@ -34,23 +34,24 @@ def self.already_enrolled_full_message
     'Client has another enrollment in this project that conflicts with this entry date.'
   end
 
-  def self.find_conflict_severity(enrollment)
+  def self.find_conflicting_enrollment(enrollment)
```
gigxz, line 37: Is there test coverage for the changes here?

It looks like all the Enrollment creation tests are jammed into the submit_form_spec which is big https://github.com/greenriver/hmis-warehouse/blob/a65d9cddea0badf29b3fb1bc96ab025e4e18472c/drivers/hmis/spec/requests/hmis/submit_form_spec.rb#L468-L681

This seems like the right place to add assertions on the `data` that is returned:  https://github.com/greenriver/hmis-warehouse/blob/a65d9cddea0badf29b3fb1bc96ab025e4e18472c/drivers/hmis/spec/requests/hmis/submit_form_spec.rb#L540-L545

gigxz (commented):
Looks good!! Mostly minor comments. The idea of hmis_household_event_enrollments  may need some discussion, I'm not convinced that we need it. The Let me know if you have any questions on the review!

db/warehouse/migrate/20250121204933_add_hmis_household_events_table.rb
martha: Thanks, sounds good. I'll remove the confusing comment; I just meant for it to indicate why we are not adding a fk relationship (referencing [this migration when we added staff assignments](https://github.com/greenriver/hmis-warehouse/blob/e10d40a6224bcb2b244cd745f867625d534c700f/db/warehouse/migrate/20240717205642_add_staff_assignment_tables.rb#L17-L19)).

app/models/grda_warehouse/utility.rb
martha: Yes -- that's weird, I did need it at some point to get the spec tests to pass. but anyway, I will remove it per https://github.com/greenriver/hmis-warehouse/pull/5047#discussion_r1939465299.

drivers/hmis/app/graphql/mutations/join_household.rb
martha: The frontend does need to decide whether or not to show the button "View [donor's] enrollment". "May Smith" in the mocks [here](https://www.figma.com/design/kqZiUPjXvqSvJVslJCnlTB/OP-HMIS-%2F-Tickets-%2F-Figma?node-id=9454-23856&t=vX9j189HEg1BxLcC-0). (Logic [here](https://github.com/greenriver/hmis-frontend/pull/1022/files#diff-9d96cff77d6ee71b82b0fa8139813a2a9b76af074b82c109fcc668cdfb09d72fR260)) But yes, it doesn't need to resolve the whole household. I can make that change.

drivers/hmis/app/graphql/mutations/join_household.rb
martha: Sounds good I agree! Sorry, I misinterpreted something you said about assigning the units deterministically at our 1:1 last time we discussed this

drivers/hmis/app/graphql/mutations/join_household.rb
martha: Thanks yes this was the conclusion I was coming to as well! Earlier I wrote a comment about this that I never hit publish on 🤦‍♀️ 

Here's what I found when investigating this locally:
- It's possible for a household ID to be duplicated across projects; neither the db nor the rails code prevents this
- But it's also unhandled elsewhere. For example, the Enrollment Overview and Household pages return all enrollments with this household ID, regardless of project

<details> 
<summary>Data standards relevant excerpts</summary>

Data standards manual:
> "A Household Identifier is an
automatically generated identifier
created by the HMIS software. A
Household Identifier must be
permanent and unique to a single
household at each project start
within an HMIS."

Data dictionary: 
> "The Household ID must be automatically generated by an HMIS application to
ensure that it is unique. The Household ID must be unique within each HMIS
implementation regardless of how many CoCs it covers. This includes data
warehouses that may be pulling in data from multiple HMIS instances, and one or
more CoCs."
</details>

drivers/hmis/app/graphql/mutations/join_household.rb
martha: Maybe you are saying it makes more sense for the frontend to determine that independently? It does already have the info it needs to determine whether there is still a donor household. But to me it somehow seemed more graceful to receive the info directly from the backend within the mutation response, rather than separately inferring it if the mutation response is successful. Let's discuss

drivers/hmis/app/graphql/mutations/join_household.rb
gigxz: Yep that's all right, the code expects HouseholdID+DataSource to be unique, but it's not enforced in the db. Probably because of old imported data before we enforced it on import ([here](https://github.com/greenriver/hmis-warehouse/blob/fdcc6431677d258bfb80d5e0b3cd59840be7b4a5/drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/enforce_relationship_to_hoh.rb#L198)) I believe. There is a standing issue around adding more db constraints to hud hmis tables: https://github.com/open-path/Green-River/issues/977

In any case, for HMIS we can be sure that household IDs are unique in the ds. (Aside from unlikely case that the same UUID gets generated twice)

drivers/hmis/app/graphql/mutations/join_household.rb
gigxz: I see, I missed that place where we are using it for the way finding step.

drivers/hmis/app/graphql/mutations/join_household.rb
gigxz: ok! all good I'm sure I wasn't clear on that myself!

drivers/hmis/app/graphql/mutations/join_household.rb
martha: Hm, on second thought I wonder if it may actually be better to resolve the whole household after all, so that it gets updated in the GraphQL cache.

Steps to reproduce cache bug:
- Search for a client in a mult-member household with a conflicting enrollment
- Click Add to Household and try to enroll them, get the Conflicting Enrollment alert, click Join Enrollments
- In the Select Clients step, de-select this client but select one (or more) of their household members.
- Proceed with the join workflow and submit
- Click "Return to (Hoh)'s enrollment" to close the dialog
- Click again on the "Add to Household" button for the client you originally searched, but deselected
- Initiate the workflow again
  - **Expected**: You should see their household updated correctly
  - **Observed**: On the Select Clients screen, the household falsely appears to contain the removed members. If you proceed with the workflow you will get an error, "Cannot join from multiple households"

This is fixed when the whole donor household is returned from the mutation

gigxz: Yeah that makes sense! sorry about my comment, I thought that the frontend didn't look at this data. But since it does, resolving the household makes sense!

gigxz (approved):
Looks good! lets hold off on merging until this is targeted to a release branch, so the merge history is clearer.

db/warehouse/migrate/20250121204933_add_hmis_household_events_table.rb
```diff
@@ -0,0 +1,22 @@
...
+class AddHmisHouseholdEventsTable < ActiveRecord::Migration[7.0]
+  def change
+    create_table :hmis_household_events do |t|
+      t.string :HouseholdID, null: false
+      t.references :data_source, null: false
+      t.string :event_type, null: false
+      t.jsonb :event_details
+      t.references :user, null: false
+      t.timestamps
+    end
```
gigxz, line 15: Tagging @eanders as an FYI on this new HMIS table which tracks changes to households.

For now, the events are "join" and "split", with the idea that we may also want to track "add" "remove" "hoh change" etc.

In the future, this could back a neat "household history" interface.

eanders: Thanks for the tag!  That sounds like a wonderful plan.

drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -64,25 +64,24 @@ def resolve(receiving_household_id:, joining_enrollment_inputs:)
...
+        )
+
+        # ...and one representing the SPLIT *out from* the donor (remaining) household
+        leaving_event = Hmis::HouseholdEvent.new_split_event(
+          user: current_user,
+          household: donor_household,
+          receiving_household_id: receiving_household_id,
+          before_state: donor_before_state,
+          after_state: donor_after_state,
+        )
```
gigxz, line 83: Nice!

gigxz (approved):

gigxz:
Can we merge this @martha?

martha:
@gigxz Join can be merged unless you would like to wait until Split is ready so they go out in the same release. I think I still need feedback on the Split frontend PR

martha:
Thank you @gigxz ! If it sounds good to you, after CI passes, I will merge all 4 PRs into release-152

gigxz:
great, thanks!

PR #5048, opened 2025-01-10
Hotfix for Delayed Job queueing

eanders:
## _Merging this PR_

## Description
* Fixes an issue where `with_advisory_lock` doesn't return true if the lock is obtained, but `nil`

## Type of change
* Bug fix

### DIFF
--- a/app/jobs/importing/hud_zip/fetch_and_import_job.rb
+++ b/app/jobs/importing/hud_zip/fetch_and_import_job.rb
@@ -24,7 +24,7 @@ module Importing::HudZip
         true
       end
 
-      requeue_at(Time.current + WAIT_MINUTES.minutes, "Import of Data Source: #{data_source_id} already running...re-queuing job for #{WAIT_MINUTES} minutes from now") unless lock_obtained
+      requeue_at(Time.current + WAIT_MINUTES.minutes, "Import of Data Source: #{data_source_id} already running...re-queuing job for #{WAIT_MINUTES} minutes from now") if lock_obtained == false
     end
 
     def known_classes
--- a/app/jobs/importing/hud_zip/fetch_and_import_job.rb
+++ b/app/jobs/importing/hud_zip/fetch_and_import_job.rb
@@ -18,13 +18,14 @@ module Importing::HudZip
       raise "Unknown import class: #{klass}; You must add it to the list of known classes in FetchAndImportJob" unless safe_klass.present?
 
       data_source_id = options[:data_source_id]
-      lock_obtained = GrdaWarehouse::DataSource.with_advisory_lock(advisory_lock_name(data_source_id), timeout_seconds: 60) do
+      lock_obtained = nil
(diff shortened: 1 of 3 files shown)


app/jobs/importing/hud_zip/fetch_and_import_job.rb
```diff
@@ -24,7 +24,7 @@ def perform(klass:, options:)
         true
       end
 
-      requeue_at(Time.current + WAIT_MINUTES.minutes, "Import of Data Source: #{data_source_id} already running...re-queuing job for #{WAIT_MINUTES} minutes from now") unless lock_obtained
+      requeue_at(Time.current + WAIT_MINUTES.minutes, "Import of Data Source: #{data_source_id} already running...re-queuing job for #{WAIT_MINUTES} minutes from now") if lock_obtained == false
```
gigxz: On my understanding of the documentation, this one shouldn't really be needed since the block above always returns `true`?

It seems like the issue would only arise if with_advisory_lock block itself returns nil, which maybe import! or `update_cached_counts` do, in the cases below.

https://github.com/ClosureTree/with_advisory_lock/tree/master?tab=readme-ov-file#return-values
Seem like the more appropriate usage might be using `with_advisory_lock_result` then asking `lock_was_acquired?`. Then we don't need to worry about what the block returns -- since there could be a case where the block DID run but it happened to return false. For example:

```ruby
      lock_obtained = GrdaWarehouse::DataSource.with_advisory_lock(advisory_lock_name(data_source_id), timeout_seconds: 60) do
        false # some operation that happens to return false
      end

      # no way to tell if lock was obtained or not, iiuc. would need to use with_advisory_lock_result instead.
```

gigxz (approved):
lgtm as a hotfix, left some comments for possible future change

app/jobs/importing/hud_zip/fetch_and_import_job.rb
eanders: Oh 🤦 .  That explains it.  Thank you!  
Updated PR coming

ttoomey (approved):
This seems fine to deploy as an urgent hotfix.
However testing locally it seems like `with_advisory_lock` returns the value of the block (which seems expected). Looks like we have that handled in fetch_and_import_job by returning true.

I think it would be clearer and more robust to assign to lock_obtained explicitly rather than using the return value from the block
``` ruby
lock_obtained = nil 
GrdaWarehouse::DataSource.with_advisory_lock(advisory_lock_name(data_source_id), timeout_seconds: 60) do
  # ...
  lock_obtained = true
end
requeue_at(Time.current + WAIT_MINUTES.minutes, "Import of Data Source: #{data_source_id} already running...re-queuing job for #{WAIT_MINUTES} minutes from now") unless lock_obtained

```

eanders:
Thanks @ttoomey, that's an even clearer solution than what I was thinking of, but yes, I had forgotten that it was returning the block.  Fix coming shortly

gigxz (approved):

PR #5051, opened 2025-01-13
Set timeout and maximumAge for PIT geolocation

gigxz:
## _Merging this PR_

## Description
Related Issue: https://github.com/open-path/Green-River/issues/6542

* Add 10s `timeout` so page doesn't infinitely appear "loading.." in some situation where browser fails to retrieve location. 
* Allow using cached locations up to 3 minutes. This is kind of arbitrary, but I figure if you are submitting multiple household members while staying in one location, it can speed things up or save battery by not requesting location repeatedly. Downside is that there is no obvious way for the user to "force" re-request location if it's returning a cached value, so that might be annoying.
* Add comment about enableHighAccuracy. Location accuracy is important, but I'm concerned about what I'm reading about performance and the effect on battery life.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
+++ b/drivers/hmis_external_apis/app/views/hmis_external_apis/external_forms/form/_geolocation.haml
@@ -97,7 +97,15 @@
       getLocationButton.click(function(event) {
         event.preventDefault();
         setLoading(true);
-        navigator.geolocation.getCurrentPosition(setPosition, onError);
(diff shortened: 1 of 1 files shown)


gigxz:
Hoping to get this into `release-147` despite it being new work, so that we can deploy it to production 1 week prior to PIT count. cc @eanders

eanders:
I'm good with that.

martha (approved):
Looks great. I think a maximumAge of 3 minutes seems reasonable.

ttoomey (approved):

PR #5054, opened 2025-01-13
Keep User Groups When Switching from Role-Based to ACLs

dtgreiner:
## _Merging this PR_

## Description
When switching a user account from Role-Based permissions to ACLs, before it was dropping any User Groups already tied to the user. This update will keep the user assigned to those user groups.

## Type of change
Bug fix

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -102,8 +102,10 @@ module Admin
           # The User Group data is not captured for update when using the Role-Based view. This means it will not be included
           # in the params when switching from Role-Based permissions to ACLs. In order to prevent wiping out any existing
           # user_group_id data, we need to ignore this param when changing to an ACL based permissions.
+          # The reverse is true for the access_group_ids field.
           params_to_update = user_params
           params_to_update = params_to_update.except(:user_group_ids) if changing_to_acls?
+          params_to_update = params_to_update.except(:access_group_ids) if changing_to_role_based?
           @user.update!(params_to_update)
 
           # if we have a user to copy user groups from, add them
@@ -157,6 +159,10 @@ module Admin
 
     private def changing_to_acls?
       params[:user][:permission_context] == 'acls' && @user.permission_context != params[:user][:permission_context]
+    end
+
+    private def changing_to_role_based?
+      params[:user][:permission_context] == 'role_based' && @user.permission_context != params[:user][:permission_context]
     end
 
     private def user_using_or_changing_to_acls?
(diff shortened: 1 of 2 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -80,13 +80,15 @@ def update
       existing_health_roles = @user.health_roles.to_a
       begin
         User.transaction do
+          changing_to_acls = params[:user][:permission_context] == 'acls' && @user.permission_context != params[:user][:permission_context]
```
eanders: Let's pull these out into methods, for now they can just live in the controller.  
```
private def changing_to_acls?
  params[:user][:permission_context] == 'acls' && @user.permission_context != params[:user][:permission_context]
end

private def user_using_or_changing_to_acls?
  @user.using_acls? || changing_to_acls
end
```

dtgreiner: This has been included in the PR [here](https://github.com/greenriver/hmis-warehouse/pull/5054/commits/1adc0ed345317d9670e097ba0dd97f9217d1b51c)

eanders (approved):
@dtgreiner do you think it would be time consuming (> 1 hour) to add some tests?  It might be nice to just confirm that this is doing what is intended.

dtgreiner:
@eanders I've added a test case for this scenario

spec/requests/admin/users_controller_spec.rb
```diff
@@ -47,5 +47,21 @@
         expect(updated_user.reload.notify_on_client_added).to be true
       end
     end
+
+    context 'when updating user to ACLs' do
```
eanders: Can you add another test that tests going in the other direction?  That checks for `user_groups`?

Also, thanks for this!

dtgreiner: > Can you add another test that tests going in the other direction? That checks for `user_groups`?

Just so I'm understanding correctly, if the user is changing from Role-Based -> ACLs or from ACLs -> Role-Based, we want to keep the `user_groups` associated with the user. Right now this is only being done in when moving to ACLs but it shouldn't be a difficult update to include.

eanders: Correct, when going from ACLs -> Legacy, it should preserve the users's `user_groups`.
Thanks!

dtgreiner: A new test has been included for the reverse case. Also, I was wrong about it not already keeping the user_groups when switching to role-based. Because this field is included in the ACL form, it is being sent to the controller and setting the value even when switching away from ACLs.

PR #5058, opened 2025-01-15
Handle Bad Encryption IDs

dtgreiner:
## _Merging this PR_

## Description

Handle bad encrypted id values by allowing the controllers to respond as if a record with a matching value was not found (404) instead of throwing an OpenSSL::Cipher::CipherError as that could be confusing to users. 

The cipher error is still sent to Sentry.

## Type of change
Bug fix

### DIFF
--- a/lib/util/id_protector.rb
+++ b/lib/util/id_protector.rb
@@ -15,10 +15,16 @@ class IdProtector
     Rails.application.routes.router.recognize(request) do |route, params|
       decoded_key = false
       params.each do |key, value|
-        if key == :id || key.to_s.ends_with?('_id')
+        next unless key == :id || key.to_s.ends_with?('_id')
+
+        begin
           params[key] = ProtectedId::Encoder.decode(value)
-          decoded_key = true if value != params[key]
+        rescue OpenSSL::Cipher::CipherError => e
+          # Suppress Cipher Errors so the response is handled by the controller as an unfound id.
+          # Still capture the error in Sentry.
+          Sentry.capture_exception(e)
         end
+        decoded_key = true if value != params[key]
       end
       if decoded_key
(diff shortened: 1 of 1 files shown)


eanders (approved):
That looks good to me, thank you!

lib/util/id_protector.rb
```diff
@@ -15,10 +15,16 @@ def call(env)
...
+
+        begin
           params[key] = ProtectedId::Encoder.decode(value)
-          decoded_key = true if value != params[key]
+        rescue OpenSSL::Cipher::CipherError => e
+          # Suppress Cipher Errors so the response is handled by the controller as an unfound id.
+          # Still capture the error in Sentry.
+          Sentry.capture_exception(e)
         end
+        decoded_key = true if value != params[key]
```
eanders, line 27: Should this raise a NotFound error if the value and params[key] match?  Or where is that handled?

dtgreiner: That should be the same as if the `PROTECTED_IDS` setting is not being used and an invalid key is passed. Unless the controller is specifically handling it differently (e.g. [client controller](https://github.com/greenriver/hmis-warehouse/blob/a5ef47e91721eade875a52a3c687f434fb197bee/drivers/client_access_control/app/controllers/client_access_control/client_concern.rb#L14)) , it should be throwing an ActiveRecord::RecordNotFound error on the controller which responds as a 404.

Example from local environment for Data Source Controller. 
Screenshot 2025-01-15 at 1 07 12 PM

Non dev environments should show the 404 page.
Screenshot 2025-01-15 at 1 10 32 PM

eanders: Thanks! Just wanted to make sure.

PR #5062, opened 2025-01-18
Import alerting and monitoring

eanders:
## _Merging this PR_

## Description

## Type of change


dtgreiner (approved):

app/models/importers/hmis_auto_migrate/local.rb
```diff
@@ -37,6 +39,7 @@ def import!
     end
 
     delegate :loader_log, to: :@importer
+    delegate :importer_log, to: :@importer
```
ttoomey: small housekeeping item but it looks like we can use the `importer` accessor that was added
```suggestion
    delegate :importer_log, to: :importer
```

app/models/notification_types/import_count_change_threshold.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# This notification should be called when an import is paused.
+# Imports are paused when they reach a set threshold of errors or changes
+module NotificationTypes
+  class ImportRecordCountChangeThreshold < Base
+    has_many :notification_configurations
```
ttoomey: `has_many` implies this is an active record model. But `Base` does not inherit from active record.

app/models/notification_types/import_error_count_threshold.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# This notification should be called when an import is paused.
+# Imports are paused when they reach a set threshold of errors or changes
+module NotificationTypes
+  class ImportErrorCountThreshold < Base
+    has_many :notification_configurations
```
ttoomey: `has_many` implies this is an active record model. But `Base` does not inherit from active record.

app/models/notification_types/import_count_change_threshold.rb
```diff
@@ -0,0 +1,13 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# This notification should be called when an import is paused.
+# Imports are paused when they reach a set threshold of errors or changes
+module NotificationTypes
+  class ImportRecordCountChangeThreshold < Base
```
ttoomey: the class name does not match the file name so Zeitwerk doesn't load the class

app/models/concerns/service_history/builder.rb
```diff
@@ -121,7 +121,8 @@ def clients_still_processing?(client_ids:)
 
     # Class method
     private def builder_batch_job_scope
-      Delayed::Job.where(failed_at: nil).jobs_for_class('ServiceHistory::RebuildEnrollments')
+      Delayed::Job.where(failed_at: nil).jobs_for_class('ServiceHistory::RebuildEnrollments').
+        where("1!=#{rand(2..50)}") # cache buster, postgres heavily caches this query incorrectly.
```
ttoomey: Postgres is returning incorrect result for the query? That seems like it would break a lot of things. Are you sure it's not rails query cache? In which case maybe use [`uncached`](https://api.rubyonrails.org/classes/ActiveRecord/QueryCache/ClassMethods.html)
```
Delayed::Job.uncached do
  Delayed::Job.where(failed_at: nil).jobs_for_class('ServiceHistory::RebuildEnrollments')
end
```

db/warehouse/migrate/20250117174547_convert_data_source_config_to_import_threshold.rb
```diff
@@ -0,0 +1,20 @@
+class ConvertDataSourceConfigToImportThreshold < ActiveRecord::Migration[7.0]
+  def up
+    GrdaWarehouse::DataSource.where(refuse_imports_with_errors: true).find_each do |ds|
+      # Create and update in individual queries, DS count should never be more than 50
+      threshold = ds.import_threshold || GrdaWarehouse::ImportThreshold.new(data_source_id: ds.id)
+      threshold.error_count_min_threshold ||= 0
+      threshold.error_percent_threshold ||= 0
+      threshold.pause_on_error_threshold = true
+      threshold.save!
+      ds.update(refuse_imports_with_errors: false)
```
ttoomey, line 10: the intention is to deprecate this column, right? can we rename it if that's the case?
```
    rename_column :data_sources, :refuse_imports_with_errors, :deprecated_refuse_imports_with_errors
```

app/views/data_sources/_buttons.haml
```diff
@@ -20,6 +20,10 @@
     = link_to new_data_source_hmis_import_config_path(data_source_id: @data_source), class: 'btn btn-secondary btn-sm mb-4 mr-2' do
       %i.icon-file-drawn
       Automate HMIS CSV Loads
+  .config-button-wrapper
+    = link_to data_source_import_threshold_path(data_source_id: @data_source), class: 'btn btn-secondary btn-sm mb-4 mr-2' do
+      %i.icon-warning
```
ttoomey: warning icons usually indicate a condition the user needs to pay attention to. But that's not the case here right?

app/views/import_thresholds/_table.haml
```diff
@@ -0,0 +1,20 @@
+- items = import_threshold.items_for(slug)
+.card
+  %table.table.table-striped
+    %thead
+      %tr
+        %th Name
+        %th email
```
ttoomey:
```suggestion
        %th Email
```

app/views/import_thresholds/show.haml
```diff
@@ -0,0 +1,41 @@
...
+
+= simple_form_for import_threshold, url: data_source_import_threshold_path, method: :patch do |f|
+  .well
+    %h2 Changes in Record Counts
+    %p The following settings apply when an import significantly changes the number of rows in the data that already exists in the warehouse for the data source.  This will look at the aggregate change, so if 100 rows are removed and 100 added, no change is indicated.
+    .row
+      .col
+        = f.input :record_count_change_percent_threshold, label: 'Record count change threshold (%)', hint: 'Imports will pause and alert if more than the chosen percentage have changed', input_html: { style: 'width: 5em' }
+      .col
+        = f.input :record_count_change_min_threshold, label: 'Minium change count threshold', hint: 'Imports will not pause or alert below this threshold.  This is designed to prevent alerting on a 30% change when only 1 of 3 rows changes.', input_html: { style: 'width: 8em' }
```
ttoomey: Typo. TIL "Minium is a red or orange pigment made of lead"
```suggestion
        = f.input :record_count_change_min_threshold, label: 'Minimum change count threshold', hint: 'Imports will not pause or alert below this threshold.  This is designed to prevent alerting on a 30% change when only 1 of 3 rows changes.', input_html: { style: 'width: 8em' }
```

app/views/import_thresholds/show.haml
```diff
@@ -0,0 +1,41 @@
...
+        %i.icon-plus
+        Add Notification
+  .well
+    %h2 Error Counts
+    %p The following settings apply when an import contains a significant number of errors or validation issues.  The determination of change is completely contained to the import in question, looking at the errors in a given table against the number of records being processed.
+    .row
+      .col
+        = f.input :error_percent_threshold, label: 'Error count threshold (%)', hint: 'Imports will pause and alert if more than the chosen percentage of rows have errors', input_html: { style: 'width: 5em' }
+      .col
+        = f.input :error_count_min_threshold, label: 'Minium error count threshold', hint: 'Imports will not pause or alert below this threshold.  This is designed to prevent alerting on a 30% error rate when only 1 of 3 rows contains an error.', input_html: { style: 'width: 8em' }
```
ttoomey:
```suggestion
        = f.input :error_count_min_threshold, label: 'Minimum error count threshold', hint: 'Imports will not pause or alert below this threshold.  This is designed to prevent alerting on a 30% error rate when only 1 of 3 rows contains an error.', input_html: { style: 'width: 8em' }
```

app/views/import_thresholds/show.haml
```diff
@@ -0,0 +1,41 @@
+= render 'data_sources/breadcrumbs', data_source: data_source
+- content_for :title, 'Import Thresholds'
+%h1= content_for :title
+
+%p Thresholds can be configured below to pause and alert when certain criteria are met during imports for this data source.
```
ttoomey: This is orienting text for the user? Maybe replace passive with imperative to direct the use as to what they are doing.
```suggestion
%p Configure thresholds that will trigger alerts and optionally pause imports when specific criteria are met for this data source.
```

app/views/notification_configurations/create.js.erb
```diff
@@ -0,0 +1,9 @@
+<% if notification_configuration.errors.any? %>
+  console.log(<%= notification_configuration.errors.inspect %>)
```
ttoomey: remove debugging?
```suggestion
```

app/views/notification_configurations/destroy.js.erb
```diff
@@ -0,0 +1,7 @@
+<% if notification_configuration&.errors&.any? %>
+
+<% else %>
```
ttoomey: cleanup. Also is notification_configuration ever nil? maybe we don't need safe navigation operator?

```suggestion
<% if notification_configuration.errors.blank? %>

```

app/controllers/notification_configurations_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+      permit(:user_id, :active)
+  end
+
+  private def data_source_source
+    GrdaWarehouse::DataSource.viewable_by(current_user, permission: :can_view_projects)
+  end
+
+  private def data_source_scope
+    data_source_source.source
+  end
```
ttoomey: This seems like we could improve clarity here; perhaps we can reduce the use of the term 'source' which seems overloaded: We have data sources that also have a scope called source, we also have 'data_source_source' as a method here, we also have a polymorphic association on NotificationConfiguration called source. 

Maybe it would help to get rid of data_source_source and just have data_source_scope? Or if we need both, add a comment as to why

app/controllers/notification_configurations_controller.rb
```diff
@@ -0,0 +1,70 @@
...
+  helper_method :data_source
+
+  private def import_threshold
+    @import_threshold ||= data_source.import_threshold
+  end
+  helper_method :import_threshold
+
+  def notification_configuration
+    @notification_configuration ||= if params[:id].to_i.positive?
+      GrdaWarehouse::NotificationConfiguration.find(params[:id].to_i)
```
ttoomey: Is there a concern with negative ids or something? Do we need to cast to int?
```suggestion
    @notification_configuration ||= if params[:id]
      GrdaWarehouse::NotificationConfiguration.find(params[:id])
```

app/models/grda_warehouse/data_source.rb
```diff
@@ -568,6 +571,94 @@ def self.options_for_select(user:)
...
+  #
+  # This method checks if an import threshold is set and whether the number of errors
+  # exceeds the allowed limit. If so, it returns `true`, indicating that imports should
+  # be paused.
+  #
+  # @param total [Integer] The total number of rows in the import file. Required.
+  # @param errors [Integer] The number of errors encountered. Required.
+  # @return [Boolean] `true` if imports should be paused due to errors, otherwise `false`.
+  #
+  memoize def pause_imports_with_errors?(total:, errors:)
```
ttoomey, line 589: This is fine but the data_source model is creeping up to 1000 lines. Maybe consider if these methods could live in service object instead to better encapsulate and organize functionality

app/models/grda_warehouse/data_source.rb
```diff
@@ -568,6 +571,94 @@ def self.options_for_select(user:)
...
+    return unless import_threshold.present?
+
+    import_threshold.send_status_notifications(import_log_id, error_threshold_met, record_count_threshold_met, paused)
+  end
+
+  def ever_notify_for_imports?
+    return false unless import_threshold.present?
+
+    # Do we have any thresholds set?
+    import_threshold.record_count_change_percent_threshold.present? || import_threshold.error_percent_threshold.present?
```
ttoomey: I notice we have the pattern of using 'present?' everywhere. But these are integer fields right? Is there some value that might occur where presence would matter as opposed to just checking for falsyness? It doesn't seem like these would ever be empty enumerable-like values

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_logs/_import_log.haml
```diff
@@ -24,6 +20,9 @@
             %tr
               %th Imported
               %td= import_time
+            %tr
+              %th Post-Processing
+              %td= @import&.importer_log&.post_processing_time(@import)
```
ttoomey, line 25: Should there be a preload to avoid n+1? Maybe doesn't matter that much if number of rows is small

app/controllers/import_thresholds_controller.rb
```diff
@@ -0,0 +1,43 @@
...
+      permit(*GrdaWarehouse::ImportThreshold.known_params)
+  end
+
+  private def data_source_source
+    GrdaWarehouse::DataSource.viewable_by(current_user, permission: :can_view_projects)
+  end
+
+  private def data_source_scope
+    data_source_source.source
+  end
```
ttoomey: is this used?

app/models/grda_warehouse/import_threshold.rb
```diff
@@ -0,0 +1,189 @@
...
+      return false if percent_change < error_percent_threshold
+
+      true
+    end
+
+    # Gather users from notification configurations into 3 categories
+    # 1. Those that have notifications for errors
+    # 2. Those that have notifications for count changes
+    # 3. Those that have both
+    def send_status_notifications(import_log_id, error_threshold_met, record_count_threshold_met, paused)
```
ttoomey: I think keyword args are a better choice than positional once you get past one or two args. That makes it less likely that someone might call it incorrectly.
```suggestion
    def send_status_notifications(import_log_id:, error_threshold_met:, record_count_threshold_met:, paused:)
```

Also this method is kind of long

app/models/grda_warehouse/import_threshold.rb
```diff
@@ -0,0 +1,189 @@
...
+require 'memery'
+
+# This defines two thresholds for imports that will potentially trigger notifications
+# 1. When an import is going to add or remove a significant percentage of the existing records, and that change is greater than the min count threshold
+# 2. When an import contains a high percentage of errors, and that change is greater than the min count threshold
+module GrdaWarehouse
+  class ImportThreshold < GrdaWarehouseBase
+    include Memery
+    belongs_to :data_source
+
```
ttoomey, line 16: maybe could use validation?

```suggestion

validates :record_count_change_min_threshold, numericality: { greater_than_or_equal_to: 0 }
validates :record_count_change_percent_threshold, numericality: { greater_than_or_equal_to: 0, less_than_or_equal_to: 100 }

```

app/models/grda_warehouse/import_threshold.rb
```diff
@@ -0,0 +1,189 @@
...
+    # Gather users from notification configurations into 3 categories
+    # 1. Those that have notifications for errors
+    # 2. Those that have notifications for count changes
+    # 3. Those that have both
+    def send_status_notifications(import_log_id, error_threshold_met, record_count_threshold_met, paused)
+      return unless error_threshold_met || record_count_threshold_met
+
+      receive_both_user_ids = error_count_notification_user_ids & record_change_count_notification_user_ids
+
+      # Notify where the user receives both notifications
```
ttoomey:
```suggestion
      # Handle users who subscribe to both types - they get one notification containing both status types 
```

app/models/grda_warehouse/import_threshold.rb
```diff
@@ -0,0 +1,189 @@
...
+          NotifyUser.with(
+            user: user,
+            import_log_id: import_log_id,
+            data_source: data_source,
+            error: false, # never notify on errors in this scenario
+            count: record_count_threshold_met,
+            paused: paused,
+          ).import_processing.deliver_later
+        end
+      end
```
ttoomey, line 129: What happens if we get here and haven't notified anyone (no active users to receive notifications). I guess that's okay? Or is should there be a fallback behavior?

app/models/notification_types/import_paused.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# This notification should be called when an import is paused.
+# Imports are paused when they reach a set threshold of errors or changes
+module NotificationTypes
+  class ImportPaused < Base
+    has_many :notification_configurations
```
ttoomey: doesn't seem like an active record model. Maybe these are development cruft?

db/warehouse/migrate/20250116145506_create_notification_configuration.rb
```diff
@@ -0,0 +1,24 @@
+class CreateNotificationConfiguration < ActiveRecord::Migration[7.0]
+  def change
+    create_table :notification_configurations do |t|
+      t.references :user, null: false
+      t.references :source, polymorphic: true, null: false
+      t.string :notification_slug, null: false, description: 'Class name for notification logic'
```
ttoomey, line 6: the term 'slug' kind of trips me up here... usually a slug uniquely identifies a unique resource in a url. But this seems to be more of an event type. Also the class names don't appear to be usable. 
```suggestion
      t.string :notification_event, null: false, description: 'count_threshold_exceeded' | 'error_threshold_exceeded'
```

drivers/hmis_csv_importer/spec/models/importer/twenty_twenty_four/import_threshold_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+      before(:all) do
+        HmisCsvImporter::Utility.clear!
+        GrdaWarehouse::Utility.clear!
+
+        threshold = FactoryBot.create(:import_threshold, pause_on_error_threshold: true)
+        @data_source = threshold.data_source
+        @user = FactoryBot.create(:user)
+        FactoryBot.create(:notification_configuration_import_threshold, :import_error_count_slug, user: @user, source: threshold)
+
+        travel_to Time.local(2020, 1, 1) do
```
ttoomey, line 23: add a comment as to why we need to time-travel?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -792,13 +1023,13 @@ def complete_import
...
-        importer_log.save
-        data_source.update(last_imported_at: Time.current)
+        importer_log.save!
+        data_source.update!(last_imported_at: Time.current)
         elapsed = importer_log.completed_at - @started_at
         Rails.logger.tagged({ task_name: 'HMIS CSV Importer', repeating_task: true, task_runtime: elapsed }) do
           log("Completed importing in #{elapsed_time(elapsed)} #{hash_as_log_str log_ids}.  #{summary_as_log_str(importer_log.summary)}")
         end
-        @import_log&.update(importer_log: importer_log)
+        @import_log&.update!(importer_log: importer_log)
```
ttoomey, line 1022: thanks for catching these!

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+
+      # counts of expected rows in each file
+      totals_by_filename = @loader_log.summary.map { |filename, data| [filename, data['total_lines'].to_i] }.to_h
+      totals_by_filename.each do |filename, total_count|
+        met = @data_source.error_count_threshold_reached?(total_count, error_counts[filename])
+
+        return true if met
+      end
+      # No threshold met
+      false
```
ttoomey: simplification?
```suggestion
      totals_by_filename.any? do |filename, total_count|
        @data_source.error_count_threshold_reached?(total_count, error_counts[filename])
      end
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+    #
+    # The decision to pause is made based on the following conditions:
+    # - If the data source is configured to pause imports due to errors or record changes.
+    # - If the loader log status is not `'loaded'`, indicating the import has not progressed to that stage.
+    # - If any individual file has exceeded its error threshold, the import is paused.
+    #
+    # @return [Boolean] `true` if the import should be paused due to exceeding error thresholds, otherwise `false`.
+    #
+    memoize def should_pause?
+      return false unless @data_source.ever_pause_imports? # should we ever pause?
```
ttoomey: sorry, I don't understand the need for this. We check these same conditions below, why check twice? ( ever_pause_imports_with_errors? and ever_pause_imports_with_record_changes?)

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
       return true unless @loader_log.status == 'loaded'
 
-      loader_errors = @loader_log.summary.values.sum { |h| h['total_errors'].to_i }
+      return true if any_error_thresholds_met? && @data_source.ever_pause_imports_with_errors?
+
+      any_record_count_thresholds_met? && @data_source.ever_pause_imports_with_record_changes?
```
ttoomey: micro optimization, re-order conditionals so expensive operation is last
```suggestion
      return true if @data_source.ever_pause_imports_with_errors? && any_error_thresholds_met?

      @data_source.ever_pause_imports_with_record_changes? && any_record_count_thresholds_met?
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+
+      change_counts.each_value do |data|
+        total = data[:total_count]
+        changes = data[:change_count]
+        met = @data_source.record_count_threshold_reached?(total, changes)
+
+        return true if met
+      end
+      # No threshold met
+      false
```
ttoomey: simplification? 
```suggestion
      change_counts.values.any? do |data|
        total = data[:total_count]
        changes = data[:change_count]
        @data_source.record_count_threshold_reached?(total, changes)
      end
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+    #
+    # It maps error counts to their respective filenames and ensures that the counts
+    # are aggregated correctly without raising additional errors if files are missing.
+    #
+    # @return [Hash{String => Integer}] A hash where keys are filenames and values are the total error counts for each file.
+    #
+    memoize private def error_counts
+      file_lookup = self.class.importable_files_map.invert
+
+      summary = @loader_log.summary
```
ttoomey, line 188: for my own understanding, I think this is right?
```suggestion
      # serialized array persisted on the log model
      summary = @loader_log.summary
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
         importer_log_id: importer_log.id,
-      )
+      ).
+        group(:source_type).
+        count.
+        map do |k, count|
+          # convert to filename keys
+          filename = file_lookup[k.demodulize]
+          # every file _should_ be present, but let's try not to throw any errors here
+          next unless counts[filename]
```
ttoomey: So if the summary included no errors for this file name, we exclude import errors? Is this actually something that could occur?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+    # to be used in import threshold checks before committing data changes.
+    #
+    # @return [void]
+    #
+    private def precalculate_change_counts
+      # This makes an estimate of the changes that will occur as it needs to be done before the
+      # actual processing so that it can pause the import if necessary.
+      # NOTE: as written this causes 2 additional, potentially slow queries
+      importable_files.map do |file, klass|
+        incoming_data = klass.incoming_data(importer_log_id: importer_log.id).
```
ttoomey, line 240: can this be non-unique at this stage?
```suggestion
        incoming_data = klass.incoming_data(importer_log_id: importer_log.id).distinct.
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+          pluck(klass.hud_key).to_set
+
+        to_add = (incoming_data - existing_hud_keys(klass)).count
+        to_remove = (existing_hud_keys(klass) - incoming_data).count
+        # If we never delete, just pretend it will be 0
+        to_remove = 0 if klass.prevent_import_deletions?
```
ttoomey: conditional assignment to avoid unnecessary `to_remove` calculation
```suggestion
        # If we never delete, just pretend it will be 0
        to_remove = 0 if klass.prevent_import_deletions?
        to_remove ||= (existing_hud_keys(klass) - incoming_data).count

```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -101,22 +108,224 @@ def resume!
...
+    #
+    private def change_counts
+      importer_log.summary.map do |file, data|
+        klass = importable_files[file]
+        to_add = data['added']
+        to_remove = data['removed']
+        [
+          file,
+          {
+            change_count: (to_add - to_remove).abs,
```
ttoomey, line 292: I am probably not understanding but I'm unsure about this method of calculating changes. If I have 100 records, I remove all 100 and add another 100, I effectively have 0 changes even though I have replaced the entire record set? That seems expected.

I think summing them would be more intuitive, no? Or maybe max(inserts, deletions)

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -301,7 +510,10 @@ def cleanup_data_set!
     # GrdaWarehouse::Tasks::ServiceHistory::Enrollment.batch_process_date_range!(range)
     # In here, add history_generated_on date to enrollment record
     def ingest!
+      # Reset add/remove counts used for import thresholds
+      reset_import_counts
       importer_log.update(status: :importing)
+      reset_import_counts
```
ttoomey: do we actually need to reset this both before and after? It's harmless I guess, just unexpected

ttoomey (changes requested):
So sorry for the huge pile of feedback. Much of it is just suggestions and can be ignored. But it seems like we have some stray models hanging around that should be removed

drivers/hmis_csv_importer/app/views/hmis_csv_importer/import_logs/_import_log.haml
eanders: an import only has one importer log, which should be loaded already and the `post_processing_time` doesn't appear to load additional data (I may be missing something, but I don't think we need a preload here.)

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: No!  Thanks for catching that.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: Maybe the max of the two.  Generally speaking it's ok if we "replace" them.  What we're trying to catch is where the overall magnitude of the change is large.  If the sending HMIS is just renumbering the HUD Keys, but the overall change will be small, we probably don't care.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: Thanks!

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: I don't believe it would ever be non-unique at this stage.  I'm converting to a set for faster comparisons.  What made you think we need a distinct?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: This should never happen.  I think it may have been left from a previous iteration where counts might not include every file as it was based on the files present at some step.  I'll remove.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: Cool, I assumed that wouldn't short-circuit, and didn't want to risk running the calculations two dozen times when only need a handful.  Turns out I was wrong!
```
irb(main):001:0> [1, 2, 3, 4].any? { |m| puts m; m == 2 }
1
2
=> true
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
eanders: This was here because I had the checks below mis-ordered and I was optimizing to not run `any_error_thresholds_met?` or `any_record_count_thresholds_met?` if we _never_ needed them.  Your solution is much better, the checks below check ask the data source (which asks the import_threshold) this question before running the expensive calculation.

db/warehouse/migrate/20250117174547_convert_data_source_config_to_import_threshold.rb
eanders: That is the intention.  Adding to ignored columns on the data source.  We'll have to circle back after this is deployed and remove it.

db/warehouse/migrate/20250116145506_create_notification_configuration.rb
eanders: Currently `notification_slug` is either `NotificationTypes::ImportRecordCountChangeThreshold` or `NotificationTypes::ImportErrorCountThreshold`.  Maybe I'm missing something.  We could change it to `notification_event`, I'm not totally opposed to that, let me see how invasive that would be.

app/views/data_sources/_buttons.haml
eanders: I was using that here to indicate you are setting up a threshold that will trigger warnings.  Happy to use something else, but I didn't see anything obvious.  Maybe a bell?

app/models/grda_warehouse/import_threshold.rb
eanders: Totally ok, that's the current behavior.

app/models/grda_warehouse/data_source.rb
eanders: I think this comes from a history of 0 being falsey in some languages and a desire to be explicit that we are checking that there is a value set.  I know it's bitten us in the past, though I'm not remembering exactly where.  In this case, I agree, we don't need `.present?`

app/controllers/notification_configurations_controller.rb
eanders: `to_i` enforces that we aren't getting garbage in the params.  If we call `to_i` on something other than an integer, we get `0`, so `positive?` ensures we didn't get garbage.  Historically, we've avoided passing un-sanitized params to `find`.  Maybe that's no longer necessary?

app/controllers/notification_configurations_controller.rb
eanders: Yep, I think there was some weird copy/paste in there.  Thanks for catching.  That method wasn't actually used.

app/views/data_sources/_buttons.haml
ttoomey: Yes, a bell sounds like a better choice, thanks!

somehow added this to wrong comment, ignore ~~Just checking the assumption. If we had dupes it might throw off the calculation. But sounds like that's not a concern, thanks~~

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
ttoomey: thanks, just wanted to check the assumption since it would throw of the calculation.

app/models/grda_warehouse/data_source.rb
eanders: This is definitely something we should tackle.  I think I want to get this feature out and revisit.

app/models/concerns/service_history/builder.rb
eanders: I think that solved it!  Thank you.  @dtgreiner , I guess there was a built-in way to prevent this.

@ttoomey the blue CACHE is what we were fighting.
Screenshot 2025-01-31 at 10 23 14 AM

eanders:
@ttoomey I _think_ I addressed all of your wonderful comments (or at least the ones I thought needed to be addressed.)  Can you take another look?

app/controllers/notification_configurations_controller.rb
ttoomey: My understanding is the rails convention is not to sanitize the id param for find and let active record handle it; it should handle the cast for you. This is nice because it is less code to write and, while pk's are usually integers, you're not coding yourself out of uuids.

I'm not opposed to extreme paranoia when handling params. However, if we want to go down that road, I think we should dry it up. For example:

```ruby
class ApplicationRecord
  def self.find_safely(tainted_id)
    safe_id = begin
      Integer(tainted_id)
    rescue ArgumentError, TypeError
      nil
    end
    raise(ActiveRecord::RecordNotFound, "Record not found for ID: #{tainted_id}") unless safe_id?.positive?
    find(safe_id)
  end
end
```

app/models/grda_warehouse/data_source.rb
```diff
@@ -568,6 +574,99 @@ def self.options_for_select(user:)
...
+  memoize def pause_imports_with_errors?(total:, errors:)
+    return false unless import_threshold.present?
+    return false unless import_threshold.pause_on_error_threshold
+
+    error_count_threshold_reached?(total, errors)
+  end
+
+  def ever_pause_imports?
+    ever_pause_imports_with_errors? || ever_pause_imports_with_record_changes?
+  end
```
ttoomey: this probably can be removed, it's called in HmisCsvImporter::Importer but doesn't really need to be

ttoomey (approved):
Looks good, thanks for all the updates!

PR #5067, opened 2025-01-23
7229 enforce validation on Account creation; allow invalid accounts to be deactivated

ttoomey:
## _Merging this PR_

## Description
[GH Issue](https://github.com/open-path/Green-River/issues/7229)

changes:
- enforce email validation on account creation, not just update
- allows accounts to be disabled even if they are not valid (due to bad email addresses)
- create accounts in a transaction so we don't create partial or incomplete accounts
- fixes double-render error when trying to save changes to an admin user under some circumstances
- add tests

## Test instructions
- as an admin, try to create an account with an invalid email address domain (foo@some-crazydomainthatdoesntexist). It should show you the form and prompt to fix the issue. A new account should not be created
- find an user with an invalid email address. Attempt to deactivate the user, it should succeed. 

## Type of change
Bugfix

### DIFF
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -145,7 +145,8 @@ module Admin
 
     def destroy
       @user.paper_trail_event = 'deactivate'
-      @user.update(active: false)
+      # update_column() allows us to update the user even if the record is invalid
+      @user.update_column(:active, false)
       redirect_to({ action: :index }, notice: "User #{@user.name} deactivated")
     end
 
--- a/app/controllers/admin/users_controller.rb
+++ b/app/controllers/admin/users_controller.rb
@@ -56,7 +56,10 @@ module Admin
 
       @redirecting = true
       update
-      redirect_to({ action: :edit }, notice: 'User updated') and return
+      # early return if the response body was set by update(), avoid double-render error
+      return if performed?
+
+      redirect_to({ action: :edit }, notice: 'User updated')
     end
 
     def impersonate
--- a/app/controllers/users/invitations_controller.rb
+++ b/app/controllers/users/invitations_controller.rb
@@ -23,6 +23,7 @@ class Users::InvitationsController < Devise::InvitationsController
 
   # POST /resource/invitation
   def create
+    # if the account will be an admin, check that the current user has confirmed this action with their password
     if creating_admin? && current_user.confirm_password_for_admin_actions? && !current_user.valid_password?(confirmation_params[:confirmation_password])
       flash[:error] = 'User not updated. Incorrect password'
       @user = User.new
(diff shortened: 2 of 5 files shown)


app/controllers/admin/users_controller.rb
```diff
@@ -142,7 +145,8 @@ def update
 
     def destroy
       @user.paper_trail_event = 'deactivate'
-      @user.update(active: false)
+      # update_column() allows us to update the user even if the record is invalid
+      @user.update_column(:active, false)
```
ttoomey, line 149: Note; this would not trigger callbacks on user. Checked with Elliot and it sounded like there isn't anything we need to do when the user is deactivated

app/models/concerns/user_concern.rb
```diff
@@ -60,7 +60,7 @@ module UserConcern
     after_save :create_access_group
     # END_ACL
 
-    validates :email, presence: true, uniqueness: true, email_format: { check_mx: true }, length: { maximum: 250 }, on: :update
+    validates :email, presence: true, uniqueness: true, email_format: { check_mx: true }, length: { maximum: 250 }
```
ttoomey, line 63: @eanders - wanted to call out this change in behavior. It seems unexpected that we are not validating the email when the user is created. If this seems to risky we can make it conditional using an ephemeral attribute on user to control it
```ruby
@user.ensure_email_is_validated = true
@user.save
```

app/controllers/users/invitations_controller.rb
```diff
@@ -23,32 +23,52 @@ def confirm
...
   end
 
+  private def invite_and_find_or_create_user
+    # if a deleted account matches the email, restore it so devise's invitable will find it
+    email = invite_params[:email]
+    User.with_deleted.find_by_email(email)&.restore
+
+    # will find or create a user record
+    user = User.invite!(invite_params.except(:legacy_role_ids), current_user) do |u|
+      # I guess we use invite for the side-effect of user creation in this case
```
eanders:
```suggestion
      # If the user creating the account explicitly indicated they didn't need to send an invitation 
      # skip sending the invitation.  This is used when bulk creating accounts ahead of when 
      # someone should be given access to the application.
```

app/controllers/users/invitations_controller.rb
```diff
@@ -23,32 +23,52 @@ def confirm
...
+      u.skip_invitation = invite_params[:skip_invitation]&.to_i == 1
+    end
+    return user if user.errors.present?
+
+    # Roles need to be added as a second pass
```
eanders:
```suggestion
    # Roles need to be added as a second pass since devise invitable doesn't know how to handle them
```

app/models/concerns/user_concern.rb
eanders: @gigxz do you know if we ever create users without a valid email intentionally?  I think maybe for OKTA integrations, or imports?  I agree with @ttoomey we should be validating on create, but maybe that's not helpful in some scenarios.

spec/controllers/users/invitations_controller_spec.rb
```diff
@@ -0,0 +1,62 @@
+require 'rails_helper'
+
+RSpec.describe Users::InvitationsController, type: :controller do
+  let(:admin_user) { create(:acl_user) }
+  let(:admin_role) { create :admin_role }
+  let(:no_data_source_collection) { create :collection }
+
+  let(:agency) { create(:agency) } # Assuming a factory for Agency
```
eanders:
```suggestion
  let(:agency) { create(:agency) } 
```

eanders (approved):
Thanks @ttoomey !  
I left a few thoughts, and I think we should retarget 149.

app/models/concerns/user_concern.rb
gigxz: No, not that I know of. For both Okta and import scenarios, I think it makes sense to validate the email right away. The only different on account creation for Okta is that we can skip sending the initial email invitation (if "Skip email invitation" checked). The email address still needs to be entered initially and should be valid.

I see that on the AccountRequest model, there is a similar validation for email that is validated on `create`. So maybe this is already something we validate in a different place? I don't know how the account request model is used though. https://github.com/greenriver/hmis-warehouse/blob/release-148/app/models/account_request.rb#L11

ttoomey: > The only different on account creation for Okta is that we can skip sending the initial email invitation

I don't think this validation would impact sending an invitation. I could be wrong but it seems like invitations are sent by manually calling `User.invite!` - we have a different path for user creation in `OmniauthUserProvisioner`

In fact, the omniauth provisioner skips all validations, which would include this email validation.

>  I don't know how the account request model is used though.

Good find; I didn't realize that we support account requests. I wonder if it's still used?

PR #5068, opened 2025-01-23
Bugfix and cleanup in AC Warehouse custom data element exporter

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7094

- Remove "Pathways" from the custom data element export. It is already exported in another file which they prefer to use.
- Add logic to CDE export to translate Source Client IDs into Destination IDs before exporting. Source Client IDs are not understood by the party importing because they only have destination ids.
  - This is only applicable to CDEs where the owner type is client. This is future-proofing for future CDEs more than anything, since we are removing the Pathways. There is one remaining Client-tied CDE in this file which is the "has active referrals in LINK", which I don't believe is being looked at.

## Type of change
Bug fix, cleanup

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/cde_export.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/cde_export.rb
@@ -78,12 +78,13 @@ module HmisExternalApis::AcHmis::Exporters
     end
 
     # { source client id => destination warehouse id }
-    # only includes source clients that are the Owner of any CDEs in scope
+    # only includes source clients that are the Owner of any CDEs
     def client_id_to_warehouse_id
-      @client_id_to_warehouse_id ||= begin
-        source_client_ids = cdes.where(data_element_definition: { owner_type: 'Hmis::Hud::Client' }).pluck(:owner_id)
-        Hmis::Hud::Client.where(id: source_client_ids).joins(:warehouse_client_source).pluck(c_t[:id], wc_t[:destination_id]).to_h
-      end
+      @client_id_to_warehouse_id ||=
+        Hmis::Hud::Client.where(data_source: data_source).
+          joins(:custom_data_elements).    # Only include clients that have CDEs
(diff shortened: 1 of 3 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/cde_export.rb
```diff
@@ -59,13 +66,26 @@ def columns
...
 
     def walk_in_cded
       @walk_in_cded ||= Hmis::Hud::CustomDataElementDefinition.where(data_source: data_source, key: :direct_entry, owner_type: 'Hmis::Hud::Project').first!
     end
 
+    # { source client id => destination warehouse id }
+    # only includes source clients that are the Owner of any CDEs in scope
+    def client_id_to_warehouse_id
+      @client_id_to_warehouse_id ||= begin
+        source_client_ids = cdes.where(data_element_definition: { owner_type: 'Hmis::Hud::Client' }).pluck(:owner_id)
```
ttoomey: The id set of all CDEs could be kind of large right? Maybe distinct so there's only one owner (client_id)
```suggestion
        source_client_ids = cdes.where(data_element_definition: { owner_type: 'Hmis::Hud::Client' }).distinct.pluck(:owner_id)
```

ttoomey (approved):
looks good!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/cde_export.rb
gigxz: Thanks, this could definitely be improved! I updated this to just join the client scope to `custom_data_elements` which should have the same effect of limiting scope to only clients that are "owners" of custom data elements. https://github.com/greenriver/hmis-warehouse/pull/5068/commits/ef2cc0dc131777e2d40fb8957c28d8b5142a7fee

gigxz:
test failure is unrelated todordie, merging

PR #5069, opened 2025-01-23
kick the can on removing hello-sign

ttoomey:
we'll fix this eventually but not now


eanders (approved):
Thanks @ttoomey, just waiting on tests to pass.

PR #5080, opened 2025-01-24
SPM 2024 Version in Longitudinal SPM Report

dtgreiner:
## _Merging this PR_

## Description

Update the Longitudinal SPM report to be aware of the new SPM 2024 version.

## Type of change
Bug fix

### DIFF
--- a/drivers/longitudinal_spm/spec/models/longitudinal_spm/report_spec.rb
+++ b/drivers/longitudinal_spm/spec/models/longitudinal_spm/report_spec.rb
@@ -1,0 +1,27 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+require 'rails_helper'
+
+RSpec.describe LongitudinalSpm::Report, type: :model do
+  before(:all) do
+    cleanup_test_environment
+  end
+  after(:all) do
+    cleanup_test_environment
+  end
+
+  let!(:user) { create :user }
+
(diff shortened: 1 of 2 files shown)


eanders (approved):
Thanks @dtgreiner !

eanders:
How hard do you think it would be to add a smoke test so if the default SPM changes to FY2025, say, tests would begin failing until we updated the code you just touched?

dtgreiner:
> How hard do you think it would be to add a smoke test so if the default SPM changes to FY2025, say, tests would begin failing until we updated the code you just touched?

I should be able to get something together here soon and I'll add it to this PR

dtgreiner:
@eanders - I've added a test that runs the default report, and testing it without the previous change in the PR was resulting in a failing test as we would want. It was getting a bit hairy setting up the data to look for the correct version of the SPM, so I opted for this higher level test instead.

drivers/longitudinal_spm/spec/models/longitudinal_spm/report_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  let!(:user) { create :user }
+
+  describe 'default report' do
+    it 'running the report with default values does not fail' do
+      report = LongitudinalSpm::Report.new(
+        user_id: user.id,
+      )
+      expect { report.run_and_save! }.not_to raise_error
```
eanders, line 24: Did you test that this does raise an error if the changes in the report hadn't been made? 
I think the more high-level approach is fine, but since it doesn't have any data to work with, I'm unclear if this is sufficient.

PR #5087, opened 2025-01-24
Cleanup and bug fixes for cohort and project group Access Controls

eanders:
## _Merging this PR_

## Description
* Updates entity access concern to correctly re-use collections and user groups if the names of project groups or cohorts change
* Updates the edit UI for cohorts and project groups to hide irrelevant access fields
* Updates the list views for access controls, collections, and user groups, to ignore items that aren't editable and are maintained by the system elsewhere (on cohorts or project groups)
* Adds tests to confirm changing entity name doesn't break the underlying connections
* Adds documentation to various methods 
* Migrates existing system collections and user groups to have a direct connection to their source

## Type of change
* Bug fix
* Documentation

### DIFF
--- a/app/controllers/admin/access_controls_controller.rb
+++ b/app/controllers/admin/access_controls_controller.rb
@@ -13,7 +13,7 @@ class Admin::AccessControlsController < ApplicationController
 
   def index
     @access_controls = access_control_scope.
-      joins(:role, :collection).
+      left_outer_joins(:collection, :role, :user_group).
       order(r_t[:name].asc, collection_t[:name].asc).
       filtered(params[:filter])
     @pagy, @access_controls = pagy(@access_controls)
@@ -53,7 +53,7 @@ class Admin::AccessControlsController < ApplicationController
   end
 
   private def access_control_scope
-    AccessControl.all
+    AccessControl.editable
   end
 
   private def filter_params
--- a/app/controllers/admin/access_controls_controller.rb
+++ b/app/controllers/admin/access_controls_controller.rb
@@ -13,7 +13,6 @@ class Admin::AccessControlsController < ApplicationController
 
   def index
     @access_controls = access_control_scope.
-      left_outer_joins(:collection, :role, :user_group).
       order(r_t[:name].asc, collection_t[:name].asc).
       filtered(params[:filter])
     @pagy, @access_controls = pagy(@access_controls)
@@ -53,7 +52,7 @@ class Admin::AccessControlsController < ApplicationController
   end
 
   private def access_control_scope
-    AccessControl.editable
+    AccessControl.user_managed
   end
 
   private def filter_params
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -38,7 +38,8 @@ class AccessControl < ApplicationRecord
   end
 
   # Where not ALL role, collection, user group are system
-  scope :editable, -> do
+  # These access controls will show up on the admin page for editing
+  scope :user_managed, -> do
     left_outer_joins(:collection, :role, :user_group).
       where.not(
         collection_id: Collection.system.select(:id),
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -39,9 +39,6 @@ class AccessControl < ApplicationRecord
 
   # Where not ALL role, collection, user group are system
   scope :editable, -> do
-    # r_t
-    # collection_t
-    # ug_t
     left_outer_joins(:collection, :role, :user_group).
       where.not(
         collection_id: Collection.system.select(:id),
--- a/app/models/access_control.rb
+++ b/app/models/access_control.rb
@@ -14,6 +14,7 @@
 class AccessControl < ApplicationRecord
   include ActionView::Helpers::TagHelper
   include UserPermissionCache
+  include ArelHelper
 
   acts_as_paranoid
   has_paper_trail
@@ -34,6 +35,19 @@ class AccessControl < ApplicationRecord
   # hide previous declaration of :system (from Kernel), we'll use this one
   replace_scope :system, -> do
     joins(:collection).merge(Collection.system)
+  end
+
+  # Where not ALL role, collection, user group are system
+  scope :editable, -> do
+    # r_t
+    # collection_t
+    # ug_t
+    left_outer_joins(:collection, :role, :user_group).
+      where.not(
+        collection_id: Collection.system.select(:id),
+        role_id: Role.system.select(:id),
+        user_group_id: UserGroup.system.select(:id),
+      )
   end
 
   scope :not_system, -> do
--- a/app/models/collection.rb
+++ b/app/models/collection.rb
@@ -41,6 +41,11 @@ class Collection < ApplicationRecord
   has_many :supplemental_data_sets, through: :group_viewable_entities, source: :entity, source_type: 'HmisSupplemental::DataSet'
 
   belongs_to :user, optional: true
+
+  # Collections may be "owned" by a single "source" entity.
+  # These eponymous collections are created through side-effects (see the EntityAccess concern) and
+  # are hidden from the administrative view.  They back the access given to users on the entity edit pages.
+  # Currently, these are used for ProjectGroups and Cohorts.
   belongs_to :source, optional: true, polymorphic: true
 
   validates_presence_of :name, unless: :user_id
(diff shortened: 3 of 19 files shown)


app/views/cohorts/edit.haml
```diff
@@ -16,25 +16,37 @@
...
-      - if users.count > 0
-        %p The following users already have access to view this cohort:
-        %ul
-          - users.each do |u|
-            %li= u.name_with_email
-      = f.input :viewer_ids, collection: @users, as: :select_two, selected: @viewer_ids, input_html: {multiple: :multiple, style: 'width: 100%;'}, label: 'Grant read-only-level access to users', hint: 'Specified user will have access to view, but not change clients on this cohort.'
+        -# TODO: START_ACL remove after ACL transition
+        -# if everyone is using Access Controls, hide this section, but keep it in the form
+        -# so we don't lose data during the change over
+        - visibility_class = if User.all_using_acls? then 'd-none' else '' end
```
ttoomey: This syntax is legal but it seems pretty unusual to me; I don't know that I've seen this in other code bases. I have a feeling that it's best to stay within the conventions of the language particularly for such a common type of expression

Would a ternary be okay since it's a short line?
```suggestion
        - visibility_class = User.all_using_acls? ? 'd-none' : nil
```
or

```suggestion
        - visibility_class = 'd-none' if User.all_using_acls?
```

app/views/cohorts/edit.haml
```diff
@@ -16,25 +16,37 @@
...
+        - hint_text = if User.anyone_using_acls? then 'Specified user and members of specified groups will have access to this cohort. This method of access is being deprecated, to be replaced with the fields below.' else 'Specified user and members of specified groups will have access to this cohort.' end
```
ttoomey: similar concern, can we avoid this 'if..else..end' construction?
```suggestion
        - hint_text = 'Specified user and members of specified groups will have access to this cohort.'
        - hint_text += ' This method of access is being deprecated, to be replaced with the fields below.' if User.anyone_using_acls?
```

app/views/cohorts/edit.haml
```diff
@@ -16,25 +16,37 @@
...
+        %div{ class: visibility_class }
+          = f.input :user_ids, collection: @groups, as: :grouped_select_two, group_method: :last, selected: @group_ids, input_html: {multiple: :multiple, style: 'width: 100%;'}, label: 'Grant access to users and groups', hint: hint_text
+        - if User.anyone_using_acls? && ! User.all_using_acls?
+          %p The following are the new mechanism for adding access to cohorts.  During the transition period, please assign both methods of access.
+        -# END_ACL
+
+        -# if no one is using Access Controls, hide this section, but keep it in the form
+        -# so we don't lose data during the change over
+        - visibility_class = if User.anyone_using_acls? then '' else 'd-none' end
+        %div{ class: visibility_class }
```
ttoomey: can we avoid reusing the `visibility_class` var from 23? Maybe we don't need a variable?

```suggestion
        %div{ class: (User.anyone_using_acls? && 'd-none') }
```

app/controllers/admin/access_controls_controller.rb
```diff
@@ -13,7 +13,7 @@ class Admin::AccessControlsController < ApplicationController
 
   def index
     @access_controls = access_control_scope.
-      joins(:role, :collection).
+      left_outer_joins(:collection, :role, :user_group).
```
ttoomey: do we need these joins? It looks like the `editable` scope is doing that already

app/models/access_control.rb
```diff
@@ -36,6 +37,16 @@ class AccessControl < ApplicationRecord
     joins(:collection).merge(Collection.system)
   end
 
+  # Where not ALL role, collection, user group are system
+  scope :editable, -> do
```
ttoomey: I wonder if there's a more descriptive name than `editable`. Maybe `not_system_managed` or `user_managed`

app/models/collection.rb
```diff
@@ -41,6 +41,7 @@ class Collection < ApplicationRecord
   has_many :supplemental_data_sets, through: :group_viewable_entities, source: :entity, source_type: 'HmisSupplemental::DataSet'
 
   belongs_to :user, optional: true
+  belongs_to :source, optional: true, polymorphic: true
```
ttoomey, line 49: Out of context I think I would be confused as to what `source` means here as the term is very generic; maybe add some docs (also on user_group)

```suggestion
    # Collections may be dedicated to single "source" entity. These eponymous collections are created through side-effects (see the EntityAccess concern). However they may be managed by admins like other collections
    belongs_to :source, optional: true, polymorphic: true
```

app/models/concerns/entity_access.rb
```diff
@@ -49,26 +82,76 @@ def editable_access_control
...
+  # also updates the collection's name to match the entity's +name+.
+  # Note that updating the name may cause an extra query, and in the future we may wan to
+  # refactor that into an after save hook or similar.
+  #
+  # @return [Collection]
+  #
   def system_collection
     @system_collection ||= begin
-      collection = Collection.where(system: ['Entities'], name: name, collection_type: collection_type).first_or_create
+      collection = Collection.system.where(source: self).first_or_create
```
ttoomey: It's not optimal we have this pattern of accessors with side-effects but so it goes. But I think it would be an improvement to raise if the side-effect fails
```suggestion
      collection = Collection.system.where(source: self).first_or_create!
```

app/models/concerns/entity_access.rb
```diff
@@ -49,26 +82,76 @@ def editable_access_control
...
       collection.set_viewables(entity_relation_type => [id])
+      collection.update(name: name) # ensure the collection name still matches
```
ttoomey: (same throughout this file)
```suggestion
      collection.update!(name: name) # ensure the collection name still matches
```

app/models/concerns/entity_access.rb
```diff
@@ -15,6 +21,21 @@ module EntityAccess
   #   3. first or create an access control that includes above role and collection
   #   4. add user to the access control
   # NOTE: this will remove any existing who are not included
+  ##
```
ttoomey: Thanks for docs!! The existing comment above this maybe could get adjusted or removed? I can't make much sense of what it means, for example:
>  Need two cohort users choosers, editors and read-only users

app/models/concerns/entity_access.rb
```diff
@@ -89,6 +172,63 @@ def editable_role
...
+  private def viewable_user_group_name
+    "#{name} [viewable]"
+  end
+
+  private def editable_user_group_name
+    "#{name} [editable]"
+  end
+
+  # NOTE: this only needs to be run once, and only if the installation was created
+  # prior to release-151
```
ttoomey: maybe we should add a todo-or-die to remove dead code once everyone has migrated?
```suggestion
  # prior to release-151
  TodoOrDie('Remove fix_entity_associations', by: '2025-04-01')
```

Or alternatively move the method into the migration where we will purge it eventually

db/migrate/20250124170335_add_source_to_collections_and_user_groups.rb
```diff
@@ -0,0 +1,9 @@
+class AddSourceToCollectionsAndUserGroups < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+
+  def change
+    add_reference :collections, :source, polymorphic: true, index: { algorithm: :concurrently }
+    add_reference :user_groups, :source, polymorphic: true, index: { algorithm: :concurrently }
```
ttoomey: collections and user groups are pretty small tables, we probably could just build the index non-concurrently and stay safe with our ddl tx?

 I also wonder if we should make these indexes unique. There shouldn't be two collections referencing the same source entity right?

```suggestion
   safety_assured do
      add_reference :collections, :source, polymorphic: true, index: { unique: :true }
      add_reference :user_groups, :source, polymorphic: true, index: { unique: :true }
    end
```

app/models/concerns/entity_access.rb
```diff
@@ -89,6 +172,63 @@ def editable_role
...
+  def fix_entity_associations
+    # If we already have a system collection in the new mechanism, do nothing
+    existing_collection = Collection.system.where(source: self).exists?
+    unless existing_collection
+      # Find the last collection that would have been created under the old mechanism
+      collection = Collection.system.where(name: name).
+        order(id: :desc).
+        first_or_create
```
ttoomey: Are these normally created on-demand? Do we expect every project group or cohort to have a collection/usergroup? If not maybe we don't need to make new records unless the collection/user group already exists.

Also probably should use bang variant:
```suggestion
        first_or_create!
```

ttoomey (approved):
Looks good! I left a few comments and suggestions

dtgreiner (approved):

PR #5089, opened 2025-01-27
Implement Split Households mutation

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/5800

Summary of changes:
- Add the Split Household mutation
- Add spec and system tests

## Type of change
New feature


drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Mutations
+  class SplitHousehold < BaseMutation
```
gigxz: Prefer using `CleanBaseMutation` to skip the input wrapper thing - I missed that on join households. Going forward we should use CleanBaseMutation but it would be too much churn to retroactively updated all the BaseMutation mutations.

drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
...
+      access_denied! unless splitting_enrollments.size == splitting_enrollment_inputs.size
+
+      project = splitting_enrollments.map(&:project).uniq.sole
+      access_denied! unless current_permission?(permission: :can_split_households, entity: project)
+
+      donor_household = splitting_enrollments.map(&:household).uniq.sole
+      remaining_enrollments = donor_household.enrollments.where.not(id: splitting_enrollment_ids)
+
+      raise 'Splitting all clients to a new household is invalid' if remaining_enrollments.empty?
+      raise 'This operation would leave behind a household with no HoH, which is not allowed' unless remaining_enrollments.exists?(relationship_to_hoh: 1)
```
gigxz: Micro optimization: check against the array rather than making another db query with `exists?`:
```suggestion
      raise 'This operation would leave behind a household with no HoH, which is not allowed' unless remaining_enrollments.any?(&:head_of_household?)
```

drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
...
+
+      donor_before_state = donor_household.snapshot_household_state
+      new_household_id = Hmis::Hud::Base.generate_uuid
+
+      Hmis::Hud::Enrollment.transaction do
+        splitting_enrollments.each do |enrollment|
+          enrollment.household_id = new_household_id
+          enrollment.relationship_to_hoh = map_enrollment_id_to_relationship[enrollment.id.to_s]
```
gigxz, line 42: Should we validate the relationships at all here, for example ensuring there is exactly 1 hoh? We can `raise` if we don't expect it from the frontend.

drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
...
+
+          enrollment.active_unit_occupancy&.assign_attributes(occupancy_period_attributes: { end_date: Date.current })
```
gigxz, line 46: Maybe add comment: release the unit when splitting household members, so that we don't end up in a situation where there is a unit occupied by 2 different households (which could occur if any of the remaining enrollments occupy the same unit)

drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
...
+        event = Hmis::HouseholdEvent.new
+        event.user = current_user
+        event.household = donor_household
+        event.event_type = Hmis::HouseholdEvent::SPLIT
+        event.event_details = {
+          'receiving_household_id': new_household_id,
+          'before': donor_before_state,
+          'after': donor_after_state,
+        }
+        event.save!
```
gigxz: It might be nice to have this logic contained in the `HouseholdEvent` class, so we can have a consistent place whether the event detail shapes are defined. Right now this json shape is duplicated in both mutations, so the shapes could get out of sync.

```ruby

def self.new_split_event(user:, household:, receiving_household_id:, before_state:, after_state:)
   new(...)
end

def self.new_join_event(user:, household:, donor_household_id:, before_state:, after_state:)
   new(user: user, household: household, ...)
end
```

drivers/hmis/app/graphql/mutations/split_household.rb
```diff
@@ -0,0 +1,69 @@
...
+
+        remaining_enrollments.invalidate_processing!
```
gigxz, line 65:
```suggestion
        # Invalidate remaining enrollments and trigger re-processing, since household composition has changed.
        # Processing for the new household should already be triggered based on the warehouse_trigger_processing after_create hook.
        remaining_enrollments.invalidate_processing!
        queue_service_history_processing!
```

gigxz (approved):
Looks great! Thanks for the system test! 💯

PR #5099, opened 2025-01-30
Redact names on all client dashboard pages (and some reports)

eanders:
## _Merging this PR_

## Description
* This applies the PII Provider to additional sections of the client dashboard
* Additionally, the PII Provider is applied to reporting pages where the change was easy to identify and implement

## Type of change
* New Feature

- [x] Provided testing instructions

### DIFF
--- a/app/controllers/clients/anomalies_controller.rb
+++ b/app/controllers/clients/anomalies_controller.rb
@@ -66,7 +66,7 @@ module Clients
     end
 
     protected def title_for_show
-      "#{@client.name} - Anomalies"
+      "#{@client.pii_provider(user: current_user).full_name} - Anomalies"
     end
 
     private def anomaly_params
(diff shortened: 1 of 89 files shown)


dtgreiner (approved):
Is there a reason why we don't modify/override the client fields (e.g. name, dob, ssn) to use the pii_provider? The concern I'd have is that it may be easy to miss this when adding a new view/method that calls `client.name` instead of `client.pii_provider(user: current_user).name`.

Otherwise this looks good to me

eanders:
> Is there a reason why we don't modify/override the client fields (e.g. name, dob, ssn) to use the pii_provider? The concern I'd have is that it may be easy to miss this when adding a new view/method that calls `client.name` instead of `client.pii_provider(user: current_user).name`.
> 
> Otherwise this looks good to me

Only because there's no `current_user` on client, so we wouldn't know what to show.  We could probably provide helper methods like `client.protect_name(user: current_user)` but I'm not sure that's significantly better (and they'd just be wrappers for the PII provider calls.

PR #5100, opened 2025-01-31
Remove duplicates from pathways file

gigxz:
## _Merging this PR_

## Description
Issue (customer req): https://github.com/open-path/Green-River/issues/7094

Adjusting this export so that we don't export "duplicate" values for the same Warehouse Client.

For situations where a destination client has multiple HMIS clients, _AND_ multiple of those HMIS clients has Pathways custom fields, we used to export all of the pathway data for each source client –– resulting in duplicate destination IDs in the export. With this change, we now choose to only export Pathways from **one** of those HMIS clients. We select the HMIS client who's pathways have been most recently updated.

## Type of change

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/pathways_export.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/pathways_export.rb
@@ -92,12 +92,13 @@ module HmisExternalApis::AcHmis::Exporters
     end
 
     private def pathway_cded_key_to_id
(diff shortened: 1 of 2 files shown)


ttoomey (approved):
Looks good!

The only suggestion I have is that it might be a good idea to include an additional condition on owner_type in pathways_by_client_id. Otherwise, if the CDE was somehow associated with an enrollment or something, you could end up with bugs that would be difficult to diagnose

```diff
    private def pathways_by_client_id
      @pathways_by_client_id ||= Hmis::Hud::CustomDataElement.
        where(data_element_definition_id: pathway_cded_key_to_id.values). # All Pathway-related definitions
+        where(owner_type: 'Hmis::Hud::Client').
        group_by(&:owner_id) # By Client ID
    end
```

gigxz:
Thanks for the review and suggestion @ttoomey, added!

PR #5101, opened 2025-01-31
max threshold for source clients

dtgreiner:
## _Merging this PR_

## Description

Add a max threshhold for number of source clients added to a destination client when automatically merging clients. 

## Type of change
New feature

### DIFF
--- a/app/models/grda_warehouse/tasks/identify_duplicates.rb
+++ b/app/models/grda_warehouse/tasks/identify_duplicates.rb
@@ -9,6 +9,8 @@ module GrdaWarehouse::Tasks
     include ArelHelper
     include NotifierConfig
 
+    MAX_SOURCE_CLIENTS = 50
+
     def initialize(run_post_processing: true)
       setup_notifier('IdentifyDuplicates')
       @run_post_processing = run_post_processing
@@ -144,8 +146,10 @@ module GrdaWarehouse::Tasks
         # This has already been fully merged
         next if source_id == destination_id
 
-        destination = client_destinations.find(destination_id)
-        source = client_destinations.find(source_id)
+        destination = client_destinations.find_by(id: destination_id)
+        source = client_destinations.find_by(id: source_id)
+        next unless destination.present? && source.present?
+
         begin
           destination.merge_from(source, reviewed_by: user, reviewed_at: DateTime.current)
         rescue Exception => e
@@ -241,9 +245,13 @@ module GrdaWarehouse::Tasks
       )
     end
 
+    private def matchable_destination_ids
+      GrdaWarehouse::WarehouseClient.group(:destination_id).having('count(*) < ?', MAX_SOURCE_CLIENTS).select(:destination_id)
+    end
+
     # fetch a list of existing clients from the DND Warehouse DataSource (current destinations)
     private def client_destinations
-      GrdaWarehouse::Hud::Client.destination
+      GrdaWarehouse::Hud::Client.destination.where(id: matchable_destination_ids)
     end
 
     # Look for really clear matches (2 of the following 3 should be good):
--- a/app/models/grda_warehouse/tasks/identify_duplicates.rb
+++ b/app/models/grda_warehouse/tasks/identify_duplicates.rb
@@ -245,6 +245,8 @@ module GrdaWarehouse::Tasks
       )
     end
 
+    # Destination clients are limited to MAX_SOURCE_CLIENTS number of source clients. This is to prevent runaway duplicate merges.
+    # This will return the ids of any destination clients that are eligible to have additional source clients added.
     private def matchable_destination_ids
       GrdaWarehouse::WarehouseClient.group(:destination_id).having('count(*) < ?', MAX_SOURCE_CLIENTS).select(:destination_id)
     end
--- a/app/models/grda_warehouse/tasks/identify_duplicates.rb
+++ b/app/models/grda_warehouse/tasks/identify_duplicates.rb
@@ -246,14 +246,16 @@ module GrdaWarehouse::Tasks
     end
 
     # Destination clients are limited to MAX_SOURCE_CLIENTS number of source clients. This is to prevent runaway duplicate merges.
-    # This will return the ids of any destination clients that are eligible to have additional source clients added.
-    private def matchable_destination_ids
-      GrdaWarehouse::WarehouseClient.group(:destination_id).having('count(*) < ?', MAX_SOURCE_CLIENTS).select(:destination_id)
+    # This will return the ids of any destination clients that is at or beyond this threshold so they can be filtered out of future matching.
+    # We are using unmatchable destinations here to more easily include destination clients who do not have a warehouse client record.
+    # This should also be a smaller list than the full list of matchable destination ids.
+    private def unmatchable_destination_ids
+      GrdaWarehouse::WarehouseClient.group(:destination_id).having('count(*) >= ?', MAX_SOURCE_CLIENTS).select(:destination_id)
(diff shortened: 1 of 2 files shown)


app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -241,9 +245,13 @@ def match_existing!
       )
     end
 
+    private def matchable_destination_ids
```
eanders: Can you add a comment for this method about the intention?

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -127,6 +127,58 @@
         end
       end
     end
+
+    describe 'source client threshold' do
+      before(:each) do
+        client_in_destination.destroy
+        client_in_source.update(first_name: 'A', last_name: 'Client', dob: '2000-01-01', ssn: 'XXX-XX-1234')
```
eanders: `SSN` field should be a 9 char string (so I think you want to leave out the dashes)

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -127,6 +127,58 @@
...
+        GrdaWarehouse::Tasks::IdentifyDuplicates.new.identify_duplicates
```
eanders: It looks like you call this in the tests below, maybe you don't need the `before(:each)` at all?

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -127,6 +127,58 @@
...
+
+        number_sample_clients = 125
+
+        (1..(number_sample_clients - 1)).each do |n|
+          client = create :grda_warehouse_hud_client, data_source: source_data_source
+          date = Date.new(2000, 1, 1) + n.days - n.months
+          str_n = n.to_s.rjust(4, '0')
+          ssn = "#{str_n.last(3)}-#{str_n.last(2)}-#{str_n.last(4)}"
+          client.update(first_name: "client_#{n}", last_name: 'Test', dob: date, ssn: ssn)
+        end
```
eanders: Can you add a comment above this something like:
> Generate unique clients that won't match each other

Then further down:
> Test that each unique client received a destination client

Then for the next set:
> Generate non-unique clients such that they should all match each other

and
> Test that while they all match each other, only 3 destination clients are generated. 

I also think `expected_number_destination_clients = ((number_sample_clients * 1.0) / GrdaWarehouse::Tasks::IdentifyDuplicates::MAX_SOURCE_CLIENTS).ceil` might be nice to just write as 3 so we are sure we aren't ending up with `1`.  Though I suppose just testing that that is > 1 would be sufficient too.

eanders (changes requested):
Thanks @dtgreiner, this looks generally good, but I think some notes about why things are would be helpful, and one note about ensuring we're testing what we think we are.

app/models/grda_warehouse/tasks/identify_duplicates.rb
dtgreiner: added in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5101/commits/0c3456738bfd97dbd00e7d4c18b38c36aea53673)

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
dtgreiner: added in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5101/commits/0c3456738bfd97dbd00e7d4c18b38c36aea53673)

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
dtgreiner: Comments and test updated in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5101/commits/0c3456738bfd97dbd00e7d4c18b38c36aea53673)

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
dtgreiner: cleaned up this section in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5101/commits/c8d30a2b23810dbfb07f79f32207236af44bd4e1)

dtgreiner:
@eanders this should be good for another check when you have  a chance. Thanks!

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -127,6 +127,69 @@
...
+          str_n = n.to_s.rjust(4, '0')
+          ssn = "#{str_n.last(3)}#{str_n.last(2)}#{str_n.last(4)}"
+          client = GrdaWarehouse::Hud::Client.source.find_by(SSN: ssn)
+          client.update(first_name: 'A', last_name: 'Client', dob: '2000-01-01', ssn: 'XXXXX1234')
+        end
+
+        GrdaWarehouse::Tasks::IdentifyDuplicates.new.run!
+        GrdaWarehouse::Tasks::IdentifyDuplicates.new.match_existing!
+
+        # # The code below will calculate the number of expected destination clients in the case that MAX_SOURCE_CLIENTS changes enought to affect this number
```
eanders, line 170:
```suggestion
        # # The code below will calculate the number of expected destination clients in the case that MAX_SOURCE_CLIENTS changes enough to affect this number
```

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -127,6 +127,69 @@
...
+        # expected_number_destination_clients = ((number_sample_clients * 1.0) / GrdaWarehouse::Tasks::IdentifyDuplicates::MAX_SOURCE_CLIENTS).ceil
+
+        # With MAX_SOURCE_CLIENTS set to 50, we are expecting 3 destiantion clients. 2 with 50 source clients and 1 with 25 source clients.
+        # We are setting this specifically instead of using the calculated number in case `MAX_SOURCE_CLIENTS` gets set to a number larger than
+        # `number_sample_clients`. If that happened, we wouldn't be reaching the threshold for the numebr of source clients that we are testing.
+        expected_number_destination_clients = 3
+
+        destination_clients = [].tap do |clients|
+          clients << GrdaWarehouse::Hud::Client.destination
+        end.flatten
```
eanders: I'm not understanding what's happening here.  This looks to be generating an array of three scopes all the same, and then calling flatten (which should do nothing).
You should end up with:
`destination_clients = [scope_of_destination_clients, scope_of_destination_clients, scope_of_destination_clients]`

I think you just wanted `destination_clients = GrdaWarehouse::Hud::Client.destination.to_a`

dtgreiner: You're right, and it's updated in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5101/commits/95588c08570d6703dee9427a5f815d4736b3c2ba). 

This was an artifact from a different way I was grouping the clients for testing. It slowly was optimized away and left with what we are seeing now which is essentially `destination_clients = GrdaWarehouse::Hud::Client.destination.to_a`

dtgreiner:
working on cleaning up the spec issues popping up. Should have something ready soon.

eanders (approved):

PR #5113, opened 2025-02-06
7210 client search performance

ttoomey:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7210

details:
- attempted performance fix fox acl-user search. Perform text filtering first before authorization
- fix render performance due to n+1 queries on client search page using an 'accessor' class
- cleanup client_names/aliases

QA instructions in issue

## Type of change
Performance fix


app/models/grda_warehouse/hud/client.rb
```diff
@@ -754,26 +754,6 @@ def alternate_names
       names.join(',')
     end
 
-    def client_names(user:, health: false)
```
ttoomey, line 757: moved to SourceClientNameSet. Note, afaict, the 'health' flag is always true given current usage. I removed it as a simplification

app/views/assigned/clients/_client_table.haml
```diff
@@ -1,24 +1,25 @@
 .o-page__main-content.mt-3
   .clients__list
-    - clients.each do |user_client|
-      - client = user_client.client
+    - source_clients = clients.map(&:user_client)
+    - accessor = current_user.client_view_accessor.preload_source_clients(source_clients)
+    - source_clients.each do |client|
```
ttoomey, line 5: @eanders I'm not sure how this view is used. Can you suggest how I can exercise it? Maybe it's part of cohorts but those don't work for me locally

eanders:
@ttoomey I think we should retarget 151 so we can get this on staging for some QA

app/views/assigned/clients/_client_table.haml
eanders: I believe you need to assign some relationships to see data there.
Screenshot 2025-02-06 at 5 04 41 PM

app/models/grda_warehouse/source_client_view_accessor.rb
```diff
@@ -0,0 +1,53 @@
...
+
+  # @param user [User] The current user accessing the client data
+  def initialize(user:)
+    @user = user
+    @source_clients = {}
+  end
+
+  # Retrieves source clients for a given destination client
+  # @param client [DestinationClient] The destination client to look up
+  # @return [Array<HudClient>] Array of source client records associated with the destination client
```
eanders: Maybe?  HudClient is weird, but so is DestinationClient and SourceClient since the model is GrdaWarehouse::Hud::Client.
```suggestion
  # @return [Array<SourceClient>] Array of source client records associated with the destination client
```

app/models/grda_warehouse/source_client_view_accessor.rb
```diff
@@ -0,0 +1,53 @@
...
+  # @param client [DestinationClient] The destination client to get names for
+  def client_names(client)
+    GrdaWarehouse::SourceClientNameSet.new(
+      destination_client: client,
+      source_clients: source_clients(client),
+      user: @user,
+    )
+  end
+
+  def preload_source_clients(clients)
```
eanders: Should this method name indicate it is only returning those that are "searchable_to" the user?  Sometimes people can search for a client that they don't have access to through `clients_source_visible_to`.  

Since this returns the full client objects, it feels a bit dangerous.  Maybe it should be:

`preload_source_clients(clients, context: :searchable)` and `preload_source_clients(clients, context: :viewable)` to indicate what context the returned data should be used in.

app/models/grda_warehouse/source_client_name_set.rb
```diff
@@ -0,0 +1,46 @@
...
+      @names = source_clients.map do |client|
+        SourceClientName.new(
+          ds_name: client.data_source.short_name,
+          ds_id: client.data_source.id,
+          value: client.pii_provider(user: user).full_name,
+        )
+      end
+
+      patient = destination_client.patient
+      if patient && source_clients.none? { |sc| sc.data_source&.authoritative_type == 'health' }
```
eanders, line 35:
```suggestion
     # if the destination client has a patient record, and we haven't already added a source client 
     # from the health data source, include the name of the patient
      if patient && source_clients.none? { |sc| sc.data_source&.authoritative_type == 'health' }
```

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
```diff
@@ -35,7 +35,14 @@ def index
...
     elsif current_user.can_search_own_clients? && params[:q].present?
-      @clients = client_source.text_search(params[:q], client_scope: client_search_scope, sorted: sorted)
+      # FIXME: optimization and refactor needed, this is hack
+      # It seems that searchable_to scope is not-performant under some configurations. Perform text search first
+      upper_bound_limit = 8_000 # protect from overly broad queries
+      unauthorized_client_ids = client_source.text_search(params[:q], sorted: sorted).limit(upper_bound_limit).pluck(:id)
+      Sentry.capture_message("Client search from User##{current_user.id} for #{params[:q]} was too broad") unless unauthorized_client_ids.size < upper_bound_limit
+
+      # performs the text search twice to sort the results
+      @clients = client_source.where(id: unauthorized_client_ids).text_search(params[:q], client_scope: client_search_scope, sorted: sorted)
```
eanders: Doesn't performing this search twice just double the query time?

In an effort to prevent people trying to use the client search for reporting (or accidentally pulling the entire client set) I wonder if we could decrease the max limit to 500.  I can't think of a reason you would ever want more than a couple hundred results from a client search anyway.

Maybe (untested):
```suggestion
      # FIXME: optimization and refactor needed, this is hack
      # It seems that searchable_to scope is not-performant under some configurations. Perform text search first
      upper_bound_limit = 500 # protect from overly broad queries
      unauthorized_client_ids = client_source.text_search(params[:q], sorted: sorted).limit(upper_bound_limit).pluck(:id)
      if unauthorized_client_ids.size == upper_bound_limit
        flash[:alert] = "Client search returned more than #{upper_bound_limit} records, showing #{upper_bound_limit} records, please narrow your search and try again."
        Sentry.capture_message("Client search from User##{current_user.id} for #{params[:q]} was too broad") 
      end
      
      # performs the text search twice to sort the results
      @clients = client_source.where(id: unauthorized_client_ids).text_search(params[:q], client_scope: client_search_scope, sorted: sorted)
```

eanders (changes requested):
@ttoomey thanks for the new approach!

I dropped a handful of comments inline, but mainly I like the approach.  We should:
1. Make it more clear that the clients being returned are just for the search context, either by renaming the class or the preload method (or something else.)
2. Cap the number of returned results quite a bit smaller and alert the user they are seeing truncated results.  Without a warning I think the sort could potentially not find the most relevant maybe?

app/models/grda_warehouse/source_client_view_accessor.rb
ttoomey: that's more consistent, thanks!

app/models/grda_warehouse/source_client_view_accessor.rb
ttoomey: The naming could be more clear for sure. Also just to clarify this preload method is not actually intended to have a significant return type, it just warms the cache. I'll add an explicit return to make that more clear.

I think adding a context argument could be a nice api hint but it's some extra fuss to support such conditional behavior; we would have to maintain different caches for each context. I suppose we could raise if context != 'searchable' but I think that would make it confusing

How about we just rename the accessor methods?
- client_names -> searchable_client_names
- source_clients -> searchable_clients

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
ttoomey: > Doesn't performing this search twice just double the query time?

No, it's not doubling the query time as far as I can tell. In my testing the text search is not the performance problem. The performance problem appears to be the searchable_to auth scope, which is why we are splitting it out into phases like this

This is a hack but I think it would be hard to avoid running the text search twice as you need to filter to matching records, filter for authorization, and then rank the results (second pass text search).

> In an effort to prevent people trying to use the client search for reporting (or accidentally pulling the entire client set) I wonder if we could decrease the max limit to 500. I can't think of a reason you would ever want more than a couple hundred results from a client search anyway.

So the reason to have a greater upper bound is that we don't know which records the user has access to at this point. Let's say I have a limit of 10 and there are 20 matching clients; however the clients I can see are all in the last 10 so I will get an empty result set. A generous upper bound makes this less likely to occur.

But I guess selecting 8k ids sounds too dangerous? In that case we can reduce the limit but show an error to the user and refuse to return incomplete results if we exceed the limit. That seems safer although it means you will no longer be able to search for common first-names on most real data sets (for acl users at least)

app/models/grda_warehouse/source_client_view_accessor.rb
ttoomey: oh I see where you got the idea the return type was meaningful, I had some old incorrect code in the client_table partial. I'll fix that up

ttoomey:
> I think we should retarget 151 so we can get this on staging for some QA

Sure, I was thinking we might want to try selective deploy to the impacted production instance in 150. But we can alway cherry-pick this out and and apply it to 150 as a one-off

drivers/client_access_control/app/views/client_access_control/clients/index.haml
```diff
@@ -29,7 +29,12 @@
     = render 'common/pagination_top', item_name: 'client'
   .o-page__main-content.mt-3
     .clients__list
-      = render 'clients/new_client'
+      - if @client_search_returned_too_much
+        .alert.alert-danger(role="alert")
+          Client search returned too many records, please narrow your search and try again.
```
ttoomey, line 34: choosing to not using flash here, rather position the alert inline below the search form to be more clear

app/models/grda_warehouse/source_client_view_accessor.rb
eanders: The bit on line 42-43 below:

```
clients = GrdaWarehouse::Hud::Client.arbiter(@user).
clients_source_searchable_to(@user, client_ids: source_client_ids)
```

 is what makes me think we should rename.  

I like your options above. `searchable_client_names` and `searchable_clients` but I think `preload_source_clients` should be `preload_searchable_source_clients` because of the use of `arbiter.clients_source_searchable_to`.

eanders: Alternatively, if `preload_source_clients` were private, that would suffice.

OR 

If the class name were `SourceClientSearchAccessor`

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
```diff
@@ -76,6 +76,23 @@ def index
     end
   end
 
+  # FIXME: optimization and refactor needed, this is hack
+  # It seems that searchable_to scope is not-performant under some configurations.
+  protected def temporary_acl_search_hack
+    upper_bound_limit = 500 # protect from overly broad queries
+    # 1. Perform text search first
+    unauthorized_client_ids = client_source.text_search(params[:q], sorted: sorted).limit(upper_bound_limit).pluck(:id)
```
eanders, line 84: I see, I'd missed that the first query wasn't bounded by the user's permission.  That makes more sense, and I agree, this should probably cast a wider net, though if your search matches 500, you probably need to refine it.

Maybe we should leave it big (5,000?).  

Maybe the first query should be 5,000 and the second could be limited to 500.  Maybe we just go with 500 as you have it written and see if there are issues.

app/views/assigned/clients/_client_table.haml
```diff
@@ -1,24 +1,25 @@
 .o-page__main-content.mt-3
   .clients__list
-    - clients.each do |user_client|
-      - client = user_client.client
+    - source_clients = clients.map(&:client)
+    - current_user.client_view_accessor.preload_source_clients(source_clients)
```
eanders: maybe `client_view_accessor` should take an array of destination clients as an optional initialization argument?  Then this could be something like:

```
- view_clients = current_user.client_view_accessor(clients: clients)
...
- akas = view_clients.searchable_client_names(client).to_a
```

Essentially allowing the preload to be private.

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
ttoomey: Yeah, let's up the limit to 5k.

I don't think there's a reason to limit the second query here - it should already be limited by pagination in the view afaik. Or do you think we should need an additional limit? That would be a change in behavior and might need to be reflected in the UI (presenting pagination as fuzzy number "thousands" rather than a specific and inaccurate value)

app/views/assigned/clients/_client_table.haml
ttoomey: While it would be more concise, this API feels less clear to me. Exposing 'preload' feels clear and flexible. Packing the clients array into the accessor for the adapter feels ambiguous to me and raises questions around how I'm supposed to interact with the cache. Also with this api you'd need to pass in the whole client array to use the name accessor, which feels like extra work.

I think I can do a better job communicating the function of this class, I'll try and enrich the inline docs. Or happy to huddle if you have some suggestions

ttoomey: per conversation, renamed preload_source_clients to preload_searchable_clients

eanders (approved):
Thanks @ttoomey !

PR #5116, opened 2025-02-06
Additional RRH Sub Type Adjustments

eanders:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

## Description
[//]: # (Summarize changes and include links related issue)
[//]: # (List any new dependencies or relevant ADRs)
* Updates some key Project and SHE scopes to handle RRH with RRH Sub Types
* Adds RRH sub-type and project type to the projects on the client history calendar 

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)
* New feature

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] I have added spec tests (or not applicable)
- [x] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #5118, opened 2025-02-07
Disallow same-day exit from residential projects

gigxz:
## _Merging this PR_

## Description
Issue and QA instructions: https://github.com/open-path/Green-River/issues/7238

* Prevent Auto-exit from creating same-day exits in residential projects
* Update AutoExitJob to allow passing scope (projects or data source)
* Prevent manual same-day exits in HMIS for residential projects

## Type of change
Bug fix, HUD compliance

### DIFF
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -49,7 +49,7 @@ module Hmis
           Hmis::Hud::Base.transaction do
             # Auto-exit all household members together, setting the exit date equal to the most recent contact for any household member
             household.enrollments.each do |e|
-              auto_exit(e, most_recent_contact, project_type: project.project_type)
+              auto_exit(e, most_recent_contact, project: project)
             end
           end
         end
@@ -77,12 +77,12 @@ module Hmis
       end
     end
 
-    def auto_exit(enrollment, most_recent_contact, project_type:)
+    def auto_exit(enrollment, most_recent_contact, project:)
       exit_date = contact_date_for_entity(most_recent_contact)
       # If most recent contact was a Bed Night service, the Exit Date should be the day after they received service
       exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
       # If most recent contact was on Entry Date and this is a residential project, add 1 day to avoid Same-day-exit data quality errors
-      exit_date += 1.day if exit_date == enrollment.entry_date && HudUtility2024.residential_project_type_ids.include?(project_type)
+      exit_date += 1.day if exit_date == enrollment.entry_date && !project.allows_same_day_exit?
 
       user = Hmis::Hud::User.system_user(data_source_id: enrollment.data_source_id)
 
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -12,6 +12,7 @@ module Hmis
       Hmis::ProjectAutoExitConfig.exists?
     end
 
+    # todo add project id param
     def perform
       return unless self.class.enabled?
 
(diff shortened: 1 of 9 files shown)


drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -68,10 +77,13 @@ def get_most_recent_contact(enrollment, project)
       end
     end
 
-    def auto_exit(enrollment, most_recent_contact)
+    def auto_exit(enrollment, most_recent_contact, project_type:)
       exit_date = contact_date_for_entity(most_recent_contact)
       # If most recent contact was a Bed Night service, the Exit Date should be the day after they received service
       exit_date += 1.day if most_recent_contact.is_a?(Hmis::Hud::Service) && most_recent_contact.record_type == 200
+      # If most recent contact was on Entry Date and this is a residential project, add 1 day to avoid Same-day-exit data quality errors
+      exit_date += 1.day if exit_date == enrollment.entry_date && HudUtility2024.residential_project_type_ids.include?(project_type)
```
eanders: This is probably fine but with RRH SSO (RRH Sub Type 1) it might make sense to check for that, and allow same day exits when `project_type == 13 && rrh_sub_type == 1`.  That said, I doubt there's any harm making it a one night enrollment.

eanders (approved):
Thanks @gigxz !

martha (approved):
looks good, thank you @gigxz !

drivers/hmis/app/models/hmis/hud/validators/base_validator.rb
```diff
@@ -44,6 +44,10 @@ def self.before_entry_message(entry_date)
     "cannot be before entry date (#{entry_date.strftime('%m/%d/%Y')})"
   end
 
+  def self.on_entry_message(entry_date)
```
martha, line 47: (though, I'm sure you saw, it looks like this introduces some other test failures that are probably easily fixed)

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
gigxz: Thanks, good point.

I set it up to use the same logic as `EntryAfterExit` validator, which looks only at `residential_project_type_ids` if I traced the code correctly. 

I am concerned that this might not be the right set to use though, since residential_project_type_ids includes Street Outreach, which isn't in line with the glossary:

> A “residential project” provides overnight accommodations and includes projects that are meant to be
long-term, such as permanent supportive housing.


Perhaps I should make an another project id list (or function) for `project_type_ids_allowing_same_day_exit` so it's clear what it's used for, and use it for both EntryAfterExit validator and this new code? I think it would be the same list just minus street outreach:

```
irb(main):003:0> HudUtility2024.residential_project_type_ids.map { |id| HudUtility2024.project_type(id) }
=>
["Emergency Shelter - Entry Exit",
 "Emergency Shelter - Night-by-Night",
 "Transitional Housing",
 "PH - Permanent Supportive Housing",
 "Street Outreach",
 "Safe Haven",
 "PH - Housing Only",
 "PH - Housing with Services (no disability required for entry)",
 "PH - Rapid Re-Housing"]
```

eanders: I believe Street Outreach should also require one CLS and an enrollment length of at least one night.  No harm in making a new method.  

What I was getting at is that RRH probably only needs the one night requirement if `RRHSubType` is **not** `1`.
Sadly, this can't be solved by ONLY looking a the project type.

gigxz: Yeah sorry so for RRHSubType, I feel like we should be consistent with the CSV validator, do you want me to update it in both places? The new method can accept both project type and subtype

Odd I wouldn't have expected that for street outreach. Do you have a reference?

gigxz: proposed change to move it to the hmis project class https://github.com/greenriver/hmis-warehouse/pull/5118/commits/9970124d9f68df65798f262b83a8e5193f29b3f1

eanders: I like that.  
We also have some [changes coming in 151](https://github.com/greenriver/hmis-warehouse/pull/5116/files) that begin to take this into account more systematically on the warehouse side.

eanders: Oh, and the CSV validator.... should probably be updated as well.  I'm happy to do that. I'll make an issue.

gigxz: ok thanks!

PR #5120, opened 2025-02-10
Improve performance of File viewable_by scope

martha:
## _Merging this PR_

## Description
GH issue https://github.com/open-path/Green-River/issues/7157

## Type of change
Performance improvement


drivers/hmis/app/models/hmis/file.rb
```diff
@@ -43,14 +43,23 @@ def data_source
...
+
+    # significantly speeds up the Client scope if we restrict which client IDs we care about.
+    # if the caller passed specific client IDs to check, just check those; otherwise check any clients that have files
+    client_ids_to_check = client_ids || Hmis::File.where(enrollment_id: nil).pluck(:client_id).uniq
+
     client_scope = Hmis::Hud::Client.
-      viewable_by(user).
-      with_access(user, :can_view_any_nonconfidential_client_files, :can_view_any_confidential_client_files)
+      where(id: client_ids_to_check).
+      # recombining can_view_clients into the with_access call, as opposed to separately adding the .viewable_by scope, does seem to save some time
```
martha: Proposal 2: Remove `viewable_by(user)` and add `can_view_clients` to the `with_access` check. 

Reasoning: I _think_ this should give the same results, because `viewable_by` is just making a `with_access` call under the hood:
https://github.com/greenriver/hmis-warehouse/blob/release-152/drivers/hmis/app/models/hmis/hud/client.rb#L107-L113

In QA with 500k clients:
```
Benchmark.measure { Hmis::Hud::Client.viewable_by(user).with_access(user, :can_view_any_nonconfidential_client_files, :can_view_any_confidential_client_files).count }
=> #<Benchmark::Tms:0x00007efc5ef4c700 @cstime=0.0, @cutime=0.0, @label="", @real=9.596826205030084, @stime=0.0, @total=0.03800899999999885, @utime=0.03800899999999885>

Benchmark.measure { Hmis::Hud::Client.with_access(user, :can_view_clients, :can_view_any_nonconfidential_client_files, :can_view_any_confidential_client_files).count }
=> #<Benchmark::Tms:0x00007fb5efb807c0 @cstime=0.0, @cutime=0.0, @label="", @real=2.370837351307273, @stime=0.0, @total=0.018767000000000422, @utime=0.018767000000000422>
```

Questions:
- I do think this is a hit to readability, but to me it seems worth it, right?

drivers/hmis/app/models/hmis/file.rb
```diff
@@ -43,14 +43,23 @@ def data_source
...
+  # todo @martha - need to add tests for this after validating the approach
+  scope :viewable_by, ->(user, client_ids = nil) do
     # NOTE: it's okay that confidential files are included in this scope even if the user
     # doesn't have permission to read the file. Users can see the existence of confidential
     # files but they can't read them. Reference:
     # https://github.com/open-path/Green-River/issues/5184
+
+    # significantly speeds up the Client scope if we restrict which client IDs we care about.
+    # if the caller passed specific client IDs to check, just check those; otherwise check any clients that have files
+    client_ids_to_check = client_ids || Hmis::File.where(enrollment_id: nil).pluck(:client_id).uniq
```
martha: Proposal 1: Allow the caller of the `viewable_by` scope to pass in specific client IDs to check. Even if none are provided, restrict the scope to only clients that have any files.

Reasoning: From the perspective of the Client Files page, where we are seeing slowness, we already know that we care about only 1 client, so it doesn't make sense to make a time-consuming sub-query to find _all_ clients this user has access to.

Questions:
- Does this change seem like an antipattern? It feels like the usual railsy way to solve similar problems would be to combine scopes, but I couldn't quite figure out how that would work in this case, where the `viewable_by` scope needs to query `Hmis::Hud::Client`s in order to find out which files are viewable by this user.
- Maybe it's more elegant to pass a `client_scope` rather than `client_ids`
- Should we do something similar with enrollments, in the `HasFiles` concern, for parity? From what I can tell we're never querying the enrollments `files` field

drivers/hmis/app/models/hmis/file.rb
martha: 🤦‍♀️ after adding spec tests, I learned that this won't work, since the `permissions` passed to `with_access` are ORed not ANDed -- i.e. `with_access` returns records for which the user has any of the permissions, not all. I'll revert this.

drivers/hmis/app/models/hmis/file.rb
martha: Discussed on dev standup and agreed that we will go forward with this approach

drivers/hmis/app/graphql/types/hmis_schema/has_files.rb
```diff
@@ -33,7 +34,9 @@ def resolve_files(scope = object.files, **args)
       private
 
       def scoped_files(scope, sort_order: :date_created)
-        scope = scope.viewable_by(current_user)
+        # Passing the client ID to the viewable_by scope improves performance
+        client_id = object.instance_of?(Hmis::Hud::Client) ? object.id : object.client.id
+        scope = scope.viewable_by(current_user, client_ids: [client_id])
         scope = scope.sort_by_option(sort_order) if sort_order.present?
         scope
       end
```
gigxz: This places as implicit constraint that `HasFiles` can only be used on an object that has a `client` relation, or else this will throw. For example if we wanted to resolve files on a `household` it would break here, because `object.client` doesn't exist. To future-proof this, maybe it can just accept the `client_ids` optionally, and have it be up to the caller to specify them:

```ruby
      def scoped_files(scope, sort_order: :date_created, client_ids: nil)
        # Passing the client ID to the viewable_by scope improves performance
        # client_id = object.instance_of?(Hmis::Hud::Client) ? object.id : object.client.id
        scope = scope.viewable_by(current_user, client_ids: client_ids)
        scope = scope.sort_by_option(sort_order) if sort_order.present?
        scope
      end
```
and on the client
```ruby
resolve_files(object.files, client_ids: [object.id], **args)
```

gigxz (approved):

drivers/hmis/app/models/hmis/file.rb
```diff
@@ -43,17 +43,23 @@ def data_source
...
+
     client_scope = Hmis::Hud::Client.
       viewable_by(user).
       with_access(user, :can_view_any_nonconfidential_client_files, :can_view_any_confidential_client_files)
+    client_scope = client_scope.where(id: client_ids) if client_ids.present?
+
     enrollment_scope = Hmis::Hud::Enrollment.
       viewable_by(user).
       with_access(user, :can_view_any_nonconfidential_client_files, :can_view_any_confidential_client_files)
+    enrollment_scope = enrollment_scope.joins(:client).where(client: { id: client_ids }) if client_ids.present?
```
ttoomey: Personally I favor this style of using a hash to express a condition on a joined table and it's standard practice in active record.

However we have a practice to using arel instead with global variables (c_t). So I think this is more compliant with our practices:

```suggestion
    enrollment_scope = enrollment_scope.joins(:client).where(c_t[:id].in(client_ids)) if client_ids.present?
```

ttoomey (approved):

drivers/hmis/app/models/hmis/file.rb
eanders: The arel version protects against our non-standard (for rails) table naming conventions.  I prefer the hash style _if_ the table name and relation name match.

martha: In this case they do match, plus or minus a capital letter ... so I take it this means we should keep the hash style here?

martha:
Should be retargeted to release-153
Edit: ✅

drivers/hmis/app/models/hmis/file.rb
eanders: No, the OP code **always uses the arel version**.  The capitalization and pluralization are the main reasons to use it.
That was probably unclear above, _I_ (myself, not the OP engineer) prefer the hash version, but as OP engineers, we have standardized on the arel version.

martha: I see, thanks!

PR #5121, opened 2025-02-10
Hotfix for importer core dump

eanders:
## _Merging this PR_

## Description
* Possible fix for a data loading issue that is causing a core dump in some situations

## Type of change
* Bug fix

### DIFF
--- a/app/models/concerns/service_history/builder.rb
+++ b/app/models/concerns/service_history/builder.rb
@@ -122,7 +122,9 @@ module ServiceHistory::Builder
     # Class method
     private def builder_batch_job_scope
       Delayed::Job.uncached do
-        Delayed::Job.where(failed_at: nil).jobs_for_class('ServiceHistory::RebuildEnrollments')
+        Delayed::Job.where(failed_at: nil).
+          where("1 != #{rand(2..50)}"). # Cache buster
+          jobs_for_class('ServiceHistory::RebuildEnrollments')
       end
     end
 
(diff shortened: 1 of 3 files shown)


app/models/concerns/service_history/builder.rb
```diff
@@ -122,7 +122,11 @@ def clients_still_processing?(client_ids:)
     # Class method
     private def builder_batch_job_scope
       Delayed::Job.uncached do
-        Delayed::Job.where(failed_at: nil).jobs_for_class('ServiceHistory::RebuildEnrollments')
+        Delayed::Job.where(failed_at: nil).
+          # Cache buster - for whatever reason `uncached` is completely inconsistent.
+          # We ALWAYS want this to query the database.
+          where("1 != #{rand(2..50)}").
+          jobs_for_class('ServiceHistory::RebuildEnrollments')
```
ttoomey: I wonder if this is because AR evaluation is lazy? So builder_batch_job_scope just returns the scope but it's the caller that executes it (which would be outside of the uncached block). I think that might be what's going on. Can we just return an array?
```suggestion
        Delayed::Job.where(failed_at: nil).
          jobs_for_class('ServiceHistory::RebuildEnrollments').
          to_a # force execution within uncached block
```

FWIW I think 2..50 is too small a space for cache busting, it seems like it would be unreliable as you're likely to generate the same random number repeatedly. You'd probably want a counter of some kind (`@counter += 1`) to avoid repeats

ttoomey (commented):
looks good generally; left some comments on cache busting issue

ttoomey (approved):

PR #5135, opened 2025-02-13
User access query reduction

eanders:
## _Merging this PR_

## Description
* Provides an alternate way of checking project policies which allows for preloads
* Preserves existing ability to check by ID
* memoizes a few more methods to prevent extra duplicate queries

## Type of change
* Bug fix


app/models/grda_warehouse/auth_policies/user_acl_context.rb
```diff
@@ -52,7 +52,7 @@ def permissions_for_collection_ids(collection_ids)
       flat_map(&:granted_permissions).to_set.freeze
   end
 
-  def data_source_collection_ids(data_source_id)
+  memoize def data_source_collection_ids(data_source_id)
```
ttoomey: I was trying to be a little conservative with memoization in the permissions so we don't have n-tiers of redundant memoization.

Should this be memoized? It's only called by data_source_role_permissions which is itself memoized. Alternatively, we could not memoize `data_source_role_permissions` and instead memoize the private methods instead. I'm fine either way

app/models/grda_warehouse/auth_policies/user_legacy_context.rb
```diff
@@ -51,15 +51,15 @@ def initialize(user)
...
+  memoize def permissions_for_access_group_ids(access_group_ids)
     access_group_ids += system_access_group_ids(:data_sources)
     return EMPTY_SET if access_group_ids.blank?
     return EMPTY_SET unless user.access_groups.where(id: access_group_ids).exists?
 
     legacy_permissions
   end
 
-  def data_source_access_group_ids(data_source_id)
+  memoize def data_source_access_group_ids(data_source_id)
```
ttoomey: see note above in user acl context about memoization

app/views/projects/_clients.haml
```diff
@@ -13,7 +13,7 @@
           - @clients.each do |service|
             %tr
               %td
-                - if ! @project.confidential? || can_edit_projects? || ! project.confidential_for_user?(user)
+                - if ! @project.confidential? || can_edit_projects? || ! @project.confidential_for_user?(user)
```
ttoomey, line 16: style without the space after the ! operator is more conventional ([ref](https://rubystyle.guide/#no-space-bang))
```suggestion
                - if !@project.confidential? || can_edit_projects? || !@project.confidential_for_user?(user)
```

dtgreiner (approved):

app/models/grda_warehouse/auth_policies/user_acl_context.rb
```diff
@@ -32,18 +35,53 @@ def initialize(user)
...
+      where(project_id: project_ids).
+      pluck(:project_id, :collection_id).
+      group_by(&:shift).
+      transform_values(&:flatten)
+    @collection_ids_by_project.merge!(results)
+  end
+
+  def collection_ids_for_project(project_id)
+    preload_collection_ids_by_project([project_id]) unless @collection_ids_by_project.key?(project_id)
+    @collection_ids_by_project[project_id] || []
```
ttoomey: Looking at this more closely, I think the null result isn't cached. (preload_collection_ids_by_project wouldn't have ids for projects without collections). So you could cache the null result here (same for coc_codes_for_project and equiv methods in legacy context) to avoid n+1s in that case
```suggestion
    @collection_ids_by_project[project_id] ||= []
```

ttoomey (approved):
looks good, noticed one issue with cache misses on nulls

dtgreiner (approved):

PR #5142, opened 2025-02-14
Fix DisablingCondition and Race compliance issues

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7266

Summary of changes:
- Add new `post_process` method on the base processor, implemented only on `ClientProcessor` for now, updating RaceNone and GenderNone to 99 when no race/gender values have been submitted.
- Add a before_save hook on Enrollment to set DisablingCondition to 99 if it is not present
- Save race/gender values that are not collected as 0, not 99
- Add scripts to the data cleanup util that can be used to update existing data
- Add spec tests:
  - To the form processor, to test the changed race/gender behavior
  - To the `UpdateExternalFormSubmission` mutation, to test the new changes to the ClientProcessor which are relevant to external forms
  - To the enrollment model, to test the changed DisablingCondition behavior
  - To the data cleanup util, to test the new cleanup scripts.

Still to do:
- [x] Finish investigating, and probably also add a fix and a cleanup script for, this related data issue: https://github.com/open-path/Green-River/issues/7266#issuecomment-2660273451
- [x] Run the cleanup scripts with task queue (once the fixes are in prod so more bad data won't be generated?) - added to the task queue

## Type of change


drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -75,6 +75,24 @@ def schema
     def information_date(_)
     end
 
+    def post_process
+      # RaceNone and GenderNone should be set to 99 (Data Not Collected) if no Race/Gender fields are 1.
+      # input_to_multi_fields handles this when data is submitted as an empty list, but
+      # if race/gender data is submitted as individual fields ('Client.Woman' => '0', 'Client.Transgender' => '0')
+      # then we need to update RaceNone/GenderNone to 99 here in post-processing.
```
gigxz, line 82: thanks for the explanation 💯

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
```diff
@@ -75,6 +75,24 @@ def schema
...
+      client = @processor.send(factory_name)
+
+      race_fields = HudUtility2024.races.keys.excluding('RaceNone')
+      any_race = race_fields.any? { |field| client.send(field) == 1 }
```
gigxz: I believe this can be written as 
```suggestion
      any_race = client.race_fields.excluding('RaceNone').any?
```

Using the helper from this concern (shared with wh and hmis client models) https://github.com/greenriver/hmis-warehouse/blob/44113f9cba11910aff1f64b9918ccb472b463dae/app/models/concerns/hud_concerns/client.rb#L16-L24

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -99,6 +99,64 @@ def self.fix_relationship_to_hoh_99s!
...
+        "#{quoted_field} = CASE WHEN #{quoted_field} = 99 THEN 0 ELSE #{quoted_field} END"
+      end.join(', ')
+
+      rows_affected = clients_with_bad_99s.update_all(update_sql)
+      Rails.logger.info "Set #{category} fields 99=>0 on #{rows_affected} Clients"
+
+      # Build a scope that matches clients where ALL of the <Race|Gender> fields are 0
+      clients_with_all_fields_empty = fields.
+        map { |field| clients.where(field => 0) }.
+        reduce { |memo, scope| memo.and(scope) }
```
gigxz: This seems good since it's consistent with how the `or` was constructed above, but it could be done with a one liner (just for fun, take or leave it ) 
```
clients.where(**fields.to_h { |f| [f, 0] })
```

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -99,6 +99,64 @@ def self.fix_relationship_to_hoh_99s!
...
+
+      # IF ALL <Race|Gender> fields are 0 but the <Race|Gender>None field is nil, it should be set to 99 Data Not Collected
+      clients_with_bad_nonefield = clients.where(none_field => nil).merge(clients_with_all_fields_empty)
```
gigxz: can this just be `clients_with_all_fields_empty.where(none_field => nil)` ? it should already be scoped to the correct clients

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -129,6 +129,138 @@ def expect_leaves_non_hmis_data_alone(&block)
     end
   end
 
+  context 'enrollments with DisablingCondition' do
```
gigxz:
```suggestion
  context 'enrollments without DisablingCondition' do
```
?

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
```diff
@@ -129,6 +129,138 @@ def expect_leaves_non_hmis_data_alone(&block)
...
+        :hmis_hud_client,
+        data_source: other_source_ds,
+        RaceNone: 99,
+        GenderNone: 99,
+        **race_fields.map { |field| [field, 99] }.to_h,
+        **gender_fields.map { |field| [field, 99] }.to_h,
+      )
+    end
+
+    it 'updates race/gender values of 99 to 0' do
```
gigxz: This seems like it would be well suited to the test style
```
expect do 
end.to change { client.reload.field }.from(x).to(y).and
  not_change { ... }.and
  etc
```

Since it looks like you're asserting on some fields that _weren't_ supposed to change, and some fields that _were_. It makes it a little easier to read the test because you can see which fields were expected to change.

gigxz (commented):
Looks great! Thanks for the thorough testing!

drivers/hmis/lib/hmis_data_cleanup/util.rb
martha: Cool 😎

drivers/hmis/app/models/hmis/hud/processors/client_processor.rb
martha: Yes looks good, thanks!

drivers/hmis/spec/models/hmis/hmis_data_cleanup_util_spec.rb
martha: Sure sounds good, I wasn't sure if there was a neat way to iterate through all the fields I wanted to make assertions on within a `change` call, but I am happy to make it more explicit

martha: Ok, it is quite a bit more verbose now, but hopefully clearer as well. Let me know if you think it looks okay.

drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -99,6 +99,61 @@ def self.fix_relationship_to_hoh_99s!
       end
     end
 
+    # Change any DisablingCondition values on the Enrollment that are nil to 99
```
gigxz, line 102: We need to add a task to run this on deploy for environments with any HMIS Data Sources. There is some code called like "task queue" etc where you can add a few lines of code to call this, and it will run on the next deploy. Sorry I forget the name of it _Again_ and I can never find it in the code (I know I asked Elliot recently) but Elliot or Dave should know what it's called!

martha: Ok, sounds good! Tagging @eanders for review to make sure `drivers/hmis/config/initializers/hmis_feature.rb` looks okay. No need to deeply review the rest of the PR unless you feel like it. Thank you!

drivers/hmis/config/initializers/hmis_feature.rb
```diff
@@ -15,3 +15,8 @@
 Rails.application.config.queued_tasks[:hmis_check_constraints] = -> do
   Hmis::Tasks::CheckConstraints.check_hud_constraints
 end
+
+Rails.application.config.queued_tasks[:hmis_7266_data_cleanup] = -> do
```
eanders: This seems fine, though I'm not sure I'd use an internal ID in the key.  Can we just call it:
```suggestion
Rails.application.config.queued_tasks[:hmis_disabling_condition_and_race_cleanup] = -> do
```

That said, it does tag it to a particular point in time, so maybe it should just be a rough date:
```
Rails.application.config.queued_tasks[:hmis_disabling_condition_and_race_cleanup_2_2025] = -> do
```

eanders (approved):
TaskQueue setup looks good, just one comment about the naming.  Not critical though.

drivers/hmis/config/initializers/hmis_feature.rb
martha: Thanks! Updated to your second suggestion

PR #5144, opened 2025-02-17
Override authentication salt to allow for forced logouts

eanders:
## _Merging this PR_

## Description
* Fixes an issue in deployed environments where setting `unique_session_id` on a user isn't sufficient to force logout

## Type of change
* New feature
* Bug fix

### DIFF
--- a/app/models/concerns/user_concern.rb
+++ b/app/models/concerns/user_concern.rb
@@ -384,11 +384,16 @@ module UserConcern
     # Enforce a known value is included in the devise salt
     # this allows us to invalidate sessions even though they are stored in redis
     def authenticatable_salt
-      "#{super}#{session_token}"
+      base_salt = super
+      return base_salt if custom_session_invalidator.blank?
+
+      # Poison the salt to force the user to re-login by changing custom_session_invalidator
+      # Make sure the salt isn't changing length
+      Digest::MD5.hexdigest("#{base_salt}#{custom_session_invalidator}")[0, base_salt.length]
     end
 
     def force_logout!
-      update_attribute(:session_token, SecureRandom.hex)
+      update_attribute(:custom_session_invalidator, SecureRandom.hex)
     end
 
     # Dependent on devise expire_password_after being set to a value other than false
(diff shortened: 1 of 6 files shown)


app/views/admin/users/index.haml
```diff
@@ -40,8 +40,8 @@
       %tbody
         - @users.each do |user|
           %tr
-            %td= user.last_name
-            %td= user.first_name
+            %td= link_to user.last_name, edit_admin_user_path(id: user)
+            %td= link_to user.first_name, edit_admin_user_path(id: user)
```
ttoomey, line 44: This is fine although may be confusing that the user name link goes to edit. I know we had some iterations on table nav links in the HMIS; I can't remember if we ended up with a clear rule. @gigxz - do you recall?

db/migrate/20250217181347_add_session_token_to_users.rb
```diff
@@ -0,0 +1,5 @@
+class AddSessionTokenToUsers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :users, :session_token, :string
```
ttoomey, line 3: Calling this "session_token" may be a bit too generic. Perhaps 'custom_session_token'? 

Or include comment that clarifies that it's a mechanism we invented and not part of devise?

app/models/concerns/user_concern.rb
```diff
@@ -381,12 +381,22 @@ def invitation_status
       end
     end
 
+    # Enforce a known value is included in the devise salt
+    # this allows us to invalidate sessions even though they are stored in redis
+    def authenticatable_salt
+      "#{super}#{session_token}"
```
ttoomey: A potential issue is that authenticatable_salt will be different lengths for different users. A user who has had their session reset will get a much longer salt.

I don't know if this matters but it might be worth combining them to get a consistent lengths. I think bitwise OR or a truncated hash could work

Perhaps something like this:
```ruby 
  # Override authenticatable_salt to include our custom token
  def authenticatable_salt
    base_salt = super
    return base_salt if custom_session_token.blank?

    Digest::MD5.hexdigest("#{super}#{session_invalidation_token}")[0, base_salt.length]
  end

ttoomey (approved):
Looks like a good solution! It seems like the only other consumer of authenticatable_salt is rememberable, which we don't use

app/views/admin/users/index.haml
eanders: [This was restored as part of a direct request.](https://github.com/open-path/Green-River/issues/6934#issuecomment-2663214539)
Happy to talk through other mechanisms.

PR #5152, opened 2025-02-20
Prevent bug of auto-exiting already exited client - Hotfix

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/7335

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
+++ b/drivers/hmis/app/jobs/hmis/auto_exit_job.rb
@@ -49,7 +49,7 @@ module Hmis
           Hmis::Hud::Base.transaction do
             # Auto-exit all household members together, setting the exit date equal to the most recent contact for any household member
             household.enrollments.each do |e|
-              auto_exit(e, most_recent_contact, project: project)
+              auto_exit(e, most_recent_contact, project: project) unless e.exit.present?
             end
           end
         end
--- a/drivers/hmis/spec/jobs/auto_exit_job_spec.rb
+++ b/drivers/hmis/spec/jobs/auto_exit_job_spec.rb
@@ -71,10 +71,11 @@ RSpec.describe Hmis::AutoExitJob, type: :model do
 
     context 'with a multi member household' do
       let!(:c2) { create :hmis_hud_client, data_source: ds1, user: u1 }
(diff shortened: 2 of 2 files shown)


drivers/hmis/app/jobs/hmis/auto_exit_job.rb
```diff
@@ -49,7 +49,7 @@ def perform(project_ids: nil, data_source_id: nil)
           Hmis::Hud::Base.transaction do
             # Auto-exit all household members together, setting the exit date equal to the most recent contact for any household member
             household.enrollments.each do |e|
-              auto_exit(e, most_recent_contact, project: project)
+              auto_exit(e, most_recent_contact, project: project) unless e.exit.present?
```
gigxz, line 52: Thinking through the approach` # Get the most recent contact date for the whole household`, this includes contacts that occurred for clients that are already exited. For example:

* Member A: Entered 1/1, served on 2/1
* Member B: Entered on 1/1, exited on 3/1 (with exit assessment)

As a result, Member A would be auto-exited to 3/1 because that's the last contact date in the household. I think that is correct-- if we considered "most recent contact" based on only open enrollments, it could result in a HoH being auto-exited prior to other members existing Exit Dates.

gigxz (approved):

drivers/hmis/app/jobs/hmis/auto_exit_job.rb
martha: Hm, thanks for pointing that out! I think what you're saying makes sense. But there is something still bothering me about this, what if
- Member A: Entered 1/1, no services
- Member B: Entered 1/1, no services, exited on 3/1 because an exit record was created, but _not_ from an exit assessment? <-- We think this shouldn't happen in our HMIS, but is it valid data? Do we need to worry about it? 
- Member A would be auto-exited with exit date of 1/1 or 1/2 (depending on project type)

gigxz: Yeah, good point. Yes it's valid data, but no it can't (currently) happen via our HMIS because Exits are always generated from Exit Assessments. So I think what we have here is reasonable.

PR #5153, opened 2025-02-20
Add last contact date to enrollment

martha:
## _Merging this PR_

## Description
GH issue: https://github.com/open-path/Green-River/issues/6924

Summary of changes:
- Adds a new LastContact schema type with date and type
- ~Aliases `contact_date` on records that are used for determining last contact, so that we can use `map(&:contact_date)`. This replaces the Auto Exit job's [`contact_date_for_entity`](https://github.com/greenriver/hmis-warehouse/blob/51128f221668d2447644bb6e2457ec5b0d5c6ecc/drivers/hmis/app/jobs/hmis/auto_exit_job.rb#L107-L120). Let me know what you think!~
- Adds the `last_contact` field to the Enrollment schema object. 
  - This has similar to the logic for the Auto Exit job, but it also includes Custom Case Notes, which auto exit does not because it's not a HUD field.
  - See inline comment thread about HUD service names
- Adds spec test

## Type of change
New feature

  - [x] ensured permissions and visibility checks are performed in the right places - yes, we determined that Last Contact date/type can be a summary field so anyone with `can_view_limited_enrollment_details` can see.


drivers/hmis/app/models/hmis/hud/service.rb
```diff
@@ -17,6 +17,15 @@ class Hmis::Hud::Service < Hmis::Hud::Base
   belongs_to :client, **hmis_relation(:PersonalID, 'Client')
   belongs_to :user, **hmis_relation(:UserID, 'User'), optional: true, inverse_of: :services
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
+  has_one :hud_service_type, class_name: 'Hmis::Hud::CustomServiceType', primary_key: [
```
martha: I'd like to discuss or pair on this please! The idea is that from the Enrollment schema object, if we have a Service as last contact, we want to be able to get the string representing the service type, like "Bed Night". I was trying to adapt the code [here](https://github.com/greenriver/hmis-warehouse/blob/51128f221668d2447644bb6e2457ec5b0d5c6ecc/drivers/hmis/app/graphql/types/hmis_schema/service.rb#L60-L69) from the Service schema object and not introduce n+1s or get into an infinite loop with the data loader. I'm not sure I've done it correctly or in an acceptable way.

gigxz: I think it would be sufficient to just say "Service" for non-bed-night services. Sorry that wasn't specified! Custom services have names that we don't control, and we don't know if the service name (without the category name) is enough context to explain the thing. Similarly for other HUD services, I don't think it's necessary to list the service type (and they can get really long).

drivers/hmis/app/graphql/types/hmis_schema/last_contact.rb
```diff
@@ -0,0 +1,16 @@
...
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::LastContact < Types::BaseObject
+    skip_activity_log
+
+    field :date, GraphQL::Types::ISO8601Date, null: false
+    field :type, String, null: false
```
gigxz: If we drop the dynamic service name then this should be a fixed set of contact types that we can be represented as an enum, right?  If you think it shouldn't be an enum, I would give this field a different name like "description" or "name". `type` to me feels more like an enum field, rather than a string.

martha: Hm, yeah that seems reasonable! It can be an enum if we drop both dynamic service name and dynamic assessment name too

gigxz: Yeah I don't think we want the assessment name dynamic either, too much variability in length

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -150,6 +150,8 @@ def title
     title
   end
 
+  alias contact_date assessment_date
```
gigxz: I'm not sure if this is worth adding aliases to the HMIS models, since we're only using it in a couple places. This could cause issues later if there were to be a "contact date" added to any hud model (top of mind due to recent discussions around a new date field on Service).

Could this logic be encapsulated in a utility function instead? e.g. `contact_date_for_entity` on Enrollment

drivers/hmis/spec/requests/hmis/enrollment_last_contact_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+    <<~GRAPHQL
+      query EnrollmentsWithLastContactDate(
+        $clientId: ID!
+      ) {
+        client(id: $clientId) {
+          enrollments {
+            nodesCount
+            nodes {
+              id
+              lastContact {
```
gigxz, line 36: `drivers/hmis/spec/requests/hmis/client_enrollments_visibility_spec.rb` this spec tests `can_view_limited_enrollment_details` permission, maybe start resolving lastContact on that test and assert its presence on limited-access enrollments ?

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -5652,6 +5653,63 @@ type KeyValue {
   value: String
 }
 
+type LastContact {
+  date: ISO8601Date!
+  type: LastContactType!
```
gigxz: `contactDate` and `contactType` would make this more descriptive and searchable. With such generic names "date" and "type" it can be tricky to find usages in the codebase. (Sorry, I should've suggested that on my first review)

drivers/hmis/spec/requests/hmis/enrollment_last_contact_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+              type: 'CASE_NOTE',
+            ),
+          )
+        end
+      end
+
+      it 'minimizes n+1' do
+        30.times do
+          enrollment = create :hmis_hud_enrollment, data_source: ds1, client: c1
+          create(:hmis_hud_service_bednight, data_source: ds1, client: c1, enrollment: enrollment)
```
gigxz: Thanks for adding a test case for db queries! I think it could be expanded to include (1) multiple records per enrollment, and (2) other record types beyond bed night.

I was playing around with it just to assess the n+1 for reviewing, and this seems happy. (Tho I had to add a default info date in the case note factory - see other comment about null comparison error)


```ruby
      it 'minimizes n+1' do
        10.times do
          enrollment = create :hmis_hud_enrollment, data_source: ds1, client: c1
          10.times do
            create(:hmis_hud_service_bednight, data_source: ds1, client: c1, enrollment: enrollment)
            create(:hmis_hud_custom_case_note, data_source: ds1, client: c1, enrollment: enrollment)
            create(:hmis_custom_assessment, data_source: ds1, client: c1, enrollment: enrollment)
            create(:hmis_current_living_situation, data_source: ds1, client: c1, enrollment: enrollment)
            create(:hmis_custom_service, data_source: ds1, client: c1, enrollment: enrollment)
          end
        end

        expect do
          response, result = post_graphql(client_id: c1.id) { query }
          enrollments = result.dig('data', 'client', 'enrollments', 'nodes').map(&:deep_symbolize_keys)
          expect(enrollments.map { |e| e[:lastContact] }.compact.size).to eq(enrollments.size)
        end.to make_database_queries(count: 10..30)
      end
```

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
```diff
@@ -449,5 +450,57 @@ def intake_assessment
...
+        # AssessmentDate is guaranteed non-null by DB
+        load_ar_association(object, :custom_assessments, scope: Hmis::Hud::CustomAssessment.order(:assessment_date)).last,
+        # CustomCaseNote's information_date can be null in DB, so sort with nulls_first using arel
+        load_ar_association(
+          object,
+          :custom_case_notes,
+          scope: Hmis::Hud::CustomCaseNote.order(Hmis::Hud::CustomCaseNote.arel_table[:information_date].asc.nulls_first),
+        ).last,
+      ].compact.
+        max_by { |entity| Hmis::Hud::Enrollment.contact_date_for_entity(entity) } # max_by treats nils as lower than any non-nil value
```
gigxz: I'm seeing this raise `comparison of NilClass with Date failed` if any entities have a missing date field. 
Is there a reason not to just omit records with null values from the scopes above?

drivers/hmis/spec/requests/hmis/enrollment_last_contact_spec.rb
```diff
@@ -0,0 +1,175 @@
...
+          expect(query_last_contact).to match(
+            a_hash_including(
+              date: assessment.assessment_date.strftime('%Y-%m-%d'),
+              type: 'ANNUAL_ASSESSMENT', # Specifies the custom assessment name
+            ),
+          )
+        end
+      end
+
+      context 'when custom assessment is the last contact' do
```
gigxz:
```suggestion
      context 'when case note is the last contact' do
```

gigxz (commented):
Looking good!! Nice use of data loaders to prevent n+1 🙌 
There was one issue I noticed with an exception for null dates on case note.

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
martha: Hm, so, maybe I am misunderstanding the comments on [load_ar_association](https://github.com/greenriver/hmis-warehouse/blob/6c41a9d74b937594b6dc868833cb06afd927f680/drivers/hmis/app/graphql/concerns/graphql_application_helper.rb#L35-L38):
> Note: 'scope' is intended for ordering or to modify the default association in a way that is constant with respect to the resolver, for example `scope: FooBar.order(:name)`. It is NOT used to filter down results.

This is the reason I avoided adding a `.where` clause that omits null values from the scopes. But I tried it just now and it does seem to work.

drivers/hmis/spec/requests/hmis/enrollment_last_contact_spec.rb
martha: Cool, thanks! This is more robust 🦾

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
martha: Ok, I updated it https://github.com/greenriver/hmis-warehouse/pull/5153/commits/cccb7094822dc88dac8fa6d71b1b13707eee9b7a and it does make the code a lot more sensible and readable. I got rid of all the confusing comments that were trying to make up for the lack of clarity. It passed the tests, including the n+1, so I guess it should be functional. 

I still feel a bit nervous about it though, since the load_ar_association so explicitly states that the "scope" passed should not be used to filter down results...

gigxz: Good point, sorry for the confusion! That comment should be updated. The "modify the default association in a way that is constant with respect to the resolver" part is accurate. Maybe changing the end to:
> It should not be used to filter down results, _unless_ the filter is constant with respect to the resolver.
> For example: `FooBar.where.not(foo: nil).order(:name)` is ok, but `FooBar.where(name: object.name)` is not


Does that make sense? It just can't be dependent on the resolver at all. (And my bad for missing your original code comment about this)

gigxz (approved):
👍 lgtm!

PR #5156, opened 2025-02-21
6617 filter scopes improvement

ttoomey:
## _Merging this PR_

## Description
* The refactor moves individual filter methods from FilterScopes into separate criteria classes
* Each criteria class follows a consistent pattern with applies? and apply methods
* The original functionality is preserved through a delegation system in FilterScopes that maps to the new criteria classes
* The filter logic itself remains largely unchanged, just reorganized into discrete classes

### Key Improvements:
* Better separation of concerns with each filter in its own class
* More testable code structure with clear boundaries
* Configuration is now explicit through the Configuration class
* Consistent interface across all filter types

### Notes:
* The original race_ethnicity_alternative method was moved to ClientRaceAndEthnicityMixin
* I haven't completed test for all filters

## Type of change
Code clean-up and refactoring


app/models/filters/criteria/filter_for_age.rb
```diff
@@ -0,0 +1,44 @@
+class Filters::Criteria::FilterForAge < Filters::Criteria::Base
+  AGE_RANGES = {
```
ttoomey, line 4: translated from `FilterBase.age_range(description)` - note that I pruned out elements that appeared to be no longer used (five_to_nine, ten_to_fourteen, etc)

app/models/filters/criteria/filter_for_sub_population.rb
```diff
@@ -0,0 +1,9 @@
+class Filters::Criteria::FilterForSubPopulation < Filters::Criteria::Base
+  def applies? = input.sub_population.present?
+
+  def apply(scope)
+    raise "input #{input.sub_population} not allowed" unless input.sub_population.in?(AvailableSubPopulations.available_sub_populations.values)
```
ttoomey, line 8: added this safety check

drivers/performance_measurement/app/models/performance_measurement/client.rb
```diff
@@ -103,68 +103,76 @@ def report_scope_source # needed for filter scopes
       self.class
     end
 
-    private def join_clients_method
-      :source_client
+    protected def criteria_configuration(**opts)
+      defaults = {
+        join_clients_method: :source_client,
```
ttoomey, line 110: I think this is the only place we actually use join_clients_method

app/models/filters/criteria.rb
```diff
@@ -0,0 +1,113 @@
...
+# It uses a factory pattern to instantiate criteria objects based on filter definitions, and provides utilities
+# for working with groups of related criteria.
+#
+# Each criteria class represents a specific type of filter (e.g., age range, project type, etc.) and follows
+# a consistent interface with `applies?` and `apply` methods.
+#
+# @example Creating and applying a single criterion
+#   filter = Filters::FilterBase.new(user_id: current_user.id, age_ranges: [:under_eighteen])
+#   criterion = Filters::Criteria.factory(:filter_for_age, input: filter)
+#   filtered_scope = criterion.apply(scope) if criterion.applies?
```
eanders, line 13: Would it make sense to have `apply` just start with `return scope unless applies?`?
I can see the benefit of being explicit about checking applicability before calling apply, but since they are both instance methods, it seems like the behavior of `apply` when it doesn't apply, would just be to return the initial scope, unless it returns a `none` scope, but I _think_ that would be unexpected.

app/models/filters/criteria.rb
```diff
@@ -0,0 +1,113 @@
...
+  end
+
+  # Criteria definitions mapping criterion IDs to their metadata
+  # Each definition includes:
+  # - tags: Array of symbols categorizing the criterion (e.g., :hud, :warehouse, :client)
+  # - id: Symbol identifying the criterion
+  # - class_name: String name of the criterion class
+  #
+  # @api private
+  DEFINITIONS = [
```
eanders, line 65: Eventually I think this should introspect on the criteria classes, something like:
```
Filters::Criteria::Base.descendants.map(&:definition)
```

app/models/filters/criteria/filter_for_age.rb
eanders: Thanks, it looks like those are in use in `PerformanceMeasurement::EquityAnalysis` but it includes its own definition. Until/if we upgrade that section of the report to use this class, we can safely drop them from here.

app/models/filters/criteria/filter_for_ca_homeless.rb
```diff
@@ -0,0 +1,15 @@
+class Filters::Criteria::FilterForCaHomeless < Filters::Criteria::Base
+  def applies? = input.coordinated_assessment_living_situation_homeless
+
+  # This needs to work correctly with project type filters, where it adds the
+  # potentially additional type of CA, but only if LivingSituation (3.917.1) is
+  # of a homeless type (6, 1, 18)
```
eanders: Probably worth renaming this, CA was the previous name for CE.  Since we're re-writing a bit, can we make this:
```suggestion
class Filters::Criteria::FilterForCeHomeless < Filters::Criteria::Base
  def applies? = input.coordinated_assessment_living_situation_homeless

  # This needs to work correctly with project type filters, where it adds the
  # potentially additional type of CE, but only if LivingSituation (3.917.1) is
  # of a homeless type (6, 1, 18)
```

app/models/filters/criteria.rb
ttoomey: Maybe, I feel like the configuration to add a new criteria is pretty minimal so the maintenance shouldn't be too bad. The explicit configuration also makes it convenient to add attributes (like the tags). It also avoids flaky class-loader issues.

app/models/filters/criteria/filter_for_age.rb
ttoomey: thanks, given how the old filter_for_age_method was written, it seemed like those other attrs were never used

app/models/filters/criteria.rb
ttoomey: Maybe. Right now calling apply() without checking applies? first has undefined behavior. If we want to be defensive we could add a guard against this if it seems like too much of a sharp edge. We could add that on the base class and require children to call super.

```ruby
  def applies?
    super && input.widget_id.present?
  end

  def apply(scope)
    # raises if applies? has not been called
    scope = super(scope)
    # ...filter scope
    scope.where(widget: input.widget_id)
  end
```

app/models/filters/criteria/filter_for_projects.rb
```diff
@@ -0,0 +1,27 @@
...
+    scope.merge(viewable_project_scope).in_project(project_ids)
+  end
+
+  protected
+
+  def visible_project_ids
+    project_ids = []
+    project_ids += input.project_ids || [] if user.report_filter_visible?(:project_ids)
+
+    # note, this is from the original logic: filter by groups even if user cannot filter by project_id
```
eanders:
```suggestion
    # Note: since filtering by project groups amounts to filtering by sets of project ids, and they need to be OR'd with projects so that they aren't limiting to overlaps with chosen projects, we extract the project_ids from the chosen project groups and add them to those from the projects input.
```

This is interesting, and not something I remembered.  I see why it works this way, but I can also see a solid argument for breaking project groups out into their own criterion.  
Looking at `FilterBase#effective_project_ids` the inputs are all OR'd, and very few reports use `anded_effective_project_ids`.  
Given that, I guess I'd be inclined to move ProjectGroups out of this filter for projects for consistency.  What do you think?

app/models/filters/criteria/filter_for_projects_hud.rb
```diff
@@ -0,0 +1,7 @@
+class Filters::Criteria::FilterForProjectsHud < Filters::Criteria::Base
+  def applies? = input.project_ids.present?
+
+  def apply(scope)
+    scope.in_project(input.project_ids).merge(viewable_project_scope)
```
eanders: In app/models/filters/criteria/filter_for_projects.rb there's a note about the order of this scope, does that not apply here as well?

drivers/performance_measurement/app/models/performance_measurement/client.rb
eanders: I agree.  It looks like the default is to join `:clients`, if you have a proposed change to make this less awkward, I'm all ears.
The rest of this model looks like a fantastic cleanup.

spec/models/filters/criteria/filter_for_prior_living_situation_spec.rb
```diff
@@ -0,0 +1,44 @@
+require 'rails_helper'
+require_relative 'shared_filter_criteria_context'
+
+RSpec.describe Filters::Criteria::FilterForPriorLivingSituation do
+  include_context 'filter criteria setup'
+
+  let(:prior_living_situation_ids) { [1, 16] } # Emergency shelter and Place not meant for habitation
```
eanders: These are the FY2022 numbers, they should be `101` and `116`.  Below the `3` for Permanent Housing, should probably be something like `335` (Staying or living in a family member’s room, apartment, or house).

spec/models/filters/criteria/filter_for_projects_spec.rb
```diff
@@ -0,0 +1,65 @@
...
+        create(:role, can_view_project_related_filters: false)
+      end
+      before { allow(user).to receive(:report_filter_visible?).with(:project_ids).and_return(false) }
+      let(:project_ids) { [project.id] }
+
+      it 'ignores project_ids filter' do
+        result = criteria.apply(scope)
+        expect(result.pluck(:id)).to contain_exactly(*enrollments.map(&:id))
+      end
+    end
```
eanders, line 65: I'm probably missing something, but my reading of this was that the user doesn't have access to any of the projects, but the filter is returning all enrollments?  Is that correct?  Maybe that's ok because you should never be able to access a report in this context?

spec/models/filters/criteria/filter_for_race_spec.rb
```diff
@@ -0,0 +1,69 @@
...
+      let(:races) { ['Asian', 'MultiRacial'] }
+
+      it 'returns clients matching either criteria' do
+        result = criteria.apply(scope)
+
+        # Should return white client and multi-racial client
+        expect(result.count).to eq(1)
+        expect(result.pluck(:client_id)).to contain_exactly(multi_racial_client.id)
+      end
+    end
```
eanders: I think we need one more multi-racial client who isn't Asian.  Line 52 should return 2 clients so we can confirm line 64 is correctly filtering down further.

eanders (commented):
Thanks @ttoomey !  This is huge, and a step (maybe many) in the right direction.  Let me know if you want to discuss any of the comments/questions.

app/models/filters/criteria/filter_for_projects_hud.rb
ttoomey: great catch, I'll add a test for it.

drivers/performance_measurement/app/models/performance_measurement/client.rb
ttoomey: great, I don't know that I have a better solution for it- since it's only needed in this once place maybe it's okay to leave it a bit awkward for now

spec/models/filters/criteria/filter_for_prior_living_situation_spec.rb
ttoomey: I missed those, thanks!

spec/models/filters/criteria/filter_for_projects_spec.rb
ttoomey: I tried to make it match is the current behavior of `filter_for_projects`. My understanding of the current filter logic is that if the use lacks the reporting permission, it doesn't apply the filter. If the filter is applied, then it filters for project visibility. Perhaps we rely on the permissions applied in FilterForUserAccess.

Happy to change this behavior if it seems incorrect

```ruby
# current filter
private def filter_for_projects(scope)
  return scope if @filter.project_ids.blank? && @filter.project_group_ids.blank?

  project_ids = if @filter.user.report_filter_visible?(:project_ids)
    @filter.project_ids || []
  else
    []
  end
  project_groups = GrdaWarehouse::ProjectGroup.where(id: @filter.project_group_ids)
  project_groups.each do |group|
    project_ids += group.projects.pluck(:id)
  end

  return scope if project_ids.blank?

  scope.in_project(project_ids.uniq).merge(GrdaWarehouse::Hud::Project.viewable_by(@filter.user, permission: :can_view_assigned_reports))
end
```

app/models/filters/criteria/filter_for_projects.rb
ttoomey: Thanks for the note, I'm not quite clear what effective_project_ids or anded_effective_project_ids are doing in filter base. Let's try and discuss when we huddle next

In principal I agree that breaking these into additional criteria makes sense. However I think I need some additional context. I was hoping to limit scope of this refactor (it's already pretty wide already) so I'm a little cautious about splitting the filter

app/models/filters/criteria/filter_for_data_sources.rb
```diff
@@ -1,11 +1,14 @@
...
   def applies?
     input.data_source_ids.present? && user.report_filter_visible?(:data_source_ids)
   end
 
   def apply(scope)
-    scope.in_data_source(input.data_source_ids).
-      joins(:data_source).
-      merge(GrdaWarehouse::DataSource.viewable_by(user))
+    scope = super(scope)
+    scope.merge(GrdaWarehouse::DataSource.viewable_by(user)).
```
eanders, line 11: Does this need an order of operation comment?

app/models/filters/criteria/filter_for_organizations.rb
```diff
@@ -1,10 +1,13 @@
...
 class Filters::Criteria::FilterForOrganizations < Filters::Criteria::Base
   def applies?
     input.organization_ids.present? && user.report_filter_visible?(:organization_ids)
   end
 
   def apply(scope)
-    scope.in_organization(input.organization_ids).
-      merge(GrdaWarehouse::Hud::Organization.viewable_by(user))
+    scope = super(scope)
+    scope.merge(GrdaWarehouse::Hud::Organization.viewable_by(user)).
```
eanders, line 11: Same here, is order important?

app/models/filters/criteria/filter_for_race.rb
```diff
@@ -1,20 +1,24 @@
+# frozen_string_literal: true
+
+# frozen_string_literal: true
+
```
eanders:
```suggestion
```

spec/models/filters/criteria/filter_for_projects_spec.rb
ttoomey: from our conversation on Friday I think we decided to leave this as-is for now

app/models/filters/criteria.rb
ttoomey: we decided to leave this as-is, adjust in future work as needed

app/models/filters/criteria/filter_for_projects.rb
ttoomey: added some comments to filter base to document effective_project_ids; it's used for project-scoped reports, not enrollments.

I believe we decided to leave the filter as-is to avoid changing behavior

ttoomey:
@eanders summarizing recent (hopefully final) work on this branch
* defensive check on criteria against calling apply if applies? is false
* include report_scope_source in config if it's defined
* factor out add_alternative
* adjust order or permissions and input conditions to avoid clobbering scopes when using merge()
* add criteria specs for data source, destination, disabilities, dv, funders, org, psh & rrh move in, sub pops, times homeless, cocs

spec/models/filters/criteria/filter_for_projects_spec.rb
eanders: Yep

app/models/concerns/filter/filter_scopes.rb
```diff
@@ -7,519 +9,30 @@
...
-      return scope unless @filter.hoh_only
+    # run the criteria on scope if applicable using the current filter as the input
+    def run_applicable_criteria(criterion_id, scope)
+      criterion = Filters::Criteria.factory(criterion_id, input: @filter, config: criteria_configuration)
+      criterion.applies? ? criterion.apply(scope) : scope
+    end
 
-      scope.where(she_t[:head_of_household].eq(true))
+    def criteria_configuration(**opts)
+      # special case handling to allow for @project_types instance var which is seems to override filter.project_type_ids
```
eanders, line 26: I'd like to track down why we have this and if we still need it.  That could be a second phase, but let's make an issue to follow-up.

It looks like: 
```
drivers/core_demographics_report/app/models/core_demographics_report/core.rb:33
/drivers/core_demographics_report/app/models/core_demographics_report/demographic_summary/report.rb:34
drivers/destination_report/app/models/destination_report/report.rb:20
drivers/disability_summary/app/models/disability_summary/disability_summary_report.rb:19
drivers/income_benefits_report/app/models/income_benefits_report/report.rb:140
drivers/prior_living_situation/app/models/prior_living_situation/prior_living_situation_report.rb:20
``` 
implement sets of default project types and a few others set `@project_types = filter.project_type_ids` (which should not affect the filter scopes.)

I'm guessing this was put in place before we had a a good way to set defaults on the filter.  Maybe those could just be pivoted to set the default value of `filter`.

It's also possible, at least some of those are using `@project_types` for something else, I just found places we have `include Filter::FilterScopes` and `@project_types =`

spec/models/filters/criteria/filter_for_dv_currently_fleeing_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+      currently_fleeing: currently_fleeing,
+      start: start_date,
+      end: end_date,
+    )
+  end
+  let(:criteria) { described_class.new(input: filter, config: config) }
+
+  # Create enrollments with different currently fleeing statuses
+  let!(:enrollments) do
+    fleeing_statuses = [1, 0, 2] # Yes, No, Client doesn't know
```
eanders: Doesn't know should be 8 throughout.
```suggestion
    fleeing_statuses = [1, 0, 8] # Yes, No, Client doesn't know
```

spec/models/filters/criteria/filter_for_dv_status_spec.rb
```diff
@@ -0,0 +1,114 @@
...
+    [
+      # Enrollment with DV survivor status (1)
+      create_enrollment_for_client(
+        create(:hud_client),
+      ),
+      # Enrollment without DV survivor status (0)
+      create_enrollment_for_client(
+        create(:hud_client),
+      ),
+      # Enrollment with different DV status (2)
```
eanders, line 31: `DomesticViolenceSurvivor` uses list 1.8, which includes:
```
0 => yes
1 => no
8 => Client Doesn't know
9 => Client prefers not to answer
99 => Data not collected
```
For the purposes of the tests `2` is probably fine, but for consistency we should probably use valid options unless we're testing for bad data.

spec/models/filters/criteria/filter_for_gender_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+
+  let(:genders) { [0, 1] } # Female and Male
+  let(:filter) { ::Filters::FilterBase.new(user_id: user.id, genders: genders) }
+  let(:criteria) { described_class.new(input: filter, config: config) }
+
+  let!(:enrollments) do
+    [
+      create_enrollment_for_client(create(:hud_client, Gender: 0, Woman: 1)),
+      create_enrollment_for_client(create(:hud_client, Gender: 1, Man: 1)),
+      create_enrollment_for_client(create(:hud_client, Gender: 4, NonBinary: 1)),
```
eanders: I'm probably missing something, but why are we passing `Gender` here?  That field was removed with the FY2024 updates.

spec/models/filters/criteria/filter_for_household_type_spec.rb
```diff
@@ -0,0 +1,66 @@
...
+        expect(result.pluck(:id)).to contain_exactly(enrollments[0].id)
+      end
+    end
+
+    context 'when filtering for households with children' do
+      let(:household_type) { :with_children }
+
+      it 'returns only households with children' do
+        result = criteria.apply(scope)
+        expect(result.pluck(:id)).to contain_exactly(enrollments[1].id)
```
eanders: It might be worth adding a test (or source data) to support that this brings back everyone in the multi-person household.  I realize the check is working on the SHE table, so may be unnecessary, but returning both non-HoH and HoH records might add some confidence in this test.

spec/models/filters/criteria/filter_for_indefinite_disabilities_spec.rb
```diff
@@ -0,0 +1,93 @@
...
+      indefinite_disabilities: indefinite_disabilities,
+      start: start_date,
+      end: end_date,
+    )
+  end
+  let(:criteria) { described_class.new(input: filter, config: config) }
+
+  # Create enrollments with different indefinite disability statuses
+  let!(:enrollments) do
+    indefinite_statuses = [1, 0, 2] # Yes, No, Client doesn't know
```
eanders: Same in this file
```suggestion
    indefinite_statuses = [1, 0, 8] # Yes, No, Client doesn't know
```

spec/models/filters/criteria/filter_for_dv_currently_fleeing_spec.rb
ttoomey: thanks!

spec/models/filters/criteria/filter_for_gender_spec.rb
ttoomey: this was slop, removed it. Thanks!

app/models/concerns/filter/filter_scopes.rb
ttoomey: Yeah, this was unclear to me too.

> Maybe those could just be pivoted to set the default value of filter.

For classes that include FilterScopes and set @project_types, I'm not sure there's a clean way to set default filters- maybe I'm missing it though. I suppose we could do it in the initializer by creating a new filter before mutating but I don't love that either. For example:


```diff
--- a/drivers/core_demographics_report/app/models/core_demographics_report/core.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/core.rb
@@ -29,9 +29,10 @@ module CoreDemographicsReport
     attr_accessor :comparison_pattern, :project_type_codes
 
     def initialize(filter)
-      @filter = filter
-      @project_types = filter.project_type_ids || HudUtility2024.homeless_project_types
-      @comparison_pattern = filter.comparison_pattern
+      @filter = filter.dup
+      @filter.project_type_ids.presence ||= HudUtility2024.homeless_project_types
+      @project_types = filter.project_type_ids
+      @comparison_pattern = @filter.comparison_pattern
     end
 
     # The CoC breakdowns aren't relevant to the Core Demographics report right now.
```

So maybe the new issue would be to create a mechanism to set default filters cleanly?

spec/models/filters/criteria/filter_for_race_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+    context 'when filtering for both specific races and MultiRacial' do
+      let(:races) { ['Asian', 'MultiRacial'] }
+
+      it 'returns clients matching either criteria' do
+        result = criteria.apply(scope)
+
+        # Should return white client and multi-racial client
+        expect(result.count).to eq(1)
+        expect(result.pluck(:client_id)).to contain_exactly(asian_multi_racial_client.id)
```
eanders, line 68: This comment doesn't seem to align with the expectations.  I think maybe:
```suggestion
        # Should return asian multi-racial client, but exclude the non-asian multi-racial client.
        expect(result.count).to eq(1)
        expect(result.pluck(:client_id)).to contain_exactly(asian_multi_racial_client.id)
```

spec/models/filters/criteria/filter_for_sub_population_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+    available_sub_populations.each do |method|
+      allow(scope).to receive(method).and_return(scope)
+    end
+
+    # Set up specific return values for our test cases
+    allow(scope).to receive(:veterans).and_return(scope.where(id: enrollments[0].id))
+    allow(scope).to receive(:non_veterans).and_return(scope.where(id: enrollments[1].id))
+    allow(scope).to receive(:adult_only_households).and_return(scope.where(id: enrollments[2].id))
+  end
```
eanders: I'm probably missing something, but by specifying which enrollment to return doesn't this bypass calling the actual scope?

This feels more like a test of the test code than the described class.

Maybe this could be something like:
```
available_sub_populations = AvailableSubPopulations.available_sub_populations
```

And then we could loop over those calling `criteria.apply(scope)` and checking against the expected set of enrollments?  Maybe the enrollments need to be grouped into sets keyed by each population?

eanders (commented):
Thanks @ttoomey , sorry this review was so broken up.  I've made it through again, and I think there were just a few more items that should be addressed.

spec/models/filters/criteria/filter_for_sub_population_spec.rb
ttoomey: Good point, I've improved it to cover all subpopulations

spec/models/filters/criteria/filter_for_household_type_spec.rb
ttoomey: adjusted, lmk if it seems reasonable (testing scope in AdultsWithChildrenSubPop::GrdaWarehouse::ServiceHistoryEnrollmentExtension)

spec/models/filters/criteria/filter_for_race_spec.rb
ttoomey: thanks for catching!

ttoomey:
>  I think there were just a few more items that should be addressed.

I think I got the adjustments to the specs, thanks!

eanders (approved):

PR #5158, opened 2025-02-21
Client Activity Report Optimization

dtgreiner:
* prevent income benefits report from breaking when SHE has not been cleaned up correctly

* filter out orphaned SHEs with join instead of next

## _Merging this PR_

## Description

Add some additional optimization to the client activity report

## Type of change
Bug fix


dtgreiner:
This may not fix everything, but it looks like it should prevent at least one long running query from running twice

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
```diff
@@ -69,7 +69,7 @@ def total_client_count
     end
 
     def clients
-      GrdaWarehouse::Hud::Client.where(id: report_scope.select(:client_id)).
+      @clients ||= GrdaWarehouse::Hud::Client.where(id: report_scope.select(:client_id)).
```
eanders: Memoizing a scope shouldn't actually do anything since it won't be instantiated until called.  Do you know where `clients` is called?  If it's always the same, then maybe it makes sense to drop a `to_a` on the end, but if it's paginated or limited, then it won't.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
```diff
@@ -207,10 +207,13 @@ def max_assessment_by_client_id
...
-      scope = filter.apply(report_scope_base, report_scope_base, include_date_range: false)
-      # Apply a single date filter
-      scope.ongoing(on_date: filter.on)
+      @report_scope ||= {}
+      @report_scope[limited] ||= begin
+        filter.personal_ids_for_days_since_contact_calculations = personal_ids if limited
+        scope = filter.apply(report_scope_base, report_scope_base, include_date_range: false)
+        # Apply a single date filter
+        scope.ongoing(on_date: filter.on)
+      end
```
eanders: I'm not convinced this will return the same data.  `scope.ongoing` is being called within the memoization... maybe it is, but that isn't clear to me.  I _think_ this has the same issue as `clients` in that it's returning a scope so memoizing shouldn't do anything, but I'm not quite as convinced.  
Maybe we could bring in memery and just `memoize` the method to avoid the possible change in function?

dtgreiner:
@eanders hen you have a chance to to a review, this PR should help out with the Client Activity Report. I don't think it fixes all of the optimization needs, but it should speed things up a bit.

eanders (commented):
Sorry @dtgreiner looks like I started a review and never submitted it.  I don't think these are going to work, but maybe you can convince me?

dtgreiner:
@eanders I've added some memoization updates [here](https://github.com/greenriver/hmis-warehouse/pull/5158/commits/bdcdddb06851f78c9b53dd8c2d2d37811c674b6f) that should be a bit more solid.

app/jobs/background_render/inactive_client_report_job.rb
```diff
@@ -15,7 +17,8 @@ def render_html(filters:, user_id:, page:)
     current_user = User.find(user_id)
     @filter = ::Filters::FilterBase.new(user_id: user_id).set_from_params(JSON.parse(filters).with_indifferent_access)
     set_report
-    @pagy, @clients = pagy(@report.clients.order(:last_name, :first_name), page: page, params: @filter.for_params)
+    @clients = @report.clients.sort_by { |c| [c.last_name, c.first_name] }
```
eanders: What happens when `@report.clients` returns 5,000+ clients?  
Is the name sort what is slow?
I _think_ doing this means every page you look at needs to load ALL clients.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
```diff
@@ -71,11 +72,11 @@ def total_client_count
     end
 
     def clients
-      @clients ||= GrdaWarehouse::Hud::Client.where(id: report_scope.select(:client_id)).
+      @clients ||= GrdaWarehouse::Hud::Client.where(id: report_scope_client_ids).
```
eanders: If you're including `memery` I'd say this should be `memoize def clients` instead of setting the instance variable.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
```diff
@@ -122,7 +123,7 @@ def activities
     # Clients the user can see an enrollment for who are also in the report scope
     # This ignores report scope to look for contacts at any project viewable by the user
     def client_scope
-      c_ids = client_ids.presence || report_scope.select(:client_id)
+      c_ids = client_ids.presence || report_scope_client_ids
```
eanders: So this turned a sub-query into a hard-coded list.  That's generally fine if that list is less than a few thousand... if it gets bigger though, that can be an issue.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
dtgreiner: makes sense, I'll include this in the next commit

app/jobs/background_render/inactive_client_report_job.rb
dtgreiner: you may be right, I'll do some testing and make sure its not rechecking this list every page. This pattern was also in the [attachment three client data report](https://github.com/greenriver/hmis-warehouse/blob/bdcdddb06851f78c9b53dd8c2d2d37811c674b6f/app/jobs/background_render/attachment_three_client_data_report_job.rb#L19), so if it does need to be cleaned up, I may adjust that location also.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
dtgreiner: I'm going to revert this one. This subquery was showing slowness when running tests, but the main culprit seemed to be the clients query

dtgreiner:
@eanders I've gone through the recommendations and added some additional cleanup/optimization.

app/jobs/background_render/inactive_client_report_job.rb
```diff
@@ -15,7 +17,7 @@ def render_html(filters:, user_id:, page:)
     current_user = User.find(user_id)
     @filter = ::Filters::FilterBase.new(user_id: user_id).set_from_params(JSON.parse(filters).with_indifferent_access)
     set_report
-    @pagy, @clients = pagy(@report.clients.order(:last_name, :first_name), page: page, params: @filter.for_params)
+    @pagy, @clients = pagy_array(@report.clients.sort_by { |c| [c.last_name, c.first_name] }, page: page, params: @filter.for_params)
```
eanders: I'm still confused as to how this is helping.  This is going to pull back all `@report.clients` every time and then sort them, whereas before it was pulling back only the selected page of 25.

drivers/inactive_client_report/app/models/inactive_client_report/report.rb
```diff
@@ -68,12 +71,12 @@ def total_client_count
       @total_client_count ||= clients.count
     end
 
-    def clients
+    memoize def clients
```
eanders: Do we call `.clients` more than once in a request cycle?  If we do, this might have a big benefit, if we don't, we've just pulled al the data and we'll have to pull it again on the next request.

dtgreiner: yes, I believe it is calling it twice, but it could be three times. I'd have to double check to be sure

eanders: I guess I'd like to see some effort in determining why it is slow, and if there's any way we can speed up the queries `.clients` is making.

drivers/inactive_client_report/app/controllers/inactive_client_report/warehouse_reports/reports_controller.rb
```diff
@@ -35,7 +37,9 @@ def index
     end
 
     def data
-      @pagy, @clients = pagy(@report.clients.order(:last_name, :first_name))
+      query_string = params[:query_string]
+      page = query_string[:page] if query_string
+      @pagy, @clients = pagy(@report.clients.order(:last_name, :first_name), page: page, params: @filter.for_params)
```
eanders: This is odd, (needing to set page).  Can you leave a comment as to why if this is necessary?

dtgreiner: This was breaking when attempting to view the report through the direct link instead of the background job. It was more to make sure this ran during local testing and wasn't part of what I was looking for with the background job response time. I knew I'd be coming back to fix later it so left it in for the QA deployment test.

eanders (approved):
Looks good to me.  Thank you!

dtgreiner:
I might add the backgrounding back in, the report on Boston Staging is feeling slow, so what I was seeing yesterday may have been due to caching

eanders:
> I might add the backgrounding back in, the report on Boston Staging is feeling slow, so what I was seeing yesterday may have been due to caching

That's fine by me, can you add in the ability to run it inline (you've done that before with a query param) as well?  It might be helpful to be able to try both until we've nailed down the query logic.

dtgreiner:
backgrounding has been added back in, including the render_inline functionality for development use

drivers/inactive_client_report/app/controllers/inactive_client_report/warehouse_reports/reports_controller.rb
```diff
@@ -25,7 +27,11 @@ class ReportsController < ApplicationController
     def index
       @excel_export = ::InactiveClientReport::DocumentExports::ReportExcelExport.new
       respond_to do |format|
-        format.html {}
+        format.html do
+          # This is being set so the render_inline feature can be used in development
+          # to override the background running of the report
+          set_pagy_data if params[:render_inline] == '1' && Rails.env.development?
```
eanders: In this case, maybe we should remove the development restriction so we can see performance in production?  At least temporarily?

drivers/inactive_client_report/app/views/inactive_client_report/warehouse_reports/reports/index.haml
```diff
@@ -19,5 +18,4 @@
         = render 'background_render', url: render_section_inactive_client_report_warehouse_reports_reports_path, fetch_params: { filter: @filter.for_params[:filters], query_string: params } do
```
eanders, line 18: Does this need an adjustment to handle the inline version?

dtgreiner: I'll need to pull out the development check for the if part of the if/else controlling this rendering a couple lines above.

PR #5161, opened 2025-02-23
Extend width of client dq fields

gigxz:
## _Merging this PR_

## Description

Fix visual bug by setting fixed width on the DQ fields.

It would be better to fix the default layout behavior more holistically on the frontend (I think the cause is the horizontal group, with the "info group" within it), but I'm taking the less risky/less impactful approach. Looks ok on mobile too.

before:
Screenshot 2025-02-23 at 5 02 50 PM

after:
Screenshot 2025-02-23 at 4 19 48 PM

## Type of change
config change


martha (approved):
Ok, sounds good to me!

PR #5162, opened 2025-02-24
7281 importer performance hotfix and  logging

ttoomey:
## _Merging this PR_

## Description

Details
* hack around bad query plan using postgres configuration
* increase postgres working memory, allow it it to configured via env var
* instrument importer to capture slow query performance and metrics in 'phases'
* add conditional check to skip expensive pre-calculation step ([added in pr 5062](https://github.com/greenriver/hmis-warehouse/pull/5062/files)) 
* add missing index on youth-education-status
* attempt to capture memory use stats

This should not change behavior except for performance

## Type of change
New feature


eanders (approved):
@ttoomey this looks great, thank you!  I'm especially enamored with the SQL logging.  
Is there any chance that's going to make those records HUGE?

ttoomey:
> @ttoomey this looks great, thank you! I'm especially enamored with the SQL logging.
Is there any chance that's going to make those records HUGE?

Thanks @eanders - I've adjusted my use of the notifications subscription after a closer reading of the rails docs. Requesting re-review

It's possible that the captured sql queries could be large but I'm hopeful that compression will keep it manageable. I've added a condition to skip the capture for small queries (less than 100ms).

According to the docs there is overhead to subscribing/unsubscribing as I guess it does mess with rails caching. I think this shouldn't matter but I'd want to keep an eye on it

eanders:
> > @ttoomey this looks great, thank you! I'm especially enamored with the SQL logging.
> > Is there any chance that's going to make those records HUGE?

Running locally with a relatively small file, this ended up with ~4MB of data in that column.  Given that we do load those more than one at a time (occasionally) see `app/controllers/source_data_controller.rb:32` for one example, we should probably add a `diet` scope or similar to make sure we aren't bringing that back unnecessarily.

> It's possible that the captured sql queries could be large but I'm hopeful that compression will keep it manageable. I've added a condition to skip the capture for small queries (less than 100ms).
Sounds good, I'm locally testing with the updated code.

> 
> According to the docs there is overhead to subscribing/unsubscribing as I guess it does mess with rails caching. I think this shouldn't matter but I'd want to keep an eye on it
Yeah, that overhead feels totally fine.

eanders:
@ttoomey I don't think this is quite capturing `post_processing` correctly.

Screenshot 2025-02-24 at 12 14 23 PM

ttoomey:
> @ttoomey I don't think this is quite capturing post_processing correctly.

That's odd, it seems like the only way you'd get that would be that `post_process` didn't return, either it's still running or there was an exception

db/warehouse/migrate/20250226145730_create_idx_on_youth_eds.rb
```diff
@@ -0,0 +1,10 @@
+class CreateIdxOnYouthEds < ActiveRecord::Migration[7.0]
+  disable_ddl_transaction!
+  def change
+    add_index 'YouthEducationStatus', ['EnrollmentID', 'data_source_id', 'PersonalID'], name: 'idx_youth_eds_hud_keys', algorithm: :concurrently
```
eanders, line 4: `algorithm: :concurrently` makes strong migration happy, but often causes issues on deployment.  We should keep an eye on it.

ttoomey: Yeah, maybe we just want to lock YouthEducationStatus instead. I guess it does seem to build pretty quickly even for a few million rows.

ttoomey:
@eanders I made an attempt to refactor `precalculate_change_counts` since metrics suggest it's the most expensive step. Trying a temp table and keeping the whole calculation in the db. Rebuilding fixpoints as a smoke test

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer_log.rb
```diff
@@ -57,5 +60,78 @@ def import_validation_errors_count(filename, files)
...
+      raise if phase.blank?
+
+      self.phase_metrics ||= {}
+      phase_metrics[phase] ||= {}
+      phase_metrics[phase].deep_merge!(args.stringify_keys)
+      save!
+    end
+
+    # for debugging, parse the logs generated by HmisCsvImporter::Importer::Importer#with_sql_log
+    def debug_phases(show_sql: false)
```
ttoomey, line 75: this is helpful for debugging, not sure we want to keep it around long term

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/import_change_calculator.rb
```diff
@@ -0,0 +1,92 @@
...
+        # Clean up temp table even if an error occurs
+        drop_temp_table(temp_table_name)
+      end
+      results
+    end
+
+    private
+
+    # Create a temporary table with incoming data
+    def create_temp_table(klass, incoming_scope)
```
eanders: I wonder if it would make sense to keep these around for more of the process than just the counting.  I was also hoping we could pull in the primary key so we could later join back to the source and destination tables in a more performant way (though maybe `where data_source_id = 1 and ProjectID = '1bc'` is performant enough.)  
If we do the calculations here, and store a table of ids for each process, then later when we go to make the updates/additions/deletions it could just join the tmp table to the appropriate other tables instead of having to recalculate the entire scope.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -80,8 +82,9 @@ def import!(import_log = nil)
       log_timing :cleanup_data_set!
       log_timing :analyze_tables
 
-      # determine what changes will be made and make note for alerting and monitoring
-      log_timing :precalculate_change_counts
+      # Determine what changes will be made and make note for alerting and monitoring. This is only needed if the data source is configured to pause on errors.
+      log_timing :precalculate_change_counts if @data_source.ever_pause_imports_with_errors?
```
eanders: Thinking this through a bit more, I think this should be triggered if there are any import thresholds that would either send a notification OR pause the import.  
I'm good with this for now to prove it works, but before it gets merged we should update.  That will probably require a method or scope on `ImportThreshold` to indicate if it would ever pause or notify.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -588,6 +634,7 @@ def add_new_data
         batch = []
         # This is the same query as `existing_hud_keys` but the memoization of that method prevents this from
         # using it.
+        #  maybe slow (?)
```
eanders: Should we add a `with_sql_log` to this query?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
ttoomey: good point, is this just an additional check for notification:
```suggestion
      log_timing :precalculate_change_counts if @data_source.ever_pause_imports_with_errors? || @data_source.ever_notify_for_imports?
```

eanders: Yes!

PR #5177, opened 2025-03-03
Support rough-out of CE frontend pages and add CE starter pack

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add a "CE starter pack" which isn't meant to be run in production, but can be run in dev databases to generate some basic CE records like templates, swimlanes, and tasks. My understanding is that these should all eventually be configurable by users in-app, but for now, it's helpful to have a quick and re-runnable script that we can share.
- Add a pick list of Workflow Definition Templates to enable selecting a template when creating an opportunity
- for Opportunity schema object,
  - Add pagination to Opportunity Candidates schema field
  - Add accepted referral, alongside active referral
- for Project schema object,
  - add paginated CE Opportunities schema field
- Add `can_view_coordinated_entry` to root access, which is true if CE is enabled in config
- Fix bugs in `Hmis::Ce::Configuration` and `Hmis::Ce::Match::Applicability`

## Type of change
- New feature
- Bug fix

- [na] I have added spec tests (or not applicable)

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -301,7 +301,7 @@ GEM
       momentjs-rails (>= 2.8.1)
     brakeman (7.0.0)
       racc
-    browser (6.2.0)
+    browser (5.3.1)
     builder (3.3.0)
     bumbler (0.9.0)
     bundler-audit (0.9.2)
@@ -1115,7 +1115,7 @@ GEM
       yabeda (~> 0.8)
     yajl-ruby (1.4.3)
     yaml-safe_load_stream3 (0.1.2)
-    zeitwerk (2.7.2)
+    zeitwerk (2.6.18)
 
 PLATFORMS
   arm64-darwin-22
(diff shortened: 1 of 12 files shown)


drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -161,6 +161,12 @@ def self.static_options_for_type(pick_list_type, user:)
           pluck(:OtherFunder).uniq.sort.map do |other_funder|
             { code: other_funder, label: other_funder }
           end
+      when 'WORKFLOW_DEFINITION_TEMPLATES'
+        return [] unless Hmis::Ce.configuration.enabled?
+
+        Hmis::WorkflowDefinition::Template.all.map do |template|
```
martha, line 168: Question about workflow definition templates -- Should they be separated across hmis data sources? Looking in the db, if I'm understanding correctly, the `owner` field is the one that would do this, but it's not enforced right now because that field is nullable. Should it be made required?

drivers/hmis/app/models/hmis/ce/match/opportunity_applicability.rb
```diff
@@ -4,6 +4,9 @@
 # @attr [Object] owner Entity that owns/manages this applicability rule
```
martha, line 4: This is renamed from `applicability_context.rb` to resolve errors: `uninitialized constant Hmis::Ce::Match::Rule::OpportunityApplicability (NameError) in rule.applies_to_opportunity?`

drivers/hmis/app/models/hmis/ce/configuration.rb
```diff
@@ -27,11 +29,11 @@ def days_to_retain_orphan_candidate_pools
     def values
       @values ||= AppConfigProperty.
         where(key: PROPERTIES.map { |attr| key_for(attr) }).
-        index_by { |p| p.key.to_sym }
+        index_by(&:key)
```
martha, line 32: This fixes a bug with `Hmis::Ce.configuration.enabled?` returning false even when there is a value for it in the config table.

Gemfile.lock
```diff
@@ -301,7 +301,7 @@ GEM
       momentjs-rails (>= 2.8.1)
     brakeman (7.0.0)
       racc
-    browser (6.2.0)
+    browser (5.3.1)
```
martha, line 304: I am not sure about the gemfile changes here, I think they were just generated after running `bundle install` since merging in the ruby upgrade changes in release-153, but maybe they should be reverted, or some of these gems should actually be updated to later versions? To investigate more

ttoomey: I think it's fine, we can reconcile with the gemfile lock from 154 before we merge

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
ttoomey: It's a good question, I think the templates would be managed by GR initially so we might not need to solve it for a while. Perhaps the templates will be managed at the data-source level but that's just a guess. I suggest leaving it for now, maybe add a 'fixme' comment

drivers/hmis/app/models/hmis/ce/configuration.rb
ttoomey: thanks!

drivers/hmis/app/models/hmis/ce/match/opportunity_applicability.rb
ttoomey: Thanks for catching! I should add some tests for this

ttoomey (approved):
This all looks great, thanks!

PR #5178, opened 2025-03-03
CVE-2025-27221

ttoomey:
address CVE in URI


eanders (approved):
Pending addition of CGI update

PR #5179, opened 2025-03-04
7348 active client details performance, add pii protection

ttoomey:
## _Merging this PR_

## Description
Fix performance issue on active clients report. It doesn't seem reasonable to render thousands of clients on page
* replace async rendering with paginated list of clients
* fix can_view_clients check
* will change the sort order to be by date, rather than client_id, date
* implements pii protection

## Type of change
Bug/performance fix


app/views/warehouse_reports/client_details/actives/index.haml
```diff
@@ -5,14 +5,16 @@
...
-.d-flex
-  .ml-auto.mb-2
-    - if can_view_clients?
-      = render 'report_downloads/report_download', export: @pdf_export, excel_download_path: warehouse_reports_client_details_actives_path({format: :xlsx}.merge(link_params)), excel_download_tooltip: 'Download includes the following additional fields: Length of Stay (to-date), Living Situation, Destination, Race.'
-= render 'background_render', url: render_section_warehouse_reports_client_details_actives_path, fetch_params: { filter: @filter.for_params[:filters] } do
-  .rollup
-    .rollup-container.well
 
-= render 'common/section_not_loading', path_to_section: section_warehouse_reports_client_details_actives_path(link_params)
+- if can_view_clients?
```
ttoomey: this puts the download button and the list of clients behind can_view_clients

eanders: Does this mean the entire report is behind `can_view_clients?`
I don't think that's the intention, though it does show clients, it is ok that someone sees clients within the context of a report without having `can_view_clients?`.  `can_view_clients?` is intended to cover the client dashboard.  Reports are assigned explicitly, so can potentially expose some client data without the report runner having access to the full client dashboard.

ttoomey: Thanks for the explanation, I'll adjust it back so that only the download button is protected by the condition

ttoomey:
Trivy is still throwing an error even though we have updated the vulnerable gems. Probably a cache but not sure how to clear it.

ttoomey:
moving back to draft

app/views/warehouse_reports/client_details/actives/_list.haml
```diff
@@ -0,0 +1,18 @@
...
+        %th Client
+        %th Project Type
+        %th Project
+        %th Entry Date
+        %th Exit Date
+    %tbody
+      - list.preload(:client).each do |enrollment|
+        - client = enrollment.client
+        %tr
+          %td= link_to_if current_user.can_view_clients?, "#{client.FirstName} #{client.LastName}", appropriate_client_path(id: client.id)
```
eanders: Let's implement the PII provider for name

app/models/active_client_report.rb
```diff
@@ -32,23 +34,8 @@ def value_for_display(_key, value)
...
-      preload(:source_clients).
-      index_by(&:id)
-  end
-
-  def active_client_service_history
-    enrollment_scope.distinct.group_by(&:client_id)
-  end
-
-  private def enrollment_scope
+  def enrollment_scope
```
eanders, line 37: Change order to `first_date_in_program, last_date_in_program`

app/views/warehouse_reports/client_details/actives/index.xlsx.axlsx
```diff
@@ -15,29 +15,27 @@ wb.add_worksheet(name: "Active #{GrdaWarehouse::WarehouseReports::Dashboard::Bas
...
+    length_of_stay = if enrollment.last_date_in_program
+      enrollment.last_date_in_program - enrollment.first_date_in_program
+    else
+      Date.current - enrollment.first_date_in_program
     end
+
+    row = [
+      enrollment.client_id,
+      ::GrdaWarehouse::Config.get(:include_pii_in_detail_downloads) ? client.FirstName : 'Redacted',
+      ::GrdaWarehouse::Config.get(:include_pii_in_detail_downloads) ? client.LastName : 'Redacted',
```
eanders: This should use PII provider too

eanders (approved):
Just a few notes from our discussion, this looks great, thank you!

app/views/warehouse_reports/client_details/actives/index.xlsx.axlsx
```diff
@@ -15,29 +16,30 @@ wb.add_worksheet(name: "Active #{GrdaWarehouse::WarehouseReports::Dashboard::Bas
...
-        enrollment.last_date_in_program,
-        length_of_stay.to_i,
-        enrollment.enrollment_group_id,
-        HudUtility2024.living_situation(enrollment.enrollment&.LivingSituation),
-        HudUtility2024.destination(enrollment.destination),
-      ] + GrdaWarehouse::Hud::Client.race_fields.map { |m| HudUtility2024.no_yes_reasons_for_missing_data(client.public_send(m.to_sym)&.to_i) }
-      sheet.add_row(row)
+  @report.enrollment_scope.preload(:client, :project, :enrollment).find_each do |enrollment|
+    client = enrollment.client
+    pii = client.pii_provider(user: current_user, context: :download)
```
ttoomey: this introduces a lot of n+1 queries. Perhaps we should use ProjectPiiPolicy ?

eanders: That works for me.  I wonder if we should plan to use ProjectPiiPolicy for reports for consistency.

ttoomey: yeah, we could do that- I also wonder if we should plan implement bulk-preload for SouceClientPolicy and then replace ProjectPiiPolicy. If that seems reasonable I might leave this as-is for now. But if that seems unrealistic then I'll switch this to ProjectPiiPolicy, the only issue is that it doesn't handle all the other auth logic (roi, etc)

ttoomey: per team check in about reporting perms, the ProjectPiiPolicy sounds like the best option

app/models/grda_warehouse/auth_policies/project_pii_policy.rb
```diff
@@ -21,6 +23,11 @@ class GrdaWarehouse::AuthPolicies::ProjectPiiPolicy < GrdaWarehouse::AuthPolicie
     end
   end
 
+  def initialize(context:, resource:)
```
ttoomey, line 26: allow this policy to take a project instead of just an id

app/models/grda_warehouse/auth_policies/null_pii_policy.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class GrdaWarehouse::AuthPolicies::NullPiiPolicy
```
eanders: Is this really a `false` policy?  I would think a `null` policy might have unknown consequences.

Either way, can you add a comment about what this is doing and when it is appropriate to use?

app/models/grda_warehouse/hud/client.rb
```diff
@@ -1318,11 +1318,29 @@ def policy_class
       end
     end
 
+    # pii provider for use on client dashboard
     memoize def pii_provider(user:)
       policy = user.policy_for(self)
       GrdaWarehouse::PiiProvider.new(self, policy: policy)
     end
 
+    # pii provider for use in reports and bulk view
```
eanders, line 1327: This looks like a really nice helper, thanks!

eanders (approved):
Thanks @ttoomey

app/models/grda_warehouse/auth_policies/null_pii_policy.rb
ttoomey: Sure thing, added a comment, renamed to to DenyPiiPolicy to better match the behavior

ttoomey:
@eanders my inclination would be to merge this to 154 and send to QA; or do you think this should go in as a hotfix to 153?

eanders:
> @eanders my inclination would be to merge this to 154 and send to QA; or do you think this should go in as a hotfix to 153?

I agree, let's retarget 154.  Looks good to me!

PR #5180, opened 2025-03-04
Analytics views for cohorts and tool to maintain them

eanders:
## _Merging this PR_

## Description
* Provides views for cohort data to be consumed by OP Analytics
* Provides some intermediate tables to represent cohort data in the right shape for OP Analytics
* Provides a mechanism for maintaining the intermediate tables

## Type of change
* New feature


db/views/analytics_cohort_client_changes_v01.sql
```diff
@@ -0,0 +1,2 @@
+SELECT *
```
ttoomey: for all views, I think we should enumerate the column names (that's how they end up in the schema sql anyway)

app/models/grda_warehouse/cohorts/cohort_client_data.rb
```diff
@@ -0,0 +1,58 @@
...
+module GrdaWarehouse::Cohorts
+  class CohortClientData < GrdaWarehouseBase
+    # Maintains cohort data by deleting existing records and inserting updated data.
+    #
+    # @param cohort [Cohort] The cohort for which client data should be maintained.
+    # @return [void]
+    def self.maintain_data(cohort)
+      transaction do
+        where(cohort_id: cohort.id).delete_all
+        batch = []
```
eanders: Batch inserts

db/views/analytics_cohort_client_changes_v01.sql
eanders: Thanks for the reminder!

app/models/cohort_columns/client_id.rb
```diff
@@ -25,5 +27,13 @@ def display_read_only(user)
       html += link_to_if(user.can_access_some_version_of_clients?, value(cohort_client), appropriate_client_path(cohort_client.client), target: '_blank')
       html
     end
+
+    def analytics_value
+      value(cohort_client)
+    end
+
+    def analytics_data_type
+      'integer'
```
ttoomey, line 36: are ids always going to be integers?

app/models/cohort_columns/most_recent_date_to_street.rb
```diff
@@ -21,7 +23,11 @@ def value(cohort_client) # OK
       return unless date.present?
 
       days = (Date.current - date).to_i
-      "#{date&.to_s} (#{days} days)"
+      "#{date} (#{days} days)"
```
ttoomey, line 26: should Date.current be memoized so it can't change mid-run? (@toda ||= Date.current)

db/warehouse/migrate/20250304141815_add_cohort_cache_tables.rb
```diff
@@ -0,0 +1,41 @@
...
+      t.string :column_name
+
+      t.integer :value_integer
+      t.boolean :value_boolean
+      t.string :value_string
+      t.text :value_text
+      t.date :value_date
+      t.jsonb :value_json
+
+      t.string :data_type, comment: 'Indicates which column is in-use for this row of data'
```
ttoomey: should be non-nullable?

app/models/grda_warehouse/cohorts/cohort_analytics_generation.rb
```diff
@@ -0,0 +1,57 @@
...
+        generator = create!(
+          cohort_id: cohort.id,
+          started_at: Time.current,
+          process_name: __method__,
+        )
+        GrdaWarehouse::Cohorts::CohortColumnTitle.maintain_titles(cohort)
+        GrdaWarehouse::Cohorts::CohortClientTab.maintain_tabs(cohort)
+        generator.update!(completed_at: Time.current)
+      end
+    end
```
ttoomey: This is fine but extract move the repeated code to make it clearer /dryer?
```suggestion
    def self.maintain_titles
      act_on_each_cohort(__method__) do |cohort|
        GrdaWarehouse::Cohorts::CohortColumnTitle.maintain_titles(cohort)
        GrdaWarehouse::Cohorts::CohortClientTab.maintain_tabs(cohort)
      end
    end
    
    # yield or each cohort and log processing time 
    def self.act_on_each_cohort(action)
      GrdaWarehouse::Cohort.find_each do |cohort|
        generator = create!(
          cohort_id: cohort.id,
          started_at: Time.current,
          process_name: action,
        )
        yield(cohort)
        generator.update!(completed_at: Time.current)
      end
    end
```

app/models/grda_warehouse/cohorts/cohort_column_title.rb
```diff
@@ -0,0 +1,32 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# This model is an intermediate model used to hold data generated from existing cohorts.
+# The table is maintained by a script and is in the shape we expect to exist after the
+# next cohort rewrite.
+module GrdaWarehouse::Cohorts
+  class CohortColumnTitle < GrdaWarehouseBase
```
ttoomey: I probably am missing something but is 'title' the right name for this model? The properties don't align with with what I would expect to find in a title, (title.title, title.data_type).

app/models/grda_warehouse/cohorts/cohort_column_title.rb
```diff
@@ -0,0 +1,32 @@
...
+          next if col.class.in?(cohort.class.excluded_from_analytics)
+
+          {
+            cohort_id: cohort.id,
+            data_type: col.analytics_data_type,
+            name: col.column,
+            title: col.title,
+            description: col.description,
+          }
+        end.compact
```
ttoomey: I recently learned about `filter_map` which could obviate the call to compact

ttoomey (approved):
Looks good assuming this structure works for the data team

dtgreiner (approved):

app/models/cohort_columns/client_id.rb
eanders: Yes, these are our internal IDs used to link back to the client dashboard.

app/models/cohort_columns/most_recent_date_to_street.rb
eanders: maybe.  It would only mean some people might get an extra day counted, so I think the risk is pretty small.  It would mean the view was generated at the overlap of the day and took a significant amount of time. 

Any thoughts on where we might memoize that?  This model is also used in display of a cohort, so it might be complicated to inject it from the outside.

ttoomey: I guess I was thinking it could be memoized right on the column (or base class); maybe I'm misunderstanding but I thought the same col instance is re-used across clients. Looks like columns get new'd up again each batch of clients so would need an adjustment.

I'm also fine with leaving it as-is if there aren't any realistic risks to inconsistent dates.

PR #5182, opened 2025-03-05
7347 performance dashboard fixes

ttoomey:
## _Merging this PR_

## Description
* fix performance CoC breakdown to use Enrollment.EnrollmentCoc
* prune references to deprecated EnrollmentCoC model
* fix width of "disaggregated by" drop down

## Type of change
bug fixes


app/models/concerns/performance_dashboard/overview/coc.rb
```diff
@@ -24,12 +26,12 @@ def coc_bucket_titles
   end
 
   def coc_bucket(coc)
-    coc
+    coc.presence || 'N/A'
```
ttoomey: seemed like a better label than 'nil'

app/views/clients/rollup/_enrollment_cocs.haml
```diff
@@ -1,12 +1,11 @@
 :ruby
   et = GrdaWarehouse::Hud::Enrollment.arel_table
-  ec_t = GrdaWarehouse::Hud::EnrollmentCoc.arel_table
   enrollment_cocs = @client.
     source_enrollment_cocs.
     joins(enrollment: :client).
     merge(GrdaWarehouse::Hud::Enrollment.visible_to(current_user, client_ids: @client.source_client_ids)).
     preload( enrollment: [:client, :project] ).
-    order(et[:EntryDate].desc, ec_t[:InformationDate].desc)
+    order(et[:EntryDate].desc, et[:id].desc)
```
ttoomey, line 8: guessing that the intention was to provide a tie breaker

app/controllers/performance_dashboards/overview_controller.rb
```diff
@@ -71,7 +73,9 @@ def download
     end
 
     private def set_key
-      @key = report_class.detail_method(params.dig(:filters, :key))
+      key_from_params = params.dig(:filters, :key).presence
+      # FIXME: this may crash because key can be missing from params
```
eanders, line 77: Can we protect against failure by picking a default? I believe the initial page load would have the logic for that.

app/models/concerns/performance_dashboard/overview/coc.rb
eanders: Does `Any` make sense in this context?

app/models/grda_warehouse/hud/enrollment_coc.rb
```diff
@@ -1,9 +1,16 @@
...
+
 ###
 # Copyright 2016 - 2025 Green River Data Analysis, LLC
 #
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+#####
+# DEPRECATED 🚨
+# The EnrollmentCoC class has been deprecated as of 2024. The class is still needed by the
```
eanders:
```suggestion
# The EnrollmentCoC class has been deprecated as of 10/1/2024. The class is still needed by the
```

app/models/report_generators/data_quality/fy2017/q3.rb
```diff
@@ -52,7 +54,7 @@ def columns
         project_name: she_t[:project_name],
         RelationshipToHoH: e_t[:RelationshipToHoH],
         household_id: she_t[:household_id],
-        CoCCode: ec_t[:CoCCode],
+        CoCCode: e_t[:enrollment_coc],
```
eanders: So this is an interesting situation.  In all likelihood, you can no longer run a 2017 DQ report (I'm not 100% sure though) and viewing an old one should show you the data as it was at the time.  However, if you _can_ run one, it should run with the logic from 2017.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2020/base.rb
```diff
@@ -159,7 +161,7 @@ def needs_ce_assessments?
             drug_abuse_entry: [2, 3].include?(disabilities_at_entry.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_exit: [2, 3].include?(disabilities_at_exit.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_latest: [2, 3].include?(disabilities_latest.detect(&:substance?)&.DisabilityResponse),
-            enrollment_coc: enrollment.enrollment_coc_at_entry&.CoCCode,
+            enrollment_coc: enrollment.enrollment_coc,
```
eanders: I don't think we want to change this.  The logic for HUD reports is well specified.  This changes that logic if we ever need to go back and see why an answer is what it is.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2021/base.rb
```diff
@@ -205,7 +207,7 @@ def needs_ce_assessments?
             drug_abuse_entry: [2, 3].include?(disabilities_at_entry.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_exit: [2, 3].include?(disabilities_at_exit.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_latest: [2, 3].include?(disabilities_latest.detect(&:substance?)&.DisabilityResponse),
-            enrollment_coc: enrollment.enrollment_coc_at_entry&.CoCCode,
+            enrollment_coc: enrollment.enrollment_coc,
```
eanders: Same in this file.  This logic should remain  as it was. (and below)

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2023/base.rb
```diff
@@ -210,7 +212,7 @@ def needs_ce_assessments?
             drug_abuse_entry: [2, 3].include?(disabilities_at_entry.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_exit: [2, 3].include?(disabilities_at_exit.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_latest: [2, 3].include?(disabilities_latest.detect(&:substance?)&.DisabilityResponse),
-            enrollment_coc: enrollment.enrollment_coc_at_entry&.CoCCode,
+            enrollment_coc: enrollment.enrollment_coc,
```
eanders: Same in this file

drivers/hud_data_quality_report/app/models/hud_data_quality_report/generators/fy2020/base.rb
```diff
@@ -137,7 +139,7 @@ class Base < ::HudReports::QuestionBase
             drug_abuse_entry: [2, 3].include?(disabilities_at_entry.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_exit: [2, 3].include?(disabilities_at_exit.detect(&:substance?)&.DisabilityResponse),
             drug_abuse_latest: [2, 3].include?(disabilities_latest.detect(&:substance?)&.DisabilityResponse),
-            enrollment_coc: enrollment.enrollment_coc_at_entry&.CoCCode,
+            enrollment_coc: enrollment.enrollment_coc,
```
eanders: and here.

app/controllers/performance_dashboards/overview_controller.rb
ttoomey: My understanding is this is only used when rendering the details modal; at that point we should expect the key to be present, right? I guess my assumption is that this **should** be an error, it would be incorrect to use a default value. If that's not the case, can you help me understand how I could replicate this behavior?

app/models/concerns/performance_dashboard/overview/coc.rb
ttoomey: Perhaps "Data not collected" would be clearer as it borrows the HUD language

app/models/grda_warehouse/hud/enrollment_coc.rb
ttoomey: thanks, I'll added this note to the other EnrollmentCoC classes / mixins

app/models/report_generators/data_quality/fy2017/q3.rb
ttoomey: got it, I'll revert this. I guess if we keep the old code functional we don't need to migrate legacy data.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2020/base.rb
ttoomey: Okay. Sounds like we also need to keep the enrollment_coc_at_entry association on enrollment. I had hoped to remove it as that would better identify places where we missed updating code to reflect 2024 changes.

ttoomey:
Thanks for the review @eanders - I also reverted the change to change on the dq and pit report from prior years:
(drivers/hud_pit/app/models/hud_pit/generators/pit/fy2023/base.rb drivers/h
ud_pit/app/models/hud_pit/generators/pit/fy2022/base.rb  drivers/hud_data_quality_report/app/models/hud_dat
a_quality_report/generators/fy2022/base.rb)

eanders (approved):
@ttoomey I think this looks good.  It's pretty wide ranging, so we should definitely target 154 or maybe even 155.

PR #5183, opened 2025-03-05
Add PII protection for HUD reports

dtgreiner:
## _Merging this PR_

## Description

Adds PII Protection to HUD reports based on the project. This affects APR, CAPER, CE APR, HOPWA CAPER, PIT, PATH, SPM, DQ

## Type of change

New feature


app/helpers/application_helper.rb
```diff
@@ -32,24 +32,36 @@ def site_menu
...
+    return 'Not Specified' if boolean.nil?
+
     case boolean
-    when nil
-      'Not Specified'
     when true, 'Yes'
-      capture do
-        concat content_tag :span, nil, class: 'icon-checkmark o-color--positive' if include_icon
-        concat ' Yes'
+      if include_content_tag
```
eanders: I wonder if this would be cleaner if written such that it determines the text first.
Returns if it doesn't `include_content_tag` and then wraps the value.
Something like
```
text = case boolean
when true, 'Yes
  'Yes'
...
return text unless include_content_tag

css_classes = case text
when 'Yes'
  'icon-checkmark o-color--positive'
...

capture do
  concat content_tag :span, nil class: css_classes if include_icon
  concat " #{text}"
end

File: app/helpers/application_helper.rb [hunk 2]
```diff
@@ -148,10 +160,13 @@ def ssn(number)
...
 
-  def masked_ssn(number)
+  def masked_ssn(number, include_content_tag: true)
     # pad with leading 0s if we don't have enough characters
     number = number.to_s.rjust(9, '0') if number.present?
-    content_tag :span, number.to_s.gsub(HudUtility2024::SSN_RGX, 'XXX-XX-\3')
+    value = number.to_s.gsub(HudUtility2024::SSN_RGX, 'XXX-XX-\3')
+    return value unless include_content_tag
+
+    ActionController::Base.helpers.content_tag :span, number.to_s.gsub(HudUtility2024::SSN_RGX, 'XXX-XX-\3')
```
eanders: Since this is in `ApplicationHelper`, I don't think you need:
```suggestion
    content_tag :span, number.to_s.gsub(HudUtility2024::SSN_RGX, 'XXX-XX-\3')
```

app/models/grda_warehouse/auth_policies/default_pii_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# Some reports now require a pii policy. In cases where that is not available, this policy matches
+# the system behavior prior to this implementation, allowing all fields to be visible.
+class GrdaWarehouse::AuthPolicies::DefaultPiiPolicy < GrdaWarehouse::AuthPolicies::BasePolicy
```
eanders: Can we call this `LegacyPiiPolicy` instead of Default?  I think we want the "default" to be the new policy eventually.

app/models/hud_reports/report_client_base.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
```
eanders: I think I missed a few of these, but they need to go below the copyright, or our copyright update script will be really unhappy.  (And as of rails 6, I think, these are fine being further down in the file.

app/models/hud_reports/report_client_base.rb
```diff
@@ -7,8 +9,107 @@
 # Base class for a report type's clients.
 module HudReports
   class ReportClientBase < GrdaWarehouseBase
+    include ActionView::Helpers
+    include ActionView::Context
```
eanders: If we can get around bringing these in as includes, we should.  It doesn't look like you can just call `ApplicationHelper.yes_no` so I guess we should make a generic ruby class the includes it, something like: ```
class ModelApplicationHelper
  include ActionView::Helpers
   include ActionView::Context
  include ApplicationHelper
end
``` 

Then we can call `ModelApplicationHelper.new.yes_no(false)`.

This probably isn't a fantastic way forward in the future, but should DRY things up a bit for now.

File: drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml [hunk 6]
```diff
@@ -30,11 +32,14 @@
               %th= header
         %tbody.thead-dark
           - records.each.with_index do |record, i|
+            - pii_policy = current_user.policy_for(nil, policy_class: GrdaWarehouse::AuthPolicies::DefaultPiiPolicy)
+            - pii_policy = current_user.policy_for(record.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy) if record.project_id
```
eanders: Can you work with @ttoomey to see if there's a mechanism for preloading the policies/contexts?  It sounds like he may have a solution that would allow us to not use the project policy here.

drivers/hud_apr/app/models/hud_apr/fy2020/apr_client.rb
```diff
@@ -47,5 +49,9 @@ def self.detail_headers_for_export
 
       detail_headers.except('first_name', 'last_name', 'dob', 'ssn')
     end
+
+    def display_value(col, pii_policy:, include_content_tag: true)
+      super
+    end
```
eanders: If it doesn't need to make any changes to super's functionality, I don't think we need this.

app/helpers/application_helper.rb
dtgreiner: this has been cleaned up [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/3a52aa0755560c964a26f94e516edfd1904b94a3)

app/helpers/application_helper.rb
dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/3a52aa0755560c964a26f94e516edfd1904b94a3)

app/models/grda_warehouse/auth_policies/default_pii_policy.rb
dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/3a52aa0755560c964a26f94e516edfd1904b94a3)

app/models/hud_reports/report_client_base.rb
dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/048cb1e9050ed6ecfa744d5c93ae74b1e1c7150c)

app/models/hud_reports/report_client_base.rb
dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/3a52aa0755560c964a26f94e516edfd1904b94a3)

drivers/hud_apr/app/models/hud_apr/fy2020/apr_client.rb
dtgreiner: these have been removed [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/3a52aa0755560c964a26f94e516edfd1904b94a3)

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
dtgreiner: These are being preloaded on Line 20 with `current_user.policy_context.preload_project_dependencies(project_ids)` but we are investigating whether or not this needs to be addressed further due to VA's ROI process.

dtgreiner: It sounds like we should be good with this project-based setup in regards to VA and how they are using ROIs

dtgreiner:
@eanders This should be ready for another check.

app/models/hud_reports/report_client_base.rb
```diff
@@ -4,11 +4,62 @@
...
+
 # Base class for a report type's clients.
 module HudReports
   class ReportClientBase < GrdaWarehouseBase
     self.abstract_class = true
 
     has_many :report_clients, as: :universe_membership, dependent: :destroy
+
+    def display_value(col, pii_policy:, include_content_tag: true)
+      # We are expecting some columns to com through with a dot notation.
```
eanders:
```suggestion
      # We are expecting some columns to come through with a dot notation.
```

db/warehouse/migrate/20250304143108_add_project_id_to_hud_clients.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddProjectIdToHudClients < ActiveRecord::Migration[7.0]
+  def change
+    ['hud_report_dq_clients', 'hud_report_path_clients'].each do |table|
```
eanders, line 5: Do we need a data migration as well, or does the view handle a missing `project_id` appropriately?

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
```diff
@@ -30,11 +32,14 @@
               %th= header
         %tbody.thead-dark
           - records.each.with_index do |record, i|
+            - pii_policy = current_user.policy_for(nil, policy_class: GrdaWarehouse::AuthPolicies::LegacyPiiPolicy)
+            - pii_policy = current_user.policy_for(record.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy) if record.project_id
```
eanders: Maybe this is a bit cleaner?
```suggestion
            - pii_policy = if record.project_id
              - current_user.policy_for(record.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy) 
             - else 
               - pii_policy = current_user.policy_for(nil, policy_class: GrdaWarehouse::AuthPolicies::LegacyPiiPolicy)
            
```
Mostly I'm concerned that calling `current_user.policy_for` may incur a cost.  As written it gets called twice **if we have a project_id**.  Alternatively we could set up a hash of arguments and call `**args`

Not sure which is better.

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
```diff
@@ -30,11 +32,14 @@
...
             %tr
               %th= i + 1
               - headers.each_key do |field|
                 %td{style: 'max-width: 600px;'}
-                  = formatted_cell(record.send(field))
+                  =# formatted_cell(record.send(field))
```
eanders:
```suggestion
```

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_base.rb
```diff
@@ -4,10 +4,13 @@
...
 ###
 
+# frozen_string_literal: true
+
 module HudSpmReport::Generators::Fy2023
   class MeasureBase < ::HudReports::QuestionBase
     def self.client_class
-      HudSpmReport::Fy2023::SpmEnrollment
+      HudSpmReport::Fy2023::SpmEnrollment.
+        preload(enrollment: :project)
```
eanders: This is probably fine, but it seems slightly odd to put this in a method named `client_class`, is there a `client_scope` class we could use?

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_one.rb
```diff
@@ -4,16 +4,18 @@
...
   class MeasureOne < MeasureBase
     def self.question_number
-      'Measure 1'.freeze
+      'Measure 1'
     end
 
     def self.client_class
       HudSpmReport::Fy2023::Episode.
-        joins(:enrollments).preload(:enrollments)
+        joins(:enrollments).preload(:enrollments, enrollments: { enrollment: :project })
```
eanders: I _think_ rails will just eat the first declaration, but it should be unnecessary.
```suggestion
        joins(:enrollments).preload(enrollments: { enrollment: :project })
```

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/measure_one.rb
```diff
@@ -4,16 +4,18 @@
...
   class MeasureOne < MeasureBase
     def self.question_number
-      'Measure 1'.freeze
+      'Measure 1'
     end
 
     def self.client_class
       HudSpmReport::Fy2024::Episode.
-        joins(:enrollments).preload(:enrollments)
+        joins(:enrollments).preload(:enrollments, enrollments: { enrollment: :project })
```
eanders: Same here

drivers/hud_spm_report/app/views/hud_spm_report/cells/show.haml
```diff
@@ -39,8 +42,8 @@
                     %a{href: '#', class: ['btn', 'btn-sm'], data: { toggle: :popover, title: 'Client Information', url: api_client_path(client.client_id), html: 'true' }}
                       %i.icon-info
                   - else
-                    = formatted_cell(cell_value(client, k), k)
-  = content_for :page_js do
+                    = client.display_value(k, pii_policy: pii_policy, include_content_tag: true)
+  = content_for :page_jexits do
```
eanders: I don't think that's quite right.  maybe:
```suggestion
  = content_for :page_js do
```

eanders (commented):
@dtgreiner I think this is getting close.  
Can you check in with @ttoomey on how this may interact with changes in https://github.com/greenriver/hmis-warehouse/pull/5179?

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
ttoomey: jumping in a little here:
* I'm not sure under what circumstance we have blank project_ids but wouldn't it be safest to hide PII in that case?
* if we really do need LegacyPiiPolicy, suggest using a [singleton pattern](https://ruby-doc.org/stdlib-3.1.0/libdoc/singleton/rdoc/Singleton.html) for [example](https://github.com/greenriver/hmis-warehouse/blob/4e85bbc6e83d4d727f37496ed7f5310e6110b91a/spec/models/auth_policies/deny_pii_policy_spec.rb)


```
            - pii_policy = GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance if record.project_id.blank?
            - pii_policy ||= current_user.policy_for(record.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy)
```

db/warehouse/migrate/20250304143108_add_project_id_to_hud_clients.rb
dtgreiner: The views will handle any previously run reports with the Legacy PII Provider, but the migration would be needed to have new reports show/hide PII for the current user based on the project. Without the migration, the DQ and PATH reports would continue show PII solely based on the LegacyPiiProvider.

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
dtgreiner: I've updated the legacy pii policy to use the singleton pattern - [commit](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/200b67b6f9e1ad8a08e0f4a50fde5d0bb4f3bec9)

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/200b67b6f9e1ad8a08e0f4a50fde5d0bb4f3bec9)

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_base.rb
dtgreiner: I've refactored the SPM report to use a `client_scope` method instead of relying solely on the `client_class` method - [commit](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/07d8b1689eb107c18e42a8bb86c2b241804c169a)

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2023/measure_one.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/200b67b6f9e1ad8a08e0f4a50fde5d0bb4f3bec9)

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/measure_one.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/200b67b6f9e1ad8a08e0f4a50fde5d0bb4f3bec9)

drivers/hud_spm_report/app/views/hud_spm_report/cells/show.haml
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5183/commits/200b67b6f9e1ad8a08e0f4a50fde5d0bb4f3bec9).

dtgreiner:
> @dtgreiner I think this is getting close. Can you check in with @ttoomey on how this may interact with changes in #5179?

Nothing is immediately popping out as a conflict between these two PRs, but I'm going to do some further digging to be sure.

dtgreiner:
I've pulled in the lates that includes https://github.com/greenriver/hmis-warehouse/pull/5179 and everything appears to be running as expected. 

@eanders when you have the chance, can you give this another checkover?

drivers/hud_pit/app/views/hud_pit/cells/show.xlsx.axlsx
```diff
@@ -2,12 +2,15 @@ wb = xlsx_package.workbook
 wb.add_worksheet(name: @name.slice(0, 30).gsub(':', ' ')) do |sheet|
   title = sheet.styles.add_style(sz: 12, b: true, alignment: { horizontal: :center })
   sheet.add_row(HudPit::Fy2022::PitClient.detail_headers_for_export.values, style: title)
+  current_user.policy_context.preload_project_dependencies(@clients.distinct.pluck(:project_id))
   @clients.preload(:source_client).each do |client|
+    pii_policy = GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance if client.project_id.blank?
+    pii_policy ||= current_user.policy_for(client.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy)
```
eanders: Let's leave this as is for now, but I wonder if it makes sense in some future version to create a method that returns either the legacy policy or the project policy.  Something like:
```
current_user.reporting_policy_for(project_id: client.project_id)
```
That could return `GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance` if `project_id` is blank.  

It would DRY this up significantly and allow us to change the reporting policy behavior in one central location.

eanders (approved):
Looks good with me.  
Just one comment about future cleanup.  
We can discuss if you think that's a good idea or if we should implement that before merging.

dtgreiner:
> Looks good with me. Just one comment about future cleanup. We can discuss if you think that's a good idea or if we should implement that before merging.

I like the idea and it doesn't sound like it would be a big lift. I'll take a look and see if it's something I can get into this PR. It would be better to do it now if its not too much effort

ttoomey:
Looks like a ~300k line diff in the schema. That's probably not right?

app/helpers/model_application_helper.rb
```diff
@@ -0,0 +1,13 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class ModelApplicationHelper
```
ttoomey: This isn't following the typical pattern for rails helpers in app/helpers, usually those are modules that provide behavior to views and controllers. This seems to be a utility class. It's harmless, just unexpected

ttoomey:
I'll try and get a review completed on this today

app/models/grda_warehouse/auth_policies/legacy_pii_policy.rb
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Some reports now require a pii policy. In cases where that is not available, this policy matches
+# the system behavior prior to this implementation, allowing all fields to be visible.
+class GrdaWarehouse::AuthPolicies::LegacyPiiPolicy
```
ttoomey: My understanding of why this is needed, does this seem right?
> we have reports that were generated before PII protections were implemented. These reports do have the per-client project references needed to determine permissions to see PII. To minimize operational disruption, we fallback to this policy to preserve existing behavior.

```suggestion
# We have reports that were generated before PII protections were implemented. These reports do not have the project references needed to determine permissions for PII. To minimize operational disruption, we fallback to this policy to preserve existing behavior.
class GrdaWarehouse::AuthPolicies::LegacyPiiPolicy
```

If that's correct, I'd suggest we plan to phase this out so we aren't exposing PII perpetually. Maybe programatically we should just raise in the initializer if `Date.today >  LEGACY_BEHAVIOR_GRACE_PERIOD_EXPIRES_ON`

However if this is just an allow-all policy with no other logic, then maybe we should call this 'AllowPiiPolicy' (the inverse of the DenyPiiPolicy) since "legacy" is also how we refer to role-based access.

drivers/hopwa_caper/app/controllers/hopwa_caper/cells_controller.rb
```diff
@@ -13,11 +15,11 @@ def show
       @cell = @report.valid_cell_name(params[:id])
       @table = @report.valid_table_name(params[:table])
       @enrollments = @report.hopwa_caper_enrollments.
-        preload(:enrollment).
+        preload(:enrollment, { enrollment: :project }).
         joins(hud_reports_universe_members: { report_cell: :report_instance }).
         merge(::HudReports::ReportCell.for_table(@table).for_cell(@cell))
       @services = @report.hopwa_caper_services.
-        preload(enrollment: :enrollment).
+        preload(enrollment: [:enrollment, { enrollment: :project }]).
```
ttoomey, line 22: Wow, this association chain is unfortunate (enrollment.enrollment.enrollment.project). Not much to be done about it though

db/warehouse/migrate/20250304143108_add_project_id_to_hud_clients.rb
```diff
@@ -0,0 +1,9 @@
...
+      add_column table, :project_id, :integer
```
ttoomey: My understanding is that Todd is migrating us away from integer pks on production as the part of pg/rds updates.
```suggestion
      add_reference table, :project
```

app/models/user.rb
```diff
@@ -225,6 +225,12 @@ def unique_role_names
     roles.map(&:name).uniq
   end
 
+  memoize def reporting_policy_for_project(project_id:)
```
eanders: Can you add a comment describing what the goal of this method is?

drivers/hopwa_caper/app/views/hopwa_caper/cells/show.haml
```diff
@@ -32,8 +32,7 @@
               %th= header
         %tbody.thead-dark
           - records.each.with_index do |record, i|
-            - pii_policy = GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance if record.project_id.blank?
-            - pii_policy ||= current_user.policy_for(record.project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy)
+            - pii_policy = current_user.reporting_policy_for_project(project_id: record.project_id)
```
eanders, line 35: This sure feels cleaner to me.  Thanks!

app/models/hud_reports/report_client_base.rb
```diff
@@ -4,11 +4,62 @@
...
+      when /prior_living_situation$/
+        HudUtility2024.living_situation(cell)
+      when /.*destination$/
+        HudUtility2024.destination(cell)
+      when /_days_/
+        number_with_delimiter(cell)
+      when /.*length_of_stay$/
+        HudUtility2024.residence_prior_length_of_stay(cell)
+      when /^ssn$/
+        ModelApplicationHelper.new.masked_ssn(cell, include_content_tag: include_content_tag)
```
ttoomey: I think we're trying to retire masked ssn in favor of the pii provider pattern.
* Are we doing this because we only ever want masked ssns in the report so we don't want to obey the policy?
* It looks like `include_content_tag` would just add an enclosing span. Is that span necessary?

```suggestion
         GrdaWarehouse::PiiProvider.viewable_ssn(cell, policy: pii_policy)
```

app/models/user.rb
ttoomey: I don't think we should memoize this as policy_for is already memoized.

```suggestion
 def reporting_policy_for_project(project_id:)
```

ttoomey: Also `GrdaWarehouse::Hud::Client#project_pii_provider` is doing something kind of similar. Maybe we could centralize it? That let's us respect the include_pii_in_detail_downloads configuration too


```diff
diff --git a/app/models/grda_warehouse/hud/client.rb b/app/models/grda_warehouse/hud/client.rb
index a198809f41..9466d9f88a 100644
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -1327,18 +1327,7 @@ module GrdaWarehouse::Hud
 
     # pii provider for use in reports and bulk view
     def project_pii_provider(project:, user:, mode:)
-      allowed = false
-      case mode.to_sym
-      when :download
-        allowed = ::GrdaWarehouse::Config.get(:include_pii_in_detail_downloads)
-      when :browse
-        allowed = true
-      else
-        raise ArgumentError, "Bad mode #{mode}"
-      end
-
-      policy = user.policy_for(project, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy) if allowed
-      policy ||= GrdaWarehouse::AuthPolicies::DenyPiiPolicy.instance
+      policy = user.reporting_policy_for_project(project, mode: mode)
       GrdaWarehouse::PiiProvider.new(self, policy: policy)
     end
 
diff --git a/app/models/user.rb b/app/models/user.rb
index e07b485c6d..65a616cb32 100644
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -225,10 +225,21 @@ class User < ApplicationRecord
     roles.map(&:name).uniq
   end
 
-  memoize def reporting_policy_for_project(project_id:)
-    return GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance if project_id.blank?
+  def reporting_policy_for_project(project_id:, mode: :browse)
+    return GrdaWarehouse::AuthPolicies::LegacyPiiPolicy.instance if project_id.nil?
+
+    allowed = false
+    case mode.to_sym
+    when :download
+      allowed = ::GrdaWarehouse::Config.get(:include_pii_in_detail_downloads)
+    when :browse
+      allowed = true
+    else
+      raise ArgumentError, "Bad mode #{mode}"
+    end
 
-    policy_for(project_id.to_i, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy)
+    policy = policy_for(project, policy_class: GrdaWarehouse::AuthPolicies::ProjectPiiPolicy) if allowed
+    policy || GrdaWarehouse::AuthPolicies::DenyPiiPolicy.instance
   end
 
   memoize def policy_for(resource, policy_class: nil)
```

app/helpers/model_application_helper.rb
ttoomey: some other thoughts:
* this probably can be a singleton(?)
* This feels like a lot just to access the `yes_no` method but I guess we can factor it out later pretty easily. I might tuck this away under app/models/reports for now, I'm not sure this should be used more generally

dtgreiner:
I don't think the release-155 rebase is completely happy here. I may have to rebuild the PR in a new branch.

dtgreiner:
Closing this PR. Instead changes will be targeted [here](https://github.com/greenriver/hmis-warehouse/pull/5203) in order to start with a clean release-155 base.

ttoomey:
> Closing this PR. Instead changes will be targeted https://github.com/greenriver/hmis-warehouse/pull/5203 in order to start with a clean release-155 base.

Fwiw, I might just force-push over the old branch, that we don't need a new PR. Either way is fine though

PR #5187, opened 2025-03-06
Switch from FileMagic to Marcel

eanders:
## _Merging this PR_

## Description
* Hotfix to fight unmaintained file magic gem

## Type of change
* Dependency update

### DIFF
--- a/app/uploaders/ad_hoc_data_source_uploader.rb
+++ b/app/uploaders/ad_hoc_data_source_uploader.rb
@@ -5,6 +5,7 @@
 ###
 
 # encoding: utf-8
+# frozen_string_literal: true
 
 class AdHocDataSourceUploader < CarrierWave::Uploader::Base
   # we will use mini magics API to process attachments
@@ -82,6 +83,7 @@ class AdHocDataSourceUploader < CarrierWave::Uploader::Base
   def whitelist_mime_type_pattern
     Regexp.union WHITELIST
   end
+
   # not sure how this works internally, so adding this duplicate method for now...
   def allowlist_mine_type_pattern
     Regexp.union WHITELIST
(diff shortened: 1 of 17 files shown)


dtgreiner (approved):
Do we want to try and dry up all of the duplicated `content_type_from_bytes` methods while we're in here? Other than that, it looks good to me

eanders:
> Do we want to try and dry up all of the duplicated `content_type_from_bytes` methods while we're in here? Other than that, it looks good to me

I debated cleaning it up, either by removing the method since it only does one thing now, or adding a concern.  Turns out there isn't a base class we can affect easily.  Since we're working toward removing CarrierWave (which is the source of the Uploader models) it didn't feel worth the risk right now.

app/uploaders/ad_hoc_data_source_uploader.rb
```diff
@@ -98,12 +100,9 @@ def extract_file_metadata!
   end
 
   private def content_type_from_bytes(_file_to_test = file)
-    @filemagic ||= FileMagic.new(FileMagic::MAGIC_MIME_TYPE)
-    begin
-      @filemagic.buffer(file.read)
-    rescue StandardError
-      nil
-    end
+    Marcel::MimeType.for(file.read, name: file.name)
```
dtgreiner, line 103: Do we need to include `name: file.name` in all instances of `content_type_from_bytes` or was this specific on running into issues during testing?

dtgreiner (approved):
Looks good to me with just one question

app/uploaders/ad_hoc_data_source_uploader.rb
eanders: My understanding is that you can hint Marcel with a filename (which it uses the extension from) and/or a filetype (presumably fetched by other means).  We only need it when we're passing a string instead of a file, which we mostly do in tests.

PR #5191, opened 2025-03-08
Fix for string encoding regex incompatability

eanders:
## _Merging this PR_

## Description
Fixes [this sentry issue](https://green-river.sentry.io/issues/6368488680/?alert_rule_id=12523843&alert_type=issue&environment=production&notification_uuid=fae9f7f6-955d-4b8b-bae3-f2b5b1c4e202&project=4503977346662400&referrer=slack)
This might be related to the recent change from FileMagic to Marcel, but I'm unclear on that.

## Type of change
* Bug fix

### DIFF
--- a/app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
+++ b/app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
@@ -38,8 +38,9 @@ module GrdaWarehouse::ClientMatcherLookups
     def normalize(str)
       return nil unless str.present?
 
-      # Convert to UTF-8 as the regex expects it
-      str = str.encode('UTF-8', 'ASCII-8BIT', invalid: :replace, undef: :replace, replace: '')
+      # Convert to UTF-8 if we received as ASCII-8BIT as I18n.transliterate expects UTF-8
+      # This conversion will strip weirdly encoded characters silently
+      str = str.encode('UTF-8', 'ASCII-8BIT', invalid: :replace, undef: :replace, replace: '') if str.encoding == 'ASCII-8BIT'
 
       str = I18n.transliterate(str) if transliterate
       str.downcase.strip.gsub(/[^a-z0-9]/, '').presence
--- a/app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
+++ b/app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
@@ -38,9 +38,9 @@ module GrdaWarehouse::ClientMatcherLookups
     def normalize(str)
       return nil unless str.present?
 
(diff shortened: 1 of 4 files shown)


app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
```diff
@@ -36,6 +38,9 @@ def add(client)
     def normalize(str)
       return nil unless str.present?
 
+      # Convert to UTF-8 as the regex expects it
+      str = str.encode('UTF-8', 'ASCII-8BIT', invalid: :replace, undef: :replace, replace: '')
```
ttoomey: This is forcing the source to be treated as ascii even if it is utf8. I might be missing something but it looks like this would strip out valid utf8 chars:
```
> utf8_str = "José"
=> "José"
> result = utf8_str.encode("UTF-8", "ASCII-8BIT", invalid: :replace, undef: :replace, replace: '')
=> "Jos"
```

So maybe we should make this conditional. Assuming that str.valid_encoding can detect this issue. Also maybe we should replace with '?' so that users can see that there was an encoding problem and the data is incomplete
```suggestion
      str = str.encode('UTF-8', 'ASCII-8BIT', invalid: :replace, undef: :replace, replace: '?') unless str.valid_encoding?
```

eanders: Thanks!
Looks like if we don't set `transliterate = true` it throws out non-ascii as well.  Not sure why we aren't transliterating for Identify Duplicates, but for now, I'm not changing it.  
I didn't add `?` as these are never seen, just used in automated matching routines.

I've added a check to only re-encode if the source is ASCII-8BIT, and some tests to expose the current functionality.  Let me know if you think it should be handled differently.

ttoomey: got it, that should be fine. If we trust str.encoding we could probably leave off the condition and let it auto detect the source too, I guess that would handle other encodings potentially
```
str = str.encode('UTF-8',  invalid: :replace, undef: :replace, replace: '')
```

spec/models/grda_warehouse/tasks/identify_duplicates_spec.rb
```diff
@@ -128,6 +136,31 @@
       end
     end
 
+    describe 'When the client has non-ascii characters in their name' do
```
ttoomey, line 139: thanks for adding tests 🙌

ttoomey (approved):

app/models/grda_warehouse/client_matcher_lookups/proper_name_lookup.rb
eanders: I'm up for removing the source encoding, I've adjusted the condition to only run if it isn't already UTF-8.

PR #5192, opened 2025-03-10
Support CE workflow progression

martha:
## _Merging this PR_
- please squash-merge this PR into the long lived feature branch, `tt-ce-spike-3`

## Description
GH issue: https://github.com/open-path/Green-River/issues/7309
Frontend PR https://github.com/greenriver/hmis-frontend/pull/1096

Summary of changes:
- Make schema changes to support frontend CE workflow progression, such as:
  - Turn step and referral statuses into enums instead of strings
  - Return the referral when a step has been completed, so the frontend can display wayfinding step if the step submission has caused the overall referral's status to change
  - Add the project name to the opportunity
    - I added this when referrals were navigationally "hanging out at the root", but later moved them under project, so it's not needed for my current frontend work. However, I haven't removed it, because I think it could still be useful for My Dashboard later on?
    - I used `inverse_of` to help rails out, so hopefully it isn't causing extra db queries anyway. 
  - Add the Client to the referral so frontend can resolve client name and details
- Add a new form role for CE Referral Steps
  - This is a "data collection role" so it can be edited using the form builder. I've explicitly excluded it from the data collection features logic on projects and enrollments.
- Add a simple one-task workflow to the starter pack

Testing instructions:
- Please run the starter pack and let me know how it goes! `rails driver:hmis:ce_starter_pack_20250302`
- I'll add more in-depth testing instructions on the frontend PR.

Still to be done:
- I'd like to add a more complex workflow to the starter pack, but I wanted to put up this PR for feedback on what I have so far.

## Type of change
New feature

  - [ ] ensured the API can't leak data from other data sources - not yet, I think this is something that will need to be done more systematically across the newly introduced CE models. See [thread](https://github.com/greenriver/hmis-warehouse/pull/5177#discussion_r1978054826)
  - [ ] ensured permissions and visibility checks are performed in the right places - not yet, permissions are still to do


drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -91,6 +93,7 @@ class Hmis::Form::Definition < ::GrdaWarehouseBase
     :REFERRAL,
     :REFERRAL_REQUEST,
     :EXTERNAL_FORM,
+    :CE_REFERRAL_STEP,
```
gigxz: To discuss - do we want this under DATA_COLLECTION_FEATURE_ROLES. Do we want referral functionality to be "toggled on and off" using form instances?

I think CE Referral may be too complicated for this data collection feature concept. Maybe we could leave it out for now, until we have more of an intentional conversation about how Coordinated Entry is configured.

martha: Remove from Data Collection Feature roles, and add as "misc" role next to occurrence point form.

Add issue under https://github.com/open-path/Green-River/issues/7190 for CE configuration, questions like:
- turn on ce for project
- choose which templates apply to the project
- does it accept and/or send out direct referrals
- prioritized client lists

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -48,7 +48,139 @@ task ce_starter_pack_20250302: [:environment] do
...
+        "text": "Notes",
+        "type": "TEXT",
+        "link_id": "notes",
+        "disabled_display": "HIDDEN",
+        "mapping": {"custom_field_key": "ce_one_task_notes"}
+      }
+    ]
+  }
+  form_definition.save! if form_definition.changed?
+  # TODO - add CDEDs? Not necessary/used for CE, but the form builder prevents edits if it can't find them
```
martha, line 161: If these forms are created using the editor, then they will have (unused) cdeds created for them on publish anyway.

I guess we could either:
- create some cdeds here in the starter pack, just to help down the road with any form builder edits
- update the form builder to make an exception for certain form types that don't need cdeds

drivers/hmis/app/models/hmis/form/definition.rb
martha: Created https://github.com/open-path/Green-River/issues/7190

gigxz: Thank you! Linking https://github.com/open-path/Green-River/issues/7409 which is I think what you meant!

drivers/hmis/app/graphql/mutations/ce_referral_step_submit.rb
```diff
@@ -4,12 +4,15 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module Mutations
   class CeReferralStepSubmit < CleanBaseMutation
```
gigxz: This mutation class name doesn't match the mutation field name, which is unlike our usual pattern. The mutation field is `submit_ce_referral_step`. Can you rename these mutation classes to match?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -11,7 +11,8 @@ class HmisSchema::CeReferral < Types::BaseObject
     field :id, ID, null: false
     field :opportunity, HmisSchema::CeOpportunity, null: false
     field :steps, [HmisSchema::CeReferralStep], null: false
-    field :status, String, null: false
+    field :status, HmisSchema::Enums::CeReferralStatus, null: false
+    field :client, Types::HmisSchema::Client, null: false
```
gigxz, line 15: Any reason not to load this association with a data loader?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
```diff
@@ -8,19 +8,32 @@
 
 module Types
   class HmisSchema::CeReferralStep < Types::BaseObject
-    field :id, ID, null: false
+    # object is a Hmis::WorkflowExecution::Step
+
+    field :id, ID, null: false # unique identifier for this step based on node and instance
+    field :step_id, ID, null: true # the DB identifier of this step, if it is persisted
```
gigxz: this is ok, a more concise option (instead of specifying `def step_id` down below) is to use `method`:
```suggestion
    field :step_id, ID, null: true, method: :id
```

gigxz: Suggest putting comments into `description:` instead so they are accessible in the graphql schema

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
```diff
@@ -8,19 +8,32 @@
...
+    def step_id
+      object.id
+    end
+
     def workflow_node
       object.node
     end
+
+    def swimlane
+      load_ar_association(object, :node).swimlane.name
```
gigxz: We should use safe accessors here because swimlane is optional on `Task`. And the relation isn't specified at all on `Node`, so I'm not sure how this would behave if the `node` was a StartEvent or EndEvent or some other Node subclass. Do we know what model to expect here?

gigxz: Is it possible to define a `:swimlane` association to the Step model (going `through` node)?
I can see that being useful, but I'm not totally understanding the subclasses and why only certain subclasses have swimlanes.

But that would let us load it directly here:
```
load_ar_association(object, :swimlane)&.name
```
(Current proposed implementation would have n+1 with the `swimlane` lookup which could also be fixed with adding a second call to  use `load_ar_association` to load the swimlane. But it seems preferable to me to just have the association directly)

drivers/hmis/app/graphql/types/hmis_schema/enums/ce_referral_status.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class HmisSchema::Enums::CeReferralStatus < Types::BaseEnum
+    value 'initialized'
+    value 'in_progress'
+    value 'accepted'
+    value 'rejected'
```
gigxz: Maybe generate this enum based on  `Hmis::Ce::Referral.aasm.states.map(&:name)` so its not defined twice

drivers/hmis/app/graphql/types/hmis_schema/enums/ce_referral_step_status.rb
```diff
@@ -0,0 +1,21 @@
...
+#
+#  License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+#
+
+module Types
+  class HmisSchema::Enums::CeReferralStepStatus < Types::BaseEnum
+    value 'unavailable'
+    value 'available'
+    value 'in_progress'
+    value 'completed'
```
gigxz: same here with `Hmis::WorkflowExecution::Step.aasm.states.map(&:name)`

gigxz (commented):
I can't immediately tell where its coming from, but I get this error when I try to load the project page (GetProjects operation). I did run the started pack method. Any ideas?


```
{
    "errors": [
        {
            "message": "Rails couldn't find a valid model for Hmis::Ce::Opportunity association. Please provide the :class_name option on the association declaration. If :class_name is already provided, make sure it's an ActiveRecord::Base subclass.",
            "backtrace": [
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:441:in `rescue in compute_class'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:432:in `compute_class'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:382:in `klass'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:230:in `inverse_of'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:235:in `check_validity_of_inverse!'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/reflection.rb:497:in `check_validity!'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/associations/association.rb:41:in `initialize'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/associations.rb:304:in `new'",
                "/bundle/ruby/3.3.0/gems/activerecord-7.0.8.7/lib/active_record/associations.rb:304:in `association'",
                "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:24:in `block (2 levels) in fetch'",
                "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:23:in `each'",
                "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:23:in `block in fetch'",
                "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:22:in `each'",
                "/app/drivers/hmis/app/graphql/sources/active_record_association.rb:22:in `fetch'",
                "/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader/source.rb:136:in `run_pending_keys'",
                "/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:297:in `block (2 levels) in spawn_source_fiber'",
                "/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:295:in `each'",
                "/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:295:in `block in spawn_source_fiber'",
                "/bundle/ruby/3.3.0/gems/graphql-2.3.20/lib/graphql/dataloader.rb:248:in `block in spawn_fiber'"
            ]
        }
    ],
    "data": {}
}
```

martha:
> this error when I try to load the project page (GetProjects operation)

Hm, I'm not currently able to reproduce the error you're describing :(. But I think it implies there is some typo or un-runnable ruby code in `drivers/hmis/app/models/hmis/ce/opportunity.rb`. The other day I saw a similar error (on different models) during local development, and the issue was related to referencing a validator before its definition.

To help narrow it down, could you please check whether reverting this PR's changes to `drivers/hmis/app/models/hmis/ce/opportunity.rb` fixes the issue for you?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
martha: > I'm not totally understanding the subclasses and why only certain subclasses have swimlanes.

My understanding is that only `Hmis::WorkflowDefinition::Task`s have swimlanes and other types (gateway, start/end event) don't.

I think your suggestion will work if we add the `swimlane` association to Node. I fiddled around with that and got it working here: https://github.com/greenriver/hmis-warehouse/pull/5192/commits/a2c182dc863794a59aa41c170749c7046c3cf71e

If we don't want to do that, because swimlane is only relevant to a specific type of node, then I think we will need to go with your 2nd suggestion using 2 calls to `load_ar_association`

drivers/hmis/app/graphql/mutations/create_ce_opportunity.rb
```diff
@@ -4,8 +4,10 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module Mutations
-  class CeOpportunityCreate < CleanBaseMutation
+  class CreateCeOpportunity < CleanBaseMutation
```
ttoomey: This is fine, it's consistent with our mutation naming. 

Some context: our convention is a awkward because it doesn't group related mutations together alphabetically; it's hard to find them as the number of mutations grow.

If this was greenfield I might advocate more strongly for changing our convention but maybe it's not worth trying to change at this point.

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
ttoomey: It's a good question; I wonder if there's a reason we might need cdeds. If not it would be cleaner to make them optional

ttoomey (approved):
Looks good, thanks!

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
martha: @gigxz @ttoomey Relatedly, have we discussed using Form Processors for ce steps? Does it sound like too much effort? I haven't played around with it and don't really have a clear idea of how it would work exactly. But the CE workflow engine will eventually create enrollments and CE events, sort of reminiscent of the Form Processor's "related records"

drivers/hmis/app/graphql/mutations/create_ce_opportunity.rb
gigxz: Agreed alphabetical grouping is nice. I suggested this because you had the mutation type name and the class name having different names, which I think is confusing for looking up mutations. Changing the mutation field name would be fine with me too, as long as they match!

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
gigxz: I would like to discuss it more, particularly re: reporting needs for CE. (Responses to CE referral task fields will need to be reportable in OP Analytics).

For the related record changes, I think we agreed that it would be sufficient to have a fixed set of "side effects" that CE forms could perform (e.g. create Enrollment, set Move-in Date, create CE Event, set CE Event result..), a.k.a. it's not necessary to make it so flexible that it's able to update _any_ field on Enrollment (for example).

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
gigxz: I suppose I'm a little confused on the **step `belongs_to` node** relationship. (cc @ttoomey ). I think it seems like steps are supposed to belong to _task_ nodes specifically.

If so I think we could adjust the belongs_to relationship- or add an additional one- that specifically is for the Task. Something like the below. Does that make sense? (I haven't been able to fully test this because I'm still working on figuring out creating steps in by db)

```diff
module Hmis::WorkflowExecution
  class Step < GrdaWarehouseBase

    belongs_to :node, class_name: 'Hmis::WorkflowDefinition::Node'
+    belongs_to :task, -> { tasks }, class_name: 'Hmis::WorkflowDefinition::Task', foreign_key: 'node_id'
+    has_one :swimlane, through: :task, class_name: 'Hmis::WorkflowDefinition::Swimlane'
```

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -48,7 +45,139 @@ task ce_starter_pack_20250302: [:environment] do
...
+  gateway = Hmis::WorkflowDefinition::Gateway.find_or_initialize_by(
+    template: one_task_template,
+    gateway_type: 'exclusive',
+    name: 'client acceptance'
+  )
+  start_workflow_event.connect_to!(client_acceptance_task) unless start_workflow_event.outflows.where(target_node_id: client_acceptance_task.id).exists?
+  client_acceptance_task.connect_to!(gateway) unless client_acceptance_task.outflows.where(target_node_id: gateway.id).exists?
+  gateway.connect_to!(reject_workflow_event, condition: 'client_accepted = 0') unless gateway.outflows.where(target_node_id: reject_workflow_event.id).exists?
+  gateway.connect_to!(accept_workflow_event, condition: 'client_accepted = 1') unless gateway.outflows.where(target_node_id: accept_workflow_event.id).exists?
+
```
gigxz, line 179: This comment is for line 223 below
```
clients = Hmis::Hud::Client.all.limit(100) # modify this if you want to include different or specific clients
```
Suggest changing to `Hmis::Hud::Client.hmis.limit(100)`  to avoid issues with trying to resolve clients from different data sources, which causes an error in the frontend:



Screenshot 2025-03-13 at 8 32 04 AM

gigxz:
@martha the error I was seeing seemed to resolve from pulling down the latest changes!

drivers/hmis/app/graphql/mutations/create_ce_opportunity.rb
martha: Would we consider starting to make subdirectories within `drivers/hmis/app/graphql/mutations/`? I don't have too strong a preference, but I'm reflecting that I usually navigating by search, or fuzzy finding filenames, rather than navigating the file structure. I think I'm more likely to search `CreateCeOpportunity` since the majority of our mutations follow the `CreateSomething` pattern.

If not then I can change it back to `CeCreateOpportunity` per Theron's preference.

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
martha: Ok, sounds good, let's discuss more. I think for this PR I'd like to leave this as-is. I'll split out a separate ticket to fully enable editing form definitions for CE using the form builder, since I think there's more discussion needed there.

Edit: Created https://github.com/open-path/Green-River/issues/7414

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
martha: I _think_ that a step is more like an instance of a node. Something like this:
- `WorkflowDefinition::Template` has many `Node`s
- A `Node` can be a `StartEvent`, `EndEvent`, `Gateway`, or `Task` -- anything that subclasses `Node`
- `WorkflowExecution::Instance` has many `Step`s, each of which corresponds to one `Node` from the instance's `Template`

martha: Hm I think I'm wrong about the last part - it looks like you're right that Steps are only instantiated for Tasks and not for other node types. I think your suggestion will probably work, I'm testing it now.

gigxz: Thanks, I agree that Step is an "instance" of a node. I'm wondering if Step is _always_ an instance of a node with type **Task**, and _never_ an instance of a node with type **StartEvent**, **EndEvent**, or **Gateway**.

Looking back at Theron's PR description https://github.com/greenriver/hmis-warehouse/pull/5013 it does confirm that Step is related to the Task node type specifically.

Also looking at where `Step`s get initialized ([here](https://github.com/greenriver/hmis-warehouse/blob/45ccaf6acc35163371d992ee7e507a164367acff/drivers/hmis/app/models/hmis/workflow_execution/engine.rb#L105-L121
) and [here](https://github.com/greenriver/hmis-warehouse/blob/45ccaf6acc35163371d992ee7e507a164367acff/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb#L19-L20)) they are initialized **only** with nodes of type Task.

If my understanding is correct, I'd suggest reverting the changes from https://github.com/greenriver/hmis-warehouse/commit/a2c182dc863794a59aa41c170749c7046c3cf71e and adding the suggested relationships from https://github.com/greenriver/hmis-warehouse/pull/5192#discussion_r1993413633. It seems clearer to me because the swimlane relation only exists on Task.
Now that I have data I did test it and it seems to work! Additional changes are:
* new scope on node `scope :tasks, -> { where(type: 'Hmis::WorkflowDefinition::Task') }` which is used to filter down the `task` relation on step, and

drivers/hmis/app/graphql/mutations/create_ce_opportunity.rb
ttoomey: Sub directories do sound appealing, maybe worth a brief exploration? Thanks for fixing the class/filename-mismatch!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
gigxz: ok looping back from huddle, I misunderstood class_name magic and we don't actually need the scope

gigxz (approved):
looks good!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
martha: I _think_ that's the case although I haven't really found any good documentation for it! https://guides.rubyonrails.org/association_basics.html#class-name

ttoomey: Sorry for the confusion! To confirm:
* Yes, a Step is an "instance" of Task, Task is a subclass of node
* Other node-types do not have Steps because the aren't intractable by the user. Tasks are for user actions (forms) while the other node subclasses are for control flow.

Happy to jump on a huddle and try to answer questions!

I tried to call this out a bit in the inline-comments, I can try to make another pass on them.
```
# task.rb
# Represents a task that needs to be performed by a user in an execution step
# Tasks require form completion to proceed which is tracked corresponding execution steps
# Tasks may include a swimline which can be used to determine which user is assigned to the corresponding step
```
```
# step.rb
# Represents an individual task instance within a workflow execution.
# Tracks the state, assignments, and completion data for a specific
# task node.
```

drivers/hmis/app/graphql/mutations/create_ce_opportunity.rb
martha: Sounds good! Moving them into the `ce` subdirectory involved giving them their own namespace too, so `Mutations::CreateCeReferral` => `Mutations::Ce::CreateCeReferral`, etc. I think this is OK, but let me know if any objections.

 https://github.com/greenriver/hmis-warehouse/pull/5192/commits/e5dec6bcb500d83523aa8343c1db5084b09ac440

martha:
I'll merge this to the feature branch now, so I can more easily rebase and prepare new PRs; but happy to keep addressing feedback on future PRs!

PR #5195, opened 2025-03-11
Validate step responses in CE engine

martha:
## _Merging this PR_
- please squash-merge this branch into the long-lived feature branch tt-ce-spike-3

## Description
Summary of changes:
- ~Adapt the Referral Execution Engine `Hmis::WorkflowExecution::Engine` to follow a "dependency injection"-type pattern, enabling us to instantiate a dry-run engine. Dependencies injected are:~
  - ~Message handler - the normal `ReferralMessageHandler` responds to messages like `start_referral` and `create_wip_enrollment` by making changes and saving them in the database, while the `DryRunMessageHandler` just collects and reports on what messages were called.~
  - ~"Stepper" - this could probably have a clearer name, or maybe needs more general refactor. It's responsible for calling methods on `step`s -- so the real referral stepper calls e.g. `step.start!` and `step.complete!` while the dry run stepper calls `step.start` and `step.complete` without saving.~
  - ~Assignment handler: nil for dry run (no assignment)~
  - ~Audit logger: nil for dry run (no audit logs)~
- Add a `confirmed` flag, which is still unused for now.
- Add validation in the engine, using the validator already built into the definition. This should catch required fields that are not passed, and raise if unrecognized fields are passed.
- Workflow change: Enable looping back to a previously completed step, as in admin referral denial workflow.
- Spec files:
  - Add new spec for the `SubmitCeReferralStep` mutation
  - Refactor some of the fixtures from `ce_referral_spec.rb` into a shared helper

How to test:
- Using the template and one-task opportunity created by the starter pack (documented [here](https://github.com/greenriver/hmis-frontend/pull/1096)),
- ~Try submitting the step with the answer "No, client does not accept." You should see a warning modal, have the opportunity to confirm, and then see the referral get declined.~
- Try submitting the step with no answer selected. You should see error validations.

## Type of change
Feature build-out


ttoomey:
sorry I haven't completed a review this yet, I'll try and get that done today!

martha:
Thanks @ttoomey , no worries and no rush! It's still in draft, and I'm still mulling over some todos. I mainly just wanted to put up the PR in order to get early feedback on this "dependency injection" engine idea (https://github.com/greenriver/hmis-warehouse/pull/5195/files#diff-0a2d03daccfa09bc5306391cf90c599a515152152db533ed6ebad81ad5d8874aR32-R38), but I'm not blocked or anything

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
```diff
@@ -16,23 +16,65 @@ class Ce::SubmitCeReferralStep < CleanBaseMutation
...
+        )
+
+        step = dry_run_engine.active_steps.find(step_id)
+
+        errors.push(*validate(step, input))
+        return { errors: errors } if errors.any? # return early - if there are validation errors, they may cause the dry run to fail
+
+        dry_run_engine.complete_step!(step, user: current_user, submitted_values: input)
+
+        # todo @martha - this isn't flexible enough, it needs to be configurable which types of messages require a warning.
```
martha: I'd like to think this through together before proceeding. 

Currently, the MessageHandler only handles events happening to the whole _referral_ (like start referral, reject referral, etc.) But it doesn't handle events happening on individual steps, like start step, complete step.

That seems easy enough to address, maybe using the new Stepper I've introduced here, to collect messages similarly to how DryRunMessageHandler does. However, once we do, we still need some additional configuration, maybe a new boolean column on Nodes, indicating whether starting a given node is a warning-level event.

Does that sound right? Does it sound worth it? Or should we just say that only referral events can trigger warnings, not step events?

drivers/hmis/app/models/hmis/workflow_execution/step.rb
```diff
@@ -26,7 +26,7 @@ class Step < GrdaWarehouseBase
 
       # node can be started by a user
       event :enable do
-        transitions from: :unavailable, to: :available
+        transitions from: [:unavailable, :completed], to: :available
```
ttoomey, line 29: what's the need for this?

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
ttoomey: Yes, I imagine you'd probably want to just as the message_handler for the relevant messages rather than handling each case in the mutation.

It may not be worth if we could make do with warnings or guidance inlined into the steps forms using existing form definitions.

If the need is less for automatic confirmation based on the type of action (email, enrollment, etc) then perhaps we do need another approach. Your suggestion to build add flags to to nodes could work. However if we do need to support a richer, more integrated confirmation mechanism, I'd lean towards extending the already existing trigger configuration. Maybe:

```ruby
# In node.rb
def confirmation_config
  (trigger_config || []).filter { |trigger| trigger['event'] == 'confirmation' }
end

def requires_confirmation?(submitted_values = {})
  confirmation_config.any? do |config|
    evaluate_condition(config['condition'], submitted_values)
  end
end
```

Then engine.complete_step could return a result object.
```ruby
def resolve(referral_id:, step_id:, input:, confirmed: false)

  referral = Hmis::Ce::Referral.viewable_by(current_user).find(referral_id)
  submission_result = nil
  
  referral.opportunity.with_lock do
    engine = referral.workflow_engine
    step = engine.active_steps.find(step_id)
    submission_result = engine.complete_step!(
      step, user: current_user, submitted_values: input, confirmed: confirmed
    )
  end

  {
    step: submission_result.step,
    referral: submission_result.completed? ? referral.reload : referral,
    confirmation_messages: submission_result.confirmation_messages
  }
end
```

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -6,47 +6,59 @@
 # task assignments, and message processing according to the workflow template.
 module Hmis::WorkflowExecution
   class Engine
-    attr_reader :template, :instance, :message_handler, :assignment_handler
+    attr_reader :template, :instance, :message_handler, :stepper, :assignment_handler, :audit_logger
 
-    def initialize(workflow_instance, message_handler:, assignment_handler:)
+    def initialize(workflow_instance, message_handler:, stepper:, assignment_handler: nil, audit_logger: nil)
```
ttoomey: The additional dependency injection is fine although it does make invoking the engine more complicated. If we keep these, we might want to consider default options that provide expected behavior (log and save steps by default).

I had exposed dependencies that I thought might require different behavior if we were to adapt the engine to serve workflows outside CE.

ttoomey (commented):
Looks good, thanks! I wanted to confirm that the dry-run behavior is still correct. Maybe we can catch up on this on Monday.

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
martha: Thanks Theron, this code sketch is really helpful! I like this idea. Let's discuss at dev standup, since I also agree with
> It may not be worth if we could make do with warnings or guidance inlined into the steps forms using existing form definitions.

My recollection from our design meetings is that:
- We first added the dry-run/confirm warnings idea based on the idea that configurable forms means entrusting any warning elements for irreversible actions to end users, and we want to avoid that
- However, after more discussion with design, we agreed that having a warning in the form (pre-submit) is a requirement, and we may add new form elements to support this(?)
- so we may have elided the need for an infinitely configurable warning confirmation?

drivers/hmis/app/models/hmis/workflow_execution/step.rb
martha: Ah, yeah I should have flagged this for additional discussion too. This enables looping back to a previous step, for example in the Admin Review Denial workflow, where an Admin can deny the denial, sending the workflow back to the previous step.

I played with this a little locally and it seems to work ok, but I understand the workflows were originally designed to be acyclic, so let's discuss more and figure out if this is going to break something unexpected.

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
```diff
@@ -16,23 +16,64 @@ class Ce::SubmitCeReferralStep < CleanBaseMutation
...
         referral: referral.reload,
       }
     end
+
+    def validate(step, submitted_values)
+      definition = step.node.form_definition
+      return unless definition && submitted_values
+
+      definition.validate_form_values(submitted_values)
+    end
```
gigxz: as discussed at dev standup - makes sense to move this logic into the step

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
martha: Discussion notes from dev standup
- With the current approach, we could just use the `send_notification` event rather than notifying on step change
- But we would also want this to be configurable per referral template. Not all templates want to warn/confirm on every email notification
- In the future let's consider building confirmation into the trigger config
- For now let's keep the API changes with the `confirmed` flag, but revert the dry run engine since it adds unneeded complexity
- Move validation logic into the step itself, later this will be flexible to return warnings as well as errors

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
martha: Thanks yep! I've added a new method on the engine, rather than doing this as part of `complete_step`, in order to more gracefully respond to it in the mutation. This was roughly based on the submit form mutation, which [collects validation from the form processor](https://github.com/greenriver/hmis-warehouse/blob/ab8d55969a9a737fe4df3bbd6f44d593a5e9d116/drivers/hmis/app/graphql/mutations/submit_form.rb#L68-L71) before running. Let me know if you think it looks okay.

martha:
Ok, @gigxz @ttoomey thank you for the helpful discussion earlier. I think this is ready for you guys to take another look. Thank you!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -32,6 +32,11 @@ def start_step!(step, user:)
       log_event('start_step', user: user, step: step)
     end
 
+    def validate_step(step, submitted_values:)
+      definition = step.node.form_definition
+      definition.validate_form_values(submitted_values) if definition.present? && submitted_values.present?
+    end
```
ttoomey: Not sure about the conditional validation:
- I think task nodes are required to have a definition. We shouldn't need to check it's present unless I'm missing something
- we should validate even if the submitted values were empty, right?

```suggestion
    def validate_step(step, submitted_values:)
      definition = step.node.form_definition
      definition.validate_form_values(submitted_values) 
    end
```

ttoomey (approved):
Looks good, thank you!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Makes sense, thanks!

gigxz (approved):

PR #5196, opened 2025-03-12
7375 spm days homeless

ttoomey:
## _Merging this PR_

## Description
Bug fixes and tests for SPM measure 1

## Type of change
Bug fix


drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
```diff
@@ -107,12 +110,18 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
         )
       end
 
-      dates = filtered_bed_nights.map(&:last)
+      first_date, last_date, days_homeless = compute_min_max_and_unique(filtered_bed_nights.map(&:last))
```
ttoomey: several issues here:
* filtered_bed_nights were not sorted as seems to have been the assumption. This led to inaccurate start/end dates for episodes
* the dates themselves were not de-duplicated before counting to find the days_homeless which led to inaccurate durations

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
```diff
@@ -107,12 +110,18 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
...
+
+      # filtered_bed_nights appears to be an array of [enrollment, index(?), date]
+      literally_homeless_at_entry = filtered_bed_nights.any? do |enrollment, _, date|
+        enrollment_literally_homeless_at_entry(enrollment) if date == first_date
+      end
```
ttoomey: previously this was only looking a the enrollment on one date that matched.

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/measure_one.rb
```diff
@@ -213,12 +213,15 @@ def run_question!
     end
 
     private def median(values)
```
ttoomey, line 217: This implementation was incorrect:
```
> median([1,2,3])
=> 1 # should be 2
```

ttoomey:
@eanders I need to give this another self-review but I think I fixed a few relevant bugs; maybe I should expand the tests further? Or put a pin in it and get it into 154

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
eanders: Can you add a comment about what this is trying to do? 
My, probably incorrect, understanding is that you are looking for an enrollment that had an entry date on the earliest bed-night.  Specifically, you are looking for an enrollment that was literally homeless at entry (based on project type or 3.197).

Could this be simplified/clarified by filtering down enrollments to those starting on `first_date` and then for those `enrollment_literally_homeless_at_entry(enrollment)`?  

This is a total guess, but I'd say there are probably fewer enrollments than filtered_bed_nights, and they each only have one entry date.  That said, I'm sure I'm missing something.  Maybe the enrollment could just be overlapping the `first_date`...

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
```diff
@@ -107,12 +110,18 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
         )
       end
 
-      dates = filtered_bed_nights.map(&:last)
```
eanders, line 110: Do we know why this isn't being sorted as expected?  Looking at `filter_episode(calculated_bed_nights)`, the first thing it does is sort:
```
calculated_bed_nights = calculated_bed_nights.sort_by(&:last)
```

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
```diff
@@ -107,12 +110,18 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
...
+      first_date, last_date, days_homeless = compute_min_max_and_unique(filtered_bed_nights.map(&:last))
+
+      # filtered_bed_nights appears to be an array of [enrollment, index(?), date]
```
eanders: Comment from `candidate_bed_nights:
```
# return a triple [enrollment, id, date], using the index in place of the id so we don't need to load
# it from the DB (middle item needs to be unique within the enrollment)
```

Oddly, sometimes it returns `nil` as the middle item.  Not sure what that's about.
Also, looking at the code, I think you are correct, it is an index, not an ID.

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/measure_one.rb
eanders: Good catch.  I'm still unclear why ruby doesn't have median, mean methods for arrays.  This seems like something a low-level maths class or gem should have so people don't make stupid mistakes.

eanders (approved):
@ttoomey, this looks great.
Thank you for the many tests, and especially for the shared context.  I wonder if that could be made more generic eventually, or at least centralized.

I dropped a bunch of thoughts/questison on the episode file, but happy to punt any additional cleanup to round 2.

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
ttoomey: I assume the issue is on line 89 where we add to the sorted array but don't re-sort:
`filtered_bed_nights += pre_entry_dates`

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
ttoomey: Yes, I think your understanding is correct. The reason I changed it is that if there are multiple enrollments for that first date, it seems like each of those enrollments should be checked; not just the first enrollment (but this seems like an edge case). 

> Could this be simplified/clarified by filtering down enrollments to those starting on first_date and then for those enrollment_literally_homeless_at_entry(enrollment)?

The conditional execution based on date should accomplish that right? It should only evaluate enrollments that overlap first_date. Or is it just that the code could be clearer?

eanders: I think it was a combination of "performance" (or perceived performance) and clarity.  It is looping over all `filtered_bed_nights`, but really only cares about those that occur on the first date, right?  That seems like it might need to loop over a bunch of days (thousands for a single person sometimes) when it only cares about one of the first few.  Maybe `any?` will short circuit, but if it doesn't find a match, it may have to loop through the whole set.

Some additional clarity would be nice too.

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
eanders: Ah! missed that.  Thanks

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
ttoomey: > I think it was a combination of "performance" (or perceived performance) and clarity. It is looping over all filtered_bed_nights, but really only cares about those that occur on the first date, right? That seems like it might need to loop over a bunch of days (thousands for a single person sometimes) when it only cares about one of the first few. Maybe any? will short circuit, but if it doesn't find a match, it may have to loop through the whole set.

I see, yes ideally we would only iterate over the enrollments, not all the days, thanks!

I suggest that since we do need to iterate over all the days at least once, I'll add this logic into that loop to address the performance concern:
* renamed `compute_min_max_and_unique` to `compute_episode_statistics`
* adjusted `compute_episode_statistics` to also return `first_date_enrollments`
* adjusted `literally_homeless_at_entry` to only check  `first_date_enrollments`

However... reviewing the spec again I'm unsure as to why we only consider enrollments that overlap the first date. Shouldn't we just consider all enrollments with any bed nights?

eanders: Maybe we should look at the spec together tomorrow.  I'm guessing we only care about those that overlap the first date (really started on the first date) because it is looking for enrollments that are homeless at entry, which might influence the start of the episode.

ttoomey: That would be great, thanks

PR #5200, opened 2025-03-13
Move relevant state into the database

eanders:
## _Merging this PR_

## Description
* Moves `RELEVANT_COC_STATE` from ENV to `GrdaWarehouse::Config`

## Type of change
* Code clean-up

### DIFF
--- a/app/models/grda_warehouse/config.rb
+++ b/app/models/grda_warehouse/config.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module GrdaWarehouse
   class Config < GrdaWarehouseBase
     serialize :client_details, Array
@@ -17,7 +19,7 @@ module GrdaWarehouse
 
     def self.relevant_state_codes
       # NOTE: memoized class variable, only changes on restart/redeploy
-      @relevant_state_codes ||= ENV['RELEVANT_COC_STATE']&.split(',') || ['MA']
+      @relevant_state_codes ||= GrdaWarehouse::Config.get(:relevant_state_codes)&.split(',') || ['MA']
     end
 
     def self.available_cas_methods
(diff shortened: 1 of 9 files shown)


db/warehouse/migrate/20250313125655_add_relevant_state_to_site_config.rb
```diff
@@ -0,0 +1,8 @@
+
+# frozen_string_literal: true
+
+class AddRelevantStateToSiteConfig < ActiveRecord::Migration[7.0]
+  def change
+    add_column :configs, :relevant_state_codes, :string, default: ENV['RELEVANT_COC_STATE'] || 'MA', null: false
```
gigxz, line 6: Defaulting to MA for new installations feels a little weird, I thought we are trying to move away from MA-defaulted text and code? Is the issue that everything will break if there's no state?

app/models/grda_warehouse/config.rb
```diff
@@ -17,7 +19,7 @@ def self.cas_enabled?
 
     def self.relevant_state_codes
       # NOTE: memoized class variable, only changes on restart/redeploy
-      @relevant_state_codes ||= ENV['RELEVANT_COC_STATE']&.split(',') || ['MA']
+      @relevant_state_codes ||= GrdaWarehouse::Config.get(:relevant_state_codes)&.split(',') || ['MA']
```
gigxz, line 22: dont think ` || ['MA']` is necessary since the new column is non-nullable? (I realize this conflicts with my other question)

gigxz (approved):

db/warehouse/migrate/20250313125655_add_relevant_state_to_site_config.rb
eanders: That's exactly the issue.  There's an assumption that it is set.  It shouldn't actually default to MA unless we don't have `ENV['RELEVANT_COC_STATE']`, which should be set correctly prior to DB setup... Except, MA probably ends up in the structure file.  Yeah, not ideal, but things start to break if there's no state.  At least it's easily fixable now.

app/models/grda_warehouse/config.rb
eanders: That's probably true.  I was just going for minimal code changes.  I don't think it hurts, but I can remove if you think we should.

dtgreiner (approved):

PR #5204, opened 2025-03-14
Prevent returning CE conditional tasks

martha:
## _Merging this PR_
- squash and merge into the feature branch `tt-ce-spike-3`

## Description

Summary of changes:
- In the query for a Referral's Steps,
  - Add a filter so we don't return steps that might not happen
  - Add a `stop_when` condition so we don't return the children of such steps
- Fix n+1 issue with step form definitions
- Optimize the Graph's `walk` method to reduce unneeded db queries
- Add spec

To test:
- Using the ['Admin Approve Denial' workflow](https://github.com/greenriver/hmis-warehouse/blob/564fc77255ca7c68cf53ddf656711df69fd76ca7/drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake#L181-L187) in the starter pack, start a new referral
- You should not see the 'Review Denial' step until it is activated

## Type of change
- Feature update


drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
```diff
@@ -31,5 +31,9 @@ def workflow_node
     def swimlane
       load_ar_association(object, :swimlane)&.name
     end
+
+    def form_definition
+      load_ar_association(object.node, :form_definition)
```
gigxz: is there a reason not to use data loader to load association `object.node`? this is a little funny to me because it will be N+1 anyway when loading the node, I think.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -17,9 +17,16 @@ class HmisSchema::CeReferral < Types::BaseObject
...
+
+      graph = instance.template.graph(preloads: :inflows) # preload inflows so we can check conditions without n+1
+      graph.
+        # Stop the search when the node doesn't exist yet and is conditional. We don't want to return this node, or any of its children, if it won't definitely happen.
+        walk(stop_when: ->(node) { steps_by_node_id[node.id].nil? && node.inflows.map(&:condition).any? }).
+        filter(&:task?).
+        map do |node|
+        # If this node exists already, return it; otherwise, return a non-persisted version, ONLY IF it has no conditionals (will definitely happen).
+        steps_by_node_id[node.id] || (node.inflows.map(&:condition).any? ? nil : instance.steps.new(node: node).freeze)
+      end.compact
```
gigxz: Looks good to me!
Suggestion for readability improvement, since this is pretty complicated:

```ruby
        walk(stop_when: ->(node) { steps_by_node_id[node.id].nil? && node.has_conditional_inflows? }).
        filter(&:task?).
        map do |node|
            return steps_by_node_id[node.id] if steps_by_node_id[node.id] # task instance already exists
            
            return nil if node.has_conditional_inflows? # task is conditional, don't show it yet
            
            # initialize step to display in the UI
            instance.steps.new(node: node).freeze
      end.compact
```

drivers/hmis/app/models/hmis/workflow_definition/graph.rb
```diff
@@ -13,6 +15,8 @@ def initialize(nodes)
     # stop_when allows us to perform a bounded depth-first search:
     # matches = graph.walk(entrypoint_ids: ids, stop_when: ->(node) { node.type == 'task' })
     def walk(entrypoint_ids: nil, stop_when: nil)
+      nodes_by_id = nodes.index_by(&:id) # map nodes by ID so we can get them without re-querying the database
+
```
gigxz, line 19: nice improvement, thanks!

drivers/hmis/spec/requests/hmis/ce/ce_referral_spec.rb
```diff
@@ -98,13 +98,130 @@
...
+      end
+
+      before do
+        client_acceptance_task.outflows.destroy_all
+        client_acceptance_task.connect_to!(gateway)
+        gateway.connect_to!(provider_acceptance_task, condition: 'needs_provider_acceptance = 1')
+        gateway.connect_to!(accept_referral, condition: 'needs_provider_acceptance = 0')
+      end
+
+      it 'does not return the conditional step until it is enabled' do
```
gigxz: nit on test naming, since this case _doesn't_ test that the step  is returned when its enabled
```suggestion
      it 'does not return the conditional step' do
```

gigxz (approved):
looks good!

ttoomey (approved):
Looks good to me, thanks!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
martha: Hrm! I made this improvement, but there must still be something I'm not understanding either in the dataloader or how I'm setting up [this spec test](https://github.com/greenriver/hmis-warehouse/pull/5204/files#diff-2dd19f7adb4deb398e998a7ecfb0e2214ba3abf99c74b883af00cc77edbd8d89R201-R233). After running that test with both versions of the code, I didn't see a change in the # of db queries made.

PR #5207, opened 2025-03-15
7357 Performance Measurement Dashboard  issues

ttoomey:
## _Merging this PR_

## Description
* Fix issues with missing distinct when joining pm clients to client_projects
* Use correct cell references in spm_fields
* Fixes what seems like a misalignment where the spm_enrollment.destination is not 0 for stayers
* Adds tests to Performance Measurement Dashboard
* Add documentation (README)

## Type of change
Bug fix, Documentation


app/jobs/reporting/hud/run_report_job.rb
```diff
@@ -37,7 +39,7 @@ def perform(class_name, report_id, email: true)
         end
       end
 
-      puts "Running: #{@generator.class.name} Report ID: #{report_id}"
+      # puts "Running: #{@generator.class.name} Report ID: #{report_id}"
```
ttoomey, line 42: remove cruft from test output

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv.rb
```diff
@@ -87,7 +89,7 @@ def log_timing(message)
       phase = message.to_s.gsub(/!$/, '')
       importer_log&.log_phase(phase, started_at: Time.current)
 
-      memory_used = -> { `ps -o rss= -p #{Process.pid}`.to_i }
+      memory_used = -> { File.read("/proc/#{Process.pid}/status")[/VmRSS:\s+(\d+)/, 1].to_i }
```
ttoomey, line 92: seems like this invocation of ps doesn't work on apline busybox, leading to errors in the test output when importing fixtures

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -1104,7 +1108,7 @@ def spm_fields
           ],
         },
         {
-          cells: [['1a', 'B2']],
+          cells: [['1a', 'D2']],
```
ttoomey, line 1132: I believe these are incorrect. However it doesn't seem to matter in practice as cell is just used to get the universe from the SPM, and all all cells have the same universe for a given question

drivers/performance_measurement/app/models/performance_measurement/result_calculation.rb
```diff
@@ -119,12 +121,13 @@ def client_count_present(field, period, project_id: nil)
 
     def client_sum(field, period, project_id: nil)
       column = "#{period}_#{field}"
-      return clients.joins(:client_projects).merge(PerformanceMeasurement::ClientProject.where(period: period, for_question: field)).sum(column) if project_id.blank?
+      return clients.joins(:client_projects).merge(PerformanceMeasurement::ClientProject.where(period: period, for_question: field)).distinct.sum(column) if project_id.blank?
```
ttoomey, line 125: the missing distinct seems to be a cause of the misalignment

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+
+      # PM report success rate
+      pm_result = @pm_report.result_for(:es_sh_th_rrh_positive_destinations)
+      pm_percentage = pm_result.primary_value.to_f
+
+      # They should match
+      expect(pm_percentage).to be_within(0.1).of(spm_percentage)
+    end
+
+    xit 'has the same PH retention or exit rate in both reports' do
```
ttoomey: Getting 0 from the PM and 100 from the SPM, not sure why

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+
+      # PM report 6-month return rate
+      pm_result = @pm_report.result_for(:returned_in_six_months)
+      pm_percentage = pm_result.primary_value.to_f
+
+      # They should match
+      expect(pm_percentage.round).to eq(spm_percentage.round)
+    end
+
+    xit 'has the same 12-month return rate in both reports' do
```
ttoomey: getting 33 from SPM and 66 from PM

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -0,0 +1,311 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../../../../hud_spm_report/spec/models/fy2024/shared_context'
+
+RSpec.describe 'Performance Measurement and SPM Alignment', type: :model do
+  include_context 'SPM test setup'
+
+  # Shared context for running both reports with the same data
```
eanders:
```suggestion
```

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+      spm_report_filter.update(project_ids: project_ids)
+
+      # Step 1: Run the Performance Measurement report
+      @pm_report = PerformanceMeasurement::Report.new(
+        user_id: user.id,
+      )
+      @pm_report.filter = spm_report_filter
+      @pm_report.update_goal_configuration!
+      @pm_report.save!
+      @pm_report.update_goal_configuration!
```
eanders: This is probably some quirk of the report, but why do we need to call `update_goal_configuation!` twice?

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -0,0 +1,311 @@
...
+    it 'has the same permanent housing exit rate in both reports' do
+      # SPM report successful placement percentage for ES/TH/SH/RRH exits
+      spm_percentage = @spm_report.answer(question: '7b.1', cell: 'C4').summary.to_f
+
+      # PM report success rate
+      pm_result = @pm_report.result_for(:es_sh_th_rrh_positive_destinations)
+      pm_percentage = pm_result.primary_value.to_f
+
+      # They should match
+      expect(pm_percentage).to be_within(0.1).of(spm_percentage)
```
eanders: Should we also confirm it is non-zero?

eanders (commented):
Thanks @ttoomey !
I'm glad the bug was actually relatively small.  Sorry for all of the discovery required to find it.

Thanks for the extra documentation and new tests!

I think it's worth tracking down the test failures before we merged, but looks good to me otherwise.

drivers/performance_measurement/app/models/performance_measurement/result_calculation.rb
```diff
@@ -77,10 +79,11 @@ def percent_of(numerator, denominator)
       ((numerator / denominator.to_f) * 100).round
     end
 
+    # it appears we only store integers so round the average
     def average(value, count)
       return 0 unless count.positive?
 
-      value.to_f / count
+      (value.to_f / count).round
```
ttoomey, line 86: previously we were writing floats into an int db field, resulting in truncation.

drivers/performance_measurement/app/models/performance_measurement/details.rb
```diff
@@ -269,7 +271,25 @@ def show_row?(key)
...
+    # - category/sub_category: Classification ('Rare', 'Brief', 'Non-Recurring') and subgroup
+    # - column: Where metric is displayed (:system, :project, or :both)
+    # - year_over_year_change: Whether to display year-over-year comparison
+    # - title: Human-readable title
+    # - goal_description: Description with %{goal} placeholder
+    # - goal_direction: Direction of improvement ('>', '<', '+', or '-')
+    # - goal_calculation: Method reference for goal calculation
+    # - calculation_description: How the metric is calculated
+    # - calculation_column: Client data column(s) for calculation
+    # - detail_columns: Client-level columns for drill-down views
```
ttoomey, line 288: seems like these are for drill downs as opposed to the client_project

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -933,7 +937,7 @@ def can_see_client_details?(user)
     end
 
     PERMANENT_DESTINATIONS = HudSpmReport::Generators::Fy2023::MeasureSeven::PERMANENT_DESTINATIONS
-    PERMANENT_DESTINATIONS_OR_STAYER = (PERMANENT_DESTINATIONS + [0]).freeze
+    # PERMANENT_DESTINATIONS_OR_STAYER = (PERMANENT_DESTINATIONS + [0]).freeze
```
ttoomey: it sure seems like destination is not set to 0 for stayers, I've tried to adapt the spm_fields config to account for what I assume is a change in behavior in the spm implementation

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -1196,21 +1227,27 @@ def spm_fields
...
             {
-              name: :moved_in_destination, # NOTE: destination 0 == stayer in the SPM
-              value_calculation: destination_calculation,
+              name: :moved_in_destination,
+              value_calculation: default_calculation,
             },
           ],
           client_project_rows: [
             ->(spm_enrollment) {
-              return unless destination_calculation.call(spm_enrollment)
```
ttoomey, line 1205: I think 7b.2 universe is only enrollments with a move-in, so this check isn't needed

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
ttoomey: removed it, thanks

drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
ttoomey: I improved the tests to add an expected value so we don't rely just on spm value

drivers/performance_measurement/app/models/performance_measurement/details.rb
eanders: I believe drill-downs are only shown for client_project (project-level) data.  So, yes, but in relation to `client_project`)

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -957,7 +961,7 @@ def can_see_client_details?(user)
           value_calculation: ->(client, variant_name) {
             return true if PERMANENT_DESTINATIONS.include?(client.send("#{variant_name}_so_destination"))
             return true if PERMANENT_DESTINATIONS.include?(client.send("#{variant_name}_es_sh_th_rrh_destination"))
-            return true if PERMANENT_DESTINATIONS_OR_STAYER.include?(client.send("#{variant_name}_moved_in_destination"))
+            return true if client.send("#{variant_name}_stayer")
```
eanders: `_stayer` is not the same as someone who was a stayer or exited to a permanent destination.  I'm not sure this change is quite right.

drivers/performance_measurement/app/models/performance_measurement/report.rb
eanders: We have a comment `name: :moved_in_destination, # NOTE: destination 0 == stayer in the SPM` on `cells: [['7b.2', 'C2']],`.  Before we remove this I'd like to confirm this is no longer true.  
Looking at some data in production I'm not seeing any SPM Enrollments with `destination = 0` so probably this change should be:
```
PERMANENT_DESTINATIONS_OR_STAYER = (PERMANENT_DESTINATIONS + [nil]).freeze
```

drivers/performance_measurement/app/models/performance_measurement/report.rb
eanders: I think `m7b_history` includes data for both 7b1 and 7b2, which might mean we need to limit to those with destinations.

drivers/performance_measurement/app/models/performance_measurement/report.rb
ttoomey: yeah, stayer doesn't seem to be populated in the current config, I hijacked it. I'll back that out

drivers/performance_measurement/app/models/performance_measurement/report.rb
ttoomey: seems reasonable, I'll try it, thanks. I looked around to see if we had been setting this to 0 in older SPMs but couldn't find any clues.

drivers/performance_measurement/app/models/performance_measurement/report.rb
ttoomey: I'm not quite sure I follow, my mental model is that there's an implied filter on the lamba so that it only runs for enrollments who would fall into the universe defined in cell (7b.1 C2). However I'll add logic to check for stayers here in addition to destination

eanders (approved):
Let's see if this fixes things.  Thanks for all the tests @ttoomey and for your thoughtful approach.

ttoomey:
Thanks, merge to 155, yes? We don't have much time left for QA on 155 but we could get this out in a hotfix if needed

eanders:
@ttoomey I think 155 is fine.
@agougherty do you have time to QA tomorrow or Wednesday if we get it deployed later today?

agougherty:
@eanders @ttoomey yes, I'll have time tomorrow!

PR #5210, opened 2025-03-17
Release opportunity when workflow is rejected

martha:
## _Merging this PR_
- squash-merge into the feature branch tt-ce-spike-3

## Description
Addresses comment on frontend PR: https://github.com/greenriver/hmis-frontend/pull/1096#discussion_r1995827168

~Although this change is related to the earlier GH issue https://github.com/open-path/Green-River/issues/7309, I opened this PR against the validations branch, since I had made changes in the mutation there already, and added a spec file for this mutation there too. Sorry for the tangle!~

How to test: 
- In the frontend, deny a referral. You should see the status of the opportunity return to "open."

Still to do:
- The "return to opportunity" button in the wayfinding step doesn't reload the opportunity so that the new status is visible. I'd like to address that in https://github.com/open-path/Green-River/issues/7322, the opportunity page ticket.

## Type of change
Bug fix on feature branch (unreleased bug)

- [na] If adding a new endpoint / exposing data in a new way, I have:
  - [na] ensured the API can't leak data from other data sources
  - [na] ensured this does not introduce N+1s
  - [na] ensured permissions and visibility checks are performed in the right places


drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
```diff
@@ -63,6 +63,8 @@ def resolve(referral_id:, step_id:, input:, confirmed: false)
         engine.complete_step!(step, user: current_user, submitted_values: input)
       end
 
+      referral.opportunity.release! if referral.status.to_sym == :rejected
```
gigxz: suggest using or adding a helper so we can call `referral.rejected?` (not sure if these are magically present or need to be added)

gigxz: More importantly, though, I wonder if this logic should live in the engine rather than in the mutation. 
For example another place this release could go is [here](https://github.com/greenriver/hmis-warehouse/blob/9f7e1a5131e9a23e447bb557141eb3742f16c573/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb#L18-L19) so it occurs as part of the EndEvent reject_referral. Lets discuss. I  _think_ its preferable to keep logic in the engine or message handler, and not have logic peppered into the mutations

gigxz: oh yeah looking at [these implementations](https://github.com/greenriver/hmis-warehouse/blob/9f7e1a5131e9a23e447bb557141eb3742f16c573/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb#L35-L420) @martha it definitely makes sense to me for the opportunity release to go here too.

gigxz (approved):

ttoomey (approved):
Thanks!

PR #5214, opened 2025-03-18
include coc_codes in collection edit form

dtgreiner:
## _Merging this PR_

## Description

The coc_codes column on collections was being overwritten with an empty set on the collection update page.

## Type of change

Bug fix


app/views/admin/collections/_form.haml
```diff
@@ -11,6 +11,7 @@
   .col-6
     = f.input :name, required: true, disabled: @collection.entities_locked?
     = f.input :description, disabled: @collection.entities_locked?
+    = f.hidden_field :coc_codes, multiple: true, value: @collection.coc_codes
```
eanders: Hmm, what do you think about just excluding it from the permitted params on the update method instead?
I'm concerned this potentially opens up a way of adding CoC Codes that shouldn't be there.

dtgreiner: That should work too, there was some logic specifically setting the coc_codes in the permitted params which made me think it was being utilized elsewhere, but after a deeper look, I think that was leftover logic from a different change. I'll do a little more digging to be sure, but likely will be switching it over to making the change in the params.

dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/5214/commits/79f09cdcb2086735fdf968718d2732d6bbac39c1)

app/controllers/admin/collections_controller.rb
```diff
@@ -110,10 +112,7 @@ def bulk_entities
         :name,
         :description,
         :collection_type,
-        coc_codes: [],
-      ).tap do |result|
-        result[:coc_codes] ||= []
-      end
+      )
```
eanders, line 126: I'm pretty sure for updating the viewables of the collection, we'll need to include the CoC codes.  `viewable_params` doesn't include CoCs, so I think we'll need to fix that, unless that's in a different controller?

Might be worth adding some tests so we can confirm the various save methods are working as expected.

dtgreiner: It looks like this is only coming into play for legacy collections but does need updating. I'll get an update together to allow the legacy and current collection updates to both work and add some tests for creating/updating collections and collection viewables.

dtgreiner:
@eanders whenever you have the chance, this has been updated with new tests and should be working for legacy and new collections

spec/requests/admin/collections_controller_spec.rb
```diff
@@ -0,0 +1,411 @@
...
+        collection: legacy_project_collection,
+        coc_codes: ['XX-001'],
+        data_sources: [data_source.id],
+        organizations: [],
+        projects: [],
+        project_groups: [],
+        project_access_groups: [],
+        reports: [],
+        cohorts: [],
+      )
```
eanders: I feel like there's probably some serious cleanup that could happen with a loop here, but it looks good for now.  Thanks!

eanders (approved):

spec/requests/admin/collections_controller_spec.rb
dtgreiner: just added some test clean up [here](https://github.com/greenriver/hmis-warehouse/pull/5214/commits/5932beee9f5daa0fde8d37cb5d0bcc8ca679a45a)

eanders (approved):

PR #5216, opened 2025-03-19
7424 replace aasm

ttoomey:
## _Merging this PR_

## Description
Discovered aasm no longer seems to be maintained (last commit is > 2 years ago). Replace with api-compatible module. 

## Type of change
Dependency update


app/models/concerns/simple_state_machine.rb
```diff
@@ -0,0 +1,121 @@
+# frozen_string_literal: true
+
+# SimpleStateMachine module provides a lightweight alternative to the AASM gem
+module SimpleStateMachine
+  # Extends the including class with class methods
+  def self.included(base)
+    base.extend(ClassMethods)
+    base.class_attribute :aasm_column, :aasm_states, :aasm_events, :aasm_initial_state, instance_accessor: false
```
eanders: Given that this is now SSM, maybe we should go ahead and rename these methods?
```suggestion
    base.class_attribute :ssm_column, :ssm_states, :ssm_events, :ssm_initial_state, instance_accessor: false
```

eanders (approved):
@ttoomey, this looks like a great path forward.  (I do wonder if we should develop it as a gem, I know other GR projects would use it.)

My only suggestion would be to embrace the SimpleStateMachine name and force a change away from `aasm_` prefixed methods and naming.

eanders (approved):
Thanks!

gigxz (approved):
looks great!

PR #5218, opened 2025-03-19
Update schema and starter pack for Opportunity Page

martha:
## _Merging this PR_
- Squash-merge into the feature branch tt-ce-spike-3

## Description
GH issue: https://github.com/open-path/Green-River/issues/7322

Summary of changes:
- Schema changes
  - Add `MatchRule` type
  - Add `rules` to Opportunity
  - Add `topCandidate` to the Opportunity, which is loaded as part of the summary
  - Make candidate's `client` field nullable, if user doesn't have viewing permission on this client. Add a non-nullable `clientId` field for identifying clients anonymously.
- Add some example Match Rules to the starter pack, and sketch out support for them in the in-code config. This is incomplete, I just wanted to get something that will demo the Opportunity page's basic functionality
- Add a `for_opportunity` class method on `Hmis::Ce::Match::Rule`
- Fix some misc bugs in Match Rules applicability logic, and add spec tests

How to test:
- I'll add instructions on the frontend PR

## Type of change
Feature build-out

  - [ ] ensured this does not introduce N+1s - to discuss
  - [x] ensured permissions and visibility checks are performed in the right places - I think so, Match Rules are not sensitive and they are only resolved on Opportunities that the user already has perms on


drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
```diff
@@ -0,0 +1,31 @@
...
+
+module Types
+  class HmisSchema::CeMatchRule < Types::BaseObject
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :type, HmisSchema::Enums::CeMatchRuleType, null: false, method: :rule_type
+    field :owner_type, String, null: false
+
+    def owner_type # todo @martha - discuss
+      # config = object.applicability_config.symbolize_keys
```
martha: I got a bit stuck here and would like to discuss. Match rule owner is polymorphic, it can be an opportunity, project, or organization. Match rule applicability is also configured in the `applicability_config` object which indicates whether it applies to a given funder or project type.

I had a half baked idea that maybe this new schema field should be a flexible `owner_string` and can return something like:
- "Applies to this opportunity"
- "Inherited from [project | organization]"
- "Applies to this project's type"
- "Applies to this project's funder(s)"

etc. But this seems a little fishy somehow, we don't often return whole strings that the UI should display from the backend.

drivers/hmis/app/models/hmis/ce/match/rule.rb
```diff
@@ -22,13 +22,31 @@ def priority_scheme?
...
       )
       applicability.call(opportunity)
     end
+
+    def self.for_opportunity(opportunity)
+      all_rules = Hmis::Ce::Match::Rule.order(:owner_type, :id).to_a
+      all_rules.filter { |rule| rule.applies_to_opportunity?(opportunity) }
+    end
+
+    # todo @martha - discuss, this makes many fewer db queries, but we probably don't want to introduce the complexity?
```
martha: I'd like to discuss this part further! chatGPT and I played around with this until we came up with a scope that queries all the rules for a given opportunity in very few db queries. Gig and I discussed at a 1:1 earlier this week, and more or less agreed that it wasn't worth the complexity, since we don't expect millions of match rule records. (And the scope is a bit ugly with the inlined sql code). 

I've included it here commented out, just to show the difference in this [test](https://github.com/greenriver/hmis-warehouse/pull/5218/files#diff-e70d7e88dd404e9d0ba5186bd63fdfca6e0ef47850122789f91d8150f4174d89R113-R143) (that I can also remove/modify before merge) and confirm that we want to keep the simpler code, but less performant version.

martha:
I'm converting this back to Draft for now while working on the correct behavior for anonymized clients (per https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP?node-id=2697-72703#1182177067)!

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -9,8 +9,11 @@
 module Types
   class HmisSchema::CeCandidate < Types::BaseObject
     field :id, ID, null: false
-    # TODO - do we want to expose client here? Maybe we can show candidates without revealing PII
-    field :client, HmisSchema::Client, null: false
+    field :client, HmisSchema::Client, null: true
```
gigxz: A few suggestions here just for future understanding
- can you add a comment stating that object is `Hmis::Ce::Match`?
- maybe `description` on the client field explaining why it can be null

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
gigxz: I think your proposed approach here is good for now! I don't have anything against returning a user-facing description from the API, but you're right we don't do it much. I think if we end up doing more of a design deepdive on how to display requirement lineage, maybe we'll revisit this.

The only tweak I would put on your implementation is to return "Funder" or "Project Type" if that's where the rule is coming from, something like:
```ruby
      if object.owner_type
        object.owner_type.demodulize # May be Project, Organization, etc
      elsif object.applicability_config[:project_types]&.any?
        'Project Type'
      elsif object.applicability_config[:project_funders]&.any?
        'Funder'
      end
```

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
```diff
@@ -0,0 +1,31 @@
...
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeMatchRule < Types::BaseObject
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :type, HmisSchema::Enums::CeMatchRuleType, null: false, method: :rule_type
+    field :owner_type, String, null: false
+
```
gigxz, line 15: Can you add a comment here like `# object is a Hmis::Ce::Match::Rule`. It's maybe obvious from the type naming, but I always find it useful when we have a comment confirming what the underlying record is

drivers/hmis/app/graphql/mutations/ce/create_ce_referral.rb
```diff
@@ -17,7 +17,7 @@ def resolve(opportunity_id:, client_id:, input:)
       raise unless Hmis::Ce.configuration.enabled?
 
       opportunity = Hmis::Ce::Opportunity.viewable_by(current_user).find(opportunity_id)
-      client = Hmis::Hud::Client.viewable_by(current_user).find(client_id)
+      client = Hmis::Hud::Client.find(client_id) # Doesn't need to be viewable by the current user
```
gigxz, line 20: We should confirm data source against `current_user.hmis_data_Source_id`. Otherwise the API would support referring a client from a different data source, which could get weird.

martha:
I'm re-marking this as Ready for Review. Thanks!

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -10,20 +10,26 @@ module Types
...
-    field :status, String, null: false
+    field :status, HmisSchema::Enums::CeOpportunityStatus, null: false
     field :expires_at, GraphQL::Types::ISO8601DateTime, null: true
     field :active_referral, Types::HmisSchema::CeReferral, null: true
     field :accepted_referral, Types::HmisSchema::CeReferral, null: true
     field :candidates, Types::HmisSchema::CeCandidate.page_type, null: false
+    field :top_candidate, Types::HmisSchema::CeCandidate, null: true
     field :project_id, ID, null: false
     field :project_name, String, null: false
+    field :rules, [HmisSchema::CeMatchRule], null: true
```
gigxz: I wonder if we will ever want them split out the rule types in the API. For example if there is different supplemental information that we want to resolve on eligibility requirements, then it might make sense for eligibility requirements to have their own GQL type with that shape.

```graphql
opportunity {
    eligibilityRequirements { .... }
    prioritizationSchemes { ... }
}
```

Actually, now that I dig into it a bit more, it looks like there should only be 1 prioritization scheme per opportunity. The candidate pool builder appears to only use [one expression](https://github.com/greenriver/hmis-warehouse/blob/349829daabf47dcf25b464ae7b92b94ffdbad400/drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb#L31) to calculate priority score. (Which makes sense because there is nothing in place to rank/prioritize different prioritization schemes against each other).

Given that, a more appropriate API might be:

```graphql
opportunity {
    eligibilityRequirements { .... }
    prioritizationScheme { ... }
}
```

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -10,20 +10,26 @@ module Types
...
     def candidates
       Hmis::Ce::Match::Candidate.
         for_opportunity(object).
         order(priority_score: :desc, client_id: :desc)
     end
 
+    def top_candidate
+      Hmis::Ce::Match::Candidate.for_opportunity(object).order(priority_score: :desc, client_id: :desc).first
+    end
+
```
gigxz: This feels maybe unnecessary, the frontend could query the same thing with `candidates(limit: 1) { id }`. That might be a more maintainable approach so we don't need to worry about maintaining the logic for both the `candidates` and `top_candidate` method to match exactly.

drivers/hmis/app/models/hmis/ce/match/rule.rb
gigxz: @martha I was poking at this a while to try to understand _why_ the existing method made SO many queries. I noticed that in the test you added, it is making a ton of repeated DB queries to look up the same project, organization, etc.

I finally realized the N+1 was coming from the `rule.owner` lookup that occurs on line 27. Adding a preload `Hmis::Ce::Match::Rule.preload(:owner)` on line 35 brings the query count down to like 8 from 100+. 

So my suggestion would be making that change, and continuing to perform the applicability evaluation in-memory (aka removing the commented-out scope). I am recommending this because I think it is preferable to have the logic in one place (OpportunityApplicability) rather than duplicate across a scope and a class, and generally the class logic is easier to understand.

---

In line with that, there could be some perf improvements added to the candidate pool builder [here](https://github.com/greenriver/hmis-warehouse/blob/349829daabf47dcf25b464ae7b92b94ffdbad400/drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb#L26-L28), I think something like:
```ruby
      all_rules = Hmis::Ce::Match::Rule.preload(:owner).order(:owner_type, :id).to_a
      # group opportunities by unique priority and eligibility rules
      active_opportunities.preload(project: [:organization, :funders]).each do |opportunity|
```
First preload the `owner` so we don't have the same n+1 looking up the owner record, then preload project/organization/funder so we don't have an N+1 on that either. If you can try that and throw it in too, that would be great!

drivers/hmis/app/models/hmis/ce/match/opportunity_applicability.rb
```diff
@@ -26,11 +26,11 @@ def matches_owner?(opportunity)
     end
 
     def matches_project_type?(project)
-      project_types.blank? || project.project_type.in(project_types)
+      project_types.blank? || project_types.include?(project.project_type)
     end
 
     def matches_project_funders?(project)
-      project_funders.blank? || project_funder_ids & project.project_funders.map(&:funder_id)
+      project_funders.blank? || (project_funders & project.funders.map(&:id)).any?
```
gigxz: `project.funders.map(&:id)` should actually by `project.funders.map(&:funder)` since this should be the HUD Funder ID, not the ID of the funder record itself.

gigxz (commented):
looks great!

drivers/hmis/app/models/hmis/ce/match/rule.rb
martha: Thank you, amazing!

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
martha: Thanks! yeah this sounds good to me too, I was side-eying [this from the frontend PR](https://github.com/greenriver/hmis-frontend/pull/1101/commits/2b0dc3365b0e8cb54339971fe9e5ba9bfdcf5606#diff-c3d9400c977b85f261b2a19c77a5e214b36f0a2f511be3a86b85371c73d2167bR35-R40) as a bit smelly too. Now that's no longer needed.

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
```diff
@@ -0,0 +1,29 @@
...
+module Types
+  class HmisSchema::CeMatchRule < Types::BaseObject
+    # object is a Hmis::Ce::Match::Rule
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :type, HmisSchema::Enums::CeMatchRuleType, null: false, method: :rule_type
+    field :owner_type, String, null: false
+
+    def owner_type
```
ttoomey, line 16: I believe my original intent was that a given rule might be owned by a polymorphic owner, where owner is the entity under which the rule is managed (coc, project). This would determine who could manage the rule and where it might appear in the UI. 

But a rule might apply based on multiple attributes related to the opportunity (project_types, funders, etc). This applicability is different from ownership.

So if the intention is to show the owner type, this resolver isn't quite right, it would instead return object.owner_type.

If the intention is to describe the types of opportunities this rule would target, this isn't quite right either, it wouldn't be project_type or funder but it would need to describe all the non-empty keys of applicability_config.

drivers/hmis/app/models/hmis/ce/match/rule.rb
```diff
@@ -17,18 +17,31 @@ def eligibility_requirement?
...
       applicability = OpportunityApplicability.new(
         owner: owner,
         project_types: config[:project_types],
         project_funders: config[:project_funders],
       )
       applicability.call(opportunity)
     end
+
+    def self.for_opportunity(opportunity)
+      all_rules = Hmis::Ce::Match::Rule.preload(:owner).order(:owner_type, :id).to_a
```
ttoomey: adjust to allow class method to apply to a scope. Otherwise the behavior might be non-intuitive (should support invocation like `Hmis::Ce::Match::Rule.where(owner: owner).for_opportunity(opportunity)`)
```suggestion
      all_rules = preload(:owner).order(:owner_type, :id).to_a
```

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
gigxz: I'm not really seeing the need for distinguishing between "ownership" and "applicability". This probably needs some more discussion and requirement-gathering. I would suggest leaving this as-is for now, and revisit when we get into the "requirements management" chunk of work.

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
```diff
@@ -0,0 +1,29 @@
...
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeMatchRule < Types::BaseObject
+    # object is a Hmis::Ce::Match::Rule
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :type, HmisSchema::Enums::CeMatchRuleType, null: false, method: :rule_type
```
gigxz: Is this still needed?

drivers/hmis/app/graphql/types/hmis_schema/enums/ce_opportunity_status.rb
```diff
@@ -0,0 +1,18 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+#  Copyright 2016 - 2024 Green River Data Analysis, LLC
```
gigxz: double copyright comment on this page

drivers/hmis/app/models/hmis/ce/match/rule.rb
gigxz: I think the convention would also be to make this a `scope`

ttoomey: My general understanding is that if you're returning an ActiveRecord::Relation, then a scope is most correct.

If you're returning something else that can't be chained (such as an array in this case), a class method is the more appropriate choice.

Of course either will work, I don't think there's much of a functional difference; it's more about conveying intent.

So if changing this to a scope, I'd suggest returning a scope of Opportunity.where(id:...) rather than an array

[rails docs](https://api.rubyonrails.org/classes/ActiveRecord/Scoping/Named/ClassMethods.html#method-i-scope)
> scope() 
Adds a class method for retrieving and querying objects. The method is intended to return an [ActiveRecord::Relation](https://api.rubyonrails.org/classes/ActiveRecord/Relation.html) object, which is composable with other scopes

gigxz: oh I see, thanks. seems like it can remain a class method then.

drivers/hmis/app/graphql/types/hmis_schema/enums/ce_opportunity_status.rb
martha: Thanks!

drivers/hmis/app/graphql/types/hmis_schema/ce_match_rule.rb
martha: No, I'll remove it. Thanks!

gigxz (approved):

PR #5220, opened 2025-03-20
webconsole-for-qa

ttoomey:
## _Merging this PR_

## Description
Adds an admin web console for QA (admin/console). There is a web-console gem but I don't get the sense it's intended for production

Why? I can't seem to get a console into QA for debugging, this is my questionable work-around

## Type of change
New feature

### DIFF
--- a/app/controllers/admin/console_controller.rb
+++ b/app/controllers/admin/console_controller.rb
@@ -1,0 +1,60 @@
+# frozen_string_literal: true
+
+if Rails.env.staging? || Rails.env.development?
+  require 'irb'
+  require 'irb/completion'
+
+  class Admin::ConsoleController < ApplicationController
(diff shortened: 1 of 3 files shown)


ttoomey:
@eanders does this seem like a bad idea? My thought is I could just deploy this to QA to get access temporarily. Maybe we don't merge it though

eanders (changes requested):
@ttoomey I don't think we should do this.  We already have mechanisms to access Rails Consoles in deployed environments.  If we need to get into the web process, then maybe we need something like this, but if we just need access to the environment, we should be able to get you up and running with our existing infrastructure.

ttoomey:
Thanks, I was hoping for a quick & temporary way to work around my access problem on QA; but you're right, it's not a good idea

PR #5227, opened 2025-03-21
Give HMIS HUD models "strict" int and decimal attributes

martha:
## _Merging this PR_

## Description

By default in Rails/Postgres, when saving a non-numerical string like "bad string" onto an int or decimal database field, the string gets cast to 0 instead of raising an exception. We want to override that behavior for all HUD models. (See more details on the linked ticket.) This PR defines new types `StrictInteger` and `StrictDecimal`, and applies them to HUD models in a new concern `WithStrictAttributes`.

## Type of change
Bug protection


drivers/hmis/app/models/hmis/hud/income_benefit.rb
```diff
@@ -18,4 +20,5 @@ class Hmis::Hud::IncomeBenefit < Hmis::Hud::Base
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
 
   validates_with Hmis::Hud::Validators::IncomeBenefitValidator
+  validates :unemployment_amount, numericality: { only_integer: true }
```
martha: Use https://api.rubyonrails.org/classes/ActiveRecord/Attributes/ClassMethods.html#method-i-attribute instead

drivers/hmis/app/models/hmis/hud/base.rb
```diff
@@ -25,6 +27,29 @@ def self.without_optimistic_locking
 
   before_validation :ensure_id
 
+  # def self.inherited(subclass)
```
martha: Approach 1: In the base Hmis::Hud abstract class, add a callback `self.inherited`
Why it doesn't work: Rails hasn't loaded the db schema yet at the point when the models inherit from the base class, so the `table_exists?` is always false.

drivers/hmis/app/models/hmis/hud/concerns/with_strict_attributes.rb
```diff
@@ -0,0 +1,32 @@
...
+    if table_exists?
+      columns.each do |column|
+        case column.type
+        when :integer
+          attribute column.name, Hmis::StrictInteger.new
+        when :decimal
+          attribute column.name, Hmis::StrictDecimal.new
+        end
+      end
+    else # rubocop:disable Style/EmptyElse
```
martha: Approach 2: In a concern, which we add to the `Shared` concern.
Pro of this approach: It's probably the most clear, because it makes more intuitive sense for these attributes to be defined on the models (aka in a concern that is applied to the models) rather than in some hook somewhere else.
Why it doesn't work (perfectly): Apparently _some_ models have the table loaded at this point, but not all.

drivers/hmis/config/initializers/strict_attributes.rb
```diff
@@ -0,0 +1,37 @@
...
+# frozen_string_literal: true
+
+# discussed with Gig and agreed that it would be more clear to put this in the models themselves (aka in a concern)
+# rather than here in a hook. However, I'm leaving this in the PR for discussion since I need to better understand rails model loading.
+# In this strategy, as long as we call Rails.application.eager_load!, all the models' tables are loaded.
+
+# # if changing this code and then testing it using rspec, need to DISABLE_SPRING=1
+# Rails.application.config.after_initialize do
+#   # without eager_load, Hmis::Hud::Base.descendants == [].
+#   Rails.application.eager_load!
```
martha: Approach 3: In an `after_initialize` hook.
Cons of this approach: 
- It's less clear because it defines model code somewhere far-ish away from the models themselves.
- I don't really know about the ramifications of calling `Rails.application.eager_load!` here. Will that have bad performance effects on application load?

Pro of this approach:
- Using eager loading, it's the only approach I've managed to find so far that actually defines the attributes on all of the models.

drivers/hmis/spec/requests/hmis/assessments/submit_hud_assessments_spec.rb
```diff
@@ -436,6 +438,29 @@ def expect_assessment_dates(assessment, expected_assessment_date:, expected_entr
     errors = result.dig('data', 'submitAssessment', 'errors')
     expect(errors).to contain_exactly(a_hash_including(expected_error))
   end
+
+  it 'prevents saving IncomeBenefit with invalid strings' do
```
martha, line 442: (Just noting that this test isn't closely related to the other changes in this PR; it's confirming existing behavior)

ttoomey (commented):
Thanks for investigating this! I hadn't considered the load order issue, that's a good finding.

I wonder if just hooking into the active record lifecycle would be a bit simpler and sufficient?

```ruby

module StrictNumericValidation
  extend ActiveSupport::Concern

  included do
    before_validation :enforce_strict_numeric_fields!
  end

  private

  INT_RGX = /\A-?\d+(\.\d*[1-9])?\z|\A-?\d+\.0+\z/  # Allows "123.0"
  DEC_RGX = /\A-?\d+(\.\d+)?\z/
  STRICT_TYPE_PATTERNS = {
    big_integer: INT_RGX,
    decimal: DEC_RGX,
    float: DEC_RGX,
    integer: INT_RGX,
  }.freeze

  def enforce_strict_numeric_fields!
    self.class.columns.each do |column|
      pattern = STRICT_TYPE_PATTERNS[column.type]
      next unless pattern
      # next unless attribute_changed?(column.name)

      value = self[column.name]
      next unless value.is_a?(String) && value.present?

      raise ArgumentError, "Invalid value for #{column.name}: \"#{value.inspect}\"" unless value =~ pattern
    end
  end
end
```

drivers/hmis/app/models/hmis/hud/concerns/with_strict_attributes.rb
gigxz: @martha I put a `binding.pry` here, and this is what I found:

```
[1] pry(Hmis::Hud::Project)> self
=> Hmis::Hud::Project(Table doesn't exist)
[2] pry(Hmis::Hud::Project)> table_name
=> "hmis_projects"
```

Which pointed me to the fact that the concern is being included BEFORE the table_name is specified on the model. Moving [these 2 lines](https://github.com/greenriver/hmis-warehouse/blob/5924157d0de8533f663d35ad912e0c4ef4de621b/drivers/hmis/app/models/hmis/hud/project.rb#L15-L16) up to the top of the model fixes the issue, at least for Project. It looks like the same issue applies to  the rest of the models listed below

martha: Wow 😮 cool sleuthing, thank you Gig!

drivers/hmis/app/models/hmis/strict_decimal.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Hmis::StrictDecimal < ActiveModel::Type::Decimal
+  def cast(value)
+    return if value.nil?
+
+    raise ArgumentError, "Invalid decimal value: #{value.inspect}" if value.is_a?(String) && value !~ /\A[+-]?\d+(\.\d+)?\z/
```
gigxz: There are a couple cases where this became too strict I think:

old behavior: `"500."`=>`500.0`
new behavior: `"500."`=> raises error

old behavior: `".5"`=>`0.5`
new behavior: `"0.5"`=> raises error

drivers/hmis/app/models/hmis/strict_decimal.rb
```diff
@@ -0,0 +1,11 @@
...
+
+    super
```
gigxz, line 11: the default behavior scast non-String values as well:
```
irb(main):026> value = {"foo" => "bar" }; ib.TotalMonthlyIncome = value; ib.save!; ib.TotalMonthlyIncome.to_s
=> "0.0"
```

I don't think the code here will address that since the value is not a string?

Maybe we should adjust it to 
```
raise if value.to_s !~ DECIMAL_REGEX
```

drivers/hmis/app/models/hmis/strict_integer.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class Hmis::StrictInteger < ActiveModel::Type::Integer
+  def cast(value)
+    return if value.nil?
+
+    raise ArgumentError, "Invalid integer value: #{value.inspect}" if value.is_a?(String) && value !~ /\A[+-]?\d+\z/
```
gigxz: Same question here regarding invalid non-string values

drivers/hmis/app/models/hmis/hud/concerns/with_strict_attributes.rb
```diff
@@ -0,0 +1,24 @@
...
+  included do
+    raise "#{self} cannot include Hmis::Hud::Concerns::WithStrictAttributes because the table #{table_name} does not exist. Try moving `self.table_name` assignment to the top of the class." unless table_exists?
+
+    columns.each do |column|
+      case column.type
+      when :integer
+        attribute column.name, Hmis::StrictInteger.new
+      when :decimal
+        attribute column.name, Hmis::StrictDecimal.new
+      end
```
gigxz, line 25: We should handle big ints and floats here too per Therons comment. It looks like CustomService FAAmount is a float for some reason (not ideal that thats different from Service.FAAmount, dont think it was intentional)

I think the possible values for type are the values from `NATIVE_DATABASE_TYPES ` here https://api.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/PostgreSQLAdapter.html

drivers/hmis/app/models/hmis/strict_decimal.rb
martha: Sounds good, I think I will propose something like `raise unless (value.is_a? Float) || DEC_RGX.match?(value.to_s)` so we don't have stringify-then-match if the value doesn't need to be cast at all.

drivers/hmis/app/models/hmis/hud/concerns/with_strict_attributes.rb
martha: Interesting, it looks like there is a `money` type supported as well, but I think we don't use it?

gigxz: Not that I know of

drivers/hmis/app/models/hmis/strict_decimal.rb
martha: Thanks good point. While testing I also noticed it became more strict about booleans. I updated it to allow casting `false` to 0 and `true` to 1 for int cols.

drivers/hmis/app/models/hmis/strict_integer.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class Hmis::StrictInteger < ActiveModel::Type::Integer
+  INT_RGX = /\A-?\d+(\.0+)?\z/
```
martha, line 4: This is a different regex than you had proposed @ttoomey:
- this one `/\A-?\d+(\.0+)?\z/` allows '123.0' but rejects '123.45'
- your proposal `/\A-?\d+(\.\d*[1-9])?\z|\A-?\d+\.0+\z/` accepts both and casts them to 123

gigxz (approved):
This looks great and certainly gives me peace of mind! 

If you pull in the latest release-159, the system tests should pass again. I think probably worth waiting for that before merging, to make sure the assessment tests are happy.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -2200,7 +2245,7 @@ def process_assessment(new_hud_values: nil)
           end
 
           it 'should not raise when verified_by_project_id is unrecognized' do
-            values = hud_values.merge({ 'CurrentLivingSituation.verifiedByProjectId' => 'a random string that isnt an id' })
+            values = hud_values.merge({ 'CurrentLivingSituation.verifiedByProjectId' => '1' }) # a random ID that doesn't correspond to a project
```
gigxz, line 2248: @martha based on the comments [here](https://github.com/greenriver/hmis-warehouse/blob/c17fe8231d4aac722a4b08b077987256755a36ef/drivers/hmis/app/models/hmis/hud/processors/current_living_situation_processor.rb#L20-L22) and [here](https://github.com/greenriver/hmis-warehouse/blob/c17fe8231d4aac722a4b08b077987256755a36ef/drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb#L59) I think it's supposed to be allowed to submit a string (in practice, a project name) for this field.  Did that behavior break? If it was trying to store a string in the `verified_by_project_id` then that seems like a bug that should be fixed in the CLS processor.

martha: @gigxz thanks for catching that, I should have thought it through more carefully. I opened a PR https://github.com/greenriver/hmis-warehouse/pull/5276

PR #5230, opened 2025-03-22
Collections use GrdaWarehouse::Lookups::CocCode

eanders:
## _Merging this PR_

## Description
This adjusts the way we provision access to projects through CoC Codes. Instead of using a json array on the Collection, we use the `CoCCode` model in the same way we provision through `Organization`.  

## Type of change
* Code clean-up


drivers/hmis/spec/requests/hmis/update_referral_posting_spec.rb
```diff
@@ -17,7 +19,7 @@
   let!(:link_creds) do
     create(:ac_hmis_link_credential)
   end
-  let!(:coc) { create :hud_project_coc, data_source: ds1, project_id: destination_project.project_id, state: 'MA' }
+  let!(:coc) { create :hud_project_coc, data_source: ds1, project_id: destination_project.project_id, state: 'MA', coc_code: 'MA-500' }
```
gigxz: I tested this without the diff and it doesn't seem like it's needed for the test to pass, am I missing something? The factory sets it to `XX-500`

eanders: I think there was a period of time where the lookup CoC values for XX-xxx weren't correctly set.  It can probably be removed now.  I'll give it a go.

app/models/grda_warehouse/lookups/coc_code.rb
```diff
@@ -51,8 +66,12 @@ def self.options_for_select(user:, permission: :can_view_projects)
...
   def as_select_option
     [
-      "#{preferred_name || official_name} (#{coc_code})",
+      name,
       coc_code,
     ]
   end
+
+  def name
+    "#{preferred_name || official_name} (#{coc_code})"
```
ttoomey: do we need to handle blank strings?

```suggestion
      "#{preferred_name.presence || official_name.presence || 'Unknown'} (#{coc_code.presence || 'No CoC'})"
```

app/views/admin/collections/_projects.haml
```diff
@@ -23,8 +23,8 @@
                         %span{data: { toggle: :tooltip, title: "Additional Sources <br /> #{additional_sources}", html: :true }}
                           %i.icon-warning
                     .project.pl-4.mb-4= entity.ProjectName
-                - elsif entity_type == :coc_codes
-                  .project.pl-4.mb-4= entity
+                  -# - elsif entity_type == :coc_codes
+                  -#   .project.pl-4.mb-4= entity
```
dtgreiner: Do we want to remove this commented out code?

dtgreiner (approved):

app/models/concerns/user_concern.rb
```diff
@@ -666,11 +666,11 @@ def add_viewable(*viewables)
 
     def coc_codes(force_calculation: false)
       key = [self.class.name, __method__, id]
-      Rails.cache.delete(key) if force_calculation
+      Rails.cache.delete(key) if force_calculation || Rails.env.test?
```
ttoomey, line 669: is this needed? In `config/environments/test.rb` we use the null store:
```
config.cache_store = :null_store
```

app/models/collection.rb
```diff
@@ -21,6 +23,10 @@ class Collection < ApplicationRecord
   acts_as_paranoid
   has_paper_trail
 
+  self.ignored_columns = [
+    :coc_codes,
+  ]
```
ttoomey, line 30: we'd be removing these right? maybe add todo or die to drop the col

ttoomey (approved):
Looks good! 

I think we should add a unique constraint to `coc_codes.coc_code`, it seems like we can have duplicates which could cause bugs.

app/models/concerns/user_concern.rb
eanders: I'm unclear why, but yes, this seems to be cached for long enough to cause problems when running the tests as a whole.  Individually, they all pass.

app/models/grda_warehouse/lookups/coc_code.rb
eanders: Official name and CoC Code should never be blank, adding presence call on preferred_name to handle empty strings.

PR #5231, opened 2025-03-22
7357 bugfix de-duplication

ttoomey:
## _Merging this PR_

## Description
* fixes handling of duplicates in client_sum and client data. Previous use of 'distinct' was summing / plucking the distinct values of `column` whereas the intention was to sum/pluck `column` on distinct clients
* adds tests to prove fix
* uses a 'query' object pattern to clean up code
* comments

## Type of change
Bug fix

### DIFF
--- a/drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
+++ b/drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
@@ -644,4 +644,267 @@ RSpec.describe 'Performance Measurement and SPM Alignment', type: :model do
       expect(pm_count).to eq(2) # 2 exits (client 2 from ES and client 4 from RRH)
     end
   end
+
+  describe 'result_calculation#client_sum method' do
+    include_context 'SPM test setup'
+
+    before do
+      # Create projects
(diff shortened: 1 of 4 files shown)


drivers/performance_measurement/spec/models/performance_measurement/pm_spm_alignment_spec.rb
```diff
@@ -644,4 +644,267 @@ def setup_reports(projects:)
...
+
+    it 'should correctly sum values from different clients with the same value' do
+      # Get the result from client_sum for the days_homeless_es_sh_th field
+      sum_result = @pm_report.client_sum(:days_homeless_es_sh_th, :reporting)
+
+      # What we should get is the sum of all client values, even if they're the same
+      expected_result = @days_homeless_value * 2 # 60
+
+      # Check that we're getting the correct sum (both clients' values)
+      expect(sum_result).to eq(expected_result),
```
eanders: Is it worth adding a test that worked the old way to confirm it returns a different result (essentially to confirm we didn't just accidentally get the right result?)

eanders (approved):
Thanks @ttoomey, that looks like it was probably tricky to track down.

PR #5232, opened 2025-03-23
7268 custom hmis data

ttoomey:
## _Merging this PR_

## Description
Enrich analysis view:

- [x] export account users, periodically synced from app db
- [x] update `Exit view  to include auto_exited 
- [x] add `clh_locations` (renamed go `client_geolocations`)
- [x] add `CustomCaseNote` (renamed to `hmis_case_note`, snake_cased fields)
- [x] add `hmis_external_form_submissions` table, excluding form submission blob
- [x] add `hmis_staff_assignments`
- [x] add `hmis_staff_assignment_relationships`
- [ ] reviewed current view for `CustomDataElements` - it appears to be unconstrained so all data should already be available
- [x] add `hmis_client_alerts`

## Type of change
Feature


db/views/analytics_client_geolocations_v01.sql
```diff
@@ -0,0 +1,18 @@
...
+    source_type,
+    source_id,
+    located_on,
+    lat,
+    lon,
+    -- collected_by,
+    -- processed_at,
+    created_at,
+    updated_at,
+    enrollment_id,
```
gigxz: I think we want to drop `enrollment_id`, from my understanding this field was legacy for some imported warehouse data, but doesn't get set anymore. If the location is tied to an enrollment, it is related through `source_id`/`source_type`.

```suggestion
```

db/views/analytics_client_geolocations_v01.sql
```diff
@@ -0,0 +1,18 @@
+SELECT
+    id,
+    client_id,
+    source_type,
+    source_id,
+    located_on,
+    lat,
+    lon,
+    -- collected_by,
```
gigxz, line 10: For HMIS-collected Locations `collected_by` gets filled as the Project Name. However they are tied to an `Enrollment` via `source_id/`source_type`, so the project name is redundant. So I agree with leaving it out. (It might be useful for some warehouse installations that import(ed) location data, though - I'm unsure. But I don't think we should prioritize that for the analytics view).

db/views/analytics_client_geolocations_v01.sql
```diff
@@ -0,0 +1,18 @@
+SELECT
+    id,
+    client_id,
```
gigxz, line 3: Noting a weird thing: for imported Warehouse locations, `client_id` points to a destination client. For HMIS-collected geolocations, `client_id` points to a source client. I think the purpose of this is only to bring along HMIS data, so I think its ok for dbt to just join this to source clients (however/wherever that occurs!).

db/views/analytics_hmis_case_notes_v01.sql
```diff
@@ -0,0 +1,17 @@
+-- rename from custom_case_notes to hmis_case_notes
+SELECT
+    id,
+    "CustomCaseNoteID" AS custom_case_note_id,
```
gigxz: tbh don't think we need this, did you have an idea for why to keep it?

If we do keep it, maybe rename to `case_note_id` since we dropped the "custom" from the table.

db/views/analytics_hmis_staff_assignments_v01.sql
```diff
@@ -0,0 +1,13 @@
+SELECT
+    id,
+    user_id,
+    household_id,
+    hmis_staff_assignment_relationship_id,
```
gigxz: It seems maybe overly complicated to send out the `hmis_staff_assignment_relationships` table to superset/dbt. It is a small configuration table specifying which types of relationships are available in the installation. I think this would be a case where it may make sense to do some pre-processing in the view? Do you agree?

My ActiveRecord version of this is: `Hmis::StaffAssignment.joins(:staff_assignment_relationship).select(:id, :user_id, :household_id, Hmis::StaffAssignmentRelationship.arel_table[:name], :data_source_id, :created_at, :updated_at).to_sql`


```suggestion
    "hmis_staff_assignment_relationships"."name",
```

db/views/analytics_exits_v02.sql
```diff
@@ -0,0 +1,66 @@
+SELECT
+  "ExitID",
+  "EnrollmentID",
+  "PersonalID",
+  "ExitDate",
+  "Destination",
+  "OtherDestination",
+  "AssessmentDisposition",
+  "OtherDisposition",
```
gigxz: Suggest removing these as they are from an old version of the HUD spec
```suggestion
```

db/views/analytics_exits_v02.sql
```diff
@@ -0,0 +1,66 @@
...
+  "HousingAssessment",
+  "SubsidyInformation",
+  "ConnectionWithSOAR",
```
gigxz: Suggest removing this, not sure why its on  the Exit table, this is a field on IncomeBenefit (in the current spec)
```suggestion
```

db/views/analytics_exits_v02.sql
```diff
@@ -0,0 +1,66 @@
...
+  "WrittenAftercarePlan",
+  "AssistanceMainstreamBenefits",
+  "PermanentHousingPlacement",
+  "TemporaryShelterPlacement",
+  "ExitCounseling",
+  "FurtherFollowUpServices",
+  "ScheduledFollowUpContacts",
+  "ResourcePackage",
+  "OtherAftercarePlanOrAction",
```
gigxz: Suggest removing all of these, they are not in the current version of the HUD Spec, which leads me to believe they are likely from old versions of the spec. (I did spot-check that for some but not all).

```suggestion
```

db/views/analytics_exits_v02.sql
```diff
@@ -0,0 +1,66 @@
...
+  "ProjectCompletionStatus",
+  "EarlyExitReason",
+  "FamilyReunificationAchieved",
```
gigxz: from old hud spec, has been retired
```suggestion
```

app/models/grda_warehouse/tasks/sync_analysis_data_task.rb
```diff
@@ -0,0 +1,88 @@
...
+          sync_app_users
+          prune_removed_users
+        end
+      end
+    end
+
+    protected
+
+    def sync_app_users(batch_size: 500)
+      User.find_in_batches(batch_size: batch_size) do |batch|
```
gigxz:
```suggestion
      User.with_deleted.find_in_batches(batch_size: batch_size) do |batch|
```
I _think_ we want to retain users that have been deleted, who may have data referencing them. That way for historical data its still possible to see the related staff user information.

I don't think users get deleted often (ever? it may not be possible in the UI?) but the model and schema allow it.

gigxz (approved):
This looks great! Thanks for setting up the user sync!

I may have gone overboard with my review on the Exit view, but I think it's a good idea to exclude deprecated fields so they don't confusingly show up in OP Analytics. Were they perhaps already being excluded later in the pipeline? I know we have lots of retired HUD fields on other HUD tables as well.

eanders (approved):

db/views/analytics_hmis_case_notes_v01.sql
ttoomey: I agree, I'll exclude it from export

db/views/analytics_hmis_staff_assignments_v01.sql
ttoomey: seems like a good simplification, thanks!

ttoomey (commented):
> I may have gone overboard with my review on the Exit view, but I think it's a good idea to exclude deprecated fields so they don't confusingly show up in OP Analytics. Were they perhaps already being excluded later in the pipeline? I know we have lots of retired HUD fields on other HUD tables as well.

I think this great, thanks for highlighting out all the deprecated fields. I think it's worth it to have better hygiene in our views. I imagine those fields were exported to DBT but not in use. 

We can confirm on our Analytics WG call this week

db/views/analytics_exits_v02.sql
```diff
@@ -0,0 +1,70 @@
...
+  -- "ResourcePackage",
+  -- "OtherAftercarePlanOrAction",
+  "ProjectCompletionStatus",
+  "EarlyExitReason",
+  -- from old hud spec, has been retired
+  -- "FamilyReunificationAchieved",
+  "DateCreated",
+  "DateUpdated",
+  "UserID",
+  --"DateDeleted",
```
Jacobihi, line 32: This is the only field we're referencing currently in dbt, but because these views filter it out by definition it makes sense to remove references to it. We can do a very quick dbt update before you push this to remvoe references to date deleted

ttoomey: That sounds great; just let me know if it seems like too much trouble just for housekeeping. I think this is the same across all of the analytics views, we probably don't need to export DateDeleted as deleted records are always filtered out.

Jacobihi: Following up here https://github.com/open-path/Green-River/issues/7517

gigxz:
fyi- system test failure is unrelated that should be fixed by merging in latest 158.

gigxz:
@eanders @ttoomey would you be comfortable with this being merged into 158? (rather than 159). There is some time-sensitive need that came up (see op analytics slack)

eanders:
158 works for me.

PR #5234, opened 2025-03-24
7440-hdx-upload-corrections

ttoomey:
spec

## _Merging this PR_

## Description
Adjust HDX export to more closely meet the spec doc

Details:
* Replaced the array-based column definitions with a struct-based object model
* Added explicit data type specifications for all HDX columns
* implemented formatting per the spec
* added default values since HDX cols cannot be null 
* added tests 

## Type of change
Bug/Spec

### DIFF
--- a/drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/hdx_upload.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/hdx_upload.rb
@@ -215,7 +215,7 @@ module HudSpmReport::Generators::Fy2024
       when :string
         value.presence || ''
       else
-        raise ArgumentError, "data type \"#{date_type}\" not supported"
+        raise ArgumentError, "data type \"#{data_type}\" not supported"
       end
     end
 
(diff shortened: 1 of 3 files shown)


drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/hdx_upload.rb
```diff
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
```
eanders, line 2: This needs to go below the copyright or our update script will generate duplicate copyright notices.

eanders (approved):
Looks good to me, thanks @ttoomey !  
Much bigger refactor than I was thinking, but I think it will serve us well into the future.

ttoomey:
> Much bigger refactor than I was thinking, but I think it will serve us well into the future.

Thanks for the review, sorry it was bigger than expected.  I wanted a simpler approach initially; after digging in I thought it warranted a larger change that would give more confidence in in passing validation

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2024/hdx_upload.rb
ttoomey: thanks, I feel like we should probably fix our script right? Seems like it should handle this kind of thing. I'll make an issue to adjust it

PR #5235, opened 2025-03-24
Support the Project CE page with referrals queue

martha:
## _Merging this PR_
- squash-merge into the long lived feature branch `tt-ce-spike-3`

## Description
GH issue: https://github.com/open-path/Green-River/issues/7350

Summary of changes:
- Add `ce_referrals` on `project` schema object.
- Add `status` filter option to both Opportunity and Referral schema objects.
- Add `categories` to Opportunity schema object.
- Add `date_started` and `current_step` to Referral schema object
  -  Maybe we should discuss Current Step a bit more. In reality, there can be >1 current step, though the [WIP mocks](https://www.figma.com/board/axGI9vwmuBRC4uZ11ZJ3RP/OP-HMIS-%2F-Coordinated-Entry-%2F-FigJam?node-id=2821-78000&t=LtaE21FnWaZml1Rw-0) don't currently reflect that.
- In the starter pack, add some example Opportunity Categories
- Add spec

## Type of change
Feature build-out

  - [x] ensured the API can't leak data from other data sources - I _think_ this is true in this PR
  - [ ] ensured permissions and visibility checks are performed in the right places - todo in CE Permissions ticket

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
@@ -8,12 +8,19 @@
 
 module Types
   class HmisSchema::CeReferral < Types::BaseObject
+    # object is a Hmis::Ce::Referral
     field :id, ID, null: false
     field :opportunity, HmisSchema::CeOpportunity, null: false
     field :steps, [HmisSchema::CeReferralStep], null: false
     field :status, HmisSchema::Enums::CeReferralStatus, null: false
     field :client_id, ID, null: false
     field :client, Types::HmisSchema::Client, null: true
+    field :date_started, GraphQL::Types::ISO8601Date, null: false, method: :created_at
+    field :current_step, Types::HmisSchema::CeReferralStep, null: true
+
+    available_filter_options do
+      arg :status, [HmisSchema::Enums::CeReferralStatus]
+    end
 
     def steps
       instance = object.workflow_instance
@@ -36,6 +43,16 @@ module Types
 
     def client
       load_ar_association(object, :client, scope: Hmis::Hud::Client.viewable_by(current_user))
+    end
+
+    def current_step
+      instance = load_ar_association(object, :workflow_instance)
+      steps = load_ar_association(instance, :steps, scope: Hmis::WorkflowExecution::Step.where(status: ['available', 'in_progress']))
+      steps.first # There can be multiple steps currently in progress, but we're only going to show one in the project referrals table
+    end
+
+    def opportunity
+      load_ar_association(object, :opportunity)
     end
   end
 end
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
@@ -15,8 +15,8 @@ module Types
     field :status, HmisSchema::Enums::CeReferralStatus, null: false
     field :client_id, ID, null: false
     field :client, Types::HmisSchema::Client, null: true
-    field :date_started, GraphQL::Types::ISO8601Date, null: false, method: :created_at
-    field :current_step, Types::HmisSchema::CeReferralStep, null: true
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :current_step_name, String, null: true
 
     available_filter_options do
       arg :status, [HmisSchema::Enums::CeReferralStatus]
@@ -45,10 +45,20 @@ module Types
       load_ar_association(object, :client, scope: Hmis::Hud::Client.viewable_by(current_user))
     end
 
-    def current_step
+    def current_step_name # There can be multiple steps currently in progress, but we're only going to show one in the project referrals table
       instance = load_ar_association(object, :workflow_instance)
-      steps = load_ar_association(instance, :steps, scope: Hmis::WorkflowExecution::Step.where(status: ['available', 'in_progress']))
-      steps.first # There can be multiple steps currently in progress, but we're only going to show one in the project referrals table
+
+      step_t = Hmis::WorkflowExecution::Step.arel_table
+      step_scope = Hmis::WorkflowExecution::Step.
+        where(status: ['available', 'in_progress']).
+        # Prefer to return a step that is in_progress over available
+        order(Arel::Nodes::Case.new.when(step_t[:status].eq('in_progress')).then(1).else(2)).
+        order(step_t[:id]) # Also sort by ID, to make sure this resolves deterministically
+
+      step = load_ar_association(instance, :steps, scope: step_scope).first
+      return if step.nil?
+
+      load_ar_association(step, :node)&.name
     end
 
     def opportunity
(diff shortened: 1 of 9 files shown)


drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -8,12 +8,19 @@
...
 module Types
   class HmisSchema::CeReferral < Types::BaseObject
+    # object is a Hmis::Ce::Referral
     field :id, ID, null: false
     field :opportunity, HmisSchema::CeOpportunity, null: false
     field :steps, [HmisSchema::CeReferralStep], null: false
     field :status, HmisSchema::Enums::CeReferralStatus, null: false
     field :client_id, ID, null: false
     field :client, Types::HmisSchema::Client, null: true
+    field :date_started, GraphQL::Types::ISO8601Date, null: false, method: :created_at
```
gigxz: This seems like something we can resolve as a DateTime, since we have the time. That will allow recently-created referrals to show as "3 hours ago" instead of "today" in the relative date display. Or if we decide to only show the date in the frontend, that's fine -- but lets resolve the full datetime so that we have that flexibility as the frontend design evolves.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -37,5 +44,15 @@ def steps
     def client
       load_ar_association(object, :client, scope: Hmis::Hud::Client.viewable_by(current_user))
     end
+
+    def current_step
+      instance = load_ar_association(object, :workflow_instance)
+      steps = load_ar_association(instance, :steps, scope: Hmis::WorkflowExecution::Step.where(status: ['available', 'in_progress']))
+      steps.first # There can be multiple steps currently in progress, but we're only going to show one in the project referrals table
```
gigxz: Can you add a consistent sort order to the `scope:` above, so that this resolves deterministically when there are multiple open steps? (And, I think we should prefer to display a step that is `in_progress` over `available`. )

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -8,12 +8,19 @@
...
   class HmisSchema::CeReferral < Types::BaseObject
+    # object is a Hmis::Ce::Referral
     field :id, ID, null: false
     field :opportunity, HmisSchema::CeOpportunity, null: false
     field :steps, [HmisSchema::CeReferralStep], null: false
     field :status, HmisSchema::Enums::CeReferralStatus, null: false
     field :client_id, ID, null: false
     field :client, Types::HmisSchema::Client, null: true
+    field :date_started, GraphQL::Types::ISO8601Date, null: false, method: :created_at
+    field :current_step, Types::HmisSchema::CeReferralStep, null: true
```
gigxz: If this is only for displaying in the table, perhaps `field :current_step_name, String` would be more appropriate? 

Especially given the comment below (_There can be multiple steps currently in progress, but we're only going to show one in the project referrals table_). Since the logic for resolving this field has a heuristic like this, it seems like we _don't_ want to use the `currentStep` field for other purposes (say, for example, rendering the "current" step form on the referral page).

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -253,10 +259,20 @@ def external_form_submissions(**args)
       resolve_external_form_submissions(scope, **args)
     end
 
-    def ce_opportunities
+    def ce_opportunities(filters: nil)
+      raise unless Hmis::Ce.configuration.enabled?
+
+      scope = load_ar_association(object, :ce_opportunities)
+      scope = scope.where(status: filters&.status) if filters&.status.present?
+      scope.order(created_at: :desc)
```
gigxz: A little confusing to use the data loader since this field would have N+1 anyway from `where` and `order` clauses. Prefer removing data loader, so it's clearer to the reader that this field is N+1 and only should be resolved on a single Project.
```suggestion
      scope = object.ce_opportunities
      scope = scope.where(status: filters&.status) if filters&.status.present?
      scope.order(created_at: :desc)
```

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -253,10 +259,20 @@ def external_form_submissions(**args)
...
+    end
+
+    def ce_referrals(filters: nil)
       raise unless Hmis::Ce.configuration.enabled?
 
-      load_ar_association(object, :ce_opportunities)
+      scope = load_ar_association(object, :ce_referrals)
```
gigxz: same comment here about data loader use

drivers/hmis/spec/factories/hmis/ce/referrals.rb
```diff
@@ -2,7 +2,10 @@
 
 FactoryBot.define do
   factory :hmis_ce_referral, class: 'Hmis::Ce::Referral' do
-    association(:opportunity, factory: :hmis_ce_opportunity)
+    transient do
+      project { nil }
+    end
+    opportunity { build(:hmis_ce_opportunity, project: project) }
```
gigxz: Cool, I didn't know you could do it like this! I usually use `after(:create)` to set the transient properties, but this is way neater. Will this work for both build and create?

drivers/hmis/spec/requests/hmis/ce/project_ce_referrals_spec.rb
```diff
@@ -0,0 +1,113 @@
...
+            create(:hmis_ce_referral, status: :rejected, project: project)
+          end
+        end
+
+        it 'queries the db a reasonable amount' do
+          expect do
+            response, result = post_graphql(**variables) { query }
+            expect(response.status).to eq(200), result.inspect
+            expect(result.dig('data', 'project', 'ceReferrals', 'nodesCount')).to eq(102)
+          end.to make_database_queries(count: 20..25)
```
gigxz: Thanks for adding these tests!

drivers/hmis/spec/requests/hmis/ce/project_ce_opportunities_spec.rb
```diff
@@ -0,0 +1,100 @@
...
+        expect(opportunities.count).to eq(1)
+
+        expect(opportunities).to contain_exactly(
+          a_hash_including('id' => opportunity.id.to_s),
+        )
+      end
+
+      context 'with many opportunities' do
+        before do
+          100.times do
```
gigxz: I would bring this down from 100 just for the sake of test speed, it seems a bit high for testing the n+1. I would think something like 30-40 would suffice (since it's higher than expected db queries `15..20`)

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: And maybe we should call it `created_at` or `date_created`? I'm on the fence there. There is a difference between the datetime when the referral was created, and the datetime when the first referral _step_ was initiated.

I'm also thinking about potential future scenarios where the top prioritized eligible client gets auto-referred. In that scenario, `created_at` feels more appropriate than `date_started` since the creation was system-initiated. (Date Started feels user-initiated to me somehow, maybe I'm overthinking things).

gigxz (commented):
looks great! some minor comments on naming and such

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Sounds good! I was relying too much on the WIP mocks, which have a "started" column. But you're right, thinking about it more, it seems preferable to keep the naming of this field consistent in the schema and database, unless there is a strong need for them to diverge.

drivers/hmis/spec/factories/hmis/ce/referrals.rb
martha: Hm, yeah you are right, this doesn't work well with `create`, which I ran into when writing tests for the changes in `current_step`. I'll update this to match our usual pattern.

(You can replace `build` with `create` here on line 8, but then the factory would always save to the db, and couldn't be used to just make an in-memory object. I was also running into errors with transient properties that were required and couldn't correctly default to the factory association.)

gigxz (approved):
changes look good to me!

PR #5237, opened 2025-03-24
Add HMIS ACLs summary to user permission report

gigxz:
## _Merging this PR_

## Description
1. Add HMIS User Permissions sheet to the **User Permissions Report** download. (issue: https://github.com/open-path/Green-River/issues/7402)
    * Path for testing: `/user_permission_report/warehouse_reports/reports`
2. Update **Manage User Sessions** page to include active HMIS sessions, and specify whether most recent login was Warehouse or HMIS, and to display last login location for HMIS login, and to update confirmation warning text in HMIS-enabled installations.
    * Path for testing: `/admin/sessions`
3. Expose login activities in the HMIS GraphQL API, so they can be displayed on existing HMIS User Audit page. (The counterpart to warehouse user audit page).
    * Also include the account edit page so we can link to it. frontend changes are at https://github.com/greenriver/hmis-frontend/pull/1111 but haven't done self-review yet

## How to test
- Test the above paths with and without HMIS being "enabled". Can mock it by updating [this line](https://github.com/greenriver/hmis-warehouse/blob/a8a3cb735295bc18984563175ca9cfbe8aedc2c5/app/models/hmis_enforcement.rb#L9).

## Type of change
* new feature

### DIFF
--- a/app/models/concerns/user_concern.rb
+++ b/app/models/concerns/user_concern.rb
@@ -54,8 +54,12 @@ module UserConcern
     has_many :access_grants, class_name: 'Doorkeeper::AccessGrant', foreign_key: :resource_owner_id, dependent: :delete_all # or :destroy if you need callbacks
     has_many :access_tokens, class_name: 'Doorkeeper::AccessToken', foreign_key: :resource_owner_id, dependent: :delete_all # or :destroy if you need callbacks
 
-    # Connect users to login attempts
+    # Connect users to login attempts.
+    # Only includes Warehouse activity when called on User record, and only HMIS activity for Hmis::User record.
     has_many :login_activities, as: :user
+
+    # All login activities for user, including both HMIS and Warehouse login activity
+    has_many :all_login_activities, class_name: 'LoginActivity', foreign_key: 'user_id'
 
     # TODO: START_ACL remove when ACL transition complete
     # Ensure that users have a user-specific access group
@@ -138,9 +142,10 @@ module UserConcern
       active.not_system.where(exclude_from_directory: false)
     end
 
+    # users that have currently active sessions (either in the warehouse or in HMIS)
     scope :has_recent_activity, -> do
       where(last_activity_at: timeout_in.ago..Time.current).
-        where.not(unique_session_id: nil)
+        where.not(unique_session_id: nil, hmis_unique_session_id: nil)
     end
 
     scope :using_acls, -> do
@@ -166,6 +171,17 @@ module UserConcern
 
       # Find most recent HMIS login using cached map of LoginActivities
       LoginActivity.latest_hmis_logins[id]
+    end
+
+    # whether the user's most recent login was to the Warehouse or HMIS
+    def last_login_application
+      if last_warehouse_login && last_hmis_login
+        last_warehouse_login > last_hmis_login ? 'Warehouse' : 'HMIS'
+      elsif last_warehouse_login
+        'Warehouse'
+      elsif last_hmis_login
+        'HMIS'
+      end
     end
 
     def using_acls?
--- a/app/models/concerns/user_concern.rb
+++ b/app/models/concerns/user_concern.rb
@@ -151,6 +151,23 @@ module UserConcern
       where(permission_context: [nil, 'role_based'])
     end
 
+    # datetime of most recent warehouse login
+    def last_warehouse_login
+      # for non-HMIS installations, we can use the devise current_sign_in_at directly because there is only one scope
+      return current_sign_in_at unless HmisEnforcement.hmis_enabled?
+
+      # Find most recent Warehouse login using cached map of LoginActivities
+      LoginActivity.latest_warehouse_logins[id]
+    end
+
+    # datetime of most recent HMIS login
+    def last_hmis_login
+      return unless HmisEnforcement.hmis_enabled?
+
+      # Find most recent HMIS login using cached map of LoginActivities
+      LoginActivity.latest_hmis_logins[id]
+    end
+
     def using_acls?
       # Note using hash syntax to get around lack of column for some data migrations
       self[:permission_context].to_s == 'acls'
--- a/app/models/concerns/user_concern.rb
+++ b/app/models/concerns/user_concern.rb
@@ -156,34 +156,6 @@ module UserConcern
       where(permission_context: [nil, 'role_based'])
     end
 
-    # datetime of most recent warehouse login
-    def last_warehouse_login
-      # for non-HMIS installations, we can use the devise current_sign_in_at directly because there is only one scope
-      return current_sign_in_at unless HmisEnforcement.hmis_enabled?
-
-      # Find most recent Warehouse login using cached map of LoginActivities
-      LoginActivity.latest_warehouse_logins[id]
-    end
-
-    # datetime of most recent HMIS login
-    def last_hmis_login
-      return unless HmisEnforcement.hmis_enabled?
-
-      # Find most recent HMIS login using cached map of LoginActivities
-      LoginActivity.latest_hmis_logins[id]
-    end
-
-    # whether the user's most recent login was to the Warehouse or HMIS
-    def last_login_application
-      if last_warehouse_login && last_hmis_login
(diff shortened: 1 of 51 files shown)


drivers/user_permission_report/app/views/user_permission_report/warehouse_reports/reports/index.xlsx.axlsx
```diff
@@ -23,11 +23,27 @@ wb.add_worksheet(name: 'User Permissions') do |sheet|
...
+      user.last_warehouse_login&.to_time&.to_fs(:db),
       user.legacy_roles.map(&:name).join(', '),
       user.access_groups.filter(&:general?).map(&:name).join(', '),
       *@group_associations.keys.map { |ga| associated_items(user, ga)[:names].join(', ') }
     ]
     sheet.add_row(row, style: data_cell)
   end
 end
+
+if HmisEnforcement.hmis_enabled?
```
gigxz: TODO: Also check if this user as an `Hmis::User` would have permission `can_audit_users`. Weird to check an HMIS permission for a Warehouse report, but seems correct?

drivers/user_permission_report/app/views/user_permission_report/warehouse_reports/reports/index.haml
```diff
@@ -44,10 +44,13 @@
                   - else
                     = render 'admin/inactive_users/user_status', user: user
                 %br
-                - if user.last_sign_in_at.present?
-                  Last login: #{user.last_sign_in_at}
```
gigxz, line 48: this was showing the "last_sign_in_at" which is the login from their PREVIOUS session, not their most recent session. I assume that was unintentional so updated it to show the most recent login instead

martha (approved):
Everything looks good to me. Thanks for tagging me, it's interesting to get some exposure to this part of the codebase!

app/models/concerns/user_concern.rb
```diff
@@ -151,6 +156,34 @@ module UserConcern
...
+
+    # whether the user's most recent login was to the Warehouse or HMIS
+    def last_login_application
+      if last_warehouse_login && last_hmis_login
+        last_warehouse_login > last_hmis_login ? 'Warehouse' : 'HMIS'
+      elsif last_warehouse_login
+        'Warehouse'
+      elsif last_hmis_login
+        'HMIS'
+      end
```
eanders: What about something like:
```suggestion
      [['Warehouse', last_warehouse_login], ['HMIS', last_hmis_login]].select { |_, date| date.present? }&.max_by(&:last)&.first
```

There's probably something slightly more descriptive, but this would decrease some of the duplication.

app/models/login_activity.rb
```diff
@@ -4,14 +4,44 @@
...
   end
+
+  # Class method to preload the most recent successful Warehouse login for each user.
+  # Uses caching to avoid N+1 in reports.
+  def self.latest_warehouse_logins
+    Rails.cache.fetch('latest_warehouse_logins', expires_in: 1.minute) do
+      LoginActivity.successful.warehouse_logins.
+        select('DISTINCT ON (user_id) user_id, created_at').
+        order(:user_id, created_at: :desc).
+        map { |r| [r.user_id, r.created_at] }.to_h
```
eanders: Does it make sense to add some sort of date limit on this? (maybe 1 year, or maybe it could be based on the account expiration length?)  I'm concerned with hitting a table that might have millions of rows that this might not be performant in production.

drivers/user_permission_report/app/controllers/user_permission_report/warehouse_reports/reports_controller.rb
```diff
@@ -40,5 +46,45 @@ def associated_items(user, meth)
       }
     end
     helper_method :associated_items
+
+    def hmis_access_report_rows
```
eanders: this is fine for now as it follows the existing pattern, but it sure feels like this is getting big (and important) enough that we should build out a report class to hold the logic.  What do you think?

eanders (commented):
@gigxz I think this looks good, dropped a few thoughts, let me know if you have questions or want to discuss further.

drivers/user_permission_report/app/controllers/user_permission_report/warehouse_reports/reports_controller.rb
gigxz: The existing logic is in the view and axlsx files, so I figured it was an improvement to move the HMIS to the controller instead of the axlsx file, but I suppose a class would be even better.  Do you have an example of an existing class that I could copy from? Is something like `AccessLogs::Report` the right pattern?

gigxz: I also opened https://github.com/open-path/Green-River/issues/7483 for updating the warehouse user permission sheet to support ACLs, so maybe when we do that we can move that logic into the class as well

app/models/login_activity.rb
gigxz: Thats a good point. I didn't check the indexes on this table and it looks like it only has indices on `id`, `ip`, and `identity` (appears to hold email address) but not on `user_id`. So this will probably be inefficient.

A year limit sounds reasonable, though we would probably want to amend any wording like "never logged in" to "not in the past year", "last login"=>"last login in the past year" etc. which seems a little annoying. 

Maybe it would be better if we just added application-specific last-login fields to the `users` table... since that's really all we need here. For example `users.last_warehouse_login` and `users.last_hmis_login`. That would function like `current_sign_in_at` but scoped. I can look into that, it does sound like a better approach especially given the indices on  this table? (We would need to initially populate them with a migration based on login activities..)

gigxz: I'm going to remove this login_activities-related stuff from the PR and just keep the HMIS ACLs report, and open an issue to follow up, because its getting out of scope. I will just remove the "Last login time" from the HMIS sheet since its too expensive to calculate.

eanders: I vote for adding the fields to the user table with a migration to populate them.

drivers/user_permission_report/app/controllers/user_permission_report/warehouse_reports/reports_controller.rb
eanders: `AccessLogs::Report` looks like a great starting point.

app/views/admin/sessions/index.haml
```diff
@@ -17,13 +17,16 @@
       %th Force Logout
     %tbody
       - @users.each do |user|
+        - last_login_activity = user.all_login_activities.successful.last
```
gigxz, line 20: Dont think this is very performant either, but this page was already doing a query like this to display `location_description`. If there are many active users, this could become a big n+1.

app/models/login_activity.rb
gigxz: ok agreed, reverted those changes and opened https://github.com/open-path/Green-River/issues/7486!

drivers/user_permission_report/app/views/user_permission_report/warehouse_reports/reports/index.haml
```diff
@@ -44,16 +44,18 @@
...
                 - else
                   Never logged in
             %td
               = link_to user_permission_report_warehouse_reports_user_path(user), data: {loads_in_pjax_modal: true}, class: 'btn btn-xs btn-primary btn-muted btn-hmis' do
                 View Details
             %td
               - if user.using_acls?
-                TODO
+                %ul
+                  - user.legacy_roles.map{|r| link_to_if can_edit_roles?, r.name, edit_admin_role_path(r)}.uniq.each do |role_name|
```
eanders: This seems slightly odd to me that we'd check for using_acls and then use `legacy_roles`.  Should we adjust that?

eanders (approved):
Thanks for gathering up the report logic @gigxz, much cleaner I think.

drivers/user_permission_report/app/views/user_permission_report/warehouse_reports/reports/index.haml
gigxz: whoops I meant `roles`, good catch

PR #5238, opened 2025-03-24
7462-spm-details-m2-bugfix

ttoomey:
## _Merging this PR_

## Description
Fixes crash when trying to view client details on measure 2 cells in the SPM

## Type of change
Bug fix

### DIFF
--- a/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/return.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/fy2023/return.rb
@@ -19,7 +19,7 @@ module HudSpmReport::Fy2023
     has_many :hud_reports_universe_members, inverse_of: :universe_membership, class_name: 'HudReports::UniverseMember', foreign_key: :universe_membership_id
 
     def project_id
-      exit_enrollment&.enrollment&.project_id || return_enrollment&.enrollment&.project_id
+      [exit_enrollment, return_enrollment].detect(&:present?)&.enrollment&.project&.id
     end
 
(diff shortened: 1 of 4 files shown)


drivers/hud_spm_report/app/models/hud_spm_report/fy2024/return.rb
```diff
@@ -18,8 +18,9 @@ class Return < HudReports::ReportClientBase
 
     has_many :hud_reports_universe_members, inverse_of: :universe_membership, class_name: 'HudReports::UniverseMember', foreign_key: :universe_membership_id
 
+    # duck-types to enrollment
     def project_id
```
ttoomey, line 22: This was returning the project_id string, not the project pk. 

I reviewed the current SPM module. It seems like the return.project_id is only used here so I have to imagine this method is incorrect. Note, it would technically be safer to make a new project_pk method instead of changing this behavior but I'm choosing to fix it here in the name of maintainability.

drivers/hud_spm_report/app/models/hud_spm_report/fy2023/return.rb
```diff
@@ -19,7 +19,7 @@ class Return < HudReports::ReportClientBase
     has_many :hud_reports_universe_members, inverse_of: :universe_membership, class_name: 'HudReports::UniverseMember', foreign_key: :universe_membership_id
 
     def project_id
```
ttoomey, line 21: back-port fix so that last years spms don't crash

eanders (approved):

PR #5244, opened 2025-03-26
Implement CE Referral side-effect to enroll client

martha:
## _Merging this PR_
- squash merge into feature branch `tt-ce-spike-3` (unless that gets merged to a release branch first!)

## Description

Time tracked so far: about 6h
Summary of changes:
- Add `target_enrollment` to referral table, model, and schema object
- Refactor logic for the `BulkAssignService` mutation to determine a project CoC code, based on the project and optional input, as a method on Project so it can be shared.
- Add a new `ReferralEnroller`
  - This handles creating enrollments and assigning move-in dates as side-effects of referral steps.
  - It's moved out here to reduce complexity in `ReferralMessageHandler`
  - The code in here is heavily inspired by enrollment creation in the `BulkAssignService` mutation
- Update the Engine to pass around `user` and `step` so they are accessible where needed.
- Add a workflow template that creates enrollments to the Starter Pack.
- Spec:
  - Add spec to test the enrollment and move-in date side effect behavior
  - Refactor `ce_spec_helper` so it lives in a common place (not just under `requests`, since this newly added spec isn't testing Graphql requests.)

How to test:
- Run the starter pack
- In the frontend, create an Opportunity using the "Enrollment Creator" template
- Initiate a referral for that Opportunity. Step through the workflow, including submitting a move-in date
- You should see an enrollment with correct move-in date generated in the project

## Type of change
Feature build-out


drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,64 @@
+# frozen_string_literal: true
+
+module Hmis::Ce
+  class ReferralEnroller
+    attr_reader :referral
+    def initialize(referral)
+      @referral = referral
+    end
+
+    def create_enrollment(message, user)
```
martha: Something I want to discuss more: The ticket says,
> A referral task can be configured to generate a WIP enrollment when a certain response is entered (eg "client accepted: yes")

In this first-pass implementation, I haven't done that. At the moment, if a step has a `trigger_config` which includes the `create_enrollment` message, then an enrollment will get created, regardless of any submitted values.

I was thinking that the conditionality of creating the enrollment could be handled by branching logic and gateways that are already built into the workflow engine, but I haven't really battle tested this idea. Something like: 
- start_workflow -> 
- client_acceptance_task -> 
- gateway -> 
  - accepts? -> create_enrollment_task -> accept_referral (or even: accept_referral _is_ the step that has the create_enrollment trigger config -- but I haven't tested that idea yet.)
  - does not accept? -> reject_referral (no enrollment created.)

Alternatively, we certainly could hard-code some submitted value here to check. Or, we could make it configurable by the step definition what submitted value it's expecting (maybe somehow using `trigger_config`'s `params` property?), but I'd like to discuss/spec that idea more.

martha: Edit: I worked up a demo of the idea described here in an rspec test https://github.com/greenriver/hmis-warehouse/pull/5244/commits/b061e2504be391db1782c0d611b4c6f3293cb288

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+      project = referral.opportunity.project
+      raise 'access denied' unless user.can_enroll_clients_for?(project)
+
+      submitted_values = message.step&.submitted_values&.symbolize_keys
+      coc_code_arg = submitted_values ? submitted_values[:coc_code] : nil
```
gigxz: I'm curious about the safe accessor use `&`, is there a scenario where this evaluates but there is no step and/or no values associated with the step?

This could be a one-liner using `fetch` with a default:
```ruby
      # Step form may specify CoC code. This is required if the Project serves multiple CoCs.
      coc_code_arg = message.step&.submitted_values&.fetch('coc_code', nil)
```

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+      referral.update!(target_enrollment: enrollment)
+    end
+
+    def set_move_in_date(message, user)
+      project = referral.opportunity.project
+      raise 'access denied' unless user.can_edit_enrollments_for?(project)
+
+      # This doesn't raise if the move-in date is missing. If the field is required, it should have already been caught by form validation.
+      submitted_values = message.step&.submitted_values&.symbolize_keys
+      date_string = submitted_values ? submitted_values[:move_in_date] : nil
```
gigxz: Same clarification question for `move_in_date` and `coc_code` magic strings:


Am I correct in understanding that these match the `link_id` of the corresponding form definition item, or do they match the `mapping.field_name`?  We should make it clear with a code comment. 

I'm unsure about hard-coding special link ids since we don't use them this way for other form processing. Another option would be to switch to using `mapping`s, which are intended to be used to specify the mapping aka where the form item should be saved. That would require some additional transformation of the values, though.


If we do want to continue this linkid approach (at least temporarily), we could pull them into constants on the class to make it clearer:
```ruby
MOVE_IN_DATE_LINK_ID = 'move_in_date' # if this link id is used on a referral workflow form, the engine will attempt to store the value on the target Enrollment (and raise if there is no target enrollment)
```

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+      return unless date_string.present?
+
+      enrollment = referral.target_enrollment
+      raise "Trying to set move-in date, but referral #{referral.id} does not have a target enrollment yet. This probably indicates a mistake in the workflow configuration." unless enrollment.present?
+
+      date = HmisUtil::Dates.safe_parse_date(date_string: date_string)
+      enrollment.update!(move_in_date: date)
```
gigxz: This will set move_in_date to `nil` if the string was unparseable, that doesn't seem right. it should probably do nothing if the string is unparseable.

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+      raise 'referral generated invalid enrollment' unless enrollment.valid?
+
+      entry_date_errors = Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date(enrollment)
+      entry_date_errors.reject! { |e| e.warning? && e.type == :information }
+      error_out(entry_date_errors.first.full_message) unless entry_date_errors.empty?
+
+      # TODO(#6709) - assign enrollment to unit
+
+      enrollment.save_new_enrollment! # Saves as WIP or non-WIP, depending on auto-enter rules in the project
+      referral.update!(target_enrollment: enrollment)
```
gigxz, line 52: Is this whole thing already wrapped in a transaction somewhere? we should figure out what the appropriate level is (wrapping the whole message handler `call` in a referral, or only certain message types that update data?).

In any case these 2 lines certainly should be in a transaction. It appears `save_new_enrollment!` is expected to be called in a transaction because it saves the enrollment and the syntheic intake separately.

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def set_move_in_date(message, user)
+      project = referral.opportunity.project
```
gigxz: Might make things easier to add a has_one through so we can call `referral.project` directly (or maybe call it `referral.target_project` for clarity)

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -0,0 +1,70 @@
...
+        user: Hmis::Hud::User.from_user(user),
+        household_id: Hmis::Hud::Base.generate_uuid,
+        relationship_to_hoh: 1, # Head of Household
+        enrollment_coc: coc_code,
+      )
+      raise 'referral generated invalid enrollment' unless enrollment.valid?
+
+      entry_date_errors = Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date(enrollment)
+      entry_date_errors.reject! { |e| e.warning? && e.type == :information }
+      error_out(entry_date_errors.first.full_message) unless entry_date_errors.empty?
```
gigxz: Hmmm good thought! Raising doesn't seem ideal, but we can probably address it later. It seems like the most relevant validation that might come up is if the client is already enrolled. I'll link related issue https://github.com/greenriver/hmis-warehouse/pull/4526 for when we added a similar validation error to existing referral functionality.

I don't see any reason to raise `warning`-level valdiations.

```suggestion
      # Raise an exception if there are any entry date validation errors, which may mean that the client is already enrolled at the project. Future improvement would be to return this as a validation error instead of raising
      entry_date_errors = Hmis::Hud::Validators::EnrollmentValidator.validate_entry_date(enrollment)
      entry_date_errors.reject! { |e| e.warning? }
      error_out(entry_date_errors.map(&:full_message).join(', ')) unless entry_date_errors.empty?
```

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -72,24 +72,26 @@ def next_task_steps(step)
...
 
-    def send_message(...)
+    def send_message(user, ...)
       message = Hmis::WorkflowExecution::Message.new(...)
-      message_handler.call(message)
+      message_handler.call(message, user)
     end
 
-    def process_triggers(node, event_type)
+    def process_triggers(node, event_type, user, step = nil)
```
gigxz: This is becoming a lot of positional arguments to keep track of, I'd suggesting switching to keyword args. 

I am a little confused about needing to pass both `node` and `step` since they have a relationship to each other. Can the step be accessed from the node instead?
```
step = node.step if node.task? # if this is a task node, it should(?) have an associated step
```

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -72,24 +72,26 @@ def next_task_steps(step)
       nodes.map { |node| steps_by_node_id[node.id] }.compact
     end
 
-    def send_message(...)
+    def send_message(user, ...)
       message = Hmis::WorkflowExecution::Message.new(...)
-      message_handler.call(message)
+      message_handler.call(message, user)
```
gigxz: Is there any reason `user` shouldn't be added to the message struct itself? https://github.com/greenriver/hmis-warehouse/blob/64ac6d5448462016ffd9970a543f017f72b47b3f/drivers/hmis/app/models/hmis/workflow_execution/message.rb#L5-L8

seems reasonable enough to add the user who submitted the step

gigxz (commented):
looks good! initial comments below

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -6,6 +6,12 @@
 
 # frozen_string_literal: true
 
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
```
gigxz: there are duplicate copyright notices added in several files, maybe a merge issue?

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -11,6 +17,9 @@ class Referral < GrdaWarehouseBase
     has_many :participants, class_name: 'Hmis::Ce::ReferralParticipant'
     belongs_to :client, class_name: 'Hmis::Hud::Client'
     belongs_to :referred_by, class_name: 'Hmis::User'
+    # Referral belongs_to household, not enrollment, because we don't want to lose the association if the originally
+    # referred member leaves the household but other household members are still occupying the unit.
```
gigxz: comment no longer relevant

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
gigxz: It seems reasonable to me to implement this behavior using gateways. Are there any downsides to that approach that you're seeing?

martha: Hm. It will probably lead to more complicated workflow definitions, but I think that's okay, because the workflow definitions are built to handle complexity in a generic way (thanks Theron!).

I could see us ending up wanting more flexibility. Right now it's a trigger event that can be triggered from any of the following events:
```
start_step
complete_step
pass_gateway
start_workflow
end_workflow
```
which means that anytime we want "sometimes create an enrollment when the workflow finishes a specific step depending on the submitted values," we have to add a gateway that is followed by at least one other step.

Overall I don't think it would be too expensive, or incompatible with the approach in this PR, if we want to change this decision later.

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: Thanks for the one-liner! Yes, I'll expand in the comments.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Thanks! It was because I tried to update my IDE so it would correctly add the copyright notice on new files, and update to 2025 if it's out of date. But it's getting confused when the `frozen_string_literal` comment is below the copyright notice. I'm going to remove the behavior on my IDE for now and try to sort it out another day 😓

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: That's a good question. To me, it looks like the [Engine's](https://github.com/greenriver/hmis-warehouse/blob/tt-ce-spike-3/drivers/hmis/app/models/hmis/workflow_execution/engine.rb) methods (like start_step!, complete_step!, etc) are designed to be run inside of a transaction (because they separately save the step, the log event, and future visited nodes).

But the responsibility for actually wrapping the call in a transaction is delegated to the caller, such as the [mutation](https://github.com/greenriver/hmis-warehouse/blob/tt-ce-spike-3/drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb#L31-L40). Do we think that is safe enough? Would it make sense to make a nested transaction?

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: That's a good point about link IDs. It relates to my comment [here](https://github.com/open-path/Green-River/issues/7400#issuecomment-2755915561) about CE Events. I copied that over into the CE Event [ticket](https://github.com/open-path/Green-River/issues/7485) and added a TODO here for now.

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Okay, sounds good!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Thanks! Hm, I'm also getting confused. You are right that node and step are related, but there is no `node.step`, because node is like the template definition, so the relationship goes in the other direction. 

I guess inside of `process_triggers` we could do (pseudocode/untested)
```
step = instance.steps.find_by(node: node)
```
but that seems a little weird to me, since we already have the step.

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: I added comments here https://github.com/greenriver/hmis-warehouse/pull/5244/files#diff-1f824de3241829773396685de77de358d97b8e1e6b8887195959051542eb412fR7-R10

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: ohh I see! I usually don't worry too much about it because Elliot runs the `maintain_copyright` script every so often 👀  IDE integration sounds like a cool idea though

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
gigxz: Oh ok thanks for explaining! Nested transactions can result in some weird behavior, so I think it makes sense to keep one transaction block at the very highest level (in the mutation).

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
gigxz: thanks!!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
gigxz: Ah of course, I see. So sometimes we want to pass the `step` but sometimes we only want to pass the `node` (which may not even _have_ a step). I think your adjustment to the engine makes sense to me.

gigxz (approved):
Looks good to me

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
eanders: @ttoomey has plans to update the script, but for now, the `# frozen_string_literal: true` should go below the copyright or we'll end up with duplicates.  In the current version of Rails it doesn't care where in the file it is.  I bet your editor can be convinced it's ok somehow...

martha: I see! Okay, I'll correct the `frozen_string_literals`, there are a couple that are above the copyright because chatGPT gave me bad advice 👿  that's what I get for asking a bot instead of asking my coworkers!

ttoomey (approved):
Looks great!

There's kind of a lot of changes in db/warehouse_structure.sql, maybe that's fine, we can always reconcile it with the release branch before merging

PR #5245, opened 2025-03-27
Secure files updates

eanders:
## _Merging this PR_

## Description
* Adds the ability to send to multiple recipients 
* Adds the ability for the warehouse to send a notification to the recipient
* Exposes files you have sent as well as received
* Migrates storage to ActiveStorage while preserving access to old files until the migration has completed.

## Type of change
* New feature

### DIFF
--- a/app/controllers/secure_files_controller.rb
+++ b/app/controllers/secure_files_controller.rb
@@ -48,6 +48,7 @@ class SecureFilesController < ApplicationController
     end
 
     recipients = file_params[:recipients]&.select(&:present?)&.map(&:to_i)
+    send_notifications = file_params[:send_notifications] == '1'
     begin
       @secure_file = file_source.new
       recipients.each do |recipient_id|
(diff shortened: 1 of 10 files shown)


eanders:
Test failures appear unrelated (HMIS related)

app/controllers/secure_files_controller.rb
```diff
@@ -4,22 +4,33 @@
...
+    secure_file = @secure_file.secure_file
+    # Use ActiveStorage version if we have it
+    if secure_file.present?
+      send_data(
+        secure_file.download,
+        type: secure_file.content_type,
+        filename: secure_file.filename.to_s,
+      )
+    else
+      filename = 'secure_file'
```
dtgreiner, line 27: Looks like we're losing the old way of pulling file names. Is this intentional? I don't have an issue with it, but if we have the information, I don't see the harm in showing it.

dtgreiner (approved):
Looks good to me, just one comment but not a blocker by any means

app/controllers/secure_files_controller.rb
eanders: This should only be used for the day or so until old files are migrated to S3.  Once they have been migrated (scheduled task runs each day at 5pm) the filename would begin working again.  Since I removed the CarrierWave attachment logic it was unable to determine the file name.  Seemed like this would probably be ok given the short timeframe and infrequent use of the feature.

PR #5247, opened 2025-03-27
Prevent reopening task that had irreversible side effects

martha:
## _Merging this PR_
- squash-merge this branch into the feature branch `tt-ce-spike-3` after the base branch `martha/7400-referral-side-effects` has been merged

## Description
Summary of changes:
- Add a new classification of side effects, "irreversible" ones. This lives in the MessageHandler for now but I'm very open to suggestions on where it should move.
- Prevent re-opening a step if it was already completed and had an irreversible side effect.
- Add spec for that scenario.

## Type of change
Feature build-out, bugfix


drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -1,9 +1,21 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
 # handle workflow execution messages
 
 module Hmis::Ce
   class ReferralMessageHandler
+    IRREVERSIBLE = [
```
gigxz: Maybe IRREVERSIBLE_MESSAGE_TYPES for a more descriptive name?

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -115,6 +121,17 @@ def visit_node(node, user)
       case node
       when Hmis::WorkflowDefinition::Task
         step = instance.steps.find_or_initialize_by(node: node)
+
+        # If the step has already been completed, it may be re-openable
+        if step.status == 'completed'
+          step_side_effects = step.node.trigger_config&.map { |config| config['message'] } || []
+
+          # but _only_ if it didn't have any irreversible side effects
+          if (Hmis::Ce::ReferralMessageHandler::IRREVERSIBLE & step_side_effects).any? # rubocop:disable Style/IfUnlessModifier
```
gigxz: This is breaking the boundary that the `drivers/hmis/app/models/hmis/workflow_execution` models shouldn't have any CE-specific code.

Is there a way we can encapsulate this further, for example something like `@message_handler.contains_irreversible_triggers?(step)`. That feels a little wonky still, but better than referencing a Ce module from within the engine.

martha: Hm yeah that's a very good point. Thanks for the catch!

I'm proposing `@message_handler.contains_irreversible_messages?(messages)`, which still seems a little messy, but it keeps knowledge of the "step" out of the message handler and CE logic out of the execution engine. Let me know if it looks okay to you.

gigxz: looks good to me, thanks!

gigxz (approved):

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -115,6 +121,17 @@ def visit_node(node, user)
       case node
       when Hmis::WorkflowDefinition::Task
         step = instance.steps.find_or_initialize_by(node: node)
+
+        # If the step has already been completed, it may be re-openable
+        if step.status == 'completed'
+          step_side_effects = step.node.trigger_config&.map { |config| config['message'] } || []
+
+          # but _only_ if it didn't have any irreversible side effects
+          if @message_handler.contains_irreversible_message?(step_side_effects) # rubocop:disable Style/IfUnlessModifier
```
ttoomey: I think this will likely work well.

An alternative would be to return a status object from the message handler; the status could indicate if the message is reversible which could be persisted on the step. That might support situations where reversibility depends on dynamic behavior in the message handler, external systems, etc.

It also could be extended to provide a facility for passing success/error state back to the workflow engine if needed 

```ruby
    def process_triggers(node:, event_type:, user:, step: nil)
      results = []
      node.triggers.each do |trigger|
        next unless event_type == trigger.event

        result =  send_message(
          type: trigger.message,
          params: trigger.params,
          step: step,
          user: user,
        )
        results << result
        log_event('message_sent', event_data: trigger.to_h) if result.success?
      end
      step.update!(reversible: false) unless results.all?(:reversible?)
      results
    end
```

ttoomey (approved):
looks good, thanks!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Thanks! I like this, it does seem more extensible. Let me know if https://github.com/greenriver/hmis-warehouse/pull/5247/commits/60a00a81ff362579758ae8f08997a543fea890c9  is more or less in line with what you had in mind.

Just to think it through, only steps can be marked irreversible (as opposed to gateways and start/end events). I think that makes sense, at least for now. Even though you _can_ attach side effects to other nodes, such as adding a `create_enrollment` trigger on an `end_event`, that end event would not be reversible anyway.

PR #5248, opened 2025-03-28
Strict Client Search Whitespace

dtgreiner:
## _Merging this PR_

## Description

Trim the whitespace in the strict search db query to match the whitespace trimming being done on search field inputs. This will allow any clients with whitespace leading or trailing their first/last name to return in the strict search.

## Type of change
Bug Fix

### DIFF
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -1779,13 +1779,13 @@ module GrdaWarehouse::Hud
 
       if first_name.present?
         first_name_ids = source.where(
-          nf('LOWER', [arel_table[:FirstName]]).eq(first_name.downcase),
+          nf('TRIM', [nf('LOWER', [arel_table[:FirstName]])]).eq(first_name.downcase),
         ).pluck(:id)
       end
 
       if last_name.present?
         last_name_ids = source.where(
-          nf('LOWER', [arel_table[:LastName]]).eq(last_name.downcase),
+          nf('TRIM', [nf('LOWER', [arel_table[:LastName]])]).eq(last_name.downcase),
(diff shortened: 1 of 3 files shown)


app/models/grda_warehouse/hud/client.rb
```diff
@@ -1779,13 +1779,13 @@ def self.strict_search(criteria, client_scope: none)
 
       if first_name.present?
         first_name_ids = source.where(
-          nf('LOWER', [arel_table[:FirstName]]).eq(first_name.downcase),
+          nf('TRIM', [nf('LOWER', [arel_table[:FirstName]])]).eq(first_name.downcase),
```
eanders, line 1782: Do you think we should call `.strip` on the input as well?  Maybe we do that upstream?  
So you can search "Elliot  "?

dtgreiner: The input is being stripped at the beginning of the method:
```
1762:      first_name = criteria[:first_name]&.strip
1763:      last_name = criteria[:last_name]&.strip
```

We probably want to look at cleaning up any records with leading/trailing whitespace too, but this will fix the immediate issue the client is seeing.

eanders: excellent

eanders (approved):

eanders:
Does it make sense to add a test for this?

dtgreiner:
> Does it make sense to add a test for this?

New tests for how whitespace is handled in the client strict and text searches has been added [here](https://github.com/greenriver/hmis-warehouse/pull/5248/commits/743351418222ff33dcd9d8df35ba2792f74b3462)

eanders (approved):

PR #5249, opened 2025-03-28
Save http code for logged involvement requests

martha:
## _Merging this PR_

## Description
**Before**: 
Involvement requests' `http_status` was not saved, so they would be returned by the [failed scope](https://github.com/greenriver/hmis-warehouse/blob/2da46301e94e995992335f6c2f8a437339b7156c/drivers/hmis_external_apis/app/models/hmis_external_apis/external_request_log.rb#L21) even when successful. These scopes are used only for debugging.

**After**:
Involvement requests' `http_status` is saved to the `HmisExternalApis::ExternalRequestLog` as 200 or 400 depending on the involvement's `status_message`.

## Type of change
Bug fix

- [ ] I have run the code that is being changed under ideal conditions, and it doesn't fail - no, I only ran it in the spec test.


gigxz (approved):
thanks!

PR #5254, opened 2025-03-31
7472-spm-m1b-2x

ttoomey:
## _Merging this PR_

## Description
exclude SO from SPM measure 1

## Type of change
Bug fix, 


dtgreiner (approved):
It looks like we may be filtering to only the expected enrollment project types within the `candidate_bed_nights` and `excluded_bed_nights` methods, but possibly missing it when calculating the `pre_entry_dates`. Either way, I like this as a safer way to ensure we have the expected project types as opposed to filtering deeper in the calculation.

eanders (approved):

PR #5263, opened 2025-04-02
CAS analytics data

eanders:
## _Merging this PR_

## Description
* Adds tables and views to hold CAS data transformed into a draft of what we need for OP Analytics
* This needs to be merged before https://github.com/greenriver/boston-cas/pull/927 

## Type of change
* New feature


db/warehouse/migrate/20250331175933_create_cas_analytics_intermediate_tables.rb
```diff
@@ -0,0 +1,88 @@
...
+    create_table :cas_analytics_cas_users do |t|
+      t.string :email
+      t.references :agency, index: false
+      t.string :agency_name
+
+      t.timestamps
+    end
+
+    # Representation of client-opportunity-match combined with route
+    create_table :cas_analytics_workflows do |t|
```
gigxz: In the CE-in-HMIS world, the analog would be "referral" (`ce_referrals` table).. And a "workflow template" is equivalent to "match".
@eanders you mentioned that you would prefer if the same superset dataset was used for both CAS and CE-in-HMIS, perhaps backed by different (or the same?) analytics views. Can we do a quick huddle on this? I don't feel sure what direction we're going with CE data in superset, so I'm not sure how to review this.

db/warehouse/migrate/20250331175933_create_cas_analytics_intermediate_tables.rb
```diff
@@ -0,0 +1,88 @@
...
+      t.boolean :stalled
+      t.string :current_status
+      t.string :terminal_status
+      t.string :unsuccessful_reason
+
+      t.timestamps
+    end
+
+    # Representation of decisions
+    create_table :cas_analytics_steps do |t|
```
gigxz, line 71: note: CE analog would be `wfe_steps` table

db/warehouse/migrate/20250331175933_create_cas_analytics_intermediate_tables.rb
ttoomey: +1, that was my understanding as well.

Perhaps our plan is harmonize CE and CAS analytics views once CE is built? I imagine that building views from the ground up that support both is less likely to incur tech debt

eanders: Yeah, let's huddle on this.  I thought a workflow template was a route and a workflow instance was a match.  I chose to ignore the template for now given that routes aren't currently in the DB and we should be able to generate the parts we need from the matches...  I very much might have these wrong, and it may also be that we want a slightly different shape for analysis than we do for end-user interactions.

gigxz (approved):

db/views/analytics_cas_steps_v01.sql
```diff
@@ -0,0 +1,10 @@
+SELECT id,
+  referral_id,
+  name,
+  "order",
```
ttoomey: Clean this to avoid sql keyword? Maybe
```suggestion
  "order" as sort_order,
```

ttoomey (approved):

PR #5268, opened 2025-04-03
6417 migrate from alpine to debian

ttoomey:
## _Merging this PR_

## Description
trying to move to alpine

## Type of change
Platform change


.github/workflows/container_setup/action.yml
```diff
@@ -36,31 +29,36 @@ runs:
         REPORTING_DATABASE_DB_TEST: reporting_test
         # Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
         PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
-        PUPPETEER_EXECUTABLE_PATH: /usr/bin/chromium-browser
+        PUPPETEER_EXECUTABLE_PATH: /usr/bin/chromium
       run: |
-        apk add --no-cache $(cat .github/dependencies.txt)
+        # Add PostgreSQL 16 repository
```
eanders: I think we're running 17 in production now, so maybe we should shoot for that?

.github/workflows/audit.yml
```diff
@@ -12,10 +12,14 @@ jobs:
   # Label of the container job
   audit:
     # Containers must run in Linux based operating systems
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
+    container: ruby:3.3.8-bullseye
+
+    permissions:
```
yakloinsteak, line 18: We might not need these? Are we pushing or pulling images here?

.github/workflows/container_setup/action.yml
```diff
@@ -1,63 +1,120 @@
...
+          libxslt1-dev \
+          libyaml-dev \
+          nodejs \
+          poppler-utils \
+          postgresql-client-16 \
+          postgresql-client-17 \
+          python3 \
+          python3-pip \
+          python3-venv \
+          tmux \
```
yakloinsteak: Doubt we need tmux on CI.

.github/workflows/container_setup/action.yml
```diff
@@ -69,5 +126,5 @@ runs:
         key: ${{ runner.os }}-node_modules-${{ hashFiles('**/yarn.lock') }}
 
     - name: Install Node.js modules
-      shell: bash
+      shell: ${{ inputs.shell }}
```
yakloinsteak: Why wouldn't this just be bash?

.github/workflows/service_images.yml
```diff
@@ -35,7 +41,7 @@ jobs:
         with:
           username: ${{ secrets.DOCKER_USERNAME }}
           password: ${{ secrets.DOCKER_PASSWORD }}
-        
+
```
yakloinsteak, line 43: Why are we logging into ghcr.io below?

.github/workflows/container_setup/action.yml
```diff
@@ -1,63 +1,120 @@
...
+  using: "composite"
   steps:
-    - name: Install git
-      shell: sh
+    - name: Install dependencies
+      shell: bash
       run: |
-        apk add --no-cache git
+        # Function to retry commands
+        retry() {
```
yakloinsteak, line 24: Nice!

yakloinsteak (approved):
Maybe take a pass at reducing the permissions blocks? Some might be overly permissive for what the action is actually doing. Maybe you already did that.

I have on my long-term radar looking at slimToolkit, but that's beyond the scope of what you're doing here.

.github/workflows/audit.yml
eanders: I'll take another look, I was fighting what I thought was permissions at various points, but ended up being other things.

.github/workflows/container_setup/action.yml
eanders: At this point it probably is!

.github/workflows/service_images.yml
eanders: I don't fully know, but this failure happens when we don't.

eanders: https://github.com/greenriver/hmis-warehouse/actions/runs/15445293888

yakloinsteak: I see. We should probably put that on ECR at some point.

PR #5269, opened 2025-04-03
Add custom ac HMIS Case Note feed

gigxz:
## _Merging this PR_

## Description

Add a new feed for case notes.

How to test:
- I tested in the rails console by running in the console: `export = HmisExternalApis::AcHmis::Exporters::CaseNoteExport.new; export.run!` and writing `export.output` to a csv

## Type of change

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
@@ -1,0 +1,62 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: false
+
+module HmisExternalApis::AcHmis::Exporters
+  class CaseNoteExport
+    include ::HmisExternalApis::AcHmis::Exporters::CsvExporter
+
+    def run!
+      Rails.logger.info 'Generating content of case note export'
+
+      write_row(columns)
+      total = case_notes.count
+
+      Rails.logger.error "There are #{total} addresses to export. That doesn't look right" if total < 10
+
+      case_notes.find_each.with_index do |case_note, i|
+        Rails.logger.info "Processed #{i} of #{total}" if (i % 1000).zero?
+
+        warehouse_id = case_note.client.warehouse_id
+        next unless warehouse_id.present? # Client doesn't have a destination client ID yet. Skip since it wont be in Client.csv anyway.
+
+        values = [
+          warehouse_id, # Matches PersonalID in HMIS CSV Export
+          case_note.enrollment.id, # Matches EnrollmentID in HMIS CSV Export
+          case_note.id,
+          case_note.content,
+          case_note.information_date,
+          case_note.user.id, # Matches User.csv in HMIS CSV Export
+          case_note.date_created,
+          case_note.date_updated,
+        ]
+        write_row(values)
+      end
+    end
+
+    def columns
+      [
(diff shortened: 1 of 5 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
```diff
@@ -0,0 +1,62 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: false
```
eanders: This can probably be `true` for a new file (if not, we should adjust the string/array logic.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
```diff
@@ -0,0 +1,62 @@
...
+    def run!
+      Rails.logger.info 'Generating content of case note export'
+
+      write_row(columns)
+      total = case_notes.count
+
+      Rails.logger.error "There are #{total} case notes to export. That doesn't look right" if total < 10
+
+      case_notes.find_each.with_index do |case_note, i|
+        Rails.logger.info "Processed #{i} of #{total}" if (i % 1000).zero?
```
eanders:
```suggestion
        Rails.logger.info "Processed #{i} of #{total}" if (i % 1_000).zero?
```

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
```diff
@@ -0,0 +1,62 @@
...
+
+        warehouse_id = case_note.client.warehouse_id
```
eanders: `case_notes` isn't joining `client` so unless `case_note.enrollment` goes through a client, you can't count on having the client here, and may want:
```suggestion
        warehouse_id = case_note.client&.warehouse_id
```

drivers/hmis_external_apis/spec/models/hmis_external_apis/ac_hmis/exporters/case_note_export_spec.rb
```diff
@@ -0,0 +1,50 @@
...
+    expect(result.first['PersonalID']).to eq(client.warehouse_id.to_s)
+    expect(result.first['EnrollmentID']).to eq(enrollment.id.to_s)
+    expect(result.first['CaseNoteID']).to eq(case_note_1.id.to_s)
+    expect(result.first['NoteContent']).to eq(case_note_1.content)
+  end
+
+  it 'quotes newlines' do
+    case_note_1.update!(content: "first\nsecond")
+
+    subject.run!
```
eanders, line 43: This should also log an error (< 10 right?)  might be worth adding a test that confirms that so we can be sure we see it in the real world.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/case_note_export.rb
gigxz: Hm good point. It shouldn't happen (since when we set the enrollment we also set the PersonalID; and because the PersonalID column is non-nullable ) but I'll add the safe accessor just in case!

drivers/hmis_external_apis/spec/models/hmis_external_apis/ac_hmis/exporters/case_note_export_spec.rb
gigxz: That log was a carry-over from another export file that todd added. I don't think we really need it (it certainly doesn't alert us) so I just removed it

eanders (approved):

PR #5272, opened 2025-04-04
Default to false for warehouse perm on API

gigxz:
During local testing I observed that `user.can_edit_users?` can return `nil`, which causes an API failure because the field is non-nullable. Update it to default to false.

I am not clear on what situation causes this method to return null. I _think_ it's when the user has **no roles**. I have only seen it in rspec testing, I haven't replicated it in development.

Branch name is for testing https://github.com/greenriver/hmis-frontend/pull/1121

### DIFF
--- a/app/models/hud_reports/report_client_base.rb
+++ b/app/models/hud_reports/report_client_base.rb
@@ -13,56 +13,84 @@ module HudReports
 
     has_many :report_clients, as: :universe_membership, dependent: :destroy
 
-    def display_value(col, pii_policy:, include_content_tag: true)
+    def display_value(col, pii_policy:, include_content_tag: true, cell_val: nil, calculate_cell: true)
+      cell_val = fetch_cell_value(col) if calculate_cell
+
+      return format_complex_value(col, cell_val, pii_policy, include_content_tag) if cell_val.is_a?(Array) || cell_val.is_a?(Hash)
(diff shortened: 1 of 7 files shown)


gigxz:
@eanders maybe this should be fixed higher up the chain somewhere. It does seem odd that `can_<perm>?` would ever return nil, don't you think? Or is that intentional?

eanders:
> @eanders maybe this should be fixed higher up the chain somewhere. It does seem odd that `can_<perm>?` would ever return nil, don't you think? Or is that intentional?

I don't think that's intentional.  I'm surprised it hasn't come up until now.
I believe that _should_ be solvable by changing:


```
    define_method(permission) do
      @permissions ||= load_effective_permissions.merge(load_health_effective_permissions)
      @permissions[permission]
    end
```

to 

```
    define_method(permission) do
      @permissions ||= load_effective_permissions.merge(load_health_effective_permissions)
      @permissions[permission] || false
    end
```

Does that seem like it might work?

gigxz:
@eanders that worked great! I updated it and also added a regression test case https://github.com/greenriver/hmis-warehouse/pull/5272/commits/abd8f420dd2fc7a4fa164b190d5d47fac7de1a59

eanders (approved):
Thanks @gigxz !

PR #5273, opened 2025-04-04
Analytics view for client file

eanders:
## _Merging this PR_

## Description
* Creates views for client files and file tags
* Maintains a `url` column on client files that _should_ point to their location in S3

## Type of change
* New feature

### DIFF
--- a/app/models/grda_warehouse/client_file.rb
+++ b/app/models/grda_warehouse/client_file.rb
@@ -446,8 +446,9 @@ module GrdaWarehouse
             active_storage_url: file.client_file&.blob&.url,
           }
         rescue StandardError
-          # Ignore errors, this is less likely in production
+          # Ignore errors, in development.  We'll revisit in production
           # but if we can't find the url, we don't need to populate it
+          raise unless Rails.env.development?
         end
         upsert_all(batch, update_only: [:active_storage_url], record_timestamps: false) if batch.any?
       end
--- a/app/models/grda_warehouse/client_file.rb
+++ b/app/models/grda_warehouse/client_file.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module GrdaWarehouse
   class ClientFile < GrdaWarehouse::File
     # attr_accessor :requires_expiration_date
@@ -12,7 +14,7 @@ module GrdaWarehouse
     include ClientFileBase
     include ArelHelper
 
-    CONSENT_FORM_TAG_CACHE_KEY = 'consent_form_tagging_ids/tag_ids'.freeze
+    CONSENT_FORM_TAG_CACHE_KEY = 'consent_form_tagging_ids/tag_ids'
 
     mount_uploader :file, FileUploader # This is probably no necessary, but added to be safe
     has_paper_trail
@@ -419,6 +421,24 @@ module GrdaWarehouse
         current_user.id
       else # rubocop:disable Style/EmptyElse
         nil
+      end
+    end
+
+    def self.maintain_urls
+      where(url: nil).find_in_batches do |files|
+        batch = []
+        files.each do |file|
+          next unless file.client_file.attached?
+
+          batch << {
+            id: file.id,
+            url: file.client_file&.blob&.url,
+          }
+        rescue StandardError
+          # Ignore errors, this is less likely in production
+          # but if we can't find the url, we don't need to populate it
+        end
+        upsert_all(batch, update_only: [:url], record_timestamps: false) if batch.any?
       end
     end
 
--- a/app/models/grda_warehouse/client_file.rb
+++ b/app/models/grda_warehouse/client_file.rb
@@ -36,6 +36,17 @@ module GrdaWarehouse
     validates :data_source, presence: true, if: ->(o) { o.confidential? && o.enrollment_id.blank? }
     validates :enrollment, presence: true, if: ->(o) { o.confidential? && o.data_source_id.blank? }
 
+    # If the attached client_file is changed, clear the active_storage_url
+    # The scheduled task will re-populate it as necessary
+    before_save :clear_active_storage_url
+
+    private def clear_active_storage_url
+      # Only set the URL for S3 storage services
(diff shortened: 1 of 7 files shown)


app/models/grda_warehouse/client_file.rb
```diff
@@ -422,6 +424,24 @@ def sync_revokation_info(current_user)
...
+        files.each do |file|
+          next unless file.client_file.attached?
+
+          batch << {
+            id: file.id,
+            url: file.client_file&.blob&.url,
+          }
+        rescue StandardError
+          # Ignore errors, this is less likely in production
+          # but if we can't find the url, we don't need to populate it
```
ttoomey: Is there some way we could end up a with missing files? I imagine a missing file is not expected and indicates something has gone wrong. I imagine we wouldn't want to ignore errors

app/models/grda_warehouse/client_file.rb
```diff
@@ -422,6 +424,24 @@ def sync_revokation_info(current_user)
       end
     end
 
+    def self.maintain_urls
+      where(url: nil).find_in_batches do |files|
```
ttoomey: It seems like this logic is assuming the client_file can never change. Is that a safe assumption? We don't have any enforcement for that (such as a validation).

- option1: Could we just reprocess all files or would that be too expensive? I imagine we could calculate the url and skip the row if it matches. Also maybe it doesn't need to run every hour.
- option2: Add a rails lifecycle event to update the url if the attachment changes
- option3: Add a validation that explodes if the attachment is updated

db/warehouse/migrate/20250403232619_create_analytics_client_files.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class CreateAnalyticsClientFiles < ActiveRecord::Migration[7.0]
+  def change
+    add_column :files, :url, :string
```
ttoomey: This column name is pretty generic and might be confusing. Suggest we make it more descriptive, maybe `exported_analytics_url`

Maybe we add this to `ignored_columns` since it's not intended for use in the rails app.

db/views/analytics_client_files_v01.sql
```diff
@@ -0,0 +1,27 @@
...
+  coalesce(ft.tag_names, '{}') as tags,
+  client_files.data_source_id,
+  client_files.created_at,
+  client_files.updated_at
+from files client_files
+  left join file_tags ft on ft.file_id = client_files.id
+where client_files.type = 'GrdaWarehouse::ClientFile'
+  and client_files.deleted_at is null
+  and client_files.confidential = false
+  and client_files.consent_revoked_at is null
```
ttoomey: is it true that we don't want to export revoked rois?

db/views/analytics_client_files_v01.sql
```diff
@@ -0,0 +1,27 @@
+with file_tags as (
+  select taggings.taggable_id as file_id,
+    array_agg(tags.name) as tag_names
```
ttoomey: Maybe include order and distinct to ensure consistency?  I think the syntax is something like this:
```suggestion
    array_agg(DISTINCT tags.name ORDER BY tags.name) as tag_names
```

db/warehouse/migrate/20250403232619_create_analytics_client_files.rb
eanders: Hmm, so this is nothing special, just the active storage url.  What would you think about calling it `active_storage_url`.  In theory it would work for anyone with the appropriate credentials (in our case S3, but potentially some other backing store would work as well.

db/views/analytics_client_files_v01.sql
eanders: Good point, I was thinking we didn't want to share extra data about clients with revoked consent, but I think we may need to know who has revoked consent in OP analytics eventually.

app/models/grda_warehouse/client_file.rb
eanders: Good thoughts.  The warehouse files are immutable, but I don't think that's true in HMIS.  I'll update.  Probably best to just have it blank out the url field in a before_save hook.

This seemed light-weight enough to run every hour given that people may be adding files at any time and it should only process unprocessed files. I'm happy to move it to the every 4 hours schedule, but it should probably happen more than once a day to ensure we have as many as possible for each daily load to OP Analytics.

app/models/grda_warehouse/client_file.rb
eanders: We will only have missing files if ActiveStorage is unable to generate a url for the file (somehow the blob record was corrupted.)  I added this check since I have tons of terribly broken file data locally (from switching between file back-ends.)  
In production if we hit this, the file is essentially unusable, and for reporting and external data transfer would just be noise.

I'd love to find a more specific error to rescue, but was unable to given the errors I was seeing.

db/warehouse/migrate/20250403232619_create_analytics_client_files.rb
ttoomey: I could be thinking too defensively. I guess the thing is that it's only populated by a scheduled task so could be null; then it feels like I could accidentally call it when I meant file.client_file.url (or something) which is a bit of a foot gun.

Within the application I imagine we should use the active storage api as the single source of truth. Also this field could be hard to grep for if I was trying to figure out how it was used

app/models/grda_warehouse/client_file.rb
ttoomey: I guess if this error were to occur in production, it would indicate an inconsistency between s3 and the db, which sounds bad; my understanding is that even if the object in s3 is garbage, it should still exist.

So if this is only for development, maybe we want:
```
raise unless Rails.env.development?
```

Or if we don't want to risk blocking the export:
```ruby
bad_file_ids = []
files.each do |file|
  #...
rescue
  bad_file_ids << file.id
end
Sentry.capture_message("Invalid Client Files: #{bad_file_ids.slice(0,100).inspect}") if bad_file_ids.any?
```

eanders: I've adjusted to raise if not development.  We'll see how it goes.  If we get a ton of errors we can revisit.

ttoomey (approved):
looks good, thanks!

dtgreiner (approved):

PR #5278, opened 2025-04-07
Expose ROI related fields on client and in ROI authorizations

eanders:
## _Merging this PR_

## Description
* Expose Client ROI authorizations in analytics views
* Expose ROI related fields on client in analytics views

## Type of change
* New feature


db/views/analytics_clients_v03.sql
```diff
@@ -0,0 +1,49 @@
...
+  "DateCreated",
+  "DateUpdated",
+  "UserID",
+  "DateDeleted",
+  "ExportID",
+  "consent_form_id",
+  "housing_release_status",
+  "consent_form_signed_on",
+  "consent_expires_on",
+  "consented_coc_codes"
```
eanders: @ttoomey do you think we should include these (that was the original request) or just expose the ClientRoiAuthorizations to OP analytics?

ttoomey: My recollection is that we anticipate needing to support clients with multiple ROI authorizations. That's not supported by the current fields on Client. So I don't think we should expose those fields if we expect they will be retired in the future.

db/views/analytics_clients_v03.sql
```diff
@@ -0,0 +1,49 @@
...
+  "AfghanistanOEF",
+  "IraqOIF",
+  "IraqOND",
+  "OtherTheater",
+  "MilitaryBranch",
+  "DischargeStatus",
+  "DateCreated",
+  "DateUpdated",
+  "UserID",
+  "DateDeleted",
```
ttoomey: I've been excluding the soft deleted date from views. It's always null anyway

ttoomey (approved):

PR #5280, opened 2025-04-07
Update workflow node to refer to form definition identifier

martha:
## _Merging this PR_

## Description
Summary of changes:
- Update workflow node so it refers to the fd _identifier_ (not an individual form)
- Update workflow step so it optionally stores which form was last used to submit it
- Unrelated change: I reverted the migrations adding 'reversible' to step and 'target_enrollment' to referral. Since this work is all still on a feature branch, I figured it's beneficial to keep everything in the initial migrations for now.
- Update model associations and schema
- Update spec tests

## Type of change
New feature follow-up


drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
```diff
@@ -11,21 +11,25 @@ class Ce::SubmitCeReferralStep < CleanBaseMutation
...
 
-    def resolve(referral_id:, step_id:, input:, **_rest) # add 'confirmed' here later. For now, we ignore it
+    def resolve(referral_id:, step_id:, form_definition_id:, input:, **_rest) # add 'confirmed' here later. For now, we ignore it
       raise unless Hmis::Ce.configuration.enabled?
 
       referral = Hmis::Ce::Referral.viewable_by(current_user).find(referral_id)
       step = nil
       errors = HmisErrors::Errors.new
+      form_definition = Hmis::Form::Definition.find(form_definition_id)
+      raise unless form_definition.published?
```
gigxz: Should this be `raise if draft?` Since we're storing and resolving the form definition by id, I would expect its possible that a retired form may be used to "re"submit a step. (even if the UI doesn't allow it currently.)

drivers/hmis/app/models/hmis/workflow_definition/task.rb
```diff
@@ -5,7 +5,9 @@
 # Tasks may include a swimline which can be used to determine which user is assigned to the corresponding step
 module Hmis::WorkflowDefinition
   class Task < Node
-    belongs_to :form_definition, class_name: 'Hmis::Form::Definition'
+    # Similar to Hmis::Form::Instance, Task has a form _identifier_, not a specific form, so that referral workflow steps can use newly published form versions
+    has_many :definitions, primary_key: :form_definition_identifier, foreign_key: :identifier, class_name: 'Hmis::Form::Definition'
```
gigxz: I liked calling this `form_definition[s]` rather than `definitions` to disambiguate from other kinds of definitions (ie workflow definition)

drivers/hmis/app/graphql/mutations/ce/submit_ce_referral_step.rb
martha: Yes sounds good, we can use the existing `valid_status_for_submit?` helper

gigxz: perfect!

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
```diff
@@ -28,7 +28,7 @@ def change
       t.string :name
       t.references :swimlane, foreign_key: { to_table: :wfd_swimlanes }
       # task nodes have forms.
-      t.references :form_definition, foreign_key: { to_table: :hmis_form_definitions }
+      t.string :form_definition_identifier
       # gateway nodes have types
       t.string :gateway_type
```
gigxz, line 34: Changes look good to me, just noting I had to split out "up" and "down" functions in order to properly rerun this migration locally, in case others need to do the same:
```
  def down
    execute 'DROP TABLE IF EXISTS wfe_audit_events CASCADE'
    execute 'DROP TABLE IF EXISTS wfe_step_assignments CASCADE'
    execute 'DROP TABLE IF EXISTS wfe_steps CASCADE'
    execute 'DROP TABLE IF EXISTS wfe_instances CASCADE'
    execute 'DROP TABLE IF EXISTS wfd_flows CASCADE'
    execute 'DROP TABLE IF EXISTS wfd_nodes CASCADE'
    execute 'DROP TABLE IF EXISTS wfd_swimlanes CASCADE'
    execute 'DROP TABLE IF EXISTS wfd_templates CASCADE'
  end
```

gigxz (approved):

db/warehouse/migrate/20241207185500_create_hmis_workflows.rb
gigxz: maybe its worth adding this into the migration so its reversible for everyone

PR #5281, opened 2025-04-08
Add form definition reference to analytics_client_geolocations view

gigxz:
## _Merging this PR_

## Description

see: [superset dataset spec](https://docs.google.com/spreadsheets/d/1qiHcgfRGEBaw-e-crJXbbIoecj1HhO3W1xjdHOAqJeU/edit?gid=0#gid=0)

1. Add `form_definition_id` to geolocation view, which will allow users to see the form that collected the location in the location dataset. For example you could filter to locations collected on "2025 PIT Form" or collected on "Current living situation"
2. Add owner_type/owner_id to geolocation view. Per spreadsheet, I don't think these should necessarily be exposed on the location dataset itself,  but rather this view can be used to ADD lat/lon to the existing custom assessment & CLS data sets.

## Type of change
new feature


db/views/analytics_client_geolocations_v02.sql
```diff
@@ -0,0 +1,33 @@
...
+    -- legacy field
+    -- clh_locations.enrollment_id,
+    clh_locations.located_at,
+
+    -- If location was collected via an HMIS form, id of the form definition
+    hmis_form_processors.definition_id as form_definition_id,
+    -- If location was collected via HMIS form, type of the associated record
+    -- (eg "HmisExternalApis::ExternalForms::FormSubmission", "Hmis::Hud::CustomAssessment" or "Hmis::Hud::CurrentLivingSituation")
+    hmis_form_processors.owner_type,
+    hmis_form_processors.owner_id -- ID of associated record
```
gigxz, line 26: These 3 fields and the join are new. The rest is unchanged, aside from prefixing field names with `clh_locations.`!

db/structure.sql
eanders, line 1: I think this file can be removed from the PR

eanders (approved):
Looks good thanks!

db/views/analytics_client_geolocations_v02.sql
ttoomey: sorry I'm probably missing context but why are we prefixing the col names? My understanding is the analytics views should abstract away these kinds of details. Or is there a reason the consumer of the view needs to know the source tables involved in the wh?

gigxz: When I run this in sql, the resulting column names are not prefixed. So this shouldn't affect the "consumer of the view" at all. The prefix just tells postgres which table to look at. I needed to add them (I think) because I added a join to another table

db/views/analytics_client_geolocations_v02.sql
```diff
@@ -0,0 +1,33 @@
...
+
+FROM
+    public.clh_locations
+LEFT OUTER JOIN public.hmis_form_processors
+    ON hmis_form_processors.clh_location_id = clh_locations.id
```
ttoomey, line 31: Could this create duplicates? Looks like there's no uniqueness constraint on `hmis_form_processors.clh_location_id`

db/views/analytics_client_geolocations_v02.sql
ttoomey: Oh jeez I misread it entirely, sorry about that! 🤦

ttoomey (approved):

db/views/analytics_client_geolocations_v02.sql
gigxz: Good point. It wouldn't be possible to achieve through the application (there is no way to link a second form processor to an existing location), but you're right the db doesn't constrain it. We should probably add uniqueness constraints for some/all of the references on the form processor table.

This doesn't seem worth adding a subquery `SELECT DISTINCT ON` since it's not actually something we expect, I would rather adjust constraints on the form processor, what do you think?

ttoomey: Totally, let's just add a unique constraint to form-processor rather than changing the view. We could track it as a tech debt issue to keep the scope down here (and allow for QA)

PR #5283, opened 2025-04-08
Performance adjustments

eanders:
## _Merging this PR_

## Description
* This is an incomplete fix but contains some performance improvements to try and prevent generating excessive data and potentially adjust indexing or batching
* Rework IdentifyDuplicates to push calculations into the database
* Prevent building service history too far back in the past

## Type of change
* Code clean-up
* Performance Updates


app/models/grda_warehouse/hud/client.rb
eanders, line 1: These changes are mostly (all?) around logging for debugging.  What do you think about leaving them here?

dtgreiner: I don't see any issue with keeping the logging in here. The only thing I can think of is if this method were being called a bunch it could possibly flood the logging, but I don't believe that's the case with this split method.

dtgreiner (approved):

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -115,111 +131,511 @@ def identify_duplicates
...
+    private def find_merge_candidates_for_unprocessed
+      # Never return obvious matches if auto deduplication is disabled
+      unless GrdaWarehouse::Config.get(:enable_auto_deduplication)
+        @logger.info 'Auto deduplication disabled, returning empty set'
+        return {}
+      end
+
+      # Find all potential matches
+      matches = {}
+      if GrdaWarehouse::Config.get(:enable_auto_deduplication)
```
ttoomey: looks like this may be redundant with 531

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -115,111 +131,511 @@ def identify_duplicates
...
+          left outer join warehouse_clients on (clients.id = warehouse_clients.source_id OR clients.id = warehouse_clients.destination_id)
+          #{limits}
+        ),
+        client_two_name as (
+        SELECT clients.id as client_two_name_id,
+          concat(regexp_replace(lower(trim(clients."FirstName")), '[^a-z0-9]', '', 'g'), '_',
+          regexp_replace(lower(trim(clients."LastName")), '[^a-z0-9]', '', 'g')) as client_two_name_name
+          from "Client" as clients
+          #{limits}
+          and clients.id in (#{unprocessed.select(:id).to_sql})
```
ttoomey: so `unprocessed()` looks like "select id from (list of ids)" - just checking that's intentional; maybe it helps the query planner. The more direct thing would be to replace this with an array literal. Or change the implementation of `unprocessed`, for example
```ruby
private def unprocessed
  GrdaWarehouse::Hud::Client
    .where(id: GrdaWarehouse::Hud::Client.source.select(:id))
    .where.not(id: GrdaWarehouse::WarehouseClient.select(:source_id))
end
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -115,111 +131,511 @@ def identify_duplicates
...
+      limits = <<-SQL
+        where clients."DateDeleted" is NULL
+        and (clients."FirstName" is not null and clients."FirstName" != ''
+        or clients."LastName" is not null and clients."LastName" != '')
+      SQL
+      results = GrdaWarehouse::Hud::Client.connection.execute(<<-SQL)
+        with client_one_name as (
+        SELECT coalesce(warehouse_clients.destination_id, clients.id) as client_one_name_id,
+          concat(regexp_replace(lower(trim(clients."FirstName")), '[^a-z0-9]', '', 'g'), '_',
+          regexp_replace(lower(trim(clients."LastName")), '[^a-z0-9]', '', 'g')) as client_one_name_name
```
ttoomey: Small optimization, probably for future work: looking at this again, this replaces accented chars with underscore. I wonder using [unaccent](https://www.postgresql.org/docs/current/unaccent.html) would provide more accurate matches. 
```
unaccent(lower(trim(clients."FirstName")))
```
Also maybe we should remove quote chars so these match: "O'Neil" -> "ONeil"

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -115,111 +131,511 @@ def identify_duplicates
...
+        more_than_one_match.delete(sorted_row)
+      end
+      candidates = group_merge_chains(more_than_one_match)
+
+      source_client_counts = GrdaWarehouse::WarehouseClient.group(:destination_id).count
+      split_chains_on_max_source(candidates: candidates, counts: source_client_counts)
+    end
+
+    # Find potential client merge candidates by looking for matches across multiple criteria
+    # @return [Hash] A set of client ID pairs that are candidates for merging
```
ttoomey: I think this is incorrect
```suggestion
    # @return [Array<Array<Integer>>] Array of [destination_id, source_id] pairs that are candidates for merging
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -115,111 +131,511 @@ def identify_duplicates
...
+    # @note Only returns matches if auto deduplication is enabled in config
+    private def find_merge_candidates_for_unprocessed
+      # Never return obvious matches if auto deduplication is disabled
+      unless GrdaWarehouse::Config.get(:enable_auto_deduplication)
+        @logger.info 'Auto deduplication disabled, returning empty set'
+        return {}
```
ttoomey: always return an array
```suggestion
        return []
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
         batch.each do |client|
-          match_id = check_for_obvious_match(client)
-          if match_id.present?
-            matched += 1
-            matched_ids << match_id
-            destination_client = get_destination_client_by_id(match_id)
+          destination_client = nil
+          if found_ids.include?(client.id)
+            matched_pair = more_than_one_match.find { |pair| pair.include?(client.id) }
+            destination_id = (matched_pair - [client.id]).first # We don't know the order of the pair
```
ttoomey: This finds only the first matching pair. But there could be many?

Also maybe an opportunity for further optimization by building a two-way lookup 
```ruby
# outside of loop, build lookup
lookup = Hash.new { |h, k| h[k] = Set.new }
more_than_one_match.each d |a, b| 
  lookup[a] << b
  lookup[b] << a
end
# in loop
destination_ids = lookup[client.id]
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
-            end
-            if should_save
-              # set non-nullable fields, these aren't used because of the column limitation on import
-              # but Postgres complains if they aren't there
-              destination_client[:PersonalID] = client.PersonalID
-              destination_client[:data_source_id] = @dnd_warehouse_data_source.id
-              destination_client_updates << destination_client
-            end
+            destination_client.dob = client.dob if client.dob.present? && destination_client.dob.blank?
+            destination_client.ssn = client.ssn if client.ssn.present? && destination_client.ssn.blank?
```
ttoomey: we set dob and ssn but not name, that's intentional right?

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
       new_created = 0
+
+      # Find all potential matches
+      more_than_one_match = find_merge_candidates_for_unprocessed
+      found_ids = more_than_one_match.flatten.uniq.to_set
+
+      # At this point more_than_one_match contains pairs of destination client id and source client id where the
+      # source client should be merged into the destination client.
+      # If the source client does not exist in any of the pairs, a new destination client should be created for it.
+      destination_clients_by_id = GrdaWarehouse::Hud::Client.where(id: found_ids.to_a).index_by(&:id)
```
ttoomey: restrict for extra safety?
```suggestion
      destination_clients_by_id = GrdaWarehouse::Hud::Client.destination.where(id: found_ids.to_a).index_by(&:id)
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
-              destination_client[:data_source_id] = @dnd_warehouse_data_source.id
-              destination_client_updates << destination_client
-            end
+            destination_client.dob = client.dob if client.dob.present? && destination_client.dob.blank?
+            destination_client.ssn = client.ssn if client.ssn.present? && destination_client.ssn.blank?
+            # set non-nullable fields, these aren't used because of the column limitation on import
+            # but Postgres complains if they aren't there
+            destination_client.personal_id = client.personal_id
+            destination_client.data_source_id = @dnd_warehouse_data_source.id
+            destination_client_updates << destination_client
```
ttoomey: we're not always updating the record right?
```suggestion
            destination_client_updates << destination_client if destination_client.changed?
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
           else
             new_created += 1
             destination_client = client.dup
+            destination_client.id = nil
```
ttoomey: surprised this is needed, dup is supposed to blank the pk

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
       # If we do find a match:
       #   create the associated WarehouseClient
 
-      Rails.logger.info "Matching #{unprocessed.count} unprocessed clients"
+      unprocessed_count = unprocessed.count
+      Rails.logger.info 'Matching #unprocessed_count} unprocessed clients'
```
ttoomey: wrong quote type for interpolation
```suggestion
      Rails.logger.info "Matching #unprocessed_count} unprocessed clients"
```

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -44,55 +45,67 @@ def identify_duplicates
...
         matched_ids = []
         destination_client_updates = []
         new_destination_clients = []
         new_warehouse_clients = {}
         source_client_ids_with_new_destination_clients = []
+
         batch.each do |client|
-          match_id = check_for_obvious_match(client)
-          if match_id.present?
-            matched += 1
```
ttoomey, line 59: If we remove this then it looks like matched isn't incremented anymore? That seems incorrect

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -28,9 +30,8 @@ def run!
     end
 
     def identify_duplicates
```
ttoomey, line 41: maybe summarize what this method does
```suggestion
    # Deduplicates client records by:
    # 1. Restoring any previously deleted destination clients still referenced.
    # 2. Loading all unprocessed clients (not yet linked in warehouse_clients).
    # 3. For each unprocessed client:
    #    - If a match is found among existing destinations, update the destination and link it.
    #    - If no match, create a new destination client and link it.
    # 4. Invalidates service history for affected clients.
    # 5. Finalizes (accepts) any exact matches that were previously proposed
    # 6. Logs the deduplication run.
    def identify_duplicates
```

spec/models/grda_warehouse/tasks/service_history/enrollment_spec.rb
```diff
@@ -0,0 +1,166 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https: //github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
```
ttoomey:
```suggestion
# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
```

ttoomey (approved):
looks great! Left some comments, questions, and a few minor corrections. Also notes for additional optimization.

app/models/grda_warehouse/tasks/identify_duplicates.rb
ttoomey: note: above is probably not super efficient. A more efficient version might be:
```ruby
GrdaWarehouse::Hud::Client
  .source
  .left_outer_joins(:warehouse_client_source)
  .where(warehouse_client_source: { id: nil })
  .distinct
  ```

File: app/models/grda_warehouse/tasks/identify_duplicates.rb (hunk 3 above)
Comment (eanders) [reply app/models/grda_warehouse/tasks/identify_duplicates.rb]: I might be missing something, but I think this is replacing all non numeric non alpha chars with an empty string.  Then we concat first and last name with and underscore.

Unaccent looks like a good improvement, and appears to be installed.  Thanks!

File: app/models/grda_warehouse/tasks/identify_duplicates.rb (hunk 7 above)
Comment (eanders) [reply app/models/grda_warehouse/tasks/identify_duplicates.rb]: Hmm, not intentional. I was just porting the existing logic to the new format.  These will be set eventually in ClientCleanup, but I guess no harm adding name if it isn't there.

File: app/models/grda_warehouse/tasks/identify_duplicates.rb (hunk 8 above)
Comment (eanders) [reply app/models/grda_warehouse/tasks/identify_duplicates.rb]: That scope seems to break the tests.

File: app/models/grda_warehouse/tasks/identify_duplicates.rb (hunk 10 above)
Comment (eanders) [reply app/models/grda_warehouse/tasks/identify_duplicates.rb]: I was fighting ids, I think this may be unneeded and was an attempt to solve something unrelated.

File: app/jobs/client_cleanup_job.rb [hunk 15]
```diff
@@ -0,0 +1,46 @@
...
+  def perform(ids)
+    running_jobs = Delayed::Job.jobs_for_class('ClientCleanupJob').running.count
+    if running_jobs > 1
+
+      requeue_job
+      return
+    end
+
+    GrdaWarehouse::Tasks::ClientCleanup.run_for_clients(ids)
+  end
```
ttoomey: consider if we can just synchronize with and advisory lock to avoid race conds
```suggestion
  def perform(ids)
    lock_obtained = nil
    lock_name = self.class.name.to_s
    with_advisory_lock(lock_name, timeout_seconds: 0) do
      lock_obtained = true
      GrdaWarehouse::Tasks::ClientCleanup.run_for_clients(ids)
    end
    requeue_job unless lock_obtained
  end
```

app/jobs/client_cleanup_job.rb
```diff
@@ -0,0 +1,46 @@
...
+
+  private def requeue_job
```
ttoomey: should we use `requeue_job` from BaseJob?
```ruby
  def requeue_job
    requeue_at(Time.current + WAIT_MINUTES.minutes, "ClientCleanupJob already running...re-queuing job for #{WAIT_MINUTES} minutes from now")
  end
```

eanders: YES!  Sorry I missed this, I had pulled from another job that had some differences.  I'll update.

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -112,114 +144,546 @@ def identify_duplicates
...
+      SQL
+      results.map { |r| [r['destination_one_id'], r['destination_two_id']] }
+    end
+
+    # Finds pairs of destination and unprocessed source client IDs with exact SSN matches.
+    # Expands the search to any client, even those without a warehouse_client record.
+    # Only returns pairs where the SSN is valid and matches.
+    #
+    # @return [Array<Array<Integer>>] Array of [destination_client_id, source_client_id] pairs
+    def exact_ssn_matches_for_unprocessed
```
ttoomey, line 388: should we use SSNDataQuality (and name/dob dq fields)? Wondering if we would want to exclude partial or incomplete data

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -112,114 +144,546 @@ def identify_duplicates
...
+      # candidates now contains a hash keyed on client ids with arrays of destination client ids that will be
+      # used as the source for merges into the key client id.
+      # At this point the `key => ids` combinations have been grouped so any merges that would have
+      # required multiple iterations are all in the array associated with the destination client,
+      # all merges have been checked to confirm they won't exceed the max allowable source clients, and if
+      # they would have exceeded the limit, a new group has been added
+      candidates = find_merge_candidates_for_match_existing
+      @logger.info "Found #{candidates.size} merge candidates"
+
+      candidates.each_slice(500) do |batch|
```
ttoomey, line 186: a little unusual to call each_slice on a hash but it seems to work as intended

ttoomey (commented):
Looks good to me!!

app/models/grda_warehouse/tasks/identify_duplicates.rb
eanders: Historically those have been really inaccurate, but potentially we could exclude those that have indicated partial or not collected I guess.  That said, I don't know why you would have a valid SSN if the quality is not collected or partial.

PR #5285, opened 2025-04-08
Replace activeReferral and acceptedReferral with referral

martha:
## _Merging this PR_

## Description
While working on https://github.com/open-path/Green-River/issues/7417, I was finding our schema differentiation of an `opportunity`'s `acceptedReferral` and `activeReferral` unwieldy. This PR proposes replacing those fields with one field, `referral`, which could be
- an active referral if there is one
- an accepted referral if there is one
- null, if there is no referral or only rejected referrals (meaning the opportunity is currently open to accepting new referrals.)

The frontend can then distinguish between in-progress and accepted referrals using the `status` field. I think this makes for more graceful schema and frontend code (linked PR: https://github.com/greenriver/hmis-frontend/pull/1127/files). Let me know if you think it's an okay change! 

## Type of change
Code clean-up


drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb
```diff
@@ -194,16 +194,46 @@
...
     end
 
+    describe 'when the opportunity has several referrals' do
+      # opportunities are single-use, so there should only be one in-progress or accepted referral, but there could be many failed referrals.
+      let!(:rejected1) { create(:hmis_ce_referral, opportunity: opportunity, status: 'rejected', created_at: 1.day.ago) }
+      let!(:rejected2) { create(:hmis_ce_referral, opportunity: opportunity, status: 'rejected', created_at: 1.day.ago) }
+      let!(:rejected3) { create(:hmis_ce_referral, opportunity: opportunity, status: 'rejected', created_at: 1.day.ago) }
+
+      ['initialized', 'in_progress', 'accepted'].each do |status|
+        it 'returns the correct referral' do
```
gigxz: suggest adding status into test name so test case name isnt repeated:
```suggestion
        it "returns the #{status} referral" do
```

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb
```diff
@@ -194,16 +194,46 @@
...
+          response, result = post_graphql(**variables) { query }
+          expect(response.status).to eq(200), result.inspect
+          referral_data = result.dig('data', 'ceOpportunity', 'referral')
+          expect(referral_data).to include('id' => referral.id.to_s, 'status' => status)
+        end
+      end
+
+      context 'when there are multiple active referrals, which should not happen' do
+        let!(:referral1) { create(:hmis_ce_referral, opportunity: opportunity, status: 'in_progress', created_at: 2.days.ago) }
+        let!(:referral2) { create(:hmis_ce_referral, opportunity: opportunity, status: 'accepted', created_at: 3.days.ago) }
```
gigxz: How about adding a validation constraint for this on the AR model? I am hesitant to add a constraint to the db since there is a _chance_ we may want to allow multiple concurrent referrals in the future. But we did decide for MVP that a single active referral was appropriate.

If we had an AR constraint, the creation of `referral2` could raise a validation error like "opportunity already has a referral with an active status (initialized, in_progress, or accepted)". I think that would be preferable because it will help us catch this behavior if it occurs.

gigxz (approved):
API changes lgtm! I think its worth adding an explicit constraint as well.

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb
martha: Thanks, yeah good idea! I added an ar validation [here](https://github.com/greenriver/hmis-warehouse/pull/5285/commits/b891920f28bb52632d1ce6aa66873e2da6541542). I'll wait to see if that messed up any spec tests before merging.

PR #5288, opened 2025-04-09
Remove PITCount from PIT report, hide entirely for OP HMIS projects

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7497

* Remove `PITCount in HMIS` column from the 2024 PIT. This was causing confusion. It is often unused or used only for non-hmis-participating projects.
* Hide PITCount on the project page for HMIS-managed projects, since we dont maintain the field so it will be out of date. For imported projects, continue to show it and add a tooltip.
* Fix tables that used `table-hover` and cursor pointer even though the rows are not clickable. (accessibility fix)

## Type of change

### DIFF
--- a/drivers/hud_pit/app/models/hud_pit/generators/pit/fy2024/projects.rb
+++ b/drivers/hud_pit/app/models/hud_pit/generators/pit/fy2024/projects.rb
@@ -4,6 +4,11 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: false
+
+# Project-level PIT counts.
+#
+# This is NOT part of the HUD PIT reporting requirements; we do not expect them to be uploaded to HDX.
 module HudPit::Generators::Pit::Fy2024
   class Projects < Base
     QUESTION_NUMBER = 'Projects'.freeze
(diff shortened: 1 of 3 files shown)


drivers/hud_pit/app/models/hud_pit/generators/pit/fy2024/projects.rb
```diff
@@ -4,6 +4,11 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: false
```
gigxz: so is this what I am supposed to do when I get the error
> /Users/gig/dev/op/hmis-warehouse/drivers/hud_pit/app/models/hud_pit/generators/pit/fy2024/projects.rb:14:23: C: [Correctable] Style/RedundantFreeze: Do not freeze immutable objects, as freezing them has no effect.

Or should I instead remove the `.freeze` on line 14? 

(sorry wasn't playing close attention when we introduced the new freeze behavior)

eanders: You can just remove the `.freeze` and set the above to true.  Or do what you did, that's fine too, but is more of a band-aid approach.

eanders (approved):
Thanks @gigxz !

dtgreiner (approved):

PR #5290, opened 2025-04-10
Stricter test for readiness vs liveness

yakloinsteak:
## Description

When booting a pod, it makes sense to fail if the database can't be reached so that we don't replace a healthy set of pods with those that can't talk to the database. This bit me a couple of times during the RDS upgrades.

This puts the endpoint on a standard path so we can eventually do this in other projects.

## Type of change

Stability improvement


yakloinsteak:
@dtgreiner This came about because if you deploy a bad database config to ECS, it just boots up and replaces the web app containers with crash looping. I've hit this two or three times during the replication upgrades. I don't know if we can use this for ECS, but it will definitely help on EKS.

dtgreiner (approved):

PR #5294, opened 2025-04-11
Support Client Referrals page

martha:
## _Merging this PR_

## Description

Corresponds with frontend PR https://github.com/greenriver/hmis-frontend/pull/1133

Summary of changes:
- Opportunity schema object
  - Add `project`, and deprecate `projectId` and `projectName`. This allows the client to resolve `projectType`.
  - Add Project and Project Type to available filter options.
- Referral schema object
  - Add new fields `target_project` and `referred_by`
  - Add Project and Project Type to available filter options
- Client schema object
  - Resolve filterable eligible_ce_opportunities
  - Resolve filterable ce_referrals
- Project schema object - Should be no functional change:
  - Update filterable fields (ce_opportunities, ce_referrals) so that they omit project-related filters, since it's irrelevant - we are already querying on the project.

## Type of change
New feature

  - [n/a] ensured permissions and visibility checks are performed in the right places - not yet, todo for #7395.
- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - yes, see frontend PR


drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -83,14 +83,38 @@ def self.authorized?(object, ctx)
...
 
-    def eligible_ce_opportunities
+    field :eligible_ce_opportunities, Types::HmisSchema::CeOpportunity.page_type, null: false do
+      # Omit status, since we only return open opportunities for the client anyway
+      filters_argument Types::HmisSchema::CeOpportunity, omit: [:status], type_name: 'ClientEligibleCeOpportunity'
+    end
+    def eligible_ce_opportunities(filters: nil)
+      raise unless Hmis::Ce.configuration.enabled?
+
+      scope = Hmis::Ce::Opportunity.for_client(object)
```
ttoomey: suggest wiring up the visibility scope
```suggestion
      scope = Hmis::Ce::Opportunity.viewable_by(current_user).for_client(object)
```

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -83,14 +83,38 @@ def self.authorized?(object, ctx)
...
+    field :ce_referrals, Types::HmisSchema::CeReferral.page_type, null: false do
+      filters_argument Types::HmisSchema::CeReferral
+    end
+    def ce_referrals(filters: nil)
       raise unless Hmis::Ce.configuration.enabled?
 
-      Hmis::Ce::Opportunity.
-        for_client(object).
-        order(:id)
+      scope = object.ce_referrals
```
ttoomey: suggest wiring up the visibility scope
```suggestion
      scope = object.ce_referrals.viewable_by(current_user)
```

ttoomey (approved):
looks good, thaks!

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -14,14 +14,17 @@ class HmisSchema::CeOpportunity < Types::BaseObject
     field :expires_at, GraphQL::Types::ISO8601DateTime, null: true
     field :referral, Types::HmisSchema::CeReferral, null: true, description: 'Active or accepted referral'
     field :candidates, Types::HmisSchema::CeCandidate.page_type, null: false
-    field :project_id, ID, null: false
-    field :project_name, String, null: false
+    field :project, Types::HmisSchema::Project, null: false
+    field :project_id, ID, null: false, deprecation_reason: 'Deprecated in favor of project'
+    field :project_name, String, null: false, deprecation_reason: 'Deprecated in favor of project'
```
gigxz: Will the user resolving the opportunity always have can_view_project visibility at the project? The Project object has an authorization guard to ensure that the user has `can_view_project`, and will raise if they don't. So, if there _is_ a use-case for resolving minimal project details on the opportunity without full project access, that would be a good reason to keep minimal fields like `projectId`/`projectName`/`projectType`.

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -83,14 +83,38 @@ def self.authorized?(object, ctx)
...
-        for_client(object).
-        order(:id)
+      scope = object.ce_referrals.viewable_by(current_user)
+      scope = scope.where(status: filters&.status) if filters&.status.present?
+
+      opportunity_table = Hmis::Ce::Opportunity.arel_table
+      scope = scope.joins(:opportunity).where(opportunity_table[:project_id].in(filters&.project)) if filters&.project.present?
+
+      projects_table = Hmis::Hud::Project.arel_table
+      scope = scope.joins(opportunity: :project).where(projects_table[:project_type].in(filters&.project_type)) if filters&.project_type.present?
```
gigxz: suggestion: add `include ::Hmis::Concerns::HmisArelHelper` at the top of the class, match convention elsewhere in the codebase:
```suggestion
      scope = scope.joins(opportunity: :project).where(p_t[:project_type].in(filters&.project_type)) if filters&.project_type.present?
```

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -83,14 +83,38 @@ def self.authorized?(object, ctx)
...
     field :email_addresses, [HmisSchema::ClientContactPoint], null: false
     field :hud_chronic, Boolean, null: true, description: 'Meets the definition for HUD chronically homeless as of today (time of API request)'
-    field :eligible_ce_opportunities, Types::HmisSchema::CeOpportunity.page_type, null: false
 
-    def eligible_ce_opportunities
+    field :eligible_ce_opportunities, Types::HmisSchema::CeOpportunity.page_type, null: false do
+      # Omit status, since we only return open opportunities for the client anyway
+      filters_argument Types::HmisSchema::CeOpportunity, omit: [:status], type_name: 'ClientEligibleCeOpportunity'
+    end
+    def eligible_ce_opportunities(filters: nil)
```
gigxz: Suggest moving these resolver functions down. Convention elsewhere on graphql types is to list ALL `field`s first, and then any functions below

gigxz (approved):
lgtm! lets retarget to 161 though. one open question about the opportunity api change, but happy to leave it as-is for now. I don't think we have an immediate use case for resolving opportunities info without project access, so we can leave it strict for now.

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
martha: Oh, thanks for pointing that out, let me try to think it through. The Client Eligible Opportunities page should only return an opportunity if the user has `can_view_opportunities` (or `can_view_units` if we decide to reuse that) in the opportunity's project. However, a user could have `can_view_opportunities` but NOT `can_view_project`. Same with resolving a client's current referrals -- I guess a user may have `can_view_referrals` without `can_view_project`.

I'm not sure if we really have a use case for either of those permission profiles. But those use cases arising in the future seems somehow more likely to me than the use case of needing to resolve more fields on the project (besides just name/id/type). So maybe it would be better to design the API to leave off the `project` field, and replace it with individual name/id/type fields.

ttoomey (approved):
Looks good, thanks!

PR #5236, opened 2025-03-24
Family pathways updates

eanders:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

## Description
[//]: # (Summarize changes and include links related issue)
[//]: # (List any new dependencies or relevant ADRs)
* Adds calculations for https://github.com/open-path/Green-River/issues/7449
* Brings in [AR Lazy Preloader](https://github.com/DmitryTsepelev/ar_lazy_preload) to work around a bug where empty preloaded associations aren't getting cached.

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)
* New feature

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail

PR #5295, opened 2025-04-13
7467 rails 7.1 upgrade

ttoomey:
## _Merging this PR_

### Rails Framework Updates for 7.1
* Use rails 7.0 framework defaults for now and overlay safe 7.1 config (see new_framework_defaults_7_1.rb)
* Although config/secrets are deprecated, added a work around so we can continue using it (in application.rb)
* Replaced composite_primary_keys with rails built-in, changed foreign_key to query_constraints in associations

### Stricter defaults and deprecations
* Added/updated several HUD structure model classes
* Replaced alias_attribute with alias_method in several places
* Adjusted rubocop to prefer alias_method over alias
* Updated model serialization to use explicit typing (`serialize :data, type: Hash`)

### Database compatibility
* Added primary_abstract_class declaration in ApplicationRecord
* Added a CustomApplicationRecord module to share behavior between application models without mixing db connection info
* Fixed acts-as-taggable db config issue
* Removed schema_migration classes for each database as they seemed to be incompatible with 7.1
*  Updated database initialization logic to use helper methods

### Security
* Improved rack-attack test and configuration
* Addressed brakeman issues

### Misc
* Discovered the hmis_supplemental_field_values.data was using the wrong serialization method, added migration to correct it (I don't think this feature is in use yet)
* Improve reliability of system tests
* added preloader tests to verify our usage
* removed calls to reset in preloader (https://github.com/greenriver/hmis-warehouse/pull/5302)

### Dependency Updates
* ⚠️ Breaking or major changes
  * removed composite_primary_keys 
  * haml upgraded from 5 to 6.3. This has major implications for escaping interpolated code in :javascript blocks
* 👀 Important changes that needs testing
  * graphql upgraded from 2.3.21 to 2.5.4
  * Rack 2 to 3
  * activerecord-postgis-adapter upgraded from 8 to 9
  * acts-as-taggable-on upgraded from 11 to 12
  * business_time upgraded from 0.10.0 to 0.13.0 (why was this pinned to an old version?)
  * byebug upgraded from 11.1.3 to 12.0.0
  * removed the (almost) unused shoulda gem
  * upgraded shoulda-matchers for compatibility

Todo:
- [ ] Ensure early return from transaction generates a deprecation warning in prod/staging to sentry (check https://[warehouse]/admin/deprecation?test=test_deprecated_tx_break)
- [x] Verify brakeman's xss handling with haml 6 escaping behavior
- [x] Verify that rails alerts engineers to missing migrations in all databases in development mode
- [x] Double check enum values are correct in vispidat
- [x] Double check serialization change and migration on HmisSupplemental::FieldValue

## Type of change
Dependency update


app/controllers/warehouse_reports/cas/decision_efficiency_controller.rb
```diff
@@ -75,33 +77,25 @@ def median(array)
       @cas_c_t ||= CasAccess::Client.arel_table
     end
 
-    private def at2
```
ttoomey: review needed

app/controllers/warehouse_reports/cas/process_controller.rb
```diff
@@ -79,29 +81,50 @@ def step_times(step_range)
       else
         raise "unanticipated time unit: #{unit}"
       end
-      at = GrdaWarehouse::CasReport.arel_table
-      at2 = at.dup
-      at2.table_alias = 'at2'
```
ttoomey: review needed

app/models/application_record.rb
```diff
@@ -1,39 +1,13 @@
+# frozen_string_literal: true
+
 ###
 # Copyright 2016 - 2025 Green River Data Analysis, LLC
 #
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
 class ApplicationRecord < ActiveRecord::Base
-  include Efind
```
ttoomey, line 8: This was extracted to CustomApplicationRecord so that other models can inherit from ActiveRecord::Base rather than through ApplicationRecord. Maybe we can get away without change this but it might have been causing issues with multi db support

app/models/filters/step_range.rb
```diff
@@ -36,36 +38,38 @@ def interesting_column
...
     def ordered_steps
       @ordered_steps ||= begin
-        scope = GrdaWarehouse::CasReport.on_route(route) # .started_between(start_date: @range.start, end_date: @range.end + 1.day)
+        scope = GrdaWarehouse::CasReport.on_route(route)
         step_order = scope.distinct.
           pluck(:match_step, :decision_order).to_h
-        steps = step_order.keys
-        at = scope.arel_table
-        at2 = at.dup
-        at2.table_alias = 'at2'
```
ttoomey: needs review

lib/tasks/db.rake
```diff
@@ -2,9 +2,7 @@ namespace :db do
   namespace :schema do
     desc "Conditionally load the database schema"
     task :conditional_load, [] => [:environment] do |t, args|
-      if ApplicationRecord.connection.table_exists?(:schema_migrations)
```
ttoomey, line 5: modified this and all similar rake tasks to close the connection. It seems that db:schema:load:primary now tries to destroy and recreate the db which fails if the connection is left open.

app/models/grda_warehouse/hud/client.rb
```diff
@@ -794,41 +793,35 @@ def self.disabled_client_scope(client_ids: nil)
     # client has a disability response in the affirmative
     # where they don't have a subsequent affirmative or negative
     def self.disabled_client_because_disability_scope
-      d_t1 = GrdaWarehouse::Hud::Disability.arel_table
-      d_t2 = d_t1.dup
-      d_t2.table_alias = 'disability2'
-      c_t1 = GrdaWarehouse::Hud::Client.arel_table
-      c_t2 = c_t1.dup
-      c_t2.table_alias = 'source_clients'
```
ttoomey: needs review

app/models/grda_warehouse/hud/service.rb
```diff
@@ -36,19 +38,20 @@ class Service < Base
     scope :descending, -> { order arel_table[:DateProvided].desc }
     # really, the scope below should just be true, but it isn't; this culls things down to the most recent entry of the given type for the date
     scope :uniqueness_constraint, -> {
-      st1 = arel_table
-      st2 = st1.dup
-      st2.table_alias = 'st2'
```
ttoomey: needs review

app/models/grda_warehouse/hud/client.rb
```diff
@@ -17,7 +17,6 @@ class Client < Base
     include RandomScope
     include ArelHelper
     include HealthCharts
-    include ApplicationHelper
```
ttoomey, line 20: removed, I believe we only needed dates_overlap (and shouldn't be mixing the app helper in to models)

app/models/grda_warehouse/report/base.rb
```diff
@@ -76,26 +78,6 @@ def self.update_recent_history_table
...
-      @destination_client_table ||= begin
-        table = GrdaWarehouse::Hud::Client.arel_table.dup
-        table.table_alias = 'destination_clients'
-        table
-      end
-    end
-
-    def self.client_join_table
-      GrdaWarehouse::WarehouseClient.arel_table
-    end
```
ttoomey, line 97: I don' think these are used

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -93,15 +100,25 @@ def identify_duplicates
             data_source_id: client.data_source_id,
           )
         end
-        GrdaWarehouse::Hud::Client.import(
```
ttoomey, line 96: raise errors if the import fails

app/models/grda_warehouse/schema_migration.rb
```diff
@@ -1,13 +0,0 @@
-###
-# Copyright 2016 - 2025 Green River Data Analysis, LLC
-#
-# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
-###
-
-module GrdaWarehouse
```
ttoomey, line 7: This doesn't seem to work as is in rails 7.1

app/models/health_base.rb
```diff
@@ -1,15 +1,18 @@
...
 
-class HealthBase < ApplicationRecord
+class HealthBase < ActiveRecord::Base
+  include CustomApplicationRecord
+
   self.abstract_class = true
   connects_to database: { writing: :health, reading: :health }
   has_paper_trail versions: {class_name: 'Health::HealthVersion'}
 
-  include ArelHelper
```
ttoomey, line 12: already included

app/models/reporting_base.rb
```diff
@@ -6,13 +6,9 @@
 
 # frozen_string_literal: false
 
-class ReportingBase < ApplicationRecord
-  self.abstract_class = true
-  include ArelHelper
```
ttoomey, line 11: already included

bin/db_prep
```diff
@@ -53,28 +53,35 @@ def run_or_exit(cmd)
   exit(1)
 end
 
-def seeds_ran?(db_section)
+def with_connection(db_section)
```
ttoomey, line 56: this properly closes the db connection

bin/db_prep
```diff
@@ -16,7 +16,7 @@ NO_PRINT_DONE = (ARGV[1] == '--no-print-done')
 CREATE_DATABASES = ! DONT_CREATE_DATABASES
 
 def db(db_section)
-  @db ||=
```
ttoomey, line 19: this script-scoped ivar might cause trouble

drivers/cas_access/app/models/cas_access/filters/filter_base.rb
```diff
@@ -281,33 +283,36 @@ def ordered_steps
         scope = CasAccess::Reporting::Decisions.on_route(match_route)
         step_order = scope.distinct.
           pluck(:match_step, :decision_order).to_h
-        steps = step_order.keys
-        at = scope.arel_table
-        at2 = at.dup
-        at2.table_alias = 'at2'
```
ttoomey: review needed

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -18,7 +20,7 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
   self.table_name = :CustomAssessments
   self.sequence_name = "public.\"#{table_name}_id_seq\""
 
-  include ::HmisStructure::Assessment
```
ttoomey, line 21: use dedicated mixin to get correct hmis_configuration fields, avoids errors when aliasing

app/models/concerns/custom_application_record.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+module CustomApplicationRecord
```
ttoomey, line 3: extracted from ApplicationRecord

drivers/hmis/app/models/hmis/hud/custom_client_contact_point.rb
```diff
@@ -38,7 +40,7 @@ class Hmis::Hud::CustomClientContactPoint < Hmis::Hud::Base
   belongs_to :user, **hmis_relation(:UserID, 'User'), optional: true
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
   has_one :active_range, class_name: 'Hmis::ActiveRange', as: :entity, dependent: :destroy
-  alias_to_underscore [:NameDataQuality, :ContactPointID]
```
ttoomey, line 41: NameDataQuality doesn't exist afaict

drivers/hmis/app/models/hmis/hud/custom_service.rb
```diff
@@ -9,7 +11,7 @@
 class Hmis::Hud::CustomService < Hmis::Hud::Base
   self.table_name = :CustomServices
   self.sequence_name = "public.\"#{table_name}_id_seq\""
-  include ::HmisStructure::Service
```
ttoomey, line 12: use dedicated mixin to get correct hmis_configuration fields, avoids errors when aliasing

drivers/hmis/app/models/hmis/hud/funder.rb
```diff
@@ -27,9 +29,8 @@ class Hmis::Hud::Funder < Hmis::Hud::Base
   end
 
   # Convert funder string to int #183572073
-  def Funder # rubocop:disable Naming/MethodName
-    super&.to_i
-  end
+  # FIXME: document why we cast this to an integer
+  attribute :Funder, :integer
```
ttoomey, line 33: probably should add a unit test

spec/requests/omniauth_spec.rb
```diff
@@ -1,12 +1,13 @@
...
+
 require 'rails_helper'
 if ENV['OKTA_DOMAIN'].present?
   # Make sure that https://nvd.nist.gov/vuln/detail/CVE-2015-9284 is mitigated
   RSpec.describe Users::OmniauthCallbacksController, type: :request do
     describe 'GET /users/auth/:provider protects against CVE-2015-9284' do
       it do
-        expect do
-          get '/users/auth/okta'
-        end.to raise_error(ActionController::RoutingError)
```
ttoomey, line 9: we get a 404 now due to `show_exceptions = :rescuable`

spec/requests/rack_attack_spec.rb
```diff
@@ -32,23 +32,18 @@ def till_throttled(requests_to_send:, throttled_status: 429, mode: :default, &bl
...
-      time_scale = 0.0001
+      time_advance = 0.01
     else
       raise 'unknown mode'
     end
 
-    # travel to hour boundary so we always start at 00:00
-    Timecop.travel((Time.current + 1.hour).beginning_of_hour)
-    # adjust speed of time to reliably trigger burst throttling (10req/1sec)
-    Timecop.scale(time_scale)
```
ttoomey, line 47: replace scale with manually stepping through time with freeze for greater reliability

drivers/hmis/spec/system/hmis/enrollment_management_spec.rb
```diff
@@ -169,7 +169,8 @@ def make_household(household_id: Hmis::Hud::Base.generate_uuid, enrollment_facto
             user: u1,
           )
           # override garbage factory exit date
-          exit.update!(exit_date: today - 4.days)
```
ttoomey, line 172: This hits a validation error; I'm not sure why this validation wasn't firing before

config/initializers/rack_attack.rb
```diff
@@ -141,7 +141,8 @@ class Rack::Attack::Request
     period: 1.second,
   ) do |request|
     if request.tracking_enabled? && request.anonymous?
-      request.request_ip
```
ttoomey, line 144: looks like if we don't exclude history reqs here, the history req is not throttled as it superseded by the more general brute force throttle

config/initializers/db_warehouse.rb
```diff
@@ -1,4 +0,0 @@
-# Rails.logger.debug "Running initializer in #{__FILE__}"
-
-# save health database settings in global var
-DB_WAREHOUSE ||= YAML.safe_load(ERB.new(File.read(Rails.root.join("config","database.yml"))).result, aliases: true)[Rails.env]['warehouse']
```
ttoomey, line 4: We seemed to be using this to with establish_connection which is probably not correct (transactions aren't shared for one). Instead should use `connects_to [database]`

config/initializers/acts_as_taggable_on.rb
```diff
@@ -1,14 +1,6 @@
...
-module ActsAsTaggableOn
-  class Tag < ::ActiveRecord::Base
-    establish_connection DB_WAREHOUSE
-  end
-  class Tagging < ::ActiveRecord::Base
-    establish_connection DB_WAREHOUSE
-  end
+ActsAsTaggableOn.setup do |config|
+  # Store the tags in the warehouse
+  config.base_class = 'GrdaWarehouseBase'
```
ttoomey, line 5: lack of shared connection was leading to strange behavior in tests

app/models/grda_warehouse_base.rb
```diff
@@ -19,15 +20,6 @@ def self.postgres?
     connection.adapter_name.in?(['PostgreSQL', 'PostGIS'])
   end
 
-  def self.reset_connection
```
ttoomey, line 22: seems unused

app/models/grda_warehouse/vispdat/base.rb
```diff
@@ -11,17 +13,21 @@ class Base < GrdaWarehouseBase
...
     # Constants
     ####################
-    US_PHONE_NUMBERS = /\A(\+1)?\(?(\d{3})\)?\s*-?\s*(\d{3})\s*-?\s*(\d{4})\s*-?\s*\z/.freeze
+    US_PHONE_NUMBERS = /\A(\+1)?\(?(\d{3})\)?\s*-?\s*(\d{3})\s*-?\s*(\d{4})\s*-?\s*\z/
 
     ####################
     # enums
     ####################
 
-    enum contact_answers: [
```
ttoomey, line 20: this seems like the wrong field name for the enum

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
```diff
@@ -74,8 +74,7 @@ def sync!
             GrdaWarehouse::Hud::Project.find_each do |project|
               safe_project_names[project.id] = project.safe_project_name
             end
-            TodoOrDie('test behavior after rails upgrade', if: Rails.version !~ /\A7\.0/)
-            client_source.lazy_preload(preloads).
```
ttoomey, line 78: not sure if the preloader is still behaving; could use a test to check for n+1

app/models/importers/hmis_auto_migrate/base.rb
```diff
@@ -97,11 +99,7 @@ def upload(file_path:)
     end
 
     private def add_content_to_upload_and_save(file_path:)
-      path = Pathname.new(file_path)
-      @upload.hmis_zip.attach(
```
ttoomey, line 101: maybe this change isn't necessary but seems like newer active storage supports a less awkward syntax

drivers/hmis/app/graphql/mutations/update_relationship_to_ho_h.rb
```diff
@@ -19,12 +21,16 @@ def resolve(enrollment_id:, enrollment_lock_version: nil, relationship_to_ho_h:,
 
       enrollment.lock_version = enrollment_lock_version if enrollment_lock_version
 
+      validations = nil
       if relationship_to_ho_h == 1
         # Lock to avoid duplicate request collisions. with_lock also starts a transaction.
         enrollment.project.with_lock do
           hoh_changer = Hmis::HohChangeHandler.new(new_hoh_enrollment: enrollment, hud_user_id: hud_user.user_id)
           validations = hoh_changer.validate(include_warnings: !confirmed)
-          return { errors: validations } if validations.any?
```
ttoomey, line 27: rollback on return from a tx block is deprecated in 7.1 and removed in 7.2

config/environments/production.rb
```diff
@@ -73,8 +83,8 @@
   # the I18n.default_locale when a translation cannot be found).
   config.i18n.fallbacks = [I18n.default_locale]
 
-  # Send deprecation notices to registered listeners.
-  config.active_support.deprecation = :notify
+  # Log any deprecations.
+  config.active_support.report_deprecations = true
```
ttoomey, line 87: need to verify deprecations are reported somewhere (sentry)

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,283 @@
...
+# because we keep active storage table in a different db from uploads (add_content_to_upload_and_save)
+Rails.application.config.active_record.before_committed_on_all_records = false
+
+###
+# Disable automatic column serialization into YAML.
+# To keep the historic behavior, you can set it to `YAML`, however it is
+# recommended to explicitly define the serialization method for each column
+# rather than to rely on a global default.
+#++
+Rails.application.config.active_record.default_column_serializer = 'YAML'
```
ttoomey, line 201: we should probably revisit serializers and set the coder to YAML so we can remove this global config

config/initializers/new_framework_defaults_7_1.rb
```diff
@@ -0,0 +1,283 @@
...
+# This matches the behaviour of all other callbacks.
+# In previous versions of Rails, they ran in the inverse order.
+#++
+Rails.application.config.active_record.run_after_transaction_callbacks_in_order_defined = true
+
+###
+# Whether a `transaction` block is committed or rolled back when exited via `return`, `break` or `throw`.
+#++
+# This is a dangerous change, leaving off for now to collect deprecations
+# Rails.application.config.active_record.commit_transaction_on_non_local_return = true
```
ttoomey, line 225: this is a big change and the primary reason not to go with the 7.1 settings until we have more confidence that we won't be impacted by this change

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
ttoomey: @eanders can you point me to how I can exercise this? Maybe I should add a unit test for PushClientsToCas

app/models/grda_warehouse/hud/enrollment.rb
```diff
@@ -163,6 +163,8 @@ class Enrollment < Base
       visible_to(user)
     end
 
+    # this scope is defined in the ClientAccessControl::GrdaWarehouse::Hud::EnrollmentExtension
+    # TODO: why not just define it here
```
eanders, line 167: Initially we envisioned multiple possible `visible_to` mechanisms that might be built out in different drivers and enabled as necessary.  So far that hasn't manifested, so the main reason for defining it in the driver is that drivers should be dependent on the app, not the app on the driver.

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
eanders: This class requires the CAS database be setup.  We added `lazy_preload` here: https://github.com/greenriver/hmis-warehouse/pull/5236 because it handles some scenarios that we were unable to hit with a standard preload.  

The complexity of the calculations for what is pushed to CAS is installation and configuration specific, often relying on data not in the standard HMIS shapes and specific values of fields to determine what is valid.

I'd love some unit tests, but I think we'll need to add a CAS database (and keep the shape in-sync) to CI before that would be fully possible.  

This precursors to this model generally gets exercised as part of the calculator tests: `spec/models/grda_warehouse/cas_project_client_calculator`.

eanders: Was there a reason we don't like `lazy_preload`?

app/models/grda_warehouse/vispdat/base.rb
```diff
@@ -11,17 +13,21 @@ class Base < GrdaWarehouseBase
...
-      :contact_phone,
-      :contact_virtual,
-      :contact_in_person,
-    ]
+    enum contact_method: {
```
eanders, line 22: Should this be?
```suggestion
    enum contact_methods: {
```

I'm unsure why the enum name doesn't match the column name, but I'm also not clear that it needs to (or why we're using an enum in this case.)

Is there a problem with the array/symbol syntax?  That appears to be valid according to the docs: https://api.rubyonrails.org/v5.1.7/classes/ActiveRecord/Enum.html.  Maybe your change is to indicate that we're actually storing the text in the DB rather than an integer.

app/models/importers/hmis_auto_migrate/base.rb
eanders: As long as we can confirm we're setting `filename` I think that's great, way cleaner.

app/models/grda_warehouse_base.rb
eanders: I feel like this was used for some debugging at some point.  Happy to see it removed for now.

bin/setup
```diff
@@ -1,49 +1,60 @@
...
+  # This script is a way to set up or update your development environment automatically.
+  # This script is idempotent, so that you can run it at any time and get an expectable outcome.
+  # Add necessary setup steps to this file.
 
-Dir.chdir APP_ROOT do
   puts '== Installing dependencies =='
-  system 'gem install bundler --conservative'
-  system 'bundle check || bundle install --without seven_zip'
+  system! 'gem install bundler --conservative'
+  system('bundle check') || system!('bundle install --without seven_zip')
```
eanders: Looks like we've removed `seven_zip_ruby` from the gemfile.  We can probably remove it here as well:
```suggestion
  system('bundle check') || system!('bundle install')
```

config/initializers/acts_as_taggable_on.rb
eanders: That'a a much cleaner fix. Thanks!

drivers/cas_access/app/models/cas_access/filters/filter_base.rb
```diff
@@ -281,33 +283,36 @@ def ordered_steps
...
-              where(at2[:decision_order].lt(at[:decision_order])).
-              where(at2[:match_step].eq(step)).
-              exists,
-          ).distinct.pluck(:match_step, :decision_order).
-            map do |match_step, decision_order|
-              "(#{decision_order}) #{match_step}"
-            end
-          [step, followups]
-        end.to_h
+        step_order.keys
```
eanders: This is missing some assignment here I think.

drivers/hmis/app/models/hmis/hud/custom_service.rb
```diff
@@ -19,7 +21,12 @@ class Hmis::Hud::CustomService < Hmis::Hud::Base
   belongs_to :user, **hmis_relation(:UserID, 'User'), inverse_of: :services, optional: true
   belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
   belongs_to :custom_service_type
-  alias_attribute :service_type, :custom_service_type
+  def service_type = custom_service_type
```
eanders, line 24: I've seen this syntax start to creep in, is this somethign we want to adopt for short methods?  It feel slightly awkward to me.

drivers/hmis/app/models/hmis/hud/funder.rb
eanders: https://github.com/open-path/Green-River/issues/1017 (is the current representation of #183572073)

I believe we're casting as an integer because the spec lists the `Funder` column as an integer and we incorrectly store it as a string.  Probably better to write a migration to fix the data type in the DB.

drivers/hmis_supplemental/app/models/hmis_supplemental/field_value.rb
```diff
@@ -14,7 +16,6 @@ module HmisSupplemental
   class FieldValue < GrdaWarehouseBase
     self.table_name = 'hmis_supplemental_field_values'
     belongs_to :data_set, class_name: 'HmisSupplemental::DataSet'
-    serialize :data, type: Hash
```
eanders, line 17: Is this removal intentional?  Maybe the column is jsonb and this is the default?

Also, it appears there are duplicate copyright notices above.

eanders (commented):
Thanks @ttoomey, I think this looks generally great!  
I dropped a bunch of notes/questions inline.  Feel free to reach out if you want to discuss anything.

app/models/grda_warehouse/vispdat/base.rb
ttoomey: My understanding is that the first argument to enum should be the field name. Rails 7.1 is more strict and requires the field name to actually exist on the model; so this should the singular form "contact_method" to match the field. 

Yes, the hash syntax is to indicate that these are string and not numeric values. I'll double-check this but I think the change was needed since the col type is string and not integer.

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
ttoomey: Thanks, I removed it since we had a todo-or-die to check if the preload behavior was fixed in newer version of rails, in which case the dependency wouldn't be needed.

I was sort of hoping I could add an isolated test case to prove if we needed it or not but maybe that's too ambitious; happy to leave the lazy_preload in for now and we can plan for a future round of dependency pruning.

drivers/cas_access/app/models/cas_access/filters/filter_base.rb
ttoomey: ah oops, I thought I rolled this change back. Thanks for catching!

drivers/hmis_supplemental/app/models/hmis_supplemental/field_value.rb
ttoomey: Yes, sorry I should have called that out. This pairs with the migration in db/warehouse/migrate/20250415143753_fix_json_serialization.rb

This was mistakenly storing the serialized value as a string inside of a JSON field rather than as JSON (as intended)

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
eanders: I guess I'm inclined to leave it and then run some manual tests with and without it after 7.1 goes live.  
If it isn't needed we could remove it.  
That said, I think it might be a better overall approach than our sometimes crazy attempts at guessing at preload needs.

drivers/hmis/app/models/hmis/hud/custom_service.rb
ttoomey: It's the "endless method" syntax introduced in ruby 3. I guess I don't have a problem with it when used for simple accessors like this; it seems easy to read to me but that's subjective. If you want we could use alias_method instead which should be equiv in this case.

eanders: I don't have a strong opinion, but it feels like a formatting decision we should make actively to keep the code consistent should all one line/one variable(?) methods be written that way?

ttoomey: Maybe, I guess I would think of it as akin to using a line continuation: use it when it helps readability.

app/models/grda_warehouse/tasks/push_clients_to_cas.rb
ttoomey: fair, let's leave it in

drivers/datalab_testkit/spec/models/datalab_testkit_context.rb
```diff
@@ -69,8 +75,6 @@ def report_result
     # NOTE: SPM runs subsequent DQ reports, but sets @report_result, so we'll use that if available
     @report_result || ::HudReports::ReportInstance.last
   end
-
-  require_relative './table_comparisons'
```
ttoomey, line 73: this require seemed kind of rough, refactored to deal with naming issues

drivers/datalab_testkit/spec/models/table_comparisons.rb
```diff
@@ -1,103 +1,111 @@
...
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
-COLUMN_LABELS = ('A'..'Z').to_a.freeze
-
-def goals(file_path:, question:, csv_name:, external_column_header:)
-  csv_file = File.join(file_path, csv_name || question + '.csv')
-  data = CSV.read(csv_file)
-  data = data[1..] if external_column_header # Drop the first line if the header is outside the table in the spec, but included in the goal file
+module DatalabTestkit
```
ttoomey, line 9: Encapsulated these methods in a module

spec/models/grda_warehouse/client_notes/client_note_spec.rb
```diff
@@ -42,8 +42,8 @@
   end
 
   describe 'associations' do
-    it { should belong_to(:user).optional }
-    it { should belong_to(:client).optional }
```
ttoomey, line 46: let's not use `shoulda` (removed the gem)

spec/support/capybara/cuprite/browser_with_auto_restart.rb
```diff
@@ -0,0 +1,38 @@
+# frozen_string_literal: true
+
+require 'capybara/cuprite'
+
+module Capybara
+  module Cuprite
+    module BrowserWithAutoRestart
```
ttoomey: hopefully helps with `DeadBrowserError` in CI

spec/support/e2e_tests.rb
```diff
@@ -167,11 +167,18 @@ def truncate
     end
 
     def puts(log_str)
+      return if log_str.nil?
```
ttoomey, line 168: behavior change after upgrade to ferrum/cuprite

.github/workflows/rails_tests.yml
```diff
@@ -91,6 +91,8 @@ jobs:
           TZ: "America/New_York"
           PORT: 3333
           CONNECTION_TIMEOUT: 600000
+          FERRUM_PROCESS_TIMEOUT: 45
+          FERRUM_DEFAULT_TIMEOUT: 45
```
ttoomey: added env vars in an attempt to make system tests more reliable

drivers/hmis/app/graphql/sources/active_record_association.rb
```diff
@@ -15,16 +17,6 @@ def initialize(association_name, scope = nil)
   end
 
   def fetch(records)
-    # Rails.logger.info("preloading started #{records.first.class.name}.#{@association_name}") if records.any?
-    TodoOrDie('test behavior after rails upgrade, see #6019', if: Rails.version !~ /\A7\.0/)
```
ttoomey, line 19: This is no longer needed and may have been causing issues in some cases, ported from https://github.com/greenriver/hmis-warehouse/pull/5302

drivers/hmis/spec/models/active_record_preloader_spec.rb
```diff
@@ -0,0 +1,185 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+#  testing preloader behavior
+RSpec.describe Sources::ActiveRecordAssociation do
```
ttoomey, line 6: test against rails preloader to verify we are using it correctly

spec/support/e2e_tests.rb
```diff
@@ -62,7 +61,6 @@ def self.perform(
             window_size: [1200, 1600],
             browser_options: RemoteChrome.connected? ? { 'no-sandbox' => nil } : {},
             headless: ENV.fetch('CI', 'true') == 'true',
-            process_timeout: process_timeout,
```
ttoomey, line 65: moved to env var

ttoomey:
@gigxz tagging you on this. The PR is large so don't feel you need to review the entire thing. I just wanted to highlight changes that may be most relevant to the HMIS:
* composite key relationship now uses query_constraints
* upgrade to graphql gem
* adjustment to association loader to remove association resets
* change to GrdaWarehouseBase, it no longer inherits from ApplicationRecord
* various code changes under drivers/hmis to correct deprecations in `alias*`
* changes to make system tests more reliable (still working on this)

ttoomey:
@eanders thanks for the review! I think I've addressed all the feedback. Maybe this can go into 161 or 162?

eanders (approved):
@ttoomey, this looks great.  Thanks for all the effort.  
I wonder if it makes sense to merge this into 162 and skip 161 on staging going right to 162 (or maybe we deploy 162 to staging next Monday) so we can have 2 weeks to review.

ttoomey:
> I wonder if it makes sense to merge this into 162 and skip 161 on staging going right to 162 (or maybe we deploy 162 to staging next Monday) so we can have 2 weeks to review.

I like that, I think we would benefit from more QA time on this

ttoomey:
I'll merge this into 162

.github/workflows/rails_tests.yml
gigxz: Thanks! linking to https://github.com/open-path/Green-River/issues/7547

app/models/concerns/hmis_structure/custom_assessment.rb
```diff
@@ -0,0 +1,71 @@
+# frozen_string_literal: true
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module HmisStructure::CustomAssessment
```
gigxz, line 9: I don't think we should introduce this unless we have agreed that this is a shape we want to import/export, which we haven't yet. same with custom service

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
gigxz: I would think we can just remove this entirely. I will look at it further. we are maybe pulling something related to validations from here.

gigxz: Ok its a bit of a rabbit hole, its definitely something we need to clean up. For example CustomServiceValidator uses the service structure file to validate the presence of EnrollmentID and some other fields. Its  a weird setup from when we initially went down the direction of relying on the Structure files for validation, and we didn't fully walk it back. (Which imo we should).

What changes here are Rails 7.1-dependent?

drivers/hmis/spec/system/hmis/enrollment_management_spec.rb
gigxz: Weird.
This code points to something that needs to be addressed in the `hmis_hud_exit` factory, to make it not override exit_date if it was passed explicitly. or change line 165 to use hmis_base_hud_exit

gigxz:
Thanks for all the effort @ttoomey! this looks good to me. I tested this with the HMIS frontend as well.

My only hesitation is the addition of the custom HMIS structure files. That feels like a bigger decision that I would prefer not to be coupled to the rails upgrade, but I'm sure you have a reason for it. Its certainly less intrusive than removing the concerns entirely from those models (custom assessment, custom service) so I'm happy to go with that and have a follow-up ticket to remove them (pending discussion, curious also to hear from Elliot).

gigxz (approved):

ttoomey:
> My only hesitation is the addition of the custom HMIS structure files. That feels like a bigger decision that I would prefer not to be coupled to the rails upgrade, but I'm sure you have a reason for it. Its certainly less intrusive than removing the concerns entirely from those models (custom assessment, custom service) so I'm happy to go with that and have a follow-up ticket to remove them (pending discussion, curious also to hear from Elliot). 

Thanks for highlighting that @gigxz

The reason for the change is that rails 7.1 deprecates using alias_attribute for columns that are not attributes on the model. The CustomAssessment and CustomService use the Assessment and Service mixins respectively which result  in hmis_configuration describing fields not present in the model's attributes. This issue was surfaced in `HmisStructure::Shared` which calls `alias_attribute`, resulting in the aforementioned deprecation warnings.

My intention with adding these new mixins was to simply correct the hmis_configuration to match the attributes on the models in a way that seemed to best match the established pattern.

However it sounds like I missed the significance of these shared mixins; Sorry about that! Maybe we can create some documentation for these mixins and alert others to the potential risks with modifying them (Or perhaps add some unit tests for critical functionality)

Regardless I can back out this change and instead modify `HmisStructure::Shared` to ignore columns that are not present on the model. However that will make us blind to future cases where the model attributes and configured fields do not match.

.github/workflows/rails_tests.yml
ttoomey: these ended up in the wrong env section, fixing

ttoomey:
discussed with Elliot: going to merge to 162 as-is and then follow in another round if we need to re-org or move the HMIS structure concerns.

PR #5297, opened 2025-04-14
Fix staff assignment bug for limited visibility enrollments

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7563
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1134 (check out for testing)

* Fix bug where client enrollment table would crash when viewing enrollments with limited access, by resolving staff assignments on the Enrollment rather than on the Household. Staff assignment will always resolve as null for limited-access enrollments. This also improves performance because we do not need to resolve the household type at all.
* Fix N+1 performance issue when resolving staff assignments on the Project Households and Project Enrollments tables
* Add/update performance tests for project enrollments & project households. Queries reflect current state on the linked frontend PR

See issue for testing instructions.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/household.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/household.rb
@@ -14,6 +14,9 @@ module Types
     field :short_id, ID, null: false
     field :household_clients, [HmisSchema::HouseholdClient], null: false
     field :household_size, Int, null: false
+
+    field :current_staff_assignments, [HmisSchema::StaffAssignment], null: false
+    # historical paginated staff assignments. no longer used to fetch "current" staff assignments because of performance reasons, the currentStaffAssignments field is used instead.
     field :staff_assignments, HmisSchema::StaffAssignment.page_type, null: true do
       argument :is_currently_assigned, Boolean, required: false
     end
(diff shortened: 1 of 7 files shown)


drivers/hmis/app/graphql/types/hmis_schema/household.rb
```diff
@@ -54,16 +57,20 @@ def assessments(**args)
...
 
+    def current_staff_assignments
+      load_ar_association(object, :staff_assignments, scope: Hmis::StaffAssignment.order(created_at: :desc, id: :desc))
+    end
+
+    # This field results in N+1 because it is paginated. It is used for displaying the staff assignment history
+    # for a particular household. When loading staff assignments on a batch of households,
+    # use the 'current_staff_assignments' field instead.
     def staff_assignments(is_currently_assigned: true)
-      # There's no current use case for returning all (both currently assigned and formerly assigned)
```
martha, line 58: If I'm understanding correctly, this comment "There's no current use case for returning all (both currently assigned and formerly assigned) in the same query" is still relevant, right? That's the reason for the `is_currently_assigned` param

martha (approved):
Thanks Gig. Looks great. Let me know if you would like me to make a ticket for further optimizing the last contact date.

drivers/hmis/app/graphql/types/hmis_schema/household.rb
gigxz: We only ever call this with `isCurrentlyAssigned: false` now. I'll update the comment!

PR #5299, opened 2025-04-15
add active flag to cohort columns

dtgreiner:
## _Merging this PR_

## Description
- Add cohort columns types to database and include an active flag.
- Update logic to clear columns from cohorts when deactivating a column

## Type of change
New feature


app/models/grda_warehouse/cohort_column_type.rb
```diff
@@ -0,0 +1,310 @@
...
+    def remove_from_cohorts
+      GrdaWarehouse::Cohort.all.each do |cohort|
+        cohort.update!(column_state: cohort.column_state.reject { |col| col.cohort_column_type.class_name == class_name })
+      end
+    end
+
+    def self.maintain!
+      known_cohort_column_types.each do |class_name|
+        column_type = find_by(class_name: class_name)
+        create!(class_name: class_name, active: true) unless column_type
```
eanders: Maybe find_or_create here?  I think it should work, though I kind of like your implementation better, since you are just throwing out the record if found. Maybe that should be:

```suggestion
        create!(class_name: class_name, active: true) unless where(class_name: class_name).exists?
```
To prevent instantiating the object. 

Actually, on second thought, I think you want to pluck all existing class names and to the check against the result (converted to a set). Then this will only make one query when no new columns are needed.

app/models/grda_warehouse/cohort_column_type.rb
```diff
@@ -0,0 +1,310 @@
...
+        'CohortColumns::UserNumeric1',
+        'CohortColumns::UserNumeric2',
+        'CohortColumns::UserNumeric3',
+        'CohortColumns::UserNumeric4',
+        'CohortColumns::UserNumeric5',
+        'CohortColumns::UserNumeric6',
+        'CohortColumns::UserNumeric7',
+        'CohortColumns::UserNumeric8',
+        'CohortColumns::UserNumeric9',
+        'CohortColumns::UserNumeric10',
```
eanders: Since these are just strings now, maybe we should build the "user" columns programmatically, and loop (1..10).map { ... }. That would make this list much shorter.

db/warehouse/migrate/20250409145018_populate_cohort_column_table.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+class PopulateCohortColumnTable < ActiveRecord::Migration[7.0]
+  def up
+    GrdaWarehouse::CohortColumnType.maintain!
```
eanders: I think this needs to go in db seed as well. Eventually the data migration will be removed but new installations will need the column data. Not sure how that plays out during deployment. I can't remember if seed is called before or after migration. Hopefully after. In which case we might not want the migration at all.

app/models/grda_warehouse/cohort_column_type.rb
eanders, line 1: Can this file just be Cohorts::Column?  Maybe that's taken. I'm not a huge fan of the "Type" designation here. That might also let it become an STI table eventually with all current models inheriting from it.

app/models/grda_warehouse/cohort_column_type.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5299/commits/e36fbed7fbb24bc00231cb8cfca030ef129a67e7) to pluck the existing names ahead of time.

app/models/grda_warehouse/cohort_column_type.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5299/commits/e36fbed7fbb24bc00231cb8cfca030ef129a67e7)

db/warehouse/migrate/20250409145018_populate_cohort_column_table.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5299/commits/e36fbed7fbb24bc00231cb8cfca030ef129a67e7) 

I've added this to the seed method but still want to look into when this is getting called in case we can clean up the migration.

app/models/grda_warehouse/cohort_column_type.rb
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5299/commits/e36fbed7fbb24bc00231cb8cfca030ef129a67e7) 

I've changed this from `GrdaWarehouse::CohortColumnType` to `GrdaWarehouse::Cohorts::CohortColumn`. Using the class name 'Column' seemed to be causing issues within the migration process. This naming convention also matches other `GrdaWarehouse::Cohorts` classes (e.g. `GrdaWarehouse::Cohorts::CohortClientTab`, `GrdaWarehouse::Cohorts::CohortColumnMetadata`)

app/models/cohort_columns/base.rb
```diff
@@ -22,6 +22,15 @@ class Base < ::ModelForm
...
     attribute :editable, Boolean, lazy: false, default: true
     attribute :current_user
+    attribute :cohort_column, GrdaWarehouse::Cohorts::CohortColumn, lazy: true, default: ->(r, _) { GrdaWarehouse::Cohorts::CohortColumn.find_by!(class_name: r.class_name) }
+
+    def active?
+      cohort_column&.active?
+    end
+
+    def class_name
+      self.class.name
```
eanders, line 35: Not sure this matters, but if we're kinda sorta using these as STI models, this should probably be:
```suggestion
      self.class.sti_name
```

eanders: Actually, never mind.  After looking at the [source](https://apidock.com/rails/v7.1.3.4/ActiveRecord/Inheritance/ClassMethods/sti_name) of `sti_name`, I don't think this is necessary at this time.  I was under the impression it was doing something more to ensure it was DB safe or something.

spec/models/cohort_analytics_spec.rb
```diff
@@ -27,6 +27,10 @@
     end
   end
 
+  before(:all) do
+    GrdaWarehouse::Cohorts::CohortColumn.maintain!
+  end
+
```
eanders: Maybe this should go in `rails_helper.rb`.  There's a `before(:suite)` section that probably makes sense.

app/models/cohort_columns/base.rb
```diff
@@ -22,6 +22,15 @@ class Base < ::ModelForm
     attribute :cohort_client
     attribute :editable, Boolean, lazy: false, default: true
     attribute :current_user
+    attribute :cohort_column, GrdaWarehouse::Cohorts::CohortColumn, lazy: true, default: ->(r, _) { GrdaWarehouse::Cohorts::CohortColumn.find_by!(class_name: r.class_name) }
```
eanders: Do we need to pass `r` here?  Most of the time we should be interacting with the child class right?  So maybe?
```suggestion
    attribute :cohort_column, GrdaWarehouse::Cohorts::CohortColumn, lazy: true, default: -> { GrdaWarehouse::Cohorts::CohortColumn.find_by!(class_name: class_name) }
```

Or maybe even:
```suggestion
    def cohort_column
      GrdaWarehouse::Cohorts::CohortColumn.find_by!(class_name: r.class_name)
    end
```

I'm probably missing something, but that should mean we don't actually need to set it up in the factories.

spec/models/grda_warehouse/cohort_spec.rb
```diff
@@ -134,4 +146,49 @@
...
+
+  describe 'column_state handling' do
+    let!(:test_column) { build :user_string_cohort_column_1 }
+
+    before do
+      # Add test column to cohort's column_state
+      columns = cohort.column_state || []
+      test_column.cohort_column.activate
+      columns << test_column
+      cohort.update(column_state: columns)
```
eanders: Do you think there's any benefit to testing this against `GrdaWarehouse::Cohorts::CohortColumn.known_cohort_columns`?  
I _think_ that would enforce that all of those columns can be instantiated (and have classes that back them).  Basically:
 ```suggestion
      @all_columns = GrdaWarehouse::Cohort.available_columns
      @all_columns.each { |c| c.cohort_column.activate }
      GrdaWarehouse::Cohorts::CohortColumn.known_cohort_columns
      columns = cohort.column_state || @all_columns
      cohort.update(column_state: columns)
```

File: spec/models/grda_warehouse/cohort_spec.rb [hunk 8]
```diff
@@ -134,4 +146,49 @@
...
+    end
+
+    it 'removes inactive columns from column_state' do
+      expect(cohort.column_state.map(&:class_name)).to include('CohortColumns::UserString1')
```
eanders: These tests would have to be updated to something like:
```suggestion
      expect(cohort.column_state.map(&:class_name)).to contain_exactly(@all_columns)
```

eanders (commented):
Thanks @dtgreiner, let me know if you have questions about any of my comments.  I think this is looking good, just a couple of possible cleanups.

app/models/cohort_columns/base.rb
dtgreiner: I can give the latter suggesting a go and run through some tests. I ran into a handful of issues getting this to run tests properly, so if this possibly clears up some hurdles, it makes sense to me

spec/models/grda_warehouse/cohort_spec.rb
dtgreiner: I like it, and worst comes we should have a test checking that the classes in the database also exist in the code. I'll give it a run locally and see what falls out

db/warehouse/migrate/20250409145018_populate_cohort_column_table.rb
eanders, line 1: Do you think we need this if we call `.maintain!` at the end of db seed?

spec/models/grda_warehouse/cohort_spec.rb
```diff
@@ -134,4 +142,49 @@
...
+
+      test_column.cohort_column.activate
+
+      expect(cohort.reload.column_state.map(&:class_name)).to include('CohortColumns::UserString1')
+    end
+
+    it 'handles multiple columns in column_state' do
+      # Add another column
+      another_column = build :user_string_cohort_column_2, cohort: cohort
```
eanders: Won't this already be there now since we're including all columns?

dtgreiner: It should, but it looks like the maintain! call either isn't happening in the github action test runs or something else is being messed. All of the failing tests are showing as if these records don't exist in the database. Looking into that some more, and then can also clean up this test.

db/warehouse/migrate/20250409145018_populate_cohort_column_table.rb
dtgreiner: If I'm [reading things right](https://github.com/greenriver/hmis-warehouse/blob/3ba0d8ad82c4ce3d1901a99743332281c819f344/config/deploy.rb#L176-L189), it looks like we are seeding after the migrations are run, so we wouldn't need this. I'm not 100% familiar with this process so hopefully I'm not looking at the wrong location.

eanders: That's the old capistrano deploy code.  I believe `roll_out.rb` calls `run_deploy_tasks!` which runs [deploy_tasks.sh](https://github.com/greenriver/hmis-warehouse/blob/6a566f71224609b4b623a4a4934176e30d2eedc3/bin/deploy_tasks.sh) which calls `db:seed` after migrations.  So same result, different path.

app/models/grda_warehouse/cohorts/cohort_column.rb
```diff
@@ -0,0 +1,167 @@
...
+    end
+
+    def remove_from_cohorts
+      GrdaWarehouse::Cohort.all.each do |cohort|
+        cohort.update!(column_state: cohort.column_state.reject { |col| col.cohort_column.class_name == class_name })
+      end
+    end
+
+    def self.maintain!
+      existing_class_names = all.pluck(:class_name)
```
eanders:
```suggestion
      existing_class_names = all.pluck(:class_name).to_set
```

will be significantly more performant

eanders (approved):
Thanks @dtgreiner, only one small tweak, then let's get this out for QA.

PR #5298, opened 2025-04-15
Add candidates_generated_at (buggy)

martha:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch
- use a merge-commit or rebase strategy for PRs targeting the stable branch

## Description
[//]: # (Summarize changes and include links related issue)
[//]: # (List any new dependencies or relevant ADRs)

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [ ] I have performed a self-review of my code
- [ ] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [ ] If adding a new endpoint / exposing data in a new way, I have:
  - [ ] ensured the API can't leak data from other data sources
  - [ ] ensured this does not introduce N+1s
  - [ ] ensured permissions and visibility checks are performed in the right places
- [ ] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) 
- [ ] I have updated the documentation (or not applicable)
- [ ] I have added spec tests (or not applicable)
- [ ] I have provided testing instructions in this PR or the related issue (or not applicable)

[//]: # NOTE: system tests may fail if there is no branch on the hmis-frontend that matches the Source  or Target branch of this PR. This is expected

PR #5275, opened 2025-04-07
Implement unit availability

martha:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

## Description

https://github.com/greenriver/hmis-frontend/pull/1128
https://github.com/open-path/Green-River/issues/7417

Summary of changes:
- Add polymorphic `owner` reference to Opportunity table. This will just be Units for now.
- Add MarkUnitsAvailable and MarkUnitsUnavailable mutations. These are capable of bulk update, even though that's not implemented in the UI yet, because I figured it would help avoid future schema changes
-  Add "latest opportunity" reference (and schema attribute) on Unit. This can be the currently open opportunity, or the most recently closed opportunity.
- Add "active" helper on Opportunity and Referral schema objects. I'm curious for feedback on these. My idea was to have the logic of "which statuses mean active?" in only one place (backend).
- When a referral generates an enrollment, if the opportunity's owner is a Unit, assign unit occupancy to the enrollment.

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)
New feature

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] If adding a new endpoint / exposing data in a new way, I have:
  - [x] ensured the API can't leak data from other data sources
  - [x] ensured this does not introduce N+1s
  - [x] ensured permissions and visibility checks are performed in the right places
- [x] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) 
- [x] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - yes, see frontend PR

[//]: # NOTE: system tests may fail if there is no branch on the hmis-frontend that matches the Source  or Target branch of this PR. This is expected

PR #5301, opened 2025-04-16
Add candidates_generated_at field

martha:
## _Merging this PR_
- Targets https://github.com/greenriver/hmis-warehouse/pull/5294; when both are approved, this can be squash-merged into there before that branch is squash-merged into the release branch.

## Description
- Add field `candidatesGeneratedAt` to Opportunity schema object
- Update the Opportunity schema object to resolve `referrals` _without_ the dataloader.
  - This currently isn't resolved in batch so it's a safe change
  - There was some buggy behavior arising related to the `reset` we are doing in `Sources::ActiveRecordAssociation`. [See demo PR](https://github.com/greenriver/hmis-warehouse/pull/5298/files) showing the buggy behavior. @ttoomey and I paired and determined that it needs further investigation, but for now, avoiding the dataloader is an acceptable workaround.
- Update field `candidates_generated_at` when candidate pool is calculated (previously wasn't getting set)

## Type of change
New schema field

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - yes, see frontend PR


martha:
Needs more thought:
> This currently isn't resolved in batch so it's a safe change

I only checked this with what's currently on `release-160`. I should have pointed out that https://github.com/greenriver/hmis-warehouse/pull/5275 introduces a change which resolves referrals on opportunities that are resolved on units, which does happen in batch.

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb
```diff
@@ -41,6 +41,7 @@
                 clientId
               }
             }
+            candidatesGeneratedAt
```
martha: `referral` starts resolving wrong (and tests in this file start failing) if `candidatesGeneratedAt` is moved down below `referral`. Relates to https://github.com/greenriver/hmis-warehouse/pull/5302

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -39,7 +40,7 @@ def referral
         :referrals,
         scope: Hmis::Ce::Referral.viewable_by(current_user).where.not(status: 'rejected').
           order(created_at: :desc),
-      ).first # there should be at most 1 (enforced in an AR validation but not at the database level)
+      ).first
```
gigxz: Per https://github.com/open-path/Green-River/issues/7571 lets take this opportunity to remove the scope from the referral resolver. I think something like this is sufficient:

```ruby
    def referral
      # TODO(#7395): permissions - ensure that user has permission to view referrals at this project
      # return nil unless current_permission?(permission: :can_view_referrals, entity: project)

      load_ar_association(object, :active_referrals)&.first # move active scope to a new active_referral association
    end
```


My understanding is that should fix the weirdness you were seeing with `candidates_generated_at`.

ttoomey: 👍  to removing the scope, maybe just needs an in-memory sort?
`load_ar_association(object, :active_referrals).sort_by(&:created_at).last

martha: Thanks, I think the sort is maybe even not necessary, with the `has_one :active_referral` and `unique_referral_per_opportunity` AR validator.
https://github.com/greenriver/hmis-warehouse/blob/release-161/drivers/hmis/app/models/hmis/ce/referral.rb#L64-L76

martha: Scratch that, sorry, this should return the accepted referral too (which is not considered active)

PR #5302, opened 2025-04-17
Preloader fix

ttoomey:
## _Merging this PR_

## Description
Remove association reset. This seems to have been causing unexpected issues with custom scopes (third arg to preload) being lost under some circumstances.

Trying to reproduce the behavior that the was originally patched in #4353 I could not longer replicate the extra query. Added a test against it;

## Testing
We should check that this doesn't cause a regression in performance due

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/graphql/sources/active_record_association.rb
+++ b/drivers/hmis/app/graphql/sources/active_record_association.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 ###
 # Copyright 2016 - 2025 Green River Data Analysis, LLC
 #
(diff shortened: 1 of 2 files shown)


gigxz (commented):
@ttoomey how can I reproduce the bug that this is supposed to fix? I'm not seeing differences between this branch and release-160 when resolving `scope`d data-loaded fields.

I did however identify another bug with how we use `scope`, that I described on https://github.com/open-path/Green-River/issues/7571. Maybe we can move discussion there

ttoomey:
I'm going to pull this into into the rails 7.1 upgrade branch; that will need a detailed QA pass anyway

PR #5304, opened 2025-04-17
Remove unused methods

gigxz:
## _Merging this PR_

## Description
Remove data-loader methods on concerns that are not used. It's a good thing we don't use them because of the bug described in https://github.com/open-path/Green-River/issues/7571 - which is maybe why we didn't use them in the first place. I don't know why they were added. Data loading is not needed for these concerns because we don't resolve them in batch.

## Type of change
Code clean-up

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/has_referral_requests.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/has_referral_requests.rb
@@ -21,12 +21,6 @@ module Types
         end
       end
 
-      # FIXME: combining the extra scope isn't working here- perhaps rails 6 bug
-      # def resolve_referral_requests_with_loader(association_name, **args)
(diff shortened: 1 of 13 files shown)


drivers/hmis/app/graphql/types/hmis_schema/has_referral_requests.rb
```diff
@@ -21,12 +21,6 @@ def referral_requests_field(name, description = nil, type: Types::HmisSchema::Re
         end
       end
 
-      # FIXME: combining the extra scope isn't working here- perhaps rails 6 bug
```
gigxz, line 24: This was the hint for https://github.com/open-path/Green-River/issues/7571!  😕

ttoomey (approved):
Thank you!

PR #5309, opened 2025-04-21
Add active_referrals assoc to unit

martha:
## _Merging this PR_

## Description
Follows up on code comment https://github.com/greenriver/hmis-warehouse/pull/5275#discussion_r2050596045

## Type of change
Code clean-up

- [x] I have added spec tests (or not applicable) - I haven't added anything, but confirmed https://github.com/greenriver/hmis-warehouse/pull/5275/files#diff-86e4f5da28e828555aab1ade51832c83207366979c4b03044f64668ed84288c0 still passes so I think that should exercise this


drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -26,8 +26,9 @@ class Hmis::Unit < Hmis::HmisBase
   has_many :current_occupants, through: :active_unit_occupancies, class_name: 'Hmis::Hud::Enrollment', source: :enrollment
   has_many :opportunities, as: :owner, class_name: 'Hmis::Ce::Opportunity', inverse_of: :owner, dependent: :destroy
   # This is the latest opportunity, which could be either active or closed with an accepted referral.
-  # This relies on the assumption that if there is an active opportunity, it should be the latest one.
   has_one :latest_opportunity, -> { order(created_at: :desc, id: :desc) }, as: :owner, class_name: 'Hmis::Ce::Opportunity', inverse_of: :owner
+  # There should really be only ONE active referral for a unit, not many. It's a has_many relationship though because the data model doesn't disallow multiple.
+  has_many :active_referrals, through: :opportunities, class_name: 'Hmis::Ce::Referral', source: :active_referral
```
martha: I've added the `has_many` relationship and just added a comment to hopefully clear up confusion about the idea that there would be multiple active referrals for a unit at once.

I also thought about trying to add a validation so this just wouldn't happen, but don't feel clear exactly on how that would happen -- it would have to be a validation that runs on save referral, I think. Let me know if you think that's worthwhile to pursue.

gigxz: Yes I think its worthwhile to add the necessary validations to ensure that:
- There can be at most 1 non-`rejected` referral per Opportunity (this validation already exists, see `validate :unique_referral_per_opportunity`)
- There can be at most 1 non-`closed` Opportunity per ~~Unit~~ Owner (I think this is missing, we should add a validation to the Opportunity model)

I think once the second validation is in place, it would be safe to update this relationship to `has_one :active_referral`.

For good measure  I think we can also go ahead and add a `has_many :referrals` to retrieve all historical referrals that the unit has had, for all opportunities. We will probably want to display that at some point, and it will be helpful for debugging. I think that will be `has_many :referrals, through: :opportunities, class_name: 'Hmis::Ce::Referral'`.

drivers/hmis/app/graphql/types/hmis_schema/unit.rb
```diff
@@ -68,7 +68,7 @@ def accepting_ce_referrals
       latest_opportunity = load_ar_association(object, :latest_opportunity)
       return false if latest_opportunity.nil?
```
gigxz: this can be removed now, right?

martha: I think `accepting_ce_referrals` should still return `false` if there is no opportunity. In other words, even if `active_referral` is nil, the unit may still _not_ be accepting referrals, if it doesn't have an opportunity at all.

martha: I added some comments along these lines -- let me know if this makes sense to you!

gigxz: Oh I see, thanks.

I think we still have the problem of erroneously returning `true` if the unit only has _closed_ opportunities, though. Should this be:
```
return false if latest_opportunity.nil? || latest_opportunity.closed?
```
?

I'm thinking that if the unit only has closed opportunities, active_referral would be nil. But that doesn't mean that it is accepting new ce referrals.

martha: Ah yeah absolutely right, thanks for catching that.

drivers/hmis/app/models/hmis/unit.rb
martha: Thanks @gigxz! I'm re-requesting your review, especially eager for feedback about the preloading expectation in the validation. (https://github.com/greenriver/hmis-warehouse/pull/5309/files#diff-ba111f50055a59ac1e343e014bcb65a4c50feb17488e47064e2a8adbaa359d36R89-R92)
- Pro: Avoids n+1
- Con: Brittle, requires that the unit's `opportunites` scope is already loaded. I needed to add `unit.reload` on the tests here to get them passing: https://github.com/greenriver/hmis-warehouse/pull/5309/files#diff-aa816e925640f2be62bcc0bc88e2037a2db7c51c99f0ffee8fe3c2f6c7499559R118

Since we are not currently marking units available in batch, maybe this isn't worthwhile? Lmk what you think!

drivers/hmis/app/graphql/types/hmis_schema/unit.rb
```diff
@@ -65,10 +65,12 @@ def latest_opportunity
...
 
     def accepting_ce_referrals
+      # First check for an existing opportunity. If there is none, then this unit isn't accepting referrals
       latest_opportunity = load_ar_association(object, :latest_opportunity)
-      return false if latest_opportunity.nil?
+      return false if latest_opportunity.nil? || latest_opportunity.closed?
 
-      load_ar_association(latest_opportunity, :referrals, scope: Hmis::Ce::Referral.active).empty?
+      # Otherwise, the unit is only accepting referrals if the opportunity doesn't already have an active referral.
+      load_ar_association(object, :active_referral).nil?
```
martha, line 74: Maybe clearer to use `active_or_accepted_referral` here, even though it should never be `accepted` because we have already filtered out `closed` opportunities?

martha: I added some clarifying comments, I think it's okay to leave as is.

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -67,8 +68,31 @@ class Opportunity < GrdaWarehouseBase
...
     end
+
+    private
+
+    def unique_opportunity_per_unit
+      return if status.to_sym == :closed || owner.nil?
+
+      # This validator expects that owner's opportunities are preloaded, to avoid n+1 on save
+      conflicting_opportunity_exists = owner.opportunities.to_a.select do |existing_opp|
+        existing_opp.status != 'closed' && existing_opp.id != id
```
gigxz: do we want a to_sym or to_s here on `opp.status`, or are we sure its always a string?

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -24,10 +24,21 @@ class Hmis::Unit < Hmis::HmisBase
...
+
+  # A unit may have many historical opportunities (which represent past times when this unit was available and then filled)...
   has_many :opportunities, as: :owner, class_name: 'Hmis::Ce::Opportunity', inverse_of: :owner, dependent: :destroy
-  # This is the latest opportunity, which could be either active or closed with an accepted referral.
-  # This relies on the assumption that if there is an active opportunity, it should be the latest one.
-  has_one :latest_opportunity, -> { order(created_at: :desc, id: :desc) }, as: :owner, class_name: 'Hmis::Ce::Opportunity', inverse_of: :owner
+  # ...but it only has one "latest" opportunity, which could be either:
+  # - active and accepting referrals (open),
+  # - active with a referral in-progress (locked), or
+  # - closed with an accepted referral. This would be prioritized last, after any active opportunity.
```
gigxz, line 33: thanks for the explainer

gigxz (approved):
thanks!

PR #5313, opened 2025-04-23
HUD Report Cell Value - Account for Nil Values

dtgreiner:
## _Merging this PR_

## Description

The method fetching cell values for HUD report details was hitting an exception when reaching a nil value within the method chain being parsed from the dot notation column string. This will now return nil when this occurs.

## Type of change

Bug fix


app/models/hud_reports/report_client_base.rb
```diff
@@ -29,7 +29,9 @@ def fetch_cell_value(col)
       # This will separate the object and method in order to call the appropriate column/method
       # while similarly allowing the method to be called if the column is not passed using dot notation.
       col.to_s.split('.').inject(self) do |obj, method|
-        if obj.class.column_names.include?(method)
+        return nil if obj.nil?
```
eanders: I guess we'd usually write this:
```suggestion
        return unless obj
```

They should accomplish the same thing, but the above will also catch when obj is just empty or blank.

eanders: Maybe more to the point, why is obj empty here?

dtgreiner: It's parsing through the dot notation string (e.g. 'return_enrollment.enrollment.project.project_name') but there is a nil along the way due to the `inject(self)`. For example, if there is no `return_enrollment`, the first call `self.return_enrollment` will return nil which is then checked as `nil.enrollment` which is causing this failure.

eanders (approved):

PR #5317, opened 2025-04-24
Add analytics view for scan card codes

gigxz:
## _Merging this PR_

## Description
* Add view for scan card codes https://github.com/open-path/Green-River/issues/7610
* Included deleted client alerts in the existing view. Deleted alerts are relevant for reporting purposes to understand how alerts are used over time.
* Include deleted staff assignments in existing view. Deleted staff assignments are relevant to understanding caseload and who was assigned to what. (Assignment may change during the course of an enrollment).

For the altered views -- they are not yet being consumed. Issue for that is https://github.com/open-path/Green-River/issues/7577.

## Type of change
analytics view


db/views/analytics_hmis_scan_cards_v01.sql
```diff
@@ -0,0 +1,3 @@
+SELECT * FROM "hmis_scan_card_codes"
```
eanders: We've been explicitly specifying all fields so we can ensure the shape doesn't change without our knowledge.  Maybe enumerate the columns here?

eanders (approved):
Looks good @gigxz, but to be consistent, we should probably not do `select *`.

ttoomey (approved):
Looks good!

+1 to avoiding `SELECT *` in views

PR #5320, opened 2025-04-26
fix possible replication issue with analytics.app_users

ttoomey:
## _Merging this PR_

## Description
Fixes crash when trying to prune deleted users from app_users table

## Type of change
Bug fix


db/warehouse/migrate/20250426023314_set_replica_for_analytics_app_users.rb
```diff
@@ -0,0 +1,7 @@
+class SetReplicaForAnalyticsAppUsers < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
+      execute('ALTER TABLE analytics.app_users REPLICA IDENTITY FULL')
```
eanders, line 7: Is this somehow related to publication?
https://www.postgresql.org/docs/current/logical-replication-publication.html

How does it solve the issue?

ttoomey: Sorry, there's more info in the [linked issue](https://github.com/open-path/Green-River/issues/7600). My understanding is yes, it is related to the replication or publication mechanism. But I couldn't find the relevant config when poking around on postgresql prod; hopefully we can check in on Monday standup

The statement here came from the hint in the error message; my understanding is that this statement sends the whole row to the log when changes occur which is a requirement for replication. If it seems reasonable I might try adding it by hand to one of the impacted dbs to see if it resolves the error

ttoomey:
going to connect with devops on this first

ttoomey:
this is low priority, need to connect with devops to figure out the underlying cause

eanders:
@ttoomey do you think this is still necessary?  If not, can we close it?

ttoomey:
> @ttoomey do you think this is still necessary? If not, can we close it?

yes, it's not really resolved but low priority. 

made a new issue and plunked it in the icebox. Closing this for now
https://github.com/open-path/Green-River/issues/8423

PR #5323, opened 2025-04-26
7551-fix-double-render

ttoomey:
## _Merging this PR_

## Description
Fix double-render error when accessing client history. 

### Details:
Here's my understanding of what's happening: The history controller calls `require_can_see_this_client_demographics?` which calls `set_client`. The code expects set_client to return falsy if the client is not found. However it `set_client` returns truthy if it redirects causing the `require_can_see_this_client_demographics` to raise `not_authorized` which performs a second redirect, resulting in a double render

### testing
Log in, visit a client history page for a merged client (see linked issue for an example). You should be redirected to the most recent client

## Type of change
bug fix

### DIFF
--- a/drivers/client_access_control/app/controllers/client_access_control/history_controller.rb
+++ b/drivers/client_access_control/app/controllers/client_access_control/history_controller.rb
@@ -1,3 +1,5 @@
+# frozen_string_literal: true
+
 ###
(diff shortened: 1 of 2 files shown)


eanders (approved):
Thanks @ttoomey !

PR #5324, opened 2025-04-27
7572 validate date format

ttoomey:
## _Merging this PR_

## Description
Stricter date parsing in filter params, raise an exception for ambiguous date formats. See linked issue for exception details; it looks like there might be an ambiguous param for the `:on` param in the exception. This PR is a bit of a guess at a solution, I couldn't get the form date-picker to actually send and invalid format.

## Type of change
Bug fix

- [x] added tests

### DIFF
--- a/app/models/filters/filter_base.rb
+++ b/app/models/filters/filter_base.rb
@@ -1634,6 +1634,8 @@ module Filters
       case val.presence
       when Date, nil
         return val
+      when DateTime, Time, ActiveSupport::TimeWithZone
+        return val.to_date
       when String
         return Date.strptime(val, '%b %d, %Y') if val.match?(/\A\w{3} +\d{1,2}, +\d{4}\z/)
         return Date.strptime(val, '%Y-%m-%d') if val.match?(/\A\d{4}-\d{2}-\d{2}\z/)
--- a/app/models/filters/filter_base.rb
+++ b/app/models/filters/filter_base.rb
@@ -105,9 +105,9 @@ module Filters
 
       filters = filters.to_h.with_indifferent_access
 
-      self.on = filters.dig(:on)&.to_date || on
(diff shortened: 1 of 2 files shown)


app/models/filters/filter_base.rb
```diff
@@ -1611,5 +1622,25 @@ def to_comparison
         [start_date, end_date]
       end
     end
+
+    # Strict date parser for filter params.
+    # Accepts:
+    #   - Date objects (returns as-is)
+    #   - nil (returns nil)
+    #   - Strings in 'Mon dd, yyyy' (e.g. 'Apr 27, 2025')
+    #   - Strings in 'YYYY-mm-dd' (e.g. '2025-04-27')
```
eanders, line 1631: Can we make it accept and correctly process `mm/dd/yyyy`?  
I think sometimes people type the date in and somehow bypass the JS that should reformat to our format, but given that HMIS is only relevant in the US, it's probably ok to assume they don't mean dd/mm/yyyy.

eanders:
@ttoomey what do you mean by: **ambiguous param for the :on param in the exception**?  We don't always send that, but some reports only require a single date and often that is the `on` date.

eanders (approved):
This looks good to me, though if we can support the US standard date format, let's do it.

ttoomey:
> what do you mean by: ambiguous param for the :on param in the exception?

The localized US date format is ambiguous with UK dates ('01/12/2025' could be Jan 12th or Dec 1st). Date.parse treats it as the later, which would be incorrect in our case. I suggest not handling these formats to avoid the whole problem of ambiguity. I think the original exception is because the date in the params is US-formatted but is invalid when parsed as the UK format (for example "01/30/2025"). I'm not sure how a date in the US format was introduced

if we want to allow us-dates, we need to also fix how we parse the dates as currently they would be incorrect.

PR #5326, opened 2025-04-28
7604-fix-page-titles

ttoomey:
## _Merging this PR_

## Description
* remove client and user names from page titles
* add render_page_title helper
* previously client names were only visible based on searchable permission. In the dash context we also include names for visible clients
* add test for GrdaWarehouse::SourceClientViewAccessor
* adjust client dash tests to reflect that we show source client names in the alias


eanders (approved):

app/views/clients/coordinated_entry_assessments/edit.haml
```diff
@@ -1,13 +1,11 @@
-- title = "Editing #{_'Coordinated Entry Assessment'} for #{@client.pii_provider(user: current_user).full_name}"
+- title = "Editing #{_'Coordinated Entry Assessment'}"
```
eanders: These look like some translations we missed.  I _think_ this should be 
```suggestion
- title = "Editing #{Translation.translate('Coordinated Entry Assessment')}"
```

app/views/clients/coordinated_entry_assessments/new.haml
```diff
@@ -1,13 +1,11 @@
-- title = "New #{_'Coordinated Entry Assessment'} for #{@client.pii_provider(user: current_user).full_name}"
+- title = "New #{_'Coordinated Entry Assessment'}"
```
eanders: here too

app/views/clients/coordinated_entry_assessments/show.haml
```diff
@@ -1,13 +1,11 @@
-- title = "#{_'Coordinated Entry Assessment'} for #{@client.pii_provider(user: current_user).full_name}"
+- title = _'Coordinated Entry Assessment'
```
eanders: and here

app/views/clients/users/index.haml
```diff
@@ -1,4 +1,4 @@
-- title = "Assigned Relationships for #{@client.pii_provider(user: current_user).full_name}"
+- title = "Assigned Relationships"
```
eanders, line 1: Should this have an `H1` like the other pages?

app/views/clients/coordinated_entry_assessments/show.haml
```diff
@@ -1,13 +1,11 @@
...
 - content_for :title, title
 .o-page
   = render 'clients/breadcrumbs'
   = render 'clients/aliases'
 
 .o-page__header
-  .o-page__title
-    %h1= title
-
+  = render_page_title "#{_'Coordinated Entry Assessment'} for #{@client.pii_provider(user: current_user).full_name}"
```
eanders: Should these be:
```suggestion
  = render_page_title "#{title} for #{@client.pii_provider(user: current_user).full_name}"
```

just so we don't accidentally get a mis-match in the future?

app/helpers/application_helper.rb
```diff
@@ -527,6 +527,16 @@ def bracket_small_population(value, mask: true)
     end
   end
 
+  # page title with optional block
+  # = render_page_title "New Report" do
+  #   = render 'type_warning'
+  def render_page_title(title, &block)
```
eanders: maybe this should be:
```suggestion
  def render_page_heading(title, &block)
```
Just to differentiate from the title tag which is what we're fixing?

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
```diff
@@ -708,7 +708,7 @@ def doc_has_client?(doc, warehouse_client)
       it 'user can see any clients' do
         get client_path(non_window_destination_client)
         doc = Nokogiri::HTML(response.body)
-        expect(doc.text).to include('Bob Ross')
+        expect(doc.text).to include(non_window_destination_client.id.to_s)
```
eanders: It looks like the `can_view_clients` role 
```
let!(:can_view_clients) { create :role, can_view_clients: true, can_view_client_name: true }
```
includes can view client name, so we might want to confirm the name shows on the page.

drivers/financial/app/views/financial/clients/show.haml
```diff
@@ -1,4 +1,4 @@
-- title = @client.pii_provider(user: current_user).full_name
+- title = 'Financial Clients'
```
eanders: Maybe?
```suggestion
- title = 'Client Finances'
```

We call the records `financial_clients` but I don't think that makes sense when looking at the page.

app/views/clients/coordinated_entry_assessments/edit.haml
ttoomey: thanks! - sounds like there's more to unpack around how we use translations, seems like an area for future documentation.

app/views/clients/users/index.haml
ttoomey: I can add one, I was trying not change the existing body html

eanders: Should be fine for now then.  Let's leave it as is

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
ttoomey: yes it doesn't seem to render the name in the page, only in the title. So getting the test to pass requires modifying the view behavior to include the client name

drivers/financial/app/views/financial/clients/show.haml
ttoomey: ah okay, I'll fix up the h3 in this page to match

app/helpers/application_helper.rb
ttoomey: good call, how about render_display_title

app/views/clients/users/index.haml
ttoomey: thanks I'll leave it to limit scope. It looks like none of the dash tabs have an h1

app/views/clients/coordinated_entry_assessments/show.haml
ttoomey: I guess I would lean towards keeping them independent although it is less DRY. I feel like drying it up would be done in another pass and leaving it ugly might make future refactor easier

ttoomey: I guess I would lean towards keeping them independent although it is less DRY. I feel like drying it up would be done in another pass and leaving it ugly might make future refactor easier

app/models/grda_warehouse/source_client_view_accessor.rb
```diff
@@ -15,6 +15,38 @@
 #   end
 #
 class GrdaWarehouse::SourceClientViewAccessor
+  class Cache
```
eanders, line 21: @ttoomey can you add some code comments here to explain what this is doing?  
This feels like a fair amount of abstract logic to shove into a class inside of a class.  Does it make sense to pull this out to its own file?  (Or can we declare it private in some way?)

ttoomey: good suggestion, added docs. I guess I would leave that as an interior class if that's not too ugly.

I also replaced the method with a block pattern to make it a little cleaner and hopefully easier to understand

eanders: Thanks!  I think that change makes this much less daunting.

eanders (approved):

PR #5330, opened 2025-04-29
Rollback redis pooling

ttoomey:
rollback redis polling change to unblock rails deployment. Looks like the notifier needs additional adjustments and tests


dtgreiner (approved):

dtgreiner:
Looks good to me, outside of a rubocop complaint

PR #5331, opened 2025-04-29
Replace strict attribute casting with validations

martha:
## _Merging this PR_

## Description
- Re-add the `WithStrictAttributes` concern to `Hmis::Hud::Concerns::Shared`, and remove the special-case on `IncomeBenefit`.
- Update the concern to use validations instead of casting, so that search functionality (`where`) won't be impacted.
  - Use a custom validation that is more permissive than Rails's default `numericality` validator.
- Remove now-unused `StrictDecimal` and `StrictInteger` types.
- Add spec, including regression test for partial SSN searching functionality

## Type of change


ttoomey (approved):
This looks great, thank you!!

drivers/hmis/app/models/hmis/hud/concerns/with_strict_attributes.rb
```diff
@@ -6,22 +6,43 @@
...
+
+    # Cast the value to a string and check it against the custom regex above
+    regex = options[:integer] ? INT_RGX : DEC_RGX
+    return if raw_value.to_s.match?(regex)
+
+    record.errors.add(attribute, :not_a_number, message: 'is not a number')
+  end
+end
+
 module Hmis::Hud::Concerns::WithStrictAttributes
```
gigxz, line 31: This concern should probably be renamed to something more descriptive now, like `WithFieldTypeValidation`?

gigxz (approved):
looks good! 👍

PR #5337, opened 2025-04-30
Add CandidatePoolBuilderJob

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add `CandidatePoolBuilderJob` which builds candidate pools and runs the match engine.
- Queue the job when a unit is marked available.
- Queue the job on a schedule at 5am daily.

How to test:
- Mark a unit available locally.
- In a local rails console, run `Delayed::Worker.new.work_off`
- You should be able to visit the Opportunity page and see eligible clients there a few moments later

## Type of change
New feature


lib/tasks/grda_warehouse.rake
```diff
@@ -330,6 +330,11 @@ namespace :grda_warehouse do
       Hmis::AutoExitJob.perform_now
     end
 
+    if DateTime.current.hour == 5 && HmisEnforcement.hmis_enabled? && GrdaWarehouse::DataSource.hmis.exists? && Hmis::Ce.configuration.enabled?
```
martha, line 333: Question: How to test this change? Just on QA, or is there anything I can verify locally to make sure the change is good?

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -37,6 +37,10 @@ def resolve(unit_ids:)
         Hmis::Ce::Opportunity.import!(opportunities)
       end
 
+      job_class = Hmis::CandidatePoolBuilderJob
+      job_name = job_class.name
+      Delayed::Job.enqueue job_class.new unless Delayed::Job.queued?(job_name) || Delayed::Job.running?(job_name)
```
eanders: I think this should include the queue:

```suggestion
      Delayed::Job.enqueue(job_class.new, queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)) unless Delayed::Job.queued?(job_name) || Delayed::Job.running?(job_name)
```

drivers/hmis/app/jobs/hmis/candidate_pool_builder_job.rb
```diff
@@ -0,0 +1,28 @@
...
+  class CandidatePoolBuilderJob < BaseJob
+    include NotifierConfig
+
+    def perform
+      log("Building candidate pools for #{Hmis::Ce::Opportunity.active.count} active opportunities")
+      Hmis::Ce::Match::CandidatePoolBuilder.new.perform
+
+      log("Running the CE match engine for #{Hmis::Hud::Client.hmis.count} HMIS clients")
+      clients = Hmis::Hud::Client.hmis
+      Hmis::Ce::Match::CandidatePool.all.each do |pool|
```
eanders: If we expect candidate pool to be large in the future, a find_each might be warranted
```suggestion
      Hmis::Ce::Match::CandidatePool.all.find_each do |pool|
```

drivers/hmis/spec/requests/hmis/ce/mark_units_available_spec.rb
```diff
@@ -53,7 +53,8 @@
           expect(response.status).to eq(200), result.inspect
           expect(result.dig('data', 'markUnitsAvailable', 'units', 0, 'latestOpportunity', 'name')).to eq("Unit #{unit.id} - #{unit_type.description}")
           unit.reload
-        end.to change(Hmis::Ce::Opportunity, :count).by(1)
+        end.to change(Hmis::Ce::Opportunity, :count).by(1).
+          and change(Delayed::Job.jobs_for_class('CandidatePoolBuilderJob'), :count).from(0).to(1)
```
eanders: Depending on how jobs are being excecuted in the test environment, this check might fail intermittently.  (If the worker is running in synchronous mode.)

lib/tasks/grda_warehouse.rake
eanders: QA is probably the easiest, this config should be fine.  The only thing I might change, is that if you expect `perform_now` to take a long time to run below you should queue it with `perform_later` so it doesn't prevent other things from running.

drivers/hmis/spec/requests/hmis/ce/mark_units_available_spec.rb
martha: Thanks! Should I just remove it?

eanders: I've been fighting for a couple days with delayed jobs/active jobs and rspec.  I actually think our default behavior is to _not_ run/process delayed jobs, so I think you're ok.  

If you want to check the results of the job, I believe you'll need to call `Delayed::Worker.new.work_off`.  So I think you're good.  Sorry for the confusion.

drivers/hmis/app/jobs/hmis/candidate_pool_builder_job.rb
```diff
@@ -0,0 +1,28 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Hmis
+  class CandidatePoolBuilderJob < BaseJob
+    include NotifierConfig
+
```
gigxz: Maybe @eanders can confirm if it's best practice, but if you add
```
queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
```
It should default to queueing as long-running, which I think we want.

drivers/hmis/app/jobs/hmis/candidate_pool_builder_job.rb
```diff
@@ -0,0 +1,28 @@
...
+    def perform
+      log("Building candidate pools for #{Hmis::Ce::Opportunity.active.count} active opportunities")
+      Hmis::Ce::Match::CandidatePoolBuilder.new.perform
```
gigxz: I don't think we need to regenerate _all_ candidate pools when a new opportunity is created. That will be very slow. This wasn't clearly spec'd, so maybe we should discuss, but I would expect:
* When a new opportunity is created, we only need to regenerate the candidate pool _for that opportunity "group"._ Other candidate pools should **not** be affected by the creation of this opportunity, so there is no need to regenerate them.
* Similarly, we shouldn't need to re-run the CE Match engine for **all** candidate pools below. We should only re-run it for the affected candidate pool(s).


We do want to regenerate everything on the daily schedule. But if this job optionally accepted opportunity ids, it could be enqueued in a more limited scope from MarkUnitsAvailable.

martha: Thanks Gig!

I haven't pursued the lifecycle hooks strategy we talked about briefly at standup, since the records are currently `import!`ed. Let me know if you think that's not worth it.

drivers/hmis/spec/requests/hmis/ce/mark_units_available_spec.rb
martha: Hm, thanks! I think I do need some help with this one. Now that I'm enqueuing the job with `perform_later` (because I want to pass arguments), this assertion no longer works. Any ideas about this? Is such an assert even worthwhile? I can keep digging around in the documentation, if you're not sure. Thank you 🙏

eanders: I believe  for ActiveJob jobs (ie. `perform_later`) you want to wrap it in a block:
```
perform_enqueued_jobs do
  ...
end
```

martha: Thanks, great!

drivers/hmis/app/jobs/hmis/candidate_pool_builder_job.rb
gigxz: Yep that sounds right to keep it in the mutation. Best to avoid lifecycle hooks when  possible.

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -34,7 +34,9 @@ def resolve(unit_ids:)
           build_opportunity_for_unit(unit, template)
         end
 
-        Hmis::Ce::Opportunity.import!(opportunities)
+        result = Hmis::Ce::Opportunity.import!(opportunities)
+        opportunity_ids = result.ids
+        Hmis::MatchCandidatesJob.perform_later(opportunity_ids: opportunity_ids, backoff_time: 24.hours)
```
gigxz: We should be able to synchronously link this Opportunity to an _existing_ Candidate Pool,  if one exists that matches this opportunity's eligibility and prioritization expressions. That way, the candidate pool would appear immediately when the opportunity is created.

I think we determined during the discussion that we only need to enqueue the MatchCandidatesJob if any of the new opportunities **can't** be mapped to existing pools.

here is some pseudo code :


```ruby
# create initial opportunities (not linked to candidate pools)
opportunities = units.map do |unit|
    build_opportunity_for_unit(unit, template)
end

# group opportunities by priority+requirements which determines candidate pool
# { [priority_expression, requirement_expression ] => [opportunities] }
opportunities_by_key = #... refactor form CandidatePoolBuilder._perform to shared code

# { [priority_expression, requirement_expression ] => candidate pool }
current_pools =  # .. refactor from CandidatePoolBuilder to shared code

# link oppts to existing candidate pools, if present
opportunities_by_key.each do |key, opportunities|
  pool = current_pools.fetch(key)

  if pool
     opportunities.each { |o| o.candidate_pool = pool }
  else
    enqueue_candidate_pools_oppt_ids << opportunities.map(&:id)
   end
end

# import opportunities

# enqueue job for enqueue_candidate_pools_oppt_ids if any

# resolve

```

drivers/hmis/app/jobs/hmis/match_candidates_job.rb
```diff
@@ -0,0 +1,43 @@
...
+      opportunities = opportunity_ids ? Hmis::Ce::Opportunity.where(id: opportunity_ids) : Hmis::Ce::Opportunity.active
+      log("Building candidate pools for #{opportunities.count} opportunities")
+      Hmis::Ce::Match::CandidatePoolBuilder.new(opportunities).perform
+
+      # Find candidate pools for these opportunities
+      candidate_pool_ids = opportunities.reload.pluck(:candidate_pool_id).compact.uniq
+      candidate_pool_scope = Hmis::Ce::Match::CandidatePool.where(id: candidate_pool_ids)
+      # If backoff_time was provided, only call the match engine for candidate pools that have not been updated since backoff_time
+      cp_t = Hmis::Ce::Match::CandidatePool.arel_table
+      candidate_pool_scope = candidate_pool_scope.where(cp_t[:candidates_generated_at].lt(Time.current - backoff_time)) if backoff_time
```
gigxz: Could add a scope candidate_pool_ids so you can do 
```
Hmis::Ce::Match::CandidatePool.where(id: candidate_pool_ids).candidates_generated_before(datetime)
```
That scope might be helpful elsewhere.

drivers/hmis/spec/models/hmis/ce/match_candidate_pool_builder_spec.rb
```diff
@@ -3,10 +3,10 @@
 require 'rails_helper'
 
 RSpec.describe Hmis::Ce::Match::CandidatePoolBuilder do
-  let(:builder) { described_class.new }
-  let!(:opportunity) { create(:hmis_ce_opportunity) }
-  let(:project) { opportunity.project }
-  let(:organization) { project.organization }
+  let(:organization) { create :hmis_hud_organization }
+  let(:project) { create :hmis_hud_project, organization: organization }
+  let!(:opportunity) { create(:hmis_ce_opportunity, project: project) }
```
gigxz: is there a reason not to immediately create the organization and project using `let!` as opposed to `let`? I've seen this type of thing lead to some tricky-to-identify bugs.

gigxz: It is probably fine since opportunity uses `!` and references project, actually. I always found this distinction confusing, so I tend to always use the !

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
martha: Shared code (for key generation and etc.) can live in class methods on candidate pool builder where they live, and expose them

martha: Thanks @gigxz I updated this to follow your idea, not exactly following your pseudocode since now the `build_opportunity_for_unit` is responsible for saving the existing candidate pool if there is one. Let me know how it looks to you.

drivers/hmis/app/jobs/hmis/match_candidates_job.rb
martha: I ended up removing this since the `mark_units_available.rb` procedure to check for opportunities that are "stale" was using similar logic, and I didn't think we needed it in both places. Also, I wasn't sure how to incorporate the scope you're suggesting on CandidatePool with the arel logic on Opportunity, but let me know if I should pursue that further

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
martha: oh, also it felt a bit more graceful to me to put some of the shared code in a separate class, so I made `CandidatePoolResolver` (for "resolving" the candidate pool for a given opportunity). But let me know if you think it's ok how I've done this, I'd be glad for feedback on naming or structure

gigxz (approved):

PR #5338, opened 2025-05-01
965 user text search

ttoomey:
## _Merging this PR_

## Description
* Fixed search to work across first, last, and email
* added trigram extension to app db for ranking
* Search results can be sorted by relevance or manual sort order

## Setup Prerequisites
- Access to admin user search page
- Multiple test users with varied names/emails in the system

## Test Cases
On the admin users page, test these searches
1. **Exact Matches**
   - Search for a complete first name (e.g. "John")
   - Search for a complete last name (e.g. "Smith")
   - Search for a complete email (e.g. "john.smith@example.com")

2. **Prefix Matches**
   - Search for partial first names (e.g. "Joh" should find "John", "Joseph")
   - Search for partial last names (e.g. "Smi" should find "Smith", "Smits")
   - Search for partial emails (e.g. "john.s" should find "john.smith@example.com")

3. **Sort Order Testing**
   - Use column headers to sort results by:
     - First name
     - Last name
     - Email
   - Verify both ascending and descending sort works

4. **Edge Cases**
   - Search with mixed case (e.g. "jOhN")
   - Search with extra spaces
   - Search with empty string, should return all results
   - Search with very short strings (1-2 characters), should return no results
   - Search with special characters in names/emails

## Expected Behavior
- Results page should load quickly
- Sorting by best match is the default, but sorting by a column overrides sort  by rank

## Type of change
Bug fix, New feature


db/migrate/20250426181347_add_search_vector_to_users.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AddSearchVectorToUsers < ActiveRecord::Migration[7.0]
```
ttoomey: not strictly necessary as search will still work without indexes

eanders (approved):
@ttoomey this looks great to me.  I haven't tried to run it, but the theory seems solid.  Thanks!

Maybe eventually we can integrate some of this into the client search as well.

ttoomey:
Great, I had second thoughts about the fuzzy search and took it out as it could be confusing for someone checking for a specific email or name

app/models/concerns/user_concern.rb
```diff
@@ -470,15 +470,45 @@ def deactivation_status(user)
...
+    # @param sort_by_best_match [Boolean] whether to order results by similarity to query
+    # @param similarity_threshold [Float] minimum similarity score (0.0 to 1.0) for fuzzy matches
+    #
+    # @return [ActiveRecord::Relation] matching users
+    #
+    # @example Find users matching 'john'
+    #   User.text_search('john')
+    #
+    # @example Find users with similarity ordering
+    #   User.text_search('john smith', sort_by_best_match: true)
```
eanders: You pulled out the fuzzy match and `similarity_threshold` right?
```suggestion
    # Search for users by name or email using both prefix matching
    #
    # @param text [String] the search query
    # @param sort_by_best_match [Boolean] whether to order results by similarity to query
    #
    # @return [ActiveRecord::Relation] matching users
    #
    # @example Find users matching 'john'
    #   User.text_search('john')
    #
    # @example Find users with similarity ordering
    #   User.text_search('john smith', sort_by_best_match: true)
```

eanders (approved):

app/models/concerns/user_concern.rb
ttoomey: yes, thanks for catching the stale docs

PR #5342, opened 2025-05-01
Add `request_id` to HMIS ActivityLogs table, fix session_id on versions

gigxz:
## _Merging this PR_

## Description
1. Add `request_id` to HMIS Activity Log so we can correlate them with PaperTrail versions for debugging/support https://github.com/open-path/Green-River/issues/7639
2. Fix `session_id` on Warehouse and HMIS papertrail `versions` table. It was always being set to nil. Now it is set to the session hash. (It was previously set by `request.env['rack.session.record']&.session_id,` and the code hadn't changed in 8 years, so I assume it was just outdated.)

## Type of change
new feature, bug fix

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -137,7 +137,7 @@ class ApplicationController < ActionController::Base
   def info_for_paper_trail
     {
       user_id: warden&.user&.id,
-      session_id: session.id&.to_s,
+      session_id: session&.id&.to_s,
       request_id: request.uuid,
     }
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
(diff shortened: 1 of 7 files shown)


app/controllers/application_controller.rb
```diff
@@ -135,7 +137,7 @@ def append_info_to_payload(payload)
   def info_for_paper_trail
     {
       user_id: warden&.user&.id,
-      session_id: request.env['rack.session.record']&.session_id,
+      session_id: session&.id&.to_s,
```
eanders, line 140: are we sure this should be an integer?

eanders (approved):
Thanks @gigxz !

app/controllers/application_controller.rb
gigxz: It's a UUID. I just copied the `to_s` since Theron had that on the `GraphqlController` (when we set this field on activity logs) 👀

eanders: ok, weird that it's in `.id` but I'm good if the value is actually a UUID.

ttoomey (approved):
Looks good, thank you!

PR #5352, opened 2025-05-05
Decreasing Sequences -> Increase

dtgreiner:
## _Merging this PR_

## Description

Locate and update any decreasing sequences to increase instead now that we have the keys updated to larger integers.

## Type of change
Code clean-up


eanders (approved):
Looks good to me, thanks!

db/warehouse/migrate/20250502193442_decreasing_sequence_keys.rb
```diff
@@ -0,0 +1,34 @@
...
+  def up
+    sql = <<-SQL
+      SELECT
+        tbls.table_name, increment, seq.sequence_name
+      FROM
+        information_schema.tables AS tbls
+        INNER JOIN information_schema.columns AS cols ON tbls.table_name = cols.table_name
+        INNER JOIN information_schema."sequences" AS seq on seq.sequence_name = concat(tbls.table_name, '_id_seq')
+      WHERE
+        tbls.table_catalog = 'development_openpath_warehouse'
```
yakloinsteak: I don't think this is right. Maybe you could filter on increment equal to -1 instead?

db/warehouse/migrate/20250502193442_decreasing_sequence_keys.rb
```diff
@@ -0,0 +1,34 @@
...
+        AND tbls.table_schema = 'public'
+        AND cols.column_name = 'id'
+        AND cols.table_name NOT IN ('ClientUnencrypted', 'Site', 'bi_data_sources', 'bi_lookups_ethnicities', 'bi_lookups_funding_sources')
```
yakloinsteak: Why are these excluded? I was thinking it was issues with quoting below, but the `bi` ones wouldn't have that problem.

yakloinsteak (approved):
Looks fine to me after fixing the `table_catalog` clause. Thanks for doing this work!

db/warehouse/migrate/20250502193442_decreasing_sequence_keys.rb
dtgreiner: These were being filtered out in a previous query we had related to this, so I included them here. I don't think there is an isseu removing this clause.

db/warehouse/migrate/20250502193442_decreasing_sequence_keys.rb
dtgreiner: Thanks, looks like I missed this from my testing, I'll remove this filter.

Also the increment field was runnning into an issue when including it in the where clause (`operator does not exist: information_schema.character_data = integer`) which is why it was being filtered out in the ruby instead, but it looks like if we explicitly cast it as an integer it should work. I'll get the `instance = -1` added to the WHERE clause with the casting.

PR #5355, opened 2025-05-06
7605 search as post

ttoomey:
## _Merging this PR_

## Description
Note, only addresses the client warehouse search.

### Changes:
- Created a new `ClientSearchQuery` model to persist search parameters
- Changed client search form to use POST requests instead of GET
- Prunes saved searches that are unused
- Added fingerprinting to deduplicate identical searches
- Use uuids to obfuscate search URIs
- Implemented rate limiting for search creation and viewing
- Maintained backward compatibility by redirecting legacy GET requests
- Updated specs to test new search flow

### Impact
- PII is no longer exposed in client search URLs,  search parameters are stored in the database instead
- Rate limiting reduces possible abuse, such as scraping
- Legacy URLs still function
- search URLs are still sharable between users

### Testing
See linked issue for QA instructions

## Type of change
New feature 


config/routes.rb
```diff
@@ -434,6 +434,10 @@ def healthcare_routes
     end
   end
 
+  # Client search queries
+  resources :client_search_queries, only: [:create], controller: 'clients/search_queries'
+  get '/clients/:id/search', to: 'client_access_control/clients#search', as: 'client_search_query'
```
eanders: Are we not worried this might conflict with other client pages (eventually?) the `id` in this isn't the same as the id in `/clients/:id`.

app/models/grda_warehouse/tasks/cleanup_client_search_queries_task.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module GrdaWarehouse::Tasks
+  class CleanupClientSearchQueriesTask
```
eanders, line 14: Might be worth adding a note in [here](https://docs.google.com/document/d/12jh12ktye4XLJ1QxUYFpn23P9NoJVzMsQhkDOGwQiOQ/edit?tab=t.0#heading=h.ezuq9v3w5nts) that these are held for 2 years

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
```diff
@@ -16,29 +16,85 @@ class ClientAccessControl::ClientsController < ApplicationController
...
 
   def index
-    @show_ssn = GrdaWarehouse::Config.get(:show_partial_ssn_in_window_search_results) || can_view_full_ssn?
-    # search
-    @clients = client_scope.none
+    safe_params = GrdaWarehouse::ClientSearchQuery.permit_params(params)
+    if safe_params
+      # handle legacy get requests for search
+      query = current_user.client_search_queries.find_or_create_by_params!(safe_params)
+      redirect_to client_search_query_path(id: query.id), status: :moved_permanently
```
eanders: I wonder if we should use the fingerprint instead of the id in the url so someone can't easily guess at someone else's search terms...
Maybe with ID protection on, that will get translated anyway.
Also, I wonder about attaching the query record to a user, I like it from the reporting side, but it means it's not transferrable (you can't easily pass the URL to someone else).

drivers/client_access_control/spec/requests/client_visibility_legacy_spec.rb
```diff
@@ -29,10 +35,9 @@ def doc_has_client?(doc, warehouse_client)
 
     describe 'and the user does not have a role' do
       it 'user cannot see any clients' do
-        get clients_path(q: 'bob')
```
eanders, line 32: Is it worth keeping one of these old get versions somewhere (maybe it is below, but I haven't gotten there yet) to confirm the redirect behavior?

drivers/client_access_control/spec/requests/clients_spec.rb
```diff
@@ -160,12 +183,49 @@
...
-    it 'allows index' do
+    it 'allows index with no params' do
       sign_in user
       get clients_path
       expect(response).to render_template(:index)
     end
 
+    it 'handles legacy GET search request' do
+      sign_in user
+      get clients_path, params: { q: 'test' }
```
eanders, line 189: Legacy path is here

eanders (approved):
Thanks @ttoomey, this looks good to me.  
Dropped some questions inline, but mostly want to discuss using something other than id and the possibility of sharing a link eventually.

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
ttoomey: As-written using the id shouldn't matter as a user cannot access a search query if they don't own it; fingerprints are only unique per user. However if we need to support cross-account sharing of search links then that needs to change. 

Just checking before proceeding, do you think that's real requirement?

To make sharing work we would need to:
- replace the user id association with a join table between users and search queries, that would still allow us to still track searches. I suppose we could make it a search_events table that would give us even more granular info on search activity. 
- we'd replace PK ids with fingerprints in routing
- we'd need to ensure that the id-protection works here. If not we would need to salt the fingerprint with a secret of some kind to prevent guessing searches

config/routes.rb
ttoomey: It's a good point. I think the chance of a routing conflict is low unless there's another search action on client record. But I think we can do better. I'll put these both under 'client_searches'
```
  resources :client_searches, only: [:create], controller: 'clients/search_queries'
  get '/client_searches/:id', to: 'client_access_control/clients#search', as: 'client_search_query'
```

drivers/client_access_control/spec/requests/clients_spec.rb
ttoomey: sorry, does something need to change with that?

drivers/client_access_control/app/controllers/client_access_control/clients_controller.rb
eanders: We may want to discuss synchronously, but:
1. Being able to share URLs is really helpful for troubleshooting (and I assume occurs occasionally for internal communications)
2. We should already have access logs with the ActivityLog, so maybe we don't need a new table.  
3. I'm backpedaling on the idea of using our existing id encoder, that isn't enabled everywhere

Proposal:
1. Remove the ClientSearchQuery -> user association
2. Adjust so it uses the fingerprint as the "id" in the url
3. Allow anyone to use any ClientSearchQuery (the data will still be limited to what they can see, and if we are using the fingerprint instead of the id, they shouldn't be guessable)
4. Confirm that ActivityLog contains a reference to the `User` and `ClientSearchQuery` that we can follow if we ever want more detailed reporting

Questions:
1. Do we need to know who created a `ClientSearchQuery`, or is it sufficient to see who used it first?
2. Is there a need for further obfuscation or is the fingerprint sufficiently hard to guess.  I believe the goal would be to prevent someone snooping on someone else's query and potentially gaining information about who they were searching for.  This seems pretty unlikely and high effort-low return to me, but maybe I'm missing something.
3. Does this accomplish the original intention?  (I believe that was mostly to keep PII out of the URL, which might get logged or shared, but maybe there was additional criteria.)

ttoomey: Follow up on huddle with @eanders :
- [x] remove the direct user assoc on search query but keep a created_by_id
- [ ] try to enhance the search resolver to log additional client_ids from the search results using activity logger
- [x] try to replicate what id protector does but make it controller-scoped
- [x] we still need fingerprints to support find_or_create_by

ttoomey: Note, activity_log doesn't have a column to store extra data, such as client search result ids. So we'd need to add that col; probably would implement in a different feature

eanders: I agree, that can be future work.  Maybe it's not a column, but another table that references the activity log, one row per item and the item can include an id and type so we could use it for something other than clients.

ttoomey:
@eanders ready for re-review. 
- I've adjusted the implementation to use encrypted ids. Encryption key is generated at request time and stored in AppConfigProperty. Uses aes-128-gcm - my understanding is this adds a nonce automatically
-  searches can now be shared between users
- improvements to validation, handling bad queries, and misc cleanups

app/models/grda_warehouse/client_search_query.rb
```diff
@@ -0,0 +1,105 @@
...
+    ALLOWED_PARAMS = ['q', 'client'].freeze
+    ALLOWED_CLIENT_PARAMS = ['first_name', 'last_name', 'dob', 'ssn'].freeze
+
+    validate :validate_params
+
+    def encrypted_id
+      raise 'Attempt to encrypt NULL id' unless id.present?
+
+      ClientSearchQueryIdProtector.instance.encrypt(id.to_s)
+    end
```
eanders: This is probably a stupid question, but couldn't we just use the[ built-in attribute encryption (deterministic?)](https://guides.rubyonrails.org/active_record_encryption.html#declaration-of-encrypted-attributes).  
I believe we could just search by the params to find it.  If necessary we could manually convert to json before encrypting and searching.

ttoomey: It's a good question. I can definitely get behind replacing custom code with something built in. I haven't used rails attr encryption extensively by my understanding is that it's more for encrypting data at rest rather than url obfuscation.

The current approach requires no additional state other than a key and also produces relatively short url-friendly values. The encrypted ids are non-deterministic. Do you think we want stable ids instead?

So I guess either way could work, except with encrypted attrs we'd be adding db persistence, I'm not sure that's needed unless we want deterministic ids. For encrypted attrs, I think we'd also need to configure rails keys in our environment rather than db-based config.

> ActiveRecord::Encryption.encryptor.encrypt('test')
(irb):1:in `<main>': Missing Active Record encryption credential: active_record_encryption.primary_key (ActiveRecord::Encryption::Errors::Configuration)

I do see that rails default to AES-256-GCM, which probably is a better choice here

ttoomey:
Follow up discussion, decided to use UUIDs rather than encrypted ids. Stable IDs are desirable in this case.  This diverges from the existing id protection middleware

eanders (approved):
Thanks @ttoomey, I think we're there!

PR #5356, opened 2025-05-06
PII permission f ro Superset

eanders:
## _Merging this PR_

## Description
* Includes name, DOB, and SSN permissions for Superset access

## Type of change
* New feature

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -102,10 +102,18 @@ class User < ApplicationRecord
 
   def populate_external_reporting_permissions!
     # Projects
-    permission = :can_view_assigned_reports
-    ids = GrdaWarehouse::Hud::Project.viewable_by(self, permission: permission).pluck(:id)
-    batch = ids.uniq.map do |item_id|
-      GrdaWarehouse::ExternalReportingProjectPermission.new(user_id: id, email: email, project_id: item_id, permission: permission)
+    permissions = [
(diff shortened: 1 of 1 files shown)


app/models/user.rb
```diff
@@ -102,10 +102,18 @@ def user_project_id_prefix
...
     # Projects
-    permission = :can_view_assigned_reports
-    ids = GrdaWarehouse::Hud::Project.viewable_by(self, permission: permission).pluck(:id)
-    batch = ids.uniq.map do |item_id|
-      GrdaWarehouse::ExternalReportingProjectPermission.new(user_id: id, email: email, project_id: item_id, permission: permission)
+    permissions = [
+      :can_view_assigned_reports,
+      :can_view_full_ssn,
+      :can_view_full_dob,
+      :can_view_client_name,
```
dtgreiner, line 109: Do we want to include `:can_view_hiv_status`? I don't think superset would need `:can_view_client_photo` but that's another one thats being used in within the PII suppression policies.

eanders: We might eventually, I'm inclined to simply not expose HIV status in superset at this time, and we don't have client photos yet.  Should be easy enough to add more in the future, though we may need to background it if we bring too many more permissions over as each calculation adds some amount of time.

dtgreiner (approved):

PR #5360, opened 2025-05-06
Add backend for global CE Admin

martha:
## _Merging this PR_

frontend pr https://github.com/greenriver/hmis-frontend/pull/1151

## Description
- Opportunity:
  - Add new fields including organization name, unit, and date available
  - Add new sort options: ascending/descending based on date available. (For now this just reflects the date the opportunity was generated)
  - Add filter options including organization, available date, and workflow template
- Referral:
  - Add new fields, including resolving `current_steps` plural instead of `current_step`
  - Add papertrail to support `updated_by`. Also add papertrail to Steps
  - Add filter options including organization, template, and length of time it's been on the current step
- Factor out common code for resolving filters, etc. into concerns `HasCeReferrals` and `HasCeOpportunities`

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/has_ce_opportunities.rb
```diff
@@ -0,0 +1,52 @@
...
+      end
+
+      private
+
+      def scoped_ce_opportunities(scope, user: current_user, sort_order: :date_available_earliest_first, filters: nil)
+        raise unless Hmis::Ce.configuration.enabled? # TODO(#7506) permissions
+
+        scope = scope.viewable_by(user)
+
+        scope = scope.where(status: filters&.status) if filters.respond_to?(:status) && filters&.status.present?
```
martha: To discuss: I don't really love this messy `filter.respond_to?` check peppered in everywhere. I was wondering if we have a common pattern for dealing with this; maybe that's what the `yield_self` in `EnrollmentFilter` helps with? https://github.com/greenriver/hmis-warehouse/blob/stable/drivers/hmis/app/models/hmis/filter/enrollment_filter.rb

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -57,20 +64,24 @@ def client
...
-        order(Arel::Nodes::Case.new.when(step_t[:status].eq('in_progress')).then(1).else(2)).
-        order(step_t[:id]) # Also sort by ID, to make sure this resolves deterministically
+    def current_step_time
+      # TODO(#7647) - use more accurate timestamp field rather than updated_at
+      load_ar_association(object, :current_steps).to_a.min_by(&:updated_at)&.updated_at
+    end
 
-      step = load_ar_association(instance, :steps, scope: step_scope).first
-      return if step.nil?
+    def updated_by
```
martha, line 81: This is just a sketch of an approach we might take here (but somewhat outside of spec). Let me know what you think!

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -53,6 +59,21 @@ def project_type
...
+      load_ar_association(project, :organization).name
+    end
+
+    def unit
+      owner = load_ar_association(object, :owner)
+      owner if owner.is_a?(Hmis::Unit)
+    end
+
+    def date_available
+      # TODO(#7537) - implement "available after date". Always returns date the referral was created, for now
```
martha, line 73: My understanding, from the spec's reference to “recency”, is that this is meant to be an amalgamation of the opportunity's creation date with the (not yet implemented) date available field. So I've implemented it as `created_at` for now with todos for the future ticket.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -162,12 +162,21 @@ def self.static_options_for_type(pick_list_type, user:)
...
           end
       when 'WORKFLOW_DEFINITION_TEMPLATES'
+        # used by the dropdown in the modal when you manually create an opportunity (will be removed)
         return [] unless Hmis::Ce.configuration.enabled?
 
         # TODO(#7502) - templates are shared across data sources
         Hmis::WorkflowDefinition::Template.all.map do |template|
           { code: template.id, label: template.name }
         end
+      when 'WORKFLOW_DEFINITION_TEMPLATE_IDENTIFIERS'
```
martha: Happy to discuss/change this around; for example, we could replace the original `WORKFLOW_DEFINITION_TEMPLATES` with the new one, instead of adding `WORKFLOW_DEFINITION_TEMPLATE_IDENTIFIERS`. (Related to considerations about when to remove [this](https://github.com/greenriver/hmis-warehouse/pull/5336))

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -201,6 +201,7 @@ task ce_starter_pack_20250302: [:environment] do
   start_workflow_event = create_start_event(enrollment_creator_template, 'start referral', 'start_workflow', 'start_referral')
   accept_workflow_event = create_end_event(enrollment_creator_template, 'accept referral', 'end_workflow', 'accept_referral')
 
+  client_acceptance_task = create_task(client_accepts_form_def, enrollment_creator_template,  'Confirm Client Accepts Referral', case_managers)
```
martha, line 204: These changes to the starter pack update the Enrollment Creator template to have multiple steps in progress at the same time, just to show what that looks like. 
Screenshot 2025-05-06 at 6 27 35 PM

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: It makes sense to me that the dropdown used for filtering would be different from the dropdown used for configuration, since the filtering one includes retired templates but the configuration one should not. However we should probably indicate in the pick list name that it includes retired templates.

I would expect we end up with something like:
* `CE_WORKFLOW_TEMPLATE_IDENTIFIERS_INCLUDING_RETIRED` - unique ce workflow template identifiers, including representing workflows that are completly retired. used for filtering on existing/historical referrals..
* `CE_WORKFLOW_TEMPLATE_IDENTIFIERS` - unique ce workflow template identifiers that are currently published. used for configuring which template to use for a project/resource/group/etc (tbd https://github.com/open-path/Green-River/issues/7522)

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -162,12 +162,21 @@ def self.static_options_for_type(pick_list_type, user:)
...
+        # used by dropdowns for filters when filtering referral/opportunity by workflow template.
+        # Here we want to group templates by identifier in case there are multiple/past versions.
+        Hmis::WorkflowDefinition::Template.published.or(Hmis::WorkflowDefinition::Template.retired).
```
gigxz: These should filter on `template_type: :ce_referral`

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -162,12 +162,21 @@ def self.static_options_for_type(pick_list_type, user:)
...
+          group_by(&:identifier).map do |identifier, templates|
+          description = templates.max_by(&:updated_at).name
```
gigxz: We could alternatively just pick the published one, right? I think its a little confusing to use updated_at as a proxy for that. (Assuming we have the right validations in place to ensure exactly 1 published template per identifier)

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -19,17 +19,23 @@ class HmisSchema::CeOpportunity < Types::BaseObject
...
     field :project_name, String, null: false
     field :project_type, HmisSchema::Enums::ProjectType, null: false
+    field :organization_name, String, null: false
 
     field :eligibility_requirements, [HmisSchema::CeMatchRule], null: true
     field :priority_scheme, HmisSchema::CeMatchRule, null: true
     field :categories, [String], null: false
     field :active, Boolean, null: false, method: :active?
     field :candidates_generated_at, GraphQL::Types::ISO8601DateTime, null: true
+    field :date_available, GraphQL::Types::ISO8601Date, null: true
```
gigxz: Does this need to be nullable? All opportunities should have a created_at value

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -16,14 +16,18 @@ class HmisSchema::CeReferral < Types::BaseObject
     field :client_id, ID, null: false
     field :client, Types::HmisSchema::Client, null: true
     field :created_at, GraphQL::Types::ISO8601DateTime, null: false
-    field :current_step_name, String, null: true
-    field :target_enrollment, Types::HmisSchema::Enrollment, null: true # Don't resolve in batch
+
+    field :current_steps, [HmisSchema::CeReferralStep], null: true
+    field :current_step_time, GraphQL::Types::ISO8601DateTime, null: true
```
gigxz: I think this field name could be a little more descriptive. It's not clear to me if it's the time that the step was started, or time that it was last updated, or time it was completed, etc. 

For populating a column "Days in Current Step(s)", which it looks like this is doing in the frontend,  I think it would be simpler to just do that calculation on the backend and resolve it as a field:

```ruby
# how many days has this referral been on the current step. If there are multiple open steps, it uses the oldest one
 def days_on_open_steps
    oldest_open_step = load_ar_association(object, :current_steps).to_a.min_by(&:created_at)
    # how many days ago was this step created
    opened_days_ago = (Date.current - oldest_open_step.created_at.to_date).to_i
    opened_days_ago
end
```

What do you think?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: Thanks. I'm not sure if there is a meaningful distinction between the "updated_at" on ce_referrals vs the "updated_at"  on each step. If not, we could take the approach of updating `updated_at` on the referral table any time on of the sub tasks is touched? I also think we will want to add `last_updated_by_user_id` references (and likely created_by), because we will want to use those for reporting. That definitely feels out of scope here, though. I think you can leave it as-is or stubbed out, and add a reference to https://github.com/open-path/Green-River/issues/7678

drivers/hmis/app/graphql/types/hmis_schema/has_ce_opportunities.rb
gigxz: I think that's what the `with_filter` helper does in the filter classes (like the one you linked). It might be good to follow the pattern of having a dedicated class for filtering, since this has a lot of filter capabilities. It lets us test filters in isolation without graphql.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: Sure sounds good, but we will need a fallback since there may not be a published version. It's probably more suitable to fallback to max by version number, instead of updated timestamp.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Cool, I agree, I'll make the update to resolve this in the backend and return it in the schema as an int number of days.

I'm not sure about using `created_at`, however. Seems like that could lead to inaccuracies when the steps are not made available at the same time as they are created. For example, in the circular workflow case, a step can get re-marked as available.

`updated_at` is not a perfect fit either, but fixing this seemed out of the current ticket's scope, so I created a placeholder over [here](https://github.com/open-path/Green-River/issues/7647) to make sure we don't forget to return to this. If it's ok with you, I will go with your approach but keep using `updated_at` instead of `created_at` for now.

drivers/hmis/app/graphql/types/hmis_schema/has_ce_opportunities.rb
martha: Thanks! Hm do you have any examples in the code you can link me to where we are testing filters in isolation without graphql? I wasn't finding any instances of `apply_filters` in the spec files.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: I see, sounds like we need a new timestamp to capture the time when the step became available. Thanks for making a ticket

drivers/hmis/app/graphql/types/hmis_schema/has_ce_opportunities.rb
gigxz: Ha I guess just because we _can_ in theory  doesn't mean we are 👀

drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
```diff
@@ -0,0 +1,52 @@
...
+
+  def with_workflow_template_identifiers(scope)
+    with_filter(scope, :workflow_template) do
+      scope.joins(:opportunity).where(opp_t[:workflow_template_identifier].in(input.workflow_template))
+    end
+  end
+
+  def on_current_step_since(scope)
+    with_filter(scope, :on_current_step_since) do
+      scope.joins(:current_steps).where(wfe_step_t[:updated_at].lt(input.on_current_step_since)).distinct
```
gigxz, line 50: Add a TODO for https://github.com/open-path/Green-River/issues/7647 here?

drivers/hmis/app/models/hmis/workflow_execution/instance.rb
```diff
@@ -6,6 +6,7 @@ module Hmis::WorkflowExecution
   class Instance < GrdaWarehouseBase
     belongs_to :template, class_name: 'Hmis::WorkflowDefinition::Template'
     has_many :steps, class_name: 'Hmis::WorkflowExecution::Step', dependent: :destroy
+    has_many :open_steps, -> { open.order_by_updated_at }, class_name: 'Hmis::WorkflowExecution::Step'
```
gigxz, line 12: Does this also need a TODO for 7647 to sort by the time the step was made available? It would be great if you can add a comment explaining why the sorting (ie if theres a particular schema field that depends on it)

gigxz: Whoops nevermind I see you have the todo on the scope itself

drivers/hmis/spec/requests/hmis/ce/admin_ce_opportunities_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+        expect(response.status).to eq(200), result.inspect
+
+        opportunities = result.dig('data', 'ceOpportunities', 'nodes')
+        expect(opportunities.dig(0, 'id')).to eq(opportunity3.id.to_s)
+        expect(opportunities.dig(1, 'id')).to eq(opportunity2.id.to_s)
+        expect(opportunities.dig(2, 'id')).to eq(opportunity.id.to_s)
+      end
+    end
+
+    context 'with many opportunities' do
```
gigxz: This is more of a "with many projects" test, since it is setting up 40 projects each with 1 opportunity (Or maybe the test could be expanded to have multiple opportunities per project?)

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -554,5 +556,17 @@ def ce_referral_step(id:)
 
       Hmis::WorkflowExecution::Step.viewable_by(current_user).find(id)
     end
+
+    # Globally available CE opportunities gated by admin permission
+    ce_opportunities_field
+    def ce_opportunities(**args)
+      # TODO(#7506) - gated by admin permission
+      resolve_ce_opportunities(Hmis::Ce::Opportunity.all, **args)
```
gigxz: Oh didn't think to call this out in the spec, but this needs to be limited to Opportunities that belong to Projects in the current Data Source. Per https://github.com/open-path/Green-River/issues/6612 we never want to be resolving data from other data sources

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -554,5 +556,17 @@ def ce_referral_step(id:)
...
+    end
+
+    ce_referrals_field
+    def ce_referrals(**args)
+      resolve_ce_referrals(Hmis::Ce::Referral.all, **args)
```
gigxz: Same here needs to be limited by current data source. thank you!

drivers/hmis/app/graphql/types/hmis_schema/has_ce_opportunities.rb
martha: Ok sounds good haha! For now I'll leave this as-is with tests in the request spec files.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Hm thanks good point. Could we rely on the `viewable_by` scopes to do this restriction? They are not fully implemented yet, but I stubbed them out simply here in a way that I think achieves what we are after: https://github.com/greenriver/hmis-warehouse/pull/5360/commits/f2c91b075bcea4125c0f636d00154e4fc7b1e3f6

drivers/hmis/app/models/hmis/workflow_execution/instance.rb
martha: There's no longer a schema field that grabs just the first one - I removed that in favor of returning them all and displaying them in a truncated list. I did add a comment though to try to clarify the purpose, thanks!

martha:
@gigxz in the spirit of speedier development, I'd like to merge this now to the release branch, in order to resolve conflicts with the permissions work and get that in a good state to PR (and avoid double-dealing with conflicts, since we squash merge).

I think I've addressed all your concerns on this PR. For posterity, the outstanding question is about whether we can use `viewable_by` scope to limit queries by data source. This is also related to the permissions work, so we can continue the discussion on that PR if needed.

I plan to leave the frontend PR open for now https://github.com/greenriver/hmis-frontend/pull/1151#discussion_r2087082323 until we resolve the linked thread.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Sorry I lost track of this @martha, your implementation looks good to me.

PR #5363, opened 2025-05-07
7667-health-db-connection

ttoomey:
## _Merging this PR_

## Description
* Cleanup health db paper trail versions so that it shares a connection pool 
* extract common behavior to GrPaperTrailVersionBehavior and add HealthBaseDb to preserve the db hierarchy 

## Type of change
Bug fix


app/models/concerns/gr_paper_trail_version_behavior.rb
```diff
@@ -0,0 +1,100 @@
...
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# Common paper trail version behavior shared across app and health database records
+# Note: warehouse records use a different version table/model
+module GrPaperTrailVersionBehavior
```
ttoomey, line 11: extracted from GrPaperTrail::Version

eanders (approved):
Thanks @ttoomey .  Not sure this DB adjustment is necessary, but it feels like a good separation.
I've also pushed a fix for the incorrect input name syntax.  
This should be good to merge and deploy.

PR #5366, opened 2025-05-08
Additional job logging for HMIS jobs

gigxz:
## _Merging this PR_

## Description
Grab-bag:
* Enable PaperTrail for HMIS ExternalID, to help tracking changes to MCI and MCI Unique IDs
* Add some cloudwatch logging to PurgeSoftDeletedRecordsJob
* Add some cloudwatch logging to Hmis::MergeClientsJob to log pre-merge enrollment state. This is a quick easy change until we do https://github.com/open-path/Green-River/issues/7444
  * Also change to `update!` for changing MCI External IDs, to get paper trail versions
* Add some cloudwatch logging to `AcHmis::WarehouseChangesJob` to help investigations
* Wrap some scheduled job executions in `rescue`...`Sentry.capture_exception` since I believe this is where some errors get swallowed. I believe this is the best practice we do for other processes in this rake task. @eanders can you confirm this approach?

## Type of change
logging


gigxz:
Rack attack test failure is unrelated

app/jobs/purge_soft_deleted_records_job.rb
```diff
@@ -121,6 +125,13 @@ def process_model(model, data_source:)
...
     @total_deleted += size
-    throw :halt if @total_deleted >= @max_deleted
+    return unless @total_deleted >= @max_deleted
+
+    log("Reached maximum deletion limit of #{@max_deleted} records, halting job.")
+    throw :halt
+  end
+
+  def log(str)
+    @notifier.ping(str) # logs to cloudwatch
```
eanders: Using notifier will send to slack and the rails log.  If you don't need it in slack, I'd probably just use `Rails.logger.info(str)`

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -221,8 +226,9 @@ def merge_mci_ids
         where.not(value: current_ids_for_retained_client).
         order(:id).reverse.index_by(&:value) # de-duplicate by value, take first id
 
-      mci_ids.where(id: records_by_value.values.map(&:id)).
-        update_all(source_id: client_to_retain.id)
+      mci_ids.where(id: records_by_value.values.map(&:id)).each do |external_id|
+        external_id.update!(source_id: client_to_retain.id)
```
eanders, line 228:
```suggestion
        # save individually to trigger paper trail version creation
        external_id.update!(source_id: client_to_retain.id)
```

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -315,5 +328,9 @@ def destroy_merged_clients
         map(&:name)
       scope.preload(*preloads).each(&:destroy!)
     end
+
+    def debug_msg(str)
+      @notifier.ping(str) # logs to cloudwatch
+    end
```
eanders: Again here, I'd just send it to the log

lib/tasks/grda_warehouse.rake
```diff
@@ -325,13 +325,23 @@ namespace :grda_warehouse do
...
+    # Run HMIS Auto-Exit daily in the early morning. This is running here instead of the daily tasks because of the daily task is bloated.
     if DateTime.current.hour == 5 && HmisEnforcement.hmis_enabled? && GrdaWarehouse::DataSource.hmis.exists?
-      # Run HMIS Auto-Exit daily in the early morning. This is running here instead of the daily tasks because of the daily task is bloated.
-      Hmis::AutoExitJob.perform_now
+      begin
+        Hmis::AutoExitJob.perform_now
+      rescue StandardError => e
+        Sentry.capture_exception(e)
+        Rails.logger.error(e.message)
+      end
```
eanders, line 335: This seems reasonable to capture errors but not prevent subsequent tasks from running.  An alternative, if timing isn't critical, would be simply to use `perform_later` so it runs in its own DJ worker.

app/jobs/purge_soft_deleted_records_job.rb
gigxz: I don't see the `Rails.logger.info` logs appear in cloudwatch. I only see logs from notifier.. I can find an example to share from a job that logs both

lib/tasks/grda_warehouse.rake
gigxz: Some of these jobs that are running synchronously in this task might be pretty slow. For example:
* `driver:hmis:process_activity_logs`  - runs hmis activity log processing synchronously every hour (not sure how long this takes, could be long running during busy times with lots of users)
* `PurgeSoftDeletedRecordsJob`  - when we turn this on in prod the first time it will be huge
* `AutoExitJob` (usually fast but could be slow if new auto exit rule is enabled for a project with lots of stale enrollments)


I think we can decouple this question from the PR, but it seems to me that those should be queued as async jobs instead of being run in-line in this task. Let me know if you agree. I would probably make that another PR since its more impactful.

Edit to add: I suppose if any are time-sensitive it's probably process_activity_logs, so the audit trail in the interface doesn't get to out of date. The others could be queued imo

gigxz:
@eanders updated this to use normal logging!

lib/tasks/grda_warehouse.rake
eanders: If they aren't time-sensitive, I say we adjust to `perform_later` (another PR is fine.)

Once we get to EKS, my understanding is that cron jobs are less expensive (they can be kicked off by a single watcher instead of separate ECS scheduled tasks), so we should revisit what is run on a schedule vs async.  

That said, maybe we want to wait to adjust these if we think everything's working fine now.

eanders (approved):

ttoomey (commented):
looks good, thank you!

PR #5368, opened 2025-05-08
Implement CE permissions

martha:
## _Merging this PR_

## Description
Summary of changes:
- Migrate to add new permissions
- (Unrelated to this ticket) Migrate to remove unused column
- Implement `viewable_by` scopes on referral and opportunity
- Implement permission checks across the CE surface area of the app.

Related frontend PR (draft): https://github.com/greenriver/hmis-frontend/pull/1154

## Type of change
Feature build out

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - I'll add testing instructions on the corresponding frontend PR


drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -7,9 +7,6 @@ class Candidate < GrdaWarehouseBase
     belongs_to :candidate_pool, class_name: 'Hmis::Ce::Match::CandidatePool'
     belongs_to :client, class_name: 'Hmis::Hud::Client'
 
-    # TODO(#7395): permissions
```
martha, line 10: To discuss: Does this scope still make sense? Similar to Step (below), the Project is what governs permission to see a Candidate, unless the user has the global permission for viewing client candidacy. But there is no direct way to join to project, since candidates belong to candidate pools that may be associated with multiple projects. 

I'm proposing removing this scope in favor of checking the permissions:
- here at the client level https://github.com/greenriver/hmis-warehouse/pull/5368/files#diff-41f4ce11d41265ed85c42d8bfe3c7234e0cb66b9ecfaa898f5c75740e8f83354R424-R426
- here at the individual opportunity level https://github.com/greenriver/hmis-warehouse/pull/5368/files#diff-4191bd5a0be0f9137f0e67c2aeb2702854afd3ba99d43bb5179ba418f11d8276R36-R37

drivers/hmis/app/models/hmis/workflow_execution/step.rb
```diff
@@ -15,9 +15,6 @@ class Step < GrdaWarehouseBase
 
     has_many :assignments, class_name: 'Hmis::WorkflowExecution::StepAssignment', dependent: :destroy
 
-    # TODO(#7395): permissions
```
martha, line 20: I also think this scope may not make sense. As we've defined them, permissions on Steps are closely related to referrals. The thing that makes a step viewable _is_ the referral -- we have to join to the referral to see if any _other_ steps on this referral are assigned to this user. But we want to keep referral code out of the execution code.

I think, for now, it may be acceptable to remove this scope entirely, since everyone who has permission to view a given referral can view all of its steps.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
```diff
@@ -42,6 +43,13 @@ def form_definition # Don't resolve in batch
       workflow_task.form_definitions.published.order(version: :desc).first
     end
 
+    # Helper for the frontend to determine whether to show buttons for interacting with the step.
+    # That way the logic, which is more complicated for this permission than most, gets to live in only one place.
+    def can_current_user_perform
```
martha: To me this method looks like it would cause n+1 queries, but it doesn’t seem to -- see n+1 test in `ce_referral_spec`. Let's discuss and maybe we can pair to see what I'm missing.

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -12,6 +12,7 @@ class Referral < GrdaWarehouseBase
     include SimpleStateMachine
 
     belongs_to :opportunity, class_name: 'Hmis::Ce::Opportunity'
+    has_one :project, class_name: 'Hmis::Hud::Project', through: :opportunity
```
martha, line 15: todo - remove this, use target_project?

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -163,13 +163,70 @@ def self.permissions_with_descriptions
...
+        sub_category: 'Referrals',
+      },
+      can_assign_referral_tasks: {
+        description: 'Ability to assign contacts for any referral to the project that the user can view',
+        administrative: false,
+        access: [:editable],
+        category: 'Project Access',
+        sub_category: 'Referrals',
+      },
+      can_manage_incoming_referrals: { # TODO - Deprecate and delete
```
martha, line 250: I can't remember the steps for deprecating permissions, just that it bit me in the past 🐺. let's discuss and then I can make sure it's documented somewhere!

db/migrate/20250507192058_add_ce_permissions.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+class AddCePermissions < ActiveRecord::Migration[7.1]
+  def up
+    ::Hmis::Role.ensure_permissions_exist
+    ::Hmis::Role.reset_column_information
+  end
+
+  def down
+    remove_column :hmis_roles, :can_view_enrollment_location_map
```
gigxz: Did this sneak in by accident?

martha: Yes sorry this was a bad copy-paste, I'll remove!

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -244,6 +244,13 @@ def pick_list_group_label
     end
   end
 
+  def to_pick_list_hash
```
gigxz: We lave lots of existing methods like this called `def to_pick_list_option`, prefer updating this to consistent name

drivers/hmis/app/models/hmis/workflow_execution/step.rb
gigxz: thanks, agreed!

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -73,6 +93,16 @@ def self.apply_filters(input)
       Hmis::Filter::CeReferralFilter.new(input).filter_scope(self)
     end
 
+    # This is a helper on Referral, rather than on Step, because we want to keep Workflow Execution code encapsulated away from CE code
+    def user_can_perform_task?(user:, step:)
```
gigxz: I think this should be on the HMIS User model instead, `user.can_perform_referral_step?(step)`, to follow our existing patterns? It seems a bit arbitrary to place on the referral. We can still find the referral based on the step by doing a direct db lookup, even though there isn't a rails relation.

---

And... I'm tempted to suggest encapsulating the permission logic into a policy class. This is a pattern that @ttoomey added somewhat recently into the Warehouse, and I think it would be neat to adopt into the HMIS as well.  Maybe we can discuss next week if it makes sense to try out introducing HMIS Auth Policies for CE, or if we should hold off.


Example:

```
policy = user.policy_for(referral_task)
policy.can_perform_referral_task? # encapsulate checks against "can perform own/any" on target project
```

where policy is a class similar to https://github.com/greenriver/hmis-warehouse/blob/stable/app/models/grda_warehouse/auth_policies/data_source_policy.rb, and [policy_for](https://github.com/greenriver/hmis-warehouse/blob/fb8659592187d2234745628d60b8321851444418/app/models/user.rb#L265) is similarly implemented on the HMIS user class.

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -163,13 +163,70 @@ def self.permissions_with_descriptions
...
+        sub_category: 'Referrals',
+      },
+      can_start_referrals: {
+        description: 'Ability to initiate referrals from the client waitlist in the project',
+        administrative: false,
+        access: [:editable],
+        category: 'Project Access',
+        sub_category: 'Referrals',
+      },
+      can_view_referrals: {
```
gigxz, line 192: To discuss
- the addition of these permissions may cause confusion for existing customers, particularly those using existing referral functionality. discuss if there is an easy way to implement https://github.com/open-path/Green-River/issues/7703 aka hiding these in the UI if the feature flag is off

drivers/hmis/app/graphql/types/forms/enums/pick_list_type.rb
```diff
@@ -49,6 +49,7 @@ class Forms::Enums::PickListType < Types::BaseEnum
     value 'STAFF_ASSIGNMENT_RELATIONSHIPS', 'Staff Assignment Relationships'
     value 'USERS', 'User accounts. Deprecated in favor of AUDITABLE_USERS'
     value 'ELIGIBLE_STAFF_ASSIGNMENT_USERS', 'Current users who are eligible for staff assignment'
+    value 'ELIGIBLE_REFERRAL_STEP_ASSIGNMENT_USERS', 'Current users who can be assigned to referral steps in the specified project'
```
gigxz:
```suggestion
    value 'ELIGIBLE_STAFF_ASSIGNMENT_USERS', 'Users who are eligible for staff assignment'
    value 'ELIGIBLE_REFERRAL_STEP_ASSIGNMENT_USERS', 'Users who can be assigned to referral steps in the specified project'
```
"current users" is confusing to me, we use the term "current user" for the _currently logged in user_. If you are referring to active users we can say active users, but I think its probably sufficient to say "users" because its implied that only active users would be eligible

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -186,12 +188,19 @@ def self.static_options_for_type(pick_list_type, user:)
...
+      Hmis::User.can_edit_enrollments_for(project).
+        order(:last_name, :first_name, :id).
+        map(&:to_pick_list_hash)
+    end
+
+    def self.eligible_referral_step_assignment_user_picklist(project)
+      return [] unless Hmis::Ce.configuration.enabled?
+      return [] unless project.present? # TODO(#7409) - when project-level CE configuration exists, check it here
+
+      Hmis::User.can_perform_any_referral_tasks_for(project).
```
gigxz: I think we want to filter to active users here (on the whole scope). I don't see a use-case for assigning a referral to an inactive user

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -554,13 +554,18 @@ def ce_opportunity(id:)
...
+      access_denied! unless referral
+
+      step
     end
 
-    # Globally available CE opportunities gated by admin permission
+    # All CE opportunities the user can view, resolved on admin page
     ce_opportunities_field
     def ce_opportunities(**args)
-      # TODO(#7506) - gated by admin permission
```
gigxz, line 563: Is there a permission that gates visibility of the "admin ce" pages? (viewing all available oppts and all referrals)

martha: Thanks, yes, good question that I was also thinking about as I put up this PR. I'll start a comment thread over in the permissions grid document!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
ttoomey: active record has a query cache that may avoid hitting the db for the same query (unsure if cache hits count towards db query limits in the test)

drivers/hmis/app/graphql/types/hmis_schema/client.rb
```diff
@@ -419,7 +419,12 @@ def enabled_features
     end
 
     def eligible_ce_opportunities(**args) # Don't resolve in batch
-      resolve_ce_opportunities(Hmis::Ce::Opportunity.for_client(object), **args)
+      # Check if the user has the _global_ (not project-specific) permission to view all CE opportunities a client is eligible for.
+      access_denied! unless current_user.can_view_client_eligible_opportunities?
```
ttoomey, line 426: This is fine, it's our current practice. But it would be nice to replace global perms with either a global or data-source policy some day

drivers/hmis/app/models/hmis/ce/match/candidate.rb
ttoomey: > To discuss: Does this scope still make sense? 

This seems like it's more of a convenience scope. If there's no need to show candidates outside of individual project or client context, it may not be necessary

> But there is no direct way to join to project, since candidates belong to candidate pools that may be associated with multiple projects

I'm probably not understanding but if we did want to join to project for visibility, couldn't we do something like this? Untested:

```ruby
    scope :viewable_by, ->(user) {
      opportunity_scope = Hmis::Ce::Opportunity.viewable_by(user)
      Hmis::Ce::Match::Candidate.joins(candidate_pool: :opportunities).merge(opportunity_scope)
    }
```

drivers/hmis/app/models/hmis/ce/referral.rb
ttoomey: > I'm tempted to suggest encapsulating the permission logic into a policy class.

I think that would be great! Yeah let's discuss if it's a good time to add them

The policy can also provide friendly methods corresponding to the operations we need to authorize
```
policy_for(referral_task).start?
policy_for(referral_task).submit? 
policy_for(referral_task).view? 

```

ttoomey (commented):
This is looking good!

drivers/hmis/app/models/hmis/ce/match/candidate.rb
martha: Thanks, yeah the snippet you wrote would probably work! But, let's leave out the convenience scope for now, if that sounds OK, since we are only resolving candidates from a project or a client.

drivers/hmis/app/models/hmis/role.rb
martha: Decision from standup:
- Minimal approach that would hide these new permissions, a la https://github.com/open-path/Green-River/issues/7703#issuecomment-2891267555
- Add todos for future work around feature

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
martha: Thanks, yeah that makes sense. Do we think it's okay to rely on ActiveRecord's caching for avoiding n+1s? In this case we know that when bulk-loading steps, the project will be the same on all steps, so to me it seems acceptable.

drivers/hmis/app/models/hmis/role.rb
martha: From standup discussion: Not deprecating anything yet

drivers/hmis/app/models/hmis/ce/referral.rb
martha: Decision from standup - Separately from this PR, Theron will start a spike on adding hmis policies. In this PR I'll take your first suggestion @gigxz of moving the helper to the user instead of the referral.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
ttoomey: yeah, I think it's fine to rely on the query cache here

drivers/hmis/app/models/hmis/role.rb
martha: Thanks @gigxz  for the sketch of an approach in your linked comment. I've implemented it as you suggested and it looks like it works well enough for now.

I haven't done anything in `RolesController`, since my understanding from our standup discussion was that this is sort of a stopgap that we don't necessarily plan to continue supporting. Let me know if you don't agree!

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -99,13 +99,23 @@ def self.global_permissions
...
         perms[role[:category]] ||= {}
         perms[role[:category]][role[:sub_category]] ||= {}
         perms[role[:category]][role[:sub_category]][key] = role
       end
     end
   end
 
+  def self.show_ce_permissions?
+    # Show CE permissions if CE is enabled
+    # In test env on CI, permissions_by_group is called before the database is loaded(?)
```
martha: https://github.com/greenriver/hmis-warehouse/actions/runs/15138463065/job/42555895992

to discuss/pair

martha: Proc shouldn't evaluate when the file is loaded, it should be evaluated at runtime. Like a lambda

drivers/hmis/app/models/hmis/role.rb
martha: Discussed - sufficient for now

gigxz: Yep - updated proc change looks good!

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -42,9 +42,8 @@ class Opportunity < GrdaWarehouseBase
       end
     end
 
-    # TODO(#7395): permissions
     scope :viewable_by, ->(user) do
-      joins(:project).merge(Hmis::Hud::Project.viewable_by(user))
+      joins(:project).merge(Hmis::Hud::Project.with_access(user, :can_view_units))
```
gigxz: Why did we drop the project viewability scope here? I think we are losing the check for can_view_projects by removing this.

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -27,14 +27,34 @@ class Referral < GrdaWarehouseBase
...
 
-    # TODO(#7395): permissions
     scope :viewable_by, ->(user) do
-      joins(:target_project).merge(Hmis::Hud::Project.viewable_by(user))
+      # What makes a referral viewable by a user?
+      # - If they have can_view_referrals at the target project, OR
+      # - If they have can_view_own_referrals, AND are assigned a step in the referral.
+
+      # Start with base scope that does all necessary joins, for structural compatibility when we `or` the scopes later
+      base_scope = joins(:target_project).left_outer_joins(:steps).left_outer_joins(steps: :assignments)
```
gigxz: I think the sql is equivalent if you remove `left_outer_joins(:steps)` since its joined to get to `assignments` anyway

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -116,6 +116,16 @@ def can_view_my_dashboard?
     end
   end
 
+  def can_perform_referral_step?(step)
+    referral = Hmis::Ce::Referral.find_by(workflow_instance: step.instance)
+    project = referral.target_project
+
+    permission_from_project = can_perform_any_referral_tasks_for?(project)
```
gigxz: micro-optimization: can return early if this is true

gigxz (approved):
nice work!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_step.rb
```diff
@@ -21,6 +21,7 @@ class HmisSchema::CeReferralStep < Types::BaseObject
     field :assignees, [Application::User], null: false, description: 'User(s) currently assigned to this step'
     field :updated_by, Application::User, null: true
     field :updated_at, GraphQL::Types::ISO8601DateTime, null: true
+    field :can_current_user_perform, Boolean, null: false
```
gigxz: Didn't occur to me until reviewing the frontend, but this does break our API pattern for permissions. More in line would be to resolve an access object:
```
referralStep {
   access {
       canPerformStep
   }
}
```
I think that would be relatively simple to do:
```ruby
access_field do
      field :can_perform_step, Boolean, null: false
end

def can_perform_step # no need to call it "current user", the access fragment implies that
    true
end
```

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -163,13 +167,87 @@ def self.permissions_with_descriptions
...
         access: [:editable],
         category: 'Project Access',
         sub_category: 'Management',
       },
-      can_manage_incoming_referrals: {
+      can_view_prioritized_client_lists: {
+        description: 'Ability to view client waitlists in the project',
+        administrative: false,
+        access: [:viewable],
+        category: 'Project Access',
```
gigxz: @martha Let's move everything in the current `Referrals` sub-category into its own Category called Referrals. It feels like a stretch to call these all "Project access", especially since referrals cross-cut project access in a specific way (eg you can see a referral youre assigned to in the project, even if you cant see the project). Also, Project Access already contains a lot of perms.

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -163,13 +167,87 @@ def self.permissions_with_descriptions
...
+        sub_category: 'Referrals',
+        proc: -> { Hmis::Ce.configuration.enabled? },
+      },
+      can_view_client_eligible_opportunities: {
+        description: 'Ability to view the Client page showing all Opportunities a client is eligible for. This is a global permission (not per-project).',
+        administrative: true,
+        global: true,
+        access: [:viewable],
+        category: 'Administration',
```
gigxz: I think this should be in the referrals category. The "administrative: true" tag marks it as administrative in the ui. Same with can_administrate_coordinated_entry

gigxz:
Sorry @martha just some straggling comments here as I work through the frontend! They are non-blocking to merge, still consider this approved!

drivers/hmis/app/models/hmis/role.rb
martha: Thanks yeah that makes sense!

But raises a question for me about the `viewable_by` scope over in Hmis::Ce::Referral and Hmis::Ce::Opportunity. Based on your recent comment [here](https://github.com/greenriver/hmis-warehouse/pull/5368#discussion_r2098649101) I had added this back in both places, but now I'm realizing that is not quite correct. Does this sound right to you?

- Opportunity `viewable_by` also checks `can_view_project`. You can't view opportunities if you can't view the project
- Referral `viewable_by`:
  - If you have `can_view_referrals`, then it also checks `can_view_project`
  - If you have `can_view_own_referrals`, then it does _not_ check `can_view_project` <-- in this case you CAN view referrals in projects you can't view. But I'm not sure this is fully supported right now, since referrals live "under" projects in terms of frontend routing. I'll poke around more and create a followup ticket if necessary.

martha: I created https://github.com/open-path/Green-River/issues/7729 so we can continue the discussion there

martha:
I'll leave this unmerged for now, I plan to merge it at the same time as the frontend PR

drivers/hmis/app/models/hmis/role.rb
gigxz: Thanks. Yep I agree we need to allow that person to interact with their assigned referral even if they can't view the rest of the project. Let's run by design whether they should be accessing a "floating" referral page

PR #5376, opened 2025-05-09
Enable purging deleted records in production

gigxz:
## _Merging this PR_

## Description
Enable daily process to purge records that were deleted  >1 year ago in production.
https://github.com/open-path/Green-River/issues/6718

Also, adjust job to run asynchronously in a DJ worker.

## Type of change
new feature


lib/tasks/grda_warehouse.rake
```diff
@@ -342,7 +342,7 @@ namespace :grda_warehouse do
 
     # Purge old soft-deleted records. Enable on production when we have confidence job is correct
     begin
-      PurgeSoftDeletedRecordsJob.perform_now(dry_run: false) if DateTime.current.hour == 5 && !Rails.env.production?
+      PurgeSoftDeletedRecordsJob.perform_later(dry_run: false) if DateTime.current.hour == 5
```
gigxz: Should I set a priority here or is it OK to enqueue as the default priority?

gigxz:
Keeping as draft in case we want to put it in release-164

lib/tasks/grda_warehouse.rake
eanders: This should probably get a low priority 12 or higher.

eanders (approved):

lib/tasks/grda_warehouse.rake
gigxz: updated to 15

lib/tasks/grda_warehouse.rake
```diff
@@ -340,9 +340,9 @@ namespace :grda_warehouse do
       Hmis::MatchCandidatesJob.perform_later
     end
 
-    # Purge old soft-deleted records. Enable on production when we have confidence job is correct
+    # Purge old soft-deleted records
     begin
-      PurgeSoftDeletedRecordsJob.perform_now(dry_run: false) if DateTime.current.hour == 5 && !Rails.env.production?
+      PurgeSoftDeletedRecordsJob.set(priority: 15).perform_later(dry_run: false) if DateTime.current.hour == 5
```
ttoomey: Maybe it would be worth trying this on just one prod installation to start? Just in case something goes wrong, it would at least limit the blast radius. (via AppConfigProperty or some other flag)

```suggestion
      enabled = AppConfigProperty.where(key: 'purge_soft_deleted_records', value: '1').any? || Rails.env.staging?
      PurgeSoftDeletedRecordsJob.set(priority: 15).perform_later(dry_run: false) if enabled && DateTime.current.hour == 5
```

ttoomey (approved):
Thanks!

lib/tasks/grda_warehouse.rake
gigxz: Good idea, added a check for AppConfigProperty. We will need to add the configs gradually to each environment but I think thats a good idea to start small.

PR #5379, opened 2025-05-09
Default roles

eanders:
## _Merging this PR_

## Description
* Provides default roles
* Updates SeedMaker to create default roles 
* Adds a method to SeedMaker for setting up initial production users
* Adds the collection type to pages where we have lists with collections

## Type of change
* New feature


db/seed_maker.rb
```diff
@@ -8,55 +8,80 @@
...
+  # as part of the `production_seed_first_user` setup process
+  # To reset roles to the default values, call `seed_roles(reset_permissions: true)`
+  def seed_roles(reset_permissions: false)
+    YAML.load_file(Rails.root.join('db/seeds/roles.yaml')).each do |default_role|
+      role = Role.where(name: default_role['name']).first_or_initialize
+      # If the role already exists, skip it, we may have adjusted the permissions in the UI
+      next if role.persisted? && !reset_permissions
+
+      # Don't add any roles if you already have more than 3 roles (we can manually override with reset_permissions)
+      next if Role.count > 3 && !reset_permissions
```
dtgreiner, line 68: I don't fully follow why 3 is the magic number here

dtgreiner (approved):
Just one question, but overall looks good to me

db/seed_maker.rb
eanders: That was just a guess.  If there is an existing installation with just a few roles, it felt like it would be helpful to add the defaults, but once you have more than 3 (which s probably all of them) then you probably spent some time setting them up, so maybe we shouldn't mess with it.

PR #5387, opened 2025-05-13
Implement HMIS project groups

gigxz:
## _Merging this PR_

## Description
https://github.com/open-path/Green-River/issues/7353

* Add CRUD interfaces to the warehouse for managing HMIS Project Groups. This feature is available to HMIS administrators only (aka people who can manage HMIS permissions)
* Maintain Project Groups in the hourly schedule
* Support including HMIS Project Groups in HMIS ACL Collections (called `Hmis::AccessGroup`s in the code base still)
* Remove some unused code and templates seemingly related to HMIS ACLs

This implementation only has minimal filtering capabilities needed for this issue, I left commented-out the other attributes we can implement as needed (funder, participation status, etc)

## Type of change
New feature

## How to test
* Under HMIS Admin select Project Groups
* Test creating, editing, and deleting project groups with various criteria
* Add an HMIS Collection for a project group, give it permissions
  * I tested the use-case of a project group containing an entire data source _excluding_ one project
  * Ensure permissions are applied as expected in HMIS


gigxz:
* todos https://github.com/greenriver/hmis-warehouse/pull/5387/commits/670dba31d94496798c582d2a5dee77cd7488307f
* add data_source_id
* support deleting project groups
* QA instructions and local QA (test semi-open client access use-case)

drivers/hmis/app/models/hmis/project_group.rb
```diff
@@ -0,0 +1,110 @@
...
+        )
+    end
+
+    def any_inclusion_criteria?
+      JSON.parse(inclusion_criteria).compact_blank.any?
+    end
+
+    def any_exclusion_criteria?
+      JSON.parse(exclusion_criteria || '{}').compact_blank.any?
+    end
```
eanders: Should `any_inclusion_criteria?` also:
```suggestion
    def any_inclusion_criteria?
      JSON.parse(inclusion_criteria || '{}').compact_blank.any?
    end

    def any_exclusion_criteria?
      JSON.parse(exclusion_criteria || '{}').compact_blank.any?
    end
```

Or is that safe since we set it on create?  Maybe we should set `exclusion_criteria` to an empty hash on create as well, or make that the default for both columns?

drivers/hmis/app/models/hmis/project_group_criteria.rb
```diff
@@ -0,0 +1,169 @@
...
+    include ActionView::Context
+    include ::Hmis::Concerns::HmisArelHelper
+
+    # Source data source for this project filter. Filter cannot be used across multiple data sources.
+    attribute :source_data_source_id, Integer, default: nil
+
+    # Attributes for filtering
+    attribute :project_ids, Array, default: []
+    attribute :organization_ids, Array, default: []
+    attribute :data_source_ids, Array, default: []
```
eanders: Should this be:
```suggestion
    attribute :data_source_id, Integer, default: nil
```

The filter is tied to a single data source, so I'm not sure what it means to allow multiple here.  (Also the form appears to allow multiple selection of data sources.)

drivers/hmis/app/views/hmis_admin/project_groups/_form.haml
```diff
@@ -0,0 +1,23 @@
...
+    = ic.input :organization_ids, collection: organization_options, selected: @project_group.parsed_inclusion_criteria.organization_ids, as: :grouped_select_two, group_method: :last, input_html: {multiple: true, placeholder: 'Choose Organizations' }, label: 'Organizations', required: false
+    = ic.input :project_type_numbers, collection: Hmis::ProjectGroupCriteria.available_project_type_numbers, selected: @project_group.parsed_inclusion_criteria.project_type_numbers, input_html: { multiple: true, placeholder: 'Choose Project Types' }, label: 'Project Types', as: :select_two, required: false
+    = ic.input :data_source_ids, collection: GrdaWarehouse::DataSource.where(id: ds.id).options_for_select(user: current_user), selected: @project_group.parsed_inclusion_criteria.data_source_ids, as: :select_two, input_html: {multiple: true, placeholder: 'Choose Data Sources' }, label: 'Data Sources', required: false, hint: 'If selected, all projects in the Data Source will be included in the group.'
```
eanders: Pretty sure this should only let you choose one data source, and it should be limited to the data source of this project group... which I now see is what you have, though maybe just for consistency it is a multi-select with a plural ids name.

I'm ok with that choice.

drivers/hmis/app/views/hmis_admin/project_groups/index.haml
```diff
@@ -0,0 +1,33 @@
...
+- @prompt = "Search project groups..."
+.row.mb-4
+  .col-sm-8
+    = render 'search_form'
+  .col-sm-4.text-right
+    = link_to new_hmis_admin_project_group_path, class: 'btn btn-primary mb-2' do
+      %span.icon-plus
+      Add Project Group
+- if @pagy.count.positive?
+  = render 'common/pagination_top', item_name: 'project group'
```
eanders: This is fine as is, but it might be worth seeing if `render_paginated_list` could be used easily here.

drivers/hmis/spec/models/hmis/project_group_spec.rb
```diff
@@ -0,0 +1,51 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Hmis::ProjectGroup, type: :model do
```
eanders, line 11: I'm probably missing something, but how does this file check the project group?

drivers/hmis/app/views/hmis_admin/project_groups/_form.haml
gigxz: Yeah, good point. Yes, this is a bit messy since I added the "source" data source id concept after.

I was toying with the idea that the Project Group Criteria class might be usable/useful for other future purposes that _are_ cross-data-source, so part of me wants to leave that door open in the JSON criteria shape.

drivers/hmis/spec/models/hmis/project_group_spec.rb
gigxz: 🤦  didn't write this test yet... thanks. (its copy-pasted from another file). I forgot about this so its not quite ready for review. Everything else should be though!

drivers/hmis/app/models/hmis/project_group.rb
gigxz: I made the `inclusion_criteria` column non-nullable in the database, so we shouldn't need that. But in practice now with the view, they both always get set to something (whether that's `"{}"` or an object with contents).

drivers/hmis/app/views/hmis_admin/project_groups/index.haml
gigxz: Thanks updated!

drivers/hmis/spec/models/hmis/project_group_spec.rb
gigxz: Tests added!

drivers/hmis/app/views/hmis_admin/project_groups/_form.haml
gigxz: I'm working on a change to just make this an all projects flag

drivers/hmis/app/models/hmis/project_group_criteria.rb
```diff
@@ -0,0 +1,168 @@
...
+    include ActionView::Context
+    include ::Hmis::Concerns::HmisArelHelper
+
+    # Source data source for this project filter. ProjectGroupCriteria filter cannot be used across multiple data sources.
+    attribute :data_source_id, Integer, default: nil
+
+    # Attributes for filtering
+    attribute :project_ids, Array, default: []
+    attribute :organization_ids, Array, default: []
+    attribute :all_projects_in_data_source, Boolean, default: false
```
eanders, line 22: Maybe a minor re-factor for a future date, but it seems like choosing `all_projects_in_data_source` should be mutually exclusive with the other options.

drivers/hmis/app/models/hmis/project_group_criteria.rb
```diff
@@ -0,0 +1,168 @@
...
+    end
+
+    # Returns a unique list of project IDs that match the criteria defined by this class.
+    def effective_project_ids
+      ids = []
+      ids << project_scope.where(id: project_ids).pluck(:id) if project_ids.any?
+      ids << organization_scope.joins(:projects).where(id: organization_ids).pluck(p_t[:id]) if organization_ids.any?
+      ids << project_scope.where(project_type: project_type_numbers).pluck(:id) if project_type_numbers.any?
+      ids << data_source.projects.pluck(p_t[:id]) if all_projects_in_data_source
+      ids.flatten.uniq
```
eanders: I think we could just:
```suggestion
      return data_source.projects.pluck(p_t[:id]) if all_projects_in_data_source
      
      ids = []
      ids << project_scope.where(id: project_ids).pluck(:id) if project_ids.any?
      ids << organization_scope.joins(:projects).where(id: organization_ids).pluck(p_t[:id]) if organization_ids.any?
      ids << project_scope.where(project_type: project_type_numbers).pluck(:id) if project_type_numbers.any?
      ids.flatten.uniq
```
That's a micro performance optimization since you'll just have duplicates if they are all included.

drivers/hmis/app/views/hmis_admin/project_groups/_form.haml
```diff
@@ -0,0 +1,23 @@
...
+  %h3 Project Group Name
+  = f.input :name, required: true
+
+%div.mb-4
+  %h3 Inclusion Criteria
+  = f.simple_fields_for :inclusion_criteria do |ic|
+    = ic.input :project_ids, collection: [], as: :select_two, input_html: { multiple: true, data: { 'collection-path' => api_projects_path(selected_project_ids: @project_group.parsed_inclusion_criteria.project_ids, data_source_ids: [ds.id]) }, placeholder: 'Choose Projects'}, label: 'Projects', required: false
+    = ic.input :organization_ids, collection: organization_options, selected: @project_group.parsed_inclusion_criteria.organization_ids, as: :grouped_select_two, group_method: :last, input_html: {multiple: true, placeholder: 'Choose Organizations' }, label: 'Organizations', required: false
+    = ic.input :project_type_numbers, collection: Hmis::ProjectGroupCriteria.available_project_type_numbers, selected: @project_group.parsed_inclusion_criteria.project_type_numbers, input_html: { multiple: true, placeholder: 'Choose Project Types' }, label: 'Project Types', as: :select_two, required: false
+    = ic.input :all_projects_in_data_source, as: :boolean, label: 'Include all projects in data source', required: false
```
eanders, line 17: Maybe put this first and toggle access to the others if checked?  (Again, happy for that to come later.)

drivers/hmis/spec/models/hmis/project_group_criteria_spec.rb
```diff
@@ -0,0 +1,99 @@
...
+    context 'when organization_ids are specified' do
+      let(:criteria) do
+        Hmis::ProjectGroupCriteria.new(
+          { organization_ids: [o1.id] },
+          data_source_id: hmis_ds.id,
+        )
+      end
+
+      it 'includes all projects belonging to the specified organizations' do
+        expect(criteria.effective_project_ids).to contain_exactly(*o1.projects.pluck(:id))
```
eanders, line 51: I wonder if it's worth adding 
```suggestion
        expect(criteria.effective_project_ids.count).to eq(1)
        expect(criteria.effective_project_ids).to contain_exactly(*o1.projects.pluck(:id))
```
Otherwise, I think we might think we're getting something, but end up with empty project groups.  Same goes for the next set of tests.

eanders (approved):
Thanks @gigxz !

drivers/hmis/app/views/hmis_admin/project_groups/_form.haml
gigxz: Yeah, for sure. I have a second branch in progress already for supporting additional criteria. I can try to incorporate this. That would need custom Javascript in this template, right? (muddling my way around rails views)

eanders: Yeah, some basic custom JS in the template is probably how I'd roll it.  Alternatively, we could add the JS in `app/javascript` and then include it via the controller, but with something this small, I'd probably just add some wrapper divs with useful classes and put the toggle inline.

PR #5390, opened 2025-05-14
7579 spm missing enrollments

ttoomey:
## _Merging this PR_

## Description
Don't crash when source enrollment is missing

## Type of change
Bug fix

### DIFF
--- a/drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
@@ -200,6 +200,7 @@ module HudSpmReport::Fy2024
       end
       enrollments.each do |enrollment|
         if enrollment.project_type.in?(HudUtility2024.project_type_number_from_code(:es_nbn))
+           # FIXME, see #7473
           next unless enrollment.enrollment.present? # Skip if the enrollment has disappeared (e.g., a concurrent import deleted it)
 
           # https://files.hudexchange.info/resources/documents/System-Performance-Measures-HMIS-Programming-Specifications-September-2023.pdf - p11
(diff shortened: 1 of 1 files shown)


ttoomey:
@eanders - does this strategy seem reasonable? I wish we had better synchronization with whatever was deleting the enrollments

drivers/hud_spm_report/app/models/hud_spm_report/fy2024/episode.rb
```diff
@@ -286,6 +287,9 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
       bed_nights = existing_bed_nights.index_by(&:last)
       enrollments.each do |enrollment|
         if enrollment.project_type.in?(HudUtility2024.project_type_number_from_code(:es_nbn))
+          # FIXME, see #7473
```
eanders, line 290: @ttoomey I think this is reasonable, though it feels like this shouldn't be happening.

If we called `references` and `preload` on the scope, would that avoid this situation by making sure the data is fetched when the query starts?

ttoomey: I don't think that preload or references would be a robust solution. My thought is that we need synchronization between reporting and daily processing tasks, probably via an advisory lock

eanders: Yeah, that wasn't a proposal as a full-featured solution, but as a way to decrease the possible data collisions.  

Advisory locks might work, but the import process can take hours, as can some report jobs, so we'd need to find the right granularity that would allow both to run without affecting the other too much.

eanders (approved):

PR #5397, opened 2025-05-16
Add CE step "available at" field

martha:
## _Merging this PR_

## Description
- Add migration for adding the `available_at` timestamp field to workflow engine steps. It's required (non-null), so the migration makes it default to the record's creation timestamp - which I think is acceptable since there are no production users yet
- Populate `available_at` as well as `created_at` and `started_at` (existing cols, but previously not populated) in the workflow engine
- Update to use the new `available_at` timestamp instead of `updated_at` everywhere that we want to get at the concept of "stale" tasks

## Type of change
Feature followup


db/warehouse/migrate/20250516141401_add_step_available_at.rb
```diff
@@ -0,0 +1,27 @@
...
+
+    safety_assured do
+      execute <<~SQL
+        UPDATE wfe_steps SET available_at = created_at
+      SQL
+
+      change_column_null :wfe_steps, :available_at, false
+
+      # Index because we have to sort steps on this field
+      add_index :wfe_steps, :available_at
```
martha: I'd be glad for feedback on this index. It seemed worthwhile to me, given that we know we will be sorting on the field. But let me know if there are some additional considerations that I should take into account. Maybe our general practice is to not over-optimize on db indices until we actually see a need for them?

gigxz: When you're referring to db sorting on this column, is it just happening when we resolve `current_steps` on Referral? I'm not seeing anywhere else but maybe I'm missing something. That should only return a handful of steps max so maybe we should just sort them in memory. I can't imagine we ever have a need to sort a large number of steps by availability? (Since we only really care about that per-referral).

gigxz: By doing it in memory, I mean `load_ar_association(object, :current_steps).sort_by(&:available_at)`

martha: Yes that's what I am referring to. Sounds good, I can remove the index and sort in memory

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -73,16 +73,16 @@ def client
     end
 
     def current_steps
-      load_ar_association(object, :current_steps)
+      load_ar_association(object, :current_steps).sort_by(&:available_at)
```
gigxz: If parallel steps became available at the same time, this might not be deterministic - recommend secondary sort by ID

gigxz (approved):
Looks good, thanks!!

PR #5400, opened 2025-05-17
7714 null template error

ttoomey:
## _Merging this PR_

## Description

Fix nil source handling in HMIS CSV importer error/validation views

- Add safe navigation and fallbacks for nil source handling in Excel exports
- Add test coverage for error/validation controllers
- Fix errors in loader concern by filtering to valid attributes

## Testing
login as admin -> data sources -> data source -> click on uploads -> click on timestamp of upload -> 'Download Error Report'

## Type of change
Bug fix,  tests

### DIFF
--- a/drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
+++ b/drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
@@ -16,8 +16,8 @@ wb = xlsx_package.workbook
     sheet.add_row((['Message'] + inferred_cols), style: title)
 
     validations.each do |error|
-      source = error.source
-      row = [error.status] + (source&.hmis_data&.values || [])
+      source_data = error.source&.hmis_data || {}
+      row = [error.status] + source_data.values_at(*inferred_cols)
       sheet.add_row(row)
     end
   end
@@ -36,8 +36,8 @@ end
     sheet.add_row((['Message'] + inferred_cols), style: title)
 
     errors.each do |error|
-      source = error.source
-      row = [error.details] + (source&.values_at(inferred_cols) || [])
+      source_data = error.source&.hmis_data || {}
+      row = [error.details] + source_data.values_at(*inferred_cols)
       sheet.add_row(row)
     end
   end
--- a/drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
+++ b/drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
@@ -1,32 +1,43 @@
+# frozen_string_literal: true
+
+# Generate Excel workbook with validation and error reports
 wb = xlsx_package.workbook
+
+# process validations
 @validations.each do |file, validations|
   filename = file.split('::').last
   sheet_title = "#{filename} Validation Flags"
 
   wb.add_worksheet(name: sheet_title[0..30]) do |sheet|
     title = sheet.styles.add_style(sz: 12, b: true, alignment: { horizontal: :center })
-    row = ['Message'] + validations.first.source.class.hmis_structure.keys
-    sheet.add_row(row, style: title)
+
+    # attempt to infer the column names
+    inferred_cols = validations.detect(&:source)&.source&.class&.hmis_structure&.keys || []
+    sheet.add_row((['Message'] + inferred_cols), style: title)
+
     validations.each do |error|
       source = error.source
-      row = [error.status]
-      row += source.hmis_data.values if source
+      row = [error.status] + (source&.hmis_data&.values || [])
       sheet.add_row(row)
     end
   end
 end
+
+# process errors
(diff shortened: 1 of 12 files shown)


ttoomey:
looks like there's another issue in sentry that can get pulled into this, moving to draft

drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
```diff
@@ -1,32 +1,43 @@
...
-    row = ['Message'] + hmis_keys
-    sheet.add_row(row, style: title)
+
+    # attempt to infer the column names
+    inferred_cols = errors.detect(&:source)&.source&.class&.hmis_structure&.keys || []
+    sheet.add_row((['Message'] + inferred_cols), style: title)
+
     errors.each do |error|
-      source = error.source
-      row = [error.details] + source.values_at(hmis_keys)
```
ttoomey, line 29: this old statement seems incorrect, seems like it should have been `source.hmis_data.values_at(*hmis_keys)`

drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
```diff
@@ -1,32 +1,43 @@
...
   filename = file.split('::').last
   sheet_title = "#{filename} Validation Flags"
 
   wb.add_worksheet(name: sheet_title[0..30]) do |sheet|
     title = sheet.styles.add_style(sz: 12, b: true, alignment: { horizontal: :center })
-    row = ['Message'] + validations.first.source.class.hmis_structure.keys
-    sheet.add_row(row, style: title)
+
+    # attempt to infer the column names
+    inferred_cols = validations.detect(&:source)&.source&.class&.hmis_structure&.keys || []
```
ttoomey, line 15: try to find the first non-null source

drivers/hmis_csv_twenty_twenty_two/app/models/hmis_csv_twenty_twenty_two/loader/loader_concern.rb
```diff
@@ -12,7 +14,7 @@ module LoaderConcern
       belongs_to :loader_log, optional: true
 
       def hmis_data
-        @hmis_data ||= slice(*self.class.hmis_structure.keys)
+        @hmis_data ||= slice(*self.class.hmis_structure.keys & attributes.keys)
```
ttoomey, line 17: seeing some crashes here due to values in `hmis_structure.keys` that were not in the model's attrs

eanders:
@ttoomey does it make sense to you that this is happening because someone is trying to download the errors for an FY2022 file?
I'm able to generate the error [here](https://qa-warehouse.openpath.host/imports/3520) (on an upload from Sep 2, 2023).

But the current code seems [totally happy here](https://qa-warehouse.openpath.host/hmis_csv_importer/importer_validations/==0cJs7qjdb2qlbG8+4zkeDg==/CEParticipation) (an upload from Mar 19, 2025.)

I'll drop a couple other comments inline, but maybe I'm just looking in the wrong place.

drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
eanders: I think I agree, maybe `values_at` has some magic I'm not seeing (like some internal splat?)  Weird that it seems to be working fine sometimes.

eanders:
Any chance this is being caused by our retention policy for the loaded data?  Maybe these errors are all on slightly older data sets?

ttoomey:
> does it make sense to you that this is happening because someone is trying to download the errors for an FY2022 file

I don't think so, the original sentry report was for an import from the current year

> Any chance this is being caused by our retention policy for the loaded data? Maybe these errors are all on slightly older data sets?

Yes, I assumed it is due to our cleanup routines removing old records (HmisCsvImporter::Cleanup::ExpireLoadersJob). Some of the original code also seemed to anticipate the source might be nil (`row += source.hmis_data.values if source`). However it crashes if it happened to be the first row where it infers the headers.

drivers/hmis_csv_importer/app/views/hmis_csv_importer/importer_errors/download.xlsx.axlsx
ttoomey: oh yeah, it does seem like active record magic, that explains it. Probably not behavior we want to rely on though
```
# works on an AR instance somehow
irb(main):011> User.first.values_at('id')
=> [1]
irb(main):010> User.first.values_at(['id'])
=> [1]

# but doesn't work on a hash (as expected)
irb(main):008> User.first.attributes.values_at('id')
=> [1]
irb(main):009> User.first.attributes.values_at(['id'])
=> [nil]
```

eanders (approved):

ttoomey:
retargeting to 164 since it doesn't need to be a hotfix

PR #5409, opened 2025-05-20
Add workflow template validator

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add a `WorkflowTemplateValidator` -- see inline comments about why it is _not_ an ActiveRecord validator.
- Add helper methods to the `WorkflowTemplate` for running these validations outside of lifecycle hooks
- Validate the templates created in the starter pack
- Add spec

## Type of change
Feature build-out


drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -0,0 +1,56 @@
...
+    validate_start(record)
+    validate_end(record)
+    validate_tasks_and_gateways(record)
+    validate_nodes_reachable(record)
+  end
+
+  private
+
+  def validate_start(record)
+    unless record.nodes.entrypoints.count == 1
```
ttoomey, line 25: hypothetically a workflow could have multiple entry points but this seems like a good assumption for now

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -0,0 +1,56 @@
...
+    record.errors.add(:base, 'There must be at least one end event.') unless endpoints.any?
+
+    without_inflows = endpoints.filter { |end_event| end_event.inflows.none? }
+    record.errors.add(:base, "The following end events have no inflows: #{without_inflows.map(&:name).join(', ')}") if without_inflows.any?
+
+    with_outflows = endpoints.filter { |end_event| end_event.outflows.any? }
+    record.errors.add(:base, "The following end events have outflows: #{with_outflows.map(&:name).join(', ')}") if with_outflows.any?
+  end
+
+  def validate_tasks_and_gateways(record)
```
ttoomey, line 47: we might also want to validate that all gateways have at least one unconditional outflow that serves as the default. Otherwise if a step is completed and none of the conditions on the other outflows occur, the workflow might stall

ttoomey (approved):
looks great!

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
martha: I need to think this through a little more -- let me try to write out my thoughts.

Right now, in spec tests and starter pack, we have several WFTs that violate this rule. They usually look something like:
- Step 1 where the user submits a form with a multiple choice question, Q. ->
- Exclusive Gateway ->
  - if Q == A, go to Step 2
  - if Q == B, go to Step 3, etc.

Q is required by form validation checked on _submit,_ so if the WFT is set up correctly, the workflow won't stall. But for an incorrectly setup template to get caught at the WFT _validation_ stage, the WFT would need to "introspect" its own form definitions to find its required values. @gigxz and I had discussed that, and decided to keep it out of scope for now. We could revisit once we build out WFT definition tools in the frontend.

---
It sounds like the check you are describing is much simpler, something like,
```ruby
# if there is any gateway where *all* of its outflows are conditional,
record.nodes.gateways.filter { |gateway| gateway.outflows.map(&:condition).all? }.each do |gateway|
  # that's an error, at least one should be nonconditional
  record.errors.add(:base, "Gateway '#{gateway.name}' must have at least one non-conditional (default) outflow.")
end
```
But here's where I'm stuck. For _inclusive_ gateways, all outflows that match the condition are followed. So the default outflow would be followed, even if it was a dummy default meant to just make the validator happy. Great, then sounds like we should use an _exclusive_ gateway in that case. But what if the first outflow detected is the one with no condition (the default), even if there was a 'better match' that met the condition farther on down the list?

I think that is surmountable with some updates to our [engine code](https://github.com/greenriver/hmis-warehouse/blob/edcffdb8aa81593d3017a4b9b2b8bebefc4bde83/drivers/hmis/app/models/hmis/workflow_execution/engine.rb#L113-L114), but that starts to sound out of scope for this ticket.

ttoomey: Totally fine if this is out of scope.

I was just going by what I remember from BPMN recommendations. For exclusive gateways I think we'd always want to evaluate conditional outflows first and in a deterministic order (looks like we have this already, evaluating outflows in priority order). We might just need something like this:
```
-       outflows = node.outflows.sort_by(&:position)
+       outflows = node.outflows.sort_by { |f| [f.condition ? 0 : 1, f.position]} # evaluate outflows with conditions first

```

martha: Hm, sounds good, I've taken a shot at that here: https://github.com/greenriver/hmis-warehouse/pull/5409/commits/83017bce071811501c2b5dc9ae2003fb920c3bf3

In the api layer, we implemented some logic that stops traversing the graph when it hits a conditional task:
https://github.com/greenriver/hmis-warehouse/blob/edcffdb8aa81593d3017a4b9b2b8bebefc4bde83/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb#L42-L43

The motivating example was an "admin approve denial" template, where we only want the admin task to appear in the list if the first task has already been denied.

We might need to rethink that now. I worked around that for now in the [starter pack](https://github.com/greenriver/hmis-warehouse/pull/5409/commits/83017bce071811501c2b5dc9ae2003fb920c3bf3#diff-127bc39c8536e90cb208bc402369c7e4f7d1f13db5a66838d360fb28e524cde6R199-R200) by making the client acceptance task the default outflow, although that feels a bit strange. @gigxz curious for your thoughts on this.

gigxz:
Can we merge this or are you waiting for additional review @martha?

martha:
Thanks @gigxz. I think it is okay to merge, but I was holding off to see if you had thoughts on this thread https://github.com/greenriver/hmis-warehouse/pull/5409#discussion_r2111975799

gigxz:
lets merge thank you! that sounded reasonable to me

PR #5413, opened 2025-05-21
PIT CH Household calculation

dtgreiner:
## _Merging this PR_

## Description

PIT CH HH calculation was returning different values for different HH members when the child was CH and rest of the HH was not. This update has the HH evaluated based on the household makeup and not a per-member basis.

## Type of change
Bug fix

### DIFF
--- a/app/models/concerns/hud_reports/households.rb
+++ b/app/models/concerns/hud_reports/households.rb
@@ -83,7 +83,7 @@ module HudReports::Households
     # NOTE: Client CH status is only inherited if the client was present at the start of the enrollment.
     # per HUD guidance, the HoH should always be present for the entire stay, so we'll compare start dates to them
     # see AirTable Issue ID 30
-    private def household_chronic_status(hh_id, client_id)
+    private def calculate_household_chronic_status(hh_id, client_id, chronic_status: :chronic_status)
       household_members = households[hh_id]
       # we couldn't find a household
       return false unless household_members.present?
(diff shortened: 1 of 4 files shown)


app/models/concerns/hud_reports/households.rb
```diff
@@ -83,25 +83,40 @@ def hoh_age(household_id, date)
...
-    private def household_chronic_status(hh_id, client_id)
+    private def calculate_household_chronic_status(hh_id, client_id, chronic_status: :chronic_status)
       household_members = households[hh_id]
       # we couldn't find a household
       return false unless household_members.present?
 
       hoh = household_members.detect { |hm| hm[:relationship_to_hoh] == 1 }
       current_member = household_members.detect { |hm| hm[:client_id] == client_id }
 
+      # When no specific client_id is provided (PIT), use the HoH as the current_memberfor the household calculation
```
eanders:
```suggestion
      # When no specific client_id is provided (PIT), use the HoH as the current_member for the household calculation
```

app/models/concerns/hud_reports/households.rb
```diff
@@ -83,25 +83,40 @@ def hoh_age(household_id, date)
...
+      # This will allow the entire HH to have the same chronic status
```
eanders, line 95: Shouldn't we look at all household members?  Chronic status should be inherited from any adult or HoH.

app/models/concerns/hud_reports/households.rb
```diff
@@ -83,25 +83,40 @@ def hoh_age(household_id, date)
...
+      current_member ||= hoh
+
+      current_member_entry_date = current_member[:entry_date] if current_member.present?
+      # The current_member will not exist in cases where client_id is not provided (PIT) and the household does not have a HoH
+      # We will still want to check for a chronic adult in this case, so we'll use the earliest entry date in the HH
```
eanders: I believe the glossary says you can only inherit from non-HoH if the non-HoH member entered on the same date as the HoH.

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
```diff
@@ -158,6 +158,7 @@ def self.allowed_options(...)
             columns: pending_associations.values.first&.changes&.keys || [],
           },
         )
+        # binding.pry
```
eanders:
```suggestion
```

app/models/concerns/hud_reports/households.rb
dtgreiner: yes, that should be occurring further down. I was attempting to keep the code flow unchanged enough so I wouldn't have to split this out completely with a semi-redundant version for the PIT logic. I'll give it another run through to make sure that is happening or make sure it is more clear it is happening.

app/models/concerns/hud_reports/households.rb
dtgreiner: Thanks! I missed that here and can adjust. This was attempting to account for cases where a HOH didn't exist in the HH, but it sounds like we need to ignore other CH adults in this case as there is no HoH in which to compare.

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
dtgreiner: cleaned up in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5413/commits/2044154e993e7bc3528ff4d5b453c7257c5c05d9)

app/models/concerns/hud_reports/households.rb
dtgreiner: cleaned up in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5413/commits/2044154e993e7bc3528ff4d5b453c7257c5c05d9)

app/models/concerns/hud_reports/households.rb
dtgreiner: I removed this section and cleaned up the entry date checks in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5413/commits/2044154e993e7bc3528ff4d5b453c7257c5c05d9). The calculation was previously checking for current member entry date matching for an chronic adult to be used. The commit also updates this to check the hoh instead.

eanders (approved):
That looks right to me.  Thanks!

PR #5417, opened 2025-05-22
7559-nightly-census

ttoomey:
## _Merging this PR_

## Description
* Attempt to simplify census builder. This won't be faster but it might have a smaller peak memory footprint

## Type of change
Code clean-up


app/models/grda_warehouse/census/census_builder.rb
```diff
@@ -6,43 +6,106 @@
...
+    protected
 
-        # Move batch forward
-        batch_start_date = batch_end_date + 1.day
+    def populations
+      @populations ||= GrdaWarehouse::Census.census_populations.map { |p| p[:population] }.uniq
+    end
+
+    def relevant_projects
+      GrdaWarehouse::Hud::Project
```
ttoomey: maybe there's a smart way to limit projects

app/models/grda_warehouse/census/census_builder.rb
```diff
@@ -6,43 +6,106 @@
...
+      populations.each do |population|
+        project_population_counts(project, population).each do |date, count|
+          merged[date] ||= { project_id: project.id }
+          merged[date][population] = count
+        end
       end
+      inventories = project.inventories.within_range(start_date..end_date)
+      merged.keys.each do |date|
+        inventories.each do |inventory|
+          next unless inventory_active_on_date?(inventory, date)
```
ttoomey: I'm sure there's an optimization to be had to reduce the number of comparisons

ttoomey:
@eanders if this seems okay I'll test it on QA (current code doesn't run due to memory issues)

app/models/grda_warehouse/census/census_builder.rb
eanders: Probably something where we look to see if the project is operating (`Project.within_range(range)`, or if there are any open enrollments during the range.)

app/models/grda_warehouse/census/census_builder.rb
```diff
@@ -6,43 +6,106 @@
...
+      end_date = inventory.InventoryEndDate
+
+      # If we have a start date but no end date, active from start onward
+      return true if start_date && !end_date && start_date <= date
+
+      # If we have both start and end dates, check if date is in range (inclusive)
+      return true if start_date && end_date && date.between?(start_date, end_date)
+
+      false
+    end
```
eanders: This logic should move to the inventory record.  Analogous to the `within_range` scope already there.  It could take the same date twice or have an `active_on` method that just takes one date.

app/models/grda_warehouse/census/census_builder.rb
eanders: You could fetch the dates within the range where the inventory was active and convert it to a set, then it would just be a lookup of the date in question in the pre-calculated set.

eanders (approved):
This seems worth trying. 
I'd like to get to a place where these charts could just be generated from the source data (counts of Services or comparisons to Entry/Exit dates) but I think the performance of that in combination with the populations is probably not performant enough.

ttoomey:
@eanders I made some relatively small changes to the generator but probably worth another quick look before I merge in to 164:
* deleting only census data within start_date..end_date
* moved active_on_date? to inventory model
* only process projects within_range(start_date..end_date)
* add distinct on ServiceHistoryService count

I also tried to enhance the tests to be more comprehensive

spec/models/grda_warehouse/census/census_builder_spec.rb
```diff
@@ -9,128 +9,253 @@
...
-    @enrollment = create_enrollment(
-      client: @client,
-      project: @project,
-      entry_date: start_date,
-    )
-    (start_date..end_date).each do |date|
-      create_bed_night_service(enrollment: @enrollment, date: date)
+  # Shared context for a standard project with services
+  shared_context 'with a standard project and services' do
+    let!(:project) { create_project(project_type: 0) } # Default project type
```
eanders:
```suggestion
    let!(:project) { create_project(project_type: 0) } # ES Entry/Exit
```

spec/models/grda_warehouse/census/census_builder_spec.rb
```diff
@@ -9,128 +9,253 @@
...
+        client: client,
+        project: project,
+        entry_date: start_date, # Uses start_date from the outer scope
+      )
+    end
+
+    before(:each) do
+      # Uses start_date and end_date from the outer scope
+      (start_date..end_date).each do |date|
+        create_bed_night_service(enrollment: enrollment, date: date)
```
eanders: Where is this defined? Do we need to add bed-nights for entry/exit projects?  We shouldn't need records in the Service table, but we would want them in the service_history_services, which is what line 29 is doing.

eanders (approved):

spec/models/grda_warehouse/census/census_builder_spec.rb
ttoomey: > Where is this defined?

All these test helper methods are in spec/shared_contexts/hud_enrollment_builders.rb, I've been trying to build a library to assist with these reporting unit tests.

> Do we need to add bed-nights for entry/exit projects?

Ah nope, thanks!

app/models/grda_warehouse/census/census_builder.rb
ttoomey: thanks, I'm not sure but it may be the overhead of creating the sets would offset the gains in performance. I can make another optimization pass if it's a problem

PR #5418, opened 2025-05-22
CAS sync fix

eanders:
## _Merging this PR_

## Description
* Fix for [error in CAS Sync](https://green-river.sentry.io/issues/6621557287/?referrer=slack&notification_uuid=b0ca47e8-7c30-41f1-9d80-1a84cbf84710&environment=staging&alert_rule_id=14733065&alert_type=issue)

## Type of change
* Bug fix


ttoomey (approved):
Code looks good. So if safe_project_name is missing we exclude it rather than use a placeholder of project id or something? Assuming this is okay but I don't quite understand why the name is missing

eanders:
Yes, we're excluding if we can't find a safe name since we don't have enforcement of the confidentiality on the CAS side.  I think this ends up being a timing bug where the data isn't in one place, but is in the other, and the missing should get fixed once the cache is fully updated.  So potentially we're adding a delay to the data getting to CAS, but not completely excluding the data.

PR #5419, opened 2025-05-22
Support Referrals on User Dashboard

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add new field `ceAssignedSteps` on User.
- Resolve `referral` on `CeReferralStep`, and add a new custom data source to support this
- Add `Hmis::Ce::Referral` validation enforcing that workflow templates must be of type `ce_referral`

## Type of change
New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see frontend


drivers/hmis/app/models/hmis/user.rb
```diff
@@ -106,16 +106,35 @@ def load_effective_permissions
...
-      # This is a one-off custom logic permission for determining when to show "My Dashbord" in HMIS. It is resolved on the root access object.
+      # This is a one-off custom logic permission for determining when to show the User Dashboard in HMIS. It is resolved on the root access object.
       # This logic may evolve as we add more capabilities to the dashboard.
       # If we have more use-cases for this, we could make a helper in BaseAccess that accepts custom evaluation logic.
-      return false unless Hmis::ProjectStaffAssignmentConfig.exists? # micro optimization for installations without staff assignment
+      can_have_assigned_staff? || can_have_assigned_referral_tasks?
+    end
+  end
+
+  def can_have_assigned_staff?
```
martha: Flagging the choice to add these as root permissions, so we can discuss further.

I think they are "root" in the sense that they should be resolved on the user, because we are asking the question "does the user have the ability to see [staff assignments/referrals] in general?" which is not tied to a specific resource.

However (maybe this is more of a frontend discussion), I'm not sure they need to be loaded as part of the root permissions fragment, since they are only needed on a specific page. But I couldn't see that we have an existing pattern for this besides `useHasRootPermissions`. Should we introduce one? Maybe a new query, something like `GetDashboardRootPermissions`?

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -106,16 +106,35 @@ def load_effective_permissions
...
 
+  def can_have_assigned_referral_tasks?
+    key = [self.class.name, __method__, id]
+    Rails.cache.fetch(key, expires_in: 1.minutes) do
+      return false unless Hmis::Ce.configuration.enabled?
+
+      can_view_referrals = permissions?(:can_view_referrals, :can_view_own_referrals)
+      can_perform_referral_steps = permissions?(:can_perform_any_referral_tasks, :can_perform_own_referral_tasks)
+
+      can_view_referrals && can_perform_referral_steps
```
gigxz: another instance where policy class https://github.com/open-path/Green-River/issues/7720 can encapsulate complicated logic

drivers/hmis/app/models/hmis/user.rb
gigxz: Yeah I think it would make sense for there to be a dashboard-specific fragment.

So `can_have_assigned_staff` is really saying: do I have edit-access to any enrollments in projects that use the Staff Assignment feature? The naming is a bit confusing to me, the user can't "have" assigned staff. It's really whether they can _be_ assigned to a household. Maybe something like `can_be_assigned_to_households` is clearer?

gigxz: But, more holistically:

Since the Dashboard will become a configurable page with several additional cards eventually, I think it's worth moving away from a fake-permissions-style ("can view x thing on dashboard") towards something more structured, like a specific GraphQL query that resolves exactly what the page needs. 

That would be more flexible going forward, as we add additional cards and eventually the ability to have user- or system- configuration for which cards appear. The frontend can just display whatever it's told to.

Here is a rough idea (the field names are not proposals, just pulling out of thin air) -
```gql
query {
    myDashboard { # new GraphQL type
        # settings for what is enabled/disabled. new GraphQL type. later this might be backed by user configuration
        myDashboardConfig: { showStaffAssignment: boolean!, showProjectShortcuts: boolean!; showReferrals: boolean! }
 
        # queries to back cards/metrics on the user dashboard
        staffAssignments { } # assignments for current user
        projectShortcuts { } # projects that current user has edit-access to, if they have edit-access to <10 projects (idea discussed previously, just an example )
        reminders { } 
        ceReferrals { }
        ceReferralTasksDue { }
        # later: links or rich text configured by administrator
        # later: expiring rois for this user
        # etc
    }
    # alternative names for "myDashboard" could be "userDashboard" or "currentUserDashboard"
}
```

On the frontend, this might look like: `myDashboardConfig` fetched initially (to decide whether to even show "dashboard" on the main nav, should be inexpensive), then when the My Dashboard page is loaded, each card fetches their own query of what they need (which will get batched).

drivers/hmis/app/graphql/types/application/user.rb
```diff
@@ -42,6 +42,7 @@ class Application::User < Types::BaseObject
     field :manage_account_url, String, null: false
 
     field :staff_assignments, HmisSchema::StaffAssignment.page_type, null: true
+    field :ce_assigned_steps, HmisSchema::CeReferralStep.page_type, null: true
```
gigxz: Another benefit of the approach proposed on https://github.com/greenriver/hmis-warehouse/pull/5419/files#r2104889700 is that this field can entirely from the application `User`  type onto `MyDashboard` type, which means that access checks become simpler because we know we're only ever resolving the data (in this case, assigned steps) for the `current_user`.

In other words-- if this field is added to User, we probably need to add `raise unless current_user == object` because we shouldn't allow users to resolved assigned steps for _other_ users.

drivers/hmis/app/graphql/sources/ce_step_referral_source.rb
```diff
@@ -0,0 +1,26 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Sources
+  class CeStepReferralSource < GraphQL::Dataloader::Source
+    # Custom GraphQL Dataloader to resolve the relationship between steps and referrals.
```
gigxz: Nice!
Suggest clarifying comment to specify that this loader "fetches" `Hmis::WorkflowExecution::Step` records, and "resolves" the `Hmis::Ce::Referral` that each step belongs to. Not sure if I'm using the most accurate verbs there for Dataloader. But it would be helpful to clarify the input & output for future comprehension.

drivers/hmis/app/graphql/sources/ce_step_referral_source.rb
```diff
@@ -0,0 +1,26 @@
...
+    # Steps live under WorkflowExecution so do not have a direct relationship to CE Referrals.
+    # Since we can't use `load_ar_association`, this loader helps avoid n+1 issues.
+
+    def fetch(steps)
+      # Batch load referrals for the steps provided, and group them by workflow instance IDs
+      referrals_by_instance_id = Hmis::Ce::Referral.
+        where(workflow_instance_id: steps.map(&:instance_id)).
+        group_by(&:workflow_instance_id).
+        transform_values { |refs| refs.max_by(&:created_at) } # should be exactly 1 referral per instance ID
```
gigxz: suggest raising if there is >1 referral, since that is not expected

drivers/hmis/app/graphql/sources/ce_step_referral_source.rb
```diff
@@ -0,0 +1,26 @@
...
+
+      # In the same order as the input steps, return a list of referrals associated with each step's instance ID
+      steps.map { |step| referrals_by_instance_id[step.instance_id] }
```
gigxz: suggest raising if step doesn't have a corresponding referral, since that is not expected

drivers/hmis/app/graphql/sources/ce_step_referral_source.rb
martha: Thanks! I'm not sure of the right verbs to use either, but I don't think I would say it "fetches" Step records, those are the caller's responsibility. I will try to add more clarifying comments along the lines of what you are saying.

martha: Thank you for inspiring me to poke more around the [ruby graphql documentation for custom data sources](https://graphql-ruby.org/dataloader/sources.html). I ended up concluding that it might be more graceful to refactor this into a `CeReferralByInstanceIdSource`, so it accepts `instance_ids`, instead of `steps`. The documentation refers to the input as "keys", and having the keys be strings and not objects seems a little easier to follow logically, plus it may be useful in the future (for example, WFE audit events also store a reference to instance ID - though I don't know if we would ever resolve those outside the context of a referral).

I took a shot at the refactor and adding more clarifying comments. Let me know if you think it looks better!

drivers/hmis/app/graphql/types/application/user_dashboard.rb
```diff
@@ -0,0 +1,70 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class Application::UserDashboard < Types::BaseObject
+    # Type for resolving data needed on the HMIS user dashboard.
+    # Underlying object is Hmis::User, who should always be the currently logged-in user.
```
gigxz, line 12: Maybe it's worth adding authorization to this object so we can be sure:
```
    def self.authorized?(object, ctx)
      raise 'user dashboard can only be resolved for current user' unless object == current_user

      super
    end
```

drivers/hmis/app/graphql/types/application/user_dashboard.rb
```diff
@@ -0,0 +1,70 @@
...
+      {
+        id: object.id,
+        show_staff_assignment: show_staff_assignment,
+        show_referrals: show_referrals,
+      }
+    end
+
+    def staff_assignments
+      object.staff_assignments.
+        viewable_by(current_user).
```
gigxz: I think its a little unintuitive that we need the viewable_by scope even though we're already scoped to the users assignments, maybe worth a comment:
```suggestion
        viewable_by(current_user). # Only resolve assignments where the user has access to view the household
```

drivers/hmis/app/graphql/types/application/user_dashboard.rb
```diff
@@ -0,0 +1,70 @@
...
+        open_on_date. # This will include households that exited today
+        order(created_at: :desc, id: :desc)
```
gigxz: Scope-creeping here, but it is inconsistent that we sort households this way. It can result in "incomplete" households appearing below "active" households, which we don't do by default anywhere else.


Suggest adjusting this to use the existing household recency scope instead of sorting by assignment date:
```suggestion
        merge(Hmis::Hud::Household.sort_by_option(:most_recent))
```

drivers/hmis/app/graphql/types/application/user_dashboard.rb
```diff
@@ -0,0 +1,70 @@
...
+    end
+
+    def ce_referral_steps
```
gigxz, line 43: We still need to check whether show_referrals is true here, otherwise someone could resolve things they don't have access to (because the permission check doesn't occur on this field). Even though we know the frontend won't request it, we're still making it available in the API here.

```suggestion
    def ce_referral_steps
          return unless show_referrals
```

This could benefit from memoizing the private `show_` methods:

```
def show_referrals
  @show_referrals ||= begin

    can_view_referrals = object.permissions?(:can_view_referrals, :can_view_own_referrals)
    can_perform_referral_steps = object.permissions?(:can_perform_any_referral_tasks, :can_perform_own_referral_tasks)

    can_view_referrals && can_perform_referral_steps
  end
end
```

And I would suggest using the same pattern for staff assignments, even though that one seems safe enough since it does check viewable_by, but it would be nice to have a consistent pattern to ensure we always check if the feature is enabled before resolving it.

drivers/hmis/app/graphql/types/application/user_dashboard.rb
```diff
@@ -0,0 +1,70 @@
...
+
+    def user_dashboard_config
+      {
+        id: object.id,
+        show_staff_assignment: show_staff_assignment,
+        show_referrals: show_referrals,
+      }
+    end
+
+    def staff_assignments
```
gigxz, line 34: Same here, to be safe we should return early if user lacks permission

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -290,12 +290,13 @@ def pick_list(pick_list_type:, **args)
     end
 
     field :current_user, Application::User, null: true
+    field :user_dashboard, Types::Application::UserDashboard, null: false
```
gigxz, line 294: In this file we always(I think?) list the resolver method right below the field declaration. So I would suggest moving `def user_dashboard` up below this so its easier to find.

drivers/hmis/spec/factories/hmis/ce/steps.rb
```diff
@@ -0,0 +1,28 @@
...
+      assignees { nil }
+    end
+
+    association(:instance, factory: :hmis_workflow_execution_instance)
+    association(:node, factory: :hmis_workflow_definition_task)
+
+    status { 'available' }
+    available_at { Time.current }
+
+    after(:create) do |step, evaluator|
```
gigxz: Just a heads up, I recently was working on a spec test that used a factorybot with callback methods like this, and it caused the `make_database_queries` to include a bunch of extra queries for these callbacks. (For some reason it wasn't all happening  during setup). So it made it a bit more challenging to test performance. I ended up changing some to `after(:build)` which seemed to help, very much in-progress here https://github.com/greenriver/hmis-warehouse/pull/5429/files#diff-71efb5af46e5b2d40f2c34aaecc69124980c2a5f5b69ef66746bf26b8a152988 I am figuring out the unit factories.

Soo... no immediate feedback but more of an FYI if you hit weirdness with tests! I'm not quite sure what the best way is to resolve this, but I encountered the same concept elsewhere.

gigxz (commented):
Looks great!! some minor comments below

drivers/hmis/spec/factories/hmis/ce/steps.rb
martha: I ran into something similar this morning too, working on https://github.com/greenriver/hmis-warehouse/pull/5422.In [this file](https://github.com/greenriver/hmis-warehouse/pull/5422/commits/2a54001d24716e84edef6dafd9f274b1298f480a#diff-96007111abc00aed839d219e05f89579c77ace73340c658835c9dc01ae564d63R16-R18) I noticed that when test fixtures are built but not persisted before the test starts (`let` and not `let!` -- although I'm not totally clear on how this relates to the difference between `build` and `create`), then the `expect... to make_database_queries` block confusingly includes all queries to create those objects. At least I _think_ that is what was going on there. Anyway, adding `!` in the linked diff prevented the extra queries from screwing up the n+1 tests in that file.

martha: I'll update this to an `after(:build)` hook, and see how that impacts the test. I believe the assignment creation still needs to happen in an `after(:create)` hook.

martha:
@gigxz I think this is ready to merge, any hesitations?

gigxz:
Great! My only question mark is the addition of the `cleanup_test_environment` statements. It would be preferable to clean up whatever the source is of the problem, rather than introducing a wholesale cleanup, because it really slows the tests down, which in turn slows down CI. `admin_ce_referrals_spec.rb` takes 25 seconds for me to run locally. I'm not seeing any failures when I remove the cleanup step, are you seeing failures when its run in the suite?

martha:
@gigxz , that's a good point. I was never able to reproduce the failure locally (in fact, something is borked in my local environment with the cleanup statement so I have to comment it out to get the tests to run 🤦‍♀️ ). I added it because I was seeing intermittent failures on CI, and wasn't really sure how to troubleshoot them.

I can remove the cleanup statements for now and if the intermittent failures return I will try to troubleshoot further.

PR #5422, opened 2025-05-23
Add data source to template

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add migration to add `data_source_id` col to `wfd_template` table
- Update the `Hmis::WorkflowDefinition::Template.viewable_by` scope to check the data source ID against the current user
- Update both Template picklists to only return templates in the current data source
  - Also update the published Template picklist, so it now returns template Identifiers instead of database IDs. This is a breaking API change, but ok because these features aren't yet used in prod
- Add validation on Opportunity to enforce that the template is in the same data source as the target project

## Type of change
Code clean-up

### DIFF
--- a/db/warehouse/migrate/20250522193546_add_data_source_id_to_wfd_template.rb
+++ b/db/warehouse/migrate/20250522193546_add_data_source_id_to_wfd_template.rb
@@ -1,0 +1,32 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddDataSourceIdToWfdTemplate < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do # not yet used in prod, so safety_assured is ok
+      add_column :wfd_templates, :data_source_id, :integer, null: true
+
+      add_index :wfd_templates, :data_source_id
+
+      # We want the column to be required, but need to fill in a default value on all the existing records.
+      # Since they are non-prod records that were populated by the starter pack,
+      # we can just rerun the starter pack later to get the actual correct data source ID in each environment,
+      # so for now just use 0.
+      reversible do |dir|
+        dir.up do
+          execute <<~SQL
+          UPDATE wfd_templates
+          SET data_source_id = 0
+        SQL
+        end
+      end
+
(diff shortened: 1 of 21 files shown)


drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb
```diff
@@ -275,7 +275,7 @@
     end
 
     context 'when the opportunity has some candidates the current user lacks permission to view' do
-      let(:candidate_pool_with_anonymous) { create :hmis_ce_match_candidate_pool }
+      let(:candidate_pool_with_anonymous) { create :hmis_ce_match_candidate_pool, priority_expression: '1' } # overwrite priority expression with a different dummy value to avoid duplicate key
```
gigxz: Can we use a `sequence` in the hmis_ce_match_candidate_pool factory instead? It would be preferable if each time you create a new instance using the factory, it is valid.

db/warehouse/migrate/20250522193546_add_data_source_id_to_wfd_template.rb
```diff
@@ -0,0 +1,32 @@
...
+###
+
+# frozen_string_literal: true
+
+class AddDataSourceIdToWfdTemplate < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do # not yet used in prod, so safety_assured is ok
+      add_column :wfd_templates, :data_source_id, :integer, null: true
+
+      add_index :wfd_templates, :data_source_id
```
gigxz: suggest using add_reference instead to add a foreign key:
```
add_reference :wfd_templates, :data_source, foreign_key: true, null: true
```
(this would replace the add_column and add_index)

gigxz (approved):
👍

db/warehouse/migrate/20250522193546_add_data_source_id_to_wfd_template.rb
martha: Sounds good. We will have to split this out into 2 separate migrations, though, which are then run separately in any environment that already includes any templates. (so QA only, I believe.) The trick with setting them to 0 temporarily doesn't work if it is a foreign key.

We could also remove the nullability constraint.

martha: Opened https://github.com/greenriver/hmis-warehouse/pull/5431 to follow up.

PR #5423, opened 2025-05-23
Todo Or Die CI

eanders:
## _Merging this PR_

## Description

This PR adds CI functionality for TodoOrDie checks to systematically track and enforce TODO deadlines in the codebase. Key changes include:

- Added a new rake task to check for TodoOrDies
- Modified CI configuration to run these checks
- Configured to not fail for TodoOrDie checks during Rails tests

## Type of change
* New feature

### DIFF
--- a/lib/tasks/test.rake
+++ b/lib/tasks/test.rake
@@ -23,7 +23,7 @@ namespace :test do
     Sentry.capture_exception_with_info(
       StandardError.new("Testing Sentry.capture_exception_with_info from #{Rails.env} for hmis-warehouse"),
       'Testing custom error message',
-      { with: 'data' }
+      { with: 'data' },
     )
     sleep 1
 
@@ -31,7 +31,7 @@ namespace :test do
     Sentry.capture_exception_with_info(
       StandardError.new("Testing Sentry.capture_exception_with_info_no_exception from #{Rails.env} for hmis-warehouse"),
       'Testing custom error message',
-      { info: 'info' }
+      { info: 'info' },
     )
     sleep 1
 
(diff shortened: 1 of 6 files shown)


lib/tasks/test.rake
```diff
@@ -67,4 +67,98 @@ namespace :test do
     puts '@notifier.ping normal (Slack)'
     @notifier.ping('Testing .ping polymorphism - this should go to Slack')
   end
+
+  desc 'Evaluate all TodoOrDie calls in the codebase'
+  task todo_or_die: :environment do
```
ttoomey, line 80: Interesting, I'm a little unsure about this approach. Have you tried something like this before? 

Could we just load the whole application (eager_load) and then see what todo errors are generated? That way we don't need to use eval() or add the unparser dependency. It also seems a lot less complex

```ruby
namespace :test do
  desc 'Evaluate all TodoOrDie calls by loading the application'
  task todo_or_die: :environment do
    # Force eager loading to trigger all TodoOrDie calls
    Rails.application.eager_load!
   
  # could rescue although I think just letting the exception happen should work
  # rescue TodoOrDie::OverdueTodo => e
  #   puts "Overdue TODO found: #{e.message}"
  #   exit(1) if ENV['CI']
  end
end
```

eanders: So, I'm actually working through removing the `eval` now.  I have a better approach to that.  
I started hoping we could just eager load the app, but TodoOrDies aren't evaluated until they are called, so any inside methods were being ignored, which would mean running the test suite.  

This just searches the codebase, logs where it found them and then executes the TodoOrDie outside of context to determine if it is due.  

So far it seems to work well.
I'll see if I can remove the `unparser` dependency, I don't think I need it with the new approach.

eanders:
@ttoomey does this look better?  I know you aren't a fan of the approach, but at least this encapsulates the TodoOrDie calls into a file (no need for the dangerous `eval` or other `unparser` dependency.)

Maybe we could revisit with a different mechanism to replace TodoOrDie in the future.  

The current change:
1. Preserves the local development failure if someone is working on code that triggers the TodoOrDie
2. Prevents rails tests from failing because of a TodoOrDie
3. Exposes in CI that there are TodoOrDies that need handling, including indicating where in the code that TodoOrDie exists for easy cleanup

lib/tasks/test.rake
```diff
@@ -67,4 +75,111 @@ namespace :test do
...
+      puts "Syntax error in #{file}: #{e.message}"
+    rescue StandardError => e
+      puts "Error processing #{file}: #{e.message}"
+    end
+
+    puts "\nFound #{todo_calls.length} TodoOrDie calls"
+
+    # Create a temporary file with all TodoOrDie calls
+    require 'tempfile'
+    temp_file = Tempfile.new(['todo_or_die_calls', '.rb'])
```
ttoomey, line 111: this is fine, [there is also a block syntax](https://ruby-doc.org/stdlib-3.1.1/libdoc/tempfile/rdoc/Tempfile.html#method-c-create) that handles cleaning up the temp file
```
Tempfile.create('my_tempfile') do |file|
  #  write to file, etc
end
# temp file has been removed
```

ttoomey (commented):
Looks good, thanks!

PR #5425, opened 2025-05-24
7743 PIT corrections and tests

ttoomey:
## _Merging this PR_

## Description
* Corrects household count
* Includes households where all members are less than 18 in "parenting youth"
* Adjust condition on child counts in "parenting youth" to check relation to HOH instead of just age
* Adjust condition on child counts in "parenting youth" to check relation to HOH instead of just age
* adjust dv condition for clarity
* adds PIT unit tests

## Type of change
Bug fix, unit test


drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/parenting_youth.rb
```diff
@@ -11,6 +11,7 @@ class ParentingYouth < Base
     QUESTION_NUMBER = 'Parenting Youth Households'
 
     def self.filter_pending_associations(pending_associations)
+      # when the entire household is under 18, we don't include it
```
ttoomey: please confirm this is correct

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
```diff
@@ -82,7 +82,7 @@ def self.allowed_options(...)
           disabilities_latest = disabilities.select { |d| d.InformationDate == max_disability_date }
 
           dv_record = enrollment.health_and_dvs.
-            select { |h| h.InformationDate <= @generator.filter.on && !h.DomesticViolenceSurvivor.nil? }.
+            select { |h| h.InformationDate <= @generator.filter.on && h.DomesticViolenceSurvivor&.to_i == 1 }.
```
ttoomey: seems like this should be only 1 (Yes)

drivers/hud_pit/spec/unit/chronic_homelessness_spec.rb
```diff
@@ -0,0 +1,137 @@
+# frozen_string_literal: true
+
+# drivers/hud_pit/spec/models/chronic_homelessness_spec.rb
+
+require 'rails_helper'
+require_relative 'hud_pit_context'
+
+RSpec.describe 'Chronic Homelessness Calculations', type: :model do
```
ttoomey, line 8: note, I put these unit tests into a 'spec/unit' directory as models was already occupied by datalab tests.

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
```diff
@@ -364,7 +364,7 @@ def services_on_pit_date(batch)
         },
         chronic_households: {
           title: 'Chronically Homeless: Total number of households',
-          query: a_t[:chronically_homeless].eq(true).and(hoh_clause),
+          query: a_t[:chronically_homeless_household].eq(true).and(hoh_clause),
```
eanders, line 402:
```suggestion
          # Count heads of household in chronically homeless households (essentially count the households that are chronically homeless, multiple HoH or lack of HoH, bad data, will cause this to count incorrectly)
          query: a_t[:chronically_homeless_household].eq(true).and(hoh_clause),
```

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/parenting_youth.rb
eanders: Youth is typically 18-24 for HMIS, and parenting youth households would be households where everyone is under 24, and the parent is between 18 and 24.

That said, the glossary uses 12-24

> Youth Households
> The determination of a particular client being “youth” requires knowing who else is present with a client in the
> household at a project for a specific report date range. Using age as determined above, identify households where
> the age is known to be >= 12 and <= 24 for at least one household member and no member is known to be 25 or
> older. These are youth households. In these households, all clients >= 12 and <= 24 are considered to be “youth”.
> Note that clients in the household with an unknown age do not impact the determination of “youth.”

It might be worth looking at the programming specifications in the Glossary for **Q1. Report Validation Table** row 14.

If the entire household is <18, I believe it should be counted in the child-only category, but we should confirm.

ttoomey: Thanks! I read the spec like we need to include these households but let's check in on it tomorrow

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
eanders: I believe the intention was to find the most-recent record where the question was answered, which is different from the most-recent "yes" answer.

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/parenting_youth.rb
eanders: **Parenting Youth:**
* HoH is between 12 and 24
* Someone in the household has a *Child* relationship to the HoH
* No one over the age of 24

**Number of Parenting Youth (under age 18)**
* Filter Parenting youth for HoH < 18

**Number of Parenting Youth (age 18 to 24)**
* Filter Parenting youth for HoH 18 to 24

**Unaccompanied Youth** categories)
* Whole Household 18-24 (or maybe all <= 24)

**Child Only Household**
* Whole Household < 18

See https://www.hud.gov/sites/dfiles/CPD/documents/2024-HIC-and-PIT-Data-Collection-Notice-2023-11-02.pdf for details

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
ttoomey: backed this change out after discussion

db/warehouse/migrate/20250528000208_add_household_has_minor_children_to_hud_report_pit_clients.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+class AddHouseholdHasMinorChildrenToHudReportPitClients < ActiveRecord::Migration[7.1]
+  def change
+    add_column :hud_report_pit_clients, :household_has_minor_children, :boolean
```
ttoomey: this supports filter on parenting youth (non-adult children in household)

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
```diff
@@ -292,15 +299,29 @@ def services_on_pit_date(batch)
...
+          query: [
+            child_of_hoh_clause,
+            child_clause,
+          ].reduce(&:and),
         },
         children_of_18_to_24_parents: {
+          # https://www.hudexchange.info/faqs/3470/how-should-two-parent-parenting-youth-households-be-reported-in-the-pit/
+          # count children under oldest parent (aged 18-24)
           title: 'Children in households with parenting youth age 18 to 24 (children under age 18 with parents age 18 to 24)',
-          query: a_t[:head_of_household].eq(false).and(child_clause).and(a_t[:hoh_age].in(18..24)),
```
ttoomey, line 299: I think it's more correct to check the max age of any adult given the HUD guidance. Or perhaps we should check the max age of household members who are HOH or spouse of HOH

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/chronic_homelessness_spec.rb
```diff
@@ -0,0 +1,141 @@
...
+        expect(total_households.value).to eq(1)
+
+        total_persons = report.answer(question: question, cell: 'B3')
+        expect(total_persons.value).to eq(3)
+
+        chronic_households = report.answer(question: question, cell: 'B26')
+        expect(chronic_households.value).to eq(1)
+
+        chronic_persons = report.answer(question: question, cell: 'B27')
+        expect(chronic_persons.value).to eq(3)
```
eanders: It would be great to have some comments listing which values are going in these cells, as in "CH household" or "People CH"

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_basic_counts_spec.rb
```diff
@@ -0,0 +1,427 @@
...
+        hoh2 = create_client_with_warehouse_link(uid: 'client_tp_2b_hoh', dob: adult_dob)
+        create_enrollment(client: hoh2, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_hoh, household_id: hh_id2)
+        child2 = create_client_with_warehouse_link(uid: 'client_tp_2b_child', dob: child_dob_under_18)
+        create_enrollment(client: child2, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_child, household_id: hh_id2)
+      end
+
+      it 'counts total persons across all households' do
+        report = run_report(questions: [question])
+        total_persons = report.answer(question: question, cell: 'B3')
+        expect(total_persons.value).to eq(4) # 2 from hh1 + 2 from hh2
```
eanders: I don't think the comments align with the data.  It looks like `hh_total_persons_2b` has two people and the test if for 4 total, but the comment `# Household 2 (1 HoH) -> 1 person` indicates only one.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_basic_counts_spec.rb
```diff
@@ -0,0 +1,427 @@
...
+        create_enrollment(client: hoh_client, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_hoh, household_id: household_id)
+
+        child_1 = create_client_with_warehouse_link(uid: 'client_child_c1', dob: dob_child_5)
+        create_enrollment(client: child_1, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_child, household_id: household_id)
+
+        child_2 = create_client_with_warehouse_link(uid: 'client_child_c2', dob: dob_child_17)
+        create_enrollment(client: child_2, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_child, household_id: household_id)
+
+        # Add an adult who is not a child to ensure they are not counted
+        adult_member = create_client_with_warehouse_link(uid: 'client_child_adult', dob: dob_adult_18)
```
eanders: I'm not clear on why we have the `uid` value, but it is weird to call this one `client_child_adult` when the test is to confirm they are an adult.  Could it just be `client_adult`?

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_basic_counts_spec.rb
```diff
@@ -0,0 +1,427 @@
...
+        adult_member = create_client_with_warehouse_link(uid: 'client_youth_adult', dob: dob_adult_25)
+        create_enrollment(client: adult_member, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_other_adult, household_id: household_id)
+      end
+
+      it 'counts only persons aged 18-24' do
+        report = run_report(questions: [question])
+        # Number of Persons (18 - 24) is B5 for AdultAndChild question
+        youth_count = report.answer(question: question, cell: 'B5')
+        expect(youth_count.value).to eq(3)
+      end
```
eanders: Since you bothered to setup a complicated family, it seems like testing for `0` in some of the youth questions would be relevant.  I _think_ having someone 25+ should ensure those are all 0.  Also, it might be worth having some clients of unknown age (there might be later, just haven't seen any) to confirm the households get counted in the expected buckets.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_basic_counts_spec.rb
```diff
@@ -0,0 +1,427 @@
...
+    end
+
+    context 'for Persons Ages 55-64' do
+      let(:question) { HudPit::Generators::Pit::Fy2025::Adults::QUESTION_NUMBER }
+      let(:dob_age_55) { pit_date - 55.years } # Exactly 55
+      let(:dob_age_60) { pit_date - 60.years } # Exactly 60
+      let(:dob_age_64) { pit_date - 64.years } # Exactly 64
+      let(:dob_age_54) { pit_date - 54.years } # Exactly 54 (too young)
+      let(:dob_age_65) { pit_date - 65.years } # Exactly 65 (too old)
```
eanders, line 411: It feels like some of these age bucket tests could be DRYed up significantly by creating a hash of ages and looping over them.  No harm in the way it is written, it just feels like a LOT of code.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+      end
+
+      it 'counts the HoH as a parenting youth 18-24' do
+        report = run_report(questions: [question])
+        parenting_youth_18_24_count = report.answer(question: question, cell: 'B8')
+        expect(parenting_youth_18_24_count.value).to eq(1)
+      end
+    end
+
+    context 'when HoH (18-24) and Spouse (18-24) are present with a child (qualifying PYH with two parents)' do
```
eanders, line 55: Generally we would only count the HoH as a parent, as in "parent" is a stand-in for "household" but maybe this count is different.  Is there some documentation that lead you to include the spouse relationship?

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+        create_enrollment(client: hoh_client, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_hoh, household_id: household_id)
+        other_youth_adult = create_client_with_warehouse_link(uid: 'py_other_youth_no_child_adult', dob: dob_youth_18)
+        create_enrollment(client: other_youth_adult, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_other_adult, household_id: household_id)
+      end
+
+      it 'counts 0 due to household_type filter (adults_only)' do
+        report = run_report(questions: [question])
+        parenting_youth_18_24_count = report.answer(question: question, cell: 'B8')
+        expect(parenting_youth_18_24_count.value).to eq(0)
+      end
```
eanders: Can you add a confirmation check that the household _is_ counted in the adult only category?  I believe this check is just confirming they aren't counted in the parenting youth.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+        create_enrollment(client: child_client, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_child, household_id: household_id)
+
+        older_adult = create_client_with_warehouse_link(uid: 'py_adult_25', dob: dob_adult_25)
+        create_enrollment(client: older_adult, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_other_adult, household_id: household_id)
+      end
+
+      it 'counts 0 due to max_age filter' do
+        report = run_report(questions: [question])
+        parenting_youth_18_24_count = report.answer(question: question, cell: 'B8')
+        expect(parenting_youth_18_24_count.value).to eq(0)
```
eanders: Again, I think a positive check here for the expected category would be nice, otherwise, we don't really know where they are ending up, just guessing.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+      it 'counts HoH and spouse as Number of parenting youth (under age 18)' do
+        report = run_report(questions: [question])
+        parenting_youth_under_18_count = report.answer(question: question, cell: 'B6')
+        expect(parenting_youth_under_18_count.value).to eq(2)
+        children_in_parenting_youth_under_18_count = report.answer(question: question, cell: 'B7')
+        expect(children_in_parenting_youth_under_18_count.value).to eq(2)
+      end
+    end
+
+    context 'when HoH is <18, but no child present (household is adults_only)' do
```
eanders: This context doesn't match the test
```suggestion
    context 'when HoH is <18, and no other child present (household is child_only)' do
```

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+        older_adult = create_client_with_warehouse_link(uid: 'py_adult_25_u18hh', dob: dob_adult_25)
+        create_enrollment(client: older_adult, project: es_project, entry_date: pit_date, relationship_to_ho_h: rel_other_adult, household_id: household_id)
+      end
+
+      it 'counts 0 due to max_age filter' do
+        report = run_report(questions: [question])
+        parenting_youth_under_18_count = report.answer(question: question, cell: 'B6')
+        expect(parenting_youth_under_18_count.value).to eq(0)
+      end
+    end
```
eanders: It is unclear to me which question we are testing.  Maybe instead of "filters out HH" in the context it should say, `HH is excluded from xxx question` (and again, I think a positive check would be helpful as well.)

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
```diff
@@ -0,0 +1,507 @@
...
+
+    context 'when HoH is 18, with a child (single adult HH, but HoH wrong age for this cell)' do
```
eanders, line 381: Same here, some indication of the question being checked would be helpful

eanders (approved):
@ttoomey I think the code changes look good.  The tests are somewhat verbose and could use a bit of a human touch. 

My only remaining question is if we think a spouse could count as a parenting youth, and where that's documented.

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_basic_counts_spec.rb
ttoomey: good catch. The helper uses the uid to set the personal_id of the source client. Descriptive ids are helpful when debugging these report tests

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/pit_parenting_youth_spec.rb
ttoomey: per our discussion this morning, this is a reasonable interpretation of:
https://www.hudexchange.info/faqs/3470/how-should-two-parent-parenting-youth-households-be-reported-in-the-pit/

But it's not totally clear. This PR changes the logic to treat any adult as "parenting" so children are counted under the oldest adult in the household, not just the HOH

drivers/hud_pit/spec/models/hud_pit/generators/pit/fy2025/chronic_homelessness_spec.rb
ttoomey: added a helper method 
```diff
-  report.answer(question: question, cell: 'B27')
+ report_value(report, question: question, row: :chronically_homeless_persons)
```

ttoomey:
Thanks for the feedback @eanders - I tried to replace all the cell numbers with descriptive names.

This should probably get QA, maybe can go out in 164 tomorrow?

eanders (approved):

drivers/hud_pit/app/models/hud_pit/generators/pit/fy2025/base.rb
```diff
@@ -77,50 +78,68 @@ def self.allowed_options(...)
...
-          hoh_veteran = household_member_data(last_service_history_enrollment).detect do |member|
-            next unless member
-
+          hoh_veteran = household_member_data(last_service_history_enrollment).any? do |member|
             member['veteran_status'] == 1 && member['relationship_to_hoh'] == 1
-          end.present?
+          end
+
+          # calculate household (hh) stats.
+          # Note these should be the same across all households and could be cached
```
eanders: Should this be:?
```suggestion
          # Note these should be the same across all household members and could be cached
```

eanders (approved):
Thanks @ttoomey !

PR #5427, opened 2025-05-27
7751-axlsx-fix

ttoomey:
## _Merging this PR_

## Description
Second attempt at fixing intermittent render bug. Guessing that the axlsx template renderer might be missing

## Type of change
Bug fix


app/models/grda_warehouse/cohorts/document_exports/cohort_excel_export.rb
```diff
@@ -63,7 +67,6 @@ def perform
         write_tmp_file(
           renderer.render(
             template: 'cohorts/export',
-            action: :show,
```
ttoomey, line 66: this got left by mistake but I don't think it matters

app/models/grda_warehouse/cohorts/document_exports/cohort_excel_export.rb
```diff
@@ -81,6 +84,24 @@ def downloadable_file=(file_io)
       self.mime_type = EXCEL_MIME_TYPE
     end
 
+    private def preflight_checks
```
eanders, line 88: This all feels somewhat excessive given that we have many XLSX downloads across the app.  Is this just some additional instrumentation that we should take out in the future if it doesn't come up as an issue again?

ttoomey: > This all feels somewhat excessive given that we have many XLSX downloads across the app. Is this just some additional instrumentation that we should take out in the future if it doesn't come up as an issue again?

For sure. It's unclear what is causing the issue. We can remove this if it doesn't come up but it should be harmless

ttoomey, line 88:
```suggestion
    # additional instrumentation that we should take out in the future once we resolve #7751
    private def preflight_checks
```

eanders (approved):

PR #5429, opened 2025-05-28
HMIS Unit Groups

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7529
* models/db:
  * Add `Hmis::UnitGroup` model and table
  * Add unit group relationship to `Hmis::Unit`
* CE engine:
  * Rework `OpportunityApplicability` into `MatchApplicability` so we can determine the applicable rules for all entity types (not just opportunities) for display purposes. Also support evaluating rules on Unit Groups
* GraphQL:
  * Added a type for UnitGroup, but mostly left TODOs for issue 7753
  * Added a spec test for resolving the UnitGroup and associated rules/units

**Open question** about `priorityScheme`: it seems like we still need logic for choosing one scheme (or constraints for only allowing 1 scheme). Left those as comments for now.

**How to test**: I tested relationships and tables in the rails console, and spec test for the new graphql types. The graphql side is expanded in https://github.com/open-path/Green-River/issues/7753 which covers frontend management for unit groups. That one is more testable in the ui. https://github.com/greenriver/hmis-warehouse/pull/5444

**Not included:**
* unit group "tag" concept for exclusivity in matching
* unit "variant" concept for different behavior for housing units / voucher units / service capacity units. can be added later.

- [x] update starter pack to set up unit group
- [x] update wfd template selection when creating opportunities from mutation

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/match/match_applicability.rb
```diff
@@ -0,0 +1,99 @@
...
+#   rules for each entity type.
+# - Global and Data Source-level rules are not yet supported (e.g. "all PSH projects in a data source")
+#
+#
+#
+# @attr [Object] owner Entity that owns/manages this applicability rule
+# @attr [Array<String>] project_types List of valid project types
+# @attr [Array<String>] project_funders List of valid project funder IDs
+module Hmis::Ce::Match
+  MatchApplicability = Struct.new(:owner, :project_types, :project_funders, keyword_init: true) do
```
gigxz, line 30: This replaces the `OpportunityApplicability` class. The main difference is that it can resolve applicability for any entity type, not just opportunities.

So you can ask:
* Does this rule apply to this Project?
* Does this rule apply to this Unit Group?
* etc.

This allows us to support future interfaces where we can display applicable rules on the Project page, on the Unit Group page, etc. As opposed to just on the Unit page.

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -40,9 +57,26 @@ class Hmis::Unit < Hmis::HmisBase
...
+  alias_attribute :date_updated, :updated_at # remove?
   alias_attribute :date_created, :created_at
 
+  attribute :variant, :string
+  enum variant: {
+    dwelling: 'dwelling',
+    voucher: 'voucher',
+    service_slot: 'service_slot',
+    shelter: 'shelter',
+  }
```
gigxz: These are a bit TBD, still need to get feedback/workshop

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -26,6 +26,7 @@ def resolve(unit_ids:)
 
       # TODO(#7522) - template should be determined by context (project, unit type, ...)
       # For now, if you are using the "starter pack," this picks the template that creates an enrollment
+      # TODO if units belong to unit group, pick the wfd from that group
```
gigxz: todo @gigxz in scope

gigxz:
I have a couple things to finish up, marking as draft again. But feel free to begin reviewing, and sorry its big! This should match what we discussed yesterday @ttoomey @martha .

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
gigxz: added

db/warehouse/migrate/20250523195117_add_hmis_unit_groups.rb
```diff
@@ -0,0 +1,24 @@
...
+  def change
+    create_table :hmis_unit_groups do |t|
+      t.string :name, null: false
+      t.references :project, null: false, foreign_key: { to_table: :Project }
+      t.string :workflow_template_identifier, null: true # which workflow template to use to fill units in this group. use identifier instead of db id to avoid tying to a particular version of the template. nullable for non-CE usage
+      t.timestamps
+      t.timestamp :deleted_at
+    end
+
+    add_column :hmis_units, :hmis_unit_group_id, :integer, null: true
```
gigxz, line 19: to discuss: populating all existing units into groups, so that we don't need to deal with supporting "floating" units in the frontend. can be deferred to another pr

drivers/hmis/app/models/hmis/unit_group.rb
```diff
@@ -0,0 +1,45 @@
...
+      Hmis::Ce::Match::Rule.priority_scheme.for_entity(self).first # TODO enforce 1 priority scheme?
+    end
+
+    private
+
+    def workflow_template_is_valid
+      return unless workflow_template
+
+      errors.add(:workflow_template_identifier, 'must be published') unless workflow_template.published?
+      # errors.add(:workflow_template_identifier, 'must belong to the same data source') if workflow_template.data_source_id != project.data_source_id
```
gigxz: pull in https://github.com/greenriver/hmis-warehouse/pull/5431 and add back

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -24,15 +24,21 @@ def resolve(unit_ids:)
       project = Hmis::Hud::Project.find_by(id: project_ids.first)
       access_denied! unless current_user.permissions_for?(project, :can_manage_units)
 
-      # TODO(#7529) - template should be determined by context (project, unit type, ...)
-      # For now, if you are using the "starter pack," this picks the template that creates an enrollment
-      template = Hmis::WorkflowDefinition::Template.viewable_by(current_user).last
-      raise unless template.present?
+      # Determine which template to use for each unit based on Unit Group configuration.
+      # TODO(#7753) Once we finalize the frontend and adjust this mutation, we may want to:
+      # - Raise if called for units belonging to several unit groups (maybe allowed, tbd?)
```
martha: To me it seems ok to allow bulk-marking-available units from different unit groups. For example, say my project has a building with 10 walkup units and 2 accessible units. (I think) those go in 2 different unit groups, since they require different workflow templates, but it sounds fine for them to all be marked as available at once.

Also if I'm understanding the proposed ux for [#7773](https://github.com/open-path/Green-River/issues/7773) correctly, then it's just a bulk action from the unit management screen, so it's certainly simpler to just allow it.

drivers/hmis/app/models/hmis/unit_group.rb
```diff
@@ -0,0 +1,45 @@
...
+    end
+
+    private
+
+    def workflow_template_is_valid
+      return unless workflow_template
+
+      errors.add(:workflow_template_identifier, 'must be published') unless workflow_template.published?
+      errors.add(:workflow_template_identifier, 'must belong to the same data source') if workflow_template.data_source_id != project.data_source_id
+      errors.add(:workflow_template_identifier, 'must have a template type of ce_referral') if workflow_template.template_type == :ce_referral
```
ttoomey: Sorry if I'm not understanding but is this right? Should it be:
```suggestion
      errors.add(:workflow_template_identifier, 'must have a template type of ce_referral') unless workflow_template.template_type == :ce_referral
```

db/warehouse/migrate/20250523195117_add_hmis_unit_groups.rb
ttoomey: I think we should use either `bigint` or `references` for pks, we're trying to get away from mismatched pk types.
```suggestion
    add_reference :hmis_units, :hmis_unit_group # adds index by default
```

db/warehouse/migrate/20250523195117_add_hmis_unit_groups.rb
```diff
@@ -0,0 +1,24 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddHmisUnitGroups < ActiveRecord::Migration[7.1]
+  def change
+    create_table :hmis_unit_groups do |t|
+      t.string :name, null: false
+      t.references :project, null: false, foreign_key: { to_table: :Project }
```
ttoomey, line 13: do you feel confident that unit groups are per-project?

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -24,15 +24,21 @@ def resolve(unit_ids:)
...
+        raise 'Unable to mark unit available because there is no associated workflow template' unless workflow_template.present?
+
+        [unit.id, workflow_template]
+      end.to_h
 
       candidate_pool_resolver = Hmis::Ce::Match::CandidatePoolResolver.new
 
       Hmis::Unit.transaction do
         opportunities = units.map do |unit|
+          template = unit_to_template[unit.id] || random_template
```
martha: `random_template` maybe can be removed here? It looks like the raise on line 32 should mean `unit_to_template[unit.id]` is never nil

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -192,13 +193,18 @@ def unit_types
...
 
-    # TODO use dataloader
     def units(**args)
       return Hmis::Unit.none unless current_permission?(entity: object, permission: :can_view_units)
 
       resolve_units(**args)
     end
 
+    def unit_groups
+      return Hmis::Unit.none unless current_permission?(entity: object, permission: :can_view_units)
```
martha:
```suggestion
      return Hmis::UnitGroup.none unless current_permission?(entity: object, permission: :can_view_units)
```

drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb
```diff
@@ -0,0 +1,35 @@
...
+    field :workflow_template_identifier, String, null: true
+    field :workflow_template_name, String, null: true
+    # TODO(#7538) resolve default contacts for workflow template
+
+    def workflow_template_name
+      object.workflow_template&.name
+    end
+
+    def units(**args)
+      return Hmis::Unit.none unless current_permission?(entity: object.project, permission: :can_view_units)
```
martha: Doesn't seem bad to include extra permissions checks where they're not needed, but am I correctly understanding that this schema object, `HmisSchema::UnitGroup`, should already not be resolved unless the user has the `can_view_units` on this project in the first place? Maybe worth a comment, since it did give me pause -- "wait, what combination of permissions would grant the user access to view the unit group but _not_ the unit?"

drivers/hmis/app/models/hmis/ce/match/match_applicability.rb
```diff
@@ -0,0 +1,99 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# The `MatchApplicability` class determines whether a Coordinated Entry (CE) rule applies to a given entity
+# (e.g., Opportunity, Unit, Unit Group, Project, or Organization) based on project criteria.
+#
+# This is NOT determining client is eligibility itself; rather it determines which rules to use for eligibility/priority.
+#
```
martha:
```suggestion
```
repeated below

drivers/hmis/app/models/hmis/ce/match/match_applicability.rb
```diff
@@ -0,0 +1,99 @@
...
+#   should be used for eligibility/priority.
+# - It supports evaluation against various entities (eg Project, Unit, etc) to support viewing applicable
+#   rules for each entity type.
+# - Global and Data Source-level rules are not yet supported (e.g. "all PSH projects in a data source")
+#
+#
+#
+# @attr [Object] owner Entity that owns/manages this applicability rule
+# @attr [Array<String>] project_types List of valid project types
+# @attr [Array<String>] project_funders List of valid project funder IDs
```
martha: this is a nit, but the word "valid" is a little confusing here; it's more like "[project types/funders] to which this rule applies," right?

drivers/hmis/app/models/hmis/ce/match/rule.rb
```diff
@@ -2,6 +2,25 @@
...
+# and priority schemes.
+#
+# Rules are associated with an "owner" (e.g., a unit, project, or organization),
+# which determines the scope of applicability. The `applicability_config` can further refine this
+# scope by specifying conditions like project types or funders.
+#
+# Examples:
+#
+#    Rule Applies to All Emergency Shelter (ES) Projects in HMIS:
+#      { owner: hmis_data_source, applicability_config: { project_types: [0, 1] } }
```
martha, line 15: > Rule applies to all ES projects in HMIS

Based on the comments above, in `MatchApplicability`, this example is not yet supported, right?

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -24,15 +24,21 @@ def resolve(unit_ids:)
...
-      raise unless template.present?
+      # Determine which template to use for each unit based on Unit Group configuration.
+      # TODO(#7753) Once we finalize the frontend and adjust this mutation, we may want to:
+      # - Raise if called for units belonging to several unit groups (maybe allowed, tbd?)
+      unit_to_template = units.preload(:unit_group).map do |unit|
+        workflow_template = unit.unit_group&.workflow_template
+        raise 'Unable to mark unit available because there is no associated workflow template' unless workflow_template.present?
+
+        [unit.id, workflow_template]
+      end.to_h
```
ttoomey: small adjustments (untested)
* might as well preload workflow template
* present? shouldn't be needed
* can use to_h with a block rather than map
```suggestion
      unit_to_template = units.preload(unit_group: :workflow_template).to_h do |unit|
        workflow_template = unit.unit_group&.workflow_template
        raise 'Unable to mark unit available because there is no associated workflow template' unless workflow_template

        [unit.id, workflow_template]
      end
```

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
ttoomey: What's `random_template`? Maybe leftover development scaffold?
```suggestion
          template = unit_to_template.fetch(unit.id)
```

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -15,9 +15,9 @@ class Opportunity < GrdaWarehouseBase
                class_name: 'Hmis::WorkflowDefinition::Template'
 
     has_many :referrals, class_name: 'Hmis::Ce::Referral', dependent: :restrict_with_exception
-    has_many :categorizations, class_name: 'Hmis::Ce::OpportunityCategorization', foreign_key: :opportunity_id
+    has_many :categorizations, class_name: 'Hmis::Ce::OpportunityCategorization', foreign_key: :opportunity_id, dependent: :destroy
     has_many :categories, through: :categorizations
-    belongs_to :owner, polymorphic: true, optional: true # Hmis::Unit, ...
+    belongs_to :owner, polymorphic: true, optional: true # always an Hmis::Unit
```
martha, line 20: I think we originally decided to make this relationship polymorphic because we wanted Opportunity to potentially have other owners besides (the more narrowly defined Housing) Unit. And, now that we decided to reframe Unit as the more generically defined Unit of Capacity (of housing, service, etc.), we don't ever intend to make use of the polymorphic relationship.

Is that right? If so, would we want to consider making the migration from polymorphic to an explicitly named column? (migrations are still relatively inexpensive on these tables now, since they are not in production use yet) Or otherwise adding comments to try to avoid future confusion? Or do we think we still might make use of the polymorphic owner in the future?

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -6,14 +6,24 @@
...
 # frozen_string_literal: true
 
+# The `Hmis::Unit` model represents a generic unit of capacity in a project.
+# A unit is a resource that can be provided to a household or individual being served by the program.
+# Units can represent physical or virtual resources.
+#
+# Note: The term "unit" is intentionally generic and does not exclusively refer to an "apartment unit."
+# It may represent a unit of housing, a shelter bed, a shelter room, a voucher, a unit of service capacity, etc.
+# Units can optionally belong to a `UnitGroup`, and may have an associated descriptive `UnitType`.
+# Since a Unit may represent physical housing, the same Unit can be occupied, released, and re-occupied over time. (Unlike CE Occupancy records which are "single-use")
```
martha:
```suggestion
# Since a Unit may represent physical housing, the same Unit can be occupied, released, and re-occupied over time. (Unlike CE Opportunity records which are "single-use")
```
?

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -24,15 +24,21 @@ def resolve(unit_ids:)
...
-      # TODO(#7529) - template should be determined by context (project, unit type, ...)
-      # For now, if you are using the "starter pack," this picks the template that creates an enrollment
-      template = Hmis::WorkflowDefinition::Template.viewable_by(current_user).last
-      raise unless template.present?
+      # Determine which template to use for each unit based on Unit Group configuration.
+      # TODO(#7753) Once we finalize the frontend and adjust this mutation, we may want to:
+      # - Raise if called for units belonging to several unit groups (maybe allowed, tbd?)
+      unit_to_template = units.preload(:unit_group).map do |unit|
+        workflow_template = unit.unit_group&.workflow_template
+        raise 'Unable to mark unit available because there is no associated workflow template' unless workflow_template.present?
```
martha: Maybe add a comment here explaining that we do the extra db query to join to `workflow_template` (even though ultimately we only need a `workflow_template_identifier` when saving the opportunity) so that we can validate that the workflow template really exists, and is published. ?

drivers/hmis/spec/factories/hmis/unit_groups.rb
```diff
@@ -0,0 +1,15 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+FactoryBot.define do
+  factory :hmis_unit_group, class: 'Hmis::UnitGroup' do
+    sequence(:name) { |n| "Unit Group #{n}" }
+    project { association :hmis_hud_project }
+    workflow_template { association :hmis_workflow_definition_template, data_source: project.data_source }
```
martha, line 13: I guess this is another example of the `data_source` factory trickiness that we've been discussing. If this factory is called with a `workflow_template` passed in but not a `project`, it will generate an invalid unit group. But it will fail immediately due to the validations on unit group, so that seems fine.

This solves the problem slightly differently from the pattern established in the Opportunities factory, which accepts (or generates) a transient data source and then passes it to both project and template.

Not sure what pattern I prefer, just noting the discrepancy. Feel free to update the opportunity factory if you want to establish this pattern all around!

drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb
```diff
@@ -0,0 +1,35 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::UnitGroup < Types::BaseObject
```
ttoomey, line 10: maybe include some documentation on the intent of the class
```suggestion
  # Represents a logical grouping of Hmis::Units within a specific Hmis::Hud::Project.
  #   - Supports Coordinated Entry (CE) Configuration:
  #     - Associates a specific `workflow_template` (Hmis::WorkflowDefinition::Template)
  #       to be used for CE Opportunities created for units within this group.
  #     - Enables the definition of CE `eligibility_requirements` and `priority_scheme`
  #       rules (Hmis::Ce::Match::Rule) that apply to all units in the group
  class HmisSchema::UnitGroup < Types::BaseObject
```

drivers/hmis/app/models/hmis/unit_group.rb
gigxz: hmmmm yep totally, I'll fix and look into why that wasn't flagging locally! 
edit: fixed

martha (approved):
Everything looks great. Thanks Gig! 🚢

db/warehouse/migrate/20250523195117_add_hmis_unit_groups.rb
gigxz: Originally I had `add_reference` but I was nervous about using `safety_assured` (or `, index: {algorithm: :concurrently}`) since this table is in use.  It seems like in similar recent migrations we just used `safety_assured` so I can switch back to that, and do more research onto whether its really safe

drivers/hmis/app/models/hmis/ce/opportunity.rb
ttoomey, line 20: suggest simplifying the polymorphic relation if we are always referencing a unit

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
gigxz: Ok, fair enough to continue to allow it. I'll remove this comment.

Fwiw in your example, I think the "building" can often still be representable as a single unit group. From my understanding in that kind of scenario, the same workflow would be used for all units in the building, and the accessible units would just have some additional eligibility criteria.

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
gigxz: Good catch thanks!

drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb
gigxz: Good point- I will go ahead and remove this for performance and add a comment explaining why!

drivers/hmis/app/models/hmis/ce/match/match_applicability.rb
gigxz: yeah agreed, updated!

drivers/hmis/app/models/hmis/ce/match/rule.rb
gigxz: Yes you're right. I think we will likely need to support this later.

drivers/hmis/app/models/hmis/unit.rb
gigxz: Yes! good catch

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -310,6 +310,17 @@ task ce_starter_pack_20250302: [:environment] do
...
+  # create a UnitGroup in ce project
+  unit_group = Hmis::UnitGroup.find_or_create_by(
+    project: ce_project,
+    name: 'Admin Approve Denial Unit Group',
+    workflow_template: admin_approval_template,
+  )
+  # if there are any units in the project that don't have a unit group, assign them to this one
+  ce_project.units.where(unit_group: nil).each { |u| u.update!(unit_group: unit_group) }
+  # clean up any dangling opportunities that don't have an owner
+  Hmis::Ce::Opportunity.all.filter { |o| o.owner.nil? }.each { |u| u.referrals.each(&:destroy!) }.map(&:destroy!)
```
martha, line 322: Hm so the first time I ran locally this I got an error `ActiveRecord::DeleteRestrictionError: Cannot delete record because of dependent referrals` But the second time it succeeded. Weird, maybe it just needs a judicious `reload` or something so it "knows" it doesn't have any referrals left after they have all been destroyed? 

Anyway, I think this is no big deal since it's just a starter pack and not production code. But we may have to double-run it in QA too.

ttoomey (approved):
Looks good, thanks!

db/warehouse/migrate/20250523195117_add_hmis_unit_groups.rb
gigxz: yes I do

drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb
gigxz: Thanks. Adding this to the model class instead of the type class

drivers/hmis/app/models/hmis/ce/opportunity.rb
gigxz: Agreed - I will make the change to replace polymorphic owner with unit reference in a follow-up PR

gigxz: Agreed - I will make the change to replace polymorphic owner with unit reference in a follow-up PR

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
gigxz: Huh ok, good to know

drivers/hmis/spec/factories/hmis/unit_groups.rb
gigxz: Yeah good point. I don't think this is necessarily better, its just preferencing project, which we might need to change later.

gigxz:
CI breakage is unrelated, I verified the full HMIS test suite locally, so going ahead and merging

PR #5432, opened 2025-05-29
Patch project viewability granted by project group

gigxz:
## _Merging this PR_

## Description
Follow-up to https://github.com/greenriver/hmis-warehouse/pull/5387 which was incomplete because it didn't support granting access to project viewability via Project Groups. I had mistakenly only tested it for Client access.

QA comment: https://github.com/open-path/Green-River/issues/7353#issuecomment-2917365199

## Type of change
Bug fix 

### DIFF
--- a/drivers/hmis/spec/requests/hmis/ce/client_eligible_ce_opportunities_spec.rb
+++ b/drivers/hmis/spec/requests/hmis/ce/client_eligible_ce_opportunities_spec.rb
@@ -302,7 +302,7 @@ RSpec.describe Hmis::GraphqlController, type: :request do
             expect(response.status).to eq(200), result.inspect
             count = result.dig('data', 'client', 'eligibleCeOpportunities', 'nodesCount')
             expect(count).to eq(32)
-          end.to make_database_queries(count: 15..20)
+          end.to make_database_queries(count: 15..25)
         end
       end
     end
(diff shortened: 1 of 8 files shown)


eanders (approved):
Thanks @gigxz !

gigxz:
Thanks for the review! I pushed up another commit to support checking `Hmis::AccessGroup.contains_with_inherited(project_group)` because that feels comprehensive, but it doesn't actually affect behavior, since we never check visibility of a project group. The behavior of `Hmis::AccessGroup.contains_with_inherited(project)` -- where project is granted via project group-- already worked as expected.

eanders (approved):

drivers/hmis/spec/requests/hmis/ce/client_eligible_ce_opportunities_spec.rb
```diff
@@ -302,7 +302,7 @@
             expect(response.status).to eq(200), result.inspect
             count = result.dig('data', 'client', 'eligibleCeOpportunities', 'nodesCount')
             expect(count).to eq(32)
-          end.to make_database_queries(count: 15..20)
+          end.to make_database_queries(count: 15..25)
```
gigxz, line 305: count went from 20 to 21 with the new access query for the project groups table

PR #5437, opened 2025-05-30
Fix flaky CE tests

martha:
## _Merging this PR_

## Description

There were 2 interrelated bugs leading to flaky test failures in CE world.
1. The Opportunity and Referral factories were creating lots of extraneous data sources.
    - Calling `create(:hmis_ce_referral)` with no arguments would actually create _four_ data sources! (from the opportunity's project association, the opportunity's template association, the _instance's_ template association, and the referred_by user)
    - Fix this PR proposes: Instead of messing around with `after(:create)` hooks, I read the factorybot documentation more closely and found the right way to use transient attributes to ensure consistency. The downside with what I've proposed is that you need to spell the data source out explicitly if you want to pass in (for example) project but not any other attributes. I think this is acceptable, since an invalid record will fail explicitly now thanks to the recently added validations.
2. The actual bug that was leading to flaky failures where 0 records were sometimes returned ([example failure](https://github.com/greenriver/hmis-warehouse/actions/runs/15313603036/job/43083213701)): Here in [BaseController](https://github.com/greenriver/hmis-warehouse/blob/5f7258550e02670b47076d170496a94971f4152f/drivers/hmis/app/controllers/hmis/base_controller.rb#L45), when logging in the user, we find the data source by domain, but all data sources created by the data source factory had [the same domain](https://github.com/greenriver/hmis-warehouse/blob/5f7258550e02670b47076d170496a94971f4152f/drivers/hmis/spec/factories/hmis/data_sources.rb#L10). The result was that the user making the request in the spec might be attached to a different data source than the one where all the fixtures were created. It's flaky because `.find_by` is just returning any record that matches, non-deterministically.
    - Fix this PR proposes: Make the data source factory use a sequence. This does mean that we have to update spec tests that _should_ use the `hmis.dev.test` domain to make sure they do. I created a new factory `authenticated_hmis_data_source` to help with that. (let me know if you have any suggestions about the naming)

How to test: I was able to locally replicate the flaky test failure on the release branch by running the tests repeatedly, such as
```bash
for run in {1..10}; do rspec drivers/hmis/spec/requests/hmis/ce/admin_ce_referrals_spec.rb; done
```

## Type of change
Test failure fix


drivers/hmis/spec/factories/hmis/data_sources.rb
```diff
@@ -4,15 +4,21 @@
...
     sequence(:authoritative, &:zero?)
-    hmis { GraphqlHelpers::HMIS_HOSTNAME }
+    sequence(:hmis) { |n| "#{GraphqlHelpers::HMIS_HOSTNAME}.#{n}" }
     name { 'HMIS' }
     short_name { 'HMIS' }
     # association :client, factory: :hmis_hud_client
     source_type { :sftp }
   end
+
+  factory :authenticated_hmis_data_source, parent: :hmis_data_source do
```
gigxz: Nice find!!

I do want to think about the naming a bit. Nothing about this is really "authenticated" right, I mean it doesn't actually matter what the host name is? (Except for possibly in system tests?)

My understanding is that as long as we generate an HMIS user using `user.related_hmis_user(whatever_data_source)`, and then attach Access Controls to that user, the user is authorized to access that specific data source. Am I missing something about the specific hostname needing to be `GraphqlHelpers::HMIS_HOSTNAME`?

martha: Hm. I think you're right that `authenticated` is not really the right word.

On the other hand, I believe there _is_ something special about the host name "hmis.dev.test" in our (non-system) spec tests, since we hardcode it here in GraphqlHelpers: https://github.com/greenriver/hmis-warehouse/blob/4cbc578c011429cc1867da4fc2273662f57bc2dc/drivers/hmis/spec/support/graphql_helpers.rb#L67-L69

Maybe `graphql_expected_data_source`, or is that too heavy-handed?

gigxz: Oh I see, thank you! It would be helpful to have a comment on this factory stating that it matches the origin used in `post_graphql`, so this data source should be used for graphql request tests.

How about `primary_hmis_data_source` for the naming?

I think we want "find and create" behavior for the primary data source factory, since we really shouldn't have multiple data sources in the db with the same `hmis` value. That is an AR validation that is missing on `GrdaWarehouse::DataSource`, but I suspect adding a validation like that will cause a ton of tests to fail. I opened https://github.com/open-path/Green-River/issues/7775 to add that validation.

This is untested, but we have something like this on the `internal_system` factory, which had a similar need (since there is a uniqueness constraint on name for that model):

```rails
FactoryBot.define do
  factory :hmis_primary_data_source, parent: :hmis_data_source do
    hmis { GraphqlHelpers::HMIS_HOSTNAME } # matches HMIS_ORIGIN

    # if data source exists with this hmis hostname, return it instead of building a new one
    initialize_with do
      GrdaWarehouse::DataSource.find_or_create_by(hmis: hmis)
    end
end
```

I'll put a note about that idea in the linked ticket. What you have here is definitely a step in the right direction and gets the tests passing, so I think we should go ahead and merge! (maybe with the "primary" ds naming)

drivers/hmis/spec/models/hmis/hud/unit_spec.rb
```diff
@@ -109,53 +108,53 @@
...
-    let!(:unit) { create(:hmis_unit, project: project) }
+    let!(:unit) { create(:hmis_unit, project: p1) }
     let!(:today) { Date.current }
     let!(:yesterday) { today - 1.day }
 
     context 'when there are many opportunities' do
-      let!(:opportunity) { create(:hmis_ce_opportunity, owner: unit, project: project, status: :locked, created_at: today - 3.days) }
-      let!(:referral) { create(:hmis_ce_referral, opportunity: opportunity, status: :in_progress, created_at: today - 2.days) }
+      let!(:opportunity) { create(:hmis_ce_opportunity, owner: unit, project: p1, data_source: ds1, status: :locked, created_at: today - 3.days) }
+      let!(:referral) { create(:hmis_ce_referral, opportunity: opportunity, data_source: ds1, status: :in_progress, created_at: today - 2.days) }
```
gigxz, line 152: It feels less than ideal to have to pass everything, but it definitely makes the factory simpler. I wonder if we go with the "find or create" approach for a primary data source, whether we should update the HMIS factories to use the primary data source factory. If we do that, I think this consistency problem would basically go away because all the associated records would be using the same primary data source.

Lets discuss at standup today. I feel like that approach may save us headaches in the long term. We would need to revisit places where we are doing cross-data-source testing. Cross-data-source testing is (and I think will continue to be) comparatively such a small percentage of our overall test coverage, that I would rather trade-off cross-ds testing to be less ergonomic, if we can make single-ds testing simpler.

martha: That makes sense to me. It sounds like I should still go ahead and merge this, considering that work in scope for #7775?

gigxz (approved):
thank you!

drivers/hmis/spec/models/hmis/hud/unit_spec.rb
gigxz: Yes thanks!!

PR #5450, opened 2025-06-03
Outcomes report client list metric bug

dtgreiner:
## _Merging this PR_

## Description

Add a catch in the outcomes report client metrics case statement. This will more gracefully handle an unknown or empty metric being passed.

This was reaching an error due to the `rows` variable not being set to an array. Now it will log the issue and show the page with no clients found.

This appears to only be occuring when manipulating the url to change the metric.

## Type of change
Bug fix


app/models/warehouse_report/outcomes/base.rb
```diff
@@ -1096,6 +1096,9 @@ def support_for(metric, params = nil) # rubocop:disable Metrics/AbcSize
           row[:race],
         ]
       end
+    else
+      rows = []
+      Rails.logger.warn("Unknown metric passed to support_for: #{metric}")
```
eanders: Do you think this should go to sentry instead of the log maybe?

eanders (approved):

app/models/warehouse_report/outcomes/base.rb
dtgreiner: That makes sense to me, the rows being set would allow a more graceful resolution on the client side but we'd still be notified. I'll get that included before merging it in.

PR #5452, opened 2025-06-03
Validate form enable_when and autofill_when picklist conditions

martha:
## _Merging this PR_

## Description
Summary of changes: 
- Add new validations to DefinitionValidator requiring that, when a form item has a condition (`enable_when` or `autofill_when`), and that condition refers to another question with picklist options, the conditional rules refer to real picklist options.
- Add spec

## Type of change
Bug fix

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see related issue comment


drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +172,53 @@ def check_mutually_exclusive_attributes(document)
     link_check.call(document)
   end
 
+  def check_condition(condition, item_hash, link_id)
```
martha, line 185: I added a separate method `check_condition`, instead of adding this in `check_references` as suggested by the ticket, because I noticed a related todo in `check_mutually_exclusive_attributes`. I thought the additional method `check_condition` gives us a future framework for where to put all validations on conditional elements, instead of overloading `check_references`.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +172,70 @@ def check_mutually_exclusive_attributes(document)
...
+
+    return unless referenced_question['pick_list_options'] || referenced_question['pick_list_reference']
+
+    if referenced_question['pick_list_reference']
+      begin
+        options = Types::Forms::PickListOption.
+          options_for_type(referenced_question['pick_list_reference'], user: nil).
+          map(&:deep_stringify_keys)
+      rescue ArgumentError
+        # TODO - discuss, likely not worth it?
```
martha: Let's discuss the picklist_reference strategy. As mentioned in the ticket, we may decide to descope it for now.

This seems to work for picklists that are returned by the Graphql query `GetPickListType`, although I have some concerns that we're using that in a way it wasn't intended, and maybe some picklists break since we are not passing in a `user`.

Meanwhile, we would need to come up with an alternate solution for the case (currently ignored) where the picklist is based on an enum that the frontend resolves, never using `GetPickListQuery`.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -25,6 +25,12 @@ def perform(document, role = nil, skip_cded_validation: false)
     check_references(document, all_ids)
     # Check mutually exclusive attributes ("one of" on conditional objects)
     check_mutually_exclusive_attributes(document)
+
+    # Check conditions like enable_when and autofill_when.
+    # First initialize a FormDefinition to use the logic for generating an item hash, but don't persist it.
+    item_hash = Hmis::Form::Definition.new(definition: document).link_id_item_hash
```
gigxz, line 34: `link_id_item_hash` excludes Group items, which we do need to validate here since they can have conditional visibility

drivers/hmis/app/models/hmis/form/definition_validator.rb
gigxz: Ok, totally fine to descope this, if you can add comments explaining why that would be great.

There are 2 ways "pick list reference" can be set/resolved:
1. `pickListReference` may reference one of `Types::Forms::Enums::PickListType.values.keys` which are resolved on the backend. It's not really possible to perfectly validate these because many of them require context (project id, user, enrollment id, etc) to get the pick list options.
2. `pickListReference` may reference any GraphQL Enum, which are resolved against the code-generated `HmisEnums` class on the frontend. It would be possible to validate these, since we have access to the graphql enums and their values.


For # 2, here's an example of how it's possible to resolve the enum values list based on the listed graphql name. (This is super rough, just made some local changes to the existing `allowed_pick_list_references` fn and ran it locally in the rails console).

```diff
      enums = []
      collect_enums = ->(parent) {
        parent.constants.each do |name|
          child = parent.const_get(name)
          if child.is_a? Class
+            # 'name' gives class name, 'graphql_name' gives string referenced by form item
+            graphql_name = child.graphql_name
+            class_name = child.name.constantize
+            enums << [graphql_name, child.name.constantize] if child < Types::BaseEnum
          elsif child.is_a? Module
            collect_enums.call(child)
          end
        end
      }

      # Include all enums in Types::HmisSchema::Enums namespace
      collect_enums.call(Types::HmisSchema::Enums)
      # Include all enums in Types::Forms::Enums namespace
      collect_enums.call(Types::Forms::Enums)
      # Include all pick list types
      # enums += Types::Forms::Enums::PickListType.values.keys
+      graphql_name_to_enum_class = enums.to_h
+      lookup = graphql_name_to_enum_class.map { |str, klass| [str, klass.values.keys] }.to_h # <-- can validate against these
```

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: Hm, yes. But we are using the `link_id_item_hash` to find the referenced question, which we only validate against if it has a `pick_list_options` or `pick_list_reference`. Group items would not have these keys anyway, so I think it is ok for them to be missing from the hash.

If we wanted to add more validation, like on EXISTS conditions, where the referenced question can also be a group, then this wouldn't work.

I guess the alternative is to make new logic locally for a link_id_item_hash that would contain all group items. But my concern is that that would balloon in terms of space complexity. Maybe these forms are all small enough that it would not be a big deal.

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: Thanks for the hints Gig! This is interesting, it partially works, but there is some magic going on with a few specific enum types `ProjectType` and `ProjectCompletionStatus` that I haven't been able to crack. I can keep working on it a bit tomorrow. Time spent so far about 5h (which was the original estimate), so I'll timebox to another hour tomorrow.

martha: Ok! Here's what I found after a little more investigation.
- `ProjectType` is a special case because there are _two_ enum definitions, `Types::HmisSchema::Enums::Hud::ProjectType` and `Types::HmisSchema::Enums::ProjectType`, and their values look different.

<Details>

```
irb(main):002> Types::HmisSchema::Enums::Hud::ProjectType.values.keys
=>
["EMERGENCY_SHELTER_ENTRY_EXIT",
 "EMERGENCY_SHELTER_NIGHT_BY_NIGHT",
 "TRANSITIONAL_HOUSING",
 "PH_PERMANENT_SUPPORTIVE_HOUSING",
 "STREET_OUTREACH",
 "SERVICES_ONLY",
 "SAFE_HAVEN",
 "PH_HOUSING_ONLY",
 "OTHER",
 "PH_HOUSING_WITH_SERVICES",
 "DAY_SHELTER",
 "HOMELESSNESS_PREVENTION",
 "PH_RAPID_RE_HOUSING",
 "COORDINATED_ENTRY",
 "INVALID"]
irb(main):003> Types::HmisSchema::Enums::ProjectType.values.keys
=>
["ES_ENTRY_EXIT",
 "ES_NBN",
 "TH",
 "PH_PSH",
 "SO",
 "SSO",
 "SH",
 "PH_PH",
 "OTHER",
 "PH_OPH",
 "DAY_SHELTER",
 "HP",
 "PH_RRH",
 "CE",
 "INVALID"]
```

</Details>

I've added a hard-coded special case here (`EXCLUDE`) to deal with this.

- `ProjectCompletionStatus` is not a special case, it was actually a bug, as far as I can tell. I've updated the `R17_1_conditional` logic in the form json to point at the correct answer, so the conditional question appears correctly.

martha: @gigxz I think this may be okay to merge, but I'll hold off until you are back. Feel free to merge while I am out, or push more changes on this branch before merging.

martha:
How to test:
- Visit a form with a choice item that has dependent questions
- Try to change the text of the choice item, and save the form
- You see an error, and can't save the form, because the dependency is no longer valid.

Screenshot 2025-06-20 at 11 18 17 AM

Problems to discuss:
1. The user facing validation message I've created here is not friendly or understandable. Maybe we can brainstorm together what a more helpful message would be. "Cannot change the text of choice "XYZ" on question "A", because question "B" depends on it" or something like that.
2. The ergonomics of making an update like this become annoying, because you need to revert/cancel your change to this form item, go in and remove the dependency on the dependent item, come back to this item and update the text, and finally re-add the dependency. Maybe this is ok as a stopgap until we implement #7863 but it's definitely not ideal.

@gigxz

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +173,67 @@ def check_mutually_exclusive_attributes(document)
...
+      return if valid_answer_codes.nil?
+    else
+      valid_answer_codes = referenced_question['pick_list_options'].map { |opt| opt['code'].to_s }
+      valid_answer_group_codes = referenced_question['pick_list_options'].map { |opt| opt['group_code'].to_s }.uniq
+    end
+
+    answer_codes = condition.values_at('answer_code', 'answer_codes').flatten.compact.uniq
+    if answer_codes.any?
+      (answer_codes - valid_answer_codes).each do |code|
+        add_issue("Invalid answer code: #{code} in 'enable_when' prop of #{link_id}")
```
gigxz: I agree the messaging is not very easy to reason about. How about this:
```ruby
add_issue("Item '#{link_id}' has a dependency on question '#{referenced_question['link_id']}', but the dependent answer option '#{code}' is no longer a valid response choice for that question. Please update the dependency and try again.") unless valid_answer_codes.include?(code)
```

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +173,67 @@ def check_mutually_exclusive_attributes(document)
...
+  def check_condition(condition, item_hash, link_id)
+    return unless condition.key?('question') # Not currently validating conditions with local constants; see below
+
+    # Find the referenced question.
+    # item_hash does not contain groups, and the referenced question could be a group (with an EXISTS condition, for instance).
+    # But for now we are only validating against pick_list_options and pick_list_reference (which groups won't have).
+    # Therefore, if item_hash doesn't contain the referenced question, just return without raising an error.
+    referenced_question = item_hash[condition['question']]
+    return if referenced_question.nil?
+    return unless referenced_question['pick_list_options'] || referenced_question['pick_list_reference']
```
gigxz: I think this could be cleaned up a bit to be more declarative with respect to the shape we're expecting. I think its easier to reason about if we state explicitly that we're looking to only validate conditions that are comparing `code`s against Choice items:

```suggestion
    # Only validate conditions that evaluate against another question
    return unless condition.key?('question')
    # Only validate conditions that compare to the other question's answer codes (as opposed to comparing to numeric value, whether item exists, booleans, etc).
    return unless condition.values_at('answer_code', 'answer_codes', 'answer_group_condes').flatten.compact.uniq.any?

    # Find the referenced question. Since this condition evaluates against the compare questions answer code, we expect the question to be a CHOICE item with pick lists
    referenced_question = item_hash[condition['question']]
    raise 'unexpected' unless referenced_question # add_issue? not sure if checked elsewhere or in schema
    raise 'unexpected' unless referenced_question['type'] == 'CHOICE' || referenced_question['type'] == 'OPEN_CHOICE' # add_issue? not sure if checked elsewhere or in schema
    raise 'unexpected' unless referenced_question['pick_list_options'] || referenced_question['pick_list_reference'] # add_issue? not sure if checked elsewhere or in schema

```

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +173,67 @@ def check_mutually_exclusive_attributes(document)
...
+    if answer_codes.any?
+      (answer_codes - valid_answer_codes).each do |code|
+        add_issue("Invalid answer code: #{code} in 'enable_when' prop of #{link_id}")
+      end
+    end
+
+    if condition.key?('answer_group_code') && valid_answer_group_codes # rubocop:disable Style/GuardClause
+      code = condition['answer_group_code']
+      add_issue("Invalid answer group code: #{code} in 'enable_when' prop of #{link_id}") unless valid_answer_group_codes.include?(code)
+    end
```
gigxz: Could we do this instead? i'm not really understanding why we have the  `&& valid_answer_group_codes` check, because I think if the condition references an `answer_group_code`, and the referenced question DOES NOT hae any answer group codes, then we _should_ present an error.

```suggestion
    elsif condition.key?('answer_group_code')
      # This condition is checking against a group code, so we need to validate that the group code is still valid for the referenced question.
      # Use safe accessor on valid_answer_group_codes because this should also display a validation error if the referenced question has no valid_answer_group_codes
      group_code = condition['answer_group_code']
      add_issue("Item '#{link_id}' has a dependency on question '#{referenced_question['link_id']}', but the dependent answer group '#{group_code}' is no longer a valid choice group for that question. Please update the dependency and try again.") unless valid_answer_group_codes&.include?(group_code)
    end
```

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +173,67 @@ def check_mutually_exclusive_attributes(document)
...
+
+    # TODO: Additional validations. We attempt to ensure this validity in the form property editor,
+    # but we do not validate it here. For example:
+    # - if source is a question, the answer field should be compatible with the question type (eg shouldn't compare STRING=DATE)
+    # - if source is a local constant, the answer field should be compatible local constant type (eg shouldn't compare STRING=DATE)
+    # - if operator is special boolean operator (EXISTS/ENABLED), then the answer type should always be boolean
+    # - certain comparison operators should only be used for certain question types (eg can't use LESS_THAN on a STRING type)
+  end
+
+  def check_conditions(document, item_hash)
```
gigxz, line 232:
```suggestion
  # Check that all EnableWhen and AutofillWhen conditions that reference specific Choice option codes are valid.
  def check_conditions(document, item_hash)
```

Helpful to have a comment to differentiate from `check_references` method which  just checks that the link_id references are valid. Maybe adding a header comment to that method as well

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -172,6 +173,67 @@ def check_mutually_exclusive_attributes(document)
     link_check.call(document)
   end
```
gigxz, line 178:
```suggestion

  # Validate an EnableWhen condition defined on a form item, ensuring that the referenced question
  # and dependent answer codes or group codes are valid.
  #
  # @param [Hash] condition The condition object to validate (expected to match Types::Forms::EnableWhen shape)
  # @param [Hash] item_hash A hash mapping link IDs to their corresponding form items
  # @param [String] link_id The link ID of the item to validate
```

gigxz:
> Problems to discuss:
> 1. The user facing validation message I've created here is not friendly or understandable. Maybe we can brainstorm together what a more helpful message would be. "Cannot change the text of choice "XYZ" on question "A", because question "B" depends on it" or something like that.
> 2. The ergonomics of making an update like this become annoying, because you need to revert/cancel your change to this form item, go in and remove the dependency on the dependent item, come back to this item and update the text, and finally re-add the dependency. Maybe this is ok as a stopgap until we implement #7863 but it's definitely not ideal.


1) Agreed, I suggested a language change in line with what you just described.
2) Agreed, I think its OK as a stopgap

drivers/hmis/app/models/hmis/form/definition_validator.rb
martha: Sounds good, but I think not all of the "unexpected" cases here are actually unexpected. For example, an enable when condition can reference a question of type `STRING`, and it would use `answer_code` in that case.

In other words, this is a valid form definition:
```json
{
  "text": "Reference question",
  "type": "STRING",
  "link_id": "reference_q",
},
{
  "text": "Depender",
  "type": "STRING",
  "link_id": "depender",
  "enable_when": [
    {
      "operator": "EQUAL",
      "question": "reference_q",
      "answer_code": "One"
    }
  ],
}
```

gigxz (approved):
👍

PR #5455, opened 2025-06-04
Epic importer fix

eanders:
## _Merging this PR_

## Description
* Coalesces names to prevent import failures
* Unrelated linting cleanup

## Type of change
* Bug fix
*  Code clean-up

### DIFF
--- a/app/models/health/patient.rb
+++ b/app/models/health/patient.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 # Risk: Describes a patient and contains PHI
 # Control: PHI attributes documented
 module Health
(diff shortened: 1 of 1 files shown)


app/models/health/patient.rb
```diff
@@ -1064,6 +1064,9 @@ def import_epic_team_members
           member = klass.where(at[:email].lower.eq(epic_member.email.downcase).to_sql).
             where(patient_id: id).
             first_or_initialize
+          # If we have only one name, duplicate it for both first and last
+          first_name ||= last_name
+          last_name ||= first_name
```
dtgreiner, line 1069: Do we want to add a catch in case both first and last names are missing? It looks like we'd still have the email as a pseudo identifier.

dtgreiner (approved):
One question/comment, but looks good to move forward as is.

app/models/health/patient.rb
eanders: Good question.  The reason we're doing this is we recently started seeing situations where there was only one name (but an email) and the DB has `null: false` on both name fields causing the importer to fail.  
We may want a placeholder, but at least for now each record includes at least one or the other.

PR #5467, opened 2025-06-06
Add backend fields to support bulk unit actions

martha:
## _Merging this PR_

## Description

Issue: https://github.com/open-path/Green-River/issues/7773
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1172

Summary of changes:
- Add 3 new schema fields to units, similar to the recently added `deletable` helper, to help us consolidate logical complexity in the backend.
  - `canBeMarkedAvailable`, with a todo for the future "available on date" outstanding ticket
  - `canBeMarkedAvailableToday` - my idea was that only these units can be bulk-marked available.
  - `canBeMarkedUnavailable`
- Slightly change the logic for `deletable`, see inline comment

## Type of change
New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - See testing instructions on corresponding frontend PR


drivers/hmis/app/graphql/types/hmis_schema/unit.rb
```diff
@@ -64,7 +67,32 @@ def occupancy_status
 
     def deletable
       return false if occupants.any? # cannot delete unit with occupants
-      return false if load_ar_association(object, :active_referral).present? # cannot delete unit with active referral
+      return false if load_ar_association(object, :referrals).any? # cannot delete unit with any referrals (past or present)
```
martha, line 70: We can discuss this further (and maybe split out a separate ticket?) if we want users to be able to delete units that have past referrals.

Right now it fails because of `dependent: :restrict_with_exception` [here](https://github.com/greenriver/hmis-warehouse/blob/1bc5961608adfce4f2dbe75a3e1344e216fcde99/drivers/hmis/app/models/hmis/ce/opportunity.rb#L17)

gigxz: Thank you! I noticed this. This seems like a good change to prevent the frontend errors, but I agree its too restrictive. I opened https://github.com/open-path/Green-River/issues/7819 to revisit.

gigxz (approved):
looks good!!

PR #5470, opened 2025-06-08
7699 track task runs

ttoomey:
## _Merging this PR_

### execution tracking
* GrdaWarehouse::Tasks::SystemMaintenanceTask and SystemMaintenanceTaskRun have been added to record metadata for each task and log execution start and end times
* Instrumented several recurring jobs with MaintenanceTaskInstrumentation
* Failure Detection & Alerting: detect if a task has not completed within a time threshold and send an alert to Sentry
* Lifecycle Management: prunes old run records
* Admin interface to view task status

### other changes
* improve synchronization code in daily import job
* dry up rescue code in rake task
* adjust document export pruning to raise on failure to destroy records
* add smoke-test specs for many individual jobs and tasks

## Type of change
New feature, Code clean-up


ttoomey:
moving this back to draft to make a few improvements

ttoomey:
@eanders does this approach seem reasonable?

Further work:
* add MaintenanceTaskInstrumentation into all the remaining recurring tasks (hourly in lib/tasks/grda_warehouse.rake). Want to confirm the pattern seems reasonable first

Also:
* perhaps provide an admin interface to view the tasks runs?
* right now run_daily_imports_job creates a number of tasks, but maybe it's simpler if there's just one task for the whole job
* SourceEnrollments  and DaysHomeless.queue_batch probably needs adjustment to allow this type of tracking but that seems out of scope

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -197,6 +242,7 @@ def generate_logging_info
       # take snapshots of client enrollments
       GrdaWarehouse::EnrollmentChangeHistory.generate_for_date!
 
+      # FIXME - refactor the 'queue_batch' pattern to support task monitoring
```
eanders: The queuing is probably not important to track, it should only take a few seconds.  Might be worth instrumenting the processing.

app/models/concerns/maintenance_task_instrumentation.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+module MaintenanceTaskInstrumentation
+  extend ActiveSupport::Concern
```
eanders, line 11: Does this make sense to keep as a concern?  What if it were just a singleton?  Something where we could just call:

```
MaintenanceTaskInstrumentation.instrument_as_maintenance_task(...) { ... }
```

This doesn't need access to the calling class right?

app/models/concerns/maintenance_task_instrumentation.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  protected
+
+  def instrument_as_maintenance_task(job:, name:, alert_threshold_minutes: (60 * 36), &block)
```
eanders: Should we move `(60 * 36)` into a constant(`DEFAULT_ALERT_THRESHOLD_MINUTES`)?  I think that's 36 hours, but I'm not seeing the conversion between minutes and seconds anywhere.

spec/models/grda_warehouse/tasks/system_maintenance_task_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+      before do
+        create(:system_maintenance_task_run,
+               system_maintenance_task: task,
+               started_at: 30.minutes.ago,
+               completed_at: nil)
+      end
+
+      it 'returns true' do
+        expect(task.threshold_exceeded?).to be true
+      end
```
eanders, line 57: If this is only 30 minutes old, shouldn't it not trigger?  (Threshold is 60 right?)

spec/models/grda_warehouse/tasks/system_maintenance_task_spec.rb
```diff
@@ -0,0 +1,116 @@
...
+    context 'when the most recent completed run is outside threshold' do
+      before do
+        create(:system_maintenance_task_run,
+               system_maintenance_task: task,
+               started_at: 90.minutes.ago,
+               completed_at: 85.minutes.ago)
+      end
+
+      it 'returns true' do
+        expect(task.threshold_exceeded?).to be true
```
eanders, line 43: Maybe I'm reading thing backwards, but isn't this only a 15 minute run (thus within the threshold of 60 minutes?)

app/jobs/importing/run_daily_imports_job.rb
ttoomey: Sorry, I should have been clearer.  What I meant was that the "chunked batch processing" strategy used here makes it difficult to track since "completion" means all enqueued jobs succeeded; instrumenting that seems a bit tricky to me; I'm wondering if it makes sense to change strategy.

Perhaps we could use a polling/sweep-style job (periodic runs, consuming up to N records on every run). That's easier to track, since we just want to be sure the consumer runs every 60 minutes (or something). Happy to huddle about it

eanders: Ah, yes, that does seem like it would be really nice.  Let's discuss at some point.  Also fine to leave that off for this round.

spec/models/grda_warehouse/tasks/system_maintenance_task_spec.rb
ttoomey: It sounds like I may not have conveyed the intent of the alert in the code; sorry about that! You're expecting alerts to be job run-time (15 min) but that's not currently implemented; that would be a nice enhancement though.

Right now alerts are only for missing tasks. So 60 minute threshold here means that if the job hasn't run in the last 60 minutes, we should send an alert. IIRC this originated when the process was killed by the OS due to OOM issues, which we don't currently have visibility into. So this feature is only aimed at surfacing when maintenance tasks haven't occurred in the expected window.

In this test, the runtime doesn't matter, it's the completed_at time (85 min ago) which is greater than the threshold (60 min).

Let me know if this doesn't match your understanding. I'll take a pass through and clarify the code or var names to better convey this.

spec/models/grda_warehouse/tasks/system_maintenance_task_spec.rb
ttoomey: we only look at completion time for the purposes of alerts (see above comment). While we track runtime, we aren't doing anything with it in this PR yet

app/models/concerns/maintenance_task_instrumentation.rb
ttoomey: sure, we can extract to a constant. FYI the minutes conversion is in 
```ruby
    # SystemMaintenanceTask#threshold_exceeded?
    threshold = now - alert_threshold_minutes.minutes
    system_maintenance_task_runs.where(completed_at: threshold...).none?
```

eanders: Thanks!  Missed the `.minutes`

app/models/concerns/maintenance_task_instrumentation.rb
ttoomey: Sure, we can use a singleton. I had originally hoped to use around_perform to minimize code churn but that doesn't seem to work; probably due to interaction with the existing lifecycle hooks in BaseJob.

Maybe `instrument_as_maintenance_task` should just be a method on BaseJob if we need this mechanism on many jobs

app/jobs/importing/run_daily_imports_job.rb
ttoomey: tracking here: https://github.com/open-path/Green-River/issues/7815

app/models/concerns/maintenance_task_instrumentation.rb
ttoomey: It proved useful to have a concern but the logic is encapsulated in a singleton. Lmk if this seems too complex

lib/tasks/grda_warehouse.rake
```diff
@@ -288,7 +296,8 @@ namespace :grda_warehouse do
 
   desc 'Monthly tasks'
   task monthly: [:environment, 'log:info_to_stdout'] do
-    GrdaWarehouse::Tasks::CensusImport.new.run!
+    # this runs every day out of RunDailyImportsJob
+    # GrdaWarehouse::Tasks::CensusImport.new.run!
```
ttoomey: doesn't seem like we need this here?

lib/tasks/grda_warehouse.rake
```diff
@@ -299,46 +308,39 @@ namespace :grda_warehouse do
...
-    begin
-      Rake::Task['jobs:check_queue'].invoke
-    rescue StandardError => e
-      puts e.message
-      Sentry.capture_exception(e)
+    safely_execute do
+      MaintenanceTasksLifecycleJob.new.perform
     end
-    begin
-      Rake::Task['grda_warehouse:send_health_emergency_notifications'].invoke
```
ttoomey, line 309: this is probably not used

ttoomey:
@eanders I think this is ready for review. Note I didn't instrument everything running out of `config/schedule.rb` as this PR is getting pretty large already

app/controllers/admin/system_maintenance_tasks_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+module Admin
+  class SystemMaintenanceTasksController < ApplicationController
+    helper ElapsedTimeHelper
+    before_action :require_can_manage_config!
+
+    def index
+      @tasks = scope.all.order(:name)
```
eanders, line 15: Should we paginate this?

eanders: Oh, never mind, I was looking in the wrong file for the pagination.

lib/tasks/grda_warehouse.rake
eanders: Hmm, I think at some point, this monthly one was a full-rebuild.  Maybe we don't need that anymore, or maybe we should update this so that it does a full-rebuild.  Thoughts?

lib/tasks/grda_warehouse.rake
eanders: That is currently not used.  Safe to remove for now.

spec/jobs/concerns/maintenance_task_instrumentation_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+    it 'creates task run without completion time when job fails' do
+      freeze_time do
+        expect do
+          job.perform(should_fail: true)
+        end.to raise_error(StandardError, 'Test failure')
+
+        task = GrdaWarehouse::Tasks::SystemMaintenanceTask.last
+        run = task.system_maintenance_task_runs.last
+
+        expect(run.started_at).to eq(Time.current)
```
eanders, line 69: Will this check be correct if it takes more than a millisecond to run? Seems like this maybe should check for the presence of started_at rather than an exact timestamp.

eanders (approved):
Looks good to me.  
I wish we didn't need quite so much churn on the delayed jobs.  Is this because `around_perform` is only available for `ActiveJob` and `DelayedJob` doesn't have an equivalent?

ttoomey:
> I wish we didn't need quite so much churn on the delayed jobs. Is this because around_perform is only available for ActiveJob and DelayedJob doesn't have an equivalent?

Thanks!
It does feel like a lot of churn. I found a simpler approach with around_perform didn't work because:
a) our lifecycle hooks interfere with it, although I'm sure we could work around that
b) the perform action may early-exit in many cases such as if it doesn't acquire a lock. So the job itself needs to be able to signal that it completed somehow
c) many of the things we want to instrument aren't jobs, such as tasks etc. So around_perform would only work in some of the places

spec/jobs/concerns/maintenance_task_instrumentation_spec.rb
ttoomey: good point, it probably should freeze time

ttoomey: oh we do freeze time. So it should be okay

app/helpers/elapsed_time_helper.rb
```diff
@@ -22,4 +22,23 @@ def elapsed_time(total_seconds, decimal_places: 0)
       format('%ss', s)
     end
   end
+
+  def precise_distance_of_time(seconds)
```
ttoomey, line 28: added as rails time_ago_in_words is quite imprecise for some reason

lib/tasks/grda_warehouse.rake
ttoomey: discussed 6/13 and decided to just remove the monthly call as it processes 3yrs daily

ttoomey:
@eanders I made the changes we discussed, mostly adding specs and instrumenting the fetch_and_import_job and csg_engage/program_report

I think all the jobs/tasks modified should have at least a smoke test now.

Probably could use a another quick pass before merge.

spec/jobs/reporting/population_dashboard_populate_job_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+          expect(veterans_report_double).to have_received(:populate!)
+        end
+      end
+
+      context 'when a specific sub_population is given' do
+        it 'populates the specific report' do
+          described_class.new.perform(sub_population: 'youth')
+          expect(report_instance_double).to have_received(:populate!)
+        end
+      end
```
eanders: Maybe this should have a check to confirm `veterans` didn't receive `populate!` ?

```
expect(youth_report_double).to have_received(:populate!)
expect(veterans_report_double).to_not have_received(:populate!)
```

eanders (approved):
Thanks @ttoomey.  I think just one small tweak to the spec, as noted, and it looks good to me!

PR #5471, opened 2025-06-09
Demo Summary Report - Newly Entering: Youth Details Calculation

dtgreiner:
## _Merging this PR_

## Description

The details page Demographic Summary Report - Newly Entering - Youth was not returning any clients. This was due to an issue with the hash calculation where it was looking for :id instead of :client_id. It looks llike this was an artifact from a previous way of doing the calculations.

## Type of change
Bug fix


drivers/core_demographics_report/app/models/core_demographics_report/first_time_calculations.rb
```diff
@@ -13,18 +13,11 @@
     def no_recent_homelessness_detail_hash
       {}.tap do |hashes|
         available_no_recent_homelessness_types.invert.each do |key, title|
-          # These need to use enrollment.id to capture age correctly
-          id_field = if key.to_sym.in?([:with_children, :with_children, :unaccompanied_youth])
```
dtgreiner, line 17: Note: `:with_children` is not in `the available_no_recent_homelessness_types` hash, and I am not seeing `no_recent_homelessness_detail_hash` being called anywhere it may be overwritten.

eanders: I'm guessing that was supposed to be `adults_with_children` and `without_children` since youth can fall into all three categories.

eanders: But maybe that's irrelevant?

drivers/core_demographics_report/app/models/core_demographics_report/first_time_calculations.rb
```diff
@@ -13,18 +13,11 @@
...
-          else
-            :client_id
-          end
-
           hashes["no_recent_homelessness_#{key}"] = {
             title: "No Recent Homelessness - #{title}",
             headers: client_headers,
             columns: client_columns,
-            scope: -> { report_scope.joins(:client, :enrollment).where(id_field => no_recent_homelessness_client_ids(key)).distinct },
+            scope: -> { report_scope.joins(:client, :enrollment).where(client_id: no_recent_homelessness_client_ids(key)).distinct },
```
eanders, line 23: Just want to check this.  So my understanding is that someone gets tracked as having recent homelessness here:

```
private def set_recently_homeless_client_counts(clients, client_id, enrollment_id, coc_code = base_count_sym)
      # Only count them in one category.
      if !clients[:client][coc_code].include?(client_id)
        clients[:chronic][coc_code] << client_id if chronic_ids.include?(client_id)
        clients[:hoh_chronic][coc_code] << client_id if hoh_chronic_ids.include?(client_id)
        clients[:high_acuity][coc_code] << client_id if high_acuity_ids.include?(client_id)
        clients[:hoh_high_acuity][coc_code] << client_id if hoh_high_acuity_ids.include?(client_id)
        # These need to use enrollment.id to capture age correctly, but needs the client for summary counts
        clients[:adult_and_child][coc_code] << [enrollment_id, client_id] if adult_and_child_ids.include?(enrollment_id)
        clients[:hoh_adult_and_child][coc_code] << [enrollment_id, client_id] if hoh_adult_and_child_ids.include?(enrollment_id)
        clients[:unaccompanied_youth][coc_code] << [enrollment_id, client_id] if unaccompanied_youth_ids.include?(enrollment_id)
      end
      # Always add them to the clients category
      clients[:client][coc_code] << client_id
      clients[:household][coc_code] << client_id if hoh_client_ids.include?(client_id)
    end
```

I believe then we may be looking for `no_recent_homelessness_client_ids(:unaccompanied_youth)`.

which will use the following to look that up:
```
private def no_recent_homelessness_client_ids(key, coc_code = base_count_sym)
      # These are stored as Set[[enrollment_id, client_id]]
      if key.in?([:adult_and_child, :hoh_adult_and_child, :unaccompanied_youth])
        no_recent_homelessness_clients[key][coc_code].to_a.map(&:last).uniq
      else
        # fetch client_ids from [client_id]
        no_recent_homelessness_clients[key][coc_code]
      end
    end
```

Which should _always_ be pulling the `client_id` out of the hash, which I think makes your change correct.  

Can you add some comments to the methods in this file to try and hint our future selves as to how these work?

dtgreiner: Yes, that's what I was finding. I'll get some comments in here, and I believe the other similar files are already working as the updated setup. I'll look at adding comments there too.

dtgreiner: I've added method headers for each of the `*_calculations.rb` files. These were AI generated and manually adjusted. Some were a little vague or redundant with the method name, but figured it was worth having them in for consistency. Happy to adjust them if needed.

drivers/core_demographics_report/app/models/core_demographics_report/age_calculations.rb
```diff
@@ -81,49 +89,64 @@ def age_detail_hash
       end
     end
 
+    # Counts the number of adult clients
+    # @return [Integer] The count of adult clients, masked if population is small
```
eanders:
```suggestion
    # @return [Integer] The count of adult clients, masked if appropriate
```

eanders (approved):
Thanks @dtgreiner, I think the comments look good (spot checked.)
I'm sure some could be more descriptive, but I'm not seeing anything that would be problematic.

PR #5474, opened 2025-06-10
fix tz issue leading to intermittent test failures

ttoomey:
## _Merging this PR_

## Description
fixes what appears to be date-to-time tz related issue in CE search filter date conversion, showing up in test failures 

## Type of change
Bug fix


gigxz (approved):
thanks! looks like there were a couple other failures based on the time when CI ran near midnight - but this change looks good to me!

PR #5479, opened 2025-06-10
CE Project Configuration, revamp global feature flags

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7409
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1176

* Add a `GlobalFeatureFlags` type to the GraphQL API, which replaces the global feature flags that are resolved on the App Settings Controller
* Add a Project Config for enabling Coordinated Entry features on a per-project basis. Disabling the app config property overrides the project-level configurations.

How to test: see instructions on https://github.com/greenriver/hmis-frontend/pull/1176
## Type of change
New Feature


drivers/hmis/app/graphql/mutations/create_project_config.rb
```diff
@@ -18,16 +20,7 @@ def resolve(input:)
...
-      when 'AUTO_ENTER'
-        Hmis::ProjectAutoEnterConfig
-      when 'AUTO_EXIT'
-        Hmis::ProjectAutoExitConfig
-      when 'STAFF_ASSIGNMENT'
-        Hmis::ProjectStaffAssignmentConfig
-      else raise "Unsupported type: #{input.config_type}"
-      end
-
+      class_name = input.config_type.constantize
```
gigxz: This is a breaking change but I think it's OK since the project settings page is only used by a couple people and sparingly

drivers/hmis/app/controllers/hmis/app_settings_controller.rb
```diff
@@ -36,6 +38,7 @@ def show
       revision: Git.revision,
       branch: Git.branch,
       theme: theme&.hmis_value,
+      # deprecated, use globalFeatureFlags on the GraphQL API instead
       globalFeatureFlags: {
         # Whether to show MCI ID in client search results
         mciId: HmisExternalApis::AcHmis::Mci.enabled?,
```
gigxz, line 44: to remove in a future release

drivers/hmis/app/graphql/mutations/create_project_config.rb
ttoomey: While this is concise, it's not obvious to the reader that config_type is an enum who's values are constrained to certain ruby class names.

This should be functional but I'd suggest adding a comment here calling out what's going on (explain we aren't calling `constantize` on a random user-provided string).

If I were nit-pick, I would favor preserving the previous key values and explicit mapping. It feels like it better preserves the boundary between API and implementation details but this is quite subjective and shouldn't block this pr. I imagine centralizing it in the ProjectConfig

```ruby
class Hmis::ProjectConfig
  # ...
  # string keys keep the factory classes internal
  CONFIG_TYPE_FACTORIES = [
    'AUTO_EXIT_CONFIG' => 'Hmis::ProjectAutoExitConfig',
    'AUTO_ENTER_CONFIG' => 'Hmis::ProjectAutoEnterConfig',
    'STAFF_ASSIGNMENT_CONFIG' => 'Hmis::ProjectStaffAssignmentConfig',
    'CE_CONFIG' => 'Hmis::ProjectCeConfig',
  ].freeze

  # Hmis::ProjectConfig.config_factory(input.config_type)
  def self.config_factory(config_type)
    CONFIG_TYPE_FACTORIES.fetch(config_type).constantize.new
  end

```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -291,6 +291,11 @@ def pick_list(pick_list_type:, **args)
 
     field :current_user, Application::User, null: true
 
+    field :global_feature_flags, Types::HmisSchema::GlobalFeatureFlags, null: false
+    def global_feature_flags
+      current_user
```
ttoomey: sorry I'm probably missing something, why are we passing in current user? It looks like the GlobalFeatureFlags resolver doesn't reference `object`

Also, it would be nice to include some tests for this endpoint

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -486,11 +491,15 @@ def form_rule(id:)
       Hmis::Form::Instance.find_by(id: id)
     end
 
-    field :project_configs, Types::HmisSchema::ProjectConfig.page_type, null: false
-    def project_configs
-      raise 'not allowed' unless current_user.can_configure_data_collection?
+    field :project_configs, Types::HmisSchema::ProjectConfig.page_type, null: false do
+      filters_argument Types::HmisSchema::ProjectConfig
+    end
+    def project_configs(filters: nil)
```
ttoomey, line 497: suggest adding some basic tests that also exercise the filters

drivers/hmis/app/models/hmis/project_ce_config.rb
```diff
@@ -0,0 +1,11 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class Hmis::ProjectCeConfig < Hmis::ProjectConfig
+  def config_type = 'COORDINATED_ENTRY'
```
ttoomey, line 10: sorry, I'm not following how this method is used. I assume it was to populate the field on the graphql resolver:
```ruby
    field :config_type, Types::HmisSchema::Enums::ProjectConfigType, null: false, method: :type
```

Although now we are using the STI type for that instead.

ttoomey (approved):
Looks good, thanks! I left some suggestions.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: I was passing that in to resolve an `id` on the global feature flag object, to use as a cache key for apollo. There is probably a better way to achieve this, maybe by specifying on the frontend that this object type doesn't have a cache key? That wasn't immediately obvious so I went with resolving an id. It doesn't really have anything to do with the current user, though- I will look into a better way to handle this.

drivers/hmis/app/graphql/mutations/create_project_config.rb
gigxz: I agree- I will make that change. Thanks!

drivers/hmis/app/models/hmis/project_ce_config.rb
gigxz: I'll revert that change per your other suggestion

PR #5480, opened 2025-06-12
Handle duplicate medicaid ids more appropriately

eanders:
## _Merging this PR_

## Description
* If a client merge is attempted where the clients both have patient records and the medicaid ids don't match, don't fail completely.
* Mark the two clients as distinct, and carry on
* tests 

## Type of change
* Bug fix

### DIFF
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -2257,59 +2257,71 @@ module GrdaWarehouse::Hud
       setup_notifier('PatientMerger') unless @notifier
       moved = []
       to_clean = [id]
-      transaction do
-        # get the existing destination client for other_client
-        prev_destination_client = if other_client.destination_client
-          other_client.destination_client
-        elsif other_client.destination?
-          to_clean << other_client.id
-          other_client
-        end
-        # if it had sources then move those over to us
-        # and say who made the decision and when
-        other_client.source_clients.each do |m|
-          m.warehouse_client_source.update!(
-            destination_id: id,
-            reviewed_at: reviewed_at,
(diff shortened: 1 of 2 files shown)


eanders:
I've run this in production and confirmed the first run logs the failure to merge:
```
I, [2025-06-12T13:29:08.208223 #7471]  INFO -- : Attempting merge of 123 into 234
I, [2025-06-12T13:29:08.677150 #7471]  INFO -- : Non-matching Medicaid IDs on patient merge
```

And subsequent runs no longer see the issue as there is a split history added to prevent re-attempting the match.

ttoomey (approved):
Looks good! 

* should retarget to 166
* we still don't know exactly why the rescue wasn't triggered; could be a side-effect of multi-db queries or advisory locks but we didn't find that yesterday when pairing. I guess we don't know if this will fix the issue in prod

One thing, it looks like this code got moved into the transaction but it probably should be outside, right?
```
      # and invalidate our own service history
        force_full_service_history_rebuild
        # and invalidate any cache for these clients
        self.class.clear_view_cache(prev_destination_client.id) if prev_destination_client.present?

```

ttoomey:
oh sorry I missed that you'd tested in prod already. That's great!

eanders:
For better or worse, it looks like that was in the transaction on 165, so leaving for now.
> 
> One thing, it looks like this code got moved into the transaction but it probably should be outside, right?
> 
> ```
>       # and invalidate our own service history
>         force_full_service_history_rebuild
>         # and invalidate any cache for these clients
>         self.class.clear_view_cache(prev_destination_client.id) if prev_destination_client.present?
> ```

PR #5486, opened 2025-06-13
Add Exclusion criteria to project groups

eanders:
## _Merging this PR_

## Description
Adds:
1. Exclusions of projects
2. Exclusions of project types
3. Updated UI for Project Group editor
4. Tests

## Type of change
* New feature


app/controllers/project_groups_controller.rb
```diff
@@ -32,33 +34,38 @@ def new
...
-        @project_group.maintain_projects!
-        Collection.maintain_system_groups(group: :project_groups)
-
-        # TODO: START_ACL remove when ACL transition complete
-        users = user_params[:users]&.reject(&:empty?)&.map(&:to_i)
-        users << current_user.access_group.id
-        @project_group.update_access(users)
-        AccessGroup.maintain_system_groups(group: :project_groups)
-        # END_ACL
+    begin
```
eanders, line 37: This adjustment is mostly to get the tests happy, the UI seemed ok, but I think might not have been if it had raised.  There are two changes here:
1. move the rescue outside of the transaction
2. call `@project_group.instance_variable_set(:@filter, nil)` to force a recalculation of the filter before calling `@project_group.maintain_projects!` (Maybe that should just be part of the method, but it only seemed to be an issue in the controller/tests.

app/models/filters/filter_base.rb
```diff
@@ -125,42 +128,42 @@ def update(filters) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/Perc
         self.heads_of_household = filter_hoh
         self.hoh_only = filter_hoh
       end
-      self.default_project_type_codes = Array.wrap(filters.dig(:default_project_type_codes))&.reject(&:blank?) if filters.key?(:default_project_type_codes)
+      self.default_project_type_codes = cleanup_input_array(filters.dig(:default_project_type_codes), input_type: :to_s, default: default_project_type_codes) if filters.key?(:default_project_type_codes)
```
eanders: I tried to DRY this up a bit, I think it's easier to read, but would love some confirmation.

eanders:
Looks like I need to fix some other tests, pulling this back to draft for now.

eanders:
Tests are passing now.

app/models/grda_warehouse/project_group.rb
```diff
@@ -140,7 +150,7 @@ def self.options_for_select(user:)
     end
 
     def filter
-      @filter ||= begin
```
ttoomey, line 143: did you mean to remove the memoization here?

app/controllers/project_groups_controller.rb
```diff
@@ -32,33 +34,38 @@ def new
...
+
+          # TODO: START_ACL remove when ACL transition complete
+          users = user_params[:users]&.reject(&:empty?)&.map(&:to_i)
+          users << current_user.access_group.id
+          @project_group.update_access(users)
+          AccessGroup.maintain_system_groups(group: :project_groups)
+          # END_ACL
+        end
       end
     rescue Exception => e
```
ttoomey: Not in the change set here but it seems incorrect that we rescue exception, if we're going to use the save! & rescue pattern I think it should rescue StandardError.

app/controllers/project_groups_controller.rb
```diff
@@ -67,21 +74,23 @@ def edit
...
+          # END_ACL
+          if user_params.key?(:editor_ids)
+            user_ids = user_params[:editor_ids]&.reject(&:empty?)&.map(&:to_i)
+            @project_group.replace_access(User.where(id: user_ids).to_a, scope: :editor) if user_ids.present?
+          end
+          @project_group.maintain_projects!
         end
-        @project_group.maintain_projects!
       end
     rescue Exception => e
```
ttoomey:
```suggestion
    rescue => e # StandardError
```

app/models/grda_warehouse/project_group.rb
```diff
@@ -45,6 +49,12 @@ class ProjectGroup < GrdaWarehouseBase
     has_many :contacts, through: :projects
     has_many :organization_contacts, through: :projects
 
+    after_initialize do
+      # Initialize filter with excluded project ids and types
+      self.excluded_project_ids = filter&.excluded_project_ids || []
+      self.excluded_project_type_numbers = filter&.excluded_project_type_numbers || []
+    end
```
ttoomey: this works. Probably better to use discrete getters
```
   def excluded_project_ids
     filter.excluded_project_ids || []
   end
   def excluded_project_type_numbers
      filter.excluded_project_type_numbers || []
   end
```

app/controllers/project_groups_controller.rb
```diff
@@ -32,33 +34,38 @@ def new
...
-        AccessGroup.maintain_system_groups(group: :project_groups)
-        # END_ACL
+    begin
+      @project_group = project_group_source.new
+      project_group_source.transaction do
+        @project_group.assign_attributes(name: group_params[:name])
+        filter = ::Filters::HudFilterBase.new(user_id: current_user.id, project_type_numbers: []).update(filter_params.merge(coc_codes: []))
+        filter.coc_codes = []
+        @project_group.options = filter.to_h
+        if @project_group.save!
```
ttoomey: this condition doesn't really make sense. It either succeeds or it raises an exception. Note in the update action doesn't use the bang variant.

app/controllers/project_groups_controller.rb
```diff
@@ -32,33 +34,38 @@ def new
...
+          editors = user_params[:editor_ids]&.reject(&:blank?)&.map(&:to_i)
+          # If the user can't edit all project groups, make sure we add the user so they can access it later
+          editors << current_user.id
+
+          @project_group.replace_access(User.find(editors), scope: :editor)
+          @project_group.maintain_projects!
+          Collection.maintain_system_groups(group: :project_groups)
```
ttoomey: it looks like maintain_system_groups is called in an after_create already. But we need to call it again here?

app/models/filters/filter_base.rb
ttoomey: Thanks! I like this direction. Suggestions
* I don't see a reason to use dig(), seems better to use [] since we only go one level deep
* I think that passing in a default arg seems extra; the conditional assignment seemed better to me
* rather than input_type, suggest passing in a proc as that's more flexible and concise
* naming: I might call this 'normalize_input_array` but cleanup is fine too
```ruby
self.default_project_type_codes = normalize_input_array(filters[:default_project_type_codes], &:to_s) || default_project_type_codes if filters.key?(:default_project_type_codes)

private def normalize_input_array(dirty, &normalizer)
  normalizer ||= proc(&:to_i)  # default transformation
  # maybe sort.uniq too?
  Array.wrap(dirty).compact_blank.map(&normalizer).presence
end
```

app/models/filters/filter_base.rb
```diff
@@ -573,10 +589,25 @@ def effective_project_ids
...
 
-        project_ids.uniq.reject(&:blank?)
+        reject_excluded_project_ids(project_ids.uniq.reject(&:blank?))
       end
     end
 
+    def reject_excluded_project_ids(project_ids)
+      return project_ids if excluded_project_ids.empty? && excluded_project_type_numbers.empty?
+
+      project_ids.reject! { |id| excluded_project_ids.include?(id) }
```
ttoomey: probably better not to mutate our inputs, could cause weird bugs
```suggestion
      project_ids = project_ids.reject { |id| excluded_project_ids.include?(id) }
```

ttoomey (commented):
Thanks! Left a few notes that probably should get addressed

I think I need a better understanding of project groups, I'm surprised by the amount of complexity around this model

app/models/grda_warehouse/project_group.rb
eanders: Yes, the memorization was breaking tests at least, possibly the functionality as well.

app/models/filters/filter_base.rb
```diff
@@ -173,7 +176,7 @@ def update(filters) # rubocop:disable Metrics/CyclomaticComplexity, Metrics/Perc
       vispdat_limit = filters.dig(:limit_to_vispdat)&.to_sym
       self.limit_to_vispdat = vispdat_limit if vispdat_limit.present? && available_vispdat_limits.values.include?(vispdat_limit)
       self.ph = filters.dig(:ph).in?(['1', 'true', true]) unless filters.dig(:ph).nil?
-      self.times_homeless_in_last_three_years = filters.dig(:times_homeless_in_last_three_years)&.reject(&:blank?)&.map(&:to_i) unless filters.dig(:times_homeless_in_last_three_years).nil?
+      self.times_homeless_in_last_three_years = normalize_input_array(filters[:times_homeless_in_last_three_years]) unless filters
```
ttoomey: probably?
```suggestion
   self.times_homeless_in_last_three_years = normalize_input_array(filters[:times_homeless_in_last_three_years]) unless filters.dig(:times_homeless_in_last_three_years).nil?
```

app/models/grda_warehouse/project_group.rb
ttoomey: got it, maybe just add a comment that it's intentionally not memoized

ttoomey (approved):
Looks great! Looks like just one small correction needed to a condition in filter base

app/models/filters/filter_base.rb
eanders: Good catch, thanks!

PR #5492, opened 2025-06-16
Superset User Env

yakloinsteak:
- Fetch superset params for warehouse

## Description
Pulls in two superset environment variables to allow usage of superset API.

## Testing Notes
I copied key files to a running demo container and saw the two values added to the environment. It does not fail if the values aren't present. Testing this on QA isn't a valid test since EKS pods get their secrets via a different mechanism.

## Type of change

Environment variables update

## Testing/Todo
- [x] Update roles to allow decryption of these values
- [x] Update roles to allow fetching the parameters
- [x] Try on demo via deployment. That's probably the safest to try since it has to be production.
- [x] See actual values in .env file


eanders (approved):
Thanks @yakloinsteak, looks good at first glance.

PR #5497, opened 2025-06-17
add sentry monitoring to WH front-end

dtgreiner:
## _Merging this PR_

## Description

Adds Sentry error monitoring to capture JavaScript errors in the front-end. It's currently set to use the same Sentry project as our backend, so any JS errors will populate there. It's also setup so that it should be automatically capturing un-caught errors.

## Type of change
New feature


app/javascript/application_esbuild.js
```diff
@@ -1,2 +1,8 @@
-// entry new point
+// Import Sentry for error tracking
```
eanders, line 1: This is great that you're adding this here.  
I'm being forced to embrace this pattern with the bootstrap upgrade as well, and I think it will serve us much better long-term.

app/javascript/sentry.js
```diff
@@ -0,0 +1,53 @@
...
+const hostname = window.location.hostname;
+
+// The base trace rate for performance monitoring.
+// This should be provided during the build process. It can be set to 0 to disable tracing.
+// It defaults to 1.0 in staging and 0.0 in other environments if not set.
+const getBaseTraceRate = () => {
+  if (config.traceRate) {
+    return parseFloat(config.traceRate);
+  }
+  if (config.environment === 'staging') {
```
eanders: Do we want a default for `production`?  or is that something we plan to add later?  Maybe we could add it but comment it out for now so we can know where we're going?

eanders (approved):

app/javascript/sentry.js
dtgreiner: This was based off of how the sentry.rb works which only defaults the 'qa_hmis' staging. With that said, I don't see any issues in defaulting production to 1.0 as well.

app/views/layouts/application.html.haml
```diff
@@ -95,3 +95,10 @@
     = yield :page_js
     = render "inactive_session_modal"
     = svg 'symbol-defs'
+
+    :javascript
+      window.sentryConfig = {
+        dsn: "#{ENV['WAREHOUSE_SENTRY_DSN']}",
+        environment: "#{Rails.env}",
+        traceRate: "#{ENV['SENTRY_PERFORMANCE_TRACE_RATE']}"
+      };
```
ttoomey: * It's safer to use to_json rather than string interpolation. Suggest putting the config in application controller to make it easier to override.
* using WAREHOUSE_SENTRY_DSN here will send all the FE sentry alerts into the same project as the warehouse. That might be good as it's easier to report on related errors; but it also might be very noisy initially.
* we should also include the id or email of the current user and if the user was impersonating someone
* suggest we also include the current git revision
* 

```ruby
  # ApplicationController
  def sentry_frontend_config
      {
        dsn: ENV['WAREHOUSE_SENTRY_DSN']'
        environment: Rails.env,
        traceRate: ENV['SENTRY_PERFORMANCE_TRACE_RATE'],
      }
  end
  helper_method :sentry_frontend_config
```

```suggestion
    :javascript
      window.sentryConfig = #{sentry_frontend_config.to_json};
```

app/javascript/sentry.js
```diff
@@ -0,0 +1,53 @@
...
+};
+
+if (config.dsn) {
+  Sentry.init({
+    dsn: config.dsn,
+    environment: config.environment || 'development',
+    initialScope: {
+      tags: { hostname },
+    },
+    tracesSampler: (samplingContext) => {
```
ttoomey, line 27: IIRC this is only for performance traces. This probably isn't something we'd need initially

app/javascript/sentry.js
```diff
@@ -0,0 +1,53 @@
...
+  if (config.environment === 'production' || config.environment === 'staging') {
+    return 1.0;
+  }
+  return 0.0;
+};
+
+if (config.dsn) {
+  Sentry.init({
+    dsn: config.dsn,
+    environment: config.environment || 'development',
```
ttoomey, line 22: suggest adding `release: config.gitRevision,`

app/javascript/sentry.js
```diff
@@ -0,0 +1,53 @@
...
+      }
+      return traceWeight * baseTraceRate;
+    },
+    // Add integrations here if needed
+    allowUrls: [hostname],
+    ignoreErrors: [],
+    beforeSend(event) {
+      return event;
+    },
+  });
```
ttoomey, line 50: not sure if this is exactly the right config but setting a user is something like:
```js
// var userConfig = {id: user.id, email: user.email}
    Sentry.configureScope(
      function(scope) {
        if (userConfig) {
          scope.setUser(userConfig);
        }
        scope.setExtra('true_user_id', config.trueUserId);
      }
    );

```

ttoomey (commented):
Looks good! Left a few suggestions

app/javascript/sentry.js
dtgreiner: This has been added [here](https://github.com/greenriver/hmis-warehouse/pull/5497/commits/025b7f9dd39a60a822209f4dbe3ee33bf02170a8), as well as the user & true user values

app/views/layouts/application.html.haml
dtgreiner: The config has been moved to the application controller in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5497/commits/025b7f9dd39a60a822209f4dbe3ee33bf02170a8)

app/javascript/sentry.js
dtgreiner: updated [here](https://github.com/greenriver/hmis-warehouse/pull/5497/commits/025b7f9dd39a60a822209f4dbe3ee33bf02170a8) with a couple adjustments

ttoomey (approved):
looks great, thank you!

PR #5498, opened 2025-06-17
Add events to referral

martha:
## _Merging this PR_

## Description
Summary of changes
- Add events to referral schema object
- Modify engine code to record audit events when a referral is completed.
  - It was previously only logging an event of type "message_sent" on referral completion, so I updated it to make it symmetrical with the referral start events. 
  - It will now log duplicate events for both starting and ending a referral, since it always logs an event on "message_sent," but we filter those out of the timeline.
- Related but somewhat out of scope, this PR proposes a cleanup to CE spec tests.
  - Previously, the default referral provided by `ce_spec_helper` only had one possible ending, "accept workflow." I've updated it to include a "reject workflow" possibility since this was needed by multiple tests, including the newly added `ce_referral_events_spec.rb`.
  - With this change, a few unrelated spec files also needed minor updates

https://github.com/greenriver/hmis-frontend/pull/1182

## Type of change
New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see frontend PR

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
@@ -23,7 +23,7 @@ module Types
     field :target_enrollment, Types::HmisSchema::Enrollment, null: true
     field :swimlanes, [HmisSchema::CeReferralSwimlane], null: false
     field :workflow_template_name, String, null: true
-    field :events, HmisSchema::CeReferralEvent.page_type, null: false, method: :timeline_events
+    field :audit_events, HmisSchema::CeReferralAuditEvent.page_type, null: false
 
     # Resolve project fields separately, instead of the whole project object, in case user can't view the project
     field :target_project_id, ID, null: false
@@ -165,6 +165,12 @@ module Types
       {
         can_view_target_project: project.present?,
       }
+    end
+
+    def audit_events
+      object.audit_events.
+        where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).
+        order(created_at: :desc)
     end
 
     private
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
@@ -23,6 +23,7 @@ module Types
     field :target_enrollment, Types::HmisSchema::Enrollment, null: true
     field :swimlanes, [HmisSchema::CeReferralSwimlane], null: false
     field :workflow_template_name, String, null: true
+    field :events, HmisSchema::CeReferralEvent.page_type, null: false, method: :timeline_events
 
     # Resolve project fields separately, instead of the whole project object, in case user can't view the project
     field :target_project_id, ID, null: false
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
@@ -23,19 +23,13 @@ module Types
     end
 
     def type
-      # todo @martha - add comments justifying the messiness, or make it less messy
       case object.event_type
       when 'complete_step'
         'Completed Task'
-      when 'message_sent'
-        case object.event_data['event']
-        when 'start_workflow'
-          'Started Referral'
-        when 'end_workflow'
-          object.event_data['message'] == 'reject_referral' ? 'Declined Referral' : 'Accepted Referral'
-        else
-          ''
-        end
+      when 'start_workflow'
+        'Started Referral'
+      when 'end_workflow'
+        object.event_data['result'] == 'reject_referral' ? 'Declined Referral' : 'Accepted Referral'
       else
         ''
       end
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
@@ -1,0 +1,44 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeReferralEvent < Types::BaseObject
+    # object is a Hmis::WorkflowExecution::AuditEvent
+    field :id, ID, null: false
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :user, Types::Application::User, null: true
+    field :type, String, null: false, method: :event_type
+    field :step_name, String, null: true
+
+    def step_name
+      step = load_ar_association(object, :step)
+      return nil unless step.present?
+
+      load_ar_association(step, :node).name
+    end
+
+    def type
+      # todo @martha - add comments justifying the messiness, or make it less messy
+      case object.event_type
+      when 'complete_step'
+        'Completed Task'
+      when 'message_sent'
+        case object.event_data['event']
+        when 'start_workflow'
+          'Started Referral'
+        when 'end_workflow'
+          object.event_data['message'] == 'reject_referral' ? 'Declined Referral' : 'Accepted Referral'
+        else
+          ''
+        end
+      else
+        ''
+      end
+    end
+  end
+end
--- a/drivers/hmis/app/models/hmis/ce/referral.rb
+++ b/drivers/hmis/app/models/hmis/ce/referral.rb
@@ -25,9 +25,7 @@ module Hmis::Ce
     has_many :swimlanes, through: :workflow_instance, class_name: 'Hmis::WorkflowDefinition::Swimlane'
     has_many :steps, class_name: 'Hmis::WorkflowExecution::Step', through: :workflow_instance
     has_many :timeline_events, -> {
-      # todo @martha - add comments justifying the messy logic here, or make it less messy with another strategy
-      where(event_type: 'complete_step').
-        or(where(event_type: 'message_sent').where("(event_data->>'event' = 'start_workflow' OR event_data->>'event' = 'end_workflow')")).
+      where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).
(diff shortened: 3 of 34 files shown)


drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -24,8 +24,11 @@ class Referral < GrdaWarehouseBase
     has_one :target_project, class_name: 'Hmis::Hud::Project', through: :opportunity, source: :project
     has_many :swimlanes, through: :workflow_instance, class_name: 'Hmis::WorkflowDefinition::Swimlane'
     has_many :steps, class_name: 'Hmis::WorkflowExecution::Step', through: :workflow_instance
+    has_many :timeline_events, -> {
+      where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).
+        order(created_at: :desc)
+    }, class_name: 'Hmis::WorkflowExecution::AuditEvent', through: :workflow_instance, source: :audit_events
 
-    has_many :steps, class_name: 'Hmis::WorkflowExecution::Step', through: :workflow_instance, source: :steps
```
martha, line 28: Unrelated - I'm removing this duplicate line (compare line 26)

drivers/hmis/spec/requests/hmis/ce/ce_referral_events_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+
+RSpec.describe Hmis::GraphqlController, type: :request do
+  include_context 'ce spec helper'
+
+  before(:each) do
+    hmis_login(user)
+  end
+
+  # todo @martha - this is some repeated code that maybe points to need for some restructuring
+  # it seems like THIS is the thing that should be provided by ce_referral_helper, but who knows how many tests that will break.
```
martha: Todo

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -58,9 +57,26 @@ def start_referral
...
 
-    def accept_referral
+    def accept_referral(message)
       referral.accept!
       referral.opportunity.close!
+
+      referral.workflow_instance.audit_events.create!(
+        event_type: 'end_workflow',
+        user: message.user,
+        event_data: { 'result': 'accept_referral' }.to_h,
```
martha: This is a new data shape in the `event_data` field, which the schema object reads from to determine what event text to show the user. I'm proposing "result" which can be either "accept_referral" or "reject_referral"

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -24,8 +24,11 @@ class Referral < GrdaWarehouseBase
     has_one :target_project, class_name: 'Hmis::Hud::Project', through: :opportunity, source: :project
     has_many :swimlanes, through: :workflow_instance, class_name: 'Hmis::WorkflowDefinition::Swimlane'
     has_many :steps, class_name: 'Hmis::WorkflowExecution::Step', through: :workflow_instance
+    has_many :timeline_events, -> {
+      where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).
+        order(created_at: :desc)
+    }, class_name: 'Hmis::WorkflowExecution::AuditEvent', through: :workflow_instance, source: :audit_events
```
gigxz: Suggestion to simplify this relation to fetch all audit events. That behavior is less surprising for an AcitveRecord relation. This specific logic-- selecting 3 event types and sorting-- is very specific to the GraphQL API.  As opposed to, say, debugging a referral in the rails console and wanting to see all the audit events tied to it.

(We have some other cases where graphql API logic bleeds into AR relations specifically to let us use data loader, which I'm guessing is why you went this way, but I don't think it's needed because we are not resolving these events in batch on multiple referrals).

```suggestion
    has_many :audit_events, class_name: 'Hmis::WorkflowExecution::AuditEvent', through: :workflow_instance, source: :audit_events
```

```ruby
# events resolved on the graphql referral type, for display in referral activity history timeline
# (this method could be here or on the ce_referral type. makes sense to me to be on the referral type because it is specific to the graphql API behavior)
def timeline_events
   audit_events.where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).order(created_at: :desc)
end
```

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
gigxz: Can the end_workflow event be recorded generically by the engine (like other audit events) rather than by the referral message handler? The rest of the audit events there are generic to workflows, so this feels a bit off.

I _think_ something like this would work, where we log the event after sending the message [here](https://github.com/greenriver/hmis-warehouse/blob/80fe9a30e895c832b5ad3bfedccb7af21e838ee3/drivers/hmis/app/models/hmis/workflow_execution/engine.rb#L104-L112):
```ruby
        result = send_message(
          type: trigger.message,
          params: trigger.params,
          step: step,
          user: user,
        )

        # Log audit event only if the message was successfully sent
        if result[:success?]
          log_event('message_sent', user: user, event_data: trigger.to_h, step: step)
          # Log specific message for the end of a workflow
          log_event('end_workflow', user: user, event_data: trigger.to_h, step: step) if event_type == 'end_workflow'
        end

        result
```
Then in the resolver you can look for the actual message name, which is currently hard-coded into `Hmis::Ce::ReferralMessageHandler`, i.e. `event_data["message"] == "reject_referral"` . Those strings `accept_referral` and `reject_referral` could be pulled into constants to remove that floating string introduced in the CeReferralEvent type.

What do you think of that approach? My thought here is to keep audit events localized to the workflow engine, not the message handler. And to avoid introducing new event_data shapes. (Though its a bit willy nilly currently anyway).

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -23,6 +23,7 @@ class HmisSchema::CeReferral < Types::BaseObject
     field :target_enrollment, Types::HmisSchema::Enrollment, null: true
     field :swimlanes, [HmisSchema::CeReferralSwimlane], null: false
     field :workflow_template_name, String, null: true
+    field :events, HmisSchema::CeReferralEvent.page_type, null: false, method: :timeline_events
```
gigxz: I think we might get confused later on when we start to resolve CE Events on referrals as well. I'd suggest calling these "audit events  for example:
```
referral {
  ceEvents { .. } # this will probably be added later. Types::HmisSchema::Event
  auditEvents { .. } # Types::HmisSchema::CeReferralAuditEvent
}
```

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
```diff
@@ -0,0 +1,38 @@
...
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeReferralEvent < Types::BaseObject
+    # object is a Hmis::WorkflowExecution::AuditEvent
+    field :id, ID, null: false
+    field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    field :user, Types::Application::User, null: true
+    field :type, String, null: false, method: :event_type
```
gigxz:
```suggestion
    field :type, String, null: false
```
type is defined below

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_event.rb
```diff
@@ -0,0 +1,38 @@
...
+      case object.event_type
+      when 'complete_step'
+        'Completed Task'
+      when 'start_workflow'
+        'Started Referral'
+      when 'end_workflow'
+        object.event_data['result'] == 'reject_referral' ? 'Declined Referral' : 'Accepted Referral'
+      else
+        ''
+      end
```
gigxz: I would suggest resolving these as an Enum, so the frontend can differentiate them eventually with type safety (Rather than checking against arbitrary strings). For example it might want to show Declined in red, Accepted in green, etc.

gigxz (commented):
Looks good ! some suggestions below

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
martha: This makes sense to me. I've put the constants in the engine, since they are related to referral workflow execution. But they aren't actually used there, and could definitely live elsewhere. Let me know if you prefer another place for them.

drivers/hmis/app/models/hmis/ce/referral.rb
martha: Yep makes sense, good point that it does not need to be batch resolved. (That was the reason I used "timeline" events instead of "audit" events, to use a term that is somewhat frontend specific. But agreed that if it's not needed then we should remove it)

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -12,6 +12,10 @@
 # task assignments, and message processing according to the workflow template.
 module Hmis::WorkflowExecution
   class Engine
+    # Constants for referral event types
+    ACCEPT_REFERRAL = 'accept_referral'
+    REJECT_REFERRAL = 'reject_referral'
```
gigxz: This breaks the current delineation we have where the code in `drivers/hmis/app/models/hmis/workflow_execution` and `../workflow_definition` is _not referral specific_. The referral-related stuff should remain in the `ce/` module. So I'd suggest moving these to the referral message handler, and maybe naming them something more descriptive like ACCEPT_REFERRAL_MESSAGE and REJECT_REFERRAL_MESSAGE.

ttoomey (approved):
This is looking good. I'll just echo Gig's feedback that we'd like to keep the workflow_definition and workflow_execution code generic if possible. If that feels awkward I'm happy to discuss or brainstorm a different approach. Thanks!

gigxz (approved):

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Ok, thank you, sorry

PR #5499, opened 2025-06-18
Patch income cleanup task user id

gigxz:
## _Merging this PR_

## Description
The task had code to ignore income benefits that were migrated-in, but it wasn't working correctly because it didn't check for the HUD "system user" based on data source.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -545,6 +545,7 @@ module HmisDataCleanup
           raise unless record
 
           # skip records that were likely created by migrations
+          next if fp.values.nil?
           next if record.user_id.nil?
           next if record.user_id == system_hud_users.fetch(record.data_source_id)&.user_id
 
--- a/drivers/hmis/lib/hmis_data_cleanup/util.rb
+++ b/drivers/hmis/lib/hmis_data_cleanup/util.rb
@@ -534,7 +534,10 @@ module HmisDataCleanup
         merge(data_sources).
         preload(:income_benefit)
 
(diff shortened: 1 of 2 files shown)


drivers/hmis/lib/hmis_data_cleanup/util.rb
```diff
@@ -534,15 +534,19 @@ def self.correct_total_income_records!
...
+      end.index_by(&:data_source_id)
+
       messages = []
       Hmis::Hud::IncomeBenefit.transaction do
         scope.find_each do |fp|
           record = fp.income_benefit
           raise unless record
 
           # skip records that were likely created by migrations
-          next if record.user_id.nil? || record.user_id == system_hud_user.user_id
```
gigxz, line 545: `record.user_id == system_hud_user.user_id` would always be false, because `Hmis::Hud::User.from_user(Hmis::User.system_user)` would always generate a new UserID since there was no data source id attached to the user passed. The `from_user` should probably raise in that case.

ttoomey: Thanks for catching!! Sorry I missed that

ttoomey (approved):

drivers/hmis/lib/hmis_data_cleanup/util.rb
gigxz: All good! it is sneaky, I can open another PR to have that method raise if there is no data source id, since we shouldn't be creating floating records in the `User` table with no data source id

gigxz:
The system tests failed on https://github.com/greenriver/hmis-warehouse/pull/5503 for a most likely unrelated reason, but I am running the suite locally before merging this

PR #5504, opened 2025-06-20
Sketch out rake task for defining CE workflows

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add a new rake task, `ce_define_workflows`, for iterating on client CE workflows.
- Refactor `Hmis::WorkflowDefinition::Task` into `::ScriptTask` and `::UserTask`, as discussed at dev standup.

To discuss:
- [x] Creating an enrollment as side effect of the referral end event (as opposed to a step). See comment on ticket.
- [x] Reusing forms on multiple steps. See comments added in this PR to `Hmis::WorkflowExecution::Engine`.

## Type of change
New feature


drivers/hmis/app/models/hmis/workflow_definition/template.rb
```diff
@@ -11,7 +11,7 @@ class Template < GrdaWarehouseBase
     has_many :flows, class_name: 'Hmis::WorkflowDefinition::Flow', dependent: :destroy
     has_many :instances, class_name: 'Hmis::WorkflowExecution::Instance', dependent: :restrict_with_exception, foreign_key: 'template_id'
     has_many :swimlanes, class_name: 'Hmis::WorkflowDefinition::Swimlane', dependent: :restrict_with_exception, foreign_key: 'template_id'
-    has_many :ce_opportunities, class_name: 'Hmis::Ce::Opportunity', dependent: :restrict_with_exception
+    has_many :ce_opportunities, class_name: 'Hmis::Ce::Opportunity', dependent: :restrict_with_exception, foreign_key: 'workflow_template_identifier', primary_key: 'identifier'
```
martha, line 14: Fix bug when deleting templates: `ActiveRecord::StatementInvalid: PG::UndefinedColumn: ERROR:  column ce_opportunities.template_id does not exist (ActiveRecord::StatementInvalid)`

martha:
@gigxz this is still in draft since it doesn't reflect the real workflow, but I tagged you so we can discuss the 2 open items mentioned in the PR description. We can take a look together Monday at dev standup.

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -180,11 +180,20 @@ def evaluate_condition(expression)
       calculator.evaluate!(expression, **defaults.merge(all_submitted_values.transform_keys(&:to_sym)))
     end
 
+    # Evaluate conditions against *all* submitted values. This is important because when we reach a gateway,
```
gigxz: Suggest clarifying that this method returns the values, it doesn't do any evaluation
```suggestion
    # Retrieve merged hash of *all* submitted values across all tasks in the workflow. This is important because when we reach a gateway,
```

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -0,0 +1,186 @@
+desc 'Script to create CE workflow definition'
+# This task is for developing and iterating on CE workflow definitions.
+# It will be run in staging/training environments until the workflows are ready, at which point we will run it in production.
+# CAUTION: It deletes existing referrals and opportunities, so that we don't have to worry about definitions shifting underfoot.
+# This means it should NOT be run in production after the first time!
```
gigxz, line 4: suggest adding a `raise` if this is run in production

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -0,0 +1,186 @@
...
+  Hmis::Form::Definition.where(identifier: FORM_DEFINITION_IDENTIFIERS).destroy_all
+  Hmis::Form::Definition.set_callback(:destroy, :before, :can_be_destroyed) # re-enable callback
+
+  puts 'Enabling CE in AppConfigProperty'
+  ce_enabled = AppConfigProperty.find_or_initialize_by(key: 'hmis_ce/enabled')
+  ce_enabled.value = true
+  ce_enabled.save! if ce_enabled.changed?
+
+  # Caution: if we need to run this in a multi-hmis environment, make sure this is the right data source
+  data_source = GrdaWarehouse::DataSource.hmis.order(:id).first
```
gigxz: suggest just raising if there are multiple data sources, so we can cross this bridge explicitly when we get there, rather than relying on a silent ordering behavior
```suggestion
  data_source = GrdaWarehouse::DataSource.hmis.sole
```

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -0,0 +1,186 @@
...
+
+  Hmis::WorkflowDefinition::Swimlane.where(template: templates).destroy_all
+  Hmis::WorkflowDefinition::Flow.where(template: templates).destroy_all
+  Hmis::WorkflowDefinition::Node.where(template: templates).destroy_all
+  templates.destroy_all
+
+  # Temporarily disable the callback that prevents destroying published forms
+  Hmis::Form::Definition.skip_callback(:destroy, :before, :can_be_destroyed)
+  Hmis::Form::Definition.where(identifier: FORM_DEFINITION_IDENTIFIERS).destroy_all
+  Hmis::Form::Definition.set_callback(:destroy, :before, :can_be_destroyed) # re-enable callback
```
gigxz: lets make sure this is filtered by `role` so we don't even destroy non-CE forms

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -0,0 +1,186 @@
+desc 'Script to create CE workflow definition'
+# This task is for developing and iterating on CE workflow definitions.
+# It will be run in staging/training environments until the workflows are ready, at which point we will run it in production.
+# CAUTION: It deletes existing referrals and opportunities, so that we don't have to worry about definitions shifting underfoot.
+# This means it should NOT be run in production after the first time!
+# Usage: rails driver:hmis:ce_define_workflows
+task ce_define_workflows: [:environment] do
```
gigxz: Can we refactor this in a way that will make it easier to define multiple templates in the task? We will likely end up with more than one template.

For example, pulling out a function for `def delete_template_and_associated_data(template_identifier)`, and maybe even moving specific templates to their own functions so we can more easily separate out the logic, like `def create_workflow_template_whatever_its_called!`

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -180,11 +180,20 @@ def evaluate_condition(expression)
...
       steps_by_node_id = instance.steps.index_by(&:node_id)
       template.graph.walk.each.with_object({}) do |node, result|
         step = steps_by_node_id[node.id]
+        # Caution: This logic is a little fragile. Steps may reuse the same form definition, and
+        # we want to always evaluate against the most recently submitted value.
+        # So for example, if step 1 previously submitted "move_forward = 1" but step 2 submits "move_forward = 0",
+        # the referral should not move forward.
+        # This relies on the .merge behavior that overwrites existing keys combined with the fact that we are walking
+        # through the graph sequentially.
         result.merge!(step.submitted_values) if step&.completed?
```
gigxz: >  Reusing forms on multiple steps. See comments added in this PR to Hmis::WorkflowExecution::Engine.

I think it's important to keep the behavior of being able to re-use forms across steps. The comment you added is helpful in understanding whats going on, thanks for that!

Re: the "fragile" descriptor, are there scenarios where you think this would behave differently than expected? Or is it more just, "hey this is specific behavior to pay attention to, this is how we're relying on `merge` behavior".

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: 👍 I think I'd prefer to move the comment above into `evaluate_condition` where it seems more relevant

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: The second thing. I think "fragile" might not be the right word, I'll try to update the comment with clearer language.

drivers/hmis/lib/tasks/ce_define_workflows.rake
martha: Yes, I think this refactor is nicer and hopefully easier to develop with. Thanks!

drivers/hmis/app/models/hmis/workflow_execution/step.rb
```diff
@@ -11,7 +11,7 @@ class Step < GrdaWarehouseBase
 
     belongs_to :instance, class_name: 'Hmis::WorkflowExecution::Instance'
     belongs_to :node, class_name: 'Hmis::WorkflowDefinition::Node'
-    belongs_to :task, class_name: 'Hmis::WorkflowDefinition::Task', foreign_key: 'node_id'
+    belongs_to :task, class_name: 'Hmis::WorkflowDefinition::Node', foreign_key: 'node_id' # Could be a UserTask or a ScriptTask
```
gigxz: Maybe I'm missing something here, but I don't understand why we would need two relations`:node` and `:task` that do exactly the same thing (return a Node model)?

If we want to retrieve the node as a UserTask or ScriptTask subclass specifically, it would make sense to me to have `belongs_to :user_task` and `belongs_to :script_task`.

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -0,0 +1,239 @@
...
+    template.validate!
+
+    puts(template.to_mermaid_diagram)
+
+    template
+  end
+end
+
+desc 'Script to create CE workflow definition'
+task ce_define_workflows: [:environment] do
```
gigxz, line 224: I'm getting the following error:
```
bin/rails aborted!
NoMethodError: undefined method `tasks' for an instance of ActiveRecord::Associations::CollectionProxy (NoMethodError)
/app/drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb:48:in `validate_tasks_and_gateways'
/app/drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb:18:in `validate'
/app/drivers/hmis/app/models/hmis/workflow_definition/template.rb:76:in `validate'
/app/drivers/hmis/app/models/hmis/workflow_definition/template.rb:81:in `validate!'
/app/bin/rails:5:in `<main>'
Tasks: TOP => driver:hmis:ce_define_workflows
```

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
gigxz: Updated comment is much clearer to me, thank you!

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -144,17 +157,14 @@ def traverse_node(node, user)
...
-        step.available_at = Time.current
-        step.enable!
+      when Hmis::WorkflowDefinition::UserTask
+        step = enable_step!(node)
         assign_task!(step)
+      when Hmis::WorkflowDefinition::ScriptTask
+        step = enable_step!(node)
+        # Immediately complete the step without waiting for user action
+        start_step!(step, user: user)
+        complete_step!(step, user: user, submitted_values: {})
```
gigxz, line 169: Looks great!

gigxz (approved):

PR #5511, opened 2025-06-23
7720 hmis policy class

ttoomey:
## _Merging this PR_

## Description
Introduces a policy-based authorization pattern for the HMIS driver. The goal is to make authorization logic more explicit, maintainable, and performant by consolidating scattered permission checks into dedicated policy classes. The new policy checks are limited to CE and the dashboard initially until we have confidence in them.

[Detailed README](https://github.com/greenriver/hmis-warehouse/blob/d9a906d82ed14c3b9e4c0b31dbbd00b61902c4a2/drivers/hmis/app/models/hmis/auth_policies/README.md)

This is similar to the policy pattern used on the warehouse side with some changes:
* extracted loaders out from the user_context into single-purpose classes to help reduce the complexity of the context class
* added collection or class-level actions: `can_index?` - this is a pattern I've seen in the pundit gem. The policy class does double duty for both instance can collection methods. This is a bit of a compromise of the paradigm in exchange for grouping similar code together
* some permissions check include arguments (can_perform?(step: )) to provide additional context. I think this pattern is helpful for the use-cases I've encountered in CE

## Type of change
New feature


app/models/user.rb
```diff
@@ -264,6 +264,8 @@ def reporting_policy_for_project(project_id:, mode: :browse)
 
   memoize def policy_for(resource, policy_class: nil)
     if policy_class
+      raise ArgumentError, "policy class not supported: #{policy_class.name}" unless policy_class < GrdaWarehouse::AuthPolicies::BasePolicy
```
ttoomey, line 268: prevent accidentally using the wrong type of policy

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -65,8 +68,9 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   accepts_nested_attributes_for :affiliations, allow_destroy: true
 
   has_and_belongs_to_many :project_groups,
-                          class_name: 'GrdaWarehouse::ProjectGroup',
-                          join_table: :project_project_groups
```
ttoomey, line 69: these seem incorrect to me; my understanding is that the hmis has a separate project group table

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
```diff
@@ -0,0 +1,66 @@
...
+    user.roles.flat_map(&:granted_permissions).to_set.freeze
+  end
+
+  # Project-specific permissions
+  def project_permissions(project_id)
+    access_group_ids = project_access_group_loader.get(project_id)
+    permission_loader.for_access_group_ids(access_group_ids)
+  end
+
+  def preload_project_dependencies(project_ids)
```
ttoomey, line 39: this could be used to warm the cache ahead of checking a page of projects

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -0,0 +1,62 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Determines a user's permissions for CE Referrals.
+# Key rules:
+# - A user can view a referral if they have broad `:can_view_referrals` permission on the target project.
+# - A user can also view a referral if they have `:can_view_own_referrals` and are assigned to one of its steps.
+# - Indexing (e.g., for a dashboard) requires a combination of view and perform permissions.
```
martha, line 13: It's a bit unintuitive for me that the "index" action requires "perform" permission. Just based on the name, I would've guessed that it's asking "can the user resolve a list of records?" 

It's true that in this specific case, when listing referrals on the dashboard, the user should have "perform" permission too, but I think that logic is not generic -- it might not even be generic within Referrals, if later on we add a page where the user should be able to resolve a list of referrals but without necessarily having edit access. At that point would we be stuck since `can index` already holds logic for the particular dashboard use case? Or, is this an eventuality we don't need to over-optimize for yet :)

drivers/hmis/app/graphql/mutations/ce/start_ce_referral_step.rb
```diff
@@ -20,7 +20,7 @@ def resolve(referral_id:, step_id:)
       referral.opportunity.with_lock do
```
martha, line 20: Would we want to remove the `viewable_by` scope call on line 18?

I'm curious about the ultimate goal with the `viewable_by` scopes -- is it a problem that they duplicate logic with the policies? Do we hope to get rid of them?

In this particular case, for example, we could avoid calling `viewable_by` and instead add another check to the `can_perform?` method of `CeReferralPolicy`, confirming that the user can view this referral as well as perform steps on it. That seems reasonable to me.

But in general, maybe we do need to keep the `viewable_by` scope for cases when we actually are working with a collection and need to filter results down. This mutation is acting on a single step, so it seems most graceful to me to do all permission checking in the policy.

- [`referrals.viewable_by` scope](https://github.com/greenriver/hmis-warehouse/blob/c81328eebdba6b39cb666b72f5b766dbad89218d/drivers/hmis/app/models/hmis/ce/referral.rb#L30-L53) for reference
- it duplicates logic in `Hmis::AuthPolicies::CeReferralPolicy.can_view`

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
ttoomey: Good question, I agree that this isn't intuitive. I was just trying to extract the existing checks but it's also a good time to establish healthy patterns.

Hopefully this policy structure facilitates change and doesn't feel like being 'stuck' - if your feeling is that this pattern might make our permissions more brittle we should stop and re-evaluate.

So my hope is that we can expand or adjust the permissions on these policies as needed. So to support a generic index vs a dashboard-specific check:

```ruby
def can_index?
  
  # Generic: can the user see a list of referrals?
  (context.potential_permissions & [:can_view_referrals, :can_view_own_referrals]).any?
end

def can_index_for_dashboard?
  return false unless can_index?
  
  # Dashboard-specific: user needs both view and perform permissions
  (context.potential_permissions & [:can_perform_any_referral_tasks, :can_perform_own_referral_tasks]).any?
end
```

I'm not sure about the intent behind checking for perform permissions on the dash. If that logic is more for cosmetic purposes (keeping the UI clean) then I think there's a good argument for not including it in the auth policy at all. It seems desirable for our auth policy classes to only contain security rules and UX-driven business logic should be handled elsewhere (perhaps left in the resolver or the front-end)

martha: Thanks, that makes sense. And your snippet seems easy to pivot to in the future, if the `can_index` default behavior should ever change.

I think the intent behind checking "can perform" permission on the dash is just that if the user can view, but not perform, referral tasks, then they shouldn't see referrals on "their" dashboard, because there will be nothing for them to do. I agree, that is sounding more like ux logic than a security rule. So it makes sense to me if we want to move that logic elsewhere.

drivers/hmis/app/graphql/mutations/ce/start_ce_referral_step.rb
ttoomey: So I think there is value in applying both scope and record-level authorization checks. Redundant checks can reduce the impact of errors in our implementation. 

A practice we could use is to filter what a user _can see_ with a scope, then authorize what they _can do_ with the policy.

drivers/hmis/app/models/hmis/auth_policies/README.md
```diff
@@ -0,0 +1,160 @@
...
+  # ... return assignments
+end
+```
+
+### Complex Authorization with Context
+```ruby
+# Authorization that depends on multiple factors
+def can_create_referral?(client:)
+  return false unless client.data_source_id == user.hmis_data_source_id
+  return false unless context.project_role_permissions(opportunity.project_id).include?(:can_start_referrals)
```
gigxz, line 102:
```suggestion
  return false unless context.project_permissions(opportunity.project_id).include?(:can_start_referrals)
```

(It looks like this is the HMIS equivalent of the warehouse project_role_permissions)

drivers/hmis/app/models/hmis/auth_policies/README.md
```diff
@@ -0,0 +1,160 @@
+# HMIS Authorization Policy Architecture
```
gigxz, line 1: This is a super helpful explanation and guide, thank you!

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+
+# Determines a user's permissions for CE Opportunities
+class Hmis::AuthPolicies::CeOpportunityPolicy < Hmis::AuthPolicies::BasePolicy
+  # Accepts client parameter because referral authorization requires validating
+  # the relationship between three entities: user, opportunity, and client.
+  def can_create_referral?(client:)
+    return false unless Hmis::Ce.configuration.enabled?
+
+    # Does the client record data source match the current user?
+    return false unless client.data_source_id == user.hmis_data_source_id
```
gigxz, line 17: Just noting overlap with https://github.com/open-path/Green-River/issues/7671, this will likely change since referral creation will be for a CeCandidate / CeClientProxy, data source check can most likely be removed at that point. @martha

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
```diff
@@ -0,0 +1,33 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Determines a user's permissions for CE Opportunities
+class Hmis::AuthPolicies::CeOpportunityPolicy < Hmis::AuthPolicies::BasePolicy
+  # Accepts client parameter because referral authorization requires validating
+  # the relationship between three entities: user, opportunity, and client.
+  def can_create_referral?(client:)
```
gigxz, line 13: Currently we are relying on the `Hmis::Ce::Opportunity.viewable_by(current_user)` check in the `CreateCeReferral` mutation, to ensure that the user can view this opportunity in the first place. If we wanted to have a redunant check for that here, I think it would look like:
```ruby
context.project_permissions(opportunity.project_id).include?(:can_view_units)
```


Does that seem like something we should add, or not?

This question around redundancy and separating "authorization concerns" vs "presentation concerns" still feels a bit murky to me.

For example:
* For `can_view_candidates?`, should we also check that the user has access to view this opportunity? In my view, we should always require that to be true (based on current application UI/behavior), so we should check it for redundancy. (If not, we are relying on the behavior of the application/API to ensure that opportunity visibility is checked first)
* For `HmisProjectPolicy.can_edit?`, should we also check that you can view the project, since you shouldn't be able to edit something you can't view. Etc. Not suggesting a change here necessarily, maybe this is something we revisit when we apply the project policy, but I want to surface the general concept.

drivers/hmis/app/models/hmis/auth_policies/hmis_project_policy.rb
```diff
@@ -0,0 +1,33 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
```
gigxz, line 8: Just confirming- so far this is only referenced from tests?

(I do find a bit challenging that hypothetically I can't grep for "can_edit_project_details" in the codebase to find where this permission is checked, instead I'm looking for the general `can_edit?` and/or `policy_for(variable_hopefully_named_project)`. I'm curious if that came up on the warehouse side as a developer challenge at all.)

drivers/hmis/app/models/hmis/auth_policies/README.md
ttoomey: Yes, that's accurate. Maybe I should make the names match for consistency

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
ttoomey: So I think I understand what you're getting at; it's a great point. It seems like there's a design choice here:
1) Our policies take on an additional responsibility: the policy also enforces a permission hierarchy (`can_edit` requires `can_view`, `can_delete` requires `can_edit`, etc). This centralizes auth logic in one place
2) Atomic or granular authorization: the policy imposes no hierarchy. A user can have `can_edit` without `can_view`. There either is no defined hierarchy or the permissions granted are assumed to obey the hierarchy and can be trusted

Put another way it seems like the underlying concern  is how we should handle invalid  permission configuration (a user has can_edit but not can_view). It sounds like we do have a hierarchy but maybe it's not formalized anywhere.

While putting that logic into the policies would work to enforce the hierarchy, there are some other options to consider:
* Can we assume the permissions on each role should be internally consistent? If so this could be addressed by validation on the role model to detect invalid permissions combinations
* Alternatively we could enforce the permission hierarchy at the context level, so `context.project_permissions(project_id)` would mask out invalid permissions
* These choices aren't mutually exclusive and we could still check additional permissions in the policy

I guess role-level validation is most appealing to me as it can also be used to help admins avoid creating invalid permission configuration.

Note, I see this as distinct from the authorization vs presentation concern which is more related to if policies should do double-duty and contain logic that isn't strictly auth related. Which is maybe more of a question of how/if policies should support capabilities a user has in the abstract, not connected to a particular resource (can a user perform staff assignments anywhere?)

ttoomey: examples to illustrate permission hierarchy mechanisms
```ruby
#  validate hierarchy configuration at the model level:
class Hmis::Role < ApplicationRecord
  validate :validate_permission_hierarchy
  
  PERMISSION_HIERARCHIES = {
    can_edit_project_details: [:can_view_project],
    can_delete_project: [:can_view_project, :can_edit_project_details],
    can_view_prioritized_client_lists: [:can_view_project],
    can_start_referrals: [:can_view_project]
  }.freeze

  private

  def validate_permission_hierarchy
    granted_permissions.each do |permission|
      required = PERMISSION_HIERARCHIES[permission.to_sym] || []
      missing = required - granted_permissions.map(&:to_sym)
      
      if missing.any?
        errors.add(:permissions, "#{permission} requires #{missing.join(', ')}")
      end
    end
  end
end

#  hierarchy enforcement at the context level
class Hmis::AuthPolicies::UserContext
  PERMISSION_HIERARCHIES = {
    can_edit_project_details: [:can_view_project],
    can_delete_project: [:can_view_project, :can_edit_project_details],
    can_view_prioritized_client_lists: [:can_view_project],
    can_start_referrals: [:can_view_project]
  }.freeze

  # Project-specific permissions with hierarchy enforcement
  def project_permissions(project_id)
    access_group_ids = project_access_group_loader.get(project_id)
    raw_permissions = permission_loader.for_access_group_ids(access_group_ids)
    enforce_permission_hierarchy(raw_permissions)
  end

  private

  def enforce_permission_hierarchy(permissions)
    # Remove permissions that don't have their prerequisites
    valid_permissions = permissions.select do |permission|
      required = PERMISSION_HIERARCHIES[permission.to_sym] || []
      required.all? { |req| permissions.include?(req) }
    end
    
    valid_permissions.to_set.freeze
  end
end
```

ttoomey: tagging @eanders on this too as it feels like we should have a consistent approach on both the warehouse and HMIS side

drivers/hmis/app/models/hmis/auth_policies/hmis_project_policy.rb
ttoomey: > Just confirming- so far this is only referenced from tests?

I think only can_manage_units is used in MarkUnitsAvailable/Unavailable. I'll add a comment to indicate the other permissions are not yet in use

> (I do find a bit challenging that hypothetically I can't grep for "can_edit_project_details" in the codebase to find where this permission is checked, instead I'm looking for the general `can_edit?` and/or `policy_for(variable_hopefully_named_project)`. I'm curious if that came up on the warehouse side as a developer challenge at all.)

This is a good point and I don't think we discussed it directly on the warehouse side. But yes this additional abstraction makes it harder to understand how the permissions are applied. Some level of indirection is unavoidable with any abstraction but we should consider this more carefully.

We could attempt to keep the names of the policy permission methods 1:1 with permission names, such as project_policy.can_edit_project_details? I don't favor this though as it breaks the encapsulation the policy is supposed to provide.

As constructed, I think it should be pretty easy to find the specific policies that use a given permission (HmisProjectPolicy can be discovered by grepping for "can_edit_project_details"). But the difficulty is then in finding where ProjectPolicy is used in the code base.

I could see an argument that the `policy_for(resource)` pattern is a step too far and it's too difficult to know which policy you're working with. A remedy for that might be to require passing in the policy explicitly:
`policy_for(resource, policy_class: Hmis::AuthPolicies::HmisProjectPolicy)`

Alternatively, we could adopt a more descriptive variable naming convention like `project_policy = policy_for(project)` to make the intent clearer without requiring explicit policy classes.

ttoomey: discussed 6/30. Agreed to make policy_class mandatory for better traceability

ttoomey: discussed with Elliot, agreed that policy_class should be required arg

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
ttoomey: discussed with @eanders - he's going to think about it. He mentioned that in the warehouse, permissions maybe implied. So a user with `can_edit` is implied to have `can_view` which is a different application of the hierarchy

eanders: Specifically, I think we should discuss this.  There are three approaches I see:
1. The permission you check is the permission you check, ignoring others.  This assumes the user was granted the permission appropriately.
2. The permission implies other permissions.  This is kinda sorta how the warehouse works, at least in some situations.  One example, if you have Can Edit Clients, it used to (I don't think it does any more) imply Can View Clients, since why would you be able to edit clients without seeing them?
3. The permission requires other permissions to be functional.  This is probably the most defensive/secure.  If you have Can Edit Clients, but not Can View Clients, then Can Edit Clients doesn't actually work, you would need both.  If this were to be our direction, I'd want to see some validations on the role model to reject roles where incompatible combinations were chosen, and ideally an update to the UX to disable/enable permissions.  For instance, you would have to check Can View Clients before you could check Can Edit Clients.   The other major benefit here would be the clarity to the person setting up the role about how the permissions are related.

ttoomey: yes, let's plan to discuss on dev standup next week

drivers/hmis/app/models/hmis/auth_policies/hmis_project_policy.rb
ttoomey: I made this change, it does seem easier to trace but it also looks a bit ugly imo. Does it seem okay or should we consider option 2? (I'm leaning towards option 2)

```ruby
# original: ergonomic but hard to trace. Too much magic
policy_for(referral).can_perform?(step: object)
# option 1: worse ergonomics but easy to trace
policy_for(referral, policy_class: Hmis::AuthPolicies::CeReferralPolicy).can_perform?(step: object)
# option 2: a compromise that relies on mapping symbols to policy classes
policy_for(referral, policy: :ce_referral).can_perform?(step: object)
```

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
ttoomey: introduced `requirements` property to `Hmis::Role.permissions_with_descriptions` with some initial assumptions:
* can_edit_clients requires :can_view_clients (not currently used)
* can_start_referrals requires :can_view_referrals

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -183,6 +183,7 @@ def self.permissions_with_descriptions
       },
       can_start_referrals: {
         description: 'Ability to initiate referrals from the client waitlist in the project',
+        requirements: [:can_view_referrals],
```
ttoomey, line 186: just a guess, maybe this isn't correct

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -48,6 +48,10 @@ class HmisSchema::CeReferral < Types::BaseObject
       arg :on_current_task_since, GraphQL::Types::ISO8601Date # TODO - we will discuss this with design and probably make updates
     end
 
+    def self.authorized?(object, ctx)
+      super && ctx[:current_user].policy_for(object, policy: :ce_referral).can_view?
```
ttoomey: similar to how we authorize access client; maybe we don't want this though

ttoomey:
I think this is ready for another review

app/models/grda_warehouse/auth_policies/user_acl_context.rb
```diff
@@ -15,9 +15,9 @@ class GrdaWarehouse::AuthPolicies::UserAclContext
   EMPTY_SET = Set.new.freeze
 
   def initialize(user)
-    @user = user
-    raise ArgumentError, 'must be acl user' unless @user.using_acls?
+    raise ArgumentError, 'must be acl user' unless user.is_a?(User) && user.using_acls?
```
eanders, line 18: We've been referring to Access Controls as ACLs internally, but if we think we're going to expose this error anywhere, I think we should stick to `user must be using access controls` or something else with the full name.

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
```diff
@@ -26,6 +26,10 @@ def access_denied!(message = 'access denied')
     raise message
   end
 
+  def policy_for(resource, policy:)
```
eanders: Probably too late to change this, but passing a policy to a method called `policy_for` seems weird.  I _think_ this should probably be called `access_rights_for(resource, policy:)`.   

Alternatively, maybe the thing you pass in shouldn't be called `policy` but `context`.  (Looking further down at another usage.  Then it would read something like.

`referral_policy = policy_for(referral, context: :ce_referral)`

Which makes more sense to me.

eanders: `policy_context: :ce_referral` would also be reasonable.

eanders (approved):

gigxz (approved):
looks good to me!

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
ttoomey: good point, how about:
`policy_for(referral, type: :ce_referral)`

(note ideally we should reshape our warehouse policy_for() to match)

app/models/grda_warehouse/auth_policies/user_acl_context.rb
ttoomey: I don't think this error should ever surface to end-users. It's intended as a defensive check to catch bugs

drivers/hmis/app/graphql/concerns/graphql_application_helper.rb
eanders: `type` makes me a bit nervous as it is a reserved word for STI. I know this isn't that, but it feels like it could be confusing.  do you think `policy_context:` is confusing?

ttoomey: > do you think policy_context: is confusing?

yes, "context" has a specific meaning already in our policies, and it's distinct from the policy itself. So I don't think it's a good choice here

Yes it's risky to use `type` as an attribute on an active record model. However it would be a bummer to lose access to a helpful variable name globally. I think 'type' is the correct name for this param but I could also go for 'policy_type'

eanders: ok.  I vote for `policy_type` then.

ttoomey:
@gigxz - sorry I missed test failures related n+1 issues, I've tried to correct them here but my fix could use review.

Changes:
* added a referral/project loader for the policy context
* added support for after_paginate hook to PaginationWrapperExtension which allows use 
* added manual loader calls to resolvers:
  * user dashboard
  * has_units (could further refine with a graphql lookahead but would be brittle)
  * has_ce_referrals
 * bumped the query expectations in the tests

Happy to huddle about it if you have any questions

gigxz (approved):

ttoomey:
Sorry for another iteration but I found a bug related to soft-deletion. Fixed and added tests in bad82202699e12b94ac541b14dedb0c42a732869

gigxz (approved):
looks good thanks for the additional tests

PR #5512, opened 2025-06-23
validate hmis permission mode

ttoomey:
## _Merging this PR_

## Description
Bug Fix & Consistency. Previously, using the string 'all' would have incorrectly defaulted to :any behavior. This change corrects that bug and enforces consistency.

## Type of change
Bug fix 

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/enrollment.rb
+++ b/drivers/hmis/app/models/hmis/hud/enrollment.rb
@@ -159,11 +159,11 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
   #
   # Note: use "replace_scope" hide previous declaration of :viewable_by
   replace_scope :viewable_by, ->(user, include_limited_access_enrollments: false) do
-    return with_access(user, :can_view_enrollment_details, :can_view_project, mode: 'all') unless include_limited_access_enrollments
-    return none unless user.permissions?(:can_view_enrollment_details, :can_view_limited_enrollment_details, mode: 'any')
+    return with_access(user, :can_view_enrollment_details, :can_view_project, mode: :all) unless include_limited_access_enrollments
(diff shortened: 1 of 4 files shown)


gigxz (approved):
Thanks. Opened https://github.com/open-path/Green-River/issues/7870

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -159,11 +159,11 @@ def set_hud_project_id_from_project_pk_unless_wip
   #
   # Note: use "replace_scope" hide previous declaration of :viewable_by
   replace_scope :viewable_by, ->(user, include_limited_access_enrollments: false) do
-    return with_access(user, :can_view_enrollment_details, :can_view_project, mode: 'all') unless include_limited_access_enrollments
-    return none unless user.permissions?(:can_view_enrollment_details, :can_view_limited_enrollment_details, mode: 'any')
```
gigxz, line 163: Confirmed that this line 163 was the only problematic line that causes the new validation to raise. However previously it would have defaulted to `any?` anyway, so there is no change in behavior. This only gets hit when arg `include_limited_access_enrollments: true` is passed, and we don't have great coverage for that. I proposed some more coverage here https://github.com/greenriver/hmis-warehouse/pull/5517/files

The other `mode` usages in this diff go through [Role.with_permissions](https://github.com/greenriver/hmis-warehouse/blob/bc770c59984a1f88b5b423c7c50715ac05b6b2f1/drivers/hmis/app/models/hmis/role.rb#L30-L39) which already had the `to_sym` and raise, so there is no change here. I confirmed that the test `drivers/hmis/spec/models/hmis/hud/enrollment_access_spec.rb` already covers this behavior of having one permission but not the other.

PR #5513, opened 2025-06-23
CSS Theme changes

eanders:
## _Merging this PR_

## Description
* Moves away from precompiled SASS for ENV and CLIENT 
* Converts SASS variables to CSS variables for use in Bootstrap and app CSS
* Adds in-line CSS for CSS variables on a per ENV and per CLIENT basis

## TODO:

- [x] Replace `sg-charles-blue` naming convention with something more generic
- [x] Replace `$cas-*` names with something more generic
- [x] Possibly remove or rename `$dnd`, `$bphc`, `$nechv`, `$ma-state`, `$vi-spdat` SASS variables
- [x] Confirm chart colors work, and aren't significantly different
- [x] Add seed script to pull per-client CSS into the database for the theme, use DB value if present
- [ ] Revisit "gray" variables to see if we can move away from SASS and to using [default BS variables](https://getbootstrap.com/docs/5.3/customize/css-variables/#root-variables)
- [x] Confirm print styles

## Type of change
* New feature
* Code clean-up


dtgreiner (approved):
Nothing is overtly standing out to me, but I do think it would be worth having this on QA for some extra time.

app/assets/stylesheets/application/_custom/default.css
```diff
@@ -0,0 +1,96 @@
...
+
+  .pagination {
+    --bs-pagination-active-bg: var(--brand-primary-d);
+    --bs-pagination-active-border-color: var(--brand-primary-d);
+    --bs-pagination-hover-bg: var(--bs-gray-200);
+  }
+
+  .btn {
+    --bs-btn-active-border-color: var(--brand-primary-l-darkest);
+    --bs-btn-active-border-color: var(--brand-primary-l-darkest);
```
ttoomey: duplicate line

```suggestion
```

app/assets/stylesheets/application/modules/_client.scss
```diff
@@ -829,16 +829,16 @@ $client-image-height: 200px;
...
 .enrollment__new-episode {
   >td {
-    border-bottom: 8px solid darken($table-cell-bg-positive, 10%);
+    border-bottom: 8px solid #b2dba1;
   }
 }
 
 .enrollment__new-episode--key {
-  background-color: darken($table-cell-bg-positive, 10%);
+  background-color: #b2dba1;
```
ttoomey: Does this hard-coded value break the theme? (here and also line 832). Previously this was calculated from theme variables

app/assets/stylesheets/application/_custom/default.css
```diff
@@ -0,0 +1,96 @@
...
+  --bs-warning-desaturated: #c6cdd0;
+  --brand-danger-l-darkest: #f8b87a;
+  --brand-danger-l-desaturated: #e1e2e3;
+  --bs-danger-desaturated: #c6cdd0;
+
+  /* Component Style Overrides to match legacy theme */
+  --nav-tab-active-border-color: var(--brand-primary-d);
+
+  /* Note: these are namespaced within .alert */
+  .alert {
```
ttoomey: So I don't have much experience with css properties but afaict, nested selectors like this are not valid css syntax. The .alert, .pagination, and .btn selectors should not be nested within the :root block. It seems that these should be separate rule blocks.

The css as written fails [css validator](https://jigsaw.w3.org/css-validator/#validate_by_input) 

Does this work?
```css
:root {
  /* ... rest of variables ... */
}

.alert {
  --bs-success-bg-subtle: var(--brand-success-lll);
  --bs-info-bg-subtle: var(--brand-info-lll);
  --bs-warning-bg-subtle: var(--brand-warning-lll);
  --bs-danger-bg-subtle: var(--brand-danger-lll);
}

.pagination {
  --bs-pagination-active-bg: var(--brand-primary-d);
  --bs-pagination-active-border-color: var(--brand-primary-d);
  --bs-pagination-hover-bg: var(--bs-gray-200);
}

.btn {
  --bs-btn-active-border-color: var(--brand-primary-l-darkest);
  --bs-btn-hover-bg: var(--brand-primary-l-darker);
  --bs-btn-hover-border-color: var(--brand-primary-l-darker);
}

.btn-outline-primary {
  --bs-btn-border-color: var(--brand-primary-d);
  --bs-btn-color: var(--brand-primary-d);
  --bs-btn-hover-bg: var(--brand-primary-d);
  --bs-btn-hover-color: white;
}
```

app/assets/stylesheets/application/components/_card.scss
```diff
@@ -314,7 +314,7 @@
 }
 
 .c-card__content-block--danger {
-  background-color: lighten($brand-danger-l, 5%);
+  background-color: rgba(var(--bs-danger-rgb), 0.1);
```
ttoomey: Replacing a lightened color with a transparent version seems incorrect. These are not visually the same. lighten($color, 40%) makes the color whiter, while rgba(var(--color-rgb), 0.2) makes it see-through. This will lead to UI elements having the wrong color or weird bleed through when UI elements overlap

Suggest fixing this here and in other places where we have replaced lighten with rgba.

Looks like we could achieve this with [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix):
```css
.c-card__content-block--danger {
  background-color: color-mix(in srgb, var(--brand-danger-l) 95%, white 5%);
```

Alternatively we could define the needed variants as additional static vars

app/assets/stylesheets/application/_tools/mixins/checkbox.scss
```diff
@@ -99,25 +118,30 @@
...
-    color: darken($brand-warning, 10%);
+    color: var(--bs-warning);
+    filter: brightness(90%);
   }
+
   &:hover,
   &:focus {
     &::before {
-      background: rgba($brand-warning-l, .5);
+      background: rgba(var(--brand-warning-l-rgb), .5);
```
ttoomey: Where is `--brand-success-l-rgb` defined? We have `--bs-success-rgb`.  Same for `--brand-warning-l-rgb`. These undefined vars are also referenced in buttons scss

Generally, if a css custom properties is not defined there's no build error, where as with scss vars the assets would fail to compile. So there's a risk here I'm not sure how to mitigate. Perhaps there's a linter we can bring into our build pipeline to guard against this

ttoomey: I guess you can provide fallback defaults if the var is not defined. Not sure that's great either
`color: var(--brand-primary, #457687);`

app/assets/stylesheets/application/overrides/bootstrap/table.scss
```diff
@@ -12,7 +12,7 @@
 
 .table__detail-row-rollup {
   &.open td {
-    background: lighten($tab-bg, 30%);
+    background: #ffffff;
```
ttoomey, line 15: probably should come from the theme, not hard-coded

eanders: Maybe, though I think the goal is to have theming be minimal, some highlight colors, header, footer, menu colors and logos.

app/assets/stylesheets/application/_tools/mixins/checkbox.scss
eanders: Good callout.  I'm going back through all style files and confirming we have accounted for all CSS variables.  Looks like there were 9 missing from the default.css file.  Thanks!

app/assets/stylesheets/application/components/_card.scss
eanders: Thanks!  I think we should lean on defining the various variants in the default styles explicitly.  That way if we ever want to do a dark theme, where the color mix might be reversed, we can just update the one file.

app/assets/stylesheets/application/modules/_client.scss
eanders: I'll move that to a theme color.  Thanks!

app/assets/stylesheets/application/_custom/default.css
eanders: Hmm, I thought that hadn't worked before, but it seems happy now.  Adjusting.  Thanks.

app/assets/stylesheets/application/_custom/hforward.css
```diff
@@ -0,0 +1,39 @@
...
+  --bs-danger-saturation: 75%;
+  --bs-danger-lightness: 52%;
+
+  /* Theme Background: #4F2C1D */
+  --theme-background-hue: 20;
+  --theme-background-saturation: 46%;
+  --theme-background-lightness: 21%;
+
+  /* Custom overrides */
+  --brand-primary-lll: #d39f89;
```
ttoomey, line 28: seems like there may be too many 'l's here

app/assets/stylesheets/application/_custom/allegheny.css
```diff
@@ -0,0 +1,25 @@
+:root {
+  /* Primary: #0065a6 */
```
ttoomey, line 2: this doesn't match the primary from the sass file
$brand-primary: #0A406B;

ttoomey: Also, the original sass had customized variables ($brand-success: #006E47;) etc. We are not defining those here. Was it intentional to reduce scope of the customization? This may result in a different look and feel for some themes. Perhaps the translation of these themes is incomplete?

app/assets/stylesheets/application/components/_tag.scss
```diff
@@ -98,15 +103,15 @@ $tag-icon-shadow: $white;
...
 
 // We sometimes need the light version of colors to make the tags look reasonable
 $tag-types: (
-  success: $brand-success-l,
-  warning: $brand-warning,
-  danger: $brand-danger-l,
-  info: $brand-info,
+  success: var(--brand-success-l),
+  warning: var(--brand-warning-l),
+  danger: var(--brand-danger-lll),
```
ttoomey, line 113: The danger and warning tag colors are a lighter shade I guess? Seems like many colors have been lighted so guessing this is intentional

app/assets/stylesheets/application/layouts/_main.scss
```diff
@@ -2,38 +2,41 @@
...
     flex: 0 0 calc(100vw - #{$nav-side-width});
     max-width: calc(100vw - #{$nav-side-width});
   }
 }
 
-#impersonation-warning
-{
-  background-color: $brand-danger-d;
+#impersonation-warning {
+  background-color: var(--brand-danger-l);
```
ttoomey, line 32: intentional to lighten the bg?

app/assets/stylesheets/application/modules/_health_forms.scss
```diff
@@ -122,76 +137,66 @@
...
-  color: $cob-sg-blue-005;
+.qa-valid-unpayable-text {
+  color: var(--brand-primary-lll);
 }
 
-.qa-ignored
-{
-  border-left: 4px solid $gray;
+.qa-ignored {
+  border-left: 4px solid;
```
ttoomey, line 189:
```suggestion
  border-left: 4px solid var(--bs-gray);
```

app/assets/stylesheets/application/modules/_sections_slabs_cards.scss
```diff
@@ -65,39 +76,41 @@
...
 .health__team-member__action a:hover,
 .health__team-member__action a:active,
 .health__team-member__action a:focus,
 .health-careplan__goal-action a:hover,
 .health-careplan__goal-action a:active,
-.health-careplan__goal-action a:focus
- {
-  color: darken($cob-sg-blue-006, 20);
+.health-careplan__goal-action a:focus {
+  color: var(--brand-primary-l);
```
ttoomey, line 110: this seems like the same color as the default state, so we the hover state no longer has visual differentiation.

```suggestion
  color: var(--brand-primary-l);
  filter: brightness(80%);

ttoomey (approved):
Looks good! Left some notes but nothing blocking

PR #5521, opened 2025-06-24
7675 add cdes to ce eligibility

ttoomey:
## _Merging this PR_

## Description
Support Custom Data Elements in CE Eligibility expressions

* Supports name-spaced variables in expressions. Note that using namespaces requires quoted identifiers for dentaku to parse correctly.
* Split FieldMap into specialized resolvers: ClientFieldMap for client demographics and CdeFieldMap for Custom Data Elements (CDEs)
* Support multi-valued CDEs; adds include() and exclude() custom functions
* Added CDED.read_value_from as an improved reader pattern
* Adjusted SqlExpressionTranslator to prevent invalid SQL generation when encountering unknown fields
* Adjusted engine to support expressions on destination clients

### Not addressed in this PR
* sql filters for CDEs. I think some basic filtering is relatively achievable but adds complexity

## Type of change
New feature


drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
```diff
@@ -41,4 +43,27 @@ class Hmis::Hud::CustomDataElementDefinition < Hmis::Hud::Base
   scope :for_clients, -> { for_type(Hmis::Hud::Client.sti_name) }
 
   use_enum_with_same_key :form_role_enum_map, FIELD_TYPES.map { |f| [f, f.to_s.humanize] }.to_h
+
+  def read_value_from(custom_data_element)
```
ttoomey, line 47: Alternative to the cde.value method; this offers stricter behavior and avoids iterating through all possible value fields

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+# resolve custom data elements (CDEs)
+module Hmis::Ce::Match
+  class CdeFieldMap
+    def instance_value(client, field)
```
ttoomey, line 10: this could be made more efficient. Wanted to confirm it is logically correct before optimizing

drivers/hmis/app/models/hmis/ce/match/client_expression_evaluator.rb
```diff
@@ -5,24 +5,24 @@
...
     attr_reader :expression, :field_map
+
     def initialize(expression, field_map)
       @expression = expression
       @field_map = field_map
+      @calculator = Hmis::Ce::Match::CalculatorFactory.build
     end
 
     def call(client)
-      calculator = Dentaku::Calculator.new
```
ttoomey, line 14: I think we can share a calculator safely

drivers/hmis/app/models/hmis/ce/match/client_field_map.rb
```diff
@@ -0,0 +1,59 @@
+# frozen_string_literal: true
+
+# resolve fields from the client table
+module Hmis::Ce::Match
+  class ClientFieldMap
```
ttoomey, line 5: extracted from FieldMap, FieldMap is now just a router

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
```diff
@@ -0,0 +1,62 @@
...
+      cded = parse_entity_type(field)
+
```
gigxz, line 12: I think we could use a couple more checks here

```

raise "evaluating non-assessment CDEDs not yet supported" unless cded.owner_type === "Hmis::Hud::CustomAssessment"

raise "cant resolve CDED that doesnt reference a form_definition_identifier" if cded. form_definition_identifier.nil? # could happen for shared attributes across assessments, better to raise explicitly
```

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
```diff
@@ -0,0 +1,62 @@
...
+      # choose the assessment that was most recently updated
+      record = client.custom_assessments.joins(:definition).
+        where(definition: { identifier: cded.form_definition_identifier }).
+        order(:date_updated, :id).
+        last
+      return nil unless record
+
+      # Get all CDE values for this definition
+      cdes = record.custom_data_elements.where(data_element_definition: cded)
+      return nil if cdes.empty?
```
gigxz: Maybe worth adding a comment saying that this could occur if...
- the question was left empty on the form, or
- the question was disabled by conditional logic on the form, or
- the question was asked on a previous version of this form, but is no longer asked on the latest version of the form

Those are the scenarios I can think of. We will be inevitably digging in this code wondering "why does someone not have a value for X".

It seems like it would also be possible for the CDE to _exist_ but not have a value in it, so maybe its worth mentioning that. I'm not 100% sure off the top of my head which of the above scenarios would lead to a missing CDE vs a no-value CDE. No need to dig into all that here, but some sort of comment may be helpful in the future.

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
gigxz: nevermind, seeing something like that in parse_entity_type.

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
gigxz: this looks logically correct to me!

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
```diff
@@ -0,0 +1,62 @@
...
+      end
+
+      cded = cded_lookup.dig(klass.sti_name, cde_key)
+      raise ArgumentError, "Unknown CDE in field \"#{field}\"" unless cded
+
+      cded
+    end
+
+    def cded_lookup
+      # perhaps this should be constrained by data source
```
gigxz: Hmm. As written, perhaps yes, since we're still in the world of dealing with source clients. 

However with https://github.com/open-path/Green-River/issues/7671 coming in, this will need to be slightly changed since `client` will be a destination client. The above `client.custom_assessments...` query will need to be adjusted to fetch custom assessments for all associated source clients, which may span data sources.

That's all to say, I _think_ we should leave this lookup as-is, including all data sources, in anticipation of issue 7671. For better or worse, the uniqueness constraint on `owner_type,key` spans across data sources.

cc @martha since she'll likely be working on that ticket.

gigxz: happy to discuss more, but I think leaving as-is for now makes sense

drivers/hmis/app/models/hmis/ce/match/client_field_map.rb
```diff
@@ -0,0 +1,59 @@
...
+          arel_field: age_from(current_date, arel.c_t['DOB']),
+        },
+        days_homeless: {
+          instance_value: ->(c) do
+            GrdaWarehouse::Hud::Client.days_homeless(client_id: c.id)
+          end,
+        },
+        aha_score: {
+          instance_value: ->(_) { 20 }, # TODO(#7164) this is just a mocked value to see how things look on the Opportunity page
+        },
```
gigxz: We can go ahead and remove this. Aha score will be stored in a CDE on an assessment anyway, so the new calculator you added should be sufficient. (Prioritization is based on score at the time the assessment was performed)

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
ttoomey: notes from discussion: I'm going to adjust the resolver to support destination clients in anticipation of 7671

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
ttoomey: notes from discussion: not going to invest in performance optimization yet

ttoomey:
Adjusted the match engine to accept destination clients, see [Todo](https://github.com/greenriver/hmis-warehouse/pull/5521/files#diff-21dff283a1067a329678a76a670d09dd547ae74e41f7cc948d97ce6af0637077R19)

I think this is ready for final review.

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
```diff
@@ -0,0 +1,73 @@
...
+    def arel_field(_field)
+      nil
+    end
+
+    protected
+
+    # client is a destination client
+    def custom_assessment_cdes(cded, client)
+      # choose the assessment that was most recently updated
+      Hmis::Hud::Client.arel_table
```
gigxz: does this line do anything?

drivers/hmis/app/models/hmis/ce/match/sql_expression_translator.rb
```diff
@@ -36,6 +36,10 @@ def visit(node)
       result
     end
 
+    def visit_function(_node)
```
gigxz, line 39: what's this for?

drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb
```diff
@@ -28,60 +19,161 @@
...
-      results = generate_candidates(pool, clients)
-      expect(results).to be_empty
+  shared_context 'with demographic test clients' do
+    let(:client_adult_non_veteran) { create(:hmis_hud_client, veteran_status: 0, dob: 20.years.ago) }
+    let(:client_minor_non_veteran) { create(:hmis_hud_client, veteran_status: 0, dob: 10.years.ago) }
+    let(:client_adult_veteran) { create(:hmis_hud_client, veteran_status: 1, dob: 20.years.ago) }
+    let(:client_senior_veteran) { create(:hmis_hud_client, veteran_status: 1, dob: 68.years.ago) }
+
+    let(:all_clients) do
+      [client_adult_non_veteran, client_minor_non_veteran, client_adult_veteran, client_senior_veteran]
```
gigxz, line 40: Not sure exactly where to put it, but I think it would be nice to test these expressions against a client with missing dob (and/or missing veteran status), so we have that expectation written out somewhere. (Eg `current_age > 18` excludes clients with missing dob, I think)

drivers/hmis/spec/models/hmis/ce/sql_expression_translator_spec.rb
```diff
@@ -72,5 +72,45 @@
       result = described_class.call('current_age = ((10 + 5) * (2 + 3))', field_map)
       expect(result.to_sql).to include('((10 + 5) * (2 + 3))')
     end
+
+    context 'with unknown fields' do
```
gigxz: It looks like "unknown_field" is something that's not resolvable using the field maps, is that right? I would think that we should raise an error in that case? it feels like a potential recipe for silent failures, for example if we have a typo on a cded key name or something, it would always pass? Let me know if I'm misunderstanding this!

gigxz (approved):
looks great!

martha (approved):
Thank you!

drivers/hmis/app/models/hmis/ce/match/cde_field_map.rb
ttoomey: ah oops, it's just cruft. Thanks!

drivers/hmis/app/models/hmis/ce/match/sql_expression_translator.rb
ttoomey: this is called by dentaku's expression processor when we visit a function (like visit_node, operation, etc). Needed to support expressions with includes(), excludes(), or other functions

drivers/hmis/spec/models/hmis/ce/sql_expression_translator_spec.rb
ttoomey: Sorry this code should be more clear. In the context of the SqlExpressionTranslator, an "unknown field" is one that can't be resolved into SQL, which is the case for all CDE fields currently. Because they aren't filtered in the database, the SQL filter treats them as true.

A typo in a CDE key (e.g., 'cde.custom_assessment.some_key_with_a_typo') will pass the initial SQL filter but should then be caught during the second evaluation phase. The CdeFieldMap should raise if the CDE definition can't be found.

I've adjusted the language to replace 'unknown' with 'unresolved' which hopefully clarifies the intent.

drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb
ttoomey: good idea, I'll add a test

drivers/hmis/app/models/hmis/ce/match/sql_expression_translator.rb
gigxz: gotcha, thanks!

drivers/hmis/spec/models/hmis/ce/sql_expression_translator_spec.rb
gigxz: Ah I see, thanks for the explainer!

gigxz (approved):
💯 💯 💯

drivers/hmis/app/models/hmis/ce/match/engine.rb
```diff
@@ -14,6 +14,10 @@ def self.call(...)
     # 1. Translate the eligibility requirements expression into a SQL condition and filter the clients. Uses field_map.arel_node to achieve this translation. Expression components that cannot be represented in SQL are treated as truthy. This reduces the number of client records that we need to evaluate in the more expensive second step.
     # 2. Evaluate the eligibility requirement expression against each matched client. We expect all expression variables to be defined.
     def call(pool, clients)
+      validate_clients_parameter!(clients)
+      # TODO: remove this in #7671
+      clients = translate_source_to_destination_scope(clients)
```
gigxz, line 19: heads up @martha @ttoomey in order to use this end-to-end with the frontend, I need to make the following change on line 34. Or else the candidate pool  contains destination clients, and that causes issues with visibility and permission checks down the line

```diff
-            client_id: client.id,
+            client_id: client.source_clients.first.id, # TODO: update with #7671. map back to a source client (should check for hmis data soruce too to be sure)
```

PR #5522, opened 2025-06-24
CE Workflow Tasks for QA

gigxz:
## _Merging this PR_

## Description
Follow-up to https://github.com/greenriver/hmis-warehouse/pull/5504
Issue https://github.com/open-path/Green-River/issues/7830

How to test:
* Run` rails driver:hmis:ce_define_workflows`
* In a ce-enabled project, create a unit group that uses `Housing Referral Workflow V1`
* Create units, mark as available, initiate referral
* Run through referrals with each pathway. Using generated mermaid diagram is helpful. test denial review workflow

## Type of change
configuration setup for QA

```mermaid
flowchart TD
start_referral_396(("Start Referral"))
start_referral_396--> initial_review_397
initial_review_397("Initial Review")
initial_review_397--> exclusive_gateway_initial_review_task_405
exclusive_gateway_initial_review_task_405{"Exclusive Gateway: initial_review_task"}
exclusive_gateway_initial_review_task_405-- IF move_forward = 0 --> referral_declined_404
exclusive_gateway_initial_review_task_405--> client_acceptance_398
referral_declined_404(("Referral Declined"))
exclusive_gateway_ce_offer_outcome_406-- IF move_forward = 0 --> referral_declined_404
exclusive_gateway_denial_review_408-- IF ac_workflow_v1_denial_review_decision = 1 --> referral_declined_404
confirm_success_402-- IF move_forward = 0 --> referral_declined_404
client_acceptance_398("Client Acceptance")
client_acceptance_398--> exclusive_gateway_ce_offer_outcome_406
exclusive_gateway_ce_offer_outcome_406{"Exclusive Gateway: ce_offer_outcome"}
exclusive_gateway_ce_offer_outcome_406--> provider_acceptance_399
provider_acceptance_399("Provider Acceptance")
exclusive_gateway_denial_review_408--> provider_acceptance_399
provider_acceptance_399--> exclusive_gateway_project_offer_outcome_407
exclusive_gateway_project_offer_outcome_407{"Exclusive Gateway: project_offer_outcome"}
exclusive_gateway_project_offer_outcome_407-- IF move_forward = 0 --> denial_review_401
exclusive_gateway_project_offer_outcome_407--> create_enrollment_400
denial_review_401("Denial Review")
denial_review_401--> exclusive_gateway_denial_review_408
exclusive_gateway_denial_review_408{"Exclusive Gateway: denial_review"}
create_enrollment_400("Create Enrollment")
create_enrollment_400--> confirm_success_402
confirm_success_402("Confirm Success")
confirm_success_402--> referral_accepted_403
referral_accepted_403(("Referral Accepted"))
```

### DIFF
--- a/drivers/hmis/lib/tasks/ce_define_workflows.rake
+++ b/drivers/hmis/lib/tasks/ce_define_workflows.rake
@@ -61,7 +61,7 @@ module CeWorkflowBuilder
           event: 'start_workflow',
           message: 'start_referral',
         },
-      ]
+      ],
     )
   end
 
@@ -75,14 +75,16 @@ module CeWorkflowBuilder
           message: Hmis::Ce::ReferralMessageHandler::ACCEPT_REFERRAL_MESSAGE,
         },
         *(
-          [
-            {
-              event: 'end_workflow',
-              message: 'create_enrollment',
-            },
-          ] if create_enrollment
-        )
-      ].compact
+          if create_enrollment
+            [
+              {
+                event: 'end_workflow',
+                message: 'create_enrollment',
+              },
+            ]
+          end
+        ),
+      ].compact,
     )
   end
 
@@ -103,7 +105,7 @@ module CeWorkflowBuilder
           event: 'end_workflow',
           message: Hmis::Ce::ReferralMessageHandler::REJECT_REFERRAL_MESSAGE,
         },
-      ]
+      ],
     )
   end
 
@@ -117,6 +119,7 @@ module CeWorkflowBuilder
       definition: definition,
     )
     raise 'Form definition must be present' if definition.blank?
+
     errors = Hmis::Form::DefinitionValidator.perform(definition, form_def.role, skip_cded_validation: true)
     raise "Form definition #{form_def.identifier} is not valid: #{errors.map(&:full_message)}" if errors.any?
 
@@ -137,12 +140,12 @@ module CeWorkflowBuilder
     confirm_success_task_form_identifier = 'ac_workflow_v1_confirm_success_task'
 
     delete_form_definitions([
-      initial_review_task_form_identifier,
-      ce_offer_task_form_identifier,
-      project_offer_task_form_identifier,
-      denial_review_form_identifier,
-      confirm_success_task_form_identifier,
-    ])
+                              initial_review_task_form_identifier,
+                              ce_offer_task_form_identifier,
+                              project_offer_task_form_identifier,
+                              denial_review_form_identifier,
+                              confirm_success_task_form_identifier,
+                            ])
 
     puts "Creating workflow definition template '#{identifier}'"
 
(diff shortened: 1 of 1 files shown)


drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -106,111 +105,322 @@ module CeWorkflowBuilder
...
+    create_step_form(
+      identifier: project_offer_task_form_identifier,
+      definition: {
+        "item": [
+          {
+            "text": 'Date',
+            "type": 'DATE',
+            "link_id": 'date',
+            "required": true,
+            "mapping": { "custom_field_key": 'ce_generic_date' },
```
martha, line 236: Looks like you updated these above to `ce_task_date` and `ce_task_notes`, want to make the same update here?

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -106,111 +105,322 @@ module CeWorkflowBuilder
...
+            ],
+            "mapping": { "custom_field_key": 'ac_workflow_v1_denial_review_reason' },
+            "enable_when": [{ "question": 'denial_review_decision', "operator": 'EQUAL', "answer_code": '0' }],
+          },
+        ],
+      },
+    )
+    create_step_form(
+      identifier: confirm_success_task_form_identifier,
+      definition: ce_staff_shared_form,
```
martha: For the sake of making this into a sensible demo, maybe we don't want to use the shared form for this step? At this point in the referral, a target enrollment has just been created, so asking "Continue with Referral?" sounds odd to me.

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -106,111 +105,322 @@ module CeWorkflowBuilder
...
+    denial_review_gateway = create_gateway(template, 'denial_review')
+
+    start_event.connect_to!(initial_review_task)
+
+    # Initial Review => Gateway
+    initial_review_task.connect_to!(initial_review_task_gateway)
+    # Initial Review Gateway => CE Make Offer Task OR Decline Event.
+    # Exclusive Gateway, so only the first outflow that matches condition is followed.
+    initial_review_task_gateway.connect_to!(decline_event, condition: 'move_forward = 0')
+    initial_review_task_gateway.connect_to!(ce_make_offer_task) # default outflow, so it appears under "unavailable tasks"
```
martha, line 466: Ah so let's revisit this! (Though not necessarily in this PR - I'm fine with merging as is.) The original intention behind [this logic here](https://github.com/greenriver/hmis-warehouse/blob/release-169/drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb#L56-L57) that hides conditional steps was so that, in a circular workflow, the "Review Denial" step would *not* appear unless a Denial was actually happening.

But that leads us to this weird finicky place where we need to make the explicit condition "move_forward = 0" and treat "move_forward = 1" as the default condition, which doesn't feel right at all, as you mentioned. Also, now the Denial step is correctly hidden unless it's happening, but on the other hand, the Confirm Success step is showing up even when we are in a Denial loop, which _also_ feels incorrect.

Here is a quick brainstorm of ideas:
- keep what we have, it's good enough for MVP
- remove the conditional logic, and just display all steps for MVP
- update the conditional logic: add a "happy path" flag to the task definitions (new col in the db table?), and only show "happy path" tasks. Unless any "unhappy path" tasks get started, and then stop showing happy path(?). This feels very specific to this particular workflow, I'm not sure I like it as a general solution.

martha (approved):
Awesome! thanks Gig! 🚀

drivers/hmis/lib/tasks/ce_define_workflows.rake
martha: - add a new integer "path happiness rank" field to the task definition. This could be a more generic version of the "happy path" boolean. The referral could then resolve all tasks in the "happiest possible given current progress" path. (It's generic but doesn't fit our workflow exactly, because in the Denial/Approve Denial loop, an Acceptance is still possible)

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: I added a TODO about updating this to generate custom data element definitions; we can handle updating the keys (or having them be generated) there. I'll do that in another PR because I think we want to handle that more generically and make use of the existing code from publishing forms.

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: Good point, updated:
Screenshot 2025-06-25 at 12 15 47 PM

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: Thanks for thinking through that! Would you mind putting those ideas into an issue? I think we can revisit this if/when we get feedback about the visibility of unavailable tasks, and/or when we have more experience building out different workflows.

martha: Sure yes, created https://github.com/open-path/Green-River/issues/7877

gigxz: thanks!

PR #5525, opened 2025-06-25
Add Source Enrollment to Referral

martha:
## _Merging this PR_

## Description
Summary of changes:
- Add source enrollment on the CE Referrals table
- Add a new graphql api type, `CeReferralSourceEnrollment`, with comments to distinguish it from `Enrollment`
- Update the `CreateCeReferral` mutation to accept a source enrollment instead of a client ID
- Add a new root-level query for resolving more details on a single candidate
- Add data source validation on the Referral, expecting that its constituents' data sources match

## Type of change
New feature

## How to test
- Check out the frontend code: https://github.com/greenriver/hmis-frontend/pull/1187. It's still in progress (because the designs are underway)
- Go to a CE project and navigate to any opportunity with candidates
- Click "Start Referral"
- Play around with the Select Enrollment step on the dialog

  - [n/a] ensured the API can't leak data from other data sources - in this case data from other sources is intentionally returned.


drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -132,5 +134,16 @@ def ce_template
...
+
+    def consistent_data_source
+      msg = 'must be in same data source as opportunity'
+
+      # Opportunity takes the workflow template data source as the "source of truth", so do the same here
+      data_source = opportunity.workflow_template.data_source
+
+      errors.add(:client, msg) unless data_source == client.data_source
+      errors.add(:target_enrollment, msg) if target_enrollment && data_source != target_enrollment.data_source
+      # Source enrollment doesn't necessarily need to be in the same data source as the opportunity
```
martha, line 146: To confirm - is this correct, that the source enrollment could be in a different data source?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -560,6 +560,20 @@ def ce_opportunity(id:)
...
     end
 
+    field :ce_candidate, HmisSchema::CeCandidate, null: true do
+      argument :id, ID, required: true
+    end
+
+    def ce_candidate(id:)
+      raise unless Hmis::Ce.configuration.enabled?
+
+      # Candidates aren't tied to a specific project, so check if the user can_view_prioritized_client_lists in any project.
```
martha: To discuss - Is this enough for permission guarding this method? 

Do we need to protect against data leakage across data source? I think no, because the client will be a destination client.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+        group.max_by { |assessment| [assessment.assessment_date, assessment.id] }
+      end
+
+      latest_assessments.map do |assessment|
+        OpenStruct.new(
+          id: assessment.id,
+          name: assessment.definition.title,
+          date: assessment.assessment_date,
+        )
+      end
```
gigxz: I'm not sure if this needs to be an OpenStruct. The struct data is only source from one record (assessment), so why not just resolve the type using the `assessment` model?

drivers/hmis/app/graphql/types/hmis_schema/assessment_name_with_date.rb
```diff
@@ -0,0 +1,22 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::AssessmentNameWithDate < Types::BaseObject
```
gigxz: This naming is locking us in to resolving a certain set of fields. I wonder if we can name it something more generic, maybe following the pattern of the existing `EnrollmentSummary`? We might need this concept in the future, and I imagine there is a somewhat fixed set of metadata that is "safe" to show in an extended-access scenario. 
```
module Types
  class HmisSchema::AssessmentSummary < Types::BaseObject
    # object is a Hmis::Hud::CustomAssessment
    # there is no object-level authorization because the current user may not have access to view this full assessment. this type is used to resolve "extended access" to assessments, when they would not otherwise be visible (CE referral use-case)
   # this type resolves limited data to convey limited metadata about an assessment
```

drivers/hmis/app/graphql/types/hmis_schema/assessment_name_with_date.rb
```diff
@@ -0,0 +1,22 @@
...
+    # {
+    #   id: ID,
+    #   name: String,
+    #   date: GraphQL::Types::ISO8601Date,
+    # }
+
+    field :id, ID, null: false
+    field :name, String, null: false
+    field :date, GraphQL::Types::ISO8601Date, null: false
+  end
```
gigxz: suggest resolving timestamps for when the assessment was created and updated. In case we decide those are relevant to show later.

drivers/hmis/app/graphql/types/hmis_schema/assessment_name_with_date.rb
```diff
@@ -0,0 +1,22 @@
...
+  class HmisSchema::AssessmentNameWithDate < Types::BaseObject
+    # object is an OpenStruct with the following shape:
+    # {
+    #   id: ID,
+    #   name: String,
+    #   date: GraphQL::Types::ISO8601Date,
+    # }
+
+    field :id, ID, null: false
+    field :name, String, null: false
```
gigxz: clarify that this is not the client name, project name, etc
```suggestion
    field :assessment_name, String, null: false
```

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -8,14 +8,30 @@
...
     def client
       load_ar_scope(scope: Hmis::Hud::Client.viewable_by(current_user), id: object.client_id)
     end
+
+    def enrollments # not for batch
+      form_definition_identifiers = object.candidate_pool.relevant_form_definition_identifiers
+
+      # TODO(#7671) - adjust this to fetch all enrollments from all clients
+      object.client.enrollments.
+        viewable_by(current_user). # For now, filter enrollments to those viewable by the current user. Future workflows may require more access.
```
gigxz:
```suggestion
        viewable_by(current_user). # For now, filter enrollments to those viewable by the current user. In the future we may need to update this to support resolving enrollments that are NOT viewable to the current user.
```

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -8,14 +8,30 @@
...
 
     def client
       load_ar_scope(scope: Hmis::Hud::Client.viewable_by(current_user), id: object.client_id)
     end
+
+    def enrollments # not for batch
+      form_definition_identifiers = object.candidate_pool.relevant_form_definition_identifiers
+
+      # TODO(#7671) - adjust this to fetch all enrollments from all clients
+      object.client.enrollments.
```
gigxz: Suggest additional comment:
> For now, resolve all current and historical enrollments for this client (including open, exited, and wip). Resolve broad scope because candidate pool calculation takes into account all of these enrollments, so it's possible that someone is eligible based on an old exited enrollment. In the future, we may want to limit this scope (possibly based on a configuration), but ideally this scope should align with the scope of enrollments used for candidate pool membership evaluation.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+
+    def relationship_to_ho_h
+      object.enrollment.relationship_to_ho_h
+    end
+
+    def household_size
+      household_members.map(&:personal_id).uniq.size
+    end
+
+    def other_household_member_names
```
gigxz, line 110: Just to be safe, since this is exposing client names, let's add a permission check to ensure that we only show this if the user has full enrollment access. In the future we plan to use this type to resolve enrollments that the user doesn't have full access to, and in that case I don't think we want/need to be showing other names.

```suggestion
    def other_household_member_names
      # Only resolve household member names if this user has permission to view full enrollment details.
      return [] unless can_view_enrollment_details
```
```
    private

    def can_view_enrollment_details
      current_permission?(permission: :can_view_enrollment_details, entity: object.enrollment)
    end
```

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+    field :exit_date, GraphQL::Types::ISO8601Date, null: true
+    field :in_progress, Boolean, null: false
+    field :auto_exited, Boolean, null: false
+
+    field :relationship_to_ho_h, HmisSchema::Enums::Hud::RelationshipToHoH, null: false, default_value: 99
+    field :household_size, Integer, null: false
+    field :other_household_member_names, [String], null: false
+
+    field :assessments, [Types::HmisSchema::AssessmentNameWithDate], null: false
+
```
gigxz, line 51: Can we add a data source here? We may want to display that to the user. We could also potentially use it to determine whether we can link to the client/enrollment in this HMIS.
```ruby
field :data_source, [Types::Application::DataSource], null: false # GrdaWarehouse::DataSource id and name, and maybe flag for is current data source
```

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+    # when they are looking at a list of candidate enrollments to choose for the referral's source enrollment.
+
+    # Why not use Types::HmisSchema::Enrollment?
+    # This type does NOT have object-level authorization. The user may be able to see possible source enrollments for
+    # a referral, even without permission to see full enrollments in the project.
+
+    field :id, ID, null: false
+
+    field :client_name, String, null: false
+
```
gigxz, line 31: need client_id

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+    def id
+      object.enrollment.id
+    end
+
+    def client_name
+      # Source enrollments could be from different data sources, so return the client name.
+      # TODO(#7591) - ensure that this works correctly
+      client = load_ar_association(object.enrollment, :client)
+      current_permission?(permission: :can_view_client_name, entity: client) ? client.brief_name : client.masked_name
```
gigxz: This will only resolve the name if the client is from the current data source. (That logic is encoded in current_permission?). Just calling that out, maybe add it as a comment, we can probably leave it for now

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Types
+  class HmisSchema::CeReferralSourceEnrollment < Types::BaseObject
+    # object is an OpenStruct with the following shape:
+    # {
+    #   enrollment: Hmis::Hud::Enrollment,
```
gigxz: Can we add a big call-out here stating that the Enrollment may be from another data source (HMIS or otherwise). Maybe I'm getting ahead of myself to another ticket, if you're doing that on another branch its all good

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,132 @@
...
+      @household_members ||= load_ar_association(household, :enrollments)
+    end
+
+    def exit
+      @exit ||= load_ar_association(object.enrollment, :exit)
+    end
+
+    def project
+      @project ||= load_ar_association(object.enrollment, :project)
+    end
```
gigxz: Are these instance variables necessary? I don't really see why they should be, since the data loader should handle caching those lookups.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: So, this implementation is saying: if CE is enabled, and the user has "can view prioritized client lists" in any project, then they should be able to resolve _any_ candidate across the system. That doesn't seem right to me, since we said explicitly that the "can view prioritized client lists" permission is scoped to project. (i.e. it was only supposed to grant access to viewing waitlists within the specified project)

drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
```diff
@@ -7,5 +7,15 @@ class CandidatePool < GrdaWarehouseBase
     self.table_name = 'ce_match_candidate_pools'
     has_many :candidates, class_name: 'Hmis::Ce::Match::Candidate', foreign_key: :candidate_pool_id, dependent: :destroy
     has_many :opportunities, class_name: 'Hmis::Ce::Opportunity', dependent: :restrict_with_exception
+
+    def relevant_form_definition_identifiers
+      expressions = [requirement_expression, priority_expression]
+
+      cded_keys = expressions.map do |expression|
+        expression.scan(/cde\.custom_assessment\.([a-zA-Z0-9_-]+)/).flatten
```
gigxz: I'd love to update this to use the existing parsing logic rather than scanning it with a regex. That way we don't need to worry about conflicting implementations. This would be blocked by https://github.com/greenriver/hmis-warehouse/pull/5521 going in.

The parsing happens something like 
```
calculator = Hmis::Ce::Match::CalculatorFactory.build
cded_keys = expression.map do |expression|
  fields = calculator.dependencies(expression) # ["client.age", "cde.custom_assessmenty.something", ... ]
  # use helpers on FieldMap class or CdeFieldMap  to filter/map the fields into the custom data element definition key
  keys
end.flatten.uniq
```

Here, ideally we could call something very direct like `Hmis::Ce::Match::CdeFieldMap.cdeds_referenced_by_expression(expression: x)` (which would also handle any logic surrounding the CDED scope

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
martha: I can remove them, I think I was just following from the pattern in `Types::HmisSchema::Enrollment`

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -8,14 +8,45 @@
...
+      # In the future, we may want to limit this scope (possibly based on a configuration),
+      # but ideally this scope should align with the scope of enrollments used for candidate pool membership evaluation.
+      # TODO(#7671) - adjust this to fetch all enrollments from all clients
+
+      # For now, filter enrollments to those viewable by the current user.
+      # In the future we may need to update this to support resolving enrollments that are NOT viewable to the current user.
+      base_scope = object.client.enrollments.viewable_by(current_user)
+
+      # float those with a relevant assessment to the top.
+      with_assessment_ids = base_scope.joins(custom_assessments: :definition).
```
martha: I ran out of time for improving the search functionality here. I'd like to pair, or just merge as is and improve later.

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -8,14 +8,45 @@
...
     field :client, HmisSchema::Client, null: true, description: 'Null if the user lacks permission to view the client'
     field :priority_score, Integer, null: false
+    field :enrollments, HmisSchema::CeReferralSourceEnrollment.array_page_type, null: false
 
     def client
       load_ar_scope(scope: Hmis::Hud::Client.viewable_by(current_user), id: object.client_id)
     end
+
+    def enrollments # not for batch
+      form_definition_identifiers = object.candidate_pool.relevant_form_definition_identifiers
```
gigxz, line 24:
```suggestion
      # Find any "relevant" Forms, meaning forms that collect Custom Data Elements that are used for eligibility or prioritization of this candidate pool. There may not be any associated forms, if eligibility is determined by other factors.
      form_definition_identifiers = object.candidate_pool.relevant_form_definition_identifiers
```

drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
martha: Thanks Gig, pairing was helpful. Here's what I came up with https://github.com/greenriver/hmis-warehouse/pull/5525/commits/273db12d41b8aa9ab6172ea9c1ede23aaf4819d5

Following up from our pairing conversation about whether or not to instantiate the FieldMap, here I have done so - it seems nice to be able to use the cached cded lookup, but let me know if you don't think that is appropriate

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
gigxz: for sorting, as discussed I think it makes sense only to pull up the enrollments where the _most recent_ matching-related assessment(s) were performed. I think this should work:
```rails
# Source Enrollment IDs that should be prioritized because they are the enrollment(s) where the most recently updated Eligibility/Prioritization Assessment was taken for this client
prioritized_enrollment_ids = client.custom_assessments.with_form_definition_identifier(form_definition_identifiers).preload(:definition, :enrollment).
  group_by { |a| a.definition.identifier }.
  transform_values do |assessments|
  most_recent = assessments.max_by(&:DateUpdated)
  most_recent.enrollment.id
end.values

prioritized = base_scope.where(id: prioritized_enrollment_ids).sort_by_option(:most_recent)
others = base_scope.where.not(id: prioritized_enrollment_ids).sort_by_option(:most_recent)
```

drivers/hmis/app/models/hmis/ce/match/field_type.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+module Hmis::Ce::Match::FieldType
+  CDE = 'cde'
+  CLIENT = 'client'
```
gigxz: nit: these feel like constants that can be encapsulated in the FieldMap class, rather than their own class

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -0,0 +1,154 @@
...
+      end.map do |enrollment|
+        current_permission?(permission: :can_view_client_name, entity: object.enrollment) ? enrollment.client.brief_name : enrollment.client.masked_name
+      end
+    end
+
+    def assessments
+      # Returns the most recent assessments relevant to the opportunity, grouped by form definition.
+      # Even though definition_identifiers could refer to any form type, we are hard-coding the assumption that only custom assessments will be used.
+      assessments = object.enrollment.custom_assessments.not_in_progress.
+        joins(:definition).where(definition: { identifier: object.definition_identifiers })
```
gigxz, line 129: N+1 here, but I think we are resolving small enough page sizes that maybe its not worth optimizing yet

gigxz (approved):
looks good!

drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
gigxz: it looks good to me! thanks

drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
```diff
@@ -7,5 +7,30 @@ class CandidatePool < GrdaWarehouseBase
...
+        fields = calculator.dependencies(expression)
+
+        fields.map do |field|
+          # Use the FieldMap to map each field to its type, and skip if it isn't CDE
+          field_type, resolved_field = Hmis::Ce::Match::FieldMap.field_type_for(field)
+          next unless field_type == Hmis::Ce::Match::FieldType::CDE
+
+          resolved_field
+        end.uniq
+      end.flatten.uniq
```
gigxz:
```suggestion
      end.flatten.compact.uniq
```
This needs a compact to drop nils, or else it will throw later on

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
martha: Thanks!

PR #5527, opened 2025-06-28
Fix join hh dupe assignment bug

martha:
## _Merging this PR_
- ~targets release-169 as it is a fix for work currently on QA~ targets release-170, since this hasn't been encountered in a client env

## Description
The [previously merged PR](https://github.com/greenriver/hmis-warehouse/pull/5523) for the linked ticket updates a household's staff assignment when the household is fully joined into a different household. However, while QAing, I noticed this could result in "duplicate key" errors if the receiving household already had an identical staff assignment. This PR fixes that bug.

## Type of change
Bug fix


drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -98,5 +95,31 @@ def resolve(receiving_household_id:, joining_enrollment_inputs:)
...
+        key = [donor_assignment.user_id, donor_assignment.hmis_staff_assignment_relationship_id]
+        existing_keys.include?(key)
+      end
+
+      ids_to_delete = assignments_to_delete.map(&:id)
+      Hmis::StaffAssignment.where(id: ids_to_delete).delete_all if ids_to_delete.any?
+
+      ids_to_update = assignments_to_update.map(&:id)
+      Hmis::StaffAssignment.where(id: ids_to_update).update_all(household_id: receiving_household_id) if ids_to_update.any?
+    end
```
gigxz: Nit: the partition thing feels like a bit more complexity than we need here, I would err on the side of simpler code even though it's less optimized. We don't expect many assignments to exist per household.

```ruby
      # Iterate through the donor assignments and either update or destroy them
      donor_assignments.each do |assignment|
        key = [assignment.user_id, assignment.hmis_staff_assignment_relationship_id]
        if existing_keys.include?(key)
          assignment.destroy!
        else
          assignment.update!(household_id: receiving_household_id)
        end
      end
```

gigxz (approved):

drivers/hmis/app/graphql/mutations/join_household.rb
martha: Sounds good, thanks! I propose using `really_destroy!`, to circumvent the `acts_as_paranoid` gem. We use soft-deleted assignment records to show the assignment history, but in this case I think displaying the duplicate record as part of the history would only be confusing. What do you think?

gigxz: Yeah I agree with that, probably worth a comment too explaining the rational. Thanks!

PR #5529, opened 2025-06-29
Support CE Eligibility criteria based on open enrollment and referral project types

gigxz:
## _Merging this PR_

## Description
https://github.com/open-path/Green-River/issues/7835

This PR adds support for 3 new fields in Match Expressions: `open_enrollment_project_types`, `open_referral_project_types`, and `open_enrollment_project_types_excluding_incomplete`.

These support CE eligibility based on whether a client is or isn't currently enrolled in certain project types, and if they have open referrals to certain project types.

**Future improvements:**
* Likely need performance improvements. If a candidate pool requires an open enrollment in a certain project type, that could significantly reduce the client scope, so it may be worth adding an arel representation of these fields, so they can be applied in the crude_eligibility_filter

**How to test:**
* Set up a match expression that uses open enrollments, e.g. `INCLUDES(open_enrollment_project_types, PROJECT_TYPE("CE"))`
  * rebuild candidate pools `HmisUtil::CeBuilder.build_candidate_pools(clients: optional_client_scope)`
  * view unit waitlist, ensure correct clients are included/excluded
* Set up a match expression that uses open referrals, e.g. `EXCLUDES(open_referral_project_types, PROJECT_TYPE("PH_PSH")) AND EXCLUDES(open_referral_project_types, PROJECT_TYPE("PH_PH")) AND EXCLUDES(open_referral_project_types, PROJECT_TYPE("PH_OPH")) AND EXCLUDES(open_referral_project_types, PROJECT_TYPE("PH_RRH"))`
  * rebuild candidate pools `HmisUtil::CeBuilder.build_candidate_pools(clients: optional_client_scope)`
  * view unit waitlist, ensure correct clients are included/excluded

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/match/client_field_map.rb
```diff
@@ -1,6 +1,8 @@
 # frozen_string_literal: true
 
-# resolve fields from the client table
+# resolve fields from the client, enrollment, and project table
+#
+# - Question: should we move Enrollment/Project/Referral-related fields to another FieldMap? They require additional joins
```
ttoomey: Perhaps? I think I'd be inclined to keep them in the same resolver for now. Maybe the shape of an additional resolvers be clear from the full list of fields.

drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb
```diff
@@ -219,4 +254,48 @@ def create_assessment_with_cde(client, cde_values)
...
     end
   end
+
+  context 'when evaluating enrollment-based policies' do
+    include_context 'with enrolled test clients'
+
+    describe 'project-type-based filtering' do
+      describe 'when requiring ANY open enrollment to be in a CE (14) project' do
+        # Requirement: must have open enrollment in Coordinated Entry (14) project type
+        let(:requirement_expression) { 'ANY(open_enrollment_project_types, project_type, project_type = 14)' }
```
ttoomey: This is out of scope here but I wonder if we should try and head off magic numbers in expressions by supporting lookup functions, similar to hud lists.
```diff
- 'ANY(open_enrollment_project_types, project_type, project_type = 14)' 
+ 'ANY(open_enrollment_project_types, project_type, project_type = hud_project_type('CE') )' 
```

gigxz: Cool, that was one of my questions to discuss on https://github.com/open-path/Green-River/issues/7835#issuecomment-3017237995, I will think on that syntax idea

drivers/hmis/app/models/hmis/ce/match/calculator_factory.rb
```diff
@@ -8,15 +8,20 @@ module CalculatorFactory
     def self.build
       calculator = Dentaku::Calculator.new(case_sensitive: true) # CDED keys are case sensitive, so Dentaku expressions should be too
       calculator.add_function(
-        :includes,
+        :INCLUDES,
```
gigxz, line 11: Made these uppercase since the calculator is now case sensitive (See above note)

lib/util/hud_utility_2026.rb
```diff
@@ -203,6 +203,32 @@ def project_type_group_titles
     }.freeze
   end
 
+  # Project Type keys used in HMIS GraphQL API and CE Match expressions
+  def hmis_project_type_keys
```
gigxz, line 207: I hate to introduce yet _another_ project type mapping. This one is already in use in the GraphQL API, so its not totally new, but I moved it to a shared location so we could use it in the CE Match Expressions as well.

An alternative would be for the CE Match Expressions to use the other keys we use in the warehouse like `:psh`,`:rrh`, `:sh` (see above). I would be OK with that, too, but I figured that using the Graphql Enum keys might be helpful once we get to building the frontend CE Expression Rule builder, since the frontend is already aware of these "keys" from the API.

Definitely open to other ideas here!

The proposeed usage in the Expressions for lookup is `PROJECT_TYPE("PH_PSH")`.

drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb
```diff
@@ -48,6 +48,76 @@ def generate_candidates(pool, clients)
     let(:adult_clients) { clients.where.not(id: client_minor_non_veteran.id) }
   end
```
gigxz, line 53: To test this locally you need to comment out the `skip` and replace `client_id: source_client_id,` with `client_id: client.id,` in the match engine

drivers/hmis/app/models/hmis/ce/match/sql_expression_translator.rb
```diff
@@ -11,7 +11,12 @@ class SqlExpressionTranslator
 
     def self.call(expression, field_map)
       calculator = Hmis::Ce::Match::CalculatorFactory.build
-      ast = calculator.ast(expression)
+      begin
+        ast = calculator.ast(expression)
+      rescue Dentaku::Error => e
+        err_with_context = "Error parsing expression '#{expression}': #{e.message}"
+        raise $ERROR_INFO, err_with_context, $ERROR_INFO.backtrace
```
ttoomey: Do we have to use the global ERROR_INFO here?
```suggestion
        raise e, err_with_context, e.backtrace
```

drivers/hmis/lib/hmis_util/ce_builder.rb
```diff
@@ -0,0 +1,23 @@
...
+    def self.build_candidate_pools(clients: nil)
+      # Build candidate pools
+      opportunities = Hmis::Ce::Opportunity.active
+      Hmis::Ce::Match::CandidatePoolBuilder.new(opportunities).perform
+
+      # Delete orphaned pools (to clear out any misconfigured match expressions)
+      Hmis::Ce::Match::CandidatePool.where.missing(:opportunities).find_each(&:destroy!)
+
+      # Run the match engine for each candidate pool
+      clients ||= Hmis::Hud::Client.hmis
```
ttoomey:
```suggestion
      # TODO use destinations after #7671 is implemented
      # clients ||= GrdaWarehouse::Hud::Client.destination
      clients ||= Hmis::Hud::Client.hmis
```

drivers/hmis/app/models/hmis/ce/match/client_field_map.rb
```diff
@@ -35,7 +35,36 @@ def all
...
           end,
         },
-        # Add support for more client fields here...
+        # Array of Project Types at which the Client has an open Enrollment, including WIP enrollments.
+        open_enrollment_project_types: {
+          instance_value: ->(c) do
+            Hmis::Hud::Enrollment.joins(client: :warehouse_client_source).
+              where(warehouse_clients: { destination_id: c.id }).
+              open_including_wip.
+              joins(:project).
```
ttoomey, line 44: simple performance optimization to select distinct project types (a client might have many enrollments). Would also apply to the other project type queries
```suggestion
              joins(:project).
              distinct.
```

drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb
```diff
@@ -215,12 +284,89 @@ def create_assessment_with_cde(client, cde_values)
...
+
+        it 'includes clients with open enrollments at the correct project type' do
+          results = generate_candidates(pool, clients)
+          expect(results.map(&:client_id).sort).to eq([client_enrolled_in_ce.destination_client.id])
+        end
+      end
+
+      describe 'when requiring no open enrollments in PH project types' do
+        # Requirement: must NOT have open enrollment in any Permanent Housing project (3, 9, 10, 13)
+        let(:requirement_expression) { 'EXCLUDES(open_enrollment_project_types, PROJECT_TYPE("PH_PSH")) AND EXCLUDES(open_enrollment_project_types, PROJECT_TYPE("PH_PH")) AND EXCLUDES(open_enrollment_project_types, PROJECT_TYPE("PH_OPH")) AND EXCLUDES(open_enrollment_project_types, PROJECT_TYPE("PH_RRH"))' }
```
ttoomey, line 348: this is a bit verbose but should work. Maybe in the future we might want something like an intersection method and an alias to all these project types. I guess we can see what the UI looks like.

A half-baked idea:
`NOT INTERSECTS(open_enrollment_project_types, PH_PROJECT_TYPES())`

ttoomey (approved):
Looks great!! Thanks for adding the lookup project type utility, that seems much safer and easier to manage.

drivers/hmis/lib/hmis_util/ce_builder.rb
gigxz: Thanks, my plan is to merge https://github.com/greenriver/hmis-warehouse/pull/5531 into the release branch, then I will resolve conflicts on this branch.

PR #5530, opened 2025-06-30
Superset Env Take Two

yakloinsteak:
- **Use new tenant var that will always agree with the terraform client/tenant name**

## _Merging this PR_

## Description
Some tenants (clients) are labeled inconsistently between terraform and ansible. This variable change addresses that.

## Type of change
bugfix

### DIFF
--- a/config/deploy/docker/lib/aws_sdk_helpers.rb
+++ b/config/deploy/docker/lib/aws_sdk_helpers.rb
@@ -130,8 +130,8 @@ module AwsSdkHelpers
 
     def self.superset_env_for_warehouse
       names = [
-        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('CLIENT')}/app/superset-sync-jobs/SUPERSET_ADMIN_PASS",
-        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('CLIENT')}/app/superset-sync-jobs/SUPERSET_ADMIN_USER",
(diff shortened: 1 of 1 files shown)


config/deploy/docker/lib/aws_sdk_helpers.rb
```diff
@@ -130,8 +130,8 @@ def self.get_secret(secret_arn)
 
     def self.superset_env_for_warehouse
       names = [
-        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('CLIENT')}/app/superset-sync-jobs/SUPERSET_ADMIN_PASS",
-        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('CLIENT')}/app/superset-sync-jobs/SUPERSET_ADMIN_USER",
+        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('TF_TENANT')}/app/superset-sync-jobs/SUPERSET_ADMIN_PASS",
+        "/#{ENV.fetch('RAILS_ENV')}/tenant/#{ENV.fetch('TF_TENANT')}/app/superset-sync-jobs/SUPERSET_ADMIN_USER",
```
eanders, line 134: Is this ENV already on all installations, or should we `ENV.fetch('TF_TENANT', ENV.fetch('CLIENT'))`?

eanders (approved):
Assuming we know the ENV is present and correct, looks good to me!

yakloinsteak:
I've pushed out the variables to all environments, so I think we're okay.

PR #5531, opened 2025-06-30
Use destination clients on CE waitlists

martha:
## _Merging this PR_

## Description
- Use destination clients, instead of source clients, in CE Match Engine.
  - Add `ce_client_proxies` table and `Hmis::Ce::ClientProxy` model, and update candidate reference to use this new type
  - Update the match engine to save client proxy records
  - Update `Hmis::GrdaWarehouse::Hud::ClientExtension` with new associations:
    - `ce_client_proxies` to carry over destroy logic. Client proxies and match candidates should get destroyed when a destination client is destroyed.
    - `ce_match_candidates` for determining the eligible opportunities for this client
    - `hmis_source_clients` for returning source clients as `Hmis::Hud::Client` model, which avoids n+1 issues vs. using the `as_hmis` helper.
  - Add and update spec 

## Type of change
New feature

- [ ] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) - I considered writing an ADR for this change. "Use destination client for CE waitlists" I decided against it after checking the existing ADR folder and finding that it's so far been populated with much higher-level decisions and this seems quite small in comparison. Maybe CE-in-HMIS warrants an ADR in general, but that does not seem very high priority now.

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
@@ -20,15 +20,18 @@ module Types
     end
 
     def client_name
-      # Current permission logic: For this destination client, are there any source clients whose names you can view?
-      # If so, show the *destination* client's name.
-      # In the future, we want to check permissions more broadly across data sources.
-      # (viewable_by scope only accounts for permissions in current data source.)
+      # For this destination client, are there any source clients whose names you can view? If so, return the *destination* client's name.
+      # In the future, we want to check permissions more broadly across data sources. (viewable_by scope only accounts for permissions in current data source.)
 
-      if named_source_clients.any?
-        # Use name directly, instead of PII provider.
-        # The user has permission to see this client's name in some source system.
-        # But we don't want to display the source name, because it could be inaccurate.
+      # Iterate through source clients avoids n+1 issues with viewable_by scope
+      can_view_any_source_client_name = source_clients.filter do |client|
+        current_permission?(permission: :can_view_clients, entity: client) && current_permission?(permission: :can_view_client_name, entity: client)
+      end.any?
+
+      if can_view_any_source_client_name
+        # Return the destination client's first and last names directly, instead of using the client PII provider.
+        # The user may not have permission to view the destination client name, but we know they have permission to view the client's name in some source system.
+        # We don't want to return the source client name, because it's less accurate. This Candidate represents a Destination Client, so return the destination name.
         "#{destination_client.FirstName} #{destination_client.LastName}"
       else
         "Candidate #{object.id}"
@@ -80,15 +83,7 @@ module Types
     end
 
     def source_clients
-      source_clients = load_ar_association(destination_client, :source_clients) # These are source clients, but they are GrdaWarehouse::Hud::Client objects
-      # todo @martha - this is an n+1 issue after all, because we need the Hmis::Hud::Client object, not GrdaWarehouse::Hud::Client
-      Hmis::Hud::Client.where(id: source_clients.pluck(:id))
-    end
-
-    def named_source_clients
-      source_clients.filter do |client|
-        current_permission?(permission: :can_view_clients, entity: client) && current_permission?(permission: :can_view_client_name, entity: client)
-      end
+      load_ar_association(destination_client, :hmis_source_clients)
     end
   end
 end
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
@@ -20,22 +20,15 @@ module Types
     end
 
     def client_name
-      # For this destination client, are there any source clients whose names you can view? If so, return the *destination* client's name.
+      # For this destination client, are there any source clients whose names you can view? If so, return the first viewable name.
       # In the future, we want to check permissions more broadly across data sources. (viewable_by scope only accounts for permissions in current data source.)
 
       # Iterate through source clients avoids n+1 issues with viewable_by scope
-      can_view_any_source_client_name = source_clients.filter do |client|
+      first_viewable_name = source_clients.sort_by(&:id).find do |client|
         current_permission?(permission: :can_view_clients, entity: client) && current_permission?(permission: :can_view_client_name, entity: client)
-      end.any?
+      end&.brief_name
 
-      if can_view_any_source_client_name
-        # Return the destination client's first and last names directly, instead of using the client PII provider.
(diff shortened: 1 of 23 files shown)


drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
```diff
@@ -14,6 +16,12 @@ module ClientExtension
       has_many :hmis_custom_assessments, through: :enrollments
       has_many :hmis_source_custom_assessments, through: :source_enrollments, source: :hmis_custom_assessments
       has_many :custom_services, through: :source_enrollments
+      has_many :ce_client_proxies, as: :client, class_name: 'Hmis::Ce::ClientProxy', dependent: :destroy
+      has_many :ce_match_candidates, class_name: 'Hmis::Ce::Match::Candidate', through: :ce_client_proxies
+
+      # HMIS source clients represented as Hmis::Hud::Client
+      # (Normal source_clients are represented as GrdaWarehouse::Hud::Client)
+      has_many :hmis_source_clients, -> { hmis }, class_name: 'Hmis::Hud::Client', through: :warehouse_client_destination, source: :source
```
martha: Does adding ` -> { hmis }` scope here make sense?
- for now (mvp) I think it won't have any negative effect, because in the CeCandidate schema object, we filter source enrollments to only those the current data source, anyway
- I added it because it felt weird to return non-hmis clients as `Hmis::Hud::Client`s.

drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -23,17 +23,18 @@ class Candidate < GrdaWarehouseBase
 
       scope = opportunity.candidate_pool.candidates
 
+      # TODO(#7835) - descoped for now since this is getting reworked in #7835
       # do we need to allow a referral to be re-started for the same client/opportunity?
```
martha: I've removed this for now since my understanding is that it's getting reworked soon anyway.

Also, I think that the answer to the question on line 27 "do we need to allow a referral to be re-started?" is **yes**, at least for mvp, because we don't currently have any way of cancelling a referral that was started by accident, aside from rejecting the referral.

drivers/hmis/spec/factories/hmis/ce/match_candidates.rb
```diff
@@ -2,8 +2,11 @@
 
 FactoryBot.define do
   factory :hmis_ce_match_candidate, class: 'Hmis::Ce::Match::Candidate' do
+    transient do
+      client { create :destination_client }
```
martha, line 6: I think it would be better for this to use `build` instead of `create`, but gemini and I could not quite figure it out.

drivers/hmis/app/jobs/hmis/match_candidates_job.rb
```diff
@@ -25,14 +25,16 @@ def perform(**args)
...
+
       opportunities = opportunity_ids ? Hmis::Ce::Opportunity.where(id: opportunity_ids) : Hmis::Ce::Opportunity.active
       log("Building candidate pools for #{opportunities.count} opportunities")
       Hmis::Ce::Match::CandidatePoolBuilder.new(opportunities).perform
 
       # Find candidate pools for these opportunities
       candidate_pool_ids = opportunities.reload.pluck(:candidate_pool_id).compact.uniq
       candidate_pool_scope = Hmis::Ce::Match::CandidatePool.where(id: candidate_pool_ids)
-      client_scope = Hmis::Hud::Client.hmis
+      client_scope = ::GrdaWarehouse::Hud::Client.where(data_source_id: ::GrdaWarehouse::DataSource.destination_data_source_ids)
```
ttoomey: I think we can use the existing scope
```suggestion
      client_scope = ::GrdaWarehouse::Hud::Client.destination
```

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -10,33 +10,51 @@ module Types
...
     field :enrollments, HmisSchema::CeReferralSourceEnrollment.array_page_type, null: false
 
-    def client
-      load_ar_scope(scope: Hmis::Hud::Client.viewable_by(current_user), id: object.client_id)
+    def destination_client_id
+      destination_client.id
+    end
+
+    def client_name
+      # For this destination client, are there any source clients whose names you can view? If so, return the *destination* client's name.
```
ttoomey: Are we sure this is right? The destination name could originate from a source that is not visible. Perhaps it would be safer to either pick one of the visible source names or return an array of unique visible names

ttoomey (approved):
Looks great to me, thanks! 🙌

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
gigxz: Martha and I discussed this; we landed on showing the destination record's Name, to ensure consistency (i.e. waitlist candidates always appearing with the same name). Having access to view the name of any of the source clients, in combination with `can_view_prioritized_client_lists`, is what gives you access to view that _candidate's_ name. I felt like it's important to present the candidate in a way that's accurate to the underlying record that's being evaluated for eligibility/prioritization. 

HOWEVER.. now that I'm writing it out, I suppose there is a possible scenario where someone has a fake/anonymized name and a real name, it really would be safer to err on the side of hiding the name in that case. The `CeReferralSourceEnrollment` type only reveals the name if the user otherwise has access, so this is the only place we're leaking that. I'll update it to the more conservative approach you suggest.

* Destination: `John Smith` (use has access to view this CeCandidate)
  * Source 1: `Anon Client` (user has access to view this client, including their name)
  * Source 2: `John Smith` (user does not have access to this client)

drivers/hmis/app/models/hmis/ce/match/candidate.rb
gigxz: Agreed, I made a commit to just remove this commented-out code. The new approach in this pR https://github.com/open-path/Green-River/issues/7835 doesn't touch this scope.

drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
gigxz: Updated https://github.com/greenriver/hmis-warehouse/pull/5531/commits/69e31e870de187fb7ac7412ba29785205ca916b8

drivers/hmis/extensions/grda_warehouse/hud/client_extension.rb
gigxz: I think that makes sense

gigxz (approved):
nice work!

PR #5532, opened 2025-06-30
Revert candidate pool to store source client IDs

gigxz:
## _Merging this PR_

## Description
Follow-up to https://github.com/greenriver/hmis-warehouse/pull/5521
That PR made the candidate pool reference destination clients, but we are not quite ready for that change. Map back to source client before importing, and temporarily disable tests.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/ce/match/engine.rb
+++ b/drivers/hmis/app/models/hmis/ce/match/engine.rb
@@ -15,7 +15,9 @@ module Hmis::Ce::Match
     # 2. Evaluate the eligibility requirement expression against each matched client. We expect all expression variables to be defined.
     def call(pool, clients)
       validate_clients_parameter!(clients)
-      # TODO: remove this in #7671
+      # TODO: remove this in #7671. store mapping of { destination_id => hmis_source_client_id }
+      destination_to_source_map = build_destination_to_source_map(clients)
+      # TODO: remove this in #7671. translate `clients` scope of source clients into a scope of destination clients
       clients = translate_source_to_destination_scope(clients)
 
       eligibility_evaluator = ClientExpressionEvaluator.new(pool.requirement_expression, field_map)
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/models/hmis/ce/match/engine.rb
```diff
@@ -15,7 +15,9 @@ def self.call(...)
     # 2. Evaluate the eligibility requirement expression against each matched client. We expect all expression variables to be defined.
     def call(pool, clients)
       validate_clients_parameter!(clients)
-      # TODO: remove this in #7671
+      # TODO: remove this in #7671. store mapping of { destination_id => hmis_source_client_id }
```
martha, line 18: I'm confused about the mapping suggested in this comment -- My understanding is, in #7671,
- we want the engine to continue to receive a scope of source clients
- it should translate that into a scope of dest clients (per comment on line 20)
- client proxies should be instantiated with dest client IDs
- why do we need to map back from the dest to hmis source client ID?

martha (approved):
Looks great, thank you. My comment is out of scope for this change, we can pick up the discussion on the ticket if you prefer

drivers/hmis/app/models/hmis/ce/match/engine.rb
martha: Oh, sorry, I was confused because I thought "store mapping of { destination_id => hmis_source_client_id }" was an instruction for the todo associated with #7671. But I guess it isn't, it is a description of what line 19 is doing currently. Same with comment on line 20. Both comments are _both_ adding a TODO for future removal and describing current behavior.

gigxz: Yes its a description of what its doing, in both cases - sorry for the confusion!

gigxz:
@ttoomey I'm going to go ahead and merge this to unblock other work,but feel free to leave a review anyway. I discussed with @martha that she will re-enable those tests and make the necessary changes to support https://github.com/open-path/Green-River/issues/7671. thanks!

PR #5533, opened 2025-06-30
ACL Audit History

dtgreiner:
## _Merging this PR_

## Description

Add an audit history on individual items for:
- Collections
- Roles
- User Groups
- Access Controls

The Access Controls also include a history for all Access Controls

These historys are exportable as CSV files.

## Type of change
New feature


app/controllers/admin/access_control_audits_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+
+module Admin
+  class AccessControlAuditsController < ::ApplicationController
+    before_action :require_can_audit_users!
+    before_action :set_access_control
+    include AuditHistory
+
+    def show
+      @history = Audit::Versions.new(@access_control, access_control_component_config)
+      @pagy, @versions = pagy_array(@history.version_scope)
```
eanders: Maybe we need this here, but we've been pivoting to using `render_paginated_list` in the view instead of calling pagy in the controller.  It feels like it would be a fairly light lift to update `render_paginated_list` to use `pagy_array` if that is necessary.

app/controllers/admin/access_control_audits_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    end
+
+    def export
+      @history = Audit::Versions.new(@access_control, access_control_component_config)
+      @versions = @history.version_scope
+
+      respond_to do |format|
+        format.csv do
+          send_data generate_audit_csv(@versions, @history), filename: "access-control-#{@access_control.id}-component-history-#{Date.current}.csv"
```
eanders:
```suggestion
          send_data generate_audit_csv(@versions, @history), filename: "access-control-#{@access_control.id}-component-history-#{Date.current.to_fs(:db)}.csv"
```

app/controllers/admin/access_controls_controller.rb
```diff
@@ -51,6 +54,23 @@ def assign
     redirect_to action: :index
   end
 
+  def audits
+    @pagy, @data = pagy_array(data)
```
eanders: Same here, if we can use `render_paginated_list`, it is a nice way to encapsulate the logic.

app/controllers/admin/access_controls_controller.rb
```diff
@@ -51,6 +54,23 @@ def assign
...
+
+  def export
+    respond_to do |format|
+      format.csv do
+        audit_history = []
+        histories.each_with_index do |history, index|
+          csv_data = generate_audit_csv(history.version_scope, history, include_headers: index == 0)
+          audit_history << csv_data
+        end
+        send_data audit_history.join, filename: "access-controls-component-history-#{Date.current}.csv"
```
eanders:
```suggestion
        send_data audit_history.join, filename: "access-controls-component-history-#{Date.current.to_fs(:db)}.csv"
```

app/controllers/admin/access_controls_controller.rb
```diff
@@ -77,4 +97,29 @@ def assign
   private def set_access_control
     @access_control = access_control_scope.find(params[:id].to_i)
   end
+
+  private def access_controls
+    @access_controls ||= AccessControl.all
```
eanders: I wonder if we should take this opportunity to add a `visible_to(user)` scope to `AccessControl` even if it just returns `all` for now.

app/controllers/admin/access_controls_controller.rb
```diff
@@ -77,4 +97,29 @@ def assign
...
+  end
+
+  private def histories
+    @histories ||= [].tap do |histories|
+      access_controls.each do |access_control|
+        history = Audit::Versions.new(access_control, access_control_component_config)
+        histories << history
+      end
+    end
```
eanders: Maybe?
```suggestion
    @histories ||= access_controls.map do |access_control|
        Audit::Versions.new(access_control, access_control_component_config)
    end
```

app/controllers/admin/access_controls_controller.rb
```diff
@@ -77,4 +97,29 @@ def assign
...
+  private def data
+    @data ||= histories.map do |history|
+      versions = history.version_scope
+      history.wrap_display_versions(versions).map do |version|
+        {
+          history: history,
+          version: version,
+        }
+      end
+    end.flatten(1)
```
eanders: Please double check, but I think this is just:

```suggestion
    @data ||= histories.flat_map do |history|
      versions = history.version_scope
      history.wrap_display_versions(versions).map do |version|
        {
          history: history,
          version: version,
        }
      end
    end
```

app/controllers/concerns/audit_history.rb
```diff
@@ -0,0 +1,100 @@
...
+      include_self: false,
+      excluded_fields: ['updated_at'],
+    }
+  end
+
+  def collection_audit_config
+    {
+      referenced_models: [
+        { class: AccessControl, association: :access_controls },
+        { class: GrdaWarehouse::GroupViewableEntity, association: :group_viewable_entities },
```
eanders, line 35: I don't know if it's worth cleaning up, but you shouldn't need the classes here, you can use something like: 
```
self.class.reflect_on_association(:group_viewable_entities).klass
```
On an instance of a `Collection` to get back the `GrdaWarehouse::GroupViewableEntity` class.

app/controllers/concerns/audit_history.rb
```diff
@@ -0,0 +1,100 @@
...
+        { class: AccessControl, association: :access_controls },
+        { class: UserGroupMember, association: :user_group_members },
+      ],
+      excluded_fields: ['updated_at'],
+    }
+  end
+
+  # Generate CSV for standard audit exports
+  def generate_audit_csv(versions, history, include_headers: true)
+    require 'csv'
```
eanders: The `csv` gem is not set to `require: false` in the Gemfile, so I'm 99% sure you don't need this.
```suggestion
```

app/controllers/concerns/audit_history.rb
```diff
@@ -0,0 +1,100 @@
...
+
+    CSV.generate(headers: include_headers) do |csv|
+      csv << ['Date Changed', 'Editor', 'Entity Type', 'Entity Name', 'Changed Entity Type', 'Changed Entity Name', 'Event', 'Changes'] if include_headers
```
eanders: I think `['Date Changed', 'Editor', 'Entity Type', 'Entity Name', 'Changed Entity Type', 'Changed Entity Name', 'Event', 'Changes']` should probably be captured in a method or constant to make them easier to find later.

app/models/audit/display_item.rb
```diff
@@ -0,0 +1,129 @@
...
+
+  def initialize(version, users_by_id, excluded_fields = [])
+    raise ArgumentError, "Expected GrPaperTrail::Version or GrdaWarehouse::Version, got #{version.class}" unless version.is_a?(GrPaperTrail::Version) || version.is_a?(GrdaWarehouse::Version)
+
+    @version = version
+    @excluded_fields = excluded_fields
+
+    true_username = compute_username(users_by_id, version.clean_true_user_id)
+    username = compute_username(users_by_id, version.clean_user_id)
+    @username = true_username || username || version.whodunnit.presence
```
eanders, line 24: Probably want to replace blank strings too.
```suggestion
    @username = true_username.presence || username.presence || version.whodunnit
```

eanders: Actually, it looks like `compute_username` handles the presence, so you can probably ignore this.

app/models/audit/display_item.rb
```diff
@@ -0,0 +1,129 @@
...
+    @impersonating = username if true_username && username != true_username
```
eanders: Should this check happen before calling `compute_username` so you end up checking ids instead of names?  You could potentially have user 2 "Fake Fakerson" and user 3 "Fake Fakerson" and it wouldn't recognize the impersonation I think.

app/models/audit/display_item.rb
```diff
@@ -0,0 +1,129 @@
...
+      begin
+        record = klass.with_deleted.find_by(id: version.item_id)
+        if record&.entity_type
+          record.entity_type
+        else
+          version.item_type
+        end
+      rescue StandardError
+        version.item_type
+      end
```
eanders: I wonder if it makes sense to move this logic into `GrdaWarehouse::GroupViewableEntity` for use later.  Something like:
```
GrdaWarehouse::GroupViewableEntity.item_type(version)
```

```
def self.item_type(item)
  record = with_deleted.find_by(id: item.item_id)
  if record&.entity_type
    record.entity_type
  else
    item.item_type
  end
rescue StandardError
 item.item_type  
end
```

Not so sure I like the `rescue StandardError` is there a more specific error we could be rescuing?

app/models/audit/display_item.rb
```diff
@@ -0,0 +1,129 @@
...
+      ["Modified #{entity_display_name}"]
+    end
+  end
+
+  def render_changed_value(_field, value)
+    return 'nil' if value.nil?
+    return 'true' if value == true
+    return 'false' if value == false
+    return value.to_s if value.is_a?(Numeric)
+    return value.to_s if value.is_a?(String)
```
eanders: Maybe?
```suggestion
    return value.to_s if value == true || value == false || value.is_a?(Numeric) || value.is_a?(String) 
```

eanders: Actually, I think you can just drop all lines after `return 'nil' if value.nil?` (and the `return` from the last line) as the default is to call `value.to_s`

app/models/audit/versions.rb
```diff
@@ -0,0 +1,211 @@
...
+
+    gr_paper_trail_ids = []
+    grda_warehouse_ids = []
+
+    config[:related_models].each do |model_config|
+      model_class = model_config[:class]
+      association_name = model_config[:association]
+
+      # Get the related record
+      related_record = record.send(association_name)
```
eanders: Does this need a `with_deleted`?

app/models/audit/versions.rb
```diff
@@ -0,0 +1,211 @@
...
+      model_class = model_config[:class]
+      parent_association = model_config[:parent_association]
+      nested_association = model_config[:nested_association]
+
+      # Get the parent record
+      parent_record = record.send(parent_association)
+      next unless parent_record
+
+      # Get the IDs of nested records
+      nested_record_ids = parent_record.send(nested_association).pluck(:id)
```
eanders: Maybe a `with_deleted` here as well?

app/models/concerns/history_descriptions.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def render_changed_value(_field, value)
+      return 'nil' if value.nil?
+      return 'true' if value == true
+      return 'false' if value == false
+      return value.to_s if value.is_a?(Numeric)
+      return value.to_s if value.is_a?(String)
+
+      value.to_s
```
eanders: This looks similar to the above method.  I think:
```suggestion
      return 'nil' if value.nil?

      value.to_s
```
would cover the same scenarios

app/models/grda_warehouse/group_viewable_entity.rb
```diff
@@ -21,5 +23,84 @@ class GroupViewableEntity < GrdaWarehouseBase
...
+      when 'GrdaWarehouse::Lookups::CocCode'
+        "CoC Code: #{entity&.coc_code || entity_id}"
+      when 'GrdaWarehouse::Hud::Project'
+        "Project: #{entity&.ProjectName || entity_id}"
+      when 'GrdaWarehouse::Hud::Organization'
+        "Organization: #{entity&.OrganizationName || entity_id}"
+      when 'GrdaWarehouse::DataSource'
+        "Data Source: #{entity&.name || entity_id}"
+      when 'GrdaWarehouse::ProjectAccessGroup'
+        "Project Group: #{entity&.name || entity_id}"
```
eanders: I don't know if this language is correct, but I believe this version of project groups is used to govern access to projects while the other governs access to project groups.
```suggestion
        "Project Group for Project Access: #{entity&.name || entity_id}"
```

app/models/grda_warehouse/group_viewable_entity.rb
```diff
@@ -21,5 +23,84 @@ class GroupViewableEntity < GrdaWarehouseBase
...
+    end
+
+    def self.describe_changes(version, _changes, _excluded_fields = [])
+      # Get the entity name from the version's item or object_changes
+      entity_name = if version.item
+        get_entity_display_name(version.item.entity_type, version.item.entity_id, version.item.entity)
+      elsif version.object_changes&.dig('entity_id', 1)
+        # For create events, get the entity name from the new entity_id
+        entity_id = version.object_changes['entity_id'][1]
+        entity_type = version.object_changes['entity_type']&.last || version.object_changes['entity_type']&.first
```
eanders: This position based access is really hard to follow (why are we looking for the last item, but if it isn't there, taking the first?  Or why are we grabbing the id out of the second item?  If this is part of PaperTrail, let's add a comment with an example or two to indicate how this data is stored.  If this is something added in this PR can it be converted to hashes?  If not, let's add the comment.

app/models/grda_warehouse/version.rb
```diff
@@ -42,5 +60,45 @@ def whodunnit_impersonator_pattern
       # When impersonating a user, whodunnit is recorded as "<true_user> as <current_user>"
       /^(\d+) as (\d+)$/
     end
+
+    protected
+
+    # Extracted this method from the controller. It appears to be a fallback for old versions didn't include changes
```
eanders: If we need this, I think it would be on the user audit screens or audit reports.

app/models/user_group_member.rb
```diff
@@ -4,10 +4,63 @@
...
+  end
+
+  def self.render_changed_value(_field, value)
+    return 'nil' if value.nil?
+    return 'true' if value == true
+    return 'false' if value == false
+    return value.to_s if value.is_a?(Numeric)
+    return value.to_s if value.is_a?(String)
+
+    return value.to_s
```
eanders: same here

app/views/admin/access_controls/edit.haml
```diff
@@ -1,6 +1,11 @@
 = render 'breadcrumbs'
 
 %h1 Edit Access Control #{@access_control.name}
+.pull-right
```
eanders: Might be more "modern" to do something like:

```
.d-flex
  %h1 Edit Access Control #{@access_control.name}
  .ml-auto
    = link_to admin_access_control_audit_path(@access_control), class: 'btn btn-info mb-3' do
      %i.fa.fa-history
      %span.icon-eye
      History
```

app/views/admin/audits/shared/_layout.haml
```diff
@@ -0,0 +1,11 @@
...
+- title = "#{entity_type} Audit History"
+- content_for :title, title
+= render_display_title "#{entity_name} Audit History"
+
+.text-right.mb-3
+  = link_to "Export CSV", "#{request.path}/export.csv", class: "btn btn-outline-primary btn-sm"
+
+= render 'common/pagination_top', item_name: 'audit event'
+= render 'admin/audits/shared/list'
+= render 'common/pagination_bottom', item_name: 'audit event'
```
eanders: This would be handled by `render_paginated_list`

eanders (changes requested):
@dtgreiner I think this looks pretty good.  Left a bunch of notes, though I didn't manage to comment in all the copies of things.

app/models/access_control.rb
```diff
@@ -102,6 +105,10 @@ class AccessControl < ApplicationRecord
     where(id: ids)
   end
 
+  scope :visible_to, ->(_user) do
+    return all
```
eanders: Should be sufficient:

```suggestion
   all
```

app/models/grda_warehouse/group_viewable_entity.rb
```diff
@@ -21,5 +23,99 @@ class GroupViewableEntity < GrdaWarehouseBase
...
+      when 'GrdaWarehouse::ProjectGroup'
+        "Project Group: #{entity&.name || entity_id}"
+      else
+        entity&.name || "Entity: #{entity_id}"
+      end
+
+      "Collection: #{collection_name} - #{entity_display_name}"
+    end
+
+    def self.describe_changes(version, _changes, _excluded_fields = [])
```
eanders, line 62: Why does a new method have unused arguments?

app/models/grda_warehouse/group_viewable_entity.rb
```diff
@@ -21,5 +23,99 @@ class GroupViewableEntity < GrdaWarehouseBase
...
+        if version.object.present?
+          entity_id = version.object['entity_id']
+          entity_type = version.object['entity_type']
+        else
+          entity_id = version.object_changes['entity_id'][index]
+          entity_type = version.object_changes['entity_type'][index]
+        end
+        get_entity_display_name(entity_type, entity_id)
+      else
+        'Unknown Entity'
```
eanders, line 88: This is way cleaner and clearer, thanks!

eanders (approved):
Thanks for the cleanup @dtgreiner, dropped a couple of questions, but generally this looks good!

app/models/grda_warehouse/group_viewable_entity.rb
dtgreiner: This is being called from [Audit::DisplayItem](https://github.com/greenriver/hmis-warehouse/blob/d04219f6e55a0e4471add76c7c0c65423654c189/app/models/audit/display_item.rb#L46) which is dynamically calling the method. Most classes are using the version built into the [HistoryDescriptions](https://github.com/greenriver/hmis-warehouse/blob/d04219f6e55a0e4471add76c7c0c65423654c189/app/models/concerns/history_descriptions.rb#L14) module, but this one needed to be customized.

PR #5536, opened 2025-07-01
7634 candidate pool generation

ttoomey:
## _Merging this PR_

## Description
Introduce a unified change tracking system for CE eligibility processing both clients and candidate pools

* Enable frequent incremental processing for client and pool changes
* CE eligibility is continuously updated while maintaining daily full refresh
* Minimize expensive eligibility calculations by tracking dirty state at the record level

### Changes Made
See drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md for full documentation of the change tracking system

## Type of change
New feature


drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -70,21 +65,5 @@ def build_opportunity_for_unit(unit, template, candidate_pool_resolver)
       opportunity.candidate_pool = candidate_pool_resolver.candidate_pool_for_opportunity(opportunity: opportunity)
       opportunity
     end
-
-    def opportunity_ids_to_update(opportunity_scope:) # Find any opportunities that need to be updated
```
ttoomey, line 74: I think the change-tracking system should handle this now.

* Cron should re-process all pools every 24 hours already so we shouldn't need to check for stale pools here
* The BuildCandidatePoolsJob task should mark any new pools as dirty

drivers/hmis/app/models/hmis/ce/match/candidate_pool_builder.rb
```diff
@@ -56,12 +60,13 @@ def update_pools!(values)
       end
       result = Hmis::Ce::Match::CandidatePool.import(
         attrs,
-        on_duplicate_key_update: {
```
ttoomey, line 59: small bugfix: don't update `configuration_updated_at` if the row wasn't actually updated. Requires corresponding change to check for associated records when deleting orphan pools

ttoomey:
I think I need to do a little more testing on this, moving back to draft

db/warehouse/migrate/20250627132413_create_client_change_markers.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class CreateClientChangeMarkers < ActiveRecord::Migration[7.1]
+  def change
+    safety_assured do
+      remove_column :ce_match_candidate_pools, :configuration_updated_at, :date
```
ttoomey, line 6: I don't think this needed. We can use updated_at for 'when was this record last associated with an active opportunity' for orphan tracking

ttoomey:
I think this is ready for code review

app/models/grda_warehouse/tasks/identify_duplicates.rb
```diff
@@ -218,13 +224,18 @@ def match_existing!
...
         ClientCleanupJob.set(priority: 6).perform_later(batch.flatten)
+        post_process_clients(client_ids: batch.map(&:first))
       end
       Rails.logger.info '=== Completed match_existing! ==='
       return unless @run_post_processing
 
       GrdaWarehouse::Tasks::ServiceHistory::Add.new(force_sequential_processing: true).run!
     end
 
+    private def post_process_clients(client_ids:)
```
eanders, line 236:
```suggestion
    # Marks given clients as dirty for future re-processing for:
    # * CE
    private def post_process_clients(client_ids:)
```

drivers/hmis/app/jobs/hmis/ce/build_candidate_pools_job.rb
```diff
@@ -0,0 +1,38 @@
...
+    # @param opportunity_ids [Array<Integer>, nil] IDs of opportunities to process.
+    #   If nil, processes all active opportunities.
+    def perform(opportunity_ids: nil)
+      raise unless Hmis::Ce.configuration.enabled?
+
+      if opportunity_ids
+        opportunities = Hmis::Ce::Opportunity.where(id: opportunity_ids)
+      else
+        opportunities = Hmis::Ce::Opportunity.active
+      end
```
eanders: This is tiny, and probably doesn't matter, but I've found sometimes assigning un-initialized variables within case statements can lead to unassigned variables later.  
```suggestion
      opportunities = if opportunity_ids
        Hmis::Ce::Opportunity.where(id: opportunity_ids)
      else
        Hmis::Ce::Opportunity.active
      end
```

drivers/hmis/app/jobs/hmis/ce/build_candidate_pools_job.rb
```diff
@@ -0,0 +1,38 @@
...
+
+      return if opportunities.empty?
+
+      dirty_pool_ids = Hmis::Ce::Match::CandidatePoolBuilder.new(opportunities).perform
+      if opportunity_ids
+        Hmis::Ce::Match::CandidatePool.where(id: dirty_pool_ids).mark_all_dirty if dirty_pool_ids.any?
```
eanders, line 26: Do you think it would be worth adding a helper method to "dirty" the pools?
```
Hmis::Ce::Match::CandidatePool.dirty_pools(dirty_pool_ids)
```
It could then return if no ids are passed

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
```diff
@@ -0,0 +1,171 @@
...
+
+      candidate_pool_scope = ::Hmis::Ce::Match::CandidatePool.active.where(id: markers.map(&:trackable_id))
+      client_scope = ::GrdaWarehouse::Hud::Client.destination
+
+      candidate_pool_scope.find_each do |pool|
+        Hmis::Ce::Match::Engine.call(pool, client_scope, progress: @progress)
+      end
+
+      Hmis::Ce::ChangeMarker.mark_processed(markers)
+      markers.map(&:trackable_id).max + 1
```
eanders, line 115: are `trackable_id`s always sequential?

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
```diff
@@ -0,0 +1,171 @@
...
+    end
+
+    def log(message)
+      @notifier&.ping("[#{self.class.name}] #{message}")
+    end
```
eanders: Is this being used? (I'm not seeing it.)
Do we really want to send to the Slack notifier?  Maybe just Rails.logger?

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
eanders: It looks like these are not necessarily sequential, but the value is used in a query `where(id: next_id..)` so it's just looking for any thing greater.

drivers/hmis/app/models/hmis/ce/change_marker.rb
```diff
@@ -0,0 +1,77 @@
...
+      on_duplicate_key_update: {
+        conflict_target: [:trackable_id, :trackable_type],
+        columns: [:processed_version],
+      },
+    )
+  end
+
+  # Creates new markers or increments current_version for existing ones
+  def self.upsert_or_bump_version(trackable_type, trackable_ids:)
+    raise ArgumentError, "Trackable type not supported \"#{trackable_type}\"" unless trackable_type.in?(['GrdaWarehouse::Hud::Client', 'Hmis::Ce::Match::CandidatePool'])
```
eanders: Might be worth having a `known_trackable_types` method (or constant) if we ever add more than 2.

eanders (approved):
Looks pretty solid to me.

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
gigxz: +1 I think rails logger would be sufficient

drivers/hmis/app/models/hmis/hud/disability.rb
```diff
@@ -11,6 +13,7 @@ class Hmis::Hud::Disability < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::Shared
   include ::Hmis::Hud::Concerns::EnrollmentRelated
   include ::Hmis::Hud::Concerns::ClientProjectEnrollmentRelated
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz: I don't think this is necessary, this record is collected on a hud assessment (which gets stored as a custom assessment) so the CustomAssessment model trigger should be sufficient. We also don't have any use-case yet for  using this model for eligibility

drivers/hmis/app/models/hmis/hud/employment_education.rb
```diff
@@ -14,6 +14,7 @@ class Hmis::Hud::EmploymentEducation < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::Shared
   include ::Hmis::Hud::Concerns::EnrollmentRelated
   include ::Hmis::Hud::Concerns::ClientProjectEnrollmentRelated
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz: I don't think this is necessary, this record is collected on a hud assessment (which gets stored as a custom assessment) so the CustomAssessment model trigger should be sufficient. We also don't have any use-case yet for  using this model for eligibility

drivers/hmis/app/models/hmis/hud/health_and_dv.rb
```diff
@@ -11,6 +13,7 @@ class Hmis::Hud::HealthAndDv < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::Shared
   include ::Hmis::Hud::Concerns::EnrollmentRelated
   include ::Hmis::Hud::Concerns::ClientProjectEnrollmentRelated
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz: I don't think this is necessary, this record is collected on a hud assessment (which gets stored as a custom assessment) so the CustomAssessment model trigger should be sufficient. We also don't have any use-case yet for  using this model for eligibility

drivers/hmis/app/models/hmis/hud/income_benefit.rb
```diff
@@ -14,6 +14,7 @@ class Hmis::Hud::IncomeBenefit < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::EnrollmentRelated
   include ::Hmis::Hud::Concerns::ClientProjectEnrollmentRelated
   include ::Hmis::Hud::Concerns::FormSubmittable
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz: I don't think this is necessary, this record is collected on a hud assessment (which gets stored as a custom assessment) so the CustomAssessment model trigger should be sufficient. We also don't have any use-case yet for  using this model for eligibility

drivers/hmis/app/models/hmis/hud/youth_education_status.rb
```diff
@@ -11,6 +13,7 @@ class Hmis::Hud::YouthEducationStatus < Hmis::Hud::Base
   include ::Hmis::Hud::Concerns::Shared
   include ::Hmis::Hud::Concerns::EnrollmentRelated
   include ::Hmis::Hud::Concerns::ClientProjectEnrollmentRelated
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz: I don't think this is necessary, this record is collected on a hud assessment (which gets stored as a custom assessment) so the CustomAssessment model trigger should be sufficient. We also don't have any use-case yet for  using this model for eligibility

gigxz:
@ttoomey I'm having trouble testing this, can you give me some direction? I was able to modify data and confirm that a record shows up in hmis_ce_change_markers (with current_version > processed_version) but I'm unsure how to get the job to pick it up. I tried running `rails grda_warehouse:hourly`  and  `rails jobs:work` 

edit: I got this working I think by running `Hmis::Ce::ProcessChangesJob.enqueue_if_not_already_running(wait_time: 1.minute)`  it removed more candidates from the waitlist than I was expecting so I will continue digging into that

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
```diff
@@ -0,0 +1,171 @@
...
+    def with_lock(&block)
+      lock_name = self.class.name.to_s
+      ::GrdaWarehouseBase.with_advisory_lock(lock_name, timeout_seconds: 0, &block)
+    end
+
+    private
+
+    # Safeguard to ensure data integrity: finds active pools and destination clients that should be
+    # tracked but are missing a change marker. Ensures that all relevant records are eventually processed
+    # by the CE engine.
```
gigxz: @ttoomey  Can you explain this scenario a bit, i.e. what do you mean by "should be tracked"? I don't think I'm understanding it. Is the idea that ALL destination clients will always have a row in the change marker table, so if they don't have one, its because something went wrong (they weren't added when the client was created?)

Thank you! happy to jump on a huddle

drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md
```diff
@@ -0,0 +1,98 @@
...
+    MarkClientAsDirtyBehavior-->>-RailsApp: Done
+
+    Note over RailsApp,IdentifyDuplicates: Client creation/update triggers deduplication
+    RailsApp->>+IdentifyDuplicates: Runs periodically or on client changes
+    IdentifyDuplicates->>+ChangeMarker: upsert_or_bump_version (destination client)
+    ChangeMarker-->>-IdentifyDuplicates: Marks destination client dirty
+    IdentifyDuplicates-->>-RailsApp: Done
+
+    alt Client or Pool Changes
+        UserAction->>+RailsApp: Mark Units Available
```
gigxz, line 45: There is another "Dev User Action" that requires reprocessing pools: making changes to CE Match Rules. Right now I am making Rule changes manually in the console, and then my process has been to run `HmisUtil::CeBuilder.build_candidate_pools` to rebuild the candidate pools. Is that still the right process to follow, after this change?

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
gigxz: huddle notes:
- initially, should mark all destination clients as dirty
- marking pools as dirty might be enough
- this is for the initial generation. after that, IdentifyDuplicates should always be setting a marker when it creates a new destination client (or updates one)

ttoomey: discussed 6/10: I'll double check to ensure we actually need to create these client markers, we may only need pool markers

drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md
gigxz: updated note to self: can run `Hmis::Ce::BuildCandidatePoolsJob.perform_now` instead after changing match rules. worker must be running in order to pick up the dirty marker and actually reprocess the pool(s)

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -265,6 +265,8 @@ def set_hud_project_id_from_project_pk_unless_wip
   after_create :warehouse_trigger_processing
   after_update :warehouse_trigger_processing
 
+  include ::Hmis::MarkClientAsDirtyBehavior
```
gigxz, line 268: We want to put this on `Exit` as well. Creating/removing an Exit record should trigger re-processing because it may change the value for the `open_enrollment_project_types` field in rules

gigxz: I think we also we also want to update the client ce change marker [from each of these methods](https://github.com/greenriver/hmis-warehouse/blob/dc5421eb79456ed9b46f92160dfa4d1884f93be6/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb#L58-L73) (CE Referral created, CE Referral reject, accepted) in order to recalculate rules based on `open_referral_project_types`. I realize that maybe is out of scope for the initial work here, we could split it out if preferred!

drivers/hmis/app/models/concerns/hmis/mark_client_as_dirty_behavior.rb
```diff
@@ -0,0 +1,31 @@
...
+  extend ActiveSupport::Concern
+
+  included do
+    after_save :mark_destination_client_dirty
+  end
+
+  protected
+
+  def mark_destination_client_dirty
+    # find the destination client
```
gigxz, line 21: Should we do a `return unless Hmis::Ce.configuration.enabled?` to reduce performance impact for non-CE installations?

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
```diff
@@ -0,0 +1,171 @@
...
+    # @param markers [Array<Hmis::Ce::ChangeMarker>] Dirty pool markers to process
+    # @return [Integer, nil] Next pool ID for pagination
+    def process_dirty_pools(markers)
+      return 0 if markers.empty?
+
+      candidate_pool_scope = ::Hmis::Ce::Match::CandidatePool.active.where(id: markers.map(&:trackable_id))
+      client_scope = ::GrdaWarehouse::Hud::Client.destination
+
+      candidate_pool_scope.find_each do |pool|
+        Hmis::Ce::Match::Engine.call(pool, client_scope, progress: @progress)
```
gigxz: @ttoomey Okay, here I have bug. Sorry for my very fragmented review, I have been coming back to it between meetings.

Calling `Hmis::Ce::Match::Engine.call(pool, client_scope)` will populate the pool with any clients from client_scope that match, then [remove any existing clients in the pool](https://github.com/greenriver/hmis-warehouse/blob/4fd8e13d208bc119b86e8496795915d05c0fa16c/drivers/hmis/app/models/hmis/ce/match/engine.rb#L49). So as-is it can't be called like this to evaluate a subset of dirty clients.


In practice, the bug repro is:
* Set up a unit in a group that has an eligibility requirement for a particular assessment CDE
* Populate the candidate pool manually, the right 3 clients appear
* Perform an assessment for a new 4th client, which should make them eligible
* The assessment submission triggers the candidate generation
* When completed it results in only the new client appearing on the waitlist, and the previous 3 clients are removed

gigxz: oh wait 🤦  the client_scope IS all clients. I will keep looking at why that bug is appearing

gigxz: Edit: I meant to put this comment on line 131

drivers/hmis/app/jobs/hmis/ce/build_candidate_pools_job.rb
ttoomey: I'll extract this to a method to address the concern

drivers/hmis/app/jobs/hmis/ce/build_candidate_pools_job.rb
ttoomey: Sorry I don't quite follow, isn't it sufficient to use: `CandidatePool.where(id: dirty_pool_ids).mark_all_dirty`

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
ttoomey: thanks, yeah this is cruft. I'll remove it

drivers/hmis/app/jobs/hmis/ce/build_candidate_pools_job.rb
eanders: Yeah, that's fine as is, I was thinking there might be multiple times when you want to just mark as set of pools as dirty, so something like:

```
def self.dirty_pools(ids)
  where(id: ids).mark_all_dirty
and
```

Might be convenient.

drivers/hmis/app/models/hmis/hud/enrollment.rb
ttoomey: great catch, yeah that seems easy. Thanks!

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
ttoomey: I think this is needed. Without it, if a client is added and has no dirty marker, they won't get matched immediately. We'd have to wait for the daily re-processing of all pools to match that client. Or if a pool was marked dirty by some other mechanism, the new client would only be matched to that one pool.

Maybe that's okay since adding a client without a dirty marker is unexpected and would be eventually be caught by the daily task? Although creating dirty markers for all destination clients probably isn't very expensive.

I'll add documentation here to better explain the intent

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
ttoomey: Thanks for finding this! Sorry I missed it. I've modified the behavior of the engine to support incremental updates and added some tests (drivers/hmis/spec/models/hmis/ce/match_engine_spec.rb)

drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md
ttoomey: I adjusted HmisUtil::CeBuilder.build_candidate_pools so that it calls the appropriate jobs, simulating what should happen async. Also includes more params:
```
HmisUtil::CeBuilder.build_candidate_pools(progress:true)
```

drivers/hmis/app/jobs/hmis/ce/process_changes_job.rb
gigxz: Thank you! This approach looks good to me.

gigxz (approved):
Looks great! tested a few cases locally and its working as expected.  I will retarget to release-172 for more time to QA.

PR #5538, opened 2025-07-01
7865-rm-memory-profiler-x

ttoomey:
## _Merging this PR_

## Description
* Replaced memory_profiler with PeakMemorySampler. 
* This runs in a background thread, sampling peak memory usage of a given block of code with minimal performance impact. 
* It uses the get_process_mem gem which was already in the Gemfile

task profiling:
* The meaning of memory_allocated is changing from "total allocated during execution" to "peak increase from baseline"
* no longer supporting allocation_count

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/config/initializers/hmis_feature.rb
+++ b/drivers/hmis/config/initializers/hmis_feature.rb
@@ -23,10 +23,3 @@ Rails.application.config.queued_tasks[:hmis_disabling_condition_and_race_cleanup
   HmisDataCleanup::Util.fix_disabling_condition_nils!
   HmisDataCleanup::Util.fix_race_gender_99s!
 end
-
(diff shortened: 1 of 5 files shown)


drivers/hmis/config/initializers/hmis_feature.rb
```diff
@@ -23,10 +23,3 @@
   HmisDataCleanup::Util.fix_disabling_condition_nils!
   HmisDataCleanup::Util.fix_race_gender_99s!
 end
-
-TodoOrDie('Remove one-time job', by: '2025-07-01')
```
ttoomey, line 27: removed to make the tests pass

eanders (approved):
@ttoomey let's give this a try!  Looks like a good solution. 
Makes me want to turn that into a reusable component (gem?) for light-weight memory profiling.

PR #5539, opened 2025-07-01
Mutation for recording notes on CE Referral

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7408
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1194

How to test:
- pull down frontend pr, open a referral, go to activity, record note

Screenshot 2025-07-01 at 4 36 29 PM

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/referral_note.rb
```diff
@@ -1,9 +1,10 @@
 # frozen_string_literal: true
 
-# Free-form notes a referral
+# Free-form notes a referral, optionally associated with a particular step
 module Hmis::Ce
   class ReferralNote < GrdaWarehouseBase
     belongs_to :referral, class_name: 'Hmis::Ce::Referral'
     belongs_to :user, class_name: 'Hmis::User'
+    belongs_to :wfe_step, class_name: 'Hmis::WorkflowExecution::Step', optional: true
```
ttoomey, line 8: If it's possible to destroy a step, we should probably add an association to the Wfe Step for dependent-destroy so we don't get orphaned records

ttoomey (approved):
Looks good!

Just one thing: I think there's some migration churn in the db/warehouse_structure.sql file that shouldn't get merged.

drivers/hmis/app/models/hmis/ce/referral_note.rb
gigxz: Hmm good point. I think in order to do that, I would need to add an association to the ReferralNote on the Step class, which breaks the intended barrier of not referencing referral-related code in the workflow execution code. I added the step reference as something we will likely use in the future, since we've discussed step-specific notes as an enhancement, but it's not in use yet. So I'm going to leave it as-is since' it not totally clear to me what we want for destroys.

gigxz:
Thanks for nothing the db thing- I cleaned up my local test and dev dbs schema migration tables to match the release branch!

PR #5540, opened 2025-07-02
Resolve current eligibility-related values on CE Referral

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7591
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1196

* resolve current match values on referral
* expand visibility to client name, source enrollment, and match values

## Type of change


drivers/hmis/app/graphql/types/hmis_schema/ce_match_value.rb
```diff
@@ -0,0 +1,25 @@
...
+  class HmisSchema::CeMatchValue < Types::BaseObject
+    # object is an OpenStruct
+    # used to generically resolve Key-value pairs for fields referenced by Match Rule expressions
+
+    # logic to implement:
+    # 1. Find all ce_match_rules (or expressions logged somewhere?) that were applicable at the time what this referral was created
+    # 2. For each expression, resolve it into a set of referenced fields
+    # 3. For each field, evaluate it against this client (really, their destination client) using the FieldMap
+    # 4. For each field, translate the field name into a human-readable name (using CDE Label if appropriate)
+    # 5. Resolve key-value pairs in an array on the CeReferral type (put on the Referral model, probably)
```
gigxz, line 19: Pull this out into another pr

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -113,6 +113,62 @@ def self.apply_filters(input)
...
+          OpenStruct.new(
+            rule_id: rule.id,     # ID of rule that references this field
+            rule_name: rule.name, # Name of rule that references this field
+            field_name: label,
+            field_values: Array.wrap(current_value),
+          )
+        end
+      end.flatten.compact
+    end
+
```
gigxz, line 150: Screenshot 2025-07-08 at 4 16 35 PM

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -9,23 +9,30 @@
 module Types
   class HmisSchema::CeReferral < Types::BaseObject
     # object is a Hmis::Ce::Referral
+
```
gigxz, line 12: we should probably add object-level auth but I couldn't figure out how to do it performantly yet - will create follow-up

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -79,6 +86,17 @@ def client
       load_ar_scope(scope: Hmis::Hud::Client.viewable_by(current_user), id: object.client_id)
     end
 
+    # NOTE: This field intentionally does not check can_view_clients
+    def client_name
+      c = load_ar_association(object, :client)
+      c&.brief_name || c&.masked_name
```
martha: Would this ever actually return the masked name? My understand is `client` will never be nil; when would `client.brief_name` return nil?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -79,6 +86,17 @@ def client
...
+    end
+
+    # NOTE: This field intentionally does not check can_view_clients
+    def client_age
+      load_ar_association(object, :client)&.age
```
martha: Related question here - is the safe accessor necessary given that we expect a referral will always have a client -- it's required at the database level? Maybe we are thinking ahead to the vsp use case?

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -113,6 +113,62 @@ def self.apply_filters(input)
...
+      field_map = Hmis::Ce::Match::FieldMap.new
+      cde_field_map = Hmis::Ce::Match::CdeFieldMap.new
+      destination_client = client.destination_client.as_warehouse
+      seen_field_names = Set.new
+
+      match_rules.sort_by(&:id).map do |rule|
+        evaluator = Hmis::Ce::Match::ClientExpressionEvaluator.new(rule.expression, field_map)
+        evaluator.resolve_client_values(destination_client).map do |field_name, current_value|
+          # Skip if Field has already been processed, for example expression "household_size = 1 OR household_size = 2"
+          next if seen_field_names.include?(field_name)
```
martha: Nice to have: It might make sense to add a special case here that skips returning the client age, since it's already shown as a default client attribute. Or, maybe the backend should return everything and the frontend should do this filtering, since the fact that they are shown right next to each other is a display issue, not a logical issue.

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -113,6 +113,62 @@ def self.apply_filters(input)
       Hmis::Filter::CeReferralFilter.new(input).filter_scope(self)
     end
 
+    # Returns an array of fields referenced by Match Rules for this referral's opportunity,
+    # with their current values for the referred Client's _Destination_ Client record.
+    #
+    # NOTE: This field intentionally does not check viewable_by scopes on the associated data;
+    # it is resolved in the GraphQL API to expose pertinent data that the current user may not otherwise have permission to view.
+    def resolve_match_rule_fields
```
martha, line 121: Maybe it would be helpful to exercise this logic with a spec test

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -113,6 +113,62 @@ def self.apply_filters(input)
...
+      # Fetch all match rules applicable to the opportunity
+      # TODO: this should really be the rules _as they were_ when the referral was created, not the current rules
+      match_rules = Hmis::Ce::Match::Rule.for_opportunity(opportunity)
+      field_map = Hmis::Ce::Match::FieldMap.new
```
martha, line 125: This is an amorphous not-fully-formed thought, but is there any of this code that we want to pull into a more central place, in order to reuse it for snapshotting eligibility values (#7867) and eventually displaying/resolving those snapshots? ideas:
- Refactor lines 140-158ish into something like `display_label, display_value = Hmis::Ce::Match::FieldMap.display_label_and_value(field_type, resolved_field)`
- maybe we even want an internal API surface somewhere that we can just pass in a client and an opportunity/rules list, and get back a snapshot of current client attributes relevant to those rules? Something like `Hmis::Ce::Match::Engine.client_values_for(client, rules)`, but I am not sure where this should live

(Motivation: in its existing usage, `FieldMap` acts as a router that internally creates a `CdeFieldMap` and a `ClientFieldMap`, so when scanning this code, I was puzzled about the creation of a `CdeFieldMap` outside of `FieldMap`. I started wondering about graceful ways to move the work being done here by `CdeFieldMap` back into the `FieldMap` so that this code doesn't have to know about the 2 different types of field maps.)

martha (approved):
looks good! thank you! 🚀

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: Thats true, client should never be nil. `brief_name` can return an empty string if first and last name are both missing. This should maybe read `c.brief_name.presence || c.masked_name`

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: Good point. I think I will put the logic in the frontend,  so it lives next to the other logic of deciding to show the client age.

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: thanks, working on this! i also want to move the stringification-logic into the field maps. Since there is more needed for display, like for veteran status translating "1"=>"Yes" etc.

gigxz: ok, updated on https://github.com/greenriver/hmis-warehouse/pull/5540/commits/365e1c4c629c69bf36e05f1c138ad524c5a56136:
* Each FieldMap is responsible for implementing `label_for` (whats the user-facing label for this field?) and `instance_value_for_display` (what's the value for this field, formatted for user display?)
* Instead of using the ClientExpressionEvaluator to parse and evaluate, just use a calculator to parse and the FieldMap to get the label + display value for each field

much cleaner! thanks for the suggestions!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Thanks cool, I think I like what you have, but might suggest adding a comment stating explicitly that "masked" here doesn't mean it's hidden because the user lacks permission to see it (which I think "masked" otherwise implies).

drivers/hmis/app/models/hmis/ce/referral.rb
martha: Lovely! Thanks for turning my amorphous thought into beautiful modular code ✨

PR #5543, opened 2025-07-03
Add HMIS links to HMIS DQ Report

dtgreiner:
## _Merging this PR_

## Description

Add Enrollment and Client links to the Data Source's HMIS for the drilldown in the HMIS DQ Tool Report

## Type of change
New feature


drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/items.haml
```diff
@@ -42,7 +43,14 @@
                       = link_to appropriate_client_path(item[k]) do
                         = value
                       %a{href: '#', class: ['btn', 'btn-sm'], data: { toggle: :popover, title: 'Client Information', url: api_client_path(item.client_id), html: 'true' }}
-                        %i.icon-info
+                        %i.icon-infoexit
+                    - elsif k.to_s == 'hmis_enrollment_id'
+                      - url = item.data_source.hmis_url_for(item.enrollment, user: current_user)
+                      = link_to_if data_source_link_available && url.present?, formatted_cell(value), url, target: :_blank
```
eanders: What do you think about moving this to a shared partial or into an application_helper method.  Then we could just do something like:
`= link_to_external_hmis(item: item.enrollment, user: current_user, data_source: item.data_source)`

It could return nil or an empty string if the link isn't available, and a well formatted `a` tag if it is.
I think you could probably pull the data source off of the enrollment even, and figure out `data_source_link_available` within the method.

eanders (commented):
@dtgreiner , I think this looks good.  I _think_ we should standardize this a bit more if possible and move it to a central location (it could even be on the data source if you think that makes sense.

Additionally, while you're in here, do you think it would be easy to add to the HUD reports drill-downs?

drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/items.haml
dtgreiner: That sounds like a good plan to me, I don't think it should be terribly difficult either. I'll get something pulled into this PR.

dtgreiner:
> @dtgreiner , I think this looks good. I _think_ we should standardize this a bit more if possible and move it to a central location (it could even be on the data source if you think that makes sense.
> 
> Additionally, while you're in here, do you think it would be easy to add to the HUD reports drill-downs?

I'll get a centralized method put together to call get the url and believe the HUD drill-downs are very similar to the DQ tool, so that shouldn't be too much more effort.

drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/items.haml
```diff
@@ -34,15 +34,22 @@
...
                   -# - value = item.public_send(k)
                   - value = item.download_value(k)
+                  - formatted_value = formatted_cell(value)
                   %td
                     - if k.to_s == 'destination_client_id'
                       = link_to appropriate_client_path(item[k]) do
                         = value
                       %a{href: '#', class: ['btn', 'btn-sm'], data: { toggle: :popover, title: 'Client Information', url: api_client_path(item.client_id), html: 'true' }}
-                        %i.icon-info
+                        %i.icon-infoexit
```
eanders: This looks like a typo.  I think that should just be 
```suggestion
                        %i.icon-info
```

eanders (approved):
One small typo, but otherwise I think this looks great, thanks!

PR #5551, opened 2025-07-07
Generate/update CE events from referral workflow

martha:
## _Merging this PR_

## Description
- Add a new `ReferralCeEventManager` for referral side effects related to CE events (create, update).
- Add a few specific new helpers to `HudUtility2024`, on the premise that we should keep HUD-logic centralized in helper files as much as possible, rather than hard-coding it throughout the app. But this is my first time touching the year-specific helper file, and I don't know whether/how it interacts with the upcoming 2026 changes - let me know if there's anything additional to consider!
- Update the `ReferralEnroller` so that it finds the submitted move-in date value by item mapping, not item link ID

How to test: For an opportunity/unit that uses the "AC V1" workflow template,
- Start a referral but decline the "Initial Review" step -> No CE Referral event is created.
- Start a referral and pass to the "Client Acceptance" step -> a CE event is created on the _source_ enrollment. It does not have a referral result
- Select "No, decline referral" at the "Client Acceptance" step -> the CE event should now be updated with a referral result "Unsuccessful referral: client rejected".
- Start a new referral. Pass the "Client Acceptance" step, but go through the denial workflow and "Approve Denial." ->  The CE event should be updated with a referral result "Unsuccessful referral: provider rejected".
- Start a new referral. Pass through the whole referral to the "Confirm Enrollment" step, but reject the referral at this stage. -> The CE event should be updated with a referral result "Unsuccessful referral: provider rejected".
- Start a new referral. Pass through the whole referral to successfully accept the client. The CE event should be updated with the result "Successful referral: client accepted"

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -55,11 +50,18 @@ def create_enrollment(message)
 
     # this is not actually an accessor method, even though RuboCop thinks it is
     def set_move_in_date(message) # rubocop:disable Naming/AccessorMethodName
-      project = referral.target_project
-      raise 'access denied' unless message.user.can_edit_enrollments_for?(project)
+      # No need to check permission for editing the enrollment.
+      # If the user can complete this step, they can do its side effects, even if they don't have direct permission
```
martha, line 54: Highlighting this out-of-scope permissions change for review/discussion - let me know if you agree with the change I've proposed here. (See "Permissions" section of the CE-in-HMIS spec doc)

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
```diff
@@ -0,0 +1,73 @@
...
+    def create_ce_event(message)
+      enrollment = referral.source_enrollment
+      raise 'Referral does not have a source enrollment' unless enrollment.present?
+
+      target_project = referral.target_project
+
+      enrollment.events.create!(
+        event_date: Date.current,
+        event: project_to_event_type(target_project), # TODO(#7527) determine event type from configuration if present
+        location_crisis_or_ph_housing: target_project.id,
```
martha: To discuss: Should this be the project name? Or should we implement matching logic to display the correct project? (feels like scope creep; reminds me of https://github.com/open-path/Green-River/issues/6272)
Screenshot 2025-07-08 at 4 43 48 PM
Data dictionary says: "Location of Crisis Housing or Permanent Housing Referral [Project name and/or
Project ID]"

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
```diff
@@ -0,0 +1,73 @@
...
+        user: Hmis::Hud::User.from_user(message.user),
+      )
+    end
+
+    def set_ce_event_result(message) # rubocop:disable Naming/AccessorMethodName
+      enrollment = referral.source_enrollment
+      event_type = project_to_event_type(referral.target_project)
+      event = enrollment.events.where(event: event_type).order(:date_created).last # We don't have a precise way to map to the CE Event that was created earlier in this workflow, so just expect it to be the most recently created one.
```
martha: Flagging this as a possible problem. We don't directly create a mapping from the CE Event to the referral that generated it. So when it comes time to update that event, we're guessing. Scenario where this could cause a problem:
- user starts Referral 1 for PSH Project 1. CE Event 1 is created
- user starts Referral 2 for PSH Project 2. CE Event 2 is created
- user finishes Referral 1. CE Event 2 would then be updated with the result from Referral 1, which is incorrect.

Does https://github.com/open-path/Green-River/issues/7835 sufficiently protect us from that scenario, or do we need to guard against it further? e.g., add optional `ce_referral_id` col to `Event` table?

drivers/hmis/lib/tasks/ce_define_workflows.rake
martha, line 1: Flagging something potentially confusing: With my current simple implementation, the CE Event Result we save depends only on what "branch" you took in the workflow, not on what specific answer you gave in the "Decline Reason" question. That means, for example, if you decline the referral at the "Provider Acceptance" step but give the reason "No longer interested in this program" (see screenshot), the CE Event Result would be saved as "Unsuccessful referral: provider rejected" _even though_ it might be more intuitive to think of this as a "client rejected" case.
Screenshot 2025-07-08 at 4 55 03 PM

We could update the referral definition with more complexity, gateways, etc. to deal with this. Or we could restrict the number of possible answers, so that at the "Provider Acceptance" step you are only presented with provider-initiated reasons for rejection, but that might end up being a very frustrating UX.

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
ttoomey: Good memory, yes it is like the verified-by field on CLS. So maybe we should   `location_crisis_or_ph_housing_project_id` (same pattern as `verified_by_project_id`)

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
ttoomey: That's a great point. I think adding an optional reference to the referral from on the event would be valuable.

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
```diff
@@ -0,0 +1,73 @@
...
+        return HudUtility2024.ce_events_by_code[:rrh] if HudUtility2024.residential_project_type_numbers_by_code[:rrh].include?(project_type)
+      end
+
+      psh_types = HudUtility2024.residential_project_type_numbers_by_code[:psh]
+      return HudUtility2024.ce_events_by_code[:psh] if psh_types.include?(project_type)
+
+      oph_types = HudUtility2024.residential_project_type_numbers_by_code[:oph]
+      return HudUtility2024.ce_events_by_code[:oph] if oph_types.include?(project_type)
+
+      raise "Unexpected target project type #{project_type} on project #{project.id} for referral #{referral.id}"
```
ttoomey: Since we reference the utility so heavily here, and it's hud-defined logic, I'd be tempted to just put this logic directly in the utility class.

```suggestion
      event_type  = HudUtility2026.project_to_event_type(project_type)
      raise "Unexpected target project type #{project_type} on project #{project.id} for referral #{referral.id}" unless event_type
      
      return event_type
```

I also think it should be safe to use 2026 here it's almost identical to 2024 and I believe 2026 is live on 10/1

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -170,7 +170,9 @@ def access
 
     def audit_events
       object.audit_events.
-        where(event_type: ['complete_step', 'start_workflow', 'end_workflow']).
+        where(event_type: 'end_workflow').
+        where("event_data->>'message' IN (?)", [Hmis::Ce::ReferralMessageHandler::REJECT_REFERRAL_MESSAGE, Hmis::Ce::ReferralMessageHandler::ACCEPT_REFERRAL_MESSAGE]).
```
ttoomey, line 203: Add some documentation to describe why this additional condition is needed?

drivers/hmis/lib/tasks/ce_starter_pack_20250302.rake
```diff
@@ -1,382 +0,0 @@
-# Helpers to reduce cruft and make the starter pack script more readable
```
ttoomey, line 1: note to self, moved to ce_define_workflows.rake

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
ttoomey: Makes sense to me, that's the decision we arrived at in the spec doc

ttoomey (approved):
Looks great! I left some suggestions

drivers/hmis/lib/tasks/ce_define_workflows.rake
ttoomey: Hmm, so this isn't a very well-formed idea but could we use expressions to map the variables? The workflow engine would need to evaluate the parameters before sending them to the message handler

I asked AI to scaffold this as a demonstration:
```ruby
    client_rejects_ce_event_task = Hmis::WorkflowDefinition::ScriptTask.create!(
      name: 'Update CE Event with referral result',
      template_id: template.id,
      trigger_config: [
        {
          event: 'complete_step',
          message: 'set_ce_event_result',
          # Evaluate the referral_result from the 'ce_referral_result_code' form field, defaulting to '2' (client rejected)
          params: { referral_result: "if(is_present(ce_referral_result_code), ce_referral_result_code, '2')" },
        }
      ]
    )
```

```diff
--- a/drivers/hmis/app/models/hmis/workflow_execution/engine.rb
+++ b/drivers/hmis/app/models/hmis/workflow_execution/engine.rb
@@ -116,9 +116,11 @@ module Hmis::WorkflowExecution
       results = node.triggers.map do |trigger|
         next unless event_type == trigger.event
 
+        evaluated_params = evaluate_trigger_params(trigger.params, all_submitted_values)
+
         result = send_message(
           type: trigger.message,
-          params: trigger.params,
+          params: evaluated_params,
           step: step,
           user: user,
         )
@@ -196,18 +198,24 @@ module Hmis::WorkflowExecution
       calculator.evaluate!(expression, **defaults.merge(all_submitted_values.transform_keys(&:to_sym)))
     end
 
+    def evaluate_trigger_params(params, data)
+      calculator = Dentaku::Calculator.new
+      params.transform_values do |value|
+        # For now, we only evaluate strings. If we need to support more complex data structures, we can expand this.
+        next value unless value.is_a?(String)
+
+        calculator.evaluate(value, data.transform_keys(&:to_sym))
+      end
+    end
+
     def all_submitted_values
       instance.steps.reset
```

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
martha: I think we weren't 100% happy with the solution we came up with for VerifiedBy on CLS:
- code snippet: https://github.com/greenriver/hmis-warehouse/blob/57df21f18fc931d177a8a4bcf64a6964354a7257/drivers/hmis/app/graphql/types/hmis_schema/current_living_situation.rb#L50-L60
- Outstanding bug ticket in icebox: https://github.com/open-path/Green-River/issues/6496

Now that we are finding ourselves needing to reuse that pattern, maybe it's time to revisit. But I would advocate for descoping some of this complexity (and decision-making) from CE MVP. What if we do the following:
- add a `location_crisis_or_ph_housing_project_id` column, as you proposed
- when CE generates a CE event, save the project ID to that new column, and save the project name to the existing HUD column
- don't show the new column in the UI at all. When the user reopens the CE event in the UI, they will see the "Location of Crisis Housing or Permanent Housing Referral" question as a text field, not a project dropdown, and they aren't prevented from editing its string value. But they can't change the value of the new column. (In other words, no change from existing functionality on the CE event UI)
- add a ticket to the backlog for making the CE Event UI have a Project dropdown, and standardizing the approach with VerifiedBy, with link to existing bug ticket

How does that sound? Maybe it is already what you had in mind?

gigxz: I like that approach but I would suggest also deferring the addition of `location_crisis_or_ph_housing_project_id` to another ticket, since it's scope creep for this issue. (There is also a possibility that the export guidance may change in the 2027 standard, which we may want to consider.) So - I would suggest keeping your proposed implementation, with a comment linking to the follow-up issue.

gigxz: This is also something we'll need to deal with for superset, if we don't already. (maybe the CE Event analytics view can attempt to find the real project id for the HUD field?)- if you can mention that in the follow up ticket!

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
gigxz: Yeah good point. 7835 requires a certain configuration to prevent this from happening, and even then it won't 100% prevent it because of lag in candidate pool processing.

Adding a reference is probably worthwhile but it doesn't feel like a MVP requirement, can you make a follow-up ticket? Since additions to the HUD tables should probably have additional reviewers.

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
gigxz: Agreed^!

gigxz:
```suggestion
      raise "Unable to determine CE Event Type for project type #{project_type} on project #{project.id} for referral #{referral.id}"
```

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
gigxz: 👍  thanks

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -55,11 +50,18 @@ def create_enrollment(message)
...
+
+      # Find the form item corresponding to move-in date
+      form_definition = message.step.form_definition # The form definition most recently used to complete the step
+      move_in_date_item = form_definition.link_id_item_hash.values.find do |item|
+        item['mapping']['field_name'] == 'moveInDate' && item['mapping']['record_type'] == 'ENROLLMENT'
```
gigxz: Since these are OpenStructs we can access like so
```suggestion
        item.mapping&.field_name == 'moveInDate' && item.mapping&.record_type == 'ENROLLMENT'
```

gigxz (approved):
💯  looks great

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: This seems worth exploring to make triggers more expressive! But I also think the workaround of using gateways would be sufficient for mvp.

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
ttoomey: Something to consider is that if we add a reference to referral from the event, then we can access the project that way without the need for an additional association. 

I'm unsure if exposing "Location of Crisis Housing or Permanent Housing Referral" to users makes sense given that we can determine it automatically (outside of editing historical data). Perhaps hiding it by default could avoid data-entry errors.

martha: It's currently also possible to create and edit CE Events directly in the UI without a referral. Do we plan to disable that once CE is fully in place? If not, then we probably want to add both references (`ce_referral_id` _and_ `location_crisis_or_ph_housing_project_id`).

Anyway, I created https://github.com/open-path/Green-River/issues/7954 so we can continue the discussion there. We can always split it out into 2 tickets if necessary.

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
martha: Created https://github.com/open-path/Green-River/issues/7954 (same ticket as above, as @ttoomey pointed out they are related)

drivers/hmis/lib/tasks/ce_define_workflows.rake
martha: Sounds good! Created https://github.com/open-path/Green-River/issues/7955

PR #5553, opened 2025-07-08
Move custom HUD service form to patch

gigxz:
## _Merging this PR_

## Description

Issue: https://github.com/open-path/Green-River/issues/7921

Add a "patch" to append a custom note field to the HUD Service Form. This is currently implemented in an environment as a non-managed-in-version-control form, which means its not automatically updated when the system service form changes.

When this change is deployed, we should be able to delete that "custom" HUD Service form from the environment.

I needed to add some new code to patching to support appending items to a form.

How to test: `CLIENT=client_name rails driver:hmis:seed_definitions`
## Type of change
new feature

### DIFF
--- a/drivers/hmis/lib/hmis_util/json_forms.rb
+++ b/drivers/hmis/lib/hmis_util/json_forms.rb
@@ -4,12 +4,14 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module HmisUtil
   class JsonForms
     JsonFormException = Class.new(StandardError)
     private_constant :JsonFormException
 
-    DATA_DIR = 'drivers/hmis/lib/form_data'.freeze
+    DATA_DIR = 'drivers/hmis/lib/form_data'
 
     def initialize(env_key: nil)
       @env_key = env_key if env_key.presence # allow override for testing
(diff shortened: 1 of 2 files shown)


drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -173,13 +177,36 @@ def apply_patches(tree, patches)
       [result, applied_patches]
     end
 
+    # Similar to apply_item_patches, but applies patches to the overall form
+    # instead of to an individual item. This was added to support appending an item
+    # to the Service form.
+    def apply_form_patches(definition, patches, identifier:)
+      result = definition.deep_dup
+      applied_patches = []
+      patches.filter { |patch| patch['form_identifier'] == identifier && !patch.key?('link_id')}.each do |patch|
```
ttoomey: seems like the caller already does this filtering. Maybe raise if extra safety is needed
```suggestion
      patches.filter { |patch| patch['form_identifier'] == identifier }.each do |patch|
        raise if patch.key?('link_id') # should not reach this
```

ttoomey (approved):
Looks great, thanks!

* saw some small rubocop issues in json_forms
* perhaps at some point we should add a unit test for JsonForms, I think it's exercised incidentally by other tests

martha (approved):
Cool! Thanks!

PR #5556, opened 2025-07-09
Add custom referral statuses

martha:
## _Merging this PR_

## Description
Summary of changes
- Add migration and rails models/associations to create the `ce_custom_referral_statuses` table and reference it from `ce_referrals`
- Add `customStatus` to referral object on GraphQL api
- Add CE Referral Statuses picklist that resolves both custom and default (state machine) statuses
- Update the referral filtering logic to accept both custom and default statuses. I also renamed the status filter from `status` to `referralStatus`, which [this frontend change](https://github.com/greenriver/hmis-frontend/pull/1199/files#diff-ae9c1b5394f7e4c92dab83c36fe71cdd63a1bca05f4b2b51946d8a87425883e6R61) necessitates
- Add new event trigger `make_step_available` to the workflow engine. I added a comment thread to discuss this inline

How to test
- Run the rake task `rails driver:hmis:ce_define_workflows`. This creates a new custom status "Denied Pending"
- Open a referral and progress through the steps to put it in the Denied Pending state
- On the Project/Client/Admin referrals screens, the Status filter now includes the Denied Pending custom status

## Type of change
new feature


db/warehouse/migrate/20250708132033_add_ce_custom_referral_statuses.rb
```diff
@@ -0,0 +1,26 @@
...
+###
+
+# frozen_string_literal: true
+
+class AddCeCustomReferralStatuses < ActiveRecord::Migration[7.1]
+  def change
+    create_table :ce_custom_referral_statuses do |t|
+      t.string :key, null: false, index: { unique: true }
+      t.string :name, null: false
+      t.string :treatment, null: true
```
martha, line 14: I've added optional `treatment`, but haven't done anything with it yet on the frontend, seeing it as a nice-to-have. Let me know if you think it makes sense to implement now, though!

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -434,11 +435,23 @@ module CeWorkflowBuilder
       ],
     )
 
+    denied_pending_status = Hmis::Ce::CustomReferralStatus.find_or_create_by!(
```
martha, line 473: Should there also be an Approved Pending custom status?

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -162,6 +162,7 @@ def visit_node(node, user)
       when Hmis::WorkflowDefinition::UserTask
         step = enable_step!(node)
         assign_task!(step)
+        process_triggers(node: node, event_type: 'make_step_available', user: user)
```
martha: Let's discuss pros/cons of adding this `make_step_available`. It's not strictly necessary, but maybe it makes workflows more ergonomic because it elides the need for extra ScriptTasks. But otoh maybe the ScriptTasks are nice, because they show up distinctly in workflow diagrams.

I think it has the potential to be a little confusing -- if we keep it, I might want to rename it and/or add more comments distinguishing it from `start_step`.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -86,6 +86,14 @@ def self.options_for_type(pick_list_type, user:, project_id: nil, client_id: nil
         eligible_staff_assignment_user_picklist(project)
       when 'ELIGIBLE_REFERRAL_STEP_ASSIGNMENT_USERS'
         eligible_referral_step_assignment_user_picklist(project)
+      when 'CE_REFERRAL_STATUSES'
+        custom = Hmis::Ce::CustomReferralStatus.viewable_by(user).map do |status|
+          { code: status.key, label: status.name }
+        end
+        default = Hmis::Ce::Referral.state_machine_states.map do |status|
+          { code: status, label: status.to_s.humanize }
+        end
```
ttoomey: This mixes custom and default statuses in the pick list, which creates some complexity around key collisions and filtering logic. For example a custom status of 'in_progress' could be ambiguous.

Another possibility is to generate custom statuses that are 1:1 with state machine statuses during workflow setup. That might simplify the logic and support label customization

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -93,6 +95,14 @@ def send_notification(message)
       # ).deliver_later
     end
 
+    def set_custom_referral_status(message) # rubocop:disable Naming/AccessorMethodName
+      status_key = message.params['custom_status_key']
+      return unless status_key
+
+      status = Hmis::Ce::CustomReferralStatus.find_by(key: status_key)
```
ttoomey: suggest using the bang variant to raise if not found
```suggestion
      status = Hmis::Ce::CustomReferralStatus.find_by!(key: status_key, data_source: referral.data_source)
```

drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
```diff
@@ -22,8 +22,18 @@ def filter_scope(scope)
...
+    with_filter(scope, :referral_status) do
+      all_defaults = Hmis::Ce::Referral.state_machine_states
+
+      default_statuses, custom_keys = input.referral_status.partition { |status| all_defaults.include?(status.to_sym) }
+      custom_statuses = Hmis::Ce::CustomReferralStatus.where(key: custom_keys)
+
+      missing_keys = custom_keys - custom_statuses.pluck(:key)
+      raise "Received unknown custom status(es): #{missing_keys.join(', ')}" if missing_keys.any?
+
+      scope.where(status: default_statuses, custom_status: nil).or(scope.where(custom_status: custom_statuses))
```
ttoomey: if the user searches for a "default_status" and the referral also has a custom_status set, would this filter out a referral that should match? Maybe that's intended. Perhaps add some comments to help clarify the intent.

db/warehouse/migrate/20250708132033_add_ce_custom_referral_statuses.rb
```diff
@@ -0,0 +1,26 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddCeCustomReferralStatuses < ActiveRecord::Migration[7.1]
+  def change
+    create_table :ce_custom_referral_statuses do |t|
+      t.string :key, null: false, index: { unique: true }
```
ttoomey: probably should be unique on [key,data_source_id]?

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
ttoomey: yes, let's discuss. I think clearly communicating the workflow (and all side effects) seems important.  ScriptTasks might be a better fit for that. Agree that we could probably find a better name than 'make_step_available'

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: Thanks, good point about the key collisions. I took a stab at your suggestion: https://github.com/greenriver/hmis-warehouse/pull/5556/commits/0cb2c32f32598a93ee4ba90c67d79fa420ac4fdd

Having sketched it out, I'm a little hesitant about this solution -- it felt like it required so much commentary, in several different places, in order to be understandable, which I think of as an indication that it might be a sticky/confusing spot in the code for us a few months from now.

An alternative idea would be to just validate on the custom statuses that there is no collision. I sketched that idea here: https://github.com/greenriver/hmis-warehouse/pull/5556/commits/e2ebc0d885d0b9e6c46f23e8199de8c93ab2eff2
It is less foolproof because validations can be skipped, but I think it is more straightforward, and might be good-enough and easier to understand/maintain. What do you think?

(edited to add: these two sketches are incompatible, one of them definitely needs to be reverted before merging)

drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
martha: Added comments https://github.com/greenriver/hmis-warehouse/pull/5556/commits/730869af408b6e0a17691900f0dc2b81c9bf4366

I'd like to confirm this behavior with you @gigxz . My thinking was that it's more intuitive for the user this way since default statuses and custom statuses aren't differentiated in the dropdown, but let me know if you agree!

db/warehouse/migrate/20250708132033_add_ce_custom_referral_statuses.rb
martha: https://github.com/greenriver/hmis-warehouse/pull/5556/commits/c140a28fc1cde8f3e7da49e9a3032b74c8a2a520

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Thanks, I added to our dev standup agenda. Do you mean that you would like us to generally prefer Script Tasks over side effects on User Tasks?

[if keeping], maybe `enable_step` would be a good name? (aiming for parity in imperative voice with `complete_step`, `end_workflow` etc.)

I added some updates to the mermaid diagram so that the nodes include information about their side effects. It's not that pretty but the more information the better, I guess:
```mermaid
flowchart TD
start_referral_384(("Start Referral<br/><br/>EVENT: Start Workflow → Start Referral"))
start_referral_384--> initial_review_385
initial_review_385("Initial Review")
initial_review_385--> exclusive_gateway_initial_review_task_393
exclusive_gateway_initial_review_task_393{"Exclusive Gateway: initial_review_task"}
exclusive_gateway_initial_review_task_393-- IF move_forward = 0 --> referral_declined_392
exclusive_gateway_initial_review_task_393--> client_acceptance_386
referral_declined_392(("Referral Declined<br/><br/>EVENT: End Workflow → Reject Referral"))
exclusive_gateway_ce_offer_outcome_394-- IF move_forward = 0 --> referral_declined_392
exclusive_gateway_denial_review_396-- IF ac_workflow_v1_denial_review_decision = 1 --> referral_declined_392
confirm_success_390-- IF move_forward = 0 --> referral_declined_392
client_acceptance_386("Client Acceptance")
client_acceptance_386--> exclusive_gateway_ce_offer_outcome_394
exclusive_gateway_ce_offer_outcome_394{"Exclusive Gateway: ce_offer_outcome"}
exclusive_gateway_ce_offer_outcome_394--> provider_acceptance_387
provider_acceptance_387("Provider Acceptance")
exclusive_gateway_denial_review_396--> provider_acceptance_387
provider_acceptance_387--> exclusive_gateway_project_offer_outcome_395
exclusive_gateway_project_offer_outcome_395{"Exclusive Gateway: project_offer_outcome"}
exclusive_gateway_project_offer_outcome_395-- IF move_forward = 0 --> denial_review_389
exclusive_gateway_project_offer_outcome_395--> create_enrollment_388
denial_review_389("Denial Review<br/><br/>EVENT: Make Step Available → Set Custom Referral Status<br/>PARAMS: custom_status_key: denied_pending")
denial_review_389--> exclusive_gateway_denial_review_396
exclusive_gateway_denial_review_396{"Exclusive Gateway: denial_review"}
create_enrollment_388("Create Enrollment<br/><br/>EVENT: Complete Step → Create Enrollment")
create_enrollment_388--> confirm_success_390
confirm_success_390("Confirm Success")
confirm_success_390--> referral_accepted_391
referral_accepted_391(("Referral Accepted<br/><br/>EVENT: End Workflow → Accept Referral"))
```

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
ttoomey: Thanks! I can see the benefit to both approaches. However I'd lean towards using custom-statuses to represent state-machine status. I think it better supports customization and enhancement of behavior around these custom statuses. But we could start with the blended statuses + validation and then reconsider in the future if it becomes a source of problems

If we wanted to pursue: 0cb2c32 
* in ce_define_workflows.rake  probably just generate the statuses from `Hmis::Ce::Referral.state_machine_states` rather than hard-coding
* When a referral changes status the custom status should be updated too (probably referral message handler). Then both the picklist and filter logic are simplified and only operate on custom status

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
ttoomey: Thanks, the diagram is really helpful
> Do you mean that you would like us to generally prefer Script Tasks over side effects on User Tasks?

I guess the important thing is that we can represent it back to the user clearly. I guess my assumption is the script task makes that easier to implement. But I could also see an argument that the additional step adds clutter. I guess this is pretty dependent on the treatment.

Thinking about the different approaches, script tasks might be the right tool where we need something complex, like multiple entry points to task that set different statuses. But we don't have that use-case fortunately. 

> [if keeping], maybe enable_step would be a good name? (aiming for parity in imperative voice with complete_step, end_workflow etc.)

Yes, I think "enable_step"  is good choice. We should probably describe the event sequence in a comment somewhere "enable_step → start_step → complete_step"

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: That sounds reasonable and I've taking a pass at implementing it as you described -- though, beginning to wonder if "custom" status is still the right term for this table/model.

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: - [x] rename to "enable_step"
- [x] add comment "enable_step → start_step → complete_step"
- [x] get rid of "event" language in the mermaid diagram, it's probably confusing
- [x] indicate in mermaid diagram if something is a script task
- [x] make a nice-to-have ticket for further improvements to mermaid diagram

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: Discussed and agreed to keep "custom." I added some commentary https://github.com/greenriver/hmis-warehouse/pull/5556/commits/5984f36a0f9fb5f43271e1f2230dfc5d67741191

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: We can hold off on that while workflow is being finalized - having the example for denied pending is helpful for now!

gigxz (approved):
looks great!

gigxz:
@martha can you retarget to 172? thanks!

martha:
@gigxz thanks, I think this is ready to merge, but I'll wait for frontend review on https://github.com/greenriver/hmis-frontend/pull/1199 before doing so, so that they go into the same release.

gigxz:
approved! go for it

PR #5559, opened 2025-07-10
Custom imports in the HMIS CSV importer

eanders:
## _Merging this PR_

## Description
* This implements two methods of including extra files in the standard HMIS CSV import
  1. Files can be included that augment existing HMIS classes & tables.  These only affect exiting data by providing additional columns of data.  This is being used to allow import of Gender data removed in the FY2026 data standard into the existing gender columns on Client records.
  2. Files can be included to populate existing tables that aren't part of the core HMIS CSV data exchange specification.  This is being used to bring data into Custom Data Elements and Custom Data Element Definitions.  This mechanism should be flexible enough to include any data that is associated with a particular HMIS record but only needs one attribute to represent it. 
  
## Type of change
* New feature


ttoomey (commented):
Per our discussion, here's how I understand what we are trying to achieve:

### 1) "core extension" for column/table support (called augmentation or new_table)
* supports system-wide policy decisions which are common across all installations
* allow continued support for deprecated HUD fields or tables (no case for tables yet)
* the yaml config files would be part of your the codebase along with migrations

### 2) Key-value store
* these support community-specific configuration that only impact one installation
* supports handling "one-off" requests 
* does not require migrations, configuration would most-likely be backed by the db and could be configured by admins

Looks like we're going to use dynamic loader/importer classes that are setup at boot-time. I think this works for use case 1.

I'm not sure the key-value store makes sense to include in the yaml config though. To support dynamic fields I think we'd probably want to back the configuration from custom_data_elements instead. Maybe a separate phase to implement that

eanders:
> Per our discussion, here's how I understand what we are trying to achieve:
> 
> ### 1) "core extension" for column/table support (called augmentation or new_table)
> * supports system-wide policy decisions which are common across all installations
> * allow continued support for deprecated HUD fields or tables (no case for tables yet)
> * the yaml config files would be part of your the codebase along with migrations
> 
> ### 2) Key-value store
> * these support community-specific configuration that only impact one installation
> * supports handling "one-off" requests
> * does not require migrations, configuration would most-likely be backed by the db and could be configured by admins
> 
> Looks like we're going to use dynamic loader/importer classes that are setup at boot-time. I think this works for use case 1.
> 
> I'm not sure the key-value store makes sense to include in the yaml config though. To support dynamic fields I think we'd probably want to back the configuration from custom_data_elements instead. Maybe a separate phase to implement that

I agree, the key values store should not have been left in, sorry my cleanup wasn't complete.  `custom_data_elements` is the path forward.

Do you think the extra complexity of creating the classes on the fly (at boot) is worth the simplicity of just adding a YAML file to add a new type of file to the import?  I'm thinking we'll reuse the same YAML for the exporter, but it would also be relatively easy to just create a stub class when we generate the migrations.  We could even script it, if we think having something to reference in the code would be helpful.  If that's not clear, right now, to add gender support we just:
1. Create `custom_gender.yaml` 
2. Run `HmisCsvTwentyTwentySix::CustomFileManager.generate_custom_migrations!`

That gets it working but there is no file holding `HmisCsvTwentyTwentySix::Importers::CustomGender` that is programmatically generated at boot.  In an alternate version we could:

1. Create `custom_gender.yaml` 
2. Run `HmisCsvTwentyTwentySix::CustomFileManager.generate_custom_files!`

This could create the migrations as it does now, but also a set of classes:
```
class HmisCsvTwentyTwentySix::Loader::CustomGender < Base
  self.table_name = '...'
end
```
and
```
class HmisCsvTwentyTwentySix::Importers::CustomGender < Base
  self.table_name = '...'
end
```

and
```
class HmisCsvTwentyTwentySix::Exporters::CustomGender < Base

end
```

that we could commit to the repo.

ttoomey:
> Do you think the extra complexity of creating the classes on the fly (at boot) is worth the simplicity of just adding a YAML file to add a new type of file to the import? I'm thinking we'll reuse the same YAML for the exporter, but it would also be relatively easy to just create a stub class when we generate the migrations. We could even script it, if we think having something to reference in the code would be helpful.

Yes I think generating stub classes is a great approach; and we use that pattern already for hud lists. Seems much easier to understand than code-less classes; also nice to avoid additional complexity in the boot process. Recommend just attaching a comment to the top of the class that it's auto generated.

eanders:
@ttoomey, @dtgreiner, and @gigxz there's no rush on this review, it could easily go into 173.
It's probably easiest to start with the [README](https://github.com/greenriver/hmis-warehouse/blob/e7cb48272f59eea65eba9e102f49001376f9e7a4/drivers/hmis_csv_twenty_twenty_six/README.md), which _should_ be up-to-date and give you a decent overview of the import process.

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six.rb
```diff
@@ -40,28 +47,119 @@ def self.importable_files_map
...
+        Dir.glob(File.join(custom_dir, '*.yaml')).sort.each do |config_file|
+          config = YAML.load_file(config_file, permitted_classes: [Date, Time])
+          if config && config['custom_files'].is_a?(Array)
+            all_custom_files.concat(config['custom_files'])
+          else
+            Rails.logger.warn "Custom file config #{config_file} has invalid structure - expecting 'custom_files' array"
+          end
+        rescue StandardError => e
+          Rails.logger.error "Failed to load custom file config #{config_file}: #{e.message}"
+          raise e if Rails.env.development? || Rails.env.test?
```
ttoomey: Just to confirm, are we sure it's okay to swallow missing-file errors in production? (same concern on 108, 132, and 149)

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/custom_import_concern.rb
```diff
@@ -0,0 +1,166 @@
...
+      # For augmentations, we explicitly don't update the source hash since that would cause future
+      # imports to the augmented class to appear modified
+      # We also don't set the source_id as it would cause the source data drill-down to break
+      unless self.class.augments?
+        record.source_hash = source_hash
+        # Note which record we're sending this from for error checking
+        record.source_id = id
+        # For non-augmentations, we need to set the data_source_id
+        record.data_source_id = data_source_id
+      end
```
ttoomey: minor, maybe a little easer to read with early-exit (not tested)
```suggestion
      # For augmentations, we explicitly don't update the source hash since that would cause future
      # imports to the augmented class to appear modified
      # We also don't set the source_id as it would cause the source data drill-down to break
      return record if self.class.augments?

      # For new_table, set the source hash, source ID, and data source ID.
      record.source_hash = source_hash
      # Note which record we're sending this from for error checking
      record.source_id = id
      # For non-augmentations, we need to set the data_source_id
      record.data_source_id = data_source_id

      record
```

ttoomey:
Generated flowchart to help me better understand how the components fit together, sharing for other reviewers:
```mermaid

graph TD
    subgraph "A. One-Time Setup (Code Generation)"
        direction LR
        YAML[".yaml Config File<br/>(Defines file structure, rules, and mappings)"] -- "Is read by" --> CFM("CustomFileManager.rb<br/>(The code generator)")
        CFM -- "Generates" --> GenModels["Generated Models<br/>(e.g., CustomGender.rb)"]
        CFM -- "Generates" --> Migrations["Database Migrations"]
    end

    subgraph "B. Runtime Import Process"
        Upload["User uploads CSVs in a Zip"] --> Importer("HmisCsvImporter::Importer<br/>(Main import orchestrator)")
        
        Importer -- "Uses generated model for custom file" --> GenModels
        
        GenModels -- "Includes shared logic from" --> CIC("CustomImportConcern.rb")
        
        CIC -- "Uses for data transformation" --> CM("ColumnMapper.rb")
        
        CM -- "Reads mapping rules from" --> YAML
        
        CIC -- "Returns processed data to" --> Importer
        
        Importer -- "Is this data augmenting<br/>an existing table?" --> Decision{"Check: Augmentation or New Data"}
        
        Decision -- "Yes (e.g., CustomGender)" --> BulkUpdate["Efficiently Bulk Update<br/>existing records in DB"]
        Decision -- "No (e.g., CustomDataElement)" --> Insert["Insert new records<br/>into a custom table in DB"]
        
        BulkUpdate --> DB[("Warehouse Database")]
        Insert --> DB
    end


    class YAML config
    class CFM,Importer,CIC,CM process
    class GenModels,Migrations,Upload,Decision,BulkUpdate,Insert,DB component

```

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/column_mapper.rb
```diff
@@ -0,0 +1,538 @@
...
+  #
+  # == Error Handling
+  #
+  # The mapper handles various error conditions gracefully:
+  # - Unknown mapping types are logged as warnings.
+  # - Invalid data conversions return nil or default values.
+  #
+  # @see CustomImportConcern For integration with importer classes.
+  # @see CustomFileManager For YAML configuration loading.
+  class ColumnMapper
```
ttoomey, line 73: This works. Rather than using stateless class methods, I'd suggest using a service object pattern which avoid having to pass around record and config.

```ruby
# In: .../importer/custom/column_mapper.rb

class ColumnMapper
  # Public API remains the same (or rename to `call()`)
  def self.apply_mappings(record:, config:, **)
    new(record, config, **).apply_mappings
  end

  def initialize(record, config, **)
    @record = record
    @config = config
    # ...
  end

  def apply_mappings
    # ... logic ...
    apply_direct_mapping(column_config)
  end

  private

  # Private methods are cleaner, using instance state
  def apply_direct_mapping(column_config)
    # ... logic using `@record` instance variable ...
  end
end
```

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/custom_file_manager.rb
```diff
@@ -0,0 +1,357 @@
...
+  #           type: "integer"
+  #           warehouse_column_mapping:
+  #             type: "direct"
+  #             target_column: "Woman"
+  #
+  # @example Usage
+  #   # Generate all custom models and migrations
+  #   HmisCsvTwentyTwentySix::CustomFileManager.bootstrap_custom_models!
+  #
+  class CustomFileManager
```
ttoomey: This is another candidate for refactoring into a service-object pattern
```ruby
# In: .../models/hmis_csv_twenty_twenty_six/custom_file_manager.rb

class CustomFileManager
  # The public interface orchestrates the creation of service objects
  def self.bootstrap_custom_models!
    custom_files.each do |file_config|
      new(file_config).generate_all_files
    end
  end

  def initialize(file_config)
    @file_config = file_config
    @class_name = @file_config['class_name']
    # ... other instance variables derived from the config
  end

  def generate_all_files
    generate_migration
    generate_loader_model
    generate_importer_model
  end

  private

  # Private methods are now cleaner and use instance state
  def generate_loader_model
    # ... logic using `@file_config` and `@class_name` ...
  end
end
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -967,6 +990,43 @@ def involved_project_ids
       @importer_log.import_errors.import(errors)
     end
 
+    # Use postgres bulk update for performance
+    # This will be used for data that is augmenting an existing warehouse class
+    # that never needs to add new rows.
+    # NOTE: data that augments the client table is handled differently
+    # as that table does not currently have a uniqueness constraint compatible
+    # with this method
+    private def bulk_update_batch!(klass, batch, columns)
```
ttoomey, line 1009: Looking good but this class is getting pretty wide. Maybe some pieces could be extracted in a future phase. AI says the following might be candidates for extraction, ymmv:

```
NewRecordCreator (handles the add_new_data logic)
ExistingRecordUpdater (handles the update_existing_data logic)
RecordDeleter (handles the delete_missing_data logic)
UnchangedRecordMarker (handles the mark_unchanged logic)
```

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six.rb
```diff
@@ -40,28 +47,119 @@ def self.importable_files_map
     }.freeze
   end
 
+  def self.custom_files_config
```
ttoomey, line 50: There's a lot of references file config keys. Maybe this would benefit from a lightweight wrapper class to avoid typos and hold common behaviors?

This might replace the CustomFilesConfig which doesn't seem to be doing all that much and could just be an array of `CustomFileDefinition`

AI suggests:
```ruby
  # Immutable object that provides an API for accessing custom file configuration
  CustomFileDefinition = Struct.new(
    :filename,
    :class_name,
    :description,
    :required,
    :augments_warehouse_table,
    :augment_key,
    :warehouse_class_name,
    :warehouse_key,
    :columns,
    keyword_init: true
  ) do

    # Returns the primary key field used for identifying records.
    def hud_key
      augment_key || warehouse_key || columns.first['name']
    end

    # Is this file required for a valid import?
    def required?
      required
    end

    # Does this file augment an existing warehouse table?
    def augments?
      augments_warehouse_table.present?
    end

    # Returns the ActiveRecord model class for the destination warehouse table.
    def warehouse_class
      (augments_warehouse_table || warehouse_class_name)&.safe_constantize
    end

    # Returns the fully qualified name of the importer model class.
    # e.g., "HmisCsvTwentyTwentySix::Importer::Custom::CustomGender"
    def importer_class_name
      "HmisCsvTwentyTwentySix::Importer::Custom::#{class_name}"
    end
    
    # Returns an array of column names (as strings) that are expected in the CSV header.
    def csv_headers
      columns.reject { |col| col['type'] == 'virtual' }.map { |col| col['name'] }
    end
  end
  # probably can be private?
  private_constant :CustomFileDefinition
```
```ruby
  def self.custom_files_config
    @custom_files_config ||= begin
      # ... logic to load YAML ...
      all_custom_files = # ... array of hashes from YAML
      
      # Convert array of hashes into an array of Structs
      all_custom_files.map { |file_hash| CustomFileDefinition.new(**file_hash.deep_symbolize_keys) }
    end
  end

```

ttoomey (commented):
Looking good! Left some suggestions

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_validation/unique_hud_key.rb
```diff
@@ -13,6 +15,8 @@ def self.check_validity!(klass, importer_log)
       group(klass.hud_key).
       having(nf('COUNT', [klass.hud_key]).gt(1)).
       count
+
+    return [] if incorrect_counts.is_a?(Integer) && incorrect_counts.zero?
```
ttoomey, line 19: I guess this can be nil somehow? Probably clearer to say:

```suggestion
    return [] if incorrect_counts&.zero?
```

db/warehouse/migrate/20250714172716_add_hud_fields_to_custom_data_element_definitions.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+class AddHudFieldsToCustomDataElementDefinitions < ActiveRecord::Migration[7.1]
+  def change
```
ttoomey, line 4: Do we have an immediate use case for importing CDE's? Just checking before we add all these cols.

While CustomDataElementDefinitions is small, CustomDataElements is large-ish.

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/custom_file_manager.rb
```diff
@@ -0,0 +1,357 @@
...
+    # @param importer_exists [Boolean] Whether importer table exists
+    # @param migration_name [String] Name of the migration class
+    # @return [String] Migration file content
+    # @private
+    private_class_method def self.generate_migration_content(file_config:, loader_table:, importer_table:, loader_exists:, importer_exists:, migration_name:)
+      class_name = migration_name.camelize
+
+      content = <<~MIGRATION
+        # frozen_string_literal: true
+
```
ttoomey:
```suggestion
        # THIS FILE IS AUTOGENERATED.
```

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six.rb
eanders: I believe that's expected.  The import should not fail if the custom config files aren't there or are malformed.  It should be clear in the importer what was imported.

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/custom_import_concern.rb
eanders: Looks good to me.

ttoomey (commented):
A little additional feedback I thought I submitted earlier, sorry for the delay

db/warehouse/migrate/20250714172716_add_hud_fields_to_custom_data_element_definitions.rb
eanders: We do.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_validation/unique_hud_key.rb
eanders: This weird check is a result of originally having what you suggest.  Sometimes it appears to return zero, sometimes it returns an array...

drivers/hmis_csv_twenty_twenty_six/spec/models/importer/custom_data_element_integration_spec.rb
```diff
@@ -0,0 +1,439 @@
...
+      expect(mapped_attributes['CustomDataElementID']).to eq('CDE001')
+      expect(mapped_attributes['CustomDataElementDefinitionID']).to eq('DEF123')
+      expect(mapped_attributes['value_string']).to eq('test_value')
+      expect(mapped_attributes['DataCollectionStage']).to eq(1)
+      expect(mapped_attributes['UserID']).to eq('user123')
+      expect(mapped_attributes['ExportID']).to eq('export456')
+      expect(mapped_attributes['DateDeleted']).to be_nil
+
+      # Test dates with more flexible matching - allow both Date and DateTime
+      expect(mapped_attributes['InformationDate']).to be_a_kind_of(Time) # ActiveRecord often converts to Time/DateTime
```
ttoomey: this one fails for me even after recreating my db. I get a date, not a time. Maybe:

```suggestion
            expect(mapped_attributes['InformationDate']).to satisfy { |v| v.acts_like?(:date) }
```

Same issue with the assertion on line 402

drivers/hmis_csv_twenty_twenty_six/spec/models/importer/custom_data_element_integration_spec.rb
```diff
@@ -0,0 +1,439 @@
...
+
+      edge_cases.each do |test_case|
+        test_record = source_record.dup
+        test_case[:modify]&.call(test_record)
+
+        mapped_attributes = mapper.map(test_record)
+
+        expect(mapped_attributes['owner_id']).to eq(test_case[:expected_owner_id]) if test_case.key?(:expected_owner_id)
+
+        expect(mapped_attributes['Value']).to eq(test_case[:expected_value]) if test_case.key?(:expected_value)
```
ttoomey: should this be 
```
        expect(mapped_attributes['value_string']).to eq(test_case[:expected_value]) if test_case.key?(:expected_value)
```

eanders:
@ttoomey and @dtgreiner I believe this is ready for a final review.

drivers/hmis_csv_twenty_twenty_six/README.md
```diff
@@ -1,70 +1,182 @@
...
+You can define how data from a source column is mapped to the warehouse.
 
-Row validation classes:
-* `HmisCsvValidation::Validation`: Failed checks are logged for later reference, but do not otherwise
-affect row processing.
-* `HmisCsvValidation::Error`: Failed checks are logged for later reference, and the row is excluded from the import.
+*   **Direct Mapping**: Maps the source value directly to a target column.
+*   **Value-Based Multi-Column Mapping**: Maps different source values to different target columns.
+*   **Concatenation Mapping**: Combines multiple source values into a single target field.
+
```
ttoomey, line 211: add some docs for CDEs since it's a bit more complex. Also include additional mapping types implemented in `apply_standard_mappings`

```suggestion
*   **Record Lookup Mapping**: Looks up foreign keys in other warehouse tables to establish associations.
*   **Static Value Mapping**: Assigns a fixed value to a target column, useful for setting defaults.

#### `CustomDataElement` Import Process

The import process for `CustomDataElement.csv` demonstrates how multiple mapping types are combined to handle complex requirements. The objective is to place the generic `Value` from the CSV into the correct typed `value_*` column in the `CustomDataElements` table (e.g., `value_string`, `value_date`). This is determined by the `field_type` in the associated `CustomDataElementDefinition` record.

The process is as follows:

1.  **`value_mapping` for `owner_type`**: The `RecordType` from the CSV (e.g., "Client", "Enrollment") is mapped to a full model name (`GrdaWarehouse::Hud::Client`).
2.  **`record_lookup` for `owner_id`**: Using the `owner_type` from the previous step, the system performs a lookup to find the correct record by its HMIS key (e.g., `PersonalID` for a Client) and establish the polymorphic `owner` association.
3.  **Dynamic Value Placement**: A custom routine uses the `CustomDataElementDefinitionID` to look up the definition, find its `field_type`, and place the CSV `Value` into the correct `value_string`, `value_integer`, etc., column.
4.  **Foreign Key Association**: The `CustomImportConcern` establishes the direct foreign key link by setting the `data_element_definition_id` on the `CustomDataElement` record.

This configuration-driven approach allows for complex import logic to be handled without requiring unique code for each new file type.

```

drivers/hmis_csv_twenty_twenty_six/spec/models/importer/column_mapper_spec.rb
```diff
@@ -0,0 +1,299 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe HmisCsvTwentyTwentySix::Importer::Custom::ColumnMapper do
```
ttoomey, line 11: improvement ideas:
* source_record_class is repeated, could be at top level
* do we have test coverage for all mapping_types (concat, static, record_lookup)

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/custom_file_definition.rb
```diff
@@ -0,0 +1,129 @@
...
+###
+
+# frozen_string_literal: true
+
+module HmisCsvTwentyTwentySix
+  # Encapsulates the definition and logic for a single custom file configuration
+  class CustomFileDefinition
+    attr_reader :config_data
+
+    def initialize(config_hash)
```
ttoomey: Maybe worth adding schema validation here for config? Could just re-use json schema to defend against invalid config

```suggestion
    SCHEMA_PATH = Rails.root.join('drivers', 'hmis_csv_twenty_twenty_six', 'config', 'custom_file_schema.json').to_s

    def initialize(config_hash)
      # Validate the configuration hash against our schema using the shared validator.
      # The `custom_files` key is a wrapper in the YAML, so we validate the inner hash.
      errors = HmisExternalApis::JsonValidator.perform({ 'custom_files' => [config_hash] }, SCHEMA_PATH)

      if errors.any?
        # Provide a clear, actionable error message if validation fails.
        error_message = errors.join("\n")
        raise "Invalid custom file configuration for '#{config_hash['filename']}':\n#{error_message}"
      end
```

Then AI suggests the following schema config:
```json

{
  "$id": "open-path-schemas/csv-2026-custom-file-schema.json",
  "$schema": "http://json-schema.org/draft-06/schema#",
  "additionalProperties": false,
  "title": "HMIS CSV Twenty Twenty Six Custom File Configuration",
  "type": "object",
  "properties": {
    "custom_files": {
      "type": "array",
      "description": "Array of custom file definitions",
      "items": {
        "$ref": "#/definitions/customFileDefinition"
      }
    }
  },
  "required": ["custom_files"],
  "definitions": {
    "customFileDefinition": {
      "type": "object",
      "description": "Definition for a single custom CSV file",
      "properties": {
        "filename": {
          "type": "string",
          "description": "The CSV filename (e.g., 'CustomGender.csv')",
          "pattern": "^[A-Za-z0-9_]+\\.csv$"
        },
        "class_name": {
          "type": "string",
          "description": "The class name to generate (e.g., 'CustomGender')",
          "pattern": "^[A-Z][A-Za-z0-9]*$"
        },
        "required": {
          "type": "boolean",
          "description": "Whether this file is required in imports",
          "default": false
        },
        "description": {
          "type": "string",
          "description": "Human-readable description of the file's purpose"
        },
        "warehouse_key": {
          "type": "string",
          "description": "Primary key field for warehouse tables (for new tables)"
        },
        "warehouse_class_name": {
          "type": "string",
          "description": "Full class name for warehouse model (for new tables)",
          "pattern": "^[A-Z][A-Za-z0-9:]*$"
        },
        "augments_warehouse_table": {
          "type": "string",
          "description": "Full class name of existing warehouse table to augment",
          "pattern": "^[A-Z][A-Za-z0-9:]*$"
        },
        "augment_key": {
          "type": "string",
          "description": "Key field to use for augmentation lookups"
        },
        "augment_import_class": {
          "type": "string",
          "description": "Full class name of the importer to use for augmentation",
          "pattern": "^[A-Z][A-Za-z0-9:]*$"
        },
        "columns": {
          "type": "array",
          "description": "Column definitions for the CSV file",
          "items": {
            "$ref": "#/definitions/columnDefinition"
          },
          "minItems": 1
        }
      },
      "required": ["filename", "class_name", "columns"],
      "anyOf": [
        {
          "description": "Augmentation file configuration",
          "required": ["augments_warehouse_table", "augment_key"]
        },
        {
          "description": "New warehouse table configuration",
          "required": ["warehouse_class_name"]
        }
      ]
    },
    "columnDefinition": {
      "type": "object",
      "description": "Definition for a single column in the CSV file",
      "properties": {
        "name": {
          "type": "string",
          "description": "Column name as it appears in the CSV header",
          "pattern": "^[A-Za-z_][A-Za-z0-9_]*$"
        },
        "type": {
          "type": "string",
          "description": "Data type for the column",
          "enum": ["string", "integer", "boolean", "date", "datetime", "virtual"]
        },
        "required": {
          "type": "boolean",
          "description": "Whether this column is required (cannot be empty)",
          "default": false
        },
        "max_length": {
          "type": "integer",
          "description": "Maximum length for string columns",
          "minimum": 1
        },
        "validations": {
          "type": "array",
          "description": "List of validation rules to apply",
          "items": {
            "oneOf": [
              {
                "type": "string",
                "description": "String validation (e.g., 'NonBlank' or Rails validation syntax)"
              },
              {
                "type": "object",
                "description": "Complex validation rule",
                "properties": {
                  "class": {
                    "type": "string",
                    "description": "Validation class name"
                  },
                  "arguments": {
                    "type": "object",
                    "description": "Arguments for the validation"
                  }
                },
                "required": ["class"]
              }
            ]
          }
        },
        "warehouse_column_mapping": {
          "$ref": "#/definitions/warehouseColumnMapping"
        }
      },
      "required": ["name", "type"]
    },
    "warehouseColumnMapping": {
      "type": "object",
      "description": "Configuration for how to map this column to the warehouse",
      "properties": {
        "type": {
          "type": "string",
          "description": "Type of mapping to apply",
          "enum": ["direct", "value_mapping", "value_based_multi_column", "concatenation", "record_lookup", "static_value"]
        },
        "target_column": {
          "type": "string",
          "description": "Target column name in the warehouse table"
        },
        "value_mappings": {
          "oneOf": [
            {
              "type": "object",
              "description": "Simple value mapping (key-value pairs)",
              "additionalProperties": {
                "type": ["string", "number", "boolean"]
              }
            },
            {
              "type": "array",
              "description": "Complex value mappings for multi-column",
              "items": {
                "type": "object",
                "properties": {
                  "condition": {
                    "type": "object",
                    "properties": {
                      "value": {
                        "type": ["string", "number", "boolean"]
                      }
                    },
                    "required": ["value"]
                  },
                  "target_column": {
                    "type": "string"
                  },
                  "target_value": {
                    "type": ["string", "number", "boolean"]
                  }
                },
                "required": ["condition", "target_column", "target_value"]
              }
            }
          ]
        },
        "separator": {
          "type": "string",
          "description": "Separator for concatenation mapping",
          "default": " "
        },
        "class_column": {
          "type": "string",
          "description": "Column that contains the class name for record lookup"
        },
        "lookup_field_mappings": {
          "type": "object",
          "description": "Mapping of class names to lookup fields for record lookup",
          "additionalProperties": {
            "type": "string"
          }
        },
        "value": {
          "type": ["string", "number", "boolean"],
          "description": "Static value to assign for static_value mapping"
        }
      },
      "required": ["type"],
      "allOf": [
        {
          "if": {
            "properties": {
              "type": {
                "const": "direct"
              }
            }
          },
          "then": {
            "required": ["target_column"]
          }
        },
        {
          "if": {
            "properties": {
              "type": {
                "const": "value_mapping"
              }
            }
          },
          "then": {
            "required": ["target_column", "value_mappings"]
          }
        },
        {
          "if": {
            "properties": {
              "type": {
                "const": "value_based_multi_column"
              }
            }
          },
          "then": {
            "required": ["value_mappings"]
          }
        },
        {
          "if": {
            "properties": {
              "type": {
                "const": "concatenation"
              }
            }
          },
          "then": {
            "required": ["target_column"]
          }
        },
        {
          "if": {
            "properties": {
              "type": {
                "const": "record_lookup"
              }
            }
          },
          "then": {
            "required": ["class_column", "target_column", "lookup_field_mappings"]
          }
        },
        {
          "if": {
            "properties": {
              "type": {
                "const": "static_value"
              }
            }
          },
          "then": {
            "required": ["target_column", "value"]
          }
        }
      ]
    }
  }
}

```

drivers/hmis_csv_twenty_twenty_six/config/custom/custom_data_element.yaml
```diff
@@ -0,0 +1,73 @@
...
+          lookup_field_mappings: # note: these can probably be pulled from hud_key
+            "GrdaWarehouse::Hud::Client": "PersonalID"
+            "GrdaWarehouse::Hud::Enrollment": "EnrollmentID"
+            "GrdaWarehouse::Hud::Exit": "ExitID"
+            "GrdaWarehouse::Hud::Project": "ProjectID"
+            "GrdaWarehouse::Hud::Organization": "OrganizationID"
+      - name: "Value"
+        type: "string"
+        required: false
+        warehouse_column_mapping:
```
ttoomey, line 53: missing type?
```suggestion
        warehouse_column_mapping:
          type: "direct"
```

ttoomey (approved):
Looks great! Left some suggestions

drivers/hmis_csv_twenty_twenty_six/config/custom/custom_data_element.yaml
eanders: type defaults to `direct`, but no harm in adding.

drivers/hmis_csv_twenty_twenty_six/spec/models/importer/custom_data_element_integration_spec.rb
eanders: Hmm, maybe I hadn't pushed my changes for this one.  it now checks:
```
expect(mapped_attributes['InformationDate']).to be_a_kind_of(Date) # ActiveRecord often converts to Date
```

Which is working for me.

ttoomey: yeah sorry this works for me now, I probably had an old version. Thanks

dtgreiner (approved):
Looks great to me!

PR #5561, opened 2025-07-10
Hotfix for ACL Audits

dtgreiner:
## _Merging this PR_

## Description

Add batching to the ACL audit hisotry feature. This should help out with the timeout showing up for Access Control history.

## Type of change

Bug fix

### DIFF
--- a/app/models/audit/display_item.rb
+++ b/app/models/audit/display_item.rb
@@ -52,7 +52,7 @@ class Audit::DisplayItem
   end
 
   # Class method to create multiple display items with optimized loading
-  def self.batch_create(versions, users_by_id, excluded_fields = [])
+  def self.build_batch(versions, users_by_id, excluded_fields = [])
     return [] if versions.blank?
 
     # Pre-load all records referenced in versions
--- a/app/models/audit/display_item.rb
+++ b/app/models/audit/display_item.rb
@@ -11,11 +11,12 @@ class Audit::DisplayItem
   attr_reader :version, :username, :impersonating, :error, :changes, :entity_name, :entity_display_name
   delegate :created_at, :event, :item_type, :item_id, to: :version
 
-  def initialize(version, users_by_id, excluded_fields = [])
+  def initialize(version, users_by_id, excluded_fields = [], records_by_type_and_id = {})
     raise ArgumentError, "Expected GrPaperTrail::Version or GrdaWarehouse::Version, got #{version.class}" unless version.is_a?(GrPaperTrail::Version) || version.is_a?(GrdaWarehouse::Version)
 
     @version = version
     @excluded_fields = excluded_fields
+    @records_by_type_and_id = records_by_type_and_id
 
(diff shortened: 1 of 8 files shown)


app/models/audit/display_item.rb
```diff
@@ -50,6 +51,45 @@ def initialize(version, users_by_id, excluded_fields = [])
     end
   end
 
+  # Class method to create multiple display items with optimized loading
+  def self.batch_create(versions, users_by_id, excluded_fields = [])
```
eanders: Can we rename this to something like `build_batch`.  Create has a connotation that would imply it will be saving something to the database, I read `batch_create` the same as the bulk `import` method.

Since these are just "newing" up a batch, `build` feels like a more appropriate verb to me.

app/models/audit/versions.rb
```diff
@@ -62,8 +62,145 @@ def version_array
 
   def wrap_display_versions(versions)
     users_by_id = build_user_lookup(versions)
-    versions.map do |version|
-      Audit::DisplayItem.new(version, users_by_id, config[:excluded_fields])
+    Audit::DisplayItem.batch_create(versions, users_by_id, config[:excluded_fields])
+  end
+
+  # Class method to create multiple Audit::Versions objects with optimized loading
+  def self.batch_create(records, config = {})
```
eanders: Here too, I think `build_batch` better captures the intent.

eanders (approved):
Looks good, I have not actually tested it, but before merging, I think we should rename the two methods.

dtgreiner:
I'll get a rename checked in and will push the feature branch out to a staging environment before merging it into 170. It's running more quickly locally, so I'm hopeful it will process quickly enough when deployed. if not, we may need to look at backgrounding the job.

dtgreiner:
@eanders I ended up backgrounding the access control audit history. It was still timing out with the batching.

app/models/audit/display_item.rb
```diff
@@ -68,22 +108,13 @@ def compute_entity_name(klass)
...
-        if record.respond_to?(:entity_name)
-          return record.entity_name
-        elsif record.respond_to?(:name)
-          return record.name
-        end
-      rescue StandardError
-        # If we can't load the record, fall back to humanized entity type
-      end
-    end
+    record = @records_by_type_and_id[version.item_type][version.item_id]
```
eanders: Is there ever a situation where `@records_by_type_and_id[version.item_type]` would return `nil`, if so, we should probably rewrite with tries or `dig`

eanders (approved):

app/models/audit/display_item.rb
dtgreiner: Just updated for safe navigation [here](https://github.com/greenriver/hmis-warehouse/pull/5561/commits/daa61282f499aa9bd944ca20a70a17b8233b662e)

eanders (approved):

PR #5563, opened 2025-07-11
Integrate with AHA API

martha:
## _Merging this PR_

## Description
- Add `ApiKey` as a new `RemoteCredential` type
- Add new `ApiKeyConnection`. Similar to `OauthClientConnection`, it abstracts the authentication/request logic for api connection types that use an authorization header.
- Refactor `OauthClientConnection` and related classes (logger, result) to share code with new `ApiKeyConnection`
- Add `FetchAhaScore` mutation
- Add `AHA` form component

## Type of change
new feature (custom client build)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,59 @@
...
+
+      result = conn.post('api/v1/clients/scores', payload).
+        then { |r| handle_error(r) }
+
+      return nil if result.http_status == 404
+
+      # Find first score that's not -999, return nil if all are -999 or no data
+      scores = result.parsed_body&.dig('data')&.map { |c| c.dig('score') } || []
+      scores = scores.compact.uniq.filter { |s| s != -999 }
+      scores.first || nil # If multiple non -999 scores were returned, just return the first one
```
martha: Spec question: What to do when multiple scores are returned

drivers/hmis_external_apis/app/models/hmis_external_apis/api_key_connection.rb
martha, line 1: Code style thought: I called this "connection" for parallel structure with the `OauthClientConnection`, but it's not really a "connection" in the same way, maybe it is more of a "connector".

drivers/hmis_external_apis/app/models/hmis_external_apis/api_key_connection.rb
```diff
@@ -0,0 +1,92 @@
...
+      end
+    end
+
+    def create_result(result, verb, url, merged_headers, request_log)
+      # result is a Faraday::Response
+      BaseResult.new(
+        body: result.body,
+        content_type: result.headers['content-type'],
+        error: nil,
+        error_type: nil,
```
martha, line 71: In parallel with the `OauthClientConnection` class, this doesn't fill in the `error` fields when there's an http error code returned, only when there is a connection error.

drivers/hmis_external_apis/app/models/hmis_external_apis/base_logger.rb
```diff
@@ -21,15 +23,18 @@ def capture(creds:, url:, method:, payload:, headers:)
...
       rescue StandardError => e
-        # If an exception occurred, it was on or and or a failure to connect. Storing "400" status code to indicate it was on our side (the client)
+        # If an exception occurred, it was a failure to connect. Storing "400" status code to indicate it was on our side (the client)
         update_log_record(record, { response: "#{e.class.name}: #{e.message || 'Unknown error'}", http_status: 400 })
         raise
       end
 
       if result
+        # result is either a Faraday::Response or OAuth2::Response
+        content_type = result.respond_to?(:content_type) ? result.content_type : result.headers['content-type']
```
martha: Code style question: Maybe there is a more graceful way to refactor this so that the present code in BaseLogger doesn't need to know about the two possible types of Response objects.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+      return nil if mci_ids.empty?
+
+      payload = {
+        'mc_id__in': mci_ids,
+      }
+
+      result = conn.post('api/v1/clients/scores', payload).
+        then { |r| handle_error(r) }
+
+      return nil if result.http_status == 404
```
gigxz: Add comment- does this mean client was not found? Want to be careful to not swallow server errors. For example in testing I think I was getting 404 resource not found because the API endpoint wasn't up yet. Maybe once it's working we will have a clearer way to differentiate the "no results" response.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Find and return highest score
+      scores = result.parsed_body&.dig('data')&.map { |c| c.dig('score') } || []
+
+      # Validate that all scores are numerical
+      non_numerical_scores = scores.compact.reject { |score| score.is_a?(Numeric) }
```
gigxz: can use `compact_blank` here and below (or once `scores.compact_blank!`) to drop empty strings too, just in case. (It will retain zeroes).

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+      raise(Error, "Received non-numerical scores: #{non_numerical_scores.inspect}") if non_numerical_scores.any?
+
+      highest_score = scores.compact.uniq.max
+      highest_score == -999 ? nil : highest_score # Special case for -999, which indicates no data
+    end
+
+    def self.enabled?
+      # Both MCI and AHA credentials need to exist for AHA api to be enabled, since we query the API by MCI ID.
+      aha_cred = ::GrdaWarehouse::RemoteCredentials::ApiKey.active.where(slug: SYSTEM_ID)
+      mci_cred = ::GrdaWarehouse::RemoteCredentials::Oauth.active.where(slug: HmisExternalApis::AcHmis::Mci::SYSTEM_ID)
```
gigxz: This doesn't feel _totally_ necessary, and might make the dev/testing setup more cumbersome than it needs to be? Since this doesn't make any calls to the MCI API, all we need for testing is to have some MCI ExternalIDs in the database (I think).

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+
+      # Find and return highest score
+      scores = result.parsed_body&.dig('data')&.map { |c| c.dig('score') } || []
+
+      # Validate that all scores are numerical
+      non_numerical_scores = scores.compact.reject { |score| score.is_a?(Numeric) }
+      raise(Error, "Received non-numerical scores: #{non_numerical_scores.inspect}") if non_numerical_scores.any?
+
+      highest_score = scores.compact.uniq.max
+      highest_score == -999 ? nil : highest_score # Special case for -999, which indicates no data
```
gigxz: Maybe we should blanket check for any negative value and translate that to nil. Just to have more cushion in case they decide to use -1 or something. (Speaking from having little things like that change under us, its a pain!)

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    private
+
+    def creds
+      @creds ||= ::GrdaWarehouse::RemoteCredentials::ApiKey.active.where(slug: SYSTEM_ID).first!
+    end
+
+    def conn
+      @conn ||= HmisExternalApis::ApiKeyConnection.new(creds)
```
gigxz: Looks like this is where we'll want to pass `connection_timeout` if we need to bump it. I'd suggest starting with an increased timeout in the staging environment since we've seen performance issues in the test environments before. https://github.com/greenriver/hmis-warehouse/blob/21f5326cc8147fc160f6604987565594fd943213/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/link_api.rb#L24

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+
+    def handle_error(result)
+      Rails.logger.error "AHA Error: #{result.error}" if result.error
+      Sentry.capture_exception(StandardError.new(result.error)) if result.error
+      raise(Error, result.error) if result.error
```
gigxz: Will this double-log to sentry? I think raising the exception _should_ suffice..

drivers/hmis_external_apis/app/models/hmis_external_apis/base_logger.rb
gigxz: This seems ok to me since its a pretty minor difference. BaseLogger feels too generic as class name though,  maybe ExternalApiLogger / ApiClientLogger / ApiLogger?

drivers/hmis_external_apis/app/models/hmis_external_apis/oauth_client_connection.rb
```diff
@@ -6,31 +8,14 @@
 
 module HmisExternalApis
   # https://gitlab.com/oauth-xx/oauth2/
-  class OauthClientConnection
-    attr_accessor :creds, :client_id, :scope, :headers, :base_url, :connection_timeout, :logger
+  class OauthClientConnection < BaseConnection
```
gigxz: thanks for tackling this to reduce duplication!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
martha: Hm, maybe the code I was modeling this off of https://github.com/greenriver/hmis-warehouse/blob/57df21f18fc931d177a8a4bcf64a6964354a7257/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mci.rb#L176-L180 is outdated?

gigxz: yes I think so

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -0,0 +1,87 @@
...
+          OpenStruct.new(
+            score: score_value.to_i,
+            alt_aha_flag: score_obj.dig('metadata', 'alt_aha_flag')&.to_i == 1,
+            dw_client_id: response_client.dig('dw_client_id'),
+            generator: score_obj.dig('generator'),
+          )
+        end || []
+      end || []
+
+      highest_score_object = score_objects.max_by(&:score)
```
gigxz: The `|| []` feels not railsy, I think it would be the same to do `score_objects.compact.max_by(&:score)`?

gigxz (approved):
@martha this looks great! lets merge into release-173 so we can start testing in staging?
We also want to make sure to QA the MCI API extensively (or another API that uses oauth client) since there was some code churn, as well as ensuring it still logs to ExternalRequestLog in the same way.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
martha: I think I added this because the array itself can be nil if `result.parsed_body&.dig('data')` is nil. I can try to clean it up a bit.

drivers/hmis/app/graphql/mutations/ac_hmis/fetch_aha_score.rb
```diff
@@ -0,0 +1,42 @@
...
+      client = Hmis::Hud::Client.viewable_by(current_user).find_by(id: client_id)
+      access_denied! unless client.present?
+
+      aha = HmisExternalApis::AcHmis::Aha.new
+      result = aha.fetch_score(client)
+
+      return { score: nil } if result.nil?
+
+      {
+        score: result.score,
```
gigxz, line 35: maybe return -1 in this case to represent score not found? per frontend comment https://github.com/greenriver/hmis-frontend/pull/1202 it does make me a bit nervous to have js trying to differentiate between `0` (valid score) and `nil` (score not found), so -1 feels like it may be less bug-prone

gigxz: (comment for line 32)

PR #5566, opened 2025-07-14
Referral and Custom Assessment DW exports

gigxz:
## _Merging this PR_

## Description
This PR partially completes https://github.com/open-path/Green-River/issues/7813, we still need to add some additional files (ce candidates and candidate events) and then uncomment the lines in `DataWarehouseUploadJob` to turn on the exports.

## Type of change
New feature / custom integration


drivers/hmis_external_apis/app/jobs/hmis_external_apis/ac_hmis/data_warehouse_upload_job.rb
```diff
@@ -236,5 +239,65 @@ def case_note_export
 
       uploader.run!
     end
+
+    def ce_referrals
+      export = HmisExternalApis::AcHmis::Exporters::CeReferralExport.new
+      export.run!
+      # File.open('CeReferrals.csv', 'w') do |file|
+      #   file.write(export.output.string)
+      # end
```
martha, line 248: Looks like these comments are to help with local debugging but can be removed before merge?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/ce_referral_export.rb
```diff
@@ -0,0 +1,93 @@
...
+
+      write_row(columns)
+      total = referrals.count
+
+      Rails.logger.info "There are #{total} referrals to export"
+
+      referrals.find_each.with_index do |referral, i|
+        Rails.logger.info "Processed #{i} of #{total}" if (i % 1000).zero?
+        warehouse_id = referral.client.warehouse_id
+        next unless warehouse_id.present?
```
martha, line 28: Add comment here about why we skip (rather than raise) if there is no warehouse ID on the client? I think this is unexpected, right?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/ce_referral_task_export.rb
```diff
@@ -0,0 +1,71 @@
...
+      total = referral_tasks.count
+
+      Rails.logger.info "There are #{total} referral tasks to export"
+
+      referral_tasks.find_each.with_index do |task, i|
+        Rails.logger.info "Processed #{i} of #{total}" if (i % 1000).zero?
+        referral_id = instance_id_to_referral_id[task.instance_id]
+        raise "Referral not found for instance id '#{task.instance_id}'" unless referral_id
+
+        completed_by_hud_user_pk = to_hud_user_pk(task.updated_by) if task.updated_by
```
martha:
```suggestion
        updated_by_hud_user_pk = to_hud_user_pk(task.updated_by) if task.updated_by
```
unless there is something I'm missing?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/csv_exporter.rb
```diff
@@ -15,6 +17,25 @@ def initialize(output = StringIO.new)
       self.output = output
     end
 
+    # Input: Hmis::User record
+    # Output: `id` from `User` table, which matches exported User.csv file
```
martha:
```suggestion
    # Input: `Hmis::User` record (representing an application user in the `users` table)
    # Output: `id` of `Hmis::Hud::User` record from the `User` table, which matches exported User.csv file
```
this stuff always trips me up, more comments may be helpful to future eyes?

martha: Also maybe helpful to add some additional comments about why we don't use the existing `from_user` functionality, if you don't think it's too obvious. My understanding is that it's because `from_user` creates a Hud user if there isn't one, whereas in the export we should never be creating users.

martha (approved):
Looks good!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/ce_referral_export.rb
gigxz: Added a comment to explain:
```
        # Skip if no destination client, because receiver will be unable to import this without a client id.
        # This could occur if a new client was created and directly referred and exported _before_ IdentifyDuplicates
        # had a chance to create the Destination client (unlikely, but possible)
        next unless warehouse_id.present?
```

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/exporters/csv_exporter.rb
gigxz: thanks, updated, added a comment about not wanting to create new users

PR #5570, opened 2025-07-15
Core Demo Export - Project Comparions Column Shift

dtgreiner:
## _Merging this PR_

## Description

Fix issue in Core Demo export where comparison report projects that did not exist in teh original report were showing up in the columns for the original report instead of the comparison. The export was not accounting for the requried column shift in this scenario.

## Type of change

Bug Fix

### DIFF
--- a/drivers/core_demographics_report/app/models/core_demographics_report/projects.rb
+++ b/drivers/core_demographics_report/app/models/core_demographics_report/projects.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module
   CoreDemographicsReport::Projects
   extend ActiveSupport::Concern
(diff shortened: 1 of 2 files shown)


drivers/core_demographics_report/app/models/core_demographics_report/projects.rb
```diff
@@ -55,15 +57,27 @@ def enrollment_percentage(type)
...
+      # Use project IDs for row keys to ensure same project appears on same row across reports
+      # Add report_index to ensure proper column alignment
       projects.each do |id, proj|
         title = proj[:project_name]
-        rows["_Clients in Projects_data_#{id}"] ||= []
-        rows["_Clients in Projects_data_#{id}"] += [
+        row_key = "_Clients in Projects_data_#{id}"
+        rows[row_key] ||= []
+
+        # If this is not the first report, we need to pad with nil values to align columns
```
eanders, line 74: I think this makes sense.  An alternative would be to ensure each report has the same set of keys right?
Maybe this is better as it will indicate the project wasn't included in the report run (current or prior) by leaving the section empty rather than showing zeros.
If we end up spending more time on this in the future, perhaps we should just put each run on its own tab.

eanders (approved):

drivers/core_demographics_report/app/models/core_demographics_report/projects.rb
dtgreiner: I think if we made sure each report had the same set of keys, we could be losing out on showing the data didn't exist in the report (e.g. the project didn't exist yet or ended in the prior year -- I'm assuming this is useful information). Having the project data as empty may be better than a 0 for this reason. Similarly, splitting out the data to another tab may be more work since it would involve adjusting each of the export types where this specific issue only appeared to be affecting the project data, though it may be nice to do to prevent issues like this from popping up in the future.

PR #5573, opened 2025-07-15
Process Referral Step responses into CDEs

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/7876
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1205

Summary:
* Updated the `SubmitCeReferralStep` mutation to run the FormProcessor against submitted values as custom data elements on the Step. I opted to put it in the mutation rather than in the engine because it feels like something that shouldn't be required for the workflow engine to execute. The engine should have no dependencies on the custom data elements, they are for reporting purposes.
* I opted to update the mutation to accept both formats of values, for ease, rather than implementing a backend translation function
* Resolve custom data elements on the gql CeReferralStep

How to test:
* Rerun ce_define_workflows task
   * comment-in the move-in date pieces from the PR to test that
* Create a referral, walk through tasks. Ensure in the db that CDEs are created for each value entered into Task forms.
* Ensure when you re-open tasks that you can see submitted data

## Type of change
new feature


drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -48,28 +51,36 @@ def create_enrollment(message)
...
+    # This requires a Move-in date item to be on the step form with the following attributes:
+    #    { "link_id": "move_in_date", "type": "DATE", "mapping": { "custom_field_key": "<any field key for storing date value>" } }
+    #
+    # Note: We do a lookup by link_id rather than by mapping:field_name to prevent the FormProcessor from attempting
+    # to process the move_in_date field. That would be another approach, but we decided against it in favor of having an
+    # explicit trigger to populate the collected date onto the move_in_date field. In part because, when you go back and
+    # view a previously submitted task, it should show the Move-in Date as it was when the task was performed, NOT
+    # the current value of the Move-in Date field on the target enrollment. To achieve this, the date recorded on the task
+    # would be additionally stored in the CustomDataElement field.
     def set_move_in_date(message) # rubocop:disable Naming/AccessorMethodName
```
gigxz: This maybe should be a PR comment instead of a code comment, I can clean it up before merging.  I just want to call this out explicitly since it's a revert of a very recent change (https://github.com/open-path/Green-River/issues/7485). My bad on not really thinking ahead, here.

The previous approach would be to define the item as 
`{ "link_id": "whatever", "type": "DATE", "mapping": { "record_type": "ENROLLMENT", "field_name": "moveInDate" } }`.

martha: Thanks! This makes complete sense but I had to reason it through for awhile to get things straight in my head, haha. I don't think the comment needs to be fully cleaned up, it seems helpful to keep a record of our reasoning; but maybe we can clarify something like this:

> Sets the Move-In Date on the target enrollment, based on a date value collected on the step form.
This requires a Move-in date item to be on the step form with the following attributes:
{ "link_id": "move_in_date", "type": "DATE", "mapping": { "custom_field_key": "<any field key>" } }
> - it must have the exact link_id `move_in_date` to trigger this side effect
> - it must additionally save the value to a custom data element

> We chose this approach, rather than using "mapping": { "record_type": "ENROLLMENT", "field_name": "moveInDate" }, because the form should show show the Move-in Date as it was when the task was performed, NOT the current value of the Move-in Date field on the target enrollment.

drivers/hmis/lib/tasks/ce_define_workflows.rake
```diff
@@ -123,8 +123,10 @@ module CeWorkflowBuilder
 
     errors = Hmis::Form::DefinitionValidator.perform(definition, form_def.role, skip_cded_validation: true)
     raise "Form definition #{form_def.identifier} is not valid: #{errors.map(&:full_message)}" if errors.any?
+    raise 'Step Form definition should only collect Custom Data Elements' if form_def.link_id_item_hash.values.find { |item| item.mapping.field_name }
```
martha: Related to the comment thread above, I think this could be a good place to add another brief comment about why step forms shouldn't collect data onto object fields directly. (so we aren't confused and searching around for reasoning that's explained in a different place, when we hit this validation error)

I also wonder if this should be moved into WorkflowTemplateValidator if we expect it to generally apply across workflows? -- I _think_ the reasoning you spelled out above, about wanting to show the task data as it was when the task was filled out, not the current value of the field -- applies generically, not just to referral workflows.

martha (approved):

drivers/hmis/lib/tasks/ce_define_workflows.rake
martha:
```suggestion
    raise 'Step Form definition should only collect Custom Data Elements' if form_def.link_id_item_hash.values.find { |item| item.mapping&.field_name }
```

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
gigxz: Thats clearer I will update, thank you for the suggestion!

drivers/hmis/lib/tasks/ce_define_workflows.rake
gigxz: Great point! I'll add a clarifying comment, I opened https://github.com/open-path/Green-River/issues/7971 for adding additional validations to DefinitionValidator. I don't want to make a decision on the pattern as part of this scope, because it feels potentially impactful, we should also think through what constraints exist on other record types. (For example OCCURRENCE_POINT should never collect onto HealthAndDv, etc...)

PR #5579, opened 2025-07-16
Background ACL Audit Export

dtgreiner:
## _Merging this PR_

## Description

Hotfix to background Access Control Audiit Export due to amount of data causing timeouts.

## Type of change
Bug fix

### DIFF
--- a/app/models/grda_warehouse/document_exports/access_controls_audit_export.rb
+++ b/app/models/grda_warehouse/document_exports/access_controls_audit_export.rb
@@ -28,30 +28,23 @@ module GrdaWarehouse::DocumentExports
     end
 
     private def excel_package
-      # Generate CSV data using existing logic
-      audit_history = []
-      histories = build_histories(user)
-      histories.each_with_index do |history, index|
-        csv_data = generate_audit_csv(history.version_array, history, include_headers: index == 0)
-        audit_history << csv_data
-      end
-
-      # Convert CSV to Excel
-      csv_content = audit_history.join
-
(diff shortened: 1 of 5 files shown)


app/models/grda_warehouse/document_exports/access_controls_audit_export.rb
```diff
@@ -0,0 +1,61 @@
...
+              # Add headers with styling
+              sheet.add_row(row.fields, style: title)
+            else
+              # Add data rows
+              sheet.add_row(row.fields)
+            end
+          end
+        end
+      end
+    end
```
eanders: This is probably fine as is, but requires loading the whole CSV into memory along with the Excel file I think.  Couldn't this be something more like:

```suggestion
     # Generate CSV data using existing logic
      Axlsx::Package.new do |package|
        wb = package.workbook
        wb.add_worksheet(name: 'Access Controls Audit History') do |sheet|
          title = sheet.styles.add_style(sz: 12, b: true, alignment: { horizontal: :center })

          histories = build_histories(user)
          histories.each_with_index do |history, index|
            csv_data = generate_audit_csv(history.version_array, history, include_headers: index == 0)
            sheet.add_row(csv_data)
          end
        end
      end
    end
```

Pretty sure that won't work, and doesn't include the headers, but I think you could generate each row and then immediately add it to the excel, so you wouldn't need both in memory at the same time.

eanders (approved):
Looks good, but might be worth trying to limit memory in the export

eanders (approved):

PR #5582, opened 2025-07-17
Align CE referral status filter labels with existing app language

gigxz:
## _Merging this PR_

## Description
issue: https://github.com/open-path/Green-River/issues/7833
follow-up to https://github.com/greenriver/hmis-warehouse/pull/5556

Align language in the Referral Status dropdown to the rest of the app:
* Rejected => Declined
* Initialized => REMOVED but initialized referrals will be included in the "in progress" filter. We do not have a user-facing initialized state. I don't think we even expect to have referrals in this state at this time.

## Type of change
New feature


drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
```diff
@@ -24,7 +24,11 @@ def filter_scope(scope)
 
   def with_referral_statuses(scope)
     with_filter(scope, :referral_status) do
-      custom_statuses = Hmis::Ce::CustomReferralStatus.where(key: input.referral_status)
+      status_keys = input.referral_status.map(&:to_s)
+      # inject 'initialized' state when filtering by 'in_progress'; initialized is not a user-facing state.
+      # this behavior is mirrored in the CE Referral Status pick list.
+      status_keys << 'initialized' if status_keys.include?('in_progress')
```
gigxz: Actually I don't think this is necessary, since we don't have initialized referrals. I was worried about some referrals getting lost somehow, but they will always show up if this filter is not applied.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -200,6 +200,15 @@ def self.static_options_for_type(pick_list_type, user:)
             label: enum.description,
           }
         end.compact
+      when 'CE_REFERRAL_STATUSES'
```
martha, line 196: Thanks, you are right looks like this should be a static option. I think lines 89-95 should now be removed right?

drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
martha: Yes, I think I agree it's not necessary.

martha (approved):

martha:
Test failure is unrelated, addressed here: https://github.com/greenriver/hmis-warehouse/pull/5583

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: Whoops that was a bad merge, yep!

drivers/hmis/app/models/hmis/filter/ce_referral_filter.rb
gigxz: removed

PR #5585, opened 2025-07-18
Add mutation and support for CE Direct Referrals

martha:
## _Merging this PR_

## Description
Summary of changes
- Add `referral_origin` column to `ce_referrals`.
- ~Add `delegated_handoff` flag to `wfd_nodes` table, to indicate which tasks should be "delegated" to another agent (in this case, the initial referral task that should be delegated to the sending project)~
- Add `direct_referral_entrypoint` relation to `unit_groups` table, to indicate which task should be completed by the user at the sending project
- Add `CreateDirectCeReferral` mutation
- Add picklists:
  - Projects accepting CE referrals
  - Unit groups accepting CE referrals
- Update the existing `project_can_accept_referral` query to work for both CE and legacy referrals
- Add query `direct_referral_form` to fetch the "delegated" form for a direct referral
- Add helper methods and scopes to Project, Unit, and UnitGroup for determining acceptance of direct referrals

## Type of change
New feature

  - [x] ensured this does not introduce N+1s - let's discuss; I think this PR introduces some n+1s on new pick list types.
- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - yes, see ticket


drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
```diff
@@ -0,0 +1,97 @@
...
+            return { errors: errors }
+          end
+
+          instance = opportunity.workflow_template.instances.create!
+          referral = opportunity.referrals.create!(
+            workflow_instance: instance,
+            referred_by: current_user,
+            client: source_enrollment.client,
+            source_enrollment: source_enrollment,
+            referral_origin: 'project',
```
martha: To consider: Should we make this "direct" instead of "project"? (per https://github.com/open-path/Green-River/issues/7961#issuecomment-3075752030)

drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
```diff
@@ -0,0 +1,97 @@
...
+          )
+        end
+
+        engine = referral.workflow_engine
+        engine.start_workflow!(user: current_user)
+
+        # Find the step that's marked 'delegated_handoff', aka the step that should be completed by the target project
+        step = engine.active_steps.find { |s| s.node.delegated_handoff? }
+        raise unless step.present?
```
martha, line 66: Spec question to confirm: Is it a correct assumption that the target project's workflow will _always_ specify a "delegated handoff" step? Or should we be more flexible?

Just thinking it through, I _think_ we want to keep this requirement, for the following reason:
- the config for "accepts direct referrals" is at the project level
- but the workflow template is specified at the unit group level
- if a project wants to accept both direct and waitlist referrals (not needed for mvp), this is a built-in way to differentiate between unit groups that accept direct referrals and those that should be waitlist-initiated.

Maybe what we have here is good enough for mvp, and it's not too difficult to update later if we find there's a need

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -502,13 +502,16 @@ def project_configs(filters: nil)
       scope.order(created_at: :desc, id: :asc)
     end
 
+    # Either legacy or coordinated entry referrals can use this query.
```
martha, line 505: let's discuss, is this change needed/helpful? I was thinking about it from a parity-with-legacy-referrals perspective, but I'm not sure it is a generic need.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -604,5 +610,44 @@ def self.projects_receiving_referrals(data_source_id)
...
+    def self.unit_groups_for_project_ce_referral(project_id:, user:)
+      return [] unless Hmis::Ce.configuration.enabled?
+      return [] unless user.can_manage_outgoing_referrals? # at any project
+
+      project = Hmis::Hud::Project.find(project_id)
+      return [] unless project.data_source_id == user.hmis_data_source_id
+      return [] unless project.accepts_direct_ce_referrals?
+
+      project.unit_groups.filter_map do |unit_group|
+        # this causes n+1, which is acceptable because the number of unit groups per project is expected to be small
```
martha, line 640: > this causes n+1, which is acceptable because the number of unit groups per project is expected to be small

Let's discuss/confirm

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -550,6 +555,39 @@ def ce_referral(id:)
       Hmis::Ce::Referral.viewable_by(current_user).find_by(id: id)
     end
 
+    field :direct_referral_form, Types::Forms::FormDefinition, null: true do
+      argument :target_unit_group_id, ID, required: true
+      argument :source_enrollment_id, ID, required: true
```
martha: Maybe it's unnecessary to require the source enrollment here? I've used it for permission checking only

drivers/hmis/app/graphql/mutations/ce/create_ce_referral.rb
```diff
@@ -42,6 +42,7 @@ def resolve(opportunity_id:, client_id: nil, source_enrollment_id: nil)
           referred_by: current_user,
           client: client,
           source_enrollment: source_enrollment,
+          referral_origin: 'waitlist',
```
gigxz: Suggest to move string to a constant on referral, or (better yet, I think?) making use of a scope to hide the internals of the origin string.
```ruby
scope :originated_from_waitlist, -> { where(referral_origin: 'waitlist') }
...
opportunity.referrals.originated_from_waitlist.create!(...) # no need to specify referral_origin explicitly
```

drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
```diff
@@ -0,0 +1,97 @@
...
+    argument :form_definition_id, ID, required: true # The form that was used to submit the "delegated_handoff" step
+    argument :confirmed, Boolean, required: false
+
+    field :referral, Types::HmisSchema::CeReferral, null: true # nullable in case of validation errors
+
+    def resolve(target_unit_group_id:, source_enrollment_id:, values_by_link_id:, values_by_field_name:, form_definition_id:, confirmed: false)
+      raise unless Hmis::Ce.configuration.enabled?
+
+      source_enrollment = Hmis::Hud::Enrollment.viewable_by(current_user).find(source_enrollment_id)
+      access_denied! unless current_permission?(permission: :can_manage_outgoing_referrals, entity: source_enrollment.project)
```
gigxz: Since we have the HmisProjectPolicy class now, maybe this can be:
```
access_denied! unless policy_for(source_enrollment.project, policy_type: :hmis_project).can_send_direct_referral?
```

drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
```diff
@@ -0,0 +1,97 @@
...
+
+      unit_group = Hmis::UnitGroup.find(target_unit_group_id)
+      target_project = unit_group.project # does not need to be viewable by current user
+      access_denied! unless target_project.accepts_direct_ce_referrals_from?(source_enrollment.project)
+
+      errors = HmisErrors::Errors.new
+
+      unit = unit_group.units.preload(:latest_opportunity).accepting_ce_referrals.first
```
gigxz: I wonder if we can simplify this logic, as well as the scope you add to Unit. I'm finding it a bit hard to follow with the joins on what logic is defined on the opportunity vs on the unit. Can we do something like this?

```ruby
# here
opportunity = unit_group.opportunities.accepting_referrals.first
unit = opportunity&.unit

# on Unit. unit is accepting referrals if it is unoccupied and has any opportunities that are accepting referrals (dont need to include sort logic)
scope :accepting_referrals, -> do
   active.unoccupied_on.joins(:opportunities).merge(Hmis::Ce::Opportunity.accepting_referrals)
end

# on Opportunity. opportunity is accepting referrals if it is "open"
# proposed unit scope also has a left join to active_referral, not sure if we need that I thought having an active referral would make the opportunity have status "locked"?
scope :accepting_referrals, -> do
  where(status: 'open')
end
```

drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
martha: Thanks good idea, I'd like to name it `can_send_out_direct_referral?` though to help clarify that this is the source project.

drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb
martha: Thanks, yeah I think this is a lot clearer!

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -502,13 +502,16 @@ def project_configs(filters: nil)
...
     field :project_can_accept_referral, Boolean, 'Whether the destination project is able to accept a referral for the client(s) belonging to the source enrollment', null: false do
       argument :destination_project_id, ID, required: true
       argument :source_enrollment_id, ID, required: true
+      argument :referral_mode, type: HmisSchema::Enums::ReferralMode, required: false, default_value: 'legacy'
     end
-    def project_can_accept_referral(destination_project_id:, source_enrollment_id:)
+    def project_can_accept_referral(destination_project_id:, source_enrollment_id:, referral_mode:)
       source_enrollment = Hmis::Hud::Enrollment.viewable_by(current_user).find(source_enrollment_id)
-      raise 'access denied' unless current_permission?(permission: :can_manage_outgoing_referrals, entity: source_enrollment.project)
+      access_denied! unless current_permission?(permission: :can_manage_outgoing_referrals, entity: source_enrollment.project)
```
ttoomey: perhaps use a policy check here? Would have to define `can_send_ce_referral?` if it's helpful to have distinct permissions
```suggestion
      policy = policy_for(source_enrollment.project, policy_type: :hmis_project)
      can_send = referral_mode == 'legacy' ? policy.can_send_direct_referral? : policy.can_send_ce_referral?
      access_denied! unless can_send
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -550,6 +555,39 @@ def ce_referral(id:)
       Hmis::Ce::Referral.viewable_by(current_user).find_by(id: id)
     end
 
+    field :direct_referral_form, Types::Forms::FormDefinition, null: true do
```
ttoomey: This works but the root query type is becoming quite large. Perhaps we should consider a strategy to curb its growth. For example we could group referral-related queries into a new `Referrals` type if there's no other natural home for them. This is purely for better code organization.

db/warehouse/migrate/20250716154410_add_delegated_handoff_to_wfd_node.rb
```diff
@@ -0,0 +1,18 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddDelegatedHandoffToWfdNode < ActiveRecord::Migration[7.1]
+  def change
+    # "Delegated handoff" is a flag indicating this node should be handed off to another process as part of the workflow.
+    # For example, in a Referral workflow, this would be a UserTask node that is delegated to the referring Project.
+    add_column :wfd_nodes, :delegated_handoff, :boolean, null: false, default: false
```
ttoomey: I  probably don't fully understand the intention here. But this is trying to indicate the entrypoint for the receiving project? 
```ruby
initiation_node = workflow_template.graph.nodes.find(&:delegated_handoff)
```
If several nodes had this boolean, it's likely you would get non-deterministic behavior. You could remedy it with an order-by clause/and or a unique constraint. But perhaps there's another approach that avoids creating a dependency on referral (we intended to keep workflows generic).

Instead of the boolean, could we consider adding a reference to the entrypoint node? That would be unambiguous and avoid violating the boundary. Perhaps this reference would be on unit_group?
```ruby
# migration
add_reference :hmis_unit_groups, :direct_referral_entrypoint, foreign_key: { to_table: :wfd_nodes }

# use case
initiation_node = unit_group.direct_referral_entrypoint
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Agreed I think we can use the same policy  method martha added elsewhere in this PR: `policy_for(source_enrollment.project, policy_type: :hmis_project).can_send_out_direct_referral?`

ttoomey: For sure, we should re-use the direct referral check. I guess the intent is to check both types of referrals here - So we probably don't want to check the direct referral perm unless referral_mode == 'legacy'? Something like:
```ruby
 can_send = referral_mode == 'legacy' ? policy.can_send_out_direct_referral? : true
 access_denied! unless can_send
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: I agree it is getting large. I would say that this proposed new `directReferralForm` query falls into a bucket of root queries we have for looking up Form Definitions. I would be more inclined to somehow group these together based on their function (form lookup) as opposed to grouping this in with referral queries. Because they have some similar logic. Various root queries for fetching Form Definitions:
* `assessmentFormDefinition`- select form for assessment
* `recordFormDefinition` - select form for record
* `serviceFormDefinition` - select form for service
* `staticFormDefinition` - lookup form for static form role
* `formDefinition` - direct form lookup
* `directReferralForm` (new)- select form for direct referral

Not sure what that would look like, maybe something like below, we could also refactor the lookups since several of them have different uses (Eg lookup by id vs lookup by role..) there is certainly room for improvement.
```
formDefinitionLookup {
    assessmentFormByRole(role:)
    assessmentFormById(id:)
    recordForm(..)
    serviceForm(..)
    formById(..)

}
```

This does feel out of scope to decide on a new pattern we'd like to use, that might be a longer conversation. Maybe we can capture in another ticket?

At the very least, I would suggest this query to have suffix `_definition` to be consistent with the 5 existing form lookup queries 🙂

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: Yeah good call-out, this isn't consistent with other form lookups. In other places we do not treat the forms themselves as sensitive. For example if the user is looking up a form to use for recording/editing a Service, we don't check (or know) whether the user has permission to view or edit any services. Same for looking up the client form, assessment form, etc none of these check permissions. I think that's a good thing because we don't need to protect the form in that way.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -550,6 +555,39 @@ def ce_referral(id:)
...
+
+      unless workflow_template&.published? && workflow_template.template_type.to_s == 'ce_referral'
+        error_message = "Workflow template not found. #{base_error_message}"
+        raise HmisErrors::ApiError.new(error_message, display_message: user_facing_error_message)
+      end
+
+      initiation_node = workflow_template.graph.nodes.find(&:delegated_handoff)
+      unless initiation_node&.user_task? && initiation_node.form_definition.present?
+        error_message = "Workflow template #{workflow_template.id} does not have a UserTask node with a form definition that is marked 'delegated_handoff'. #{base_error_message}"
+        raise HmisErrors::ApiError.new(error_message, display_message: user_facing_error_message)
```
gigxz: @martha can you say more about making these user-facing error messages? Can these scenarios arise? The user is unable to do anything about it, so it seems to me it would be preferable to try to prevent this scenario (and raise the usual exception if it occurs) rather than allowing it with a user-facing error.

db/warehouse/migrate/20250716154410_add_delegated_handoff_to_wfd_node.rb
martha: Thanks @ttoomey I really like this suggestion and took a stab at it here: https://github.com/greenriver/hmis-warehouse/pull/5585/commits/5930e32841c729950e18ad4eef613dc6a917f9f7 I'm not sure it's airtight yet but wanted to push up my progress!

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: In other words, shouldn't the `unit_group.accepts_direct_ce_referrals?` check in the `unit_groups_for_project_ce_referral` pick list prevent unit groups with these issues from showing up?
I guess there is a race condition that someone could modify the Workflow Template in the moments between someone loading the picklist and loading the form.. but templates are changed very infrequently, I don't think it's necessarily worth adding these user messages if it's a small race condition like that, and the user can't do anything about it. (Aside from refresh, which is what the stock error message suggests).

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -188,10 +189,30 @@ def self.apply_filters(input)
     Hmis::Filter::ProjectFilter.new(input).filter_scope(self)
   end
 
+  # Legacy referrals
   def receives_referrals?
```
gigxz: this is only used in one place, perhaps rename to `receives_legacy_referrals?`?

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -141,6 +141,7 @@ class Hmis::Hud::Project < Hmis::Hud::Base
     )
   end
 
+  # Legacy referrals
   scope :receiving_referrals, -> do
```
gigxz: this is only used in a couple places, perhaps rename to `receiving_legacy_referrals` or `accepting_legacy_referrals`? (I see the existing code uses language "receive" but the new code uses "accept" to mean the same thing. I think "accept" is fine it would be good to be consistent going forward)

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: The same permission is used to govern the sending of legacy and "ce" direct referrals, so they would both be doing the same underlying check

drivers/hmis/app/graphql/types/forms/enums/pick_list_type.rb
```diff
@@ -16,6 +16,8 @@ class Forms::Enums::PickListType < Types::BaseEnum
     value 'RESIDENTIAL_PROJECTS', 'Residential Projects'
     value 'OPEN_PROJECTS', 'Open Projects that the user can see'
     value 'PROJECTS_RECEIVING_REFERRALS', 'Open Projects that can receive referrals'
+    value 'PROJECTS_ACCEPTING_CE_REFERRALS', 'Projects that are accepting CE referrals'
+    value 'UNIT_GROUPS_FOR_PROJECT_CE_REFERRAL', 'Unit groups for the given project that can accept CE referrals'
```
gigxz: Naming-wise, if these picklists are dependent specifically on the **direct** referral configuration, I think we should put "direct" in the name to prevent ambiguity

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Thanks! Yes good point I'll clean this up -- I had added them early on and then later fiddled further with the picklist so that the situation shouldn't arise

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Thanks! I created https://github.com/open-path/Green-River/issues/7994 to follow up

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -188,10 +189,30 @@ def self.apply_filters(input)
...
     Hmis::Form::Instance.active.published.with_role(:REFERRAL).any? { |instance| instance.project_match(self) }
   end
 
+  def accepts_direct_ce_referrals?
+    config = Hmis::ProjectCeConfig.detect_best_config_for_project(self)
+
+    return false unless config.present?
+    return false unless config.accepts_direct_referrals?
+
+    true
```
gigxz: This could be simplified a bit
```suggestion
    config = Hmis::ProjectCeConfig.detect_best_config_for_project(self)
    config&.accepts_direct_referrals?
```

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -188,10 +189,30 @@ def self.apply_filters(input)
...
+  end
+
+  def accepts_direct_ce_referrals_from?(source_project)
+    return false unless accepts_direct_ce_referrals?
+
+    config = Hmis::ProjectCeConfig.detect_best_config_for_project(self)
+    # If the config specifies a list of projects that it accepts referrals from, check that this project is in that list.
+    return false if config.accepts_direct_referrals_from.present? && config.accepts_direct_referrals_from.exclude?(source_project.id)
+
+    true
```
gigxz: another nit on readability, checking for project **inclusion** in the list (as opposed to not-exclusion) is a little easier to understand (for me!). We could also optimize by not invoking `detect_best_config_for_project` twice, especially since this is called for all projects when generating the pick list, which could get slow.
Suggestion:

```ruby
    config = Hmis::ProjectCeConfig.detect_best_config_for_project(self)
    return false unless config&.accepts_direct_referrals?
    return true unless config.accepts_direct_referrals_from.present? # no projects specified, so accept from all

    config.accepts_direct_referrals_from.include?(source_project.id)
```

drivers/hmis/app/models/hmis/hud/project.rb
martha: Hm, do you have a preference between the two? Now that I'm thinking about it, I slightly prefer `receiving` over `accepting` because it's more neutral regarding the separate concept of "referral acceptance," later on down the road.

Meaning, these scopes `receiving_referrals` and flags `receives_referrals?` aren't saying anything about whether a potential direct referral will get accepted, only that the project can receive it and acceptance may or may not happen later.

db/warehouse/migrate/20250716154410_add_delegated_handoff_to_wfd_node.rb
ttoomey: Thanks! I guess a question is if it works for config management. Imagining that an admin would have to assign the unit group to a workflow template and choose the correct form entry point.

Just to pull on this a little more - we need this reference because we can't just rely on the steps sequence in the workflow and swimlane config to determine the appropriate step to show? Happy to huddle on this

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: thank you!

db/warehouse/migrate/20250716154410_add_delegated_handoff_to_wfd_node.rb
martha: Thanks for thinking it through. I'm open to the idea that we could just rely on the workflow config. I think we didn't start out with that implementation for two reasons,
1. The workflow could have a lot of steps that all become active as soon as the workflow starts. In that case, we wouldn't know which one to ask the sending project to fill out. Maybe what you're getting at is that we could require/validate that there is only one (or, only one in a specific swimlane?)? But I think we do want that flexibility to have multiple starting steps -- at least in waitlist (non direct) referrals.
2. Whether or not a workflow has a handoff step is one useful way of flagging it as "receiving direct referrals.", see my question [here](https://github.com/greenriver/hmis-warehouse/pull/5585#discussion_r2216847745). If we don't have handoff steps explicitly flagged, we may need to add an explicit flag elsewhere, like on the workflow template?

Hm, so maybe the alternative implementation you're pointing towards would be something like,
- ~add a flag on the _workflow template_ indicating whether this template can be used for direct referral.~ no, workflow templates stay generic (I _think_ we do want a unit group to be able to accept either direct or waitlist referrals, so this probably shouldn't be a mutually exclusive flag)
- if so: validate that it has only one starting step, and use that as the direct referral entrypoint

Does that sound preferable?
- pros: it would probably result in clearer code and config
- cons: it's a bit less flexible in terms of workflow definition, since we could no longer define multiple entrypoints on a workflow template if that template is ever used for direct referrals. But I'm not sure that's a real use case.

martha: Huddle discussion notes:
- we're proposing to add a new assoc on `unit_group`, `direct_referral_workflow_template_identifier`
- this would be alongside the existing `workflow_template_identifier`
- that allows flexibility for a UnitGroup to receive either direct or waitlist referrals, but to have different workflows for each, which we think makes sense

martha: I made a little progress towards the approach we had outlined on the call. Pretty quickly though, I got stumped because of the following: At referral creation time, we aren't getting the workflow template from the unit group, but rather from the opportunity. This is saved on the opportunity at opportunity creation time. I'm assuming we don't want/need the added complexity of having separate `direct_referral_workflow_template_identifier` in addition to existing `workflow_template_identifier` on _both_ the unit_group _and_ opportunity models. (At least, the idea of needing this complexity definitely made me hesitate on the approach we had discussed.)

I drafted instead (http://github.com/greenriver/hmis-warehouse/pull/5585/commits/93c237cf573b3892067449896503f9272f6de79f) an approach that just removes altogether the concept of the "direct handoff" or "task for sending project" flag. It simply expects that in the case of a direct referral, the sending project should complete the first user task, and in the workflow template there should be exactly one.

Other previous approaches we could return to:
- `delegated_handoff` flag on wfd node https://github.com/greenriver/hmis-warehouse/pull/5585/commits/dcd9f8d7b4b64ac89a471aa7a1d8a883ec5d4a1f
- `direct_referral_entrypoint` relation on `unit_group` https://github.com/greenriver/hmis-warehouse/commit/5930e32841c729950e18ad4eef613dc6a917f9f7

I'd like to flag I'm running out of time on this ticket; we had estimated about 13h, and I've spent about 15h so far.

@gigxz  curious for your directional input here; happy to pair and go over the options in more detail if the thread is a bit hard to follow (sorry!)

gigxz: @martha thanks for outlining that. I suggest moving forward with the approach you just pushed (use the first UserTask if present; raise with a descriptive message if not)

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
gigxz: I think this is fine!

drivers/hmis/app/models/hmis/hud/project.rb
gigxz: good point, looks like you made that change 👍 thanks

db/warehouse/migrate/20250716154410_add_delegated_handoff_to_wfd_node.rb
ttoomey: Thanks Martha! Sorry this took more time. 

Thinking through the design a little more:
Perhaps the assignment of a workflow to an opportunity could be deferred until referral creation time. When creating a referral, the mutation would need to get the appropriate workflow from opportunity.unit.unit_group and then persist that on the opportunity

This would support opportunities that are available for either CE or direct referrals using different workflows. Also could support other referral types in the future.  

Essentially:
* unit_group stores the appropriate workflows used when creating each type of referral
* opportunity stores the workflow appropriate to the referral type

martha: Thanks! I'd like to propose splitting that out into a future ticket - I created https://github.com/open-path/Green-River/issues/8001.

martha:
If no objection, I plan to merge this (and its corresponding frontend pr) into their respective base branches w/out squashing. Then squash merge that feature branch into probably release-174 once work for #7963 is also reviewed and merged.

PR #5586, opened 2025-07-18
Protected client search on cohort clients page

eanders:
## _Merging this PR_

## Description
* Adopts the client search logic from the main search page for the add clients page of cohorts

## Type of change
* New feature


app/controllers/cohorts/clients/search_queries_controller.rb
eanders, line 1: @ttoomey do you have any thoughts about this being its own file or just a method on cohorts/clients_controller?  I'm tempted to move it since I need to bring in all the authorization bits, but you'd built it as a separate controller for client search, so maybe there was a reason?

ttoomey: Do you mean if it should it be its own controller or should we add another action to the existing controller?

Generally I prefer to use single-purpose restful controllers when there's a resource pattern (as I think this is). I think it's easier to reason about and it's more in-keeping with rails conventions. But it does mean that shared behaviors have to be repeated of DRYed up to a concern. 

It's not a hard rule or anything; there's always room for compromise. But I think the goal is to discourage "bag-of-things" controllers which can be hard to maintain.

ttoomey:
I pushed some specs up for this, hope that's okay. fcd7999293150f01ce4ad161af9dbcdcc9c3385e

ttoomey (approved):
Looks great, thanks!

PR #5587, opened 2025-07-18
Protected search for patient pages

eanders:
## _Merging this PR_

## Description
* Adds search protection to patient search pages

## Type of change
* New feature


app/controllers/health/my_patients_controller.rb
eanders, line 1: @ttoomey this is a version that doesn't use a new controller.  I guess I like this better, at least in this situation as we'd need to adjust both the current controllers with a new method or two _and_ have a new controller for each search.  This adds one additional, relatively small method to the existing controllers.

If you think this is too much of a grab bag, I could be convinced to follow the other pattern.

eanders:
To keep the complexity down on the healthcare side, I'm leaving off the extra controllers for now.

ttoomey (approved):
Looks good, thanks

PR #5589, opened 2025-07-22
Add query for project outgoing direct referrals

martha:
## _Merging this PR_

## Description
Summary of changes:
- Schema changes:
  - Add `outgoingDirectCeReferrals` as a new field to the `Project` schema type 
  - Update `CeReferral` graphql schema model with the same `summary_field` pattern as `Enrollment`.
    - Make formerly non-nullable fields nullable, if they are not summary fields
    - Add `canViewReferralDetails` to the access object
- Policy changes 
  - Update our graphql `BaseField` with a new field-level authorization pattern. It can now accept a `policy_name` and `policy_action` as args to determine whether a field is authorized.
  - Add new `can_view_summary?` method to the `CeReferralPolicy`

https://github.com/greenriver/hmis-frontend/pull/1214

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -10,43 +10,67 @@ module Types
...
+    def self.authorized?(object, ctx)
+      user = ctx[:current_user]
+      policy = user.policy_for(object, policy_type: :ce_referral)
+      super && (policy.can_view? || policy.can_view_summary?)
+    end
+
+    # Summary fields that anyone who can view the summary of this referral has access to
+    summary_field :id, ID, null: false
+    summary_field :status, HmisSchema::Enums::CeReferralStatus, null: false
+    summary_field :custom_status, HmisSchema::CeCustomReferralStatus, null: true
```
martha, line 43: Is `custom_status` a summary field? Or an internal detail to the receiving project that the sending project doesn't need to know?

drivers/hmis/app/models/hmis/auth_policies/context_loaders/hmis_project_access_group_loader.rb
```diff
@@ -34,7 +34,12 @@ def preload(project_ids)
           # Filter out deleted access groups. ProjectAccessGroupMember can't do this due to database boundaries
           active_access_group_ids.intersection(clean_values).to_a
         end
+
       @cache.merge!(results)
+
+      # For projects that don't have any access groups, add `nil` to the cache, so that we don't check the db again
```
ttoomey, line 40: nice!

drivers/hmis/app/graphql/types/base_field.rb
```diff
@@ -10,8 +10,15 @@ module Types
   class BaseField < GraphQL::Schema::Field
     argument_class Types::BaseArgument
 
-    def initialize(*args, default_value: nil, permissions: nil, **kwargs, &block)
```
ttoomey, line 13: Could we use a lambda to provide more flexibility around authorization and simplify logic? Maybe `authorize_with.call(obj, user)`

 Also maybe take the opportunity to add some documentation


```diff
--- a/drivers/hmis/app/graphql/types/base_field.rb
+++ b/drivers/hmis/app/graphql/types/base_field.rb
@@ -7,18 +7,22 @@
 ###
 
 module Types
+  # Custom base field class that provides common functionality for all fields in the schema.
+  # This includes:
+  # - A consistent authorization layer (`permissions` or `authorize_with`)
+  # - Automatic pagination for fields that return a paginated type
+  # - Support for default values
   class BaseField < GraphQL::Schema::Field
     argument_class Types::BaseArgument
 
-    def initialize(*args, default_value: nil, permissions: nil, policy_name: nil, policy_action: nil, **kwargs, &block)
-      @permissions = Array.wrap(permissions)
-      @policy_name = policy_name
-      @policy_action = policy_action&.to_sym
+    # @param default_value [Object] A default value for the field if it resolves to nil.
+    # @param permissions [Symbol, Array<Symbol>] (Deprecated) The required perm
ission(s) to view the field. Mutually exclusive with `authorize_with`.
+    # @param authorize_with [Proc] A lambda for custom authorization logic. It receives `user` and `object` and should return a boolean. Mutually exclusive with `permissions`.
+    def initialize(*args, default_value: nil, permissions: nil, authorize_with: nil, **kwargs, &block)
+      raise ArgumentError "don't use permissions and authorize_with" if permissions && authorize_with
 
-      # Validate that policy parameters are used together
-      if (@policy_name.present? && @policy_action.blank?) || (@policy_name.blank? && @policy_action.present?) # rubocop:disable Style/IfUnlessModifier
-        raise ArgumentError, 'policy_name and policy_action must be provided to
gether'
-      end
+      @permissions = Array.wrap(permissions)
+      @authorize_with = authorize_with
 
       after_paginate = kwargs.delete(:after_paginate)
       super(*args, **kwargs, &block)
@@ -39,22 +43,19 @@ module Types
     # Field-level authorization
     # https://graphql-ruby.org/authorization/authorization.html#field-authorization
     def authorized?(object, args, ctx)
-      base_authorized = super(object, args, ctx)
+      return false unless super(object, args, ctx)
+
       if @permissions.any?
         # if `permissions:` was given, then require the current user to have the specified permissions on the object
-        base_authorized && @permissions.all? do |perm|
+        @permissions.all? do |perm|
           GraphqlPermissionChecker.current_permission_for_context?(ctx, permission: perm, entity: object)
         end
-      elsif @policy_name
+      elsif @authorize_with
         # if `policy_name:` and `policy_action:` were given, require the user t
o be able to perform that action on that policy
         current_user = ctx[:current_user]
-        policy = current_user.policy_for(object, policy_type: @policy_name)
-
-        raise ArgumentError, "Policy #{policy.class.name} does not respond to #{@policy_action}" unless @policy_action && policy.respond_to?(@policy_action)
-
-        base_authorized && policy.send(@policy_action)
+        @authorize_with.call(ctx[:current_user], object)
       else
-        base_authorized
+        true
       end
     end
 
```

martha: Thanks, yes I like that better!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -10,43 +10,70 @@ module Types
...
+    # and adds a `summary_field` method, to distinguish between fields that can be resolved
+    # by those who can_view? vs. just can_view_summary?
+    # If user lacks sufficient access, the field will be resolved as null.
+    def self.field(name, type = nil, **kwargs)
+      # See Types::BaseField `authorized?` function.
+      can_view_lambda = lambda do |current_user, object|
+        current_user.policy_for(object, policy_type: :ce_referral).can_view?
+      end
+
+      super(name, type, authorize_with: can_view_lambda, **kwargs)
```
ttoomey: Seems like this should conditionally set authorize_with?
```suggestion
      # See Types::BaseField `authorized?` function.
      unless kwargs.key?(:authorize_with)
        kwargs[:authorize_with] = lambda do |current_user, object|
          current_user.policy_for(object, policy_type: :ce_referral).can_view?
        end
      end
        
      super(name, type, **kwargs)
```

ttoomey (approved):
Looks good! There might be a bug in `field()` that should get corrected, see note. Thanks!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: definitely a summary field!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -10,43 +10,70 @@ module Types
...
+
+    access_field authorize_with: nil do
+      field :can_view_referral_details, Boolean, null: false
+      field :can_view_target_project, Boolean, null: false
+    end
 
+    # Detailed fields that only those with full view access should see. Must be nullable
+    field :opportunity, HmisSchema::CeOpportunity, null: true
+    field :steps, [HmisSchema::CeReferralStep], null: true
+    field :client_name, String, null: true, description: 'The name of the referred client. Always available to those who can view the referral, even without full client record access.'
```
gigxz: This introduces a scenario where clientName is nil, but _client_ is present. That's the first scenario below:
```graphql
# if user has access to client, but not full access to referral
client { name } // resolves
clientName // nil

# if user has full access to client and referral
client { name } // resolves
clientName // resolves

# if user has full access to referral, but not client
client { name } // nil
clientName // resolves
```
That feels pretty confusing for the API consumer, and also puts the burden on the frontend to "know" to resolve the name using the `client` object when viewing in an "direct outgoing referral" context, vs using the `clientName` in all other contexts. That might get brittle as this changes over time.

I'd suggest sticking to the logic of  `clientName` resolves as much as possible, so the frontend can continue to use that for name display. That would require making client_name a summary field, and updating the logic so it does check for viewable_by if the user only has "summary" access to the referral. what do you think of that approach?

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -10,43 +10,70 @@ module Types
...
+
+    # Summary fields that anyone who can view the summary of this referral has access to
+    summary_field :id, ID, null: false
+    summary_field :status, HmisSchema::Enums::CeReferralStatus, null: false
+    summary_field :custom_status, HmisSchema::CeCustomReferralStatus, null: true
+    summary_field :client_id, ID, null: false
+    # Special case: Client is a "summary field" because it doesn't require referral visibility, but resolving it does require permission to view the client record.
+    summary_field :client, Types::HmisSchema::Client, null: true, description: 'The full client record, if the user has permission to view it.'
+    summary_field :created_at, GraphQL::Types::ISO8601DateTime, null: false
+    summary_field :source_enrollment, Types::HmisSchema::CeReferralSourceEnrollment, null: true, description: 'Limited details about the source enrollment. Available even without full access to the source record.'
```
gigxz: This makes me a little nervous. "CeReferralSourceEnrollment" was introduced in order to resolve "expanded" details about the source enrollment on a Referral. Now, we're proposing to also resolving it when the user doesn't even have full access to the Referral itself? We should be extra careful to understand the use of these "expanded" types since they bypass the usual permission check. I worry that down the line, we might add additional fields to "CeReferralSourceEnrollment" thinking that it has the original use-case, not realizing its additionally resolved for summary-level referrals. 

Looking at your frontend changes, it seems like this field is included as a summary_field just so that we can link to the Enrollment dashboard, _not_ to expose any expanded data about the source enrollment.

Am I understanding that correctly?

If so, it feels safer to just add those needed fields as summary_fields, and remove the CeReferralSourceEnrollment from the summary_field set. For example, I think the below would be enough to construct the link (and to determine whether we can successfully link)
```graphql
ceReferral {
     clientId # summary field
     sourceEnrollmentId # summary field
     access {
         canViewReferralDetails # determine whether we can link to referral
         canViewTargetProject # determine whether we can link to target project
         canViewSourceEnrollmentDetails # determine whether can link to the source enrollment
     }
}
```
The data source check isn't needed, because "can view enrollment details" would be false if it was from another data source.

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -298,6 +299,13 @@ def ce_referrals(**args) # Don't resolve in batch
       resolve_ce_referrals(object.ce_referrals, **args)
     end
 
+    def outgoing_direct_ce_referrals(**args)
```
gigxz, line 302: I think the default sort for these should probably be sorted from the date when they were sent (recent=>oldest). Right now these are sorted by status which doesn't feel right for outgoing referrals, since its more of a log of what has been sent.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -10,43 +10,70 @@ module Types
...
-    field :swimlanes, [HmisSchema::CeReferralSwimlane], null: false
+    field :swimlanes, [HmisSchema::CeReferralSwimlane], null: true
     field :workflow_template_name, String, null: true
-    field :audit_events, HmisSchema::CeReferralAuditEvent.page_type, null: false
-    field :notes, HmisSchema::CeReferralNote.page_type, null: false
-
+    field :audit_events, HmisSchema::CeReferralAuditEvent.page_type, null: true
+    field :notes, HmisSchema::CeReferralNote.page_type, null: true
     # generically resolve current values for any fields referenced by Match Rule expressions
     field :current_match_values, [HmisSchema::CeMatchValue], null: true, description: 'Eligibility-related field values. May expose data beyond normal permissions.', method: :resolve_match_rule_fields
```
gigxz, line 81: I think we need to add referral origin and the ability to filter on that? Sorry if that is captured by a different PR or issue I may have missed it. I'm thinking of the CE Admin page in particular, which should support filtering out direct-sent referrals.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Makes sense! I think this mirrors a similar unsurfaced bug in the Enrollment schema object? https://github.com/greenriver/hmis-warehouse/blob/db323483a2d0dfe0d439a0e8e82e75138d40ad0e/drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb#L61 (Unsurfaced because we don't currently override the `permissions` kwarg on any fields in that file.) But I think I would prefer to leave that unchanged for now to avoid qa scope creep.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: That seems like a good approach, thanks for the feedback! https://github.com/greenriver/hmis-warehouse/pull/5589/commits/f1abcdea554622066cf5bf1d9b306f3fb51c9635

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Thanks, yes it's just for the link, so I agree with you. https://github.com/greenriver/hmis-warehouse/pull/5589/commits/cb920d293e5c57148dac144bbef9721af1312af1

> The data source check isn't needed, because "can view enrollment details" would be false if it was from another data source.

I agree with this too. The `isCurrentDataSource` field is currently unnecessary, but we are resolving and using it in one place at the moment, I think almost as a reminder to our future selves about the future state where we may have source enrollments from different data sources. Let me know what you think about that -- should we remove it entirely?
https://github.com/greenriver/hmis-frontend/blob/release-174/src/modules/ce/components/unit/SourceEnrollmentCard.tsx#L37-L42

drivers/hmis/app/graphql/types/hmis_schema/project.rb
martha: Sounds good https://github.com/greenriver/hmis-warehouse/pull/5589/commits/9d8846a35fbd3e7576d0e68cc861f254e9a8d52b. I haven't gone all the way down the road of enabling custom sorting on the table from the frontend, but, I guess, started down that road in the style of HasEnrollments.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Sounds good sorry for missing it! https://github.com/greenriver/hmis-warehouse/pull/5589/commits/90ed3dca49b2ebfb9c95336234ffbf6b40eb6721

martha:
I think I've addressed all the feedback, so I will merge this into the feature branch and confirm tests pass. I'll do another pass at the QA list in the morning just because there have been a lot of moving parts with this one and I want to confirm some functionality. Then I hope to merge it into release-174.

PR #5592, opened 2025-07-23
7970 warehouse auth policy fix

ttoomey:
## _Merging this PR_

## Description
This pull request addresses a potential authorization issue where soft-deleted Collection or AccessGroup records could still grant permissions to users. 

The solution is to fetch the IDs of all active Collection and AccessGroup records from the application database at the beginning of the authorization check.

Note that the issue would only impact permissions for ACL users in this edge case. The legacy context already obeys soft-deletion via `user.access_groups` but an additional filter was added as a hedge against future use of access group ids.

Tests have been added for legacy and ACL contexts, as well as additional tests on the db view

## Type of change
Bug fix


dtgreiner (approved):

eanders (approved):

ttoomey:
will merge to 174

PR #5597, opened 2025-07-24
Prevent assigning to unit with an active Opportunity

martha:
## _Merging this PR_

## Description

This PR prevents direct assignment to units that have open opportunities 

~In this PR, I'm suggesting the following fix to the bug described in the ticket:~
- ~If a unit has an open opportunity, it can still be directly assigned (from an enrollment form, a bulk service action, etc.), but doing so removes the opportunity, marking it unavailable for referrals.~
- ~If a unit has a locked opportunity (referral in progress), it should not be available for occupancy assignment.~

~I figured this would be less frustrating to users, compared to the simpler alternative of preventing users from assigning to a unit if it has an opportunity at all.~

## Type of change
Bug fix


gigxz:
@martha have you taken a look at those system test failures? they look different from the known flaky issue we've been seeing

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -476,6 +476,14 @@ def assign_unit(unit:, start_date:, user:)
     occupants = unit.occupants_on(start_date)
     raise 'Unit is already assigned to a different household' if occupants.where.not(household_id: household_id).present?
 
+    opportunity = unit.latest_opportunity
+
+    if opportunity&.locked?
+      raise 'Cannot assign to a unit with an active referral' unless opportunity.active_referral&.client == client
+    end
+
+    opportunity.destroy! if opportunity&.open?
```
gigxz: Let's discuss if we really want this, we don't have "Acts as paranoid" on opportunity currently so it would fully remove it. Right now we are keeping them around with "closed" status, is that right?

We could probably remove this line for now if it doesn't affect the underlying behavior of this ticket (preventing enrolling into units that have in progress referrals).

gigxz: The Referral Message Handler only calls opportunity.close!, not opportunity.destroy!. I guess assign_unit is called in a wider context, though. I can look at this more closely

martha: Yes, let's discuss more! My thinking was to mirror the fact that we prevent you from marking a unit available if it has occupants. So, if the principle is that in general a unit with occupants shouldn't have an open opportunity, then conversely when we assign a unit we should remove any opportunity that it has.

However... now that I'm thinking about it more, it does seem preferable to close the opportunity, instead of destroying it. I used destroy! to roughly mirror what the mark-unavailable mutation does, but (from our bug bash spreadsheet),
> A unit that’s currently accepting referrals AND has a past rejected referral raises a error when you try to “stop accepting referrals” on that unit.

martha:
I think the weird system test issue was transient - I also noticed it was different than usual, but I also observed it on the base branch PR at one point, in both cases it seems to have resolved to the usual/familiar failure.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -476,6 +476,14 @@ def assign_unit(unit:, start_date:, user:)
     occupants = unit.occupants_on(start_date)
     raise 'Unit is already assigned to a different household' if occupants.where.not(household_id: household_id).present?
 
+    opportunity = unit.latest_opportunity
+
+    if opportunity&.locked?
+      raise 'Cannot assign to a unit with an active referral' unless opportunity.active_referral&.client == client
+    end
+
+    opportunity.close! if opportunity&.open?
```
gigxz: I missed this initially, I'm not sure if `assign_unit` is the right place to close the opportunity. `assign_unit` builds the occupancy attributes that will be saved when the enrollment is saved, but it doesn't actually make any changes. Adding `opportunity.close!` makes it so this function now writes to the database. That might break assumptions for callers, and lead to inconsistencies (we might not always call this method in a transaction, for example). It also just seems weird for this function to update the opportunity record but nothing else.

A few options to consider
- see if there is a clean way to make it so the related opportunity is only closed when the underlying enrollment is saved (similar to the existing way this method sets attributes on related records)
- see whether it makes sense to update this function (assign_unit) to persist all the changes it makes? The only place I'm seeing where we don't immediately save after calling this is EnrollmentProcessor. (From a quick glance, but this would need a more thorough audit of assign_unit usage).

gigxz: I'm also wondering how this interacts with the call to `enrollment.assign_unit` that is currently made from the ReferralEnroller when an opportunity is created.

martha: Yeah that seems totally sensible! Here's my attempt at an audit of places where `assign_unit` is currently called...
- `update_referral_posting.rb:72`. 
  - Theoretically this shouldn't need to worry about opportunities, since it's the legacy referral system. Legacy and ce referrals should be mutually exclusive on a project.
  - Proposed solution: just to be safe, we could filter out units with opportunities on line 62?
- `bulk_assign_service.rb:78`
  - Should close an open opportunity if one exists
  - Proposed solution: move the `opportunity.close!` call out here into the bulk service mutation
- `join_household.rb:55`
  - The receiving unit already has occupants (the receiving household), so it probably doesn't have an open opportunity.
- `referral_enroller.rb:48`
  - You are right, this is a situation where we don't want to close the open opportunity on which the referral is happening
- `enrollment_processor.rb:87`
  - Like `bulk_assignment`, I think in this case we would want to close any open opportunity on the unit.
  - Proposed solution: following the same pattern as the `unit_occupancy_changes` enrollment after-save hook, close the open opportunity on save.

How does that sound? Commit that implements proposed solutions: https://github.com/greenriver/hmis-warehouse/pull/5597/commits/dc2e05d25b5238aeb08800965f6984aefd1f074e

If you don't agree with these changes, we can always revert to,
> the simpler alternative of preventing users from assigning to a unit if it has an opportunity at all.

- Pros: simpler; prevents buggy situations where an occupied unit has an opportunity
- Cons: more frustrating for user, if they are trying to direct enroll into a project that has units with opportunities

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -92,6 +92,12 @@ class Hmis::Unit < Hmis::HmisBase
     active.unoccupied_on.joins(:opportunities).merge(Hmis::Ce::Opportunity.receiving_referrals)
   end
 
+  scope :not_receiving_referrals, -> do
```
martha, line 95: I went back and forth about the name of this scope. Maybe `without_open_opportunity` would be clearer.

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: Papertrail - we are taking the simpler alternative

martha:
@gigxz, thanks for your patience, I think this is ready for re-review.

gigxz (approved):
thanks!

PR #5598, opened 2025-07-27
Add permission for printing case notes

gigxz:
## _Merging this PR_

## Description
https://github.com/open-path/Green-River/issues/7866
https://github.com/greenriver/hmis-frontend/pull/1212

Adds permission for printing case notes. This permission currently doesn't actually gate anything on the backend, it just determines whether or not the print button shows up in the frontend. If/when we move to PDF-generation, this permission would be evaluated on the backend as well.

## Type of change
new feature. testing instructions for feature on frontend pr.

### DIFF
--- a/db/structure.sql
+++ b/db/structure.sql
@@ -372,6 +372,7 @@ CREATE TABLE public.activity_logs (
 --
 
 CREATE SEQUENCE public.activity_logs_id_seq
+    AS integer
     START WITH 1
     INCREMENT BY 1
     NO MINVALUE
@@ -414,6 +415,7 @@ CREATE TABLE public.agencies_consent_limits (
 --
 
 CREATE SEQUENCE public.agencies_id_seq
+    AS integer
     START WITH 1
     INCREMENT BY 1
     NO MINVALUE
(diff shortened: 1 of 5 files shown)


db/structure.sql
```diff
@@ -1235,10 +938,10 @@ CREATE TABLE public.hmis_roles (
     can_manage_own_client_files boolean DEFAULT false NOT NULL,
     can_view_any_nonconfidential_client_files boolean DEFAULT false NOT NULL,
     can_view_any_confidential_client_files boolean DEFAULT false NOT NULL,
-    can_manage_client_files boolean DEFAULT false NOT NULL,
```
gigxz, line 1238: I recently updated my db files to match what was on release, but now it looks like its changed again. I think I stand by whats in my db, because `can_manage_client_files` is not a real permission. this is likely from another devs environment

db/structure.sql
```diff
@@ -472,90 +474,6 @@ CREATE TABLE public.ar_internal_metadata (
 );
 
 
---
--- Name: building_contacts_id_seq; Type: SEQUENCE; Schema: public; Owner: -
```
gigxz, line 476: should this be present? review..

db/structure.sql
martha: Makes sense! Looks like `can_manage_client_files` is still present in this file on line 1819, not sure what is up with that.

db/structure.sql
martha: for what it's worth these are also not present in my local db (and get removed from the structure file by `rails db:migrate` when I run that locally)

martha (approved):
Looks good, thank you!
> This permission currently doesn't actually gate anything on the backend, it just determines whether or not the print button shows up in the frontend. If/when we move to PDF-generation, this permission would be evaluated on the backend as well.

I think it could be reasonable to add this comment somewhere in the code, maybe on the role itself or on the client schema object? But could also be persuaded against that if you think it just adds cruft that we will have to remove.

PR #5599, opened 2025-07-29
Custom imports refactor

eanders:
## _Merging this PR_

## Description
* Follow-up on https://github.com/greenriver/hmis-warehouse/pull/5559 to use service-objects instead of stateless class methods

## Type of change
* Code clean-up

### DIFF
--- a/drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/column_mapper.rb
+++ b/drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/column_mapper.rb
@@ -7,142 +7,113 @@
 # frozen_string_literal: true
 
 module HmisCsvTwentyTwentySix::Importer::Custom
-  # Handles column mapping from custom CSV files to warehouse tables
+  # A service object that handles column mapping from custom CSV files to warehouse tables.
   #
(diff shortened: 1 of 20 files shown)


drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/importer/custom/column_mapper.rb
```diff
@@ -301,238 +247,77 @@ def self.apply_mappings(source_record, mapped_attributes, column_configs, lookup
...
+      # Preserve boolean false, since `false.blank?` is true.
+      return value if value == false
+      return nil if value.blank?
+
+      case type
+      when 'integer'
+        value.to_i
+      else # Default to string or no-op
+        value
       end
```
ttoomey: maybe worth being a little more defensive?

```suggestion
      # Default to 'string' if type is not specified in the YAML config.
      type_to_cast = type || 'string'

      case type_to_cast
      when 'integer'
        value.to_i
      when 'date'
        value.to_date
      when 'datetime'
        Time.zone.parse(value)
      when 'string'
        value
      else
        raise ArgumentError, "Unknown column type '#{type}' specified in custom file configuration."
      end
```

ttoomey (approved):
I think this is a really nice improvement, thank you!!

PR #5605, opened 2025-07-30
7937 optimize ce match engine

ttoomey:
## _Merging this PR_

## Description
Performance improvement to the CE matching engine by refactoring the data-loading strategy from individual record lookups to batch queries. Replace N+1 instance-level evaluations with bulk data fetching, pre-loading all required fields for a batch of clients at once. Performance Impact: ~100x improvement (5 → 500 clients/sec, depending on the expression)

Also:
* Implements a new `days_since_last_exit` match field for eligibility, allowing for the sql pre filter to exclude clients who have not been enrolled in the last X days. This should let us add rules that greatly reduce the number of clients we need to evaluate
* Adds new SQL view (hmis_destination_client_latest_assessments) to query the latest assessments by destination client, enabling batch evaluation
* Adds logging to match engine and process changes job
* Fixes bug where dirty pools were not marked clean if they were no-longer associated with an opportunity
* removes `days_homeless` field as it's complex to implement correctly and is not needed for mvp

## Type of change
Performance enhancement


drivers/hmis/app/models/hmis/ce/match/expression/cde_field_map.rb
```diff
@@ -32,44 +62,36 @@ def label_for(field)
...
+      return _format_for_display(field, value) unless cded.repeats?
+
+      Array.wrap(value).map { |v| _format_for_display(field, v) }
+    end
+
+    private
+
+    def _format_for_display(field, value)
+      cded = parse_entity_type(field)
+      case cded.field_type.to_sym
```
ttoomey, line 79: seemed better to use the cded type rather than trying guess from value's type

drivers/hmis/app/models/hmis/ce/match/expression/homeless_days_calculator.rb
```diff
@@ -0,0 +1,59 @@
...
+      calculate_homeless_days(client_ids, housed_dates_by_client, homeless_dates_by_client)
+    end
+
+    private
+
+    def fetch_housed_dates(client_ids)
+      # First identify any days the client was housed, as these will be excluded from the count.
+      # A client can have both homeless and housed service history records on the same day,
+      # with "housed" taking precedence.
+      housed_dates = GrdaWarehouse::ServiceHistoryService.non_homeless.
```
ttoomey: should we continue to rely on ServiceHistoryService to calculate days homeless?

drivers/hmis/lib/hmis_util/ce_builder.rb
```diff
@@ -30,7 +30,17 @@ def self.build_candidate_pools(clients: nil, opportunities: nil, progress: false
 
       # Process all dirty pools and clients using the production job
       # This populates the pools by calling the match engine with the same logic used in production
-      Hmis::Ce::ProcessChangesJob.new.perform(progress: progress) while Hmis::Ce::ChangeMarker.dirty.exists?
```
ttoomey, line 33: if pools are not all cleaned for some reason, this would never exit

db/views/hmis_destination_client_latest_assessments_v01.sql
```diff
@@ -0,0 +1,18 @@
...
+  AND c.id = wc.source_id
+INNER JOIN "CustomAssessments" ca ON ca."DateDeleted" IS NULL
+  AND ca."data_source_id" = c."data_source_id"
+  AND ca."PersonalID" = c."PersonalID"
+INNER JOIN "hmis_form_processors" fp ON fp.owner_type = 'Hmis::Hud::CustomAssessment'
+  AND fp.owner_id = ca.id
+INNER JOIN "hmis_form_definitions" def ON def."deleted_at" IS NULL
+  AND def.id = fp.definition_id
+WHERE wc.destination_id IS NOT NULL
+ORDER BY wc.destination_id, def.identifier, ca."DateUpdated" DESC, ca.id DESC;
```
eanders: Noting, this will be most-recently updated, not most recent by assessment date, which is what I would have expected.

drivers/hmis/app/models/hmis/ce/match/expression/age_calculator.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+module Hmis::Ce::Match::Expression
```
eanders, line 3: I wonder if we should think about this more globally, and relocate to the main app code.  In `app/models/concerns/arel_helper.rb` we have something very similar `def age_on_date(start_date)` that could probably be adjusted to use this internally so we don't have additional calculations of age floating around.

Additionally we have `app/models/cohort_columns/age.rb` which is also similar.

This expression doesn't appear to be specific to CE, and could be the start of some simple expressions we could begin using across the application.

That said, this is fine as is for now, but what do you think about a more generic set of expressions in the future?

drivers/hmis/app/models/hmis/ce/match/expression/cde_field_map.rb
```diff
@@ -3,22 +3,52 @@
...
+        pluck(
+          :destination_client_id,
+          cded.cde_arel_field,
+        )
+
+      result = if cded.repeats?
+        values.group_by(&:first).transform_values { |pairs| pairs.map(&:last) }
+      else
+        values.index_by(&:first).transform_values(&:last)
+      end
```
eanders: Would it make sense to move this logic into the `cded`?  Something like `result = cded.data_from_values(values)` (or as a class method.)

eanders: Also, I think:
```
values.index_by(&:first).transform_values(&:last)
```
is the same as 
```
values.to_h
```

eanders: Not sure it is better, but I think:
```
values.group_by(&:shift).transform_values(&:flatten)
```
would be the same as:
```
values.group_by(&:first).transform_values { |pairs| pairs.map(&:last) }
```

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
```diff
@@ -34,77 +46,83 @@ def arel
...
-      )
+    private
+
+    def days_since_last_exit_field
+      calculator = LastEnrolledDaysCalculator.new(@current_date)
+      {
+        query: ->(clients) { calculator.call(clients) },
+        joins: [{ hmis_source_clients: { enrollments: :exit } }],
+        arel_field: calculator.arel_expression,
+        format_for_display: ->(days) { days.nil? ? nil : "#{days} #{'day'.pluralize(days)}" },
```
eanders: do we want to add some number formatting so we don't get:
**3241 days** but **3,241 days**.

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
```diff
@@ -34,77 +46,83 @@ def arel
...
+        arel_field: calculator.arel_expression,
+      }
     end
 
-    def map_project_types(project_type_ids)
-      project_type_ids.uniq.map { |t| HudUtility2026.project_type(t) }
+    def days_homeless_field
+      {
+        query: ->(clients) { HomelessDaysCalculator.new(@current_date).call(clients) },
+        format_for_display: ->(days) { days.nil? ? nil : "#{days} #{'day'.pluralize(days)}" },
```
eanders: same here.
```
include ActionView::Helpers::NumberHelper
number_with_delimiter(days)
```

drivers/hmis/app/models/hmis/ce/match/expression/homeless_days_calculator.rb
eanders: If we can possibly avoid it, we should for 2 reasons:
1. It will always be a bit out of date
2. If you are looking by client over all time, you should use the materialized view, which will be even more out of date.  SHS is partitioned on year, SHS Materialized contains an index on client_id.

For housed dates:
1. get enrollments in PH where there is a move-in date (this is slightly complicated by needing to check the HoH for child clients.
2. Get the associated exit dates
3. Exclude all days between move-in and exit

For homeless days:
1. find entry and exit dates for all homeless entry/exit projects (project types [1, 2, 8]) all days between entry and exit (or current date if no exit) are homeless
2. find all services of type 200 (bed night) in ES NbN (project type 1)
3. find all CLS with a homeless CLS in Street Outreach (project type 4, CLS 100 .. 199) - there's some discussion we might want to have here if we count every day in the month or not for a CLS, but if you build it this way, we can expand if necessary.
4. Another discussion we should have: if we should count self-report time before entry.  Potentially we want to include time between DateToStreetESSH and EntryDate.  

I'd be happy to bat around some ideas of how we might be able to make this into a view, or at least wrap up the logic in a single purpose built tool.

eanders: I wonder if #4 should be included in an alternate field.  I'm unclear how much people expect that to be included or not in a days homeless calculation. Up until now, we have generally not included it, but that has felt somewhat incorrect as we've learned more.

db/views/hmis_destination_client_latest_assessments_v01.sql
ttoomey: good point, we probably should sort by `AssessmentDate` instead

drivers/hmis/app/models/hmis/ce/match/expression/age_calculator.rb
ttoomey: > That said, this is fine as is for now, but what do you think about a more generic set of expressions in the future?

For sure. Perhaps developing a set of [common query objects](https://thoughtbot.com/blog/a-case-for-query-objects-in-rails)

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
```diff
@@ -44,6 +44,30 @@ class Hmis::Hud::CustomDataElementDefinition < Hmis::Hud::Base
...
+      cde_t[:value_string]
+    when :text
+      cde_t[:value_text]
+    when :date
+      cde_t[:value_date]
+    when :json
+      cde_t[:value_json]
+    when :file
+      cde_t[:value_file]
+    else
```
eanders: Maybe we want this like this, but we could simplify a bit:

```suggestion
    case field_type.to_sym
    when :float, :integer, :boolean, :string, :text, :date, :json, :file
      cde_t["value_#{field_type}".to_sym]
    else
```

drivers/hmis/app/models/hmis/destination_client_latest_assessment.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# View that provides latest custom assessments for destination clients
+# This enables easy joining to get CDE values from the most recent assessments per form type
+class Hmis::DestinationClientLatestAssessment < GrdaWarehouseBase
```
eanders, line 5: Should we indicate that this class is read only?

```
def readonly?
  true
end
```

drivers/hmis/spec/models/hmis/ce/match/expression/cde_field_map_spec.rb
```diff
@@ -0,0 +1,132 @@
...
+             data_element_definition: repeating_cded,
+             value_string: 'Dust',
+             data_source: client1.data_source)
+
+      # Setup assessment for client2
+      assessment_c2 = create(:hmis_custom_assessment,
+                             client: client2,
+                             data_source: client2.data_source,
+                             assessment_date: current_date - 1.week,
+                             date_updated: current_date - 1.week,
```
eanders: Might be worth adding a data set where the date_updated is new but the assessment_date is old, to confirm it picks based on assessment date.

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
ttoomey: yes, we should take the opportunity to cleanup the display helpers. Although I'd like to avoid mixins,  probably just call `ApplicationController.helpers.pluralize(days, 'day')`

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
ttoomey: could definitely be cleaned up. Probably a lookup hash is best, I'll add that

drivers/hmis/app/models/hmis/ce/match/expression/age_calculator.rb
eanders: That sure seems like a nice approach and pattern.

drivers/hmis/spec/models/hmis/ce/match/expression/cde_field_map_spec.rb
ttoomey: * I'll adjust the spec to give confidence it's returning the most recent assessment
* we can take out date_updated here if we're sorting on assessment_date I think since it should not be relevant
* we can check the behavior when two assessments have the same assessment date that id is correctly used as a tie breaker

drivers/hmis/app/models/hmis/hud/custom_data_element_definition.rb
```diff
@@ -44,28 +55,11 @@ class Hmis::Hud::CustomDataElementDefinition < Hmis::Hud::Base
 
   use_enum_with_same_key :form_role_enum_map, FIELD_TYPES.map { |f| [f, f.to_s.humanize] }.to_h
 
-  def read_value_from(custom_data_element)
```
ttoomey, line 47: this method is unused, I'm removing it

drivers/hmis/app/models/hmis/ce/match/expression/homeless_days_calculator.rb
gigxz: We don't have a use-case for a days homeless calculation in the MVP CE implementation, so I would advocate for not spending time on it at this point. I think we can remove it altogether and create a ticket for implementing when needed.

eanders: Works for me!

drivers/hmis/app/models/hmis/ce/match/engine.rb
```diff
@@ -129,18 +150,31 @@ def call(clients, progress: false)
 
     private
 
+    def log_info(message = nil, &block)
```
gigxz, line 153: thanks for adding this logging!

drivers/hmis/app/models/hmis/ce/match/expression/last_enrolled_days_calculator.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Hmis::Ce::Match::Expression
+  # Calculates the number of days since a client's last enrollment ended.
+  # Returns 0 if the client is still enrolled (no exit date).
+  class LastEnrolledDaysCalculator
```
ttoomey, line 6: there's a bug here, fixing...

drivers/hmis/app/models/hmis/ce/match/expression/homeless_days_calculator.rb
gigxz: It's captured in https://github.com/open-path/Green-River/issues/7840

drivers/hmis/app/models/hmis/ce/match/expression/last_enrolled_days_calculator.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+module Hmis::Ce::Match::Expression
+  # Calculates the number of days since a client's last enrollment ended.
+  # Returns 0 if the client is still enrolled (no exit date).
```
gigxz: Specify that WIP enrollments are counted. I think this is true because we don't join to project at all. Probably worth having a test for it, though. 
```suggestion
  # Calculates the number of days since a client's last enrollment ended.
  # Returns 0 if the client is still enrolled (no exit date), even if the currently open enrollment is WIP
```

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
```diff
@@ -34,77 +46,88 @@ def arel
...
+        days_homeless: days_homeless_field,
+        open_enrollment_project_types: open_enrollment_project_types_field,
+        open_enrollment_project_types_excluding_incomplete: open_enrollment_project_types_excluding_incomplete_field,
+        open_referral_project_types: open_referral_project_types_field,
+      }
+    end
+
+    private
+
+    def days_since_last_exit_field
```
gigxz, line 60: This looks good! Am I correct in understanding:
* "days since last exit", "veteran status", and "current age" are currently the only fields that affect pre-filtering, since they have `arel_field` defined
* CDE fields and open_enrollment_project_types are now evaluate on a _batch_ of clients instead of an individual client. Previously they were N+1 where N is the number of clients being evaluated, and now they are N+1 where N is the number of batches being evaluated

gigxz (approved):
looks great! 🙏 💯

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
ttoomey: yes, that's all correct

drivers/hmis/app/models/hmis/ce/match/expression/last_enrolled_days_calculator.rb
ttoomey: good point, I'll add a test for it

drivers/hmis/app/models/hmis/ce/match/expression/homeless_days_calculator.rb
ttoomey: removed Days homeless

ttoomey:
I removed days_homeless and fixed an issue in last-enrolled-days calculator. I also made another pass through to tidy up and enhance the tests. 
@gigxz  - okay to merge in the changes since you approved? (4b13f5a587122c958f8de9a8b8aa88cb5b7219d1)

gigxz (approved):

PR #5610, opened 2025-07-31
Make CDED generation more flexible on publish and seeding

gigxz:
## _Merging this PR_

## Description
This change makes CDED generation more permissive on publish. If it finds an item with mapping: { custom_field_key: 'x' }, it will look for a CDED with that key. If it doesn't find one, it will create it. If it finds one that conflicts (eg for a different form) it will raise.

Benefits
* This will allow us to copy the full JSON from a staging environment, paste it into a prod environment editor, and publish. 
* This will make it easier to add managed-in-version-control forms that include custom fields. Currently we support that but we need to go manually create the CDEDs in each environment. This automates the CDED creation on seed.

Issue: https://github.com/open-path/Green-River/issues/7978

How to test:
- [ ] seed all client forms
  - [ ] any seeded forms that contain custom_field_key mappings should have CDEDs generated/updated
- [ ] test publishing form that references custom_field_keys that don't exist. it should create them.
- [ ] test publishing form that references custom_field_keys that do exist. it should use them and update the label.
- [ ] test publishing form that references custom_field_keys that do exist but have a different field_type or repeats value. it should raise
- [ ] test publishing form that has new questions without mappings. it should still create a new CDED and populate the mapping custom_field_key value with a generated key

## Type of change
new feature


drivers/hmis/app/graphql/mutations/publish_form_definition.rb
```diff
@@ -27,98 +29,34 @@ def resolve(id:)
...
+
       {
         form_identifier: definition,
       }
     end
-
-    # Adds missing `{mapping: {custom_field_key: '...'}}` to all questions in the form definition
-    # Mutates definition.items, adds `mapping.custom_field_key`
-    # Returns array of initialized CustomDataElementDefinitions to be saved (for new questions only)
-    def add_missing_custom_field_keys(definition)
```
gigxz, line 55: This logic was moved to the new CustomDataElementGenerator class

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -615,75 +615,10 @@ def updates_client_or_enrollment?
     link_id_item_hash.values.find { |item| ['ENROLLMENT', 'CLIENT'].include?(item.mapping&.record_type) }.present?
   end
 
-  # Find and/or initialize CustomDataElementDefinitions that are collected by this form.
-  # For application forms, we now rely on PublishFormDefinition to generate CDEDs, but
-  # this is still used in test fixtures and in PublishExternalFormsJob.
-  def introspect_custom_data_element_definitions(set_definition_identifier: false, data_source: GrdaWarehouse::DataSource.hmis.first)
```
gigxz, line 621: This is no longer needed,  usages replaced with CustomDataElementGenerator

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -42,6 +42,49 @@ def perform(document, role = nil, skip_cded_validation: false)
     @issues.errors
   end
 
+  # Method to validate individual CustomDataElementDefinition (CDED) against an associated form item.
+  # This is pulled out for reuse by the CustomDataElementGenerator
+  def self.validate_cded(item:, cded:)
```
gigxz, line 47: I made no changes to this logic, just pulled it out into a class method so the generator class could use it

drivers/hmis/app/models/hmis/form/custom_data_element_generator.rb
```diff
@@ -0,0 +1,182 @@
...
+          item.mapping&.field_name
+      end
+
+      def lookup_mapped_cded(item)
+        custom_field_key = item.mapping&.custom_field_key
+        return unless custom_field_key
+
+        owner_type = determine_owner_type(item)
+        found = Hmis::Hud::CustomDataElementDefinition.find_by(owner_type: owner_type, key: custom_field_key, data_source: @data_source)
+        raise "Form '#{@definition.identifier}' item '#{item.link_id}' references a CDED '#{custom_field_key}' belongs to a different form ('#{found.form_definition_identifier}')" if found&.form_definition_identifier.present? && found&.form_definition_identifier != @definition.identifier
```
martha, line 143: Do we still have any forms with cdeds kicking around from before we made the decision that forms should not share cdeds? I remember we did used to have some examples of this. If so, I think this will cause them to always fail on publish.

drivers/hmis/app/models/hmis/form/definition_validator.rb
```diff
@@ -349,51 +392,25 @@ def get_cded(item, role)
...
-        end
+        self.class.validate_cded(item: child_item, cded: cded)
       end
     end
     cded_check.call(document)
   end
 
   private def raise_bad_type_match(link_id, item_type, cded_key, cded_type)
-    raise "Item #{link_id} has type #{item_type}, but its custom field key #{cded_key} has an incompatible type #{cded_type}"
+    self.class.raise_bad_type_match(link_id, item_type, cded_key, cded_type)
```
martha: Is this still used?

martha (approved):
Great, thanks Gig!

drivers/hmis/app/models/hmis/form/custom_data_element_generator.rb
gigxz: It's OK for forms to share CDEDs. For example we have lots of Service forms that share a "service note" field in one environment, which was intentional for reporting purposes. In those cases, though, the referenced CDED should _not_ have a form_definition_identifier specified. As long as that's true, this_ shouldn't raise, because `found&.form_definition_identifier.present?` will evaluate to false.

I will look into whether it's possible that we have cross-referenced CDEDs  that _do_ have form identifiers listed. I think (hope) those would fail validation.

PR #5614, opened 2025-07-31
Add alt-AHA calculator

martha:
## _Merging this PR_
https://github.com/greenriver/hmis-frontend/pull/1222

## Description
- Add new `hmis_scoring_rules` table and `AcHmis::Scoring::Rule` model. Each row represents a rule, such as "if this link ID exactly matches X value, assign Y weight"
- Add mutation `AcHmis::CalculateAltAhaScore`
- Add calculator `HmisExternalApis::AcHmis:: AltAhaCalculator`
- Add a pattern for returning form validations from within the form processor, which we didn't previously support.
- Add script for inserting rules to the db based on a CSV

## Type of change
New feature, custom-build

- [x] I have updated the documentation (or not applicable) - see documentation in internal google doc
- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - I added a "Dev testing instructions" comment on the linked issue


drivers/hmis/app/graphql/types/forms/enums/component.rb
```diff
@@ -24,6 +24,7 @@ class Forms::Enums::Component < Types::BaseEnum
     value 'RADIO_BUTTONS_VERTICAL', 'Render a choice input item as vertical radio buttons', deprecation_reason: 'All radio buttons are now vertical'
     value 'SSN', 'SSN input component'
     value 'MINUTES_DURATION', 'Duration component with hours and minutes, value stored as minutes'
+    value 'ALT_AHA', 'Alt AHA component'
```
gigxz: move down to "Specialized input components"?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -461,7 +461,7 @@ def assessment_date_item
   # validate form_values provides against the definition
   #   * errors & warnings on missing required fields
   #   * check if the input ids match the definition
-  def validate_form_values(form_values)
+  def validate_form_values(form_values, client: nil)
```
martha: To discuss - We had previously landed on this `validate_form_values` as a potentially good place to validate the final score matches the values submitted. However, to do this we would need to pass in `client`, in order to (re)evaluate the full aha. Is this worth it? Should this validation actually be moved somewhere else, like into the form processor?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/alt_aha_calculator.rb
```diff
@@ -0,0 +1,164 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module HmisExternalApis::AcHmis
+  class AltAhaCalculator
+    ALT_AHA_NAMESPACE = 'alt_aha'
+
+    # owner can be: an enrollment (when AHA score is calculated on an unsaved assessment), or an assessment (when the assessment is being saved)
```
martha, line 13: To discuss: Do we still want to re-save the calculated score with a new owner when the assessment is saved? Where would be the best place to do that? Custom client processor code, similar to MCI?

drivers/hmis/app/models/hmis/form/definition.rb
```diff
@@ -487,15 +487,22 @@ def validate_form_values(form_values)
...
+      end
+
       # Validate required status
       if item.required && is_missing
         errors.add field_name || :base, :required, **error_context
       elsif item.warn_if_empty && (is_missing || is_data_not_collected)
         errors.add field_name || :base, :data_not_collected, severity: :warning, **error_context
       end
-
-      # TODO(##184404620): Validate ValueBounds (How to handle bounds that rely on local values like projectStartDate and entryDate?)
```
martha, line 497: Removing these, per https://github.com/open-path/Green-River/issues/5837#issuecomment-2414071045

drivers/hmis/lib/tasks/ac_hmis_scoring_import.rake
```diff
@@ -0,0 +1,211 @@
+# Import AC HMIS scoring rules from two CSVs: a Variable Definitions file and a Variable Weights file.
```
martha: name it something with the date (and move to the same dir as other one-time imports?)

drivers/hmis/app/models/ac_hmis/scoring/calculation_log.rb
```diff
@@ -0,0 +1,20 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module AcHmis
```
gigxz: AcHmis-specific models should go in the external apis driver. However it looks like these are not AcHmis-specific. Can the scoring module can be moved to `drivers/hmis/app/models/hmis`?

I see we have an existing `AcHmis::ReportApi` in here which should be moved to `drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis`

drivers/hmis/app/models/ac_hmis/scoring/rule.rb
```diff
@@ -0,0 +1,119 @@
...
+        return false if numeric_value.nil?
+
+        # gather up all the range-related criteria in a list and check if they all match.
+        criteria_config.all? do |operator, threshold|
+          case operator
+          when 'gt' then numeric_value > threshold
+          when 'gte' then numeric_value >= threshold
+          when 'lt' then numeric_value < threshold
+          when 'lte' then numeric_value <= threshold
+          else true # ignore unknown operators
```
gigxz: suggest raising here so we don't silently ignore misconfigurations. This would be unexpected since we have validations on this shape

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
```diff
@@ -58,6 +58,9 @@ class Hmis::Hud::CustomAssessment < Hmis::Hud::Base
   has_one :ce_event, through: :form_processor
   has_one :clh_location, through: :form_processor
 
+  # Calculation logs for scoring algorithms
+  has_many :calculation_logs, class_name: 'AcHmis::Scoring::CalculationLog', as: :owner, autosave: true
```
gigxz: I'm not sure about the autosave approach. It feels like a fair amount of complexity (linking them in the CustomAssessmentProcessor, exposing calculate_score and calculate_score! differently) just to avoid having some duplication in the logs.

I think I may be walking back an earlier statement, but looking at the complexity it adds, I'm not convinced that it's worth it the complexity/maintenance burden. It would be much cleaner if `calculate_score` just always handled storing a log, and none of the callers had to know/care about that. That would also be less brittle to future changes, where we would make a change an inadvertently stop the log from being persisted. What do you think, am I missing something?

drivers/hmis/app/models/hmis/hud/processors/custom_assessment_processor.rb
```diff
@@ -26,5 +22,40 @@ def schema
 
     def information_date(_)
     end
+
+    def process_custom_field(field, value)
```
gigxz: Is this validation something we can do in post-processing rather than overriding the process_custom_field function?
Overriding `process_custom_field` would be a new pattern. The `post_process` hook is explicitly there to let processors do any extra processing after the fields have all been iterated through.

here's my (untested) stab at what that would look like. It also removes the magic string of the link_id and relies on the component attribute instead (better or worse? I'm not sure... still feels a bit messy but I think I prefer it to hard-coding the link ID which doesn't appear to be hard coded anywhere else)


```ruby
    def post_process
      validate_alt_aha_score
    end

    private

      # Validate Alt-AHA score if present. For more information on Alt-AHA score see https://docs.google.com/document/d/1Gcz9-t_utRcqGV9xCzQvTehjQOCqqPv_5-JY_IhhL4Q/edit?tab=t.0#heading=h.kpe3ch74jsjj
    # If assessment contained an Alt-AHA score calculation, validate it by re-calculating the score.
    # If the scores don't match, return a validation error.
    # This prevents users from submitting scores that do not match the current form values.
    def validate_alt_aha_score
      alt_aha_item = @processor.definition.link_id_item_hash.values.find { |i| i.component == 'AltAHA' }
      return unless alt_aha_item # alt-aha not collected on this form, nothing to validate

      alt_aha_custom_field_key = alt_aha_item&.dig('mapping', 'custom_field_key')
      raise 'msg' unless alt_aha_custom_field_key # unexpected, alt aha item should map to a custom field

      return unless @values.fetch(alt_aha_custom_field_key) # something like this? we can skip if there's no score to validate

      # Recalculate score to confirm that it is unchanged
      aha_calculator = HmisExternalApis::AcHmis::AltAhaCalculator.new(
        values_by_link_id: @processor.values,
        client: @processor.client_factory,
        user: @processor.current_user,
        owner: @processor.owner_factory, # owner is the CustomAssessment that's also the owner of this FormProcessor
        form_definition_identifier: @processor.definition.identifier,
      )
      score = aha_calculator.calculate_score!

      return if score == value # submitted score and recalculated score match, nothing to validate

      @processor.add_processing_error(
        HmisErrors::Error.new(
          alt_aha_custom_field_key,
          readable_attribute: alt_aha_item.brief_text || alt_aha_item.text,
          link_id: alt_aha_item.link_id,
          section: @processor.definition.link_id_section_hash[alt_aha_item.link_id],
          message: 'value has changed, and needs to be recalculated before submission',
        ),
      )
    end
```

My example also removes the calculation log stuff, per my other comment, but that's unrelated

drivers/hmis/app/graphql/mutations/submit_assessment.rb
```diff
@@ -97,6 +99,12 @@ def resolve(input:, assessment_lock_version: nil)
       # Run processor to create/update related records
       assessment.form_processor.run!(user: current_user)
 
+      # Validation errors encountered during form processing.
+      # These are errors that should be returned to the user because they are fixable,
+      # but were surfaced during the form processor run (as opposed to the validation steps).
+      processing_validation_errors = assessment.form_processor.processing_errors&.errors
+      errors.push(*processing_validation_errors) if processing_validation_errors&.any?
```
gigxz: This will be helpful in the future. It feels wonky to only support this for individual assessments, though, and not for other record forms (`SubmitForm` mutation) and validating household assessments (`SubmitHouseholdAssessments` mutation).

 Maybe it would be better to include the processing errors in the `collect_record_validations` method? That would slightly change the meaning of that function, but I think it makes sense to group them together (record errors + processor errors) because they're both errors that are generated from the processor _running_. (As opposed to the structural "form validation" errors that we collect prior to running the processors..)

gigxz (commented):
In-progress review (I haven't yet pulled down to test) but looks great so far! Happy to talk through any suggestions

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
martha: > much cleaner if calculate_score just always handled storing a log, and none of the callers had to know/care about that

totally, I agree with this. The only thing is, we then couldn't link the calculation log item to its owner assessment -- because the assessment hasn't been persisted yet at the point when the calculation is made. That seems okay to me, I think, although we lose some ability to do precise forensics. What do you think?

drivers/hmis/app/graphql/mutations/submit_assessment.rb
martha: Okay, agreed! I can move this into `collect_record_validations` and maybe rename that function to more clearly indicate what it's doing.

drivers/hmis/app/models/ac_hmis/scoring/calculation_log.rb
martha: Sounds good, including the `ReportApi` move (https://github.com/greenriver/hmis-warehouse/pull/5614/commits/a1068b6730a376280b8023166c24b68a31377d20) but this is untested; maybe we can pair and walk through testing to make sure I haven't introduced any bugs?

drivers/hmis/app/models/hmis/hud/processors/custom_assessment_processor.rb
martha: Thanks for the suggestion! 👍

drivers/hmis/app/models/hmis/hud/processors/custom_assessment_processor.rb
```diff
@@ -26,5 +22,48 @@ def schema
...
+    def validate_alt_aha_score
+      alt_aha_item = @processor.definition.link_id_item_hash.values.find { |i| i.component == 'ALT_AHA' }
+      return unless alt_aha_item # alt-aha not collected on this form, nothing to validate
+
+      alt_aha_custom_field_key = alt_aha_item.dig('mapping', 'custom_field_key')
+      raise 'Alt-AHA form item does not map to a custom field' unless alt_aha_custom_field_key # unexpected
+
+      value = @hud_values.fetch(alt_aha_custom_field_key)
+      return unless value # Skip if there's no score to validate
+
```
gigxz, line 46: We should probably add this as well, if value is submitted as "hidden" it would still be present in hud_values obj
```suggestion
return if value == HIDDEN_FIELD_VALUE

```

drivers/hmis/app/models/hmis/scoring/rule.rb
```diff
@@ -0,0 +1,119 @@
...
+    class Rule < ::GrdaWarehouseBase
+      self.table_name = 'hmis_scoring_rules'
+
+      RANGE = 'range'
+      EXACT_MATCH = 'exact_match'
+      VALUE = 'value'
+      INCLUDE = 'include'
+      CRITERIA_TYPES = [RANGE, EXACT_MATCH, VALUE, INCLUDE].freeze
+
+      validates :link_id, :form_definition_identifier, :algorithm, :criteria_type, :weight, presence: true
```
gigxz: I don't think this is necessary since they are non-nullable in the database, this will be automatically validated

drivers/hmis/app/graphql/mutations/submit_assessment.rb
gigxz: looks great thanks!

drivers/hmis/app/models/hmis/hud/custom_assessment.rb
gigxz: Hi circling back here, the change you made looks good. I think it's ok, I'm thinking of this log analogous to the external_request_log which similarly doesn't record specific context about any entity(ies) that the request was associated with. Just which user performed it and when. I think that is OK for alt aha too. If we're trying to do forensincs on particular requests we can look at `Hmis::ActivityLog.where(operation_name: "CalculateAltAhaScore")` in a time range.

drivers/hmis/app/models/ac_hmis/scoring/calculation_log.rb
gigxz: I tested these locally and confirmed that they are both being routed to the correct controllers!

gigxz (approved):
looks great! I pushed up rubocop auto fix to the rake task. Rake task worked for me though it did log that it skipped 4 rows, which I'm not sure is something to be concerned about or not.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/alt_aha_calculator.rb
```diff
@@ -0,0 +1,162 @@
...
+      return 7 if total_points >= 7 && total_points < 8
+      return 6 if total_points >= 6 && total_points < 7
+      return 5 if total_points >= 5 && total_points < 6
+      return 4 if total_points >= 4 && total_points < 5
+      return 3 if total_points >= 3 && total_points < 4
+      return 2 if total_points >= 1 && total_points < 3
+      return 1 if total_points >= 0 && total_points < 1
+
+      0
+    end
```
gigxz: using ranges might improve the readability here
```ruby
    def convert_total_points_to_score(total_points)
      case total_points
      when 11...16 then 10
      when 9...11  then 9
      when 8...9   then 8
      when 7...8   then 7
      when 6...7   then 6
      when 5...6   then 5
      when 4...5   then 4
      when 3...4   then 3
      when 1...3   then 2
      when 0...1   then 1
      else 0
      end
    end
```

martha: Yep that seems nicer, thanks!

PR #5616, opened 2025-08-01
Support multiple priority schemes

martha:
## _Merging this PR_

## Description
- Database: 
  - Update `ce_match_candidates` table to have an array column for `priority_scores` rather than a single priority int
  - Update `ce_match_rules` with a `rank` column
- API Schema
  - Update `CeCandidate` to have a `priorityScores` int array, and deprecate the old `priorityScore` field
  - Update `CeMatchRule`, `Unit`, and `UnitGroup` which previously resolved a single `priorityScheme` to resolve an array of `prioritySchemes` ordered by rank
- Update logic for candidate pool resolving and candidate generation to evaluate multiple priority schemes using a dentaku array expression with curly braces: `{scheme1, scheme2, ...}`

## Testing notes
- Create a candidate pool that has multiple ranked priority rules.
- I haven't made this backwards-compatible with the previous single priority schemes. So, when testing this PR, you will need to update or remove the `ce_match_candidate_pools` in your database that have priority expressions that look like `cde.custom_assessment.my_score` (instead of `{cde.custom_assessment.my_score}`). Let me know if you think we should put more time into improving backwards-compatibility here.

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/match/candidate_pool_resolver.rb
```diff
@@ -53,10 +53,54 @@ def load_candidate_pools_by_key
...
+    # because `rank` is unique on owner, not unique globally.
+    # This function filters down the list of priority rules that apply to this opportunity:
+    # If there are rules defined with different owners, pick the most specific owner,
+    # and return all rules applying to that owner
+    def most_specific_rules(rules)
+      return [] if rules.empty?
+
+      # Owner specificity hierarchy: unit > unit_group > project > organization > data_source
+      specificity_order = {
+        'Hmis::Unit' => 1,
```
martha: remove Unit here per https://github.com/open-path/Green-River/issues/8044

drivers/hmis/app/models/hmis/ce/match/candidate_pool_resolver.rb
```diff
@@ -65,9 +65,52 @@ def key_for_opportunity(opportunity:)
       rules = opportunity_rules(opportunity)
 
       key = []
-      key << (rules.filter(&:priority_scheme?).first&.expression || '0')
-      key << (rules.filter(&:eligibility_requirement?).map(&:expression).join(' AND ') || 'TRUE')
+      key << priority_expression_for_rules(rules)
```
martha: @ttoomey I'm thinking through this todo from the [ticket](https://github.com/open-path/Green-River/issues/7957#issuecomment-3128159511), and would like to check with you.
> Need some special logic to determine if the rank positions change, because that should trigger recalculating the pool.

I _think_ that no additional logic is needed here, with this implementation. When the rank positions of priority rules change, the `key` based on the priority expression changes, and the candidate pool will be updated (or created, if none exists yet for that combination of priority/eligibility), and all the associated opportunities will be marked as stale, per 7992. Does that sound right to you?

db/warehouse/migrate/20250731134651_add_numeric_priority_scores_to_ce_match_candidates.rb
```diff
@@ -0,0 +1,25 @@
...
+
+# frozen_string_literal: true
+
+class AddNumericPriorityScoresToCeMatchCandidates < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
+
+  def change
+    # Add priority_scores, an array of integers for prioritization
+    add_column :ce_match_candidates, :priority_scores, :integer, array: true
+    add_index :ce_match_candidates, :priority_scores, using: :btree, algorithm: :concurrently
```
ttoomey: this might be fine; however we've had issues with async (concurrent) indexes in the past. If the migration fails it's not rolled back (due to `disable_ddl_transaction`). I suspect that none of our installations will have that many candidates yet, so building the index should be pretty fast. So I suggest removing `disable_ddl_transaction` and `algorithm: :concurrently` and enclosing with `safety_assured`

db/warehouse/migrate/20250731134651_add_numeric_priority_scores_to_ce_match_candidates.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    # Remove existing priority_score single integer column.
+    # This migration doesn't port over existing priority scores, since we can just regenerate candidates
+    safety_assured { remove_column :ce_match_candidates, :priority_score, :integer }
+
+    # Add rank to match rules, so we know the order to apply priority rules
+    add_column :ce_match_rules, :rank, :integer
+    add_index :ce_match_rules, [:owner_type, :owner_id, :rank], unique: true, name: 'index_ce_match_rules_owner_rank_unique', algorithm: :concurrently
```
ttoomey: minor: because rules can be either eligibility requirements or priorities, perhaps this field should be called `priority_rank` to make it clearer that it doesn't apply to eligibility requirements. Also the unique constraint can be conditional on rule_type

```suggestion
    add_column :ce_match_rules, :priority_rank, :integer
    add_index :ce_match_rules, [:owner_type, :owner_id, :priority_rank], unique: true, name: 'index_ce_match_rules_owner_rank_unique', where: "rule_type = 'priority_scheme'"

File: drivers/hmis/app/models/hmis/ce/match/candidate_pool_resolver.rb (hunk 2 above)
Comment (ttoomey) [reply drivers/hmis/app/models/hmis/ce/match/candidate_pool_resolver.rb]: yes, that sounds exactly right. Thanks for checking!

**Review by ttoomey — APPROVED** [review]
This looks great! I left some suggestions

Should probably retarget to r177; there is a bit of a merge conflict though. [I think I resolved the conflict in 5657](https://github.com/greenriver/hmis-warehouse/pull/5657)

File: drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb [hunk 5]
```diff
@@ -20,11 +20,17 @@ class HmisSchema::UnitGroup < Types::BaseObject
...
-    field :priority_scheme, HmisSchema::CeMatchRule, null: true
+    field :priority_scheme, HmisSchema::CeMatchRule, null: true, deprecation_reason: 'Replaced by prioritySchemes'
+    field :priority_schemes, [HmisSchema::CeMatchRule], null: true
     field :workflow_template_identifier, String, null: true
     field :workflow_template_name, String, null: true
     # TODO(#7538) resolve default contacts for workflow template
 
+    # TODO(#7957) - remove after deprecation period
+    def priority_scheme
+      priority_schemes&.first
```
gigxz, line 35: where is `priority_schemes` defined?

also: see comment https://github.com/greenriver/hmis-warehouse/pull/5657/files#r2282296605 about potentially updating this to resolve the priority schemes that would be selected for evaluation, rather than all applicable priority schemes.

drivers/hmis/app/models/hmis/ce/match/candidate_pool_resolver.rb
```diff
@@ -65,9 +65,52 @@ def key_for_opportunity(opportunity:)
...
+    # However, an opportunity can't respect multiple different rules with different owners,
+    # because `rank` is unique on owner, not unique globally.
+    # This function filters down the list of priority rules that apply to this opportunity:
+    # If there are rules defined with different owners, pick the most specific owner,
+    # and return all rules applying to that owner
+    def priority_rules(rules)
+      rules = rules.filter(&:priority_scheme?)
+      return [] if rules.empty?
+
+      # Owner specificity hierarchy: unit_group > project > organization > data_source
```
gigxz: This is duplicative with by_owner_precedence but it looks like that is resolved by Theron's merge conflict PR

drivers/hmis/app/models/hmis/ce/match/README_FOR_CE_MATCH_ENGINE.md
martha, line 1: @ttoomey was this file deleted on purpose?

ttoomey: sort of, I renamed it to `drivers/hmis/app/models/hmis/ce/README_FOR_CE_PROCESSING.md` to better reflect the documentation scope. I forgot to do a `git mv` though so it just shows up as deleted

martha: Ok gotcha, thank you!

martha: I think this is ready to be merged to release-177; I'll wait for tests to pass and then merge it in.

ttoomey: Oh wait a minute, I think it was the other markdown file that was renamed ("README_FOR_CHANGE_MARKER"). This one actually was unintentionally removed somehow in my merge. Sorry about that!

I'll push up a change to restore it. Thanks for catching!

PR #5622, opened 2025-08-04
Structure File S3 dumps

dtgreiner:
## _Merging this PR_

## Description

Include structure files in S3 metrics dumps on a weekly basis. This will allow us to more easily check for discrepancies in the deployed structure files.

## Type of change

New Feature


spec/models/app_resource_monitor/report_spec.rb
```diff
@@ -0,0 +1,173 @@
...
+
+    # Create an enrollment if the model exists
+    if defined?(GrdaWarehouse::Hud::Enrollment)
+      create(:grda_warehouse_hud_enrollment,
+             data_source: data_source,
+             personal_id: client.personal_id)
+    end
+
+    # Create a project if the model exists
+    create(:grda_warehouse_hud_project, data_source: data_source) if defined?(GrdaWarehouse::Hud::Project)
```
eanders: When would the warehouse classes not be defined?  Those aren't part of a driver, so I _think_ they should always be there.

eanders: Any reason to just just add these as `let`s at the top?  Looks like we're redefining on each test.

eanders (approved):
I think this looks good, but maybe worth cleaning up the tests a bit.

spec/models/app_resource_monitor/report_spec.rb
dtgreiner: Thanks for catching that! I meant to clean it up after I got the tests working. Changes are in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5622/commits/16108a0df4528c6da17891da592fa2c952122b0d).

PR #5632, opened 2025-08-06
7937-low-latency-ce-matches

ttoomey:
## _Merging this PR_

## Description
[See full technical description below](https://github.com/greenriver/hmis-warehouse/pull/5632#issuecomment-3161866654)

## Objective
Decouple fast client processing from slow pool processing to eliminate head-of-line blocking in Coordinated Entry updates. Replace single `ProcessChangesJob` with two concurrent, specialized jobs using per-pool advisory locks.

## Current Problem
`Hmis::Ce::ProcessChangesJob` handles both pool updates (slow, minutes-hours) and client matching (fast) sequentially. Pool processing blocks all client updates until complete.

## Solution
Two new jobs with PostgreSQL advisory locks for coordination:

### ProcessPoolsJob (long_running queue)
- Processes dirty candidate pools exclusively
- Acquires exclusive 60-second blocking locks per pool (`hmis-ce_pool-#{id}`)
- Skips pools on lock timeout, logs warning
- Yields to client job when clients become dirty

### ProcessClientsJob (default queue)  
- Processes dirty clients against all active pools
- Attempts 5-second lock acquisition per pool
- Skips locked pools, only marks clients processed if evaluated against ALL pools

## Trade-offs
- **Eventual consistency**: Client eligibility updates deferred for actively processing pools
- **Low latency**: Client updates no longer blocked by pool processing
- **Inefficient processing**: Clients remain dirty until evaluated against all pools, may be processed repeatedly

## Risk Mitigation
- Distinct lock scopes prevent deadlocks
- Skip-on-failure prevents indefinite blocking  

## Type of change
New feature


ttoomey:
# Technical Design: Concurrent CE Client and Pool Processing

## 1. Objective

To provide low-latency eligibility updates for Coordinated Entry (CE) clients by decoupling the fast-running client processing from the slow-running pool processing. This will prevent long-running pool updates from blocking timely client-matching, significantly improving responsiveness for users. We will achieve this by replacing the single `ProcessChangesJob` with two distinct, concurrent jobs that use a per-pool advisory locking mechanism to coordinate their work safely.

## 2. Background: Current Architecture & Limitations

Currently, all CE change processing is handled by a single background job, `Hmis::Ce::ProcessChangesJob`. This job runs on a fixed schedule and performs two distinct tasks:
1.  **Processes Dirty Pools:** Updates eligibility rules for candidate pools. This can be a very slow operation (minutes to over an hour).
2.  **Processes Dirty Clients:** Matches clients with newly changed data against candidate pools. This is typically a very fast operation.

The primary limitation is **head-of-line blocking**. When the job begins processing a large pool, it prevents any new client updates from being handled until the pool processing is complete.

## 3. Proposed Solution: A Two-Job, Concurrent Architecture

We will replace the single `ProcessChangesJob` with two new, specialized jobs. Coordination will be managed via PostgreSQL advisory locks at both the job level and the individual pool level. Both jobs will be wrapped in `instrument_as_maintenance_task` for consistent monitoring and will follow existing logging patterns.

### 3.1. Job-Level Locking

Each job will use a unique advisory lock to ensure only one instance of that job can run at a time. Following the existing pattern in the codebase, locks will be acquired using `GrdaWarehouseBase.with_advisory_lock` with the job class name as the lock identifier and a `timeout_seconds: 0` to prevent blocking if an instance is already running.

### 3.2. Per-Pool Locking for Coordination

The core of this design is a **per-pool advisory lock** that signals a specific pool is actively being processed. The lock name is generated with the format `hmis-ce_pool-#{id}` to ensure uniqueness.

### 3.3. The New Jobs

#### A. `ProcessPoolsJob`
*   **Purpose:** Exclusively processes dirty candidate pools.
*   **Queue:** `long_running`
*   **Frequency:** Configurable (e.g., every 5 minutes).
*   **Logic:**
    1.  Acquire a non-blocking, job-level lock.
    2.  Fetch a batch of dirty pools.
    3.  For each `pool`, acquire an **exclusive, blocking lock** with a 60-second timeout.
        *   If the lock fails (times out), log a warning and skip to the next pool. The pool will remain dirty and be retried on the next run.
    4.  Inside the lock block, execute the match engine for all clients against that single pool: `Hmis::Ce::Match::Engine.call(pool, progress: @progress)`.
    5.  Mark the pool as processed.
    6.  **Yielding Logic:** After processing each pool, check if any clients have become dirty. If so, exit the job to allow `ProcessClientsJob` to run, ensuring better responsiveness.
    7.  The lock is automatically released.
*   **Progress Support:** Supports optional progress bar display via `progress` parameter.

#### B. `ProcessClientsJob`
*   **Purpose:** Exclusively processes dirty clients for fast updates.
*   **Queue:** `default` (short-running).
*   **Frequency:** Configurable, with a short interval (e.g., every 30 seconds) for low latency.
*   **Logic:**
    1.  Acquire a non-blocking, job-level lock.
    2.  Fetch a batch of dirty clients.
    3.  Fetch all *active* candidate pools.
    4.  Iterate through each active pool.
    5.  For a given `pool`, attempt to acquire a lock with a **short, 5-second timeout**.
    6.  **If the lock is acquired:** Run the match engine for the dirty client batch against that pool (`Hmis::Ce::Match::Engine.call(pool, clients: ..., progress: @progress)`).
    7.  **If the lock is NOT acquired:** The pool is busy. Log a debug message and skip it.
    8.  After iterating through all pools, check if any pools were skipped.
        *   **If no pools were skipped:** Mark the client batch as processed.
        *   **If any pool was skipped:** The clients remain dirty to ensure they are re-evaluated against the skipped pool(s) on the next run.
*   **Progress Support:** Supports optional progress bar display via `progress` parameter.

## 4. Data Consistency Model: Eventual Consistency

This design explicitly chooses low latency over absolute, immediate consistency. A client's eligibility for a pool that is actively being processed will be deferred until the pool processing is complete. This is an acceptable and desirable trade-off.

To ensure correctness, clients will only be marked as processed if they are successfully evaluated against **all** active candidate pools. If `ProcessClientsJob` skips a locked pool, the clients remain dirty, guaranteeing they will be re-processed in a subsequent run.

## 5. Error Handling and Monitoring

### 5.1. Lock Acquisition Failures
*   **Pool Lock Failures:** If a pool lock consistently fails to be acquired, the pool will remain dirty and be retried on subsequent job runs. The job will skip the pool and continue processing other pools, logging warnings for visibility.
*   **Job Lock Failures:** If a job cannot acquire its job-level lock (because another instance is running), it will exit gracefully.

### 5.2. Monitoring and Alerting
*   **Maintenance Task Instrumentation:** Both jobs are wrapped in `instrument_as_maintenance_task` to provide consistent monitoring, alerting, and performance tracking.
*   **Lock Contention Metrics:** Warning logs will provide visibility into pools that are frequently locked, allowing for operational tuning and identification of problematic pools.

## 6. Risks and Mitigations

*   **Risk:** Increased implementation complexity.
    *   **Mitigation:** Each job's logic is small and highly focused. Following existing patterns in the codebase (advisory locks, maintenance task instrumentation) reduces implementation complexity.
*   **Risk:** Lock contention or deadlocks.
    *   **Mitigation:** The distinct lock scopes (job-level vs. per-pool) and the skip-on-failure strategy in the client processor are designed to prevent deadlocks and indefinite blocking.
*   **Risk:** Jobs failing to complete due to consistent lock acquisition failures.
    *   **Mitigation:** The maintenance task instrumentation will alert when jobs don't complete (`run.complete!` not reached), providing operational visibility into persistent issues.

lib/tasks/grda_warehouse.rake
```diff
@@ -407,8 +407,11 @@ namespace :grda_warehouse do
 
     if HmisEnforcement.hmis_enabled? && GrdaWarehouse::DataSource.hmis.exists? && Hmis::Ce.configuration.enabled?
       safely_execute do
-        # enqueue CE processor if it's not already running. Once enqueued it should self-sustain
-        Hmis::Ce::ProcessChangesJob.enqueue_if_not_already_running(wait_time: 5.minutes)
+        # enqueue CE processors if they're not already running. Once enqueued they should self-sustain
+        # ProcessPoolsJob handles expensive pool processing on a longer interval
+        Hmis::Ce::ProcessPoolsJob.enqueue_if_not_already_running(wait_time: 5.minutes)
+        # ProcessClientsJob handles fast client processing on a shorter interval
+        Hmis::Ce::ProcessClientsJob.enqueue_if_not_already_running(wait_time: 30.seconds)
```
ttoomey: Timing could be adjusted. Maybe 30 seconds is to agressive. And perhaps pool processing could poll every 10 minutes instead of five

drivers/hmis/app/jobs/hmis/ce/process_clients_job.rb
```diff
@@ -7,19 +7,19 @@
...
 #
 # See drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md
 #
 module Hmis::Ce
-  class ProcessChangesJob < BaseJob
+  class ProcessClientsJob < BaseJob
     include NotifierConfig
 
-    queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
+    queue_as :default
```
eanders:
```suggestion
    queue_as ENV.fetch('DJ_SHORT_QUEUE_NAME', :short_running)
```

ttoomey: thanks!

drivers/hmis/app/models/hmis/ce/README_FOR_CHANGE_MARKER.md
```diff
@@ -6,12 +6,13 @@ The change tracking system for Coordinated Entry (CE) efficiently processes clie
...
+This is achieved through a versioning system managed by the `Hmis::Ce::ChangeMarker` model and two specialized self-scheduling background jobs that continuously process these changes in parallel for optimal performance.
 
 ### Core Components
 
 - **`Hmis::Ce::ChangeMarker`**: A polymorphic model that tracks the state of other records (currently `GrdaWarehouse::Hud::Client` and `Hmis::Ce::Match::CandidatePool`). It uses `current_version` and `processed_version` to determine if a record is "dirty."
-- **`Hmis::Ce::ProcessChangesJob`**: A self-scheduling job that runs continuously to process dirty records. It uses an advisory lock to prevent concurrent runs.
+- **`Hmis::Ce::ProcessPoolsJob`**: A self-scheduling job that processes dirty candidate pools on the long-running queue. It uses per-pool advisory locks to coordinate with the client processor.
+- **`Hmis::Ce::ProcessClientsJob`**: A self-scheduling job that processes dirty clients on the default queue for fast updates. It uses non-blocking per-pool locks to coordinate with the pool processor.
```
gigxz:
```suggestion
- **`Hmis::Ce::ProcessPoolsJob`**: A self-scheduling job that processes dirty candidate pools on the long-running queue. It uses per-pool advisory locks to coordinate with the client processor.
- **`Hmis::Ce::ProcessClientsJob`**: A self-scheduling job that processes dirty clients on the short-running queue for fast updates. It uses non-blocking per-pool locks to coordinate with the pool processor.
```
Clarify that ProcessClientsJob runs on the short running queue?

lib/tasks/grda_warehouse.rake
gigxz: Agreed, I think `ProcessClientsJob` could run every minute or two. Maybe we can start with 2 minutes?

For `ProcessPoolsJob`, I think 5 minutes is also too frequent, but I think it will also depends on the changes from https://github.com/greenriver/hmis-warehouse/pull/5638. Without having gone through very closely yet, I _think_ that after that change, Candidate Pools would only ever be marked as dirty when Match Rules change (infrequent dev action) or when a new Unit Group is created (infrequent user action). I don't think Candidate Pools would be marked as dirty when Opportunities are created (frequent action), given that those opportunities necessarily belong to a candidate pool that already exists (I think). But I'm definitely blurring the lines with the other PR here. Maybe we can go ahead and merge this with `10.minutes`, then we can discuss on the other PR once conflicts are resolved. (And apologies for my review being slow- I hope the merge conflicts aren't too much of a pain!).

gigxz (approved):
Looks great! 💯 
Tested locally as well with incremental changes in app, confirm waitlists are updated much quicker.

lib/tasks/grda_warehouse.rake
ttoomey: > Agreed, I think ProcessClientsJob could run every minute or two. Maybe we can start with 2 minutes?

Sounds good, I'll adjust the timing. 

>  I think that after that change, Candidate Pools would only ever be marked as dirty when Match Rules change... I don't think Candidate Pools would be marked as dirty when Opportunities are created 

Yes, that should be the case. The polling behavior should be pretty lightweight if there's no dirty pools. Initially it's just us managing the rules so I guess it's whatever seems best. We might also want to make the timeout dynamic, so that if the builder exits before completing all the pools then it uses a shorter cadence. 

> And apologies for my review being slow- I hope the merge conflicts aren't too much of a pain

No worries! I've created a small mess with these different directions but I'm hopeful it won't be too hard to reconcile. We'll see how it looks after the dust settles

PR #5634, opened 2025-08-07
HMIS CSV Custom Exports

eanders:
## _Merging this PR_

## Description
* Introduces Custom file exports to the standard HMIS CSV
* Re-uses configuration for Custom Imports to determine which custom exports are available
* Adds static file generation for custom export classes to `HmisCsvTwentyTwentySix::CustomFileManager`
* Adds the ability to export custom data with lookups in related tables
* Enables FY2026 importer and exporter for QA installation (all other staging installations are held back unitl 9/1/2025)

## Type of change
* New feature


dtgreiner (approved):

app/views/warehouse_reports/hmis_exports/_shared_filter.haml
```diff
@@ -51,6 +52,15 @@
     - if can_export_anonymous_hmis_data?
       .col-sm-4
         .pt-no-label= f.input :faked_pii, as: :pretty_boolean, label: 'Should the PII in this export be faked?', hint: 'This can be used to export data useful for developers.'
+  .row
```
ttoomey: should this be conditionally rendered based on `include_custom_files` local var? We're passing that in but not using it:
`= render 'warehouse_reports/hmis_exports/shared_filter', f: f, include_hash_status: false, include_custom_files: false`

app/javascript/controllers/custom_file_exports_controller.js
```diff
@@ -0,0 +1,61 @@
+import { Controller } from "@hotwired/stimulus"
```
ttoomey, line 26: docs?
```suggestion
/*
 * CustomFileExportsController manages the dynamic visibility of the "Custom Files"
 * section on the HMIS CSV Exports page.
 *
 * It listens for changes to the HMIS version selector and shows or hides the
 * custom files section if any custom files are available for the selected version.
 *
 * It expects the version selector element to have a `data-custom-files` attribute
 * containing a JSON object that maps HMIS versions to an object of available custom files.
 * The keys of the inner object are the display names, and the values are the filenames.
 *
 * Example of `data-custom-files`:
 * {
 *   "2026": { "Custom Genders (CustomGender.csv)": "CustomGender.csv" },
 *   "2024": {}
 * }
 *
 * It also populates a multi-select input with the available files for the selected
 * version, preserving any selections that are still valid for the new version.
 *
 * Targets:
 * - `version`: The <select> element for the HMIS version.
 * - `section`: The <div> element containing the custom files UI.
 * - `select`: The <select> element for choosing custom files.
 */
import { Controller } from "@hotwired/stimulus"
```

app/models/filters/hmis_export.rb
```diff
@@ -282,5 +285,20 @@ def chosen(key)
...
       end
     end
+
+    def available_custom_file_types
+      return {} unless version == '2026'
+
+      begin
+        HmisCsvTwentyTwentySix.custom_files_config.definitions.map(&:for_select).to_h
+      rescue StandardError => e
+        Rails.logger.warn "Failed to load available custom file types: #{e.message}"
```
ttoomey: is it correct to swallow the error here?

app/models/grda_warehouse/hud/custom_data_element.rb
```diff
@@ -18,6 +18,8 @@ class CustomDataElement < Base
     self.table_name = 'CustomDataElements'
     self.sequence_name = "public.\"#{table_name}_id_seq\""
 
+    belongs_to :custom_data_element_definition, **hud_assoc(:CustomDataElementDefinitionID, 'CustomDataElementDefinition'), optional: true
```
ttoomey, line 24: It seems like there's some potential for confusion as we have redundant associations between data element and definition. This seems unfortunate but I guess it's not practical to re-use the HMIS relationship.
I'd suggest we add some comments to this model and the HMIS model to call this out and reduce the risk that engineers use the wrong relationship. I imagine it would be easy accidentally join on the wrong field

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/exporter/custom/base.rb
```diff
@@ -0,0 +1,256 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module HmisCsvTwentyTwentySix::Exporter::Custom
+  class Base
```
ttoomey, line 24: docs?
```suggestion
  # Base class for all custom file exporters in the FY2026 specification.
  #
  # This class provides a common foundation for dynamically generated exporter
  # models that are created by `HmisCsvTwentyTwentySix::Custom::FileManager`. It
  # integrates with the YAML configuration defined for custom files to provide
  # shared functionality like generating CSV headers, mapping warehouse data to
  # export formats, and handling dynamic lookups via ActiveRecord associations.
  #
  # Subclasses are generated automatically and are expected to implement key
  # methods based on their YAML configuration, such as:
  # - `self.custom_file_name`: The name of the output CSV file.
  # - `self.export_scope`: The ActiveRecord scope for fetching data.
  # - `self.owner_class_mapping`: (Optional) A map for owner-based filtering.
  # - `self.warehouse_class_for_export`: (Optional) The primary warehouse model.
  #
  class Base
```

ttoomey (approved):
Looks great! Let some suggestions

app/views/warehouse_reports/hmis_exports/_shared_filter.haml
eanders: Yes, thank you.

app/models/filters/hmis_export.rb
eanders: I _think_ so.  My thinking was that we didn't want exports to fail if we were missing some yaml files, but it seems unlikely we'd hit this.  I'll remove the rescue for now.

PR #5644, opened 2025-08-13
Store HUD Report data as CSV files in ActiveStorage

dtgreiner:
## _Merging this PR_

## Description

Update the HUD reports to run a cleanup job after completion that stores the report data in ActiveStorage as csv files and deletes the RDS data. These csv files are then pulled down to displaying the report results. 

## Type of change
New feature


app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Concern for handling client loading in cells controllers when artifacts are stored in S3
+module HudReports
+  module ArtifactAwareCells
+    require_relative 'temporary_cell_row'
+    require_relative 'temporary_cell_collection'
```
eanders: This require setup has a bit of a code smell to me.  Can we just move those modules to the `app/hud_reports/` directory so the class loader can find them naturally?

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+        load_clients_from_s3(client_class, cell, table)
+      else
+        load_clients_from_rds(client_class, cell, table)
+      end
+    end
+
+    # Load clients using legacy database joins.
+    # This will only return clients for reports whose data HAS NOT been stored in S3.
+    # When a report is stored in S3, the data in these tables are cleared out.
+    def load_clients_from_rds(client_class, cell, table)
```
eanders: can we rename this?
```suggestion
    def load_clients_from_db(client_class, cell, table)
```

RDS is an AWS thing, but presumably, this could be any database.

eanders: Also, is this a method we expect not to need long-term, or is there an expectation that some cells will always have DB backing?

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      client_class.
+        joins(hud_reports_universe_members: { report_cell: :report_instance }).
+        merge(::HudReports::ReportCell.for_table(table).for_cell(cell)).
+        merge(::HudReports::ReportInstance.where(id: @report.id))
+    end
+
+    # Load clients using S3 data.
+    # This will only return clients for reports whose data HAS been stored in S3.
+    def load_clients_from_s3(client_class, cell, table)
```
eanders: Let's rename this method too:
```suggestion
    def load_clients_from_file(client_class, cell, table)
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      report_cell = if cell.is_a?(String)
+        @report.report_cells.find_by(cell_name: cell, question: table)
+      else
+        cell.question == table ? cell : @report.report_cells.find_by(cell_name: cell.cell_name, question: table)
+      end
+
+      return client_class.none unless report_cell
+
+      # Pull the csv file for the universe members from the universe_members_csv file for this question
+      # NOte: These files are sharded by report question/table
```
eanders:
```suggestion
      # NOTE: These files are sharded by report question/table
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      service = HudReports::S3ArtifactService.new(@report)
```
eanders:
```suggestion
      service = HudReports::FileArtifactService.new(@report)
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      cell_members = universe_csv.select { |row| row['report_cell_id'].to_i == report_cell.id }
+      member_ids = cell_members.map { |row| row['universe_membership_id'].to_i }.compact.uniq
+
+      # Load the actual report client records from the report_clients_csv file
+      # This file contains the client_id, source_enrollment_id, data_source_id, project_id, organization_id, and user_id
+      # for each client in the report.
+      clients_csv = service.retrieve_report_clients
+      return client_class.none unless clients_csv
+
+      # Filter the client datafor the specific universe members and create temporary objects to build out the cell data
```
eanders:
```suggestion
      # Filter the client data for the specific universe members and create temporary objects to build out the cell data
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: These should already be in the same directory and we shouldn't need these lines at all. I'll clear them out and verify in the next commit.

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      matching_clients = clients_csv.select { |row| member_ids.include?(row['id'].to_i) }
+      return create_temporary_cell_objects(matching_clients)
+    end
+
+    def create_temporary_cell_objects(csv_rows)
+      # Create temporary objects from CSV data and attach preloaded warehouse clients and associations
+      temporary_objects = []
+
+      # Build out data that will be used in the cell data. These items are being preloaded as are being
```
eanders:
```suggestion
      # Build out data that will be used in the cell data. These items are being preloaded and are being
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+      # used in calculations within a HUD report when displaying the cell data (e.g. PII calculations or
+      # pulling names from these objects for a more user friendly display)
+      client_ids = csv_rows.map { |r| r['client_id']&.to_i }.compact.uniq
+      source_enrollment_ids = csv_rows.map { |r| r['source_enrollment_id']&.to_i }.compact.uniq
+      data_source_ids = csv_rows.map { |r| r['data_source_id']&.to_i }.compact.uniq
+      project_ids = csv_rows.map { |r| r['project_id']&.to_i }.compact.uniq
+      organization_ids = csv_rows.map { |r| r['organization_id']&.to_i }.compact.uniq
+      user_ids = csv_rows.map { |r| r['user_id']&.to_i }.compact.uniq
```
eanders: This is looping over the CSV rows 6 times.  I think it'll be more efficient to do something like:

```suggestion
      client_ids = Set.new
      source_enrollment_ids = Set.new
      data_source_ids = Set.new
      project_ids = Set.new
      organization_ids = Set.new
      user_ids = Set.new
      
      csv_rows.each do |r| 
        client_ids << r['client_id']&.to_i 
        source_enrollment_ids << r['source_enrollment_id']&.to_i
        data_source_ids << r['data_source_id']&.to_i
        project_ids << r['project_id']&.to_i
        organization_ids << r['organization_id']&.to_i
        user_ids << r['user_id']&.to_i
      end
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: I'll update the name, and we shouldn't need this long term. This was in there for backwards compatibility with old reports and could be removed once we process all previously generated hud reports to be stored as CSV files. It felt safer to leave this here as a fallback.

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
```diff
@@ -0,0 +1,135 @@
...
+
+      actual_clients = GrdaWarehouse::Hud::Client.where(id: client_ids).
+        preload(:data_source, :source_clients).
+        index_by(&:id)
+
+      enrollments_by_id = if source_enrollment_ids.any?
+        GrdaWarehouse::Hud::Enrollment.where(id: source_enrollment_ids).preload(:client).index_by(&:id)
+      else
+        {}
+      end
```
eanders: These can all be golfed down:

```suggestion
      enrollments_by_id = {}
      enrollments_by_id = GrdaWarehouse::Hud::Enrollment.where(id: source_enrollment_ids).preload(:client).index_by(&:id) if source_enrollment_ids.any?
```

app/controllers/concerns/hud_reports/temporary_cell_row.rb
```diff
@@ -0,0 +1,78 @@
...
+    end
+
+    def display_value(key, **_options)
+      key_str = key.to_s
+      csv_row[key_str] ||
+        case key_str
+        when /^.*enrollment\.first_name$/, 'first_name', 'FirstName'
+          source_client&.first_name || client&.first_name || ''
+        when /^.*enrollment\.last_name$/, 'last_name', 'LastName'
+          source_client&.last_name || client&.last_name || ''
```
eanders, line 53: These should probably obey the PII policy right?

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,180 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Background job to store HUD report artifacts in S3 and immediately cleanup RDS data
+# This job runs after a report is completed to archive intermediate data and free up RDS space
+module HudReports
+  class StoreArtifactsAndCleanupJob < BaseJob
```
eanders, line 12: I think this should be integrated into the report run, not a separate job (fine to encapsulate the logic here, but we should be calling `perform_now` or similar since the report is already running in a background job.)

eanders: Actually, I guess I'd rather see this as a PORO that just takes the report instance.

app/controllers/concerns/hud_reports/temporary_cell_row.rb
dtgreiner: The PII policies are being called on the data after these are set (whether from the file or the db), I can run some tests to verify but the process hiding PII through the policies should not have changed in this PR

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,180 @@
...
+
+      if report_version
+        # Find the generator class from the config using the report name
+        generator_class = generator_class(report.report_name, report_version)
+
+        if generator_class
+          # Get the client classes for this specific generator
+          client_classes = client_classes_for_generator(generator_class)
+
+          client_classes.each do |table_class|
```
eanders: If this is running as part of the final step of the report, I think we could have the generator pass in the client classes (or client scopes)

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,180 @@
...
+            if table_class.column_names.include?('report_instance_id')
+              deleted_count = table_class.where(report_instance_id: report.id).delete_all
```
eanders: This feels brittle, let's force the caller to tell the cleaner-up-er-er what records to clean.

app/models/hud_reports/report_cell.rb
```diff
@@ -53,11 +55,19 @@ def user
...
+      end
     end
 
     def count
-      @count ||= universe_members.count
+      @count ||= if report_instance.artifacts_stored?
+        count_from_s3
+      else
+        universe_members.count
+      end
```
eanders: I'm probably missing something, but I think this could just be:

```suggestion
      @count ||= if report_instance.artifacts_stored?
        members.count
      else
        universe_members.count
      end
```

Since both methods are loading the data from S3.  That way it would only have to be loaded once, and then it would just pull from memory on the next request.

app/models/hud_reports/report_instance.rb
```diff
@@ -27,6 +27,16 @@ class ReportInstance < GrdaWarehouseBase
...
+
+    # ActiveStorage attachments for S3-based report artifacts
+    has_many_attached :universe_members_csv_shards
+    has_one_attached :report_clients_csv
+    has_one_attached :cell_details_csv
+    has_one_attached :report_summary_json
+
+    # Callbacks for artifact management
+    after_update :schedule_artifact_storage_and_cleanup, if: :saved_change_to_completed_at?
+
```
eanders: I think we should just move this into the `run_and_save!` method.

app/services/hud_reports/report_storage_service.rb
```diff
@@ -0,0 +1,58 @@
...
+    end
+
+    # Process a specific report
+    def process_report(report_id)
+      report = HudReports::ReportInstance.find(report_id)
+
+      Rails.logger.info "Processing report #{report_id} for S3 storage and RDS cleanup"
+
+      HudReports::StoreArtifactsAndCleanupJob.perform_now(report.id)
+      Rails.logger.info "Successfully processed report #{report_id}"
```
eanders: You're loading the report here, but then only using the ID.  (see previous comments) but I think this should pass in the report instance, and probably some other related data that we know we'll need.

app/services/hud_reports/s3_artifact_service.rb
```diff
@@ -0,0 +1,293 @@
...
+      store_universe_members_csv_shards!
+      store_report_clients_csv!
+      store_cell_details_csv!
+      store_report_summary_json!
+    end
+
+    # Store universe members as per-question shards
+    def store_universe_members_csv_shards!
+      # Avoid duplicating work if shards already exist
+      return if @report.universe_members_csv_shards.attached? && @report.universe_members_csv_shards.attachments.any?
```
eanders: We should probably have a helper method to clear and re-create the attachments, that way while we're building these reports, if we need to add new tables, there's an easy way to start over and we won't end up in a situation where half of the shards are stored.

app/services/hud_reports/s3_artifact_service.rb
```diff
@@ -0,0 +1,293 @@
...
+
+      questions = @report.report_cells.distinct.pluck(:question)
+      questions.each do |question|
+        csv_data = generate_universe_members_csv_for_question(question)
+        filename = "universe_members_#{@report.id}_#{question}_#{Time.current.to_fs(:number)}.csv"
```
eanders: These filenames should probably include the report class or table name.  Once we start adding non-HUD reports, there might be collisions with the ids.  Maybe:

```suggestion
        filename = "universe_members_#{@report.class.table_name}_#{@report.id}_#{question}_#{Time.current.to_fs(:number)}.csv"
```

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: name update included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/services/hud_reports/s3_artifact_service.rb
eanders: Oh!  Just hadn't got there.
`cleanup_artifacts!`

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/controllers/concerns/hud_reports/artifact_aware_cells.rb
dtgreiner: Included in [this commit](https://github.com/greenriver/hmis-warehouse/pull/5644/commits/189c4cdec097138b0dbd1723976918957ed2e9d8)

app/services/hud_reports/file_artifact_service.rb
```diff
@@ -0,0 +1,293 @@
...
+    # Clean up S3 artifacts (for testing or manual cleanup)
+    def cleanup_artifacts!
+      @report.report_clients_csv.purge if @report.report_clients_csv.attached?
+      @report.cell_details_csv.purge if @report.cell_details_csv.attached?
+      @report.report_summary_json.purge if @report.report_summary_json.attached?
+    end
+
+    private
+
+    def report_storage_cache_key(kind:, blob:, question: nil)
```
eanders, line 158: I think this needs to include `@report.class.name` to be safe.

app/services/hud_reports/file_artifact_service.rb
```diff
@@ -0,0 +1,293 @@
...
+
+    def generate_report_clients_csv
+      classes = report_client_classes
+      return '' if classes.blank?
+
+      # Collect headers
+      excluded = ['id', 'created_at', 'updated_at', 'deleted_at']
+      headers = classes.flat_map(&:column_names).uniq - excluded
+      # Ensure common keys present even if they’re methods
+      (headers << 'id' << 'client_id' << 'project_id' << 'data_source_id').uniq!
```
eanders:
```suggestion
      headers += [ 'id', 'client_id', 'project_id', 'data_source_id']
      headers.uniq!
```

app/models/hud_reports/report_instance.rb
```diff
@@ -27,6 +27,16 @@ class ReportInstance < GrdaWarehouseBase
     has_many :universe_cells, -> do
       universe
     end, class_name: 'ReportCell'
+
+    # ActiveStorage attachments for S3-based report artifacts
```
eanders, line 31: Let's add some convenience methods:
1. something like `associated_report_generator` which looks at the `report_name` column, does some magic (case statement for now?) and returns a generator instance

```
def populate_associated_report_generator(report)
  generator_class = case report.report_name
    ...
  end

 report.update(generator_class_name: generator_class.name)
end
```
Maybe we should add a `generator_class_name` column to `hud_report_instances`.  We could use the above to populate that.  We'd also need to update the report runner job to populate the generator class name for future runs.

After we populate `generator_class_name`, we can update `associated_report_generator` to just:

```
def associated_report_generator(report)
  generator_class = report.generator_class_name.constantize
  generator_class.new(report)
end
```

Note: we determined universe membership is a polymorphic relationship so you can get the right set of clients through that (or enrollments, or episodes or whatever.)

eanders (commented):
Plan:
1. Streamline report generator instance access
2. Add some convenience methods to the generator class to standardize access to associations needed for saving to s3 
3. Refactor job to take a report instance, this should allow easy access to the generator and "clients scopes" 
4. Take the time to define the API (does each report have a `generator_instance` method, and an `associations_for_export` method?  These methods should return a consistent shape (maybe an array of scopes) 
5. Background reports should all directly call the "archive to file" method at the end of their "run and save" routine (as perform_now).  The job could also be invoked via the rake task to migrate old reports.
6. Add a TaskQueue thingy to run the rake task.
7. Should we add another task to delete deleted records from these "client" tables or should that just be a `really_destroy!` as soon as it has moved to S3.

dtgreiner:
Looks like a bunch of HUD tests are breaking after moving the data to active storage. I'll add updates for those next.

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,143 @@
...
+    def store_artifacts_to_s3(report)
+      Rails.logger.info "Storing artifacts to S3 for report #{report.id}"
+
+      service = HudReports::FileArtifactService.new(report)
+      service.store_artifacts!
+
+      Rails.logger.info "Successfully stored artifacts to S3 for report #{report.id}"
+    end
+
+    def cleanup_rds_data(report)
```
eanders, line 45: I think these should all be `db` not `rds`, so `cleanup_db_data` in this case.

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,143 @@
...
+
+      Rails.logger.info "Successfully cleaned up RDS data for report #{report.id}"
+    end
+
+    def cleanup_universe_members(report)
+      cell_ids = report.report_cells.pluck(:id)
+
+      return unless cell_ids.any?
+
+      deleted_count = HudReports::UniverseMember.where(report_cell_id: cell_ids).delete_all
```
eanders, line 62: This will need to be updated for other "simple" backgrounds reports.  I guess I'd rather see this whole cleanup process delegated to the report, so there could be a consistent API:

`report.expunge_intermediate_data!`

or something like that.  For HUD reports this could live in the base class and be overridden as necessary.

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,143 @@
...
+      Rails.logger.info "Deleted #{deleted_count} universe members for report #{report.id}"
+    end
+
+    def cleanup_report_client_tables(report)
+      # We need to leave SpmEnrollments in the database. They are used to build other universe members.
```
eanders, line 67: Along the same lines, this feels brittle and the variation should be delegated to the report that already understands the relationships.

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,143 @@
...
+    def update_report_status(report)
+      # Update timestamp to track the process
+      report.update_columns(
+        artifacts_stored_at: Time.current,
+      )
+
+      Rails.logger.info "Updated status for report #{report.id}"
+    end
+
+    def handle_special_cleanup(table_class, report)
```
eanders, line 93: If this is all delegated to the report, we shouldn't need this big case statement.

app/controllers/concerns/hud_reports/temporary_cell_collection.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    def group(*attributes)
+      # Store the grouping attributes for use in count
+      @group_attributes = attributes
+      self
+    end
+
+    private
+
+    def evaluate_arel_condition(condition, obj)
```
eanders, line 144: Why do we need this?
This feels really brittle and unexpected.  If there are specific queries we make, maybe we can override those to look for the stored data first?

app/models/hud_reports/report_cell.rb
```diff
@@ -183,5 +201,114 @@ def write_detail(path:, generator:, question_name:)
...
+      value.match?(/\A\d{4}-\d{1,2}-\d{1,2}/) || value.match?(/\A\d{1,2}\/\d{1,2}\/\d{4}/)
+    end
+
+    private def integer?(value)
+      value.match?(/\A\d+\z/)
+    end
+
+    private def json?(value)
+      value.start_with?('[') || value.start_with?('{')
+    end
```
eanders, line 215: There must be a built-in way to determine type.

app/jobs/hud_reports/store_artifacts_and_cleanup_job.rb
```diff
@@ -0,0 +1,143 @@
...
+        store_artifacts_to_s3(report)
+
+        # Step 2: Update report status (mark as artifacts stored)
+        update_report_status(report)
+
+        # Step 3: Clean up RDS data
+        cleanup_rds_data(report)
+
+        Rails.logger.info "Successfully processed report #{report_id}: stored to S3 and cleaned RDS"
+      end
```
eanders, line 31: Did you intend to add a rescue here?
Also, it looks like if the artifacts fail to get stored, it'll still make them as stored, and cleanup the RDS.

`store_artifacts_to_s3` should probably return a true or false, and if false, it should raise and fail catastrophically.

Additionally, we should probably have some sort of wrapper around `cleanup_rds_data` so that if it fails we know we need to investigate and try again.  Maybe it could happen in a transaction?  Not sure if that's possible given the size of the changes...

PR #5656, opened 2025-08-16
better handling for delayed job with dynamic scaling/deployments

ttoomey:
## Description

Hypothesis: delayed job by default does not stop the current job when it receives SIGTERM. When containers are terminated (during deploy or scaling), k8 sends term, gives the container 30 seconds (or something) to exit, then sends SIGKILL. When delayed job receives sigkill, it does not unlock the job, resulting in stale records that remain locked.

We can change this behavior using dj config option `raise_signal_exceptions` - however when this is enabled, dj's retry logic triggers, so max_attempts should be increased to compensate.

## Type of change

### DIFF
--- a/config/initializers/delayed_job.rb
+++ b/config/initializers/delayed_job.rb
@@ -1,7 +1,12 @@
 Delayed::Worker.destroy_failed_jobs = false
 Delayed::Worker.sleep_delay = 5
-Delayed::Worker.max_attempts = 3
+Delayed::Worker.max_attempts = 6
 Delayed::Worker.max_run_time = 30.hours
+
+# Handle shutdown signals
+# This will cause the worker to catch SIGTERM and raise an error, allowing the job to be requeued
+# Note, this means deployments and auto-scaling will now consume a job's retry attempts
+Delayed::Worker.raise_signal_exceptions = :term
 
 Delayed::Worker.default_queue_name = 'default_priority'
 Delayed::Worker.read_ahead = 2


ttoomey:
@eanders - this is an idea I'm considering; wondering what you think. I'm not sure it's exactly right as-is:
* this would apply to both short and lon-term queues. It seems like the current behavior (wait for the current job to finish) is best for the short runner. But the stopping the long-term runner seems more correct
* perhaps we could achieve this by hand-rolling code to trap SIGTERM only for long-running jobs

eanders:
@ttoomey I'm confused about the description.  The scenario we were seeing was that the job was marked as locked **and** had a `failed_at` value, **but** was actually still running, and ran to completion over many minutes.

I'm very hesitant to increase the max number of attempts, especially with out also adjusting the importer and LSA, which currently are setup to only run once since they don't recover very well, and sometimes take hours.

ttoomey:
> I'm confused about the description. The scenario we were seeing was that the job was marked as locked and had a failed_at value, but was actually still running, and ran to completion over many minutes.

Sorry for the confusion @eanders . There's two parts of 8098. This was something I discovered while investigating. Regarding the behavior we saw, see the [comment](https://github.com/open-path/Green-River/issues/8098#issuecomment-3193772998) and devops issue linked in 8098 (tldr, I think there's a dj cleanup script that is marking the jobs as failed outside of the application)

> I'm very hesitant to increase the max number of attempts, especially with out also adjusting the importer and LSA, which currently are setup to only run once since they don't recover very well, and sometimes take hours.

Yes, I think that we probably don't want exactly this behavior, or at least we don't want it everywhere. However as-is there's the potential for jobs to get hard-killed with no cleanup (if my hypothesis is correct, I haven't tested it).

I guess my question with this PR is if it's worth additional time. If we wanted to pursue this further, we'd probably use per-job mix-in rather than the global setting provided by DJ.

Maybe something like this:
```ruby
module ExitOnTermSignal
  def self.execute
    original_term_handler = Signal.trap('TERM') do
      begin
        Delayed::Job.where("locked_by LIKE ?", "%pid:#{Process.pid}%")
                    .update_all(locked_at: nil, locked_by: nil)
      rescue => e
        Rails.logger&.warn("[ExitOnTermSignal] unlock failed for pid=#{Process.pid}: #{e.class}: #{e.message}")
      ensure
        Process.exit!(0)
      end
    end

    begin
      yield
    ensure
      Signal.trap('TERM', original_term_handler)
    end
  end
end
```

eanders:
Thanks @ttoomey, let's discuss, maybe with some others early next week.  I see the issues now, not sure of a resolution yet.

ttoomey:
closing in favor of [5799](https://github.com/greenriver/hmis-warehouse/pull/5799)

PR #5657, opened 2025-08-18
merge r177 into 7957 

ttoomey:
## **DO NOT SQUASH MERGE**
* resolves merge conflicts with release-177, in particular 8044
* also adds data source support for ce rules- I noticed it's part of 7957 so I modified MatchApplicability to support it. Not sure if we actually need data-source rules though


drivers/hmis/app/models/hmis/ce/match/unit_group_rule_resolver.rb
```diff
@@ -0,0 +1,71 @@
...
+    def compose_priority_expression(rules)
+      rules = rules.filter(&:priority_scheme?).sort_by { |r| [r.rank, r.id] }
+      return if rules.empty?
+
+      # Multiple ranked priority rules may apply to a unit group.
+      # These priority rules must all share the same owner type because `rank` is only unique
+      # with respect to owner; mixing owner types is ambiguous. Therefore if there are rules
+      # with different owners, pick the filter the rules to only the most specific owner
+      most_specific_level = rules.map(&:owner_precedence).min
+      rules = rules.filter { |rule| rule.owner_precedence == most_specific_level }
```
gigxz: This makes sense, but I don't think it aligns with what we resolve on the Unit/UnitGroup graphql types? Those back the Unit/Unit Group pages, which display the prioritization rule descriptions. (Now that rule's can no longer be tied to Units, we can also rethink this api choice since we don't _need_ to resolve rules on Unit anymore). 

I thinks the Unit/UnitGroup gql type is resolving ALL priority_scheme rules that apply to the Unit/UnitGroup, which isn't accurate as far as the rules that will actually be used to prioritize the candidate pool. Am I understanding that right?

gigxz (approved):

drivers/hmis/app/models/hmis/ce/match/match_applicability.rb
```diff
@@ -17,7 +21,7 @@
 # Note:
 # - This class does NOT determine client eligibility itself; it determines which rules
 #   should be used for eligibility/priority.
-# - It supports evaluation against various entities (eg Project, Unit, etc) to support viewing applicable
+# - It supports evaluation against various entities (eg Project, Unit Group, etc) to support viewing applicable
 #   rules for each entity type.
 # - Global and Data Source-level rules are not yet supported (e.g. "all PSH projects in a data source")
```
gigxz:
```suggestion
```

Thanks for adding DS we do need it, we'll want to support rules that apply to all project types in the data source

drivers/hmis/app/models/hmis/ce/match/unit_group_rule_resolver.rb
ttoomey: Thanks! Per discussion I'll address this issue in this PR

ttoomey:
Hi @martha - is it okay to merge this in?

martha:
Thanks, yes, sorry for the delay. I'll merge it now

ttoomey:
> Thanks, yes, sorry for the delay. I'll merge it now

Thanks, no worries!

PR #5660, opened 2025-08-18
Cast float to string when saving form field

martha:
## _Merging this PR_

## Description

This form snippet is saving the result of a calculation, but it's saved on the CDED as a string. So at some point, it needs to be cast to a string. I'm suggesting to cast it here, on the backend when receiving the value, as opposed to the frontend, because on the frontend within the form logic we might want to compare the value to another and show/hide some form logic based on that. (We are not doing this currently, it is just an example.) Let me know what you think about this!

Form snippet to test with:
```
{
  "text": "AMI",
  "type": "STRING",
  "link_id": "ami",
  "read_only": true,
  "autofill_values": [
    {
      "formula": "ROUNDTO((monthly_household_income / ami_constant) * 100, 2)",
      "autofill_when": [
        {
          "operator": "EXISTS",
          "question": "monthly_household_income",
          "answer_boolean": true
        },
        {
          "operator": "EXISTS",
          "question": "ami_constant",
          "answer_boolean": true
        }
      ],
      "autofill_behavior": "ALL",
      "autofill_readonly": true
    }
  ],
  "disabled_display": "HIDDEN"
},
```
## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/processors/base.rb
+++ b/drivers/hmis/app/models/hmis/hud/processors/base.rb
@@ -268,7 +268,7 @@ class Hmis::Hud::Processors::Base
       case value
       when String
         value&.presence&.strip
-      when Integer
+      when Integer, Float
         value.to_s
       else
         raise "unexpected value \"#{value}\""


ttoomey (approved):
seems reasonable to me, thanks!

PR #5662, opened 2025-08-20
Implement workflow changes

martha:
## _Merging this PR_

## Description
- Make changes described in ticket description
- ~Add support for a task side effect that marks another task as disabled. See inline comment thread~
- Disable outstanding steps when a referral is accepted or rejected
- [x] **Outstanding spec question**, see ticket comments

How to test: Run `rails driver:hmis:ce_define_workflows` and step through the new referral workflow.

## Type of change
Feature build-out


drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -114,6 +126,14 @@ def set_custom_referral_status(status_key:)
       referral.update!(custom_status: status)
     end
 
+    def disable_step(node_id:)
```
martha: This is a new `disable_step` side effect. This lets one task specify that when this task has an event, some _other_ task should change status. The trigger config for this looks like:
```
trigger_config: [
  {
    event: 'complete_step',
    message: 'disable_step',
    params: { 'node_id': other_task.id },
  },
],
```

To discuss: I'm not totally sold on doing this by `node_id`, I guess because it makes the workflow definition more opaque? Another possibility is to add a new `key` column which could be a human readable string, sort of equivalent to form IDs vs. `definition_identifier`s. Let me know if you think this is worth pursuing.

drivers/hmis/lib/ce_workflows/ac/workflow_builder.rb
```diff
@@ -337,26 +388,36 @@ def build_provider_outcome_denial_review_loop(template:, ce_staff_swimlane:, pro
...
+      # Create Enrollment (Script) => Change Provider Outcome *optional* task that re-triggers the Denial Review
+      create_enrollment_task.connect_to!(change_provider_outcome_task)
+      # Note that this re-introduces the truly circular workflow.
+      # It is theoretically possible for users to re-open a previously closed Denial Review step, like this:
+      # - Complete the Change Provider Outcome step
+      # - In the Denial Review step, select "Send Back"
+      # - In the Provider Outcome (Second Attempt) step, select "Accept/Enroll"
+      # - Complete the Change Provider Outcome step a second time
+      # - The Denial Review (1) step is re-opened, instead of Denial Review (Second Attempt).
+      change_provider_outcome_task.connect_to!(denial_review_task)
```
martha: Let's discuss, does this seem acceptable?

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -75,13 +77,22 @@ def accept_referral
...
     def reject_referral
+      disable_outstanding_steps(referral)
       referral.completed_at = Time.current
       referral.reject!
       set_custom_referral_status(status_key: 'rejected')
       referral.opportunity.release!
       mark_client_dirty(referral.client)
     end
 
+    def disable_outstanding_steps(referral)
```
ttoomey: other methods use `self.referral` rather than passing it as an arg. Suggest removing this to keep the style pattern consistent
```suggestion
    def disable_outstanding_steps
```

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
ttoomey: Yes, let's discuss. Using a pk reference might be okay for users authoring workflows from an interface. It does make cloning a workflow or export/import between environments more complex. Although there is also complexity in adding and maintaining a unique key on node

drivers/hmis/lib/ce_workflows/ac/workflow_builder.rb
ttoomey: If we wanted to avoid cycles, perhaps we could add an additional user task? Maybe the cycles are fine though

martha:
Converting to draft while I work through feedback!

drivers/hmis/lib/form_data/allegheny/ce_referral_steps/change_provider_outcome.json
```diff
@@ -0,0 +1,61 @@
...
+        {
+          "code": "Estimated vacancy no longer available"
+        },
+        {
+          "code": "Enrolled, but declined HMIS data entry"
+        }
+      ]
+    },
+    {
+      "text": "The client has already been enrolled in the project.",
```
martha, line 55:
```suggestion
      "text": "The client has already been added to the project.",
```
@gigxz I'm guessing this change is also needed here? https://github.com/greenriver/hmis-warehouse/pull/5690/files

martha:
I'm reopening this as ready for review, pending more discussion on my comment on the [issue](https://github.com/open-path/Green-River/issues/8105#issuecomment-3225733835).

gigxz (approved):

PR #5663, opened 2025-08-21
Support configuring CE Event Type on Unit Group

martha:
## _Merging this PR_

## Description

Summary of changes:
- Add `ce_event_type` to Unit Group. 
  - Update `CreateUnitGroup` mutation with the new param
  - Add a new `UpdateUnitGroup` mutation
- Add `ce_referral_id` to Event.
- Update `ReferralCeEventManager`
  - When creating a CE Event, use the CE Event specified on the Unit Group, if one exists.
  - Find the CE Event using the direct relationship to referral, rather than guessing based on recency.

Related frontend PR: https://github.com/greenriver/hmis-frontend/pull/1236

## Type of change
New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see ticket


db/warehouse/migrate/20250821194338_add_referral_id_to_ce_event.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class AddReferralIdToCeEvent < ActiveRecord::Migration[7.1]
+  disable_ddl_transaction!
```
martha: Remembering [recent feedback](https://github.com/greenriver/hmis-warehouse/pull/5616#discussion_r2280993553) about avoiding concurrent indexes, I hesitated about this one. However, the `Events` table is already in use in production, and might be large, so I thought it's probably worth it here. Does that sound right?

gigxz: @martha good question!

Re: the uniqueness index, I can imagine a future case where we want to support a Referral workflow generating multiple CE Events. (Since CE Events are HUD-defined logic, we don't know how they will change over time). So, I think I would propose making the index non-unique. I'm not sure that affects the concurrency question though.

Tagging @eanders on this: 
* We're proposing to add a reference to the `Event` table for storing the CE Referral that generated it. Does this approach look ok? (An alternative is to store the event_id reference on the ce referrals table. Storing it on the Event table feels more appropriate wrt their relationship, it also allows a one-to-many relationship in the future if ever needed)
* Thoughts on best approach to add index to this table?

eanders: My experience with `algorithm: :concurrently` is that our deployment process doesn't handle it very well.  

What would you think of a join table instead of putting it directly on event?

Maybe 

```
create_table :housing_statuses do |t|
  t.references :ce_referral, index: true
  t.references :event, index: true
  t.timestamps
end
```

And I agree, I can see a world (that already exists in CAS) where a CE referral needs to generate multiple HUD CE Events.

drivers/hmis/app/graphql/mutations/update_unit_group.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      # If name is being changed, check that it's still unique in the project.
+      # (This is also validated on the model, but we want to return a more user-friendly error message)
+      if input.name.present? && input.name != unit_group.name
+        # rubocop:disable Style/IfUnlessModifier
+        if Hmis::UnitGroup.where(project_id: unit_group.project_id, name: input.name).exists?
+          errors.add :name, :invalid, message: 'must be unique in the project'
+        end
+        # rubocop:enable Style/IfUnlessModifier
+      end
```
gigxz:
```suggestion
```

Suggest moving this validation to the model instead,
https://github.com/greenriver/hmis-warehouse/blob/1c8d8d6456ff374bc860af6a20837e2ae88ac1d3/drivers/hmis/app/models/hmis/unit_group.rb#L33
can be updated to add a custom message: `validates :name, presence: true, uniqueness: { scope: :project_id, case_sensitive: false, message: 'must be unique in the project' }`

which has the same effect!

drivers/hmis/app/graphql/mutations/create_unit_group.rb
```diff
@@ -21,7 +28,6 @@ def resolve(input:)
       # unit_type = Hmis::UnitType.find_by(id: input.unit_type_id)
       # raise 'Invalid unit type' if input.unit_type_id.present? && !unit_type.present?
 
-      errors = HmisErrors::Errors.new
       # errors.add :count, :required unless input.count.present?
       # errors.add :count, :out_of_range, message: 'must be positive' if input.count&.negative?
       # errors.add :count, :out_of_range, message: 'must be non-zero' if input.count&.zero?
```
gigxz, line 28: Can also remove the `'must be unique in the project..` and check for `unit_group.valid?` instead, per other suggestion, since it's validated on the model

drivers/hmis/app/graphql/mutations/update_unit_group.rb
```diff
@@ -0,0 +1,62 @@
...
+
+      # Prevent changing or clearing workflow template if one is already set
+      # rubocop:disable Style/IfUnlessModifier
+      if unit_group.workflow_template_identifier.present? && input.workflow_template_identifier != unit_group.workflow_template_identifier
+        errors.add :workflow_template_identifier, :invalid, message: 'cannot be changed once set'
+      end
+      # rubocop:enable Style/IfUnlessModifier
```
gigxz: Suggest moving this validation to the Model instead, [here's an example](https://github.com/greenriver/hmis-warehouse/blob/1c8d8d6456ff374bc860af6a20837e2ae88ac1d3/drivers/hmis/app/models/hmis/ce/opportunity.rb#L130-L135)

db/warehouse/migrate/20250821194338_add_referral_id_to_ce_event.rb
martha: Decision from dev standup - just use synchronous index wrapped in safety assured

drivers/hmis/app/graphql/mutations/update_unit_group.rb
martha: Cool!

gigxz (approved):

PR #5664, opened 2025-08-21
8099 active client report performance

ttoomey:
## _Merging this PR_

## Description
* adds unit tests for active client report
* adjust count queries to be more performant on active client report
* adds documentation

## Type of change
Perf fix, Bug fix, tests


eanders (approved):
Seems like it's worth trying.  
In most places we should have a date range, which should make querying SHS _much_ more performant.

ttoomey:
@eanders I backed out the change to with_service_between, it seemed like too risky. The targeted change to active client report seems like enough to address the performance concern. Can you double check it, including the updated docs I added after our discussion?

app/models/active_client_report.rb
```diff
@@ -6,6 +6,28 @@
...
+#
+# ActiveClientReport
+# Determines which clients are "active" within a user-supplied date range for the
+# Client Details → Actives report.
+#
+# Definition of "active" (HUD Active Client Method 5, augmented for street outreach):
+# - The client has an enrollment that is open at any time between filter.start and filter.end; and
+# - The client has qualifying activity within the same window:
+#   - Emergency Shelter — Non-Bed-Night (ES-NBN): requires at least one service during the window.
+#     Enrollments without services are excluded.
```
eanders:
```suggestion
#     Enrollments in ES-NBN without services are excluded.
```

app/models/active_client_report.rb
```diff
@@ -6,6 +6,28 @@
...
+#   - Street Outreach (SO): requires CLS present and valid
+#   - For other entry/exit projects types, only require entry/exit in the reporting window
```
eanders:
```suggestion
#   - For other entry/exit projects types, only require an enrollment overlapping the reporting window
```

app/models/active_client_report.rb
```diff
@@ -6,6 +6,28 @@
...
+#
+# Inclusion and exclusion rules:
+# - Qualifying events/CLS must fall within BOTH the enrollment dates and the report window.
+# - ES-NBN enrollments with no services in-range are excluded.
+#
+# This logic is kept aligned with the "core-demographics" query behavior when "require services" is enabled.
```
eanders: I'm not clear what `"core-demographics" query behavior` means.  That's just one report we refer to often since we know it gets lots of use.  I would say something more like:
```suggestion
# This logic is kept aligned with behavior in other reports when "require services" is enabled.
```

eanders (approved):
Thanks!
Left some updates to the comments, but nothing substantial.

ttoomey:
we could push this to 177 to get it out this Thurs; I'm leaning towards that

PR #5669, opened 2025-08-21
7853-hmis-hud-2026-breaking-changes

ttoomey:
## _Merging this PR_

## Description
⚠️ These are breaking API and form schema changes intended for a coordinated, cut-over deployment.

**FY2026 HUD updates across HMIS GraphQL enums and forms**
- New CoC Builds funding source
- Deprecates ESG-CV and HOPWA-CV (breaking enum name changes)
- Revised family critical issues picklists (to `NoYesReasonsForMissingData`)
- Adds V10 mental health consultation fragment to intake (with rule gating)
- Adds `#4_21_sex` demographics fragment
- Adjusts Service form to collect Date Provided for SSVF financial assistance, with additional guidance
- Remove requirement to collect specific disability fields for some VA funders

## Type of change
New feature


drivers/hmis/app/graphql/schema.graphql
```diff
@@ -5293,9 +5298,9 @@ enum FundingSource {
   HUD_COC_YOUTH_HOMELESS_DEMONSTRATION_PROGRAM_YHDP
 
   """
-  (47) HUD: ESG - CV
+  (47) HUD: ESG - CV [Deprecated]
   """
-  HUD_ESG_CV
+  HUD_ESG_CV_DEPRECATED
```
ttoomey: adding _DEPRECATED is consistent with strategy from previous standards cahnge ( see HUD_RURAL_HOUSING_STABILITY_ASSISTANCE_PROGRAM_DEPRECATED)

gigxz: ok, this is fine because the backing value is the integer, so 47 will now resolve as `HUD_ESG_CV_DEPRECATED` instead of `HUD_ESG_CV`. I think exception may raise if user is actively submitting a funder with the deprecated enum value, but refreshing the page would resolve that, so I think its OK.

gigxz:
Converting to draft since this should not yet be merged into release. Deploying to staging 9/2

drivers/hmis/app/graphql/schema.graphql
gigxz: Downside of this decision is that we need to maintain a frontend branch and keep deploying that too https://github.com/greenriver/hmis-frontend/pull/1251
Deploying it for now but we may want to back this out and keep the old enum, just to reduce overhead with maintaining and deploying both the branches.


Edit: I undid the breaking enum change here https://github.com/greenriver/hmis-warehouse/pull/5669/commits/f3518dfb91b111a52934811afab84f074ff27113. I know the approach was consistent with a previous standards change, but I think we made a mistake in that approach. Keeping keys consistent reduces the likelihood of user impact and makes it so that this branch doesn't break the API so we dont need to worry about deploying the frontend together

gigxz (approved):
Approving this for squash-merge into release-182 🙌

PR #5550, opened 2025-07-07
7049 hotfix 2026 spec change from bc770c5

ttoomey:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch
- use a merge-commit or rebase strategy for PRs targeting the stable branch

## Description
Fix breaking change that slipped in from 2026 HUD spec updates

## Type of change
Bug fix

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code

PR #5670, opened 2025-08-22
Update ServiceProcessor to be compatible with both FY24 and FY26 SSVF

gigxz:
## _Merging this PR_

## Description
Follow-up to hotfix https://github.com/greenriver/hmis-warehouse/pull/5550

https://github.com/open-path/Green-River/issues/8121

This makes the ServiceProcessor behavior compatible with both FY24 and FY26 service forms. This removes the need to have any service processor code changes on a deploy branch.

How I tested:
- Create/Update SSVF service using FY24 form, ensure FA Start Date is propagated onto Date Provided
- Create/Update SSVF service using FY26 form, ensure FA Start Date and Date Provided are collected independently (using form from https://github.com/greenriver/hmis-warehouse/pull/5669)

## Type of change
new feature

### DIFF
--- a/drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
+++ b/drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
@@ -16,13 +16,6 @@ module Hmis::Hud::Processors
       service = @processor.service_factory
 
       attributes = case attribute_name
-      when 'fa_start_date'
-        # If FA Start Date is present, the Date Provided should match it
-        if attribute_value.present?
-          { attribute_name => attribute_value, 'date_provided' => attribute_value }
-        else
-          { attribute_name => attribute_value }
-        end
       when 'sub_type_provided'
         # Enum value is set up like "144:4:6" (record type : type provided : sub type provided)
         { attribute_name => attribute_value&.split(':')&.last }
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
```diff
@@ -43,5 +36,18 @@ def schema
...
+    # have been deployed.
+    #
+    # FY2024 service form collects FA Start Date for SSVF Financial Assistance, and does NOT collect Date Provided. (Needs processor to store FA Start Date in Date Provided field)
+    # FY2026 service form collects both Date Provided and FA Start Date independently for SSVF Financial Assistance
+    def post_process
+      service = @processor.send(factory_name)
+
+      # Overwrite the DateProvided field if FA Start Date was submitted and Date Provided was not
+      date_provided_submitted = @hud_values.key?('dateProvided') && @hud_values['dateProvided'] != HIDDEN_FIELD_VALUE
+      service.date_provided = service.fa_start_date if service.fa_start_date.present? && !date_provided_submitted
```
gigxz, line 50: Noting the oddness of needing to check `@hud_values` for the hidden field;  it would be nice if the processors had easier access to "cleaned" values. But given this is somewhat temporary, I think it's not worth a refactor.

drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
```diff
@@ -43,5 +36,18 @@ def schema
...
+    # This post-processing step provides functionality needed for the FY2024 service form. It can be removed after FY2026 changes
+    # have been deployed.
+    #
+    # FY2024 service form collects FA Start Date for SSVF Financial Assistance, and does NOT collect Date Provided. (Needs processor to store FA Start Date in Date Provided field)
+    # FY2026 service form collects both Date Provided and FA Start Date independently for SSVF Financial Assistance
+    def post_process
+      service = @processor.send(factory_name)
+
+      # Overwrite the DateProvided field if FA Start Date was submitted and Date Provided was not
+      date_provided_submitted = @hud_values.key?('dateProvided') && @hud_values['dateProvided'] != HIDDEN_FIELD_VALUE
```
ttoomey, line 49: if `dateProvided` was submitted but it's blank, should it still be treated as submitted? Maybe 
```suggestion
      date_provided_submitted = @hud_values['dateProvided'].present? && date_provided_value != HIDDEN_FIELD_VALUE
```

ttoomey (approved):
looks good, thanks!

drivers/hmis/app/models/hmis/hud/processors/service_processor.rb
gigxz: Good question. I shouldn't matter in practice, because Date Provided is always required so it will be kicked back to the user if it's left blank. (I did test that case on both new and old forms). I think I'll opt to leave it as-is, since I tested it this way, and I do think we should treat it as "submitted" if its present and not hidden, regardless of the value. Again, though, it shouldn't make a difference in practice

PR #5671, opened 2025-08-23
Capybara tests for the main warehouse app

eanders:
## _Merging this PR_

## Description
* Adds capybara system tests for rails functionality
* Adds basic authentication smoke tests
* Adds tests to confirm datepicker date reformatting

## Type of change
* New Feature

### DIFF
--- a/.github/workflows/warehouse_system_tests.yml
+++ b/.github/workflows/warehouse_system_tests.yml
@@ -159,7 +159,7 @@ jobs:
           CHROME_URL: http://chrome:3333
           TODO_OR_DIE_LOG_ONLY: true
         run:
-          bundle exec rspec -fd --fail-fast=$MAX_FAILURES --color --pattern "spec/system/**/ --tag type:rails_system
+          bundle exec rspec -fd --fail-fast=$MAX_FAILURES --color --pattern "spec/system/**/" --tag type:rails_system
 
       - name: Archive artifacts
         if: always()
--- a/.github/workflows/warehouse_system_tests.yml
+++ b/.github/workflows/warehouse_system_tests.yml
@@ -116,8 +116,8 @@ jobs:
       # Tell Puppeteer to skip installing Chrome. We'll be using the installed package.
       PUPPETEER_SKIP_CHROMIUM_DOWNLOAD: true
       PUPPETEER_EXECUTABLE_PATH: /usr/bin/chromium
-      FERRUM_PROCESS_TIMEOUT: 45
-      FERRUM_DEFAULT_TIMEOUT: 45
+      FERRUM_PROCESS_TIMEOUT: 90
+      FERRUM_DEFAULT_TIMEOUT: 60
 
     steps:
       - name: Checkout
--- a/.github/workflows/warehouse_system_tests.yml
+++ b/.github/workflows/warehouse_system_tests.yml
@@ -159,7 +159,7 @@ jobs:
           CHROME_URL: http://chrome:3333
           TODO_OR_DIE_LOG_ONLY: true
         run:
-          bundle exec rspec -fd --fail-fast=$MAX_FAILURES --color --pattern "spec/system/**/" --tag type:rails_system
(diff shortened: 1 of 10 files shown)


eanders:
@ttoomey, this all works locally for me, but I'd love some thoughts on how best to implement the GitHub Action.  I _think_ currently it'll get dropped into some standard test run, but will fail as it needs `RUN_RAILS_SYSTEM_TESTS` to be set.
I think ideally, it would either be included in the other system tests or its own run.

config/environments/test.rb
```diff
@@ -88,4 +88,9 @@
   routes.default_url_options = { host: ENV['FQDN'] }
 
   config.force_ssl = false
+
+  # Enable asset compilation for system tests
+  config.assets.compile = true
+  config.assets.check_precompiled_asset = false
+  config.assets.digest = false
```
ttoomey: only needed for warehouse system tests
```suggestion
  if ENV['RUN_RAILS_SYSTEM_TESTS'] == 'true'
    config.assets.compile = true
    config.assets.check_precompiled_asset = false
    config.assets.digest = false
  emd
```

.github/workflows/warehouse_system_tests.yml
```diff
@@ -0,0 +1,173 @@
...
+          TODO_OR_DIE_LOG_ONLY: true
+        run:
+          bundle exec rspec spec/system/rails --tag type:rails_system -fd --fail-fast=$MAX_FAILURES --color
+
+      - name: Archive artifacts
+        if: always()
+        uses: actions/upload-artifact@4cec3d8aa04e39d1a68397de0c4cd6fb9dce8ec1 # v4
+        with:
+          if-no-files-found: ignore
+          name: artifacts-${{ matrix.test_group.id}}
```
ttoomey: I think this might be a copy/paste issue from the system test, probably should be fixed in both to `name: artifacts-rails-system` or something, matrix isn't defined here(?)

spec/support/rails_system_setup.rb
```diff
@@ -0,0 +1,147 @@
...
+
+  def take_screenshot_for_debugging(name = nil)
+    screenshot_name = name || "debug_#{Time.current.to_i}"
+    screenshot_path = File.join(Capybara.save_path, 'screenshots', "#{screenshot_name}.png")
+    page.save_screenshot(screenshot_path) # rubocop:disable Lint/Debugger
+    puts "Screenshot saved: #{screenshot_path}"
+  end
+end
+
+RSpec.configure do |config|
```
ttoomey: enclose in   `if rails_system_enabled`?

.github/workflows/warehouse_system_tests.yml
```diff
@@ -0,0 +1,173 @@
...
+        image: index.docker.io/browserless/chrome@sha256:efac47cfff3830d9a50b27d29f8bbb61949058ae336c823fbe9bd3c0d1debcc8 # 1-chrome-stable
+        ports:
+          - 3333:3333
+        env:
+          TZ: "America/New_York"
+          PORT: 3333
+          CONNECTION_TIMEOUT: 600000
+
+    env:
+      CLIENT: test
```
ttoomey, line 76: probably set globally (for asset compilation, etc) rather than in the step
```
      RUN_RAILS_SYSTEM_TESTS: 'true'
```

ttoomey (approved):
Thanks for working this out, that's great!

ttoomey:
> this all works locally for me, but I'd love some thoughts on how best to implement the GitHub Action. I think currently it'll get dropped into some standard test run, but will fail as it needs RUN_RAILS_SYSTEM_TESTS to be set.
I think ideally, it would either be included in the other system tests or its own run.

I'm probably missing something but we can't just set it in the global env instead of on that one step? Happy to pair on it

eanders:
@ttoomey, sorry, wrote that comment before I took a stab at it.  Everything's working now, and your comments look good.  I'll clean it up tomorrow and confirm tests work correctly.

PR #5672, opened 2025-08-24
8008 ce paper trail

ttoomey:
## _Merging this PR_

## Description
- **Soft-deletion migration**: Added `deleted_at` to CE and WorkflowDefinition/Execution tables; updated unique indexes to exclude soft-deleted rows via partial indexes
- **Enable ActsAsParanoid**: Activated `acts_as_paranoid` on CE models (`CustomReferralStatus`, `Match::Rule`, `Opportunity`, `Referral`, `ReferralNote`, `ReferralParticipant`), WorkflowDefinition (`Flow`, `Node`, `Swimlane`, `Template`), WorkflowExecution (`Instance`, `Step`, `StepAssignment`), and HMIS Units (`Unit`, `UnitGroup`, `UnitOccupancy`, `UnitType`).
- **Enable PaperTrail**: Added `has_paper_trail` to the same models; resolved method conflict on `WorkflowDefinition::Template` with `version: :paper_trail_version`.
- **Specs – shared behavior**: Added shared examples for paranoia/versioning and applied across all affected models
- **New factories**: Added factories for `ReferralNote`, `ReferralParticipant`, `WorkflowDefinition::Swimlane`, and `WorkflowExecution::StepAssignment` to support tests.

## Type of change
New feature

### DIFF
--- a/drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
+++ b/drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
@@ -27,6 +27,7 @@
 #
 module Hmis::Ce::Match
   class CandidatePool < GrdaWarehouseBase
+    # Bulk-managed, does not log to paper_trail
     self.table_name = 'ce_match_candidate_pools'
     has_one :change_marker, as: :trackable, class_name: 'Hmis::Ce::ChangeMarker', dependent: :destroy
     has_many :candidates, class_name: 'Hmis::Ce::Match::Candidate', foreign_key: :candidate_pool_id, dependent: :destroy
--- a/drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
+++ b/drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
@@ -37,6 +37,10 @@ module Hmis::Ce::Match
 
     attr_readonly :requirement_expression, :priority_expression
 
+    # Clean up the link from soft-deleted associations before deleting the pool. This avoids
+    # foreign key constraint violations
+    before_destroy :nullify_deleted_associations
+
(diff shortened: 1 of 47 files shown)


drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
```diff
@@ -36,6 +37,10 @@ class CandidatePool < GrdaWarehouseBase
 
     attr_readonly :requirement_expression, :priority_expression
 
+    # Clean up the link from soft-deleted associations before deleting the pool. This avoids
+    # foreign key constraint violations
+    before_destroy :nullify_deleted_associations
```
ttoomey, line 42: The alternative is to add soft-deletion to candidate pool. Not sure which is better. I guess I'd skip it since we keep orphan pools around for a while before we delete them, so we probably don't need long-term tracking on deleted pools ?

gigxz: This seems fine to me. I am a _little_ nervous about the risk of a pool being accidentally deleted somehow, but it could simply be re-created, which seems pretty harmless.

drivers/hmis/app/models/hmis/unit_group.rb
```diff
@@ -15,6 +15,7 @@
 #     rules (Hmis::Ce::Match::Rule) that apply to all units in the group
 module Hmis
   class UnitGroup < HmisBase
+    acts_as_paranoid
```
gigxz, line 18: thank you! good catch this table had deleted_at col but wasn't using it

gigxz (approved):
looks great thank you!


I did some local testing for deleting units and opportunities via the application

PR #5681, opened 2025-08-26
Reduce restrictions on deleting units/opportunities

martha:
## _Merging this PR_

## Description
- Reduce restrictions on deleting units:
  - On the Unit schema obj, update `deletable` to restrict deleting a unit with an _active_ referral (as opposed to any referrals, past or present)
  - Remove `dependent: :destroy` on the `Unit`'s `opportunities` relationship, but keep it on the `active_opportunity` only
- Reduce restrictions on deleting opportunities:
  - No change needed on the Unit schema `can_be_marked_unavailable` method. This already was less restrictive than the backend (hence the bug described in the ticket)
  - Remove `dependent: :restrict_with_exception` on the `Opportunity`'s `referral` relationship.

## Type of change
Bug fix


drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -17,10 +17,10 @@ class Opportunity < GrdaWarehouseBase
                primary_key: 'identifier',
                class_name: 'Hmis::WorkflowDefinition::Template'
 
-    has_many :referrals, class_name: 'Hmis::Ce::Referral', dependent: :restrict_with_exception
+    has_many :referrals, class_name: 'Hmis::Ce::Referral' # use default dependent behavior, since opportunities are soft-deleted
```
gigxz: I think this comment is a little confusing. For my own understanding, the default behavior is: when an Opportunity is destroyed, the associated referrals continue to reference the soft-deleted opportunity record (in `ce_referrals` `opportunity_id` column)

Maybe this is clearer?

```suggestion
    has_many :referrals, class_name: 'Hmis::Ce::Referral' # Referrals continue to reference the opportunity even after it is soft-deleted, as no dependent behavior is specified.
```

gigxz: @martha This adds some interesting complexity to the relationships. is this change proposed because of the following scenario:
- unit is marked available (opportunity created)
- referral is created and declined
- unit is marked unavailable (opportunity destroyed)
?



When a referral is _accepted_, [we close!](https://github.com/greenriver/hmis-warehouse/blob/66b3d16c9c539e38265a6a133e13a793bf13b71e/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb#L73) the opportunity.

However when a unit is marked as no longer available, we destroy it. I wonder if it would make more sense to call `close!` here instead of destroy: https://github.com/greenriver/hmis-warehouse/blob/66b3d16c9c539e38265a6a133e13a793bf13b71e/drivers/hmis/app/graphql/mutations/ce/mark_units_unavailable.rb#L33 ? I don't think we are deriving any value/meaning from the distinction that an opportunity was "destroyed" vs being "closed".

That approach would allow us to keep the `dependent: :restrict_with_exception` restriction, ensuring that any Referral records reference a non-deleted Opportunity. That feels less complicated to me. What do you think?

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -19,7 +19,7 @@ class Referral < GrdaWarehouseBase
     has_paper_trail
     acts_as_paranoid
 
-    belongs_to :opportunity, class_name: 'Hmis::Ce::Opportunity'
+    belongs_to :opportunity, -> { with_deleted }, class_name: 'Hmis::Ce::Opportunity'
```
gigxz: This would not be necessary if we just closed (not destroyed) opportunities when they are marked as no longer available https://github.com/greenriver/hmis-warehouse/pull/5681/files#r2304969260

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -36,14 +36,17 @@ class Hmis::Unit < Hmis::HmisBase
   has_many :active_unit_occupancies, -> { active }, class_name: 'Hmis::UnitOccupancy', inverse_of: :unit
   has_many :current_occupants, through: :active_unit_occupancies, class_name: 'Hmis::Hud::Enrollment', source: :enrollment
 
-  # A unit may have many historical opportunities (which represent past times when this unit was available and then filled)...
-  has_many :opportunities, class_name: 'Hmis::Ce::Opportunity', inverse_of: :unit, dependent: :destroy
+  # A unit may have many historical opportunities, which represent past times when this unit was available and then filled...
+  has_many :opportunities, class_name: 'Hmis::Ce::Opportunity', inverse_of: :unit
```
gigxz:
```suggestion
  # A unit may have many opportunities, representing current and historical instances of this unit being available.
  # Deleting a unit does not delete associated opportunities; they remain to preserve the relationship to referrals.
  has_many :opportunities, class_name: 'Hmis::Ce::Opportunity', inverse_of: :unit
```
perhaps?

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -36,14 +36,17 @@ class Hmis::Unit < Hmis::HmisBase
...
   # ...but it only has one "latest" opportunity, which could be either:
   # - active and accepting referrals (open),
   # - active with a referral in-progress (locked), or
   # - closed with an accepted referral. This would be prioritized last, after any active opportunity.
   has_one :latest_opportunity, -> { actives_first }, class_name: 'Hmis::Ce::Opportunity', inverse_of: :unit
 
+  # `dependent: :destroy` directive applies to the active opportunity *only* -- don't destroy past opportunities if this unit is deleted
+  has_one :active_opportunity, -> { active }, class_name: 'Hmis::Ce::Opportunity', inverse_of: :unit, dependent: :destroy
```
gigxz: Alternatively (Again in the same vein as other comments,) we could close open opportunities before destroying
```ruby
  before_destroy :close_open_opportunities
  before_destroy :ensure_no_locked_opportunities

  # close open opportunities before deleting unit
  def close_open_opportunities
    opportunities.where(status: 'open').each { |opp| opp.update!(status: 'closed') }
 end

  # I think we also need this too? I don't see anything restricting deleting a unit that has a locked opportunity which indicates an in progress referral
  def ensure_no_locked_opportunities
    if opportunities.where(status: 'locked').exists?
      errors.add(:base, 'Cannot delete Unit with locked Opportunities.')
      throw(:abort) # Prevents the destruction of the Unit
    end
  end
```

drivers/hmis/app/graphql/types/hmis_schema/unit.rb
```diff
@@ -68,7 +68,7 @@ def occupancy_status
 
     def deletable
       return false if occupants.any? # cannot delete unit with occupants
-      return false if load_ar_association(object, :referrals).any? # cannot delete unit with any referrals (past or present)
+      return false if load_ar_association(object, :active_referral).present? # cannot delete unit with active referral
```
gigxz, line 71: looks good 👍

drivers/hmis/app/models/hmis/ce/opportunity.rb
martha: That sounds reasonable to me too! Thanks!

gigxz (approved):
Looks good, thanks for the thorough tests!

PR #5698, opened 2025-08-29
Expand referral visibility through swimlane participation

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8152

Grant users with "can view own referrals" access to view referrals that they are participating in via ReferralParticipant assignment IF any of the completed steps(tasks) on that referral are assigned to their swimlane. This allows retroactively granting access to the referral for a provider even if (some of) their assigned steps have already been completed.

**Bug repro:**
* Start a referral in housing workflow
* Progress through the Provide Outcome task and complete it
* Now, assign a user to the Project Staff swimlane. It should be a user that has "can view own referrals" at the target project. (not the ability to view all referrals)
* **Current behavior**: user cannot see the referral, because they do not have an assignment to any tasks
* **Desired behavior**: user should be granted access to the referral through their swimlane assignment (ReferralParticipant record)

## Type of change
Bug fix / behavior change

### DIFF
--- a/drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
+++ b/drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
@@ -19,7 +19,7 @@ module Hmis::AuthPolicies::ContextLoaders
       (assigned_referral_steps.pluck(:instance_id) + workflow_instance_ids_with_completed_swimlane_steps).to_set
     end
 
-    memoize def assigned_referral_step_ids # this should probably exclude the completed ones?
+    memoize def assigned_referral_step_ids
       assigned_referral_steps.pluck(:id).to_set
     end
 
@@ -72,7 +72,7 @@ module Hmis::AuthPolicies::ContextLoaders
         completed_swimlane_steps = all_steps[instance_id]&.select { |step| step.user_task&.swimlane_id == participant.swimlane_id }
         assigned_instance_ids << instance_id if completed_swimlane_steps&.any?
       end
-      # binding.pry
+
       assigned_instance_ids
     end
 
--- a/drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
+++ b/drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
@@ -16,7 +16,7 @@ module Hmis::AuthPolicies::ContextLoaders
     end
 
     memoize def assigned_referral_instance_ids
-      (assigned_referral_steps.pluck(:instance_id) + workflow_instance_ids_with_completed_swimlane_steps).to_set
+      (assigned_referral_steps.pluck(:instance_id) + referrals_with_completed_swimlane_steps.pluck(:workflow_instance_id)).to_set
     end
 
     memoize def assigned_referral_step_ids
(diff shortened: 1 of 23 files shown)


gigxz:
Adding @eanders as a reviewer here since we discussed this bug and landed on the desired behavior. Adding @ttoomey  as more as an FYI but I know you're out!

There will be another PR to introduce the other behavior we discussed, namely:
- When changing swimlane participant assignment, it should remove previous assignments. Disallow removing the last user.
- Remove behavior that auto-assigns task to the person who completes it
- Interface should display "completed by X user" instead of "Assigned to" for completed tasks

Edit: linking other issue for those https://github.com/open-path/Green-River/issues/8153

drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
```diff
@@ -25,6 +25,57 @@ def initialize(user)
 
     private
 
+    # Returns workflow instance IDs where there are completed steps assigned to
+    # swimlanes that the current user participates in.
+    #
+    # This allows users to see referrals they're involved with through swimlane
+    # participation, even if no steps are currently directly assigned to them.
+    #
+    # Note: this logic is mirrored by `Referral#with_completed_steps_assigned_to_swimlane`
```
eanders: Instead of duplicating the logic, could `drivers/hmis/app/models/hmis/ce/referral.rb` have a class method that takes a user (you currently have a scope, so that's probably fine, but it might be cleaner to have some smaller methods that return the various bits you need.

Then this could just call the class methods to build up the various bit you need here to prevent duplication of logic.

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -65,6 +71,45 @@ class Referral < GrdaWarehouseBase
...
+    # Note: this logic is mirrored by `CeReferralAssignmentLoader#workflow_instance_ids_with_completed_swimlane_steps`
+    scope :with_completed_steps_assigned_to_swimlane, ->(user) do
+      # Get all referral participants for this user, with their workflow instance IDs
+      participants = user.ce_referral_participants.
+        joins(:referral).
+        select(:referral_id, :swimlane_id, 'ce_referrals.workflow_instance_id').
+        distinct
+
+      # Get all unique workflow instance IDs
+      instance_ids = participants.pluck(:workflow_instance_id).compact.uniq
```
eanders: Should this `return none if instance_ids.empty?`

eanders (approved):

drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
gigxz: Thanks @eanders. After coming back to this with fresh eyes, I removed the duplication. Instead of breaking it up into smaller methods, I kept one shared method as a class method on the referral class. I like the idea of breaking it up, but since it uses such specific grouping of records it felt more confusing to have those defined in different places.

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: Yes, updated

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -139,6 +145,41 @@ def self.apply_filters(input)
...
+    # @return [Array<Integer>] Array of referral IDs
+    def self.referral_ids_for_user_with_completed_swimlane_steps(user)
+      # Get all referral participants for this user, with their workflow instance IDs
+      participants = user.ce_referral_participants.joins(:referral).
+        select(:referral_id, :swimlane_id, 'ce_referrals.workflow_instance_id').
+        distinct
+
+      # Get all workflow instance IDs that this user is participating in
+      instance_ids = participants.map(&:workflow_instance_id)
+      return unless instance_ids.any? # user is not participating in any referrals
```
ttoomey: return consistent array?
```suggestion
      return [] unless instance_ids.any? # user is not participating in any referrals
```

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -139,6 +145,41 @@ def self.apply_filters(input)
...
+
+      # Preload all completed steps for these instances in one query, grouped by instance and swimlane
+      completed_steps = Hmis::WorkflowExecution::Step.
```
ttoomey: if a user is involved many referrals, this might become slow since it loads every step they have ever participated in?

AI suggests the following; looks like it might work but maybe not worth additional time just now

```ruby
    def self.referral_ids_for_user_with_completed_swimlane_steps(user)
      rp = Hmis::Ce::ReferralParticipant.arel_table
      ut = Hmis::WorkflowDefinition::UserTask.arel_table

      Hmis::Ce::ReferralParticipant.
        joins(referral: { steps: :user_task }).
        merge(Hmis::WorkflowExecution::Step.completed).
        where(rp[:user_id].eq(user.id)).
        where(ut[:swimlane_id].eq(rp[:swimlane_id])).
        distinct.
        pluck(:referral_id)
    end
```

ttoomey (approved):
Looks good, thank you!

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: Phew so much simpler! My AI needs to talk to your AI 😆 
I started with something similar to this, but ran into the problem of it returning all referrals that have completed steps in swimlane X, even if `user` is only assigned as a participant on one of those. But this suggested scope doesn't have that issue. I will update to use this!

gigxz:
merging despite TodoOrDies

PR #5700, opened 2025-08-29
Migrate units to unit groups

martha:
## _Merging this PR_

## Description
- Add optional Unit Type column to Unit Group.
- Add Active Record constraint on Unit enforcing that a Unit Group can only have units with one Unit Type.
- Update the Create Units mutation to raise errors if that constraint is not met
- Add one-time task to Task Queue for migrating all existing units into Unit Groups

Related frontend PR: https://github.com/greenriver/hmis-frontend/pull/1253

## Type of change
New feature


drivers/hmis/lib/hmis_data_cleanup/migrate_units_to_unit_groups_20250828.rb
```diff
@@ -0,0 +1,100 @@
...
+  def create_unit_groups_for_projects
+    puts 'Creating unit groups for projects that have units without groups'
+
+    # Find all project/unit_type combinations that have units but no unit groups
+    projects_without_unit_groups = Hmis::Hud::Project.
+      joins(:units).
+      left_joins(:unit_groups).
+      where(hmis_unit_groups: { id: nil }).
+      distinct.
+      includes(:units)
```
gigxz: This is fine, I usually find `Hmis::Hud::Project.where.missing(:unit_groups)` helpful with this kind of query since it's a bit easier to read.

Can we add a `hmis` scope to ensure we are only dealing with projects in the HMIS data source? (Others shouldn't have Units,  but just to be safe)

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -12674,7 +12675,7 @@ input UnitInput {
   """
   prefix: String
   projectId: ID!
-  unitGroupId: ID
+  unitGroupId: ID!
```
gigxz: This is a breaking change. Can we revert this and instead have CreateUnits return a validation error if unit group ID is not provided?  I'm just concerned that this kind of thing will throw up an unhandled exception, which is nice to avoid when we can.

Can you also link the corresponding frontend PR for these changes since the API changed?

gigxz (approved):
Looks good!

drivers/hmis/config/initializers/hmis_feature.rb
```diff
@@ -23,3 +23,10 @@
   HmisDataCleanup::Util.fix_disabling_condition_nils!
   HmisDataCleanup::Util.fix_race_gender_99s!
 end
+
+TodoOrDie('Remove one-time job', by: '2025-10-01')
```
gigxz: Can we say 10/6 maybe? 10/1 is a loaded day with fy26 release so just thinking to avoid mental load of new errors

drivers/hmis/app/graphql/schema.graphql
martha: Sorry for the misunderstanding, I thought it is not a breaking change because the frontend currently (already on release-178) always provides this argument. https://github.com/greenriver/hmis-frontend/blob/release-179/src/modules/units/components/CreateUnitsDialog.tsx#L188

But yes I can revert and make the change you're describing if you prefer

gigxz: I think that is a misinterpretation of the linked statement, it doesn't _always_ provide the unitGroupId currently. In cases where the project doesn't use unit groups it only provides unit type currently.

martha: I see, thanks! I updated this.

martha: Frontend PR is not ready yet. I will wait to merge this until getting approval on both.

gigxz (approved):

drivers/hmis/lib/hmis_data_cleanup/migrate_units_to_unit_groups_20250828.rb
```diff
@@ -0,0 +1,97 @@
+# frozen_string_literal: true
+
+class HmisDataCleanup::MigrateUnitsToUnitGroups20250828
+  def initialize
+  end
+
+  def perform
```
gigxz, line 7:
```suggestion
  def perform
      return unless Hmis::Unit.exists? || Hmis::UnitGroup.exists? # nothing to migrate
```

drivers/hmis/config/initializers/hmis_feature.rb
```diff
@@ -23,3 +23,10 @@
   HmisDataCleanup::Util.fix_disabling_condition_nils!
   HmisDataCleanup::Util.fix_race_gender_99s!
 end
+
+TodoOrDie('Remove one-time job', by: '2025-10-06')
+if ENV['ENABLE_HMIS_API'] == 'true'
```
gigxz:
```suggestion
if Rails.env.staging? && ENV['ENABLE_HMIS_API'] == 'true'
```

Suggesting to only run this as a TaskQueue in staging, and run it manually in production, more details at https://github.com/open-path/Green-River/issues/7814#issuecomment-3250517530

gigxz (changes requested):
Marking as Request changes for https://github.com/greenriver/hmis-warehouse/pull/5700#discussion_r2319987081 but we can discuss on the issue!

gigxz (approved):

PR #5708, opened 2025-09-01
HMIS HUD compliance: new maintainer to ensure system form rules exist in all envs

gigxz:
## _Merging this PR_

## Description
Ticket: https://github.com/open-path/Green-River/issues/8326

### PR Summary

**Pulled out logic for HUD-compliant form instance creation**
- New `HmisUtil::HudComplianceFormInstanceMaintainer` class holds all HUD-compliant form instance creation/fixup. Form definitions are still loaded only by `JsonForms` (no change to what’s loaded).
- HudComplianceFormInstanceMaintainers supports "dry run." When it creates or updates instances, it logs, prints a summary to the console, and can notify Slack.

**Deployed environments now stay in sync**
- Previously, SeedMaker only ran `create_default_occurrence_point_instances` in development, so production/staging did not get new or updated system instances when rules changed. `JsonForms.seed_all` now runs the maintainer so all required system instances are ensured for HUD compliance. 

**Instance behavior changes**
- **Current Living Situation (CLS):** HUD-compliant system instances are now created by the maintainer. All customers use a single CLS form (see https://github.com/open-path/Green-River/issues/7660). This should have **no impact** on real world CLS collection because they already have rules that extend beyond what's required by HUD.
- **Post-exit:** Instances are now created by the maintainer (funder-specific). Previously they had to be added manually, which could lead to HUD compliance gaps. (No impact on current environments, they already have the needed rules)
- **Move-in date:** Instances are now created for the new FY26 funder; previously that funder’s instance had to be added manually. (No impact on current environments, they already have the needed rules)

**Test Setup Changes**
- `spec/rails_helper.rb` still runs `HmisUtil::JsonForms.seed_all`, but **the behavior of that method has changed**:
  - Before: `seed_all` would create system instances for _assessments_, but not for other record types
  - Now: `seed_all` creates all necessary system instances for assessments and other record types (E.g. client)
- Impact:
  - If tests need to test from a "clean slate" of instances, they are responsible for deleting and re-seeding them

## Type of change
new feature / stricter HUD compliance enforcement


drivers/hmis/lib/hmis_util/hud_compliance_form_instance_maintainer.rb
```diff
@@ -0,0 +1,195 @@
...
+        identifier: FORM_IDENTIFIERS[:date_of_engagement],
+        data_collected_about: :HOH_AND_ADULTS,
+        project_types: HudHelper.util.doe_project_types,
+      )
+      # Ensure required rules exist for PATH Status (Occurrence Point form)
+      create_system_instances!(
+        identifier: FORM_IDENTIFIERS[:path_status],
+        data_collected_about: :HOH_AND_ADULTS,
+        funders: HudHelper.util.path_funders,
+      )
```
gigxz, line 81: Moved from JsonForms. Only change is adding in the new move_in_date_funders.

drivers/hmis/lib/hmis_util/hud_compliance_form_instance_maintainer.rb
```diff
@@ -0,0 +1,195 @@
...
+
+      # Ensure required rules exist for Current Living Situation
+      HudHelper.util.current_living_situation_funder_applicability_requirements.each do |spec|
+        create_system_instance!(
+          identifier: FORM_IDENTIFIERS[:current_living_situation],
+          data_collected_about: :HOH_AND_ADULTS,
+          project_type: spec[:project_type],
+          funder: spec[:funder],
+        )
+      end
```
gigxz: NEW: enforce CLS system rules for minimum HUD compliance. See HudUtility for complicated list of requirements.

lib/util/hud_utility_2026.rb
```diff
@@ -934,4 +949,65 @@ def age_range
...
+  #   Services Only, Street Outreach, or Coordinated Entry. We err on the side of collecting more data and
+  #   require CLS collection for all project types funded by CoC SSO.
+  # - YHDP: The data dictionary requires collection for any project type serving clients who meet Category 2 or 3
+  #   of the homeless definition. We err on the side of collecting more data and require CLS collection for
+  #   all YHDP programs.
+  # - Funder-only rules: When a funder is only used for one project type, we do not explicitly add the project
+  #   type to the rule. For example, HUD ESG has a specific funder for Street Outreach; we do not add project
+  #   type to that rule. The requirement is ultimately driven by the funder, and we err on the side of
+  #   collecting CLS for that funder even if the project type is misconfigured.
+  def current_living_situation_funder_applicability_requirements
```
gigxz, line 967: Some additional context here, if you are wondering why I opted to create the specific  funder rules instead of "collect CLS for all Street Outreach /CE/NBN" which is how must customers are set up.

The reason is to support the option to **not** collect CLS in non-HUD-funded projects of those types. System rules cannot be deleted, so we don't want to lock communities in to enabling CLS collection for all NBN shelters, for example. Customers are free to add additional rules (E.g. "collect CLS at all NBN shelters") where desired. (And this change does not impact such existing rules).

gigxz:
Note: Tests on commit https://github.com/greenriver/hmis-warehouse/pull/5708/commits/1bcf31cf7b670fcbdb9309ef720565b51f0b722e were green (overall build failure from minimatch vuln)

drivers/hmis/lib/hmis_util/hud_compliance_form_instance_maintainer.rb
```diff
@@ -0,0 +1,195 @@
...
+      )
+      # Ensure required rules exist for PATH Status (Occurrence Point form)
+      create_system_instances!(
+        identifier: FORM_IDENTIFIERS[:path_status],
+        data_collected_about: :HOH_AND_ADULTS,
+        funders: HudHelper.util.path_funders,
+      )
+
+      # Ensure required rules exist for Current Living Situation
+      HudHelper.util.current_living_situation_funder_applicability_requirements.each do |spec|
```
martha: Maybe add a comment here about `spec`'s expected shape? And/or rename the variable "spec" to "requirement"?

drivers/hmis/lib/hmis_util/hud_compliance_form_instance_maintainer.rb
```diff
@@ -0,0 +1,195 @@
...
+    end
+
+    def create_default_system_instance!(identifier:)
+      instance = Hmis::Form::Instance.defaults.find_or_initialize_by(definition_identifier: identifier)
+      was_new = instance.new_record?
+      instance.assign_attributes(active: true, system: true)
+      return unless instance.changed?
+
+      instance.save! unless @dry_run
+      payload = { type: :default, definition_identifier: identifier }
```
martha: Maybe this `payload` could be loosely typed as an OpenStruct

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -7,47 +7,55 @@
 # frozen_string_literal: true
 
 module HmisUtil
+  # Loads HMIS form definitions from version-controlled JSON files into the database.
+  #
+  # By default, ensures HUD-compliant system form instances exist using HudComplianceFormInstanceMaintainer. (create_instances option)
+  # Optionally, ensures custom data element definitions (CDEDs) referenced by custom_field_key mappings exist. (generate_cdeds option)
```
martha: Nice, this is much neater. Maybe add a comment saying something like "by default generate_cdeds is true except in test"

drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+require 'rails_helper'
+
+RSpec.describe HmisUtil::HudComplianceFormInstanceMaintainer do
+  let!(:data_source) { create(:hmis_data_source) }
+
+  before(:each) do
+    described_class.new.ensure_all_system_instances_exist!
```
martha, line 17:
```suggestion
    # Start all tests with the correct, hud-compliant database state.
    # In individual test setup, delete instances when testing re-creation behavior.
    described_class.new.ensure_all_system_instances_exist!
```
, or similar?

drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+      count_before = Hmis::Form::Instance.count
+
+      described_class.new.ensure_all_system_instances_exist!
+
+      expect(Hmis::Form::Instance.count).to eq(count_before + 2)
+      expect(Hmis::Form::Instance.system.where(definition_identifier: 'base-intake')).to exist
+      expect(Hmis::Form::Instance.system.where(definition_identifier: 'base-exit')).to exist
+    end
+
+    it 'updates existing rule from system: false to system: true without creating a duplicate' do
```
martha: Maybe this test could be removed since the same behavior is also tested below on line 72?

drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+    it 'recreates missing record form instances' do
+      # Delete some system instances (e.g. default system form instances)
+      Hmis::Form::Instance.system.where(definition_identifier: 'client').delete_all
+      Hmis::Form::Instance.system.where(definition_identifier: 'project').delete_all
+      deleted_count = 2
+      count_before = Hmis::Form::Instance.count
+
+      described_class.new.ensure_all_system_instances_exist!
+
+      expect(Hmis::Form::Instance.count).to eq(count_before + deleted_count)
```
martha:
```suggestion
      deleted_count = 2
      count_after_deletion = Hmis::Form::Instance.count

      described_class.new.ensure_all_system_instances_exist!

      expect(Hmis::Form::Instance.count).to eq(count_after_deletion + deleted_count)
```

drivers/hmis/lib/hmis_util/hud_compliance_form_instance_maintainer.rb
```diff
@@ -0,0 +1,195 @@
...
+        identifier: FORM_IDENTIFIERS[:path_status],
+        data_collected_about: :HOH_AND_ADULTS,
+        funders: HudHelper.util.path_funders,
+      )
+
+      # Ensure required rules exist for Current Living Situation
+      HudHelper.util.current_living_situation_funder_applicability_requirements.each do |spec|
+        create_system_instance!(
+          identifier: FORM_IDENTIFIERS[:current_living_situation],
+          data_collected_about: :HOH_AND_ADULTS,
```
martha, line 88: I'm not sure yet if this is an actionable comment, but I'm trying to think through how these new system instances interact with user-created rules, vis a vis the "detect best instance" logic. https://github.com/greenriver/hmis-warehouse/blob/4e1df78853dbfdfc9c73aadd72ce437772bbb35b/drivers/hmis/app/models/hmis/form/instance_project_match.rb#L17-L25

Example/thought experiment:
- This maintainer creates system instances for the move_in_date form. They apply to specific project types/funders, and collect move-in date for HoH.
- Later, suppose a user creates another rule for move_in_date. Let's say it applies to a specific project, and collects only for Veterans (not HoH). 
- This rule is considered a higher ranked match than the system rule, because it's project-specific. Would the project incorrectly only collect move-in date for Veterans?

I haven't been able to reproduce a scenario where this actually leads to hud non-compliance, because with occurrence point forms and current living situation, we are always choosing the "best instance for _enrollment_" which will select the system instance if it applies to the enrollment, even if there's a higher-ranked instance on the project which doesn't apply to the enrollment.

Still, wanted to write up these thoughts and see if it's something you already thought through for all form types, and/or something you think we should document somewhere.

martha (approved):
All looks great to me. Thanks for the increased test coverage!

drivers/hmis/spec/lib/hmis_util/json_forms_spec.rb
```diff
@@ -0,0 +1,230 @@
...
+  before(:all) do
+    Hmis::Form::Definition.delete_all
+    Hmis::Form::Instance.delete_all
+  end
+
+  # Clean up any seeded forms between tests
+  after(:each) do
+    Hmis::Form::Definition.delete_all
+    Hmis::Form::Instance.delete_all
+  end
```
ttoomey: This may not be needed. This is in after-each, which is happening in the transaction. So we create form records outside a tx, run the tests, delete the form records, then tx rolls back restoring the form records.

drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
```diff
@@ -0,0 +1,115 @@
...
+
+  before(:each) do
+    described_class.new.ensure_all_system_instances_exist!
+  end
+
+  after(:all) do
+    # clear and reset to original state
+    Hmis::Form::Instance.delete_all
+    HmisUtil::JsonForms.seed_all
+  end
```
ttoomey: Something to consider. I asked AI if we could DRY this up a bit; it suggested a shared `manages_hmis_form_state` tag which can be re-used in our other specs

```ruby
RSpec.configure do |config|
  # When a spec group is marked with :manages_hmis_form_state, we ensure that
  # the global HMIS form state is restored after the group completes.
  # This is necessary because many HMIS tests rely on a "global" seeded state
  # of forms and system instances for performance reasons.
  config.after(:all, :manages_hmis_form_state) do
    Hmis::Form::Instance.delete_all
    Hmis::Form::Definition.delete_all
    HmisUtil::JsonForms.seed_all
  end
end
```

which could be applied as
```diff
--- a/drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
+++ b/drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
@@ -8,19 +8,13 @@
 
 require 'rails_helper'
 
-RSpec.describe HmisUtil::HudComplianceFormInstanceMaintainer do
+RSpec.describe HmisUtil::HudComplianceFormInstanceMaintainer, :manages_hmis_form_state do
   let!(:data_source) { create(:hmis_data_source) }
 
   before(:each) do
     described_class.new.ensure_all_system_instances_exist!
   end
 
-  after(:all) do
-    # clear and reset to original state
-    Hmis::Form::Instance.delete_all
-    HmisUtil::JsonForms.seed_all
-  end
-
   describe '#ensure_all_system_instances_exist!' do
     it 'is idempotent' do
```

ttoomey (approved):
Looks great!

drivers/hmis/spec/lib/hmis_util/hud_compliance_form_instance_maintainer_spec.rb
gigxz: thanks for the suggestion!

PR #5714, opened 2025-09-04
Re-Enable ROI Revokation & Effective Date Changes

dtgreiner:
## _Merging this PR_

## Description

The ROI file revokation date and effective on date were permanantly disabled. This re-adds the javascript to allow these fields to be updated on the ROI.

## Type of change

Bug Fix


app/javascript/controllers/client_file_row_controller.js
```diff
@@ -0,0 +1,10 @@
+import { Controller } from "@hotwired/stimulus"
+
+// Connects to data-controller="client-file-row"
+export default class extends Controller {
+  connect() {
+    this.element.querySelectorAll('.enable-on-load').forEach((input) => {
+      input.disabled = false
+    })
+  }
+}
```
eanders: This feels more like it should be global, it's just targeting the global `.enable-on-load` class that we previously used to turn disabled inputs into enabled inputs after load.  

I think I'd just add this after the tooltip enabling in:
`app/javascript/application_esbuild.js`

eanders: Alternatively, we could setup a `FormInputsController` and bring it in globally to `app/javascript/controllers/index.js`.  That would give us some compartmentalization and structure, but given that it's really a one-liner, I'd probably just leave a note for a future TODO to move it if we ever need other form-related JS.

dtgreiner: makes sense to me, I was only seeing the `enable-on-load` class being used in this scenario so tried to make it more localized. I'll pull it into `application_esbuild.js` instead

eanders (approved):
Works for me, thanks!

PR #5659, opened 2025-08-18
Add support for Consolidated Waitlist and configurable hmis columns and filters

gigxz:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch
- use a merge-commit or rebase strategy for PRs targeting the stable branch

## Description
Issue: https://github.com/open-path/Green-River/issues/8043
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1235

Add queries to support the Consolidated Waitlist table. The Consolidated Waitlist is a table of ClientProxy records, and it includes any ClientProxy that currently belongs to one or more Candidate Pools.

The table also has "drill down" functionality to click into a ClientProxy and see all **Unit Groups** that the client is eligible for.

#### Changes:
- **New Table**
  - `hmis_table_configurations` stores configurations for table Columns and table Filters as json
- **New GraphQL Types**:
  - `CeClient`: backed by CeClientProxy. This is the backing type for rows of the consolidated waitlist table.
  - `CeEligibleUnitGroup`: represents eligibility relationship between a candidate and unit group. This is used to render the drilldown dialog for consolidated waitlist (eg "which unit groups is this client on the waitlist for?")
  - `TableFilterConfig` and `TableColumnConfig`: generic approach for defining dynamic (db-configured) filters and columns
- Added `Hmis::Filter::CeClientFilter` for searching/filtering ce clients (consolidated waitlist table)

#### How to test
* pull down frontend PR
* navigate to admin=>ce=>consolidated waitlist to view a list of clients that are currently belong to any candidate pool
* click on a client to see which projects and unit groups they are eligible for
* To test dynamic columns and filters, set up table configs locally. here's my local setup for reference. There are also example JSONs in the table configuration model file.
```json
[
  {
    "id": 2,
    "data_source_id": 4,
    "table_key": "ce_waitlist",
    "owner_type": "Hmis::UnitGroup",
    "owner_id": 100,
    "columns": "[{\"key\": \"cde.custom_assessment.housing_needs_aha_score\", \"type\": \"string\", \"label\": \"AHA Score\"}, {\"key\": \"cde.custom_assessment.housing_needs_chronically_homeless\", \"type\": \"string\", \"label\": \"Chronically Homeless\"}, {\"key\": \"custom_assessment.ac_housing_needs.date_created\", \"type\": \"date\", \"label\": \"Assessment Date\"}]",
    "filters": "[]",
    "created_at": "2025-08-24 14:42:48.587302",
[...]

PR #5638, opened 2025-08-09
8044 CE consolidated waitlist support

ttoomey:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

#### Major Changes

* **UnitGroup-Based Matching**: Waitlist rules and candidate pools are now managed at the `UnitGroup` level instead of individual `Opportunity` records. 

* **Real-Time Pool Updates**: Removed the periodic `BuildCandidatePoolsJob`. Rule and UnitGroup changes now trigger candidate pool rebuilds through ActiveRecord callbacks.

* **Required CandidatePool Setup**: Units can only be marked available for CE if they are associated with a UnitGroup. The `markUnitsAvailable` mutation enforces this requirement.

* **Rule Snapshots**: Referrals show eligibility criteria from when the Opportunity was created, using the `assignment_rules` on the opportunity record.

#### Implementation Details

* **Rule Ownership**: Match rules are limited to `UnitGroup`, `Project`, or `Organization` owners; rules can no longer be owned by `Unit` or `Opportunity`
* **Schema Changes**: Added `candidate_pool_id` foreign key to `hmis_unit_groups` table
* **Code Restructure**: Split `CandidatePoolResolver` into:
  * `UnitGroupRuleResolver`: Determines the effective set of rules for a `UnitGroup`
  * `CandidatePoolRepository`: Handles CandidatePool persistence and lookup
* **Rule Precedence**: Deterministic ordering (`UnitGroup > Project > Organization`, then by ID)
* **Data Integrity**: `Rule` and `UnitGroup` models prevent changes to critical attributes after creation
* **Service Refactor**: `CandidatePoolBuilder` is now stateless; callers handle locks and transactions
* **Cleanup Updates**: Orphan cleanup checks both `UnitGroup` and `Opportunity` references; `ProcessClientsJob` removes dangling `ChangeMarker` records

## Type of change
New feature

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [ ] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #5715, opened 2025-09-04
Consider unit group candidate pools active, fix global eligible client list bug

gigxz:
## _Merging this PR_

## Description
This is a follow-up to https://github.com/greenriver/hmis-warehouse/pull/5638 and https://github.com/greenriver/hmis-warehouse/pull/5659 to make a couple adjustments:

1. On the `ceClients` query (which backs the "Consolidated Waitlist"), only include clients that belong to candidate pools that are **active**. Previously it was returning clients that belonged to any candidate pool.
2. Change the definition of an **active candidate pool** to include not only pools tied to active opportunities, but also **pools tied to unit groups where waitlists are enabled**. Pools that are tied to Unit Groups reflect the clients eligible based on the current requirements, which should be maintained actively.
3. Adjust `CandidatePoolBuilder` so that it only maintains Unit Group-level candidate pools for **Unit groups belonging to projects that have "waitlists" enabled**, rather than maintaining them for all projects which is not necessary. This is needed because rules can be applied globally to all projects, but they only really need to be maintained if the project is set up to use waitlists.

This fixes a bug on the global Eligible Clients list (consolidated waitlist) that causes lots of clients to show up erroneously because they belong to inactive pools.

## Type of change
Bug fix / new feature


gigxz:
This isn't urgent, I worked around the immediate presenting "bug" in the global eligible client list by manually deleting all the inactive (by my new definition) candidate pools.

martha (approved):
Makes sense. Thank you!

drivers/hmis/app/models/hmis/ce/match/candidate_pool.rb
```diff
@@ -41,13 +41,21 @@ class CandidatePool < GrdaWarehouseBase
...
-    # pools for active opportunities
     scope :active, -> {
-      active_ids = ::Hmis::Ce::Opportunity.active.pluck(:candidate_pool_id).compact.uniq
-      where(id: active_ids)
+      # Pool is active if there are any active Opportunities that reference it
+      active_ids_for_opportunities = ::Hmis::Ce::Opportunity.active.pluck(:candidate_pool_id).compact.uniq
+      # Pool is active if there are any UnitGroups that reference it
+      active_ids_for_unit_groups = Hmis::UnitGroup.with_ce_waitlists_enabled.pluck(:candidate_pool_id).compact.uniq
+
+      where(id: active_ids_for_opportunities + active_ids_for_unit_groups)
```
ttoomey: small optimization: could use distinct on the queries and unique on the id set 
```suggestion
      active_ids_for_opportunities = ::Hmis::Ce::Opportunity.active.distinct.pluck(:candidate_pool_id).compact
      # Pool is active if there are any UnitGroups that reference it
      active_ids_for_unit_groups = Hmis::UnitGroup.with_ce_waitlists_enabled.distinct.pluck(:candidate_pool_id).compact
      all_active_ids = active_ids_for_opportunities + active_ids_for_unit_groups).sort.uniq
      where(id: all_active_ids)
```

ttoomey (approved):
Looks good, thanks!

PR #5719, opened 2025-09-04
Fix bug in Accepted Pending => Denied Pending workflow

martha:
## _Merging this PR_

## Description
- Progress a referral through to Provider Outcome, select Accept, then open the Change Provider Outcome task and submit.
- Open the Denial Review task, select Send Back
- Open the Provider Outcome (Second Attempt) task, select Accept
Previously: [error](https://green-river.sentry.io/issues/6858284403/?referrer=slack&notification_uuid=d9428290-8e21-4398-99f3-b0171c90d371&environment=staging&alert_rule_id=14733065&alert_type=issue)
With this change: Error is not raised 

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/workflow_execution/engine.rb
+++ b/drivers/hmis/app/models/hmis/workflow_execution/engine.rb
@@ -40,7 +40,7 @@ module Hmis::WorkflowExecution
         raise "Failed to reopen step #{step.id} because it had an irreversible side effect. This indicates a misconfigured workflow."
       end
 
-      return step if step.status == 'available' # If the step is already available, no action needed
+      return step if step.open? # If the step is already available, no action needed
 
       step.available_at = Time.current
       step.enable!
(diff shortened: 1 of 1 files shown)


drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -40,6 +40,8 @@ def enable_step!(node)
         raise "Failed to reopen step #{step.id} because it had an irreversible side effect. This indicates a misconfigured workflow."
       end
 
+      return step if step.open? # If the step is already available, no action needed
```
gigxz, line 43: I guess we don't need to also check that `step.persisted?` because if it was new, `find_or_initialize_by` would initialize it with `unavailable` state? I think thats right, but just calling it out.

gigxz (approved):

drivers/hmis/app/models/hmis/workflow_execution/engine.rb
martha: Yep I think that's right too!

PR #5721, opened 2025-09-05
Add Calculate Client Eligibility for form component

martha:
## _Merging this PR_

## Description
- Update `ClientPoolEvaluator` to accept overrides in order to calculate eligiblity based on form values
- Add new mutation `Ce::CalculateClientEligibility`
- Add new form component `CLIENT_ELIGIBILITY`, a type of Choice component

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1257

## Type of change
New feature

- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see linked frontend PR


drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
```diff
@@ -0,0 +1,83 @@
...
+    argument :values_by_link_id, Types::JsonObject, required: true
+
+    field :project_types, [Types::HmisSchema::Enums::ProjectType], null: false
+
+    def resolve(enrollment_id:, form_definition_identifier:, values_by_link_id:)
+      enrollment = Hmis::Hud::Enrollment.viewable_by(current_user).find(enrollment_id)
+      access_denied! unless current_user.can_edit_enrollments_for?(enrollment)
+
+      client = enrollment.client.destination_client
+      return { project_types: [] } unless client
```
martha: From a UX perspective, this button should ideally work regardless of whether the client has a destination client yet -- after all, it's basing eligibility off of the current form values. But it's more complicated than that internally, because our match engine expects a destination client, and the current form values are just overrides. I'd like to discuss possible solutions (and whether they're worthwhile to pursue)

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
```diff
@@ -0,0 +1,84 @@
...
+        ["cde.custom_assessment.#{custom_field_key}", value]
+      end.to_h
+    end
+
+    def calculate_eligible_project_types(client, field_value_overrides)
+      field_map = Hmis::Ce::Match::Expression::FieldMap.new(current_date: Date.current)
+      eligible_pools = []
+
+      # Evaluate client against active candidate pools
+      Hmis::Ce::Match::CandidatePool.active.find_each do |pool|
```
martha: Will work in conjunction with https://github.com/greenriver/hmis-warehouse/pull/5715. (Spec will fail until that is merged)

drivers/hmis/app/models/hmis/ce/match/internal/client_pool_evaluator.rb
```diff
@@ -60,10 +65,18 @@ def call(client)
     # evaluate!('current_age >= 65 AND veteran_status = 1', {current_age: 50, veteran_status: 1})
     def eval_requirement(client_values)
       @calculator.evaluate!(@pool.requirement_expression, **client_values)
+    rescue Dentaku::Error, ArgumentError => e
```
martha: I was seeing some errors with evaluating clients that had nil values, such as nil DOB/age. I _think_ they are arising now because usually those clients would've been prefiltered out. In this case, we are just evaluating against one client at a time, so prefiltering doesn't make sense, but that means the evaluator needs to become more flexible to values that can't be compared.

Let me know if you have any concerns about the behavior I've introduced here!

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
```diff
@@ -0,0 +1,84 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Mutations
+  class Ce::CalculateClientEligibility < CleanBaseMutation
```
ttoomey: Implementing this as a mutation is a little confusing as it doesn't seem to change state. So it would be more conventional to implement this as a query. But perhaps the front-end makes this inconvenient.

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
```diff
@@ -0,0 +1,84 @@
...
+    def find_project_types_from_pools(pools)
+      return [] if pools.empty?
+
+      # Get project types from Unit Groups that use these pools.
+      # (Unit Groups and not Opportunities, because we want to return results about what projects the client is eligible for, regardless of availability)
+      Hmis::UnitGroup.
+        joins(:project).
+        where(candidate_pool_id: pools.map(&:id)).
+        pluck('Project.project_type').
+        compact.uniq
```
ttoomey: I think we'll need with_ce_waitlists_enabled, once https://github.com/greenriver/hmis-warehouse/pull/5715  merged

Maybe better to pluck of off the project model, I'm not sure how pluck 'Project.project_type'  works as it's not quoted correctly
```suggestion
      Hmis::Hud::Project.
        with_ce_waitlists_enabled.
        joins(:unit_groups).
        merge(Hmis::UnitGroup.where(candidate_pool_id: pools.map(&:id)))
        distinct.
        pluck(:project_type)
```

drivers/hmis/app/graphql/types/forms/enums/component.rb
```diff
@@ -41,5 +41,6 @@ class Forms::Enums::Component < Types::BaseEnum
     value 'EMAIL', 'Email address input for ContactPoint'
     value 'SIGNATURE_GROUP', 'Signature input component'
     value 'SIGNATURE', 'Signature input component'
+    value 'CLIENT_ELIGIBILITY', 'Button to fetch information about client eligibility, based on current form values'
```
ttoomey: Maybe we should qualify this as `CLIENT_CE_ELIGIBILITY` as "eligibility" is a bit generic

drivers/hmis/app/models/hmis/ce/match/internal/client_pool_evaluator.rb
ttoomey: I think it would be better not to swallow exceptions if we can avoid it. Maybe it's okay for now but I think we should correct the calculator to handle nils if possible (same issue in `eval_priority`)

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
```diff
@@ -0,0 +1,84 @@
...
+      end.to_h
+    end
+
+    def calculate_eligible_project_types(client, field_value_overrides)
+      field_map = Hmis::Ce::Match::Expression::FieldMap.new(current_date: Date.current)
+      eligible_pools = []
+
+      # Evaluate client against active candidate pools
+      Hmis::Ce::Match::CandidatePool.active.find_each do |pool|
+        clients = GrdaWarehouse::Hud::Client.where(id: client.id)
```
ttoomey: could go out of the loop

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
martha: Say "unable to calculate because client background processing hasn't finished" or something

drivers/hmis/app/models/hmis/ce/match/internal/client_pool_evaluator.rb
martha: Remove this change from this PR, make a ticket

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
martha: I'm happy to change it to a query, but want to check first, @gigxz was there any specific reason you called this out as a mutation in the ticket description?

Just a note, alt-AHA is a bit similar. Although it does mutate data because it saves a log of the calculation.

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
martha: Screenshot 2025-09-08 at 4 10 24 PM

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
martha: Thanks I like your improvement! For my education,
> as it's not quoted correctly

what does this mean?

ttoomey: oh sorry- I was just suprised 'Project.project_type' works.

My understanding is postgres folds identifiers to lowercase unless you quote them. So `select Project.project_type from Project` is the same as `select project.project_type from project` which would be an error. Because we use case-sensitive identifiers, we'd actually want it to be `select "Project"."ProjectType" from "Project"` 

ActiveRecord does this translation/quoting for us usually. Maybe it's smart enough to handle 'Project.project_type' too

martha: I see thanks!

drivers/hmis/app/models/hmis/ce/match/internal/client_pool_evaluator.rb
martha: Papertrail https://github.com/open-path/Green-River/issues/8195

drivers/hmis/app/graphql/mutations/ce/calculate_client_ce_eligibility.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      { project_types: eligible_project_types }
+    end
+
+    private
+
+    def build_overrides(values_by_link_id, form_definition_identifier)
+      # Get the form definition to map from link_id to custom_field_key
+      form_definition = Hmis::Form::Definition.published.find_by(identifier: form_definition_identifier)
+      return {} unless form_definition
```
gigxz: Thinking through: Is this expected? In what scenario would the form not be found? The currently Published version should always be used for editing, but I suppose this could happen if a new version was published _after_ the user opened up the form for editing. In that case, there would be no way for us to know that this should be calculated against the previous (now retired) form.

Perhaps it would make more sense for this mutation to accept a `form_definition_id` so we know exactly which form we're working with? Maybe there's something I'm missing about why `identifier` was chosen as the argument.

drivers/hmis/app/graphql/mutations/ce/calculate_client_ce_eligibility.rb
```diff
@@ -0,0 +1,89 @@
...
+
+      # Build overrides hash
+      values_by_link_id.filter_map do |link_id, value|
+        form_item = form_definition.link_id_item_hash[link_id]
+        custom_field_key = form_item&.dig('mapping', 'custom_field_key')
+        next unless custom_field_key
+
+        # Map to CDE field format for CE evaluation - see FieldMap
+        ["cde.custom_assessment.#{custom_field_key}", value]
```
gigxz, line 55: Just checking-- does this `values_by_link_id` include hidden question with value `"_HIDDEN"` or are they not present in the hash or something else? It would be helpful to leave a comment about that behavior.

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
gigxz: No good reason, sorry for the confusion. I agree it would make sense as a query since it doesn't mutate anything.

drivers/hmis/app/graphql/mutations/ce/calculate_client_ce_eligibility.rb
martha: Yes that seems very reasonable, I was waffling about this too

martha: Requires a small amt of churn in the `AssessmentContext`: https://github.com/greenriver/hmis-frontend/pull/1257/commits/fd0a203d039a775f325ef916d53102d105c53e83

drivers/hmis/app/graphql/mutations/ce/calculate_client_eligibility.rb
martha: I think one reason to keep it as a mutation (and maybe why we did this in the first place for both alt-AHA and client eligibility) is that we get the `BaseMutation`'s default validation error field.

I will leave it as-is for now with comments.

drivers/hmis/app/graphql/mutations/ce/calculate_client_ce_eligibility.rb
martha: They are not present in the hash, I'll add a comment.

- behavior: https://github.com/greenriver/hmis-frontend/blob/fd0a203d039a775f325ef916d53102d105c53e83/src/modules/form/util/formUtil.ts#L1233-L1240
- called here with no `includeMissingKeys` arg: https://github.com/greenriver/hmis-frontend/blob/fd0a203d039a775f325ef916d53102d105c53e83/src/modules/form/util/formUtil.ts#L1435-L1438

gigxz: Got it, thanks!

gigxz (approved):
looks good to me!

PR #5723, opened 2025-09-05
Return one source enrollment assessment per identifier

martha:
## _Merging this PR_

## Description
- Fix bug that was causing old assessments to be returned. Now, only one assessment should be returned per form identifier.
- Ensure that if multiple assessments are returned (because the candidate pool has rules referring to custom data elements on more than one form), they are ordered by assessment date with the most recent appearing first.

## Type of change
Bug fix

- [na] If adding a new endpoint / exposing data in a new way, I have:
  - [na] ensured the API can't leak data from other data sources
  - [na] ensured this does not introduce N+1s
  - [na] ensured permissions and visibility checks are performed in the right places

### DIFF
--- a/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
+++ b/drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
@@ -131,7 +131,7 @@ module Types
       # We don't expect there to be a lot, so calculate in-memory the latest assessment per definition
       assessments.group_by { |assessment| assessment.definition.identifier }.map do |_, group|
         group.max_by { |assessment| [assessment.assessment_date, assessment.id] }
-      end.sort { |a, b| b.assessment_date <=> a.assessment_date } # sort by assessment date, descending
+      end.sort_by(&:assessment_date).reverse # sort by assessment date, descending
     end
 
     def access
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
```diff
@@ -129,9 +129,9 @@ def assessments
         joins(:definition).where(definition: { identifier: object.definition_identifiers })
 
       # We don't expect there to be a lot, so calculate in-memory the latest assessment per definition
-      assessments.group_by(&:definition).map do |_, group|
+      assessments.group_by { |assessment| assessment.definition.identifier }.map do |_, group|
         group.max_by { |assessment| [assessment.assessment_date, assessment.id] }
-      end
+      end.sort { |a, b| b.assessment_date <=> a.assessment_date } # sort by assessment date, descending
```
gigxz: Interesting I have never seen this syntax before. It does look like there are handful of places we use it in the codebase already, so its not a totally new pattern, but new to me! The array should be very small (just a handful of items max) so I would expect `sort_by(&:assessment_date).reverse` to perform just fine and is a bit more readable, but this seems OK too.

gigxz (approved):
looks good!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_enrollment.rb
martha: I'll update it! Agree that this one is more readable.

PR #5729, opened 2025-09-08
Improves database-ready.rb

yakloinsteak:
- Makes `database-ready.rb` more robust and with better logging

## Description
Attempts to fix issues with the database-ready.rb script hanging. Tested with a manual run in an existing container. Test after merging by deploying just to QA.

## Type of change
Bug fix.

### DIFF
--- a/bin/database-ready.rb
+++ b/bin/database-ready.rb
@@ -43,7 +43,7 @@ class DbTester
 
         num = result.values.flatten.first.to_i
         puts "Query result: #{num}"
-        return unless num != 6
+        return if num == 6
       rescue Timeout::Error
         puts 'Database query timed out after 30 seconds'
       rescue StandardError => e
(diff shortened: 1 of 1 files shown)


bin/database-ready.rb
```diff
@@ -31,8 +33,22 @@ def run!
...
+      begin
+        puts 'Executing test query...'
+        result = Timeout.timeout(30) do
+          model.connection.execute('select 2*3')
+        end
+        puts 'Query executed successfully, processing result...'
+
+        num = result.values.flatten.first.to_i
+        puts "Query result: #{num}"
+        return unless num != 6
```
dtgreiner: Any issue removing the double negative here? It feels like it's adding extra complexity to reading what the code is doing.
```suggestion
        return if num == 6
```

dtgreiner (approved):
Looks great to me! I've got one comment but not a blocker.

PR #5736, opened 2025-09-09
CAPER Updates

eanders:
## _Merging this PR_

## Description
* Updates to ensure only rows where a housing assessment was collected with a valid subsidy information are included in CAPER Q24a
* More comparison sum tests

## Type of change
* Bug fix


dtgreiner:
It looks like this is causing the test kits to fail, but right now TUP may be a higher priority especially since we're waiting to Test Kit changes. I'd be OK with skipping these Test Kit tests and we can come back when we have slightly more finalized test kit files.

drivers/hud_apr/spec/models/fy2026/datalab_caper/organization_f_so.rb
```diff
@@ -292,6 +320,18 @@
...
+    it 'Q24a internal integrity checks' do
+      ['B', 'C', 'D', 'E', 'F'].each do |letter|
+        check_sum(
+          question: 'Q24a',
+          source: (2..15).to_a.map { |i| "#{letter}#{i}" },
+          total: "#{letter}16",
+        )
+      end
+    end
+
```
dtgreiner, line 334: Is this needed here, or is it being handled by the validation list  from the TUP spreadsheet?

dtgreiner (approved):

PR #5738, opened 2025-09-09
Add capability of workflow to delete a wip enrollment

martha:
## _Merging this PR_

## Description
- When the "Change Provider Outcome" step is completed, delete the WIP enrollment created by the previous "Provider Acceptance" step.

## Type of change
Client request

### DIFF
--- a/drivers/hmis/app/models/hmis/ce/referral_enroller.rb
+++ b/drivers/hmis/app/models/hmis/ce/referral_enroller.rb
@@ -63,7 +63,6 @@ module Hmis::Ce
         return # in non-dev env: return, we are unable to perform the action
       end
 
-
       referral.update!(target_enrollment: nil)
       enrollment.destroy!
     end
--- a/drivers/hmis/app/models/hmis/ce/referral_enroller.rb
+++ b/drivers/hmis/app/models/hmis/ce/referral_enroller.rb
@@ -51,6 +51,18 @@ module Hmis::Ce
       referral.update!(target_enrollment: enrollment)
     end
 
+    def delete_wip_enrollment(_message)
+      enrollment = referral.target_enrollment
+      return unless enrollment
+
+      raise "target enrollment #{enrollment.id} has already had intake completed" unless enrollment.in_progress?
+
+      Hmis::Hud::Enrollment.transaction do
+        referral.update!(target_enrollment: nil)
+        enrollment.destroy!
+      end
+    end
+
     # Sets the Move-In Date on the target enrollment, based on a date value collected on the step form.
     # This requires a Move-in date item to be on the step form with the following attributes:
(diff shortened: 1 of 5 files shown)


drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -51,6 +51,18 @@ def create_enrollment(message)
       referral.update!(target_enrollment: enrollment)
     end
 
+    def delete_wip_enrollment(_message)
+      enrollment = referral.target_enrollment
+      return unless enrollment
+
+      raise "target enrollment #{enrollment.id} has already had intake completed" unless enrollment.in_progress?
```
martha: Could be raised if provider completes Change Provider Outcome step after having already completed intake

gigxz: What will be the course of action to fix the referral when it gets in this state?

Perhaps we can skip the action and report to sentry (For now) to keep track of this, and come back to it and add a user-facing validation message later..

```suggestion
      unless enrollment.in_progress?
          message = "target enrollment #{enrollment.id} has already had intake completed, unable to perform delete_wip_enrollment message"
          raise message if Rails.env.development?

          Sentry.capture_message(message)
          return # in non-dev env: return, we are unable to perform the action
     end
```

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -51,6 +51,24 @@ def create_enrollment(message)
...
+        raise message if Rails.env.development?
+
+        Sentry.capture_message(message)
+        return # in non-dev env: return, we are unable to perform the action
+      end
+
+      Hmis::Hud::Enrollment.transaction do
+        referral.update!(target_enrollment: nil)
+        enrollment.destroy!
+      end
```
gigxz: None of the other engine code uses transactions because we expect the caller to run engine steps in a transaction. Checking mutations- `SubmitCeReferralStep`, `StartCeReferralStep`, `CreateCeReferral`,`AssignReferralParticipants`, and `CreateDirectCeReferral` all run engine in a transaction. Nested transactions can get weird, so I think its best to avoid that.

```suggestion

      referral.update!(target_enrollment: nil)
      enrollment.destroy!
```

gigxz: I also noticed while looking, that `CreateDirectCeReferral` does run nested transactions (`Hmis::Ce::Referral.transaction do...` and `opportunity.with_lock do...` both create transactions) so that might be something to look further at to see if we can run that mutation in a single transaction instead of nesting.

martha: Sounds good - I'll put that on my list and create a ticket if necessary!

gigxz (approved):

PR #5747, opened 2025-09-10
Non-Manual HUD Report Priorities

dtgreiner:
## _Merging this PR_

## Description

Only bump manually run hud reports for if additional similar reports ar running. This allows hud reports started as backing for additional reports to be processed without being requeued du to priority issues.

This change also adds a check to the scorecard reports to verify the apr backing them ahs been completed before moving forward.

## Type of change

Code clean-up

### DIFF
--- a/app/jobs/reporting/hud/run_report_job.rb
+++ b/app/jobs/reporting/hud/run_report_job.rb
@@ -22,34 +22,39 @@ module Reporting::Hud
       # advisory lock to check the number of jobs running for this generator so we don't
       # all check at exactly the same time and get the same result
       @generator = class_name.constantize.new(report)
-      HudReports::ReportInstance.with_advisory_lock(@generator.class.name, timeout_seconds: 20) do
-        # We can't only count the running delayed jobs because we start a DJ every time we check
-        # So, we'll check the report class for running reports instead.
-        running_reports_count = HudReports::ReportInstance.
-          created_recently.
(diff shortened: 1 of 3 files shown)


app/jobs/reporting/hud/run_report_job.rb
```diff
@@ -22,34 +22,39 @@ def perform(class_name, report_id, email: true)
...
-      HudReports::ReportInstance.with_advisory_lock(@generator.class.name, timeout_seconds: 20) do
-        # We can't only count the running delayed jobs because we start a DJ every time we check
-        # So, we'll check the report class for running reports instead.
-        running_reports_count = HudReports::ReportInstance.
-          created_recently.
-          incomplete.
-          started.
-          for_report(report.report_name).
-          count
+      if report.manual
```
eanders: Can you add a comment here on what `report.manual` means, and why we would or wouldn't skip the advisory lock?

I really want to move the advisory lock check into a method since the run is so far down.  

Something like:

```
# this report was called directly, so we'll make sure there isn't another similar report running before we start
if report.manual
  requeued = check_and_requeue_for_running(@generator.class.name)
  # a similar report is already running, adding this back to the queue so it will be tried again in a bit
  return if requeued
end

run_report(report, @generator, email: email)
```

eanders (approved):
Thank you @dtgreiner !  Looks great!

PR #5751, opened 2025-09-10
Decrease priority of AssessmentQuestionsJob

gigxz:
## _Merging this PR_

## Description
Decrease priority of the `AssessmentQuestionsJob` which processes responses to CE assessments onto new HUD `AssessmentQuestions` records. This job is in place only for installations that use CAS, where the calculators rely on AssessmentQuestions values.

I think it is safe to decrease the priority here because CAS clients are updated daily (I think?) and so this doesn't need to be immediate.

Since this is being queued whenever a CustomAssessment that generates a CE Assessment is submitted, it may compete with other jobs that are queued related to CE candidacy. 

**Tested**: submit a custom assessment that generates a CE assessment, confirm this job is queued with specified priority

## Type of change 
New feature

### DIFF
--- a/drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
+++ b/drivers/hmis/app/jobs/hmis/assessment_questions_job.rb
@@ -4,11 +4,15 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 # Job for processing CustomAssessments that represent CE Assessments (e.g. HAT) into HUD CE AssessmentQuestions table.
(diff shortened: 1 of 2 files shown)


eanders (approved):
@gigxz this is probably fine.  CAS does get updated at least once a day, but _can_ be updated on demand by clicking the "Save" button on a CAS readiness page.  I'd guess that a couple minutes to a couple hours delay won't be an issue.

gigxz:
I see, thanks @eanders! Do you think a higher priority in the 6-10 range would be better? I don't have a sense of the timing impact for each priority, but it is being queued at `0` which seems too high.

eanders:
Let's leave it at 12 and see how it goes.  If people notice that it's slow, we can adjust.

PR #5756, opened 2025-09-11
Review of HudUtility2024

eanders:
## _Merging this PR_

## Description
* All usage of HudUtility2024 reviewed
* Added the concept of HudUtilityCurrent
* Updated any use of HudUtility2024 that, at this time, appears to be something that should reference the current FY over time to use HudUtilityCurrent
* Unit tests for HudUtilityCurrent

## Type of change
* Code clean-up
* New feature


eanders:
Note to self, that there is also logic in `Importers::HmisAutoMigrate` that should take advantage of `HudUtilityCurrent`

drivers/hmis/app/models/hmis/ce/match/expression/calculator_factory.rb
```diff
@@ -31,7 +31,7 @@ def self.build
         ->(identifier) do
           return nil if identifier.nil?
 
-          HudUtility2026.hmis_project_type_key(identifier, true)
+          HudUtilityCurrent.hmis_project_type_key(identifier, true)
```
ttoomey: this breaks if HudUtilityCurrent is 2024, `hmis_project_type_key` is not defined there. Suggest porting `hmis_project_type_key` to 2024

lib/util/hud_utility_current.rb
```diff
@@ -0,0 +1,52 @@
...
+    end
+
+    def hud_csv_version
+      # Move to 2026 in production after 2025-10-01
+      # Move to 2026 in staging after 2025-09-01
+      # Move to 2026 in test and development now
+      cutoff_date = if Rails.env.production?
+        Date.new(2025, 10, 1)
+      elsif Rails.env.staging?
+        Date.new(2025, 9, 1)
```
ttoomey: looks like we used to pin QA to 2024. Okay to change that at this point?

app/models/grda_warehouse/warehouse_reports/dashboard/entered.rb
```diff
@@ -68,16 +70,16 @@ def run!
...
+          next unless project_type_sym == project_type_key
+
+          bucket.each do |group_key, client_count|
+            @data[group_key][:data] << client_count
           end
         end
       end
 
-      data = {
+      data = { # rubocop:disable Lint/UselessAssignment
```
ttoomey: is there some reason to keep the assignment?
```suggestion
      {
```

app/models/grda_warehouse/warehouse_reports/hud_lot.rb
```diff
@@ -233,7 +233,7 @@ def details_by_range
       self_report_dates
     end
 
-    private def set_unknowns(un_processed_dates) # rubocop:disable Do not prefix writer method names with `set_`. (convention:Naming/AccessorMethodName
+    private def set_unknowns(un_processed_dates) # not prefix writer method names with `set_`. (convention:Naming/AccessorMethodName
```
ttoomey: seems like we were trying for  a magic comment to disable the linter here
```suggestion
    private def set_unknowns(un_processed_dates) # rubocop:disable Naming/AccessorMethodName
```

ttoomey (commented):
If we are already incurring code-churn, I suggest an alternative to `HudUtilityCurrent`

- Hud.util factory replaces HudUtilityCurrent: call as `Hud.util.ethnicities`
- Version override per call: `Hud.util('2026').ethnicities`
- Stable, thread-scoped current_version via Hud::Current to prevent rollover during long jobs or requests
- Avoids method_missing
- factory method is more flexible, supports replacing `HudUtility2026` with a singleton class (or something else with the same interface) without needing to change call sites

```ruby
# lib/util/hud.rb
module Hud
  class Current < ActiveSupport::CurrentAttributes
    attribute :hud_csv_version
  end

  class << self
    def util(version = nil)
      version ||= current_version

      case version.to_s
      when '2026'
        HudUtility2026
      when '2024'
        HudUtility2024
      else
        raise ArgumentError, "Unknown version #{version.inspect}"
      end
    end

    def current_version
      pinned = Hud::Current.hud_csv_version
      return pinned if pinned

      # Centralize the cutoff logic here (reuse what you have today)
      result = if Rails.env.production?
        Date.current < Date.new(2025, 10, 1) ? '2024' : '2026'
      elsif Rails.env.staging?
        Date.current < Date.new(2025, 9, 1) ? '2024' : '2026'
      else
        '2026'
      end

      # pin version for this request/job/thread so it stays consistent
      Hud::Current.hud_csv_version = result

      result
    end
  end
end
```

ttoomey:
Also as long as we renaming everything, maybe we should also rename `HudUtility` to `HudUtility2022` for clarity (assuming we are keeping it around)

lib/util/hud_utility_current.rb
eanders: Yes, we were waiting on the test kit which has been released.

eanders:
Thanks @ttoomey, I agree, that looks like a better implementation, and digging into to see if we can remove `HudUtility` at this point makes sense.  If we can't remove it completely I'm tempted to rename it `HudUtilityLegacy` as it covers everything pre-2024.

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_one.rb
```diff
@@ -528,182 +531,115 @@ def columns_for_missing_support enrollment
...
+      counts['unknown_race'] << columns_for_missing_support(client) if unknown?(client[:race_none])
+      counts['unknown_gender'] << columns_for_missing_support(client) if unknown?(client[:gender])
       return counts
     end
 
     def add_bed_utilization
       bed_utilization = []
       support = {}
-      totals = {counts: Hash.new(0), data: Hash.new(Set.new)}
+      totals = { counts: Hash.new(0), data: Hash.new(Set.new) }
```
ttoomey: this looks very much like a bug in the existing code. I suspect the intention was:
```ruby
totals = { counts: Hash.new(0), data: Hash.new { |h, k| h[k] = Set.new } }
```

example to illustrate the problem:
```ruby
irb(main):020> totals = { counts: Hash.new(0), data: Hash.new(Set.new) }
=> {:counts=>{}, :data=>{}}
irb(main):021> totals[:data][:foo].add(1)
=> #<Set: {1}>
irb(main):022> totals[:data][:bar].add(2)
=> #<Set: {1, 2}>
irb(main):023> totals[:data][:baz]
=> #<Set: {1, 2}>
```

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_one.rb
```diff
@@ -528,182 +531,115 @@ def columns_for_missing_support enrollment
...
+          where(date: filter.range).count
         data[:average_daily] = project.service_history.service.
           joins(:client, :project).
-          where(Project: {id: project.id}).
+          where(Project: { id: project.id }).
           where(date: filter.range).
           distinct.
           pluck(*client_columns)
-          counts[:average_daily] = total_count / filter.range.count
+        counts[:average_daily] = total_count / filter.range.count
```
ttoomey: probably a bug, should be `.to_f`

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_three.rb
```diff
@@ -725,7 +660,9 @@ def add_enrolled_length_of_stay
             name = project.ProjectName
             data_map[name] = []
             self.class.length_of_stay_buckets.values.each do |range|
-              data_map[name] << project_counts[project.id][:buckets][range] rescue 0
+              data_map[name] << project_counts[project.id][:buckets][range]
+            rescue StandardError
+              0
```
ttoomey: Probably not correct. I imagine the intention was
```
    begin
      data_map[name] << project_counts[project.id][:buckets][range]
    rescue StandardError
      data_map[name] << 0
    end
```

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_three.rb
```diff
@@ -1323,132 +1236,96 @@ def add_unknown_enrollment client_id:, enrollment:, counts:
...
+      counts['missing_income_at_exit'] << columns_for_destination_support(enrollment) if missing_income(client_id, enrollment, data_collection_stage: 3)
 
       return counts
     end
 
     def add_bed_utilization
       bed_utilization = []
       support = {}
-      totals = {counts: Hash.new(0), data: Hash.new(Set.new)}
+      totals = { counts: Hash.new(0), data: Hash.new(Set.new) }
```
ttoomey: looks incorrect
`Hash.new { |h, k| h[k] = Set.new }`

app/models/report_generators/data_quality/fy2017/q7.rb
```diff
@@ -67,12 +66,12 @@ def add_es_so_answers
         data: inactive.map do |id, enrollment|
           [
             id,
-            client_personal_ids[id],join(', '),
+            client_personal_ids[id], join(', '),
```
ttoomey: we probably meant:
```suggestion
            client_personal_ids[id].join(', ').
```

app/models/report_generators/system_performance/fy2018/measure_four.rb
```diff
@@ -357,29 +373,28 @@ def add_stayer_income universe_of_stayers
...
+        universe_of_stayers[index][:earliest_non_earned_income] = 0
+        if earliest_group[:TotalMonthlyIncome].present?
+          universe_of_stayers[index][:earliest_income] = earliest_group[:TotalMonthlyIncome]
+          universe_of_stayers[index][:earliest_earned_income] = earliest_group[:EarnedAmount] || 0
+          universe_of_stayers[index][:earliest_non_earned_income] = universe_of_stayers[index][:earliest_income] - universe_of_stayers[index][:earliest_earned_income]
+        elsif (amounts = earliest_group.values_at(amount_columns).compact).any?
+          universe_of_stayers[index][:earliest_income] = amounts.sum
+          universe_of_stayers[index][:earliest_earned_income] = earliest_group[:EarnedAmount] || 0
+          universe_of_stayers[index][:earliest_non_earned_income] = universe_of_stayers[index][:earliest_income] - universe_of_stayers[index][:earliest_earned_income]
+        elsif earliest_group[:IncomeFromAnySource] == 99 || earliest_group[:IncomeFromAnySource] = 8 || earliest_group[:IncomeFromAnySource] = 9
```
ttoomey: Probably intended comparison, not assignment
```suggestion
        elsif earliest_group[:IncomeFromAnySource] == 99 || earliest_group[:IncomeFromAnySource] == 8 || earliest_group[:IncomeFromAnySource] == 9
```

lib/util/hud_utility_2024.rb
```diff
@@ -200,7 +227,7 @@ def project_type_number_from_code(code)
 
   def residential_project_type_numbers_by_codes(*codes)
     codes = codes.flatten # Take either array, or multiple parameters
-    codes.map { |code| HudUtility2024.residential_project_type_numbers_by_code[code] }.
+    codes.map { |code| HudHelper.util('2024').residential_project_type_numbers_by_code[code] }.
```
ttoomey: I guess I'd probably keep these self-references as they are, within the HudUtilityX classes it seems find to use a reference to the naked class rather than going through the helper

app/helpers/application_helper.rb
```diff
@@ -148,7 +148,7 @@ def ssn(number)
     if can_view_full_ssn?
       # pad with leading 0s if we don't have enough characters
       number = number.to_s.rjust(9, '0') if number.present?
-      content_tag :span, number.to_s.gsub(HudUtility2024::SSN_RGX, '\1-\2-\3')
+      content_tag :span, number.to_s.gsub(HudHelper.util::SSN_RGX, '\1-\2-\3')
```
ttoomey: this can be done later but this reference assumes that HudHelper.util returns a class. There aren't that many places where we do this; might be worth replacing `HudHelper.util::X` - maybe just add methods directly to HudHelper
```suggestion
      content_tag :span, HudHelper.format_ssn(number)
```

lib/util/hud_helper.rb
```diff
@@ -0,0 +1,98 @@
...
+#   # Use specific version
+#   legacy_util = HudHelper.util('legacy')
+#   current_util = HudHelper.util('2026')
+#
+# @see HudUtility2026
+# @see HudUtility2024
+# @see HudUtilityLegacy
+module HudHelper
+  class Current < ActiveSupport::CurrentAttributes
+    attribute :current_fy
```
ttoomey: maybe just call it `hud_csv_version` to match the method?

spec/lib/util/hud_utility_2026_spec.rb
```diff
@@ -22,28 +22,28 @@
 
       it 'translates between race_none codes and descriptions' do
         race_none_list['values'].each do |value|
-          expect(HudUtility2026.race_none(value['key'])).to eq(value['description'])
-          expect(HudUtility2026.race_none(value['description'], true)).to eq(value['key'])
+          expect(HudHelper.util('2026').race_none(value['key'])).to eq(value['description'])
```
ttoomey: probably also okay to leave the references to `HudUtilityX` here since that's what the unit test is for (we're not testing `HudHelper`)

ttoomey (approved):
Looks good!
* left a few suggests
* left comments about existing bugs uncovered by this change; perhaps we can ignore them if the code is dead (but then perhaps we should just remove the code) 
* db/structure.sql seems to have churn but shouldn't actually have changed here

app/helpers/application_helper.rb
eanders: What do you think about `HudHelper.constants('SSN_RGX')` That could take an optional version `HudHelper.constants('SSN_RGX', version: '2024')` in case we end up having different constants per year.  For now the `constants` method could just call `util::SSN_RGX`, but we could upgrade it in the future as necessary.

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_one.rb
eanders: This is definitely dead code.  I'm tempted to add a sentry call that triggers if the method is called and add a ToDo to remove in 1.5 years.  Does that make sense to you?

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_three.rb
eanders: This is definitely dead code.  I'm tempted to add a sentry call that triggers if the method is called and add a ToDo to remove in 1.5 years.  Does that make sense to you?

app/models/report_generators/data_quality/fy2017/q7.rb
eanders: This is also dead code.

app/models/report_generators/system_performance/fy2018/measure_four.rb
eanders: This is also dead code.

app/models/grda_warehouse/warehouse_reports/project/data_quality/version_three.rb
ttoomey: if we are sure it's dead code, can we just remove it now? Maybe in a separate commit so we can revert if needed

app/helpers/application_helper.rb
ttoomey: > What do you think about HudHelper.constants('SSN_RGX')

I guess exposing constants seems okay as a bridge to something else. For SSNs in the long term I think we planned to remove the `ssn` and `masked_ssn` helpers and instead wrap data in `GrdaWarehouse::PiiProvider` - that would centralize formatting and enforce PII visibility. 

The other place we reference Util constants appears to be for ranges (`HudHelper.util::SITUATION_HOMELESS_RANGE`). 

> That could take an optional version HudHelper.constants('SSN_RGX', version: '2024') in case we end up having different constants per year. 

I guess I'd favor using simple method wrappers rather than add another access path. Methods seem more direct and consistent how the rest of the behavior is works ( so `HudHelper.util.ssn_rgx`)

```ruby
#  module HudUtility2026
#    include ::Concerns::HudUtilityCommonBehaviors
 
module Concerns
  module HudUtilityCommonBehaviors
    extend ActiveSupport::Concern

    class_methods do
      # expose internal constants
      def ssn_rgx = self::SSN_RGX
      def situation_other_range = self::SITUATION_OTHER_RANGE
      def situation_homeless_range = self::SITUATION_HOMELESS_RANGE
      def situation_institutional_range = self::SITUATION_INSTITUTIONAL_RANGE
      def situation_temporary_range = self::SITUATION_TEMPORARY_RANGE
      def situation_permanent_range = self::SITUATION_PERMANENT_RANGE
    end
  end
end
```

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
```diff
@@ -68,7 +68,7 @@ def days_since_last_exit_field
     def veteran_status_field
       {
         query: ->(clients) { clients.pluck(:id, :veteran_status).to_h },
-        format_for_display: ->(v) { HudUtility2026.veteran_status(v) },
+        format_for_display: ->(v) { HudHelper.util('2026').veteran_status(v) },
```
gigxz:
```suggestion
        format_for_display: ->(v) { HudHelper.util.veteran_status(v) },
```

No need to call out 2026 here

drivers/hmis/app/models/hmis/ce/match/expression/client_field_map.rb
```diff
@@ -118,6 +118,6 @@ def project_types_query(clients, scope, project_association)
     # display helpers
     def helpers = ApplicationController.helpers
     def format_days(days) = days.nil? ? nil : helpers.pluralize(days, 'day')
-    def format_project_types(project_type_ids) = project_type_ids.uniq.map { |t| HudUtility2026.project_type(t) }
+    def format_project_types(project_type_ids) = project_type_ids.uniq.map { |t| HudHelper.util('2026').project_type(t) }
```
gigxz, line 121:
```suggestion
    def format_project_types(project_type_ids) = project_type_ids.uniq.map { |t| HudHelper.util.project_type(t) }
```
Unless I'm missing some reason to specify 2026 here

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
```diff
@@ -62,13 +62,11 @@ def determine_event_type(capture_to_sentry: true)
 
       # Fall back to determining the event type based on the referral target project
       project = referral.target_project
-      event_type = HudUtility2026.project_to_ce_event_type(project)
-
+      event_type = HudHelper.util('2026').project_to_ce_event_type(project)
```
gigxz, line 65: I guess "2026" is called out since this helper exists in HudUtility2026 but not HudUtility2024. However the logic is not 2026-specific

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -317,18 +317,18 @@ def self.apply_filters(input)
   end
 
   # fix these so they use DATA_NOT_COLLECTED And the other standard names
-  use_enum(:gender_enum_map, ::HudUtility2026.genders) do |hash|
+  use_enum(:gender_enum_map, ::HudHelper.util('2026').genders) do |hash|
```
gigxz: Also don't think 2026 needs to be called out in this file

drivers/hmis/spec/models/hmis/ce/match/expression/client_field_map_spec.rb
```diff
@@ -183,12 +183,12 @@ def destination_client_for(hmis_client)
   describe '#format_for_display' do
     it 'formats veteran status using HUD utility' do
       formatted = field_map.format_for_display('veteran_status', 1)
-      expect(formatted).to eq(HudUtility2026.veteran_status(1))
+      expect(formatted).to eq(HudHelper.util('2026').veteran_status(1))
```
gigxz:
```suggestion
      expect(formatted).to eq(HudHelper.util.veteran_status(1))
```

drivers/hmis/spec/models/hmis/ce/match/expression/client_field_map_spec.rb
```diff
@@ -183,12 +183,12 @@ def destination_client_for(hmis_client)
...
     end
 
     it 'formats project types using HUD utility' do
       formatted = field_map.format_for_display('open_enrollment_project_types', [2, 3])
-      expected = [HudUtility2026.project_type(2), HudUtility2026.project_type(3)]
+      expected = [HudHelper.util('2026').project_type(2), HudHelper.util('2026').project_type(3)]
```
gigxz:
```suggestion
      expected = [HudHelper.util.project_type(2), HudHelper.util.project_type(3)]
```

gigxz (approved):

dtgreiner (approved):

drivers/hmis/app/models/hmis/ce/referral_ce_event_manager.rb
eanders: I'm going to leave that for now as it might break pre-10/1 if we don't support that in 2024, we can revisit on the next round.

eanders:
@ttoomey if you have time to review the final changes, I think this is ready (pending tests passing.)

eanders:
@gigxz, any thoughts on if these failures are new to this branch, or related to these changes?
https://github.com/greenriver/hmis-warehouse/actions/runs/17931549725/job/50989579129?pr=5756
https://github.com/greenriver/hmis-warehouse/actions/runs/17931550165/job/50989573170?pr=5756

I'm guessing not, but I'm unfamiliar with the code.

ttoomey (approved):
> @ttoomey if you have time to review the final changes, I think this is ready (pending tests passing.)

Looks great, thank you!!

gigxz:
> @gigxz, any thoughts on if these failures are new to this branch, or related to these changes? https://github.com/greenriver/hmis-warehouse/actions/runs/17931549725/job/50989579129?pr=5756 https://github.com/greenriver/hmis-warehouse/actions/runs/17931550165/job/50989573170?pr=5756
> 
> I'm guessing not, but I'm unfamiliar with the code.

THese look new but looks like an issue with the living situation list that you addressed already, thanks!

PR #5757, opened 2025-09-11
Update workflow to ensure enrollment exists

martha:
## _Merging this PR_
- targets `release-180`, should be re-deployed after PR is merged

## Description
- Update the Workflow with 3 separate Enrollment steps. See ticket and comments for rationale
- Propose a new referral step side effect, "raise_if_no_enrollment", that prevents users from "Confirm Success" on a referral whose enrollment has been deleted, either manually or by the workflow. (Not discussed, can revert https://github.com/greenriver/hmis-warehouse/pull/5757/commits/67905fc987981dcbb83c5c9d399268ce3726bcc5) 

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
+++ b/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
@@ -47,8 +47,6 @@ module Hmis::Ce
         reversible = false
       when 'delete_wip_enrollment'
         referral_enroller.delete_wip_enrollment(message)
-      when 'raise_if_no_enrollment'
-        referral_enroller.raise_if_no_enrollment(message)
       when 'set_move_in_date'
         # Can be triggered on the same step as create_enrollment, or a later step
         referral_enroller.set_move_in_date(message)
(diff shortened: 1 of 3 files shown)


drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -47,6 +47,8 @@ def call(message)
         reversible = false
       when 'delete_wip_enrollment'
         referral_enroller.delete_wip_enrollment(message)
+      when 'raise_if_no_enrollment'
+        referral_enroller.raise_if_no_enrollment(message)
```
gigxz: I think at this stage I'd feel more comfortable merging without https://github.com/greenriver/hmis-warehouse/commit/67905fc987981dcbb83c5c9d399268ce3726bcc5, so we're not introducing a new pattern of messages-to-raise-errors (which is unfixable by the user in this case, if I understand correctly). It can be up to the CE staff to ensure that they only perform "confirm success" if it was successful.

gigxz (commented):
Glad this works! Looks good to me but would prefer to revert the new message for an error, to keep it simple for mvp

PR #5765, opened 2025-09-12
8123 hmis file urls

ttoomey:
## _Merging this PR_

## Description
- Added authenticated endpoint `GET /hmis/clients/:client_id/files/:id` that authorizes access and redirects to a short‑lived ActiveStorage URL
- GraphQL `File.url` now returns a same‑origin path (`hmis_client_file_path`) instead of a direct service URL

Note, I have difficulty testing this locally in development; my plan is to deploy to QA for testing with s3

## Type of change
New feature


drivers/hmis/app/controllers/hmis/client_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      return head(:not_found) if redacted?(file) || !file.client_file.attached?
+
+      redirect_to object_url(file), allow_other_host: true
+    end
+
+    protected
+
+    def redacted?(file)
+      return false unless file.confidential
+      return false if file.user_id == current_hmis_user.id && current_hmis_user.can_manage_own_client_files_for?(file)
```
eanders:
```suggestion
      # This file is not confidential, therefore not redacted
      return false unless file.confidential
      # This user uploaded this file and the user still has access to files they control, therefore not redacted
      return false if file.user_id == current_hmis_user.id && current_hmis_user.can_manage_own_client_files_for?(file)
```

drivers/hmis/app/controllers/hmis/client_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+
+      !current_hmis_user.can_view_any_confidential_client_files_for?(file)
```
eanders, line 30:
```suggestion
      # Can user can see any confidential files in this data source? Redact if not
      !current_hmis_user.can_view_any_confidential_client_files_for?(file)
```

drivers/hmis/app/controllers/hmis/client_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    def object_url(file)
```
eanders, line 34:
```suggestion
    # returns the direct url to the active storage file with a short expiration 
    def object_url(file)
```

drivers/hmis/app/controllers/hmis/client_files_controller.rb
```diff
@@ -0,0 +1,56 @@
...
+      dispositions = ['inline', 'attachment']
+      params[:disposition].presence_in(dispositions) || dispositions.first
+    end
+
+    def object_file_name(file)
+      result = file.name.presence || "file-#{file.id}"
+      ActiveStorage::Filename.new(result).sanitized
+    end
+
+    def load_authorized_client_file
```
eanders, line 56:
```suggestion
    # Checks permission for both the client and the requested file
    def load_authorized_client_file
```

drivers/hmis/spec/requests/hmis/client_files_controller_spec.rb
```diff
@@ -0,0 +1,124 @@
...
+      hmis_user,
+      ds1,
+      with_permission: [
+        :can_view_clients,
+        :can_view_any_nonconfidential_client_files,
+      ],
+    )
+  end
+
+  # let!(:client) { create(:hmis_hud_client, data_source: ds1, user: u1) }
```
eanders: Maybe we don't need this because c1 is created as part of the `hmis base setup`?

eanders (approved):
Thanks @ttoomey.  I dropped some comment suggestions, but I wonder if there's also some other additional description that would be useful.  
The functionality looks good to me.

drivers/hmis/app/controllers/hmis/client_files_controller.rb
```diff
@@ -0,0 +1,61 @@
...
+# frozen_string_literal: true
+
+module Hmis
+  class ClientFilesController < Hmis::BaseController
+    before_action :attach_data_source_id
+
+    def show
+      file = load_authorized_client_file
+      return head(:not_found) if redacted?(file) || !file.client_file.attached?
+
```
gigxz, line 16: Can we insert an [activity log](https://github.com/greenriver/hmis-warehouse/blob/ce1d98b018136c5b5d54e1c107552beea685d528/drivers/hmis/app/controllers/hmis/graphql_controller.rb#L128-L147) here? Currently we only create activity logs for GraphQL requests, so we'd need to do something a bit different for the operation_name (maybe store it as `hmis_client_file_path` or `ClientFilesController#show`?).

This would allow us to audit on which files have been viewed or downloaded. I think it's worth adding, given that we have permissions around file visibility/confidential files. (I.e. you may have permission to resolve the File object, but now preview/download the file content).

Additionally, we may want to consider adding field-level audit on the `url` field on GQL File, similar to how Client [logs if ssn or dob](https://github.com/greenriver/hmis-warehouse/blob/ce1d98b018136c5b5d54e1c107552beea685d528/drivers/hmis/app/graphql/types/hmis_schema/client.rb#L275-L280) are resolved. Although I would have to dig deeper to remember the functionality, here.. I think the intention would be to record whether the `url` was resolve as non-null value. It's not sufficient to record whether the `url` field was requested, because it may have been properly resolved as nil.

This may be scope creep - happy to pull out to another ticket for ensuring sufficient audit logging.

gigxz (approved):
Looks good to me! I tested locally with my local s3 setup, but agreed further testing when deployed to QA will be necessary. Feel free to deploy this branch to hmis_qa without pulling in `7853-hmis-hud-2026-breaking-changes`  (if you haven't already), we can redploy the fy26 branch after.

drivers/hmis/app/controllers/hmis/client_files_controller.rb
ttoomey: thanks @gigxz - I've added activity logs and captured the url on file resolver, conditional on value presence

config/environments/test.rb
```diff
@@ -50,6 +50,8 @@
 
   # Store uploaded files on the local file system in a temporary directory.
   config.active_storage.service = :test
+  # Re-enable ActiveStorage routes in test so Disk service URL helpers exist
+  config.active_storage.draw_routes = true
```
ttoomey, line 54: the test adapter (`config.active_storage.service = :test`) crashes if routes are disabled ("undefined method `rails_disk_service_url`").

gigxz (approved):

PR #5768, opened 2025-09-15
When importing units, make sure they have unit groups

martha:
## _Merging this PR_
- targets 179, since this is fixes an ongoing issue

## Description
When importing Units, they should be created with Unit Groups.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis/app/models/hmis/project_unit_type_mapping.rb
+++ b/drivers/hmis/app/models/hmis/project_unit_type_mapping.rb
@@ -17,32 +17,71 @@ class Hmis::ProjectUnitTypeMapping < Hmis::HmisBase
   scope :active, -> { where(active: true) }
 
   def self.freshen_project_units(user:, today: Date.current)
+    scope = preload(:project, :unit_type).order(:id).to_a
+    create_new_units(scope, user: user)
+    destroy_inactive_units(scope, today: today)
+  end
+
+  def self.create_new_units(scope, user:)
     # { [project_id, unit_type_id] => unit_count }
     unit_counts_by_project_and_unit_type_id = Hmis::Unit.group(:project_id, :unit_type_id).count
 
-    scope = preload(:project, :unit_type).order(:id).to_a
-    unit_attrs = scope.filter(&:active?).flat_map do |record|
-      unit_type = record.unit_type
+    records_to_import = scope.
+      filter(&:active?).
+      filter { |record| record.project.present? }. # could happen if project was deleted but ProjectUnitTypeMapping wasn't properly cleaned up
+      filter do |record|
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/models/hmis/project_unit_type_mapping.rb
```diff
@@ -17,32 +17,72 @@ class Hmis::ProjectUnitTypeMapping < Hmis::HmisBase
...
-    # If this ProjectID is already mapped to this UnitTypeID in our system, but it is marked as Active=N, then remove the mapping. At this point also remove any Units for this unit type. Raise if any of those Units are occupied
+    # Batch import units
+    Hmis::Unit.import!(units_to_create, validate: false)
+  end
+
+  def self.destroy_inactive_units(scope, today: Date.current)
+    # If this ProjectID is already mapped to this UnitTypeID in our system, but it is marked as Active=N, then remove the mapping.
+    # At this point also remove any Units for this unit type. Raise if any of those Units are occupied
     scope.filter(&:inactive?).each do |record|
       existing_units = record.project.units.where(unit_type: record.unit_type)
```
gigxz, line 91: Maybe add a TODO here for `#8157`? Since this will need to be changed to look up using the unit group-unit-type relationship, ie something like `record.project.unit_groups.where(unit_type: record.unit_type).each(&:destroy!)`

drivers/hmis/app/models/hmis/project_unit_type_mapping.rb
```diff
@@ -17,32 +17,72 @@ class Hmis::ProjectUnitTypeMapping < Hmis::HmisBase
   scope :active, -> { where(active: true) }
 
   def self.freshen_project_units(user:, today: Date.current)
+    scope = preload(:project, :unit_type).order(:id).to_a
+    create_new_units(scope, user: user)
+    destroy_inactive_units(scope, today: today)
```
gigxz:
```suggestion
    scope = preload(:project, :unit_type).order(:id)
    # Create initial Units for new Active Project Unit Type Mappings
    create_new_units(scope.active, user: user)
    # Destroy Units for Project Unit Type Mappings that are Inactive
    destroy_inactive_units(scope.inactive, today: today)
```

Some minor suggestions for readability-
- add comments
- pass only relevant scope (active vs inactive)
- pass scope instead of array- since the other functions accept a variable name `scope` I would expect it to be an activerecord scope not an array. Don't think it matters much either way for perf since it's a relatively small set

gigxz (approved):
Approving this into release-179 to patch issue that is appearing from nightly custom integration on 1 installation. Feels low-risk enough because it is specific to one environment

PR #5771, opened 2025-09-16
Implement quick fixes for post-merge CE crash prevention

martha:
## _Merging this PR_

## Description
See description, bug repro instructions, and how the fixes I've proposed here should behave in my comment on the linked ticket.

## Type of change
Bug fix


drivers/hmis/app/graphql/types/hmis_schema/ce_candidate.rb
```diff
@@ -34,6 +34,8 @@ def client_attributes
     end
 
     def enrollments # not for batch
+      return [] if source_clients.empty?
```
martha, line 42: This addresses problem 1 in my comment on the ticket https://github.com/open-path/Green-River/issues/8196#issuecomment-3282562603

app/models/grda_warehouse/tasks/client_cleanup.rb
martha, line 1: This addresses problem 2 in my comment on the ticket https://github.com/open-path/Green-River/issues/8196#issuecomment-3282562603

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -900,6 +902,18 @@ def remove_unused_warehouse_clients_processed
       GrdaWarehouse::HmisClient.where(client_id: @clients).delete_all
     end
 
+    private def clean_coordinated_entry_records
+      return unless @clients.any?
+      return if @dry_run
+      return unless HmisEnforcement.hmis_enabled?
```
ttoomey:
```suggestion
      return unless HmisEnforcement.hmis_enabled? && Hmis::Ce.configuration.enabled?
```

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -900,6 +902,18 @@ def remove_unused_warehouse_clients_processed
       GrdaWarehouse::HmisClient.where(client_id: @clients).delete_all
     end
 
+    private def clean_coordinated_entry_records
```
ttoomey, line 905: suggest adding tests to `client_cleanup_spec`
```ruby
describe 'clean_coordinated_entry_records' do
  let(:dry_run) { false }
  let(:cleanup) { GrdaWarehouse::Tasks::ClientCleanup.new(dry_run: dry_run) }
  let!(:destination_client) { create(:grda_warehouse_hud_client) }

  before do
    allow(HmisEnforcement).to receive(:hmis_enabled?).and_return(true)
    cleanup.instance_variable_set(:@clients, [destination_client.id])
  end

  it 'destroys CE ClientProxies for destination clients when HMIS is enabled' do
    Hmis::Ce::ClientProxy.create!(client: destination_client)

    expect {
      cleanup.send(:clean_coordinated_entry_records)
    }.to change {
      Hmis::Ce::ClientProxy.where(
        client_id: destination_client.id,
        client_type: GrdaWarehouse::Hud::Client.sti_name,
      ).count
    }.from(1).to(0)
  end

end
```

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -900,6 +902,18 @@ def remove_unused_warehouse_clients_processed
...
+      return unless @clients.any?
+      return if @dry_run
+      return unless HmisEnforcement.hmis_enabled?
+
+      # Clean up ClientProxies, and CE Candidates via cascading destroy.
+      Hmis::Ce::ClientProxy.where(
+        client_id: @clients, # Manual join to handle polymorphic relationship
+        client_type: GrdaWarehouse::Hud::Client.sti_name,
+      ).destroy_all
```
ttoomey: if there are many unused client ids, this could be expensive as it will run for each client_proxy with n+1 destroy match candidates and events. Perhaps that doesn't occur in practice but if it's a concern, we could use `delete_all` on the associations to avoid n+1

```diff
diff --git a/drivers/hmis/app/models/hmis/ce/client_proxy.rb b/drivers/hmis/app/models/hmis/ce/client_proxy.rb
index 0ce20f56f4..9f3c6a4fa0 100644
--- a/drivers/hmis/app/models/hmis/ce/client_proxy.rb
+++ b/drivers/hmis/app/models/hmis/ce/client_proxy.rb
@@ -18,8 +18,8 @@ module Hmis::Ce
     # For now, this is the GrdaWarehouse::Hud::Client representing the *destination* client.
     # In the future, we will add more client types (e.g. VSP)
     belongs_to :client, polymorphic: true, optional: false
-    has_many :ce_match_candidates, class_name: 'Hmis::Ce::Match::Candidate', foreign_key: :client_proxy_id, dependent: :destroy
-    has_many :ce_match_candidate_events, class_name: 'Hmis::Ce::Match::CandidateEvent', foreign_key: :client_proxy_id, dependent: :destroy
+    has_many :ce_match_candidates, class_name: 'Hmis::Ce::Match::Candidate', foreign_key: :client_proxy_id, dependent: :delete_all
+    has_many :ce_match_candidate_events, class_name: 'Hmis::Ce::Match::CandidateEvent', foreign_key: :client_proxy_id, dependent: :delete_all
 
     validates :client_id, presence: true, uniqueness: { scope: [:client_type] }
     validate :client_is_destination
```

ttoomey: * deleting the proxy also deletes any events for client proxy, potentially impacting historical data for the consolidated waitlist (if the destination client were to be restored)
* suggest logging how many records were deleted `log "Deleted #{deleted} CE client proxies"`

ttoomey (approved):
Thanks! left some suggestions

app/models/grda_warehouse/tasks/client_cleanup.rb
gigxz: added

app/models/grda_warehouse/tasks/client_cleanup.rb
gigxz: added

app/models/grda_warehouse/tasks/client_cleanup.rb
gigxz: @ttoomey I pushed another WIP suggestion which was to change to using `delete_all` instead of `destroy_all` here, and explictly delete only the Proxy and Candidate records: https://github.com/greenriver/hmis-warehouse/pull/5771/commits/6cf09872671c0a6e934bf1f01c4133571ce832ad
It's WIP because I didn't realize there was a foreign key constraint on candidate events, more on that below..

Using `delete_all` felt more consistent with the ClientCleanup job which seems to exclusively delete and not destroy. I worry that with introducing a `destroy`, we might unintentionally cause a performance issue later on if we were to add additional records that hang off ClientProxy with `dependent: :destroy`. I don't think we have solid monitoring for job performance of this task, so that might be hard to surface if it occurs.


Dealing with CandidateEvents in this scenario is an interesting question. I agree they are historical records, I'd prefer to treat as an audit log that doesn't get deleted, especially since we are exporting them. That feels like scope creep, some maybe we should stick to just deleting them here and then add a follow-up issue to discuss options. (e.g. add destination client id reference directly to ce_match_candidates table so we can continue to link it to a soft-deleted client? or soft-delete client proxy records?..)


Edit: updated to delete candidate events https://github.com/greenriver/hmis-warehouse/pull/5771/commits/2450bbebfb43a49171828142ce78da0657622ace, opened https://github.com/open-path/Green-River/issues/8261 to discuss concerns around deleting CE Match Candidate Events

ttoomey: This seems reasonable to me, we can figure out retention of CE events in another issue (some kind of soft-delete or additional reference to destination client as you suggest). Thanks!

ttoomey (approved):

PR #5772, opened 2025-09-16
Alert Sentry, don't raise, for unknown CE event type

martha:
## _Merging this PR_

## Description
Enable projects that don't correspond to a CE event type to reuse the same CE workflow.
* On `create_ce_event` message, report to Sentry instead of raising if unable to determine the CE Event Type.
* On `set_ce_event_result`,
  * If CE event type can't be determined, return without reporting to Sentry or raising
  * If CE Event Type _can_ be determined, but CE Event is missing, raise

Testing instructions
- Set up a project whose project type doesn't map to a CE Event (such as 4, Street Outreach) so that it accepts referrals using a workflow that does create ce events (such as Housing Referral Workflow V1).
- Step through the workflow for a client. The referral should be able to be completed successfully and nothing should raise. On the source enrollment, no CE event should be created.

## Type of change
New feature

### DIFF
--- a/drivers/hmis/spec/models/hmis/ce/referral_ce_event_manager_spec.rb
+++ b/drivers/hmis/spec/models/hmis/ce/referral_ce_event_manager_spec.rb
@@ -233,6 +233,20 @@ RSpec.describe Hmis::Ce::ReferralCeEventManager, type: :model do
         end.to raise_error(RuntimeError, /Referral does not have a source enrollment/)
       end
     end
+
+    context 'when no CE Event Type can be determined' do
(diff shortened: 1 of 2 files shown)


drivers/hmis/spec/models/hmis/ce/referral_ce_event_manager_spec.rb
```diff
@@ -247,39 +261,82 @@ def submit_current_step(submitted_values = {})
       end
     end
 
-    before(:each) do
```
martha, line 250: I think this diff is getting exaggerated by Github, these examples are just tabbed in but otherwise unchanged. (From here until line 301)

gigxz (approved):
looks good!

PR #5784, opened 2025-09-18
Rails 7.2

eanders:
## _Merging this PR_

## Description
* Basic upgrade of rails to 7.2
* Upgrades some gems that no longer worked with 7.2

Pivots back to using `foreign_key` for multi-column relations per:
```
irb(main):002> GrdaWarehouse::Hud::Client.last
/bundle/ruby/3.4.0/gems/activerecord-7.2.2.2/lib/active_record/reflection.rb:19:in 'Class#new': DEPRECATION WARNING: Setting `query_constraints:` option on `GrdaWarehouse::Hud::Client.has_one :most_recent_ce_assessment` is deprecated. To maintain current behavior, use the `foreign_key` option instead. (called from <class:Client> at /app/app/models/grda_warehouse/hud/client.rb:54) (ActiveSupport::DeprecationException)
```

The `commit_transaction_on_non_local_return` functionality has been removed in rails 7.2.  This means transactions that contained `return` or `break` previously rolled back, but now save silently.  In reviewing the code looking at files that contain `transaction` and `return` or `break` only https://github.com/greenriver/hmis-warehouse/blob/d79e07041e1549fd6880c203383ea6b2c9ecdb63/drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb#L45-L50 appears to contain the code where the behavior has changed.  

## Type of change
* Dependency update


eanders:
@ttoomey, I was only able to find one instance where were break or return from a transaction. 
https://github.com/greenriver/hmis-warehouse/blob/d79e07041e1549fd6880c203383ea6b2c9ecdb63/drivers/hmis/app/graphql/mutations/ce/create_direct_ce_referral.rb#L45-L50

Does that look like it needs adjustment to you?

eanders:
Actually, it looks like a recent change replaced that transaction with an advisory lock, so we are probably all set.

ttoomey:
Great, sounds like we don't need to worry about that transaction behavior. I'll take another look through the codebase when I get a sec to confirm. Thanks for looking!

eanders:
Closing in favor of https://github.com/greenriver/hmis-warehouse/pull/5906

PR #5785, opened 2025-09-18
Resolve limited household member info on referral

gigxz:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch
- use a merge-commit or rebase strategy for PRs targeting the stable branch

## Description
Issue: https://github.com/open-path/Green-River/issues/8217
Frontend PR; https://github.com/greenriver/hmis-frontend/pull/1276

Resolve household member names, IDs, relationship, and access information on on `CeReferralSourceEnrollment`. This is used to display this information on the Referral details panel. Client names are only displayed if current user already has access to view them.

## Type of change
Feature

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] If adding a new endpoint / exposing data in a new way, I have:
  - [x] ensured the API can't leak data from other data sources
  - [x] ensured this does not introduce N+1s
  - [x] ensured permissions and visibility checks are performed in the right places
- [x] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) 
- [x] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [x] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #5790, opened 2025-09-19
Patch: Resolve client id on referral household members

gigxz:
## _Merging this PR_

## Description
Follow-up to https://github.com/greenriver/hmis-warehouse/pull/5785 to fix a mistake. We need to resolve the Client ID for each household member, not just the enrollment ID.
https://github.com/greenriver/hmis-frontend/pull/1278

## Type of change
Bug fix


drivers/hmis/app/graphql/types/hmis_schema/ce_referral_source_household_member.rb
```diff
@@ -21,6 +22,10 @@ class HmisSchema::CeReferralSourceHouseholdMember < Types::BaseObject
       field :can_view_clients, Boolean, null: false
     end
 
+    def client_id
+      client&.id
```
ttoomey: safe navigation operator needed because client can be nil? The graphql field is non-null

ttoomey (approved):

PR #5791, opened 2025-09-21
8234-fy26-changes

ttoomey:
## _Merging this PR_

## Description

- Removed VA rules from `disability_table` and `disability_table_r4` so elements 4.05–4.10 are not required for VA-funded projects
- Generated comments for `hud_assessment_form_rules_2026.rb`.
- Set `"warn_if_empty" to false for: translation_assistance (C4) and Sexual Orientation (R3)

### To test:
- run `rails driver:hmis:seed_definitions` and restart the Rails server to load rule changes. ⚠️
- Mixed-funder behavior remains driven by non-VA rules; confirm expected behavior if a project has both VA and non-VA funders.
- Create an intake for a VA-funded project that does not meet any other requirement for showing disability. Disability table is not shown on intake.

## Type of change
New feature


drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2026.rb
```diff
@@ -7,6 +7,42 @@
 ###
 
 module HmisUtil
+  # Purpose
```
ttoomey, line 10: generated docs, would benefit from review

drivers/hmis/lib/form_data/tarrant_county/fragments/patches/sexual_orientation.json
```diff
@@ -6,6 +6,7 @@
         "type": "CHOICE",
         "link_id": "sexual_orientation",
         "text": "Sexual Orientation",
+        "warn_if_empty": false,
```
gigxz: This is not a HUD field so doesn't need a change. It is harmless since the field defaults to false, but recommend removing the change since it's not needed.

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2026.rb
gigxz: Docs look good to me. The last note about restarting the local Rails server is something I don't need to do, but maybe that's a local setup thing.

gigxz (commented):
Looks good to me. I think we should target this against  `7853-hmis-hud-2026-breaking-changes` because release-181 will go to production on 9/25, and we shouldn't stop collecting disability for va until 10/1.

drivers/hmis/lib/hmis_util/hud_assessment_form_rules_2026.rb
ttoomey: yeah, could be something funky with my local env. It stumped me for a while that the changes weren't getting picked up

drivers/hmis/lib/form_data/tarrant_county/fragments/patches/sexual_orientation.json
ttoomey: This was part of the [issue](https://github.com/open-path/Green-River/issues/8234) "Remove "Warn if empty" behavior for Sexual Orientation field" - sounds like that was incorrect though. I can back it out

gigxz: That change referred to HUD Field R3 which you made in file `drivers/hmis/lib/form_data/default/fragments/r3_sexual_orientation.json`.

This file here is a custom installation-specific patch to add a migrated-in field on the client that _also_ collects Sexual Orientation, but it is a completely unrelated custom field. Sorry for the confusion!

ttoomey: got it, thanks!

ttoomey:
>  I think we should target this against 7853-hmis-hud-2026-breaking-changes because release-181 will go to production on 9/25, and we shouldn't stop collecting disability for va until 10/1

Good catch, thanks

gigxz (approved):

gigxz:
@ttoomey it looks like there are some test failures here, can you take a look?

gigxz:
@ttoomey so it looks like the issue was that some tests previously used projects set up with funder `20` which is HUD/VASH, and expected the full disability questions to be present. Now that disability is no longer required for HUD/VASH, those questions were missing. I pushed a commit to switch to using a different funder (`'HUD: CoC - Rapid Re-Housing' (3)`), which resolves it. I was able to run the full system test suite locally. commit: https://github.com/greenriver/hmis-warehouse/pull/5791/commits/462050422a42b382fe7b27183b628c3292644ec5

PR #5796, opened 2025-09-22
Support flag to lock edits on a form

martha:
## _Merging this PR_

## Description
- Add column `admin_editable_only` to form definition
- Create `FormDefinitionPolicy` and use it in the publish/update mutations
- Resolve form definition `access` field on the graphql schema type

## Type of change
New feature

  - [x] ensured this does not introduce N+1s - currently, the policy is only checked on publish/update mutations. In the future if we resolve the access object on several form definitions at once, we should confirm with a unit test that no n+1s are happening.
- [x] I have provided testing instructions in this PR or the related issue (or not applicable) - see test instructions on frontend pR: https://github.com/greenriver/hmis-frontend/pull/1284


drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
```diff
@@ -0,0 +1,24 @@
...
+  def can_manage_form?
+    return false if resource.managed_in_version_control?
+
+    # Only super-admins (GR users) can manage forms that are admin-editable-only
+    return false if resource.admin_editable_only? && !context.user.can_administrate_config?
+
+    context.user.can_manage_forms_for_role?(resource.role)
+  end
+
+  # TODO: incorporate other policies and permissions. For example, can_index? policy should be based on permission can_configure_data_collection?
```
martha, line 29: I definitely think the form definition policy should become more comprehensive in the future, but for now I've taken the route of minimizing code churn and leaving a todo for later.

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -16,11 +18,10 @@ def resolve(id:, input:)
       raise 'not found' unless definition
       raise 'not allowed' if definition.managed_in_version_control?
```
ttoomey: This is also checked in the policy. Maybe it can be removed from the mutation?

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -16,11 +18,10 @@ def resolve(id:, input:)
...
 
-      # The UI currently does allow changing the form role, so we make this permission check twice:
-      # - to confirm the user has permission to manage the role from the input (in case it's changing)
-      # - to confirm the user has permission to manage forms for the original role (from the definition)
+      access_denied! unless policy_for(definition, policy_type: :form_definition).can_manage_form?
+
+      # If the form role has been changed, confirm the user also has permission for the new role.
       access_denied! if input.role && !current_user.can_manage_forms_for_role?(input.role)
```
ttoomey: perhaps this additional check could be rolled up to the policy:
```ruby 
   class Hmis::AuthPolicies::FormDefinitionPolicy < Hmis::AuthPolicies::BasePolicy
      def can_update_form?(new_role: nil)
        return false unless can_manage_form?
        return true if new_role.blank?
        user.can_manage_forms_for_role?(new_role)
      end
    end
```
then
`access_denied! unless policy_for(definition, policy_type: :form_definition)
      .can_update_form?(new_role: input.role)`

db/warehouse/migrate/20250921222539_add_admin_editable_only_to_form_definition.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddAdminEditableOnlyToFormDefinition < ActiveRecord::Migration[7.0]
+  def change
+    add_column :hmis_form_definitions, :admin_editable_only, :boolean, default: false
```
ttoomey: probably doesn't matter but maybe add null:false
```suggestion
    add_column :hmis_form_definitions, :admin_editable_only, :boolean, default: false, null: false
```

ttoomey (approved):
looks good to me, thanks!

martha:
I'll retarget this to release-184 (when available) and wait to merge until the corresponding frontend PR is ready to go. https://github.com/greenriver/hmis-frontend/pull/1284

drivers/hmis/app/graphql/mutations/update_form_definition.rb
```diff
@@ -14,13 +16,8 @@ class UpdateFormDefinition < CleanBaseMutation
...
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition
-      raise 'not allowed' if definition.managed_in_version_control?
 
-      # The UI currently does allow changing the form role, so we make this permission check twice:
-      # - to confirm the user has permission to manage the role from the input (in case it's changing)
-      # - to confirm the user has permission to manage forms for the original role (from the definition)
-      access_denied! if input.role && !current_user.can_manage_forms_for_role?(input.role)
-      access_denied! unless current_user.can_manage_forms_for_role?(definition.role)
+      access_denied! unless policy_for(definition, policy_type: :form_definition).can_update_form?(new_role: input.role)
```
gigxz: Hmm. In practice, I don't think the "change form role" functionality is really needed or used. I think I would prefer to disallow changing role, as opposed to encoding special logic for it in the new policy. Changing a form's `role` could lead to lots of weird issues, like field mappings inadvertently changing to point to a different record.

gigxz: Here's a diff I tested, that updates the behavior so you can no longer change the Role after a form is created.

This lets us get rid of the distinction between "updating" and "managing" a form on the policy.

I tested from this backend branch (Rebased to 184 to fix todoordie) and release-184 on the frontend.

```diff
diff --git a/drivers/hmis/app/graphql/mutations/update_form_definition.rb b/drivers/hmis/app/graphql/mutations/update_form_definition.rb
index d33aa2ba76..e93f3865c1 100644
--- a/drivers/hmis/app/graphql/mutations/update_form_definition.rb
+++ b/drivers/hmis/app/graphql/mutations/update_form_definition.rb
@@ -17,10 +17,11 @@ module Mutations
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition

-      access_denied! unless policy_for(definition, policy_type: :form_definition).can_update_form?(new_role: input.role)
+      access_denied! unless policy_for(definition, policy_type: :form_definition).can_manage_form?

       raise 'only allowed to modify draft forms' unless definition.draft?
       raise 'not allowed to change identifier' if input.identifier.present? && input.identifier != definition.identifier
+      raise 'not allowed to change role' if input.role.present? && input.role != definition.role

       # This mutation can be used to update the definition or the title/role, which is why definition is optional.
       definition.assign_attributes(**input.to_attributes) unless input.to_attributes.blank?
diff --git a/drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb b/drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
index b66b0a9411..bf6ddb3978 100644
--- a/drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
+++ b/drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
@@ -16,14 +16,6 @@ class Hmis::AuthPolicies::FormDefinitionPolicy < Hmis::AuthPolicies::BasePolicy
     context.user.can_manage_forms_for_role?(resource.role)
   end

-  def can_update_form?(new_role: nil)
-    # If the form role has been changed, confirm the user also has permission for the new role.
-    return false unless can_manage_form?
-    return true if new_role.blank?
-
-    user.can_manage_forms_for_role?(new_role)
-  end
-
   # TODO: incorporate other policies and permissions. For example, can_index? policy should be based on permission can_configure_data_collection?

   protected
diff --git a/drivers/hmis/lib/form_data/static/form_definition.json b/drivers/hmis/lib/form_data/static/form_definition.json
index 777c9e83a2..638c36d2a7 100644
--- a/drivers/hmis/lib/form_data/static/form_definition.json
+++ b/drivers/hmis/lib/form_data/static/form_definition.json
@@ -9,7 +9,16 @@
       },
       "pick_list_reference": "FORM_TYPES",
       "required": true,
-      "component": "DROPDOWN"
+      "component": "DROPDOWN",
+      "_comment": "Only shown when creating a new FormDefinition",
+      "enable_behavior": "ALL",
+      "enable_when": [
+        {
+          "question": "definitionId",
+          "operator": "EXISTS",
+          "answer_boolean": false
+        }
+      ]
     },
     {
       "type": "STRING",
```

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
gigxz: Sounds good! One thing I probably would suggest is pulling out the "publish" action, so it's clear from looking at the policy that it allows publishing. In the future we may want to make permissions more granular to limit access to publish vs other management actions. That way the PublishFormDefinition can ask the policy specifically if this user can _publish_, and it doesn't need to care about the internals of how that is implemented.

```ruby
  def can_publish_form?
    # Currently same as can_manage_form?, but may diverge in the future
    can_manage_form?
  end
```

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+# frozen_string_literal: true
+
+class Hmis::AuthPolicies::FormDefinitionPolicy < Hmis::AuthPolicies::BasePolicy
+  def can_manage_form?
+    return false if resource.managed_in_version_control?
+
+    # Only super-admins (GR users) can manage forms that are admin-editable-only
+    return false if resource.admin_editable_only? && !context.user.can_administrate_config?
+
+    context.user.can_manage_forms_for_role?(resource.role)
```
gigxz, line 16: It would be great at some point to move this method [out of this concern](https://github.com/greenriver/hmis-warehouse/blob/44e3b0b218707dab18dd8863aefe345fe037d75f/app/models/concerns/user_permissions.rb#L229-L235) and into this policy. Looks like there are a few other mutations we'd need to update in order to achieve that. I opened a subticket in this epic https://github.com/open-path/Green-River/issues/8329 for expanding usage of this policy!

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+###
+
+# frozen_string_literal: true
+
+class Hmis::AuthPolicies::FormDefinitionPolicy < Hmis::AuthPolicies::BasePolicy
+  def can_manage_form?
+    return false if resource.managed_in_version_control?
+
+    # Only super-admins (GR users) can manage forms that are admin-editable-only
+    return false if resource.admin_editable_only? && !context.user.can_administrate_config?
```
gigxz: I would suggest leaving out the `(GR users)` part of the comment to avoid future confusion, since there's not actually a constraint in place there. We happen to only have this enabled for GR users currently (as far as I know) but that's not something inherent in the permission

gigxz (approved):

PR #5799, opened 2025-09-23
1850-delayed-job-pruning

ttoomey:
## _Merging this PR_

## Description

Adds rake task `delayed_job:prune` to clean stale Delayed::Job locks held by non-existent K8s pods.

## Type of change
New feature


eanders (approved):
Thanks @ttoomey, this looks like it should help tremendously.

lib/tasks/delayed_job.rake
```diff
@@ -0,0 +1,92 @@
...
+    candidates = scope.pluck(:id, :locked_by, :locked_at)
+
+    # Reduce Kubernetes API load by fetching live pod names once; autodetect namespace when unset
+    client = K8s::Client.in_cluster_config
+    ns = ENV['K8S_NAMESPACE']
+    if ns.blank?
+      begin
+        ns = File.read('/var/run/secrets/kubernetes.io/serviceaccount/namespace').to_s.strip
+      rescue StandardError
+        ns = 'default'
```
yakloinsteak: If we get to this point, probably best to throw an exception.

lib/tasks/delayed_job.rake
```diff
@@ -0,0 +1,92 @@
...
+
+    # Task-scoped logging; formatting/metadata handled by Rails logger
+    log = ->(msg) { Rails.logger.info("[delayed_job:prune] #{msg}") }
+
+    log.call("Starting prune ns=#{ns} action=#{action} queue=#{queue.presence || '*'} stale_minutes=#{stale}")
+
+    acted = 0
+    candidates.each do |id, locked_by, locked_at|
+      pod = extract_pod_name.call(locked_by)
+      next if pod.blank?
```
yakloinsteak: Can we log something if the pod is blank? Should that ever be expected to happen?

lib/tasks/delayed_job.rake
```diff
@@ -0,0 +1,92 @@
...
+      next if names.include?(pod)
+
+      # Concurrency guard: match on locked_by and the exact locked_at we observed, and ensure staleness
+      base = Delayed::Job.where(id: id, failed_at: nil, locked_by: locked_by).
+        where('locked_at = ? AND locked_at < ?', locked_at, cutoff)
+
+      case action
+      when 'unlock'
+        affected = base.update_all(locked_by: nil, locked_at: nil, run_at: now)
+        acted += 1 if affected == 1
```
yakloinsteak: Should we add the affected value instead? I think I gather this is always 1 or 0, but the code makes me wonder what happens when it's more than 1. Maybe log something if it's zero to clarify?

lib/tasks/delayed_job.rake
```diff
@@ -0,0 +1,92 @@
...
+      when 'fail'
+        msg = "Pruned at #{now.utc.iso8601}: pod '#{pod}' not found; locked_at=#{locked_at&.utc&.iso8601}"
+        affected = base.update_all(failed_at: now, last_error: msg)
+        acted += 1 if affected == 1
```
yakloinsteak: Similar comment as above. I guess this is sort of nit-picky. Feel free to pass on these two comments of mine.

yakloinsteak (approved):
Looks great to me. Sorry it took a while for me to get to this.

lib/tasks/delayed_job.rake
ttoomey: good idea, yes it seems like we should always be able to find pod

lib/tasks/delayed_job.rake
ttoomey: yep, good thought

ttoomey:
Thanks! I've made some updates. Also added an advisory lock for safety

lib/tasks/delayed_job.rake
```diff
@@ -0,0 +1,101 @@
...
+        case action
+        when 'unlock'
+          affected = base.update_all(locked_by: nil, locked_at: nil, run_at: now)
+        when 'fail'
+          msg = "Pruned at #{now.utc.iso8601}: pod '#{pod}' not found; locked_at=#{locked_at&.utc&.iso8601}"
+          affected = base.update_all(failed_at: now, last_error: msg)
+        else
+          msg = "Unsupported action #{action.inspect}; aborting"
+          log.call(msg)
+          exit(1)
```
eanders: Should this notify sentry or raise?  We're in a loop, so maybe we should notify sentry and next instead of exit?

eanders (approved):
Thanks @ttoomey, looks good to me, just one question about the early exit.  I assume it would just pickup on the next run, but maybe we should alert, unless you think that will be a common scenario and it's ok.

lib/tasks/delayed_job.rake
ttoomey: well it would have to be invoked with a bad argument so we definitely shouldn't continue; I think exit with non-zero code is what should happen. I could see an argument for sending to sentry as I guess we can't rely the caller to do anything reasonable if it fails.

PR #5808, opened 2025-09-25
8261 Soft-delete CE ClientProxies to preserve historical events

ttoomey:
## _Merging this PR_

## Description
- We were losing historical CE Match Candidate Events when destination clients were removed because `ClientProxy` was hard-deleted.
- These events are intended to be a durable audit log.

### What changed
- Added soft-delete to `Hmis::Ce::ClientProxy` via `acts_as_paranoid`
- Updated cleanup task to soft-delete `ClientProxy` records and delete only CE Candidates; Candidate Events are preserved.
- Adjusted associations to avoid cascading deletes of Candidate Events.
- Updated bulk import to “revive” soft-deleted proxies instead of creating new ones.

### Validation and Indexing Decisions
- Database unique index: did not add a partial unique index (`WHERE deleted_at IS NULL`). The existing unique index on `(client_type, client_id)` ensures a single row across active and soft-deleted states. Rows can be revived via an upsert. 

## Type of change


db/warehouse/migrate/20250923120000_add_deleted_at_to_ce_client_proxies.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class AddDeletedAtToCeClientProxies < ActiveRecord::Migration[7.1]
+  def change
+    add_column :ce_client_proxies, :deleted_at, :datetime
```
ttoomey, line 11: note, we chose to not include deleted_at in the unique index, this makes it convenient to restore the proxy records with an upsert

drivers/hmis/app/models/hmis/ce/match/internal/candidate_repository.rb
```diff
@@ -18,10 +18,18 @@ def import_proxies(clients, timestamp:)
...
         }
       end
 
-      result = Hmis::Ce::ClientProxy.import(values, on_duplicate_key_ignore: true)
+      result = Hmis::Ce::ClientProxy.import(
+        values,
+        on_duplicate_key_update: {
+          conflict_target: [:client_type, :client_id],
+          columns: [:deleted_at, :updated_at],
+          condition: 'ce_client_proxies.deleted_at IS NOT NULL',
```
ttoomey, line 30: The update happens only if the existing row is soft-deleted

- If no row exists: a new proxy is inserted.
- If a soft-deleted row exists: it is “revived” by setting deleted_at to NULL (from the insert values) and updating updated_at.
- If an active row exists: the WHERE condition fails, so no UPDATE occurs (the upsert degrades to DO NOTHING), leaving updated_at unchanged.

gigxz (approved):
looks great!

PR #5811, opened 2025-09-25
Clear step assignment cache when step is submitted

martha:
## _Merging this PR_

## Description
How to repro:
- Impersonate a user who has "can perform own referral steps" permission. (For example, a Provider)
- Complete a step that will enable another step that's also assigned to this user. (For example, in Housing Workflow V1, completing the Provider Outcome step enables the Change Provider Outcome step)
- Before: The step is present, but the button to start it is hidden. (Refresh the page and it appears)
- Why: Step assignments are cached for permission checking. The newly available step does now have an assignment for the current user, but the cache hasn't been updated. So the graphql step is returned from the mutation with contradictory info, something like:
`{ participants: thisUser, access: { canPerform: false } }`
- Fix: Expose a method on the user policy context to force invalidate the cache. This allows us to keep the caching in general to avoid n+1s, but in the specific case where the user performs an action that might change step assignment, we can reload the cache before returning from the mutation.

## Type of change
Bug fix


drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
```diff
@@ -23,6 +23,12 @@ def initialize(user)
       assigned_referral_steps.pluck(:id).to_set
     end
 
+    # Clear memery cache when step assignments may have changed during a request (e.g., after completing a step in a mutation).
+    def clear_cache!
+      clear_memery_cache!
+      self
+    end
```
gigxz, line 29: Why does this return `self`?

drivers/hmis/spec/requests/hmis/ce/ce_referral_spec.rb
```diff
@@ -541,13 +541,16 @@
...
+
+            referral.workflow_engine.start_workflow!(user: hmis_user)
           end
 
           it 'still does not cause n+1' do
             expect do
               _, result = post_graphql(**variables) { query }
               steps = result.dig('data', 'ceReferral', 'steps')
-              expect(steps.map { |step| step.dig('access', 'canPerformStep') }).to all(be false)
+              expect(steps.map { |step| step.dig('access', 'canPerformStep') }).to all(be true)
```
gigxz, line 553: So this is expected to be `true` now because the `hmis_user` is now assigned to the swimlane for the open steps, is that right?

drivers/hmis/spec/requests/hmis/ce/submit_ce_referral_step_spec.rb
```diff
@@ -165,6 +179,23 @@ def build_input(contact_date: nil, client_accepted: nil)
...
           end
         end
+
+        context 'when submitting this step enables another step that the user is participating in' do
+          before do
+            step.assignments.create!(user: hmis_user) # Assign the user to the current task so they have permission
+            referral.participants.create(swimlane: provider_swimlane, user: hmis_user) # Assign the current user as a participant on the next task's swimlane
+          end
+
+          it 'submits the step' do
```
gigxz: Suggest updating test name to describe the thing being tested here (that isn't tested in other cases?)
```suggestion
          it 'resolves the next available step' do
```

drivers/hmis/spec/requests/hmis/ce/submit_ce_referral_step_spec.rb
```diff
@@ -165,6 +179,23 @@ def build_input(contact_date: nil, client_accepted: nil)
...
+            response, result = post_graphql(**variables) { mutation }
+            expect(response.status).to eq(200), result.inspect
+            next_step = result.dig('data', 'submitCeReferralStep', 'referral', 'steps', 1)
```
gigxz: Maybe explicitly assert on expectation that there are 2 steps expected?

gigxz (approved):

drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_assignment_loader.rb
martha: Cursor generated this, I assume to enable chaining method calls, but we're not using that. I'll remove it! 🙃

drivers/hmis/spec/requests/hmis/ce/ce_referral_spec.rb
martha: Yes, it's due to the added participant on line 544. I made this change while troubleshooting, but it's still testing pretty much the same thing -- the change could be reverted if you prefer.

gigxz: Ok just wanted to make sure I understood the change, thanks!

ttoomey (approved):
Looks great, thanks!

martha:
I've retargeted to release-183, but before I merge, for timing reasons should it go in release-182 instead? I'm on the fence

gigxz:
Lets put this in release-182 which will go out on 10/2

PR #5818, opened 2025-09-26
Increment ChangeMarker when associated record is deleted

martha:
## _Merging this PR_

## Description
Bug repro:
- Find a client on a waitlist in a Unit Group where the eligibility depends on the client's open enrollments. I.e., the unit group has a rule such as "Must not have open Enrollment at PSH/RRH/TH"
- Create a WIP enrollment for that client in one of the project types referenced by the rule. Their eligibility will update. (ChangeMarker is incremented)
- Delete the WIP enrollment. The ChangeMarker does not get incremented, so their eligibility doesn't update until the next full refresh.

Why:
- Poking into the [`paranoia` gem source](https://github.com/rubysherpas/paranoia/blob/core/lib/paranoia.rb#L115) I learned that calling `destroy!` on a paranoid model calls `update_columns` under the hood, which bypasses ActiveRecords `after_save` hook. So the `after_save` hook was getting called on create and update, but not on destroy.
- Solution: Call the same `mark_client_as_dirty` in an `after_destroy` hook.
- This issue also impacts other records like CustomAssessment and Exit. The fix is on the shared `MarkClientAsDirtyBehavior` concern, so it should fix all of them. I haven't tested them all locally, only `Enrollment` and `CustomAssessment`.

## Type of change
Bug fix

### DIFF
--- a/.rubocop_todo.yml
+++ b/.rubocop_todo.yml
@@ -3469,7 +3469,6 @@ Style/FrozenStringLiteralComment:
     - 'drivers/hmis/spec/jobs/activity_log_processor_job_spec.rb'
     - 'drivers/hmis/spec/jobs/assessment_questions_job_spec.rb'
     - 'drivers/hmis/spec/jobs/merge_clients_job_perf_spec.rb'
-    - 'drivers/hmis/spec/jobs/merge_clients_job_spec.rb'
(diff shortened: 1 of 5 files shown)


ttoomey (approved):
looks good!

martha:
Investigating test failures 👀

PR #5819, opened 2025-09-26
8156-ssn-v4

ttoomey:
## _Merging this PR_

## Description
Implement improved SSN selection/normalization that prevents selecting lower-quality SSNs. This is not yet live but will supports an impact assessment of the proposed change via the `grda_warehouse:ssn:audit` rake task

Tun taks on to gauge churn

```sh
# prefer oldest
rails grda_warehouse:ssn:audit[true] > ssn_discrepancies_oldest_first.csv
# prefer newest
rails grda_warehouse:ssn:audit[false] > ssn_discrepancies_newest_first.csv
```

## Type of change
Investigation ahead of bug fix


ttoomey:
Does this approach seem okay? If so I'll add some more robust tests

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -464,22 +464,48 @@ def choose_best_pronouns dest_attr, source_clients
       dest_attr
     end
 
+    # Get the best SSN (has value and quality is full or partial, oldest breaks the tie)
```
eanders: I _think_ we should be using newest to break the tie.  If we use oldest, there is no way to correct bad data coming from an old location that can no longer be updated.

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -464,22 +464,48 @@ def choose_best_pronouns dest_attr, source_clients
...
+    # note, source clients might be just an array of hashes
+    SSN_DQS = [1, 2, 8, 9, 99].to_set.freeze
```
eanders: Can we lean on our friend `HudHelper`?  `HudHelper.util.ssn_data_quality_options.keys`

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -464,22 +464,48 @@ def choose_best_pronouns dest_attr, source_clients
...
-      non_blank_ssn = source_clients.select { |sc| sc[:SSN].present? }
-      if non_blank_ssn.any?
-        best = nil
-        [1, 2, 8, 9, 99].each do |dq_value|
-          best = non_blank_ssn.select { |sc| sc[:SSNDataQuality] == dq_value }.min_by { |sc| sc[:DateCreated] }
-          break if best.present?
-        end
-        best = non_blank_ssn.min_by { |sc| sc[:DateCreated] } if best.blank?
+      items = source_clients.
+        filter { |sc| sc[:SSNDataQuality].in?(SSN_DQS) }.
```
eanders: Not sure we want to throw out any source clients where we are missing SSNDataQuality.  Historically, these data quality fields have been all over the map.  Can we ignore blank `SSN` and set `SSNDataQuality` to partial if `SSN` is present, but `SSNDataQuality` is blank?

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -464,22 +464,48 @@ def choose_best_pronouns dest_attr, source_clients
...
+        map do |sc|
+          dq = sc[:SSNDataQuality]
+          value = sc[:SSN]&.strip.presence
+
+          # normalization pass
+          if dq.between?(1, 2)
+            numeric = value&.gsub(/\D/, '').presence
+            if numeric.nil?
+              dq = 99
```
eanders: just because we _said_ we didn't collect it, doesn't really mean we didn't collect it...

spec/models/grda_warehouse/tasks/client_cleanup_spec.rb
```diff
@@ -247,7 +247,7 @@
     end
 
     it 'only updates SSN from clients with a value' do
-      source_1.update(SSN: @ssn1, SSNDataQuality: 99)
```
eanders, line 250: I believe this test was specifically here to catch the situation where there's bad SSNDataQuality data, but not necessarily a missing SSN.

app/models/grda_warehouse/tasks/client_cleanup.rb
ttoomey: I've done this but I feel a bit unsure; reversing the sort order seems like a pretty big change potentially?

ttoomey:
@eanders  I revisited this to try and make it a little more performant. Seemed worth it to optimize it since it's run frequently over many clients.

ttoomey:
reverted SSN change and moved to rake task so we can assess churn

lib/tasks/grda_warehouse.rake
```diff
@@ -12,6 +12,56 @@ namespace :grda_warehouse do
...
+              source.DateCreated ||= default_date_created
+              source
+            end
+
+            next if source_clients.empty?
+
+            dest_attr = destination.attributes.with_indifferent_access.slice(:SSN, :SSNDataQuality)
+            proposed = GrdaWarehouse::SSNSelector.call(dest_attr: dest_attr.dup, source_clients: source_clients, use_oldest: use_oldest)
+
+            next if [destination.SSN, destination.SSNDataQuality] == [proposed[:SSN], proposed[:SSNDataQuality]]
```
gigxz, line 48: So this will also report on cases where SSN is unchanged but SSNDataQuality was "downgraded"?  Maybe that is intended, just wanted to flag that I think it will lead to more data to sift through to find cases where SSN actually changed. Again maybe thats fine, not sure how we are intending to use the output of this. 

For example-- if I'm understanding correctly-- if there is a Destination client with 1 source client with values `SSN: "XXXXX1234", SSNDataQuality: 1`, it will be flagged by this process because the SSNSelector would return `2` as the data quality value.

gigxz (approved):

app/services/grda_warehouse/ssn_selector.rb
```diff
@@ -0,0 +1,110 @@
...
+
+        if dq.between?(1, 2)
+          if numeric.nil?
+            dq = 99
+            value = nil
+          elsif numeric.length != 9
+            dq = 2
+          end
+        elsif numeric
+          dq = 2
```
eanders: Am I reading this right that if we have a 9 character SSN (all digits) with a blank `SSNDataQuality` we set `SSNDataQuality` to 2.  That's probably ok, but I wonder if 9 digits should == 1 if it isn't in 1 or 2.

app/services/grda_warehouse/ssn_selector.rb
```diff
@@ -0,0 +1,110 @@
...
+        end
+
+        next unless value
+
+        date_key = timestamp_for(source[:DateCreated]) || default_date_key
+        date_key *= -1 unless use_oldest?
+
+        tie_breakers = [source[:SSNDataQuality] || 99, date_key, source[:id].to_i]
+        existing = best_by_dq[dq]
+        best_by_dq[dq] = { value: value, keys: tie_breakers } if existing.nil? || (tie_breakers <=> existing[:keys]).negative?
```
eanders: Can you add some documentation for what is going on here, I'm really having trouble following it.

lib/tasks/grda_warehouse.rake
ttoomey: Yes, that's accurate. I guess if the dq changes it seems worth having visibility into that. Assuming downstream consumers care about dq but maybe it's generally ignored.

I'd probably do some quick analysis on the output to count records where dq changed, count where ssn changed. That should give us and idea of the impact.

ttoomey: Also note dq can be upgraded to 2 if the record has a numeric ssn and dq was 99 or something

app/services/grda_warehouse/ssn_selector.rb
ttoomey: That's right. I suppose we could promote to dq 1 as you suggest but that feels less "safe" - for example if the ssn was 9 0's or some other obviously fake ssn.

TBH inferring the dq based on the value might be incorrect. Maybe instead we should have some data-source level config that permits to this behavior rather than making this the default.

ttoomey: discussion with elliot:
* it's safe to demote DQ
* it seems to risky to infer a promotion to DQ based on value
* let's also leverage `valid_social?(ssn)` to screen out 9-digits socials that shouldn't be dq 1

app/services/grda_warehouse/ssn_selector.rb
ttoomey: refactored the SSNSelector class to make it easier to read (hopefully)

spec/models/grda_warehouse/tasks/client_cleanup_spec.rb
```diff
@@ -823,6 +834,18 @@
       expect(@ssn2).to eq(@dest_attr[:SSN])
     end
 
+    it 'does not normalize punctuation from SSNs before persisting them' do
```
eanders, line 837: arguably it should.  The CSV spec says numbers only, but it's probably ok this doesn't take care of that.

spec/models/grda_warehouse/ssn_selector_spec.rb
eanders, line 1: Probably not necessary, but it might be nice to have one test that confirms the behavior when the SSN is `XXXXX1234` as that is a common format we see.

eanders (approved):
Thanks @ttoomey, looks great to me.

PR #5823, opened 2025-09-28
Add client Veteran conditions for data correction, move helper text to fragments

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8316

Move two form customizations to version control:
1) Move `helper_text` values that are client-specific, into client-specific patch
2) Move client-specific veteran status condition into **default form fragment**. This allows for editing VeteranStatus values that are present for minors, which is flagged by our data quality tool. This was implemented as a one-off on a customer request on their form only. This PR suggests adding it to the default client form.
   * Accompanying system test is added as well

How to test: `CLIENT=client rails driver:hmis:seed_definitions`

## Type of change
tech debt, configuration


drivers/hmis/lib/form_data/default/fragments/client_demographics.json
```diff
@@ -118,6 +118,30 @@
           "question": "dob",
           "operator": "EXISTS",
           "answer_boolean": false
+        },
+        {
```
ttoomey, line 122: this is fine. Observation is that maybe we should support an "IN" operator to make this kind of thing more concise
```json
    {
      "operator": "IN",
      "question": "veteran_status",
      "answer_code_array": ["YES", "NO", "CLIENT_DOESN_T_KNOW", "CLIENT_PREFERS_NOT_TO_ANSWER"]
    }
```

drivers/hmis/spec/system/hmis/client_veteran_status_spec.rb
```diff
@@ -0,0 +1,200 @@
...
+          # Note: ClientProcessor processes nil into 99 because veteran status is non-nullable in HUD CSV
+        end.to change { c1.reload.VeteranStatus }.from(1).to(99)
+      end
+    end
+
+    context 'where client is a minor with existing Veteran Status = \'No\'' do
+      let(:c1) { create :hmis_hud_client, data_source: ds1, dob: 15.years.ago, veteran_status: 0 }
+
+      it 'does not show veteran status field' do
+        expect(page).not_to have_field('Veteran Status')
```
ttoomey: is this assertion right? I might be misunderstanding but seems like with this rule change the veteran status would show if the value is "No" ("show if veteran status response is YES/NO/DK/R, to support changing existing response when veteran status is present for minor")

ttoomey (approved):

drivers/hmis/lib/form_data/default/fragments/client_demographics.json
gigxz: We actually do support that... I just forgot. thanks. Updating!

drivers/hmis/spec/system/hmis/client_veteran_status_spec.rb
gigxz: Welp, good catch! This is wrong, but it was succeeding because the page hadn't fully loaded yet. Adding assertions to ensure the form is loaded causes this test to fail, which is correct, as you point out. I pushed that update!

ttoomey (commented):
Looks good, thanks!

PR #5824, opened 2025-09-29
Improve request logging

ttoomey:
## _Merging this PR_

### Changes
* Sanitize untrusted user-supplied values (host headers, X-Forwarded-For, AWS trace IDs, request_start timestamps) to prevent log injection attacks
* Remove blank/empty fields before sending to CloudWatch to save log volume and costs
* Remove redundant `process_id` field (already captured as `pid`)

## Type of change
Bug fix


config/setup_logging.rb
```diff
@@ -92,24 +94,35 @@ def _configure_lograge
...
     config.lograge.custom_options = ->(event) do
+      request = event.payload[:request]
+      headers = request&.headers
+      headers_env = headers&.env || {}
+
+      remote_ip = request&.remote_ip.presence || event.payload[:remote_ip].presence
+      client_ip = headers_env['HTTP_CLIENT_IP'].presence
+      x_forwarded_for = headers_env['HTTP_X_FORWARDED_FOR'].presence || event.payload[:x_forwarded_for].presence
+      # RFC 7239 dictates the first IP in X-Forwarded-For is the client IP
+      forwarded_ip = x_forwarded_for&.split(',')&.first&.strip.presence
```
eanders: Given that we're setting these variables here, maybe we should move the logic out of the hash and put all the logic up here.  For example:

```
remote_ip = request&.remote_ip.presence || event.payload[:remote_ip].presence || || forwarded_ip || client_ip || remote_addr
```

which would let us do:
```
{
...
  remote_ip: remote_ip
...
}
```
below.

What do you think?

eanders (approved):
Looks good, but I suggest centralizing the logic, unless we have a reason not to.

config/setup_logging.rb
ttoomey: hoisted logic up, removed untrusted vars from remote_ip, added annotations

ttoomey:
made another pass to refactor/cleanup

eanders (approved):

ttoomey:
sorry for additional iteration, made another small set of enhancements

eanders (approved):
Thanks for all the comments. Looks good to me @ttoomey

ttoomey:
going to test this on QA once 183 is on staging

lib/logging/sanitizer.rb
```diff
@@ -0,0 +1,82 @@
...
+
+    def sanitize_string(value)
+      str = value.to_s
+      # Remove null bytes and control characters (except newlines/tabs)
+      str = str.tr("\u0000-\u0008\u000B-\u001F\u007F", '')
+      # Truncate with indicator
+      str.length >= @max_string_length ? str.truncate(@max_string_length, omission: '...[TRUNCATED]') : str
+    end
+
+    def sanitize_array(array, current_depth:)
```
ttoomey, line 58: sanitize string/hash are added for completeness. As used by lograge these should mostly be strings. I think it would be nice to utilize them for manual structured logging somehow. For example, address usage like this: `Rails.logger.info "Processing #{dataset} #{year} for #{params[:level]} within #{params[:within]}"`

lib/logging/sanitizer.rb
```diff
@@ -0,0 +1,82 @@
...
+# Usage:
+#   sanitizer = Logging::Sanitizer.new
+#   sanitizer.call(user_data)
+#
+#   # With custom limits
+#   sanitizer = Logging::Sanitizer.new(max_string_length: 1000, max_hash_items: 100)
+#   sanitizer.call(user_data)
+#
+module Logging
+  class Sanitizer
```
eanders, line 16: @ttoomey are you at all concerned about the performance of this having to handle every log item?  I'm not seeing any red flags of things that would be slow except the array and hash depth checks.

lib/logging/sanitizer.rb
eanders: I can see how that might be helpful, but I'd probably just convert `params[:within]` to json since the logger only supports strings right?

I wonder a little about the 1,000 character limit since json objects can eat up characters quickly, but hopefully it is enough of the string to be useful and also a reasonable size.

lib/logging/sanitizer.rb
```diff
@@ -0,0 +1,82 @@
...
+        sanitize_array(value, current_depth: current_depth)
+      else
+        value
+      end
+    end
+
+    def sanitize_string(value)
+      str = value.to_s
+      # Remove null bytes and control characters (except newlines/tabs)
+      str = str.tr("\u0000-\u0008\u000B-\u001F\u007F", '')
```
eanders: Is there any built-in regular expression or constant that might have these characters, or characters that would be acceptable? It feels a little brittle to have these specific control characters encoded here.

eanders (approved):
@ttoomey looks good to me, but I did leave a few questions it might be nice to think about prior to merging.

lib/logging/sanitizer.rb
ttoomey: looks like we can use a character class. Maybe easier to read `/[\p{Cntrl}&&[^\t\n]]/`

lib/logging/sanitizer.rb
ttoomey: I don't see any performance concerns here, of course I might be missing something. I think for normal usage the performance impact wouldn't be measurable.

For the case of an attack or aggressive crawler, there might be some impact but I don't see a way around that. But if we wanted to mitigate that we'd probably have more ROI optimizing db queries for anonymous requests than this sanitizer.

lib/logging/sanitizer.rb
eanders: That looks more maintainable to me.  Thanks!

lib/logging/sanitizer.rb
ttoomey: > I can see how that might be helpful, but I'd probably just convert params[:within] to json since the logger only supports strings right?

Well `params[:within]` could be anything, multi-megabyte string or structured data. I think the json serializer would have to do the same thing we have here (descending, etc) but without any of the sanitization or depth protection.

I just meant it would be better to sanitize these tainted params before logging. Could just call `sanitizer.call(params[:within])` but that be kind of ugly code. 

I was kinda thinking about a new API for structured logging. Perhaps using keys
```ruby
# new API sketch
module Logging
  class Event
    SANITIZER = Sanitizer.new

    # need some way to manage or catalog "event" names
    def self.info(event, message: nil, **attrs)
      payload = { event: }
      payload[:message] = message if message
      payload[:attrs] = SANITIZER.call(attrs) if attrs.any?
      Rails.logger.info(payload)
    end
    # debug, warn, etc
  end
end

Logging::Event.info(
  'dataset.processing',
  dataset:,
  year:,
  level: params[:level],
  within: params[:within]
)
```
 

> I wonder a little about the 1,000 character limit since json objects can eat up characters quickly, but hopefully it is enough of the string to be useful and also a reasonable size.

yeah, 1k is just a guess. Maybe we should start with 2k and revisit

ttoomey:
pr feedback addressed in d94ece86bda9fcf94e7d284d665d6700d9019335

eanders (approved):
Thanks @ttoomey !

gigxz (approved):
great!

PR #5831, opened 2025-09-30
Fix for tmp table length

eanders:
## _Merging this PR_

## Description
* Postgres can't correctly handle tables with > 63 character names, adjust the naming convention for temp tables used in the exporter to always be 63 characters or shorter 

## Type of change
* Bug fix

### DIFF
--- a/drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/exporter/base.rb
+++ b/drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/exporter/base.rb
@@ -115,13 +115,16 @@ module HmisCsvTwentyTwentySix::Exporter
           exportable_files.each do |destination_class, opts|
             opts[:export] = @export
             options[:export] = @export
-            tmp_table_prefix = opts[:hmis_class].table_name.downcase
+            # Table names can't be longer than 63 characters, so we need to truncate the prefix
+            # we add roughly 13 characters below, so we'll truncate to 50 characters
+            tmp_table_prefix = opts[:hmis_class].table_name.downcase[0..50]
             dest_config = {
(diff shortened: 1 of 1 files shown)


drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/exporter/base.rb
```diff
@@ -115,13 +115,16 @@ def export!(cleanup: true, zip: true, upload: true)
           exportable_files.each do |destination_class, opts|
             opts[:export] = @export
             options[:export] = @export
-            tmp_table_prefix = opts[:hmis_class].table_name.downcase
+            # Table names can't be longer than 63 characters, so we need to truncate the prefix
+            # we add roughly 13 characters below, so we'll truncate to 50 characters
+            tmp_table_prefix = opts[:hmis_class].table_name.downcase[0..50]
```
dtgreiner, line 120: the only concern i have is we have files (e.g. project & projectcoc) that if truncated could overlap. It doesn't look like we'll run into these conflicts here, but it is still open to that.

dtgreiner (approved):
1 comment, but not a blocker.

drivers/hmis_csv_twenty_twenty_six/app/models/hmis_csv_twenty_twenty_six/exporter/base.rb
eanders: So, that's a good call out.  It _shouldn't_ matter.  We actually only ever have one tmp table per export at a time (they get cleaned up before moving on to the next table.)  In theory we could just call them `tmp_table_#{export_id}`, so the naming conflict only matters if something breaks.  For existing tables, there shouldn't be any truncation given the available length, so I think we're good.

PR #5837, opened 2025-10-01
qa fixes

ttoomey:
## _Merging this PR_

## Description

This fixes a bug where CE eligibility requirements containing "OR" in lowercase with unusual whitespace. When those expressions were combined with additional rules, the missing parentheses could change the rule’s meaning due to operator precedence. Now we use the Dentaku AST directly so every OR stays wrapped

Also:  
- `Hmis::Ce::ProcessPoolsJob`: now tracks whether `lock_for_processing` actually yielded before logging; the warning only fires when the advisory lock wasn’t acquired.  
- Resolver specs: cover lowercase `or` and nested `OR` inside functions (`NOT (...)`)
- `SqlExpressionTranslator` spec: adds coverage ensuring custom assessment fields trigger the expected join on `destination_client_latest_assessments`.

## Type of change
Bug fixes


gigxz:
Thank you for the fix! This looks much more robust. I think we can retarget this to 183 so we can do proper QA. Existing rules do not use `NOT` nor lowercase `or`.

martha (approved):
Thanks for the fix and additional QA work!

ttoomey:
Thanks, I retargeted to 183. I also fixed up an issue in the group detection to avoid redundant parens.

gigxz (approved):

PR #5841, opened 2025-10-01
Add CE system tests

martha:
## _Merging this PR_

## Description
- Adds CE system tests:
  - Unit management (Create unit groups, create units, mark units available for referrals, and view unit eligible clients list)
  - Referrals initiated from client list
  - Direct referral

## Type of change
New spec


config/environments/test.rb
```diff
@@ -92,7 +92,7 @@
   config.force_ssl = false
 
   # Enable asset compilation for system tests
-  if ENV['RUN_RAILS_SYSTEM_TESTS'] == 'true'
+  if ENV['RUN_RAILS_SYSTEM_TESTS'] == 'true' || ENV['RUN_SYSTEM_TESTS'] == 'true'
```
martha, line 95: Adding this, per claude's recommendation, resolved some cold-start issues I was having locally where the first test would always fail.

I'm still not totally sure of the root problem, or why this fixes it. But what I think might be going on is: with the recent addition of rails system tests (in addition to existing react system tests), the rails server is newly serving frontend assets, and here we tell the server _not_ to precompile those assets, so it doesn't take forever to start up.

I'd be interested to know whether other developers also encounter this when running hmis system tests locally (i.e. on release branch, not using the change I added here). And interested if @eanders you could perhaps provide more insight about this config - is my guess on the right track or totally off base?

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_direct_referral_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+
+      # Target enrollment was created
+      target_enrollment = referral.reload.target_enrollment
+      expect(target_enrollment).to be_present
+      expect(target_enrollment.project).to eq(target_project)
+      expect(target_enrollment.current_unit).to eq(unit)
+
+      # Provider can add household members to the target enrollment
+      visit "/client/#{client1.id}/enrollments/#{target_enrollment.id}/household"
+      find("[role='button']", text: 'Add Household Member').click
```
martha: While working on this I noticed/remembered that our "previously associated household members" UX returns only enrollments that the current user has permission to see. (We [say so](https://github.com/greenriver/hmis-frontend/blob/c0cd787f137528fff3106012438fed344ab999bf/src/modules/household/hooks/usePreviouslyAssociatedMembers.tsx#L71) in the UI, for clarity, although it's worth noting that that text isn't displayed if there are no previously associated members returned.) 

Maybe we could expand this so it returns previously associated members where the association is through the source enrollment of a referral that the user does have permission to see, even if they don't have permission to see the source enrollment itself.

(And/or, I was also brainstorming about somehow incorporating the existing concept of "can view limited enrollment details", but that depends on how installations want to configure their permissions.)

I think this is out of scope, but wanted to start a conversation about it. Maybe we can make a ticket if others agree it's worth pursuing?

config/environments/test.rb
eanders: My goal with the separate ENV (if memory serves) was simply to allow us to run one set of system tests in isolation.  It's totally possible I introduced some unexpected interaction with the existing system tests by doing so.  
I wasn't running the front-end (react) tests locally, except once or twice to confirm they worked early on.  I have not seen them fail early on GitHub Actions.

martha: Thanks @eanders that makes sense! Sorry I should have clarified, by "this config" what I really meant was lines 96-98 in this file. 
```ruby
    config.assets.compile = true
    config.assets.check_precompiled_asset = false
    config.assets.digest = false
```

I don't really know what they are doing, but adding them to the react system tests (as well as the rails ones) seems to have fixed the local cold start issues I was seeing.

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+
+# Shared context for CE system tests.
+# These tests are tightly coupled to the workflows defined in CeWorkflows::Ac::WorkflowBuilder.
+RSpec.shared_context 'ce system test helper' do
+  include_context 'hmis base setup'
+
+  let!(:ds1) do
+    ds1 = GrdaWarehouse::DataSource.find_or_create_by!(hmis: 'localhost', source_type: :sftp, name: 'HMIS', short_name: 'HMIS')
+
+    # This is slow because it creates the entire workflow for every test.
```
martha: I'd be grateful for feedback/pointers about this. I spent some time trying to create this workflow config only once, instead of once per test, but I was having trouble doing so in a way that didn't mess up downstream system tests. I'm sure it can be done, but I ran out of a reasonable timebox, so maybe we can discuss/pair or just keep the approach I've created here.

config/environments/test.rb
eanders: My understanding of 96-98 is that tells rails that we should compile the assets, but don't check to confirm they are precompiled, and don't rely on (or maybe create) a digest. 
My read of that is basically, just trust that the assets are correct.

drivers/hmis/spec/support/ce_system_test_helper.rb
eanders: Would it be faster to use a small test-specific form setup?  Maybe we don't need the entire allegheny set just some representative data?

eanders: Alternatively, we could use `PgFixtures` which will run the expensive import once and then can dump a sql file and restore it before each test, which should be pretty fast.

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -13,8 +13,9 @@ class JsonForms
 
     DATA_DIR = 'drivers/hmis/lib/form_data'
 
-    def initialize(env_key: nil)
+    def initialize(env_key: nil, override_generate_cdeds_in_test: false)
```
gigxz: `enable_cded_generation_in_test` might be clearer? "override" isn't really clear whether it's overriding it to generate or not generate

drivers/hmis/lib/hmis_util/json_forms.rb
```diff
@@ -298,7 +303,7 @@ def load_definition(form_definition:, identifier:, role:, title: nil)
       record.set_hud_requirements
 
       # Create/update CDEDs for items that have { mapping: { custom_field_key: '...' } }
-      unless Rails.env.test?
+      if !Rails.env.test? || override_generate_cdeds_in_test?
```
gigxz: If we are generating CDEDs, we should also validate them. Suggest pulling out like:
```
      # Determine if CDEDs should be generated and validated
      should_generate_cdeds = !Rails.env.test? || override_generate_cdeds_in_test?

     if should_generate_cdeds ...
   
     ...
     skip_cded_validation: !should_generate_cdeds, # skip validation if we didnt generate them
```

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../requests/hmis/login_and_permissions'
+
+# Shared context for CE system tests.
+# These tests are tightly coupled to the workflows defined in CeWorkflows::Ac::WorkflowBuilder.
```
gigxz: This might get confusing later when we want to add more general tests. This helper is named "ce system test helper" but it specifically sets up for testing the AC HMIS housing_workflow_template. I'd suggest naming it something specific to AC HMIS housing workflow template, or keeping it generic.

For example maybe a setup like this:

* `ce_system_test_helpers.rb` : non AC-specific helpers that are useful for navigating referrals, assigning contacts, performing steps etc. For example methods like: 
    * `def complete_referral_step(step_name:, as_user: nil, &block) # block for filling in dynamic form`
    * `def assign_contacts(swimlane_name:, user_names:)`
    * etc.
* `ac_hmis_ce_test_helpers.rb`:  AC-specific helpers, and setting up AC forms data and workflow templates
   * could private task-specific helpers like `complete_provider_outcome_step(decision:, ...)`



This isn't a fully formed suggestion, just wanted to leave those thoughts and we can discuss further!

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+        :can_view_units, # but *not* can_manage_units
+        :can_update_unit_availability,
+        :can_view_own_referrals, # only own (assigned) referrals
+        :can_perform_own_referral_tasks,
+      ],
+    )
+  end
+
+  # Eligibility and prioritization fixtures. Note that the rules defined here are not actually used to generate the pool (that's tested elsewhere, outside of system tests).
+  # But rules are present anyway so that the waitlist appears correctly in the UI.
```
gigxz: Maybe these belong in a test thats specifically testing the Unit page UI then?

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative '../requests/hmis/login_and_permissions'
+
+# Shared context for CE system tests.
+# These tests are tightly coupled to the workflows defined in CeWorkflows::Ac::WorkflowBuilder.
+RSpec.shared_context 'ce system test helper' do
+  include_context 'hmis base setup'
```
gigxz: Are we using the data loaded from this context?

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+  let!(:project_ce_config) { create(:hmis_project_ce_config, project: target_project, supports_waitlist_referrals: true, receives_direct_referrals: true) }
+  let!(:workflow_template) { Hmis::WorkflowDefinition::Template.find_by(identifier: 'housing_workflow_v1') } # created already in before_all
+
+  # User setup
+  let!(:admin) { create(:hmis_user, data_source: ds1, first_name: 'Alexandra', last_name: 'Admin') }
+  let!(:provider) { create(:hmis_user, data_source: ds1, first_name: 'Paul', last_name: 'Provider') }
+  let!(:other_user) { create(:hmis_user, data_source: ds1, first_name: 'Oliver', last_name: 'Other') } # user without permissions
+
+  # Simplified permission scenario where admin has all permissions
+  let!(:admin_ac) { create_access_control(admin, ds1) }
```
gigxz: suggest naming access_control explicitly since we use the "ac" acronym elsewhere in this file with another meaning

gigxz: Actually I'm not sure if this is an acronym for "access control" 👀  I guess this relates back to my other comment about clarifying whether this is an AcHmis helper setup file, and if so we can just call it `admin_access_control` etc

drivers/hmis/spec/support/ce_system_test_helper.rb
eanders: Sorry, meant to drop this as a possible example to crib from:
`hmis-warehouse/drivers/hud_data_quality_report/spec/models/fy2022/dq_context.rb`

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,92 @@
...
+    CeWorkflows::Shared::CeBuilderUtils.create_state_machine_custom_statuses(ds1)
+    workflow_builder = CeWorkflows::Ac::WorkflowBuilder.new(ds1)
+    workflow_builder.build_housing_workflow
+    workflow_builder.build_admin_assign_workflow
+
+    ds1
+  end
+
+  let!(:target_project) { create(:hmis_hud_project, data_source: ds1, ProjectType: 3) } # PSH
+  let!(:coc1) { create :hmis_hud_project_coc, data_source: ds1, project: target_project, coc_code: 'CO-500' }
```
gigxz: If we don't care what the coc code is, we can use this attribute to specify generating a coc
```suggestion
  let!(:target_project) { create(:hmis_hud_project, data_source: ds1, ProjectType: 3, with_coc: true) } # PSH
```

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_direct_referral_spec.rb
martha: Resolving in favor of https://github.com/open-path/Green-River/issues/8384

drivers/hmis/spec/support/ce_system_test_helper.rb
martha: Resolved! For posterity: since the data source is created in a `before(:all)` block it also needs to be cleaned up in an `after(:all)` block, otherwise it will mess with downstream tests and cause unexpected behavior (since most tests don't expect multiple hmis data sources).

drivers/hmis/spec/support/ce_system_test_helper.rb
martha: Thanks this is helpful! I ended up restructuring so that:
- ce_system_test_helpers.rb contains non-AC specific helpers, as you said
- AC-specific spec are moved into 1 file for better performance (instead of being separated into 1 for direct and 1 for waitlist referrals), so their shared helper code is just included in that spec file.

drivers/hmis/spec/support/ce_system_test_helper.rb
martha: Yep this was "access control," I've spelled it out! Thank you

drivers/hmis/spec/support/ce_system_test_helper.rb
martha: Thanks good point! I modified this so that the tests that test workflow progression have the referral pre-created in a fixture.

martha: Although, I realized that left us without a test testing referral creation from a client list, so I've added that back as part of the "unit management" test, here: https://github.com/greenriver/hmis-warehouse/pull/5841/commits/da5de9e5ff61029085297cdecd6f78c90b5ed807

gigxz:
Updating branch to pull in https://github.com/greenriver/hmis-warehouse/pull/5891

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -0,0 +1,124 @@
...
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+# Shared context for CE system tests.
+# Sets up a correctly configured target project and admin/provider users with example permissions.
+RSpec.shared_context 'ce system test helper' do
+  # Use find_or_create_by since some tests create the data source in a before(:all) block for performance reasons.
+  let!(:ds1) { GrdaWarehouse::DataSource.find_or_create_by!(hmis: 'localhost', source_type: :sftp, name: 'HMIS', short_name: 'HMIS') }
```
gigxz: I don't know why the tests use `sftp` as the source type, since OP HMIS is not an sftp source. Since its not relevant here, maybe we can change this to:

```suggestion
  let!(:ds1) { GrdaWarehouse::DataSource.find_or_create_by!(hmis: 'localhost', name: 'HMIS') }
```
?

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
```diff
@@ -0,0 +1,411 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: false
+
+require 'rails_helper'
+require_relative '../../../support/ce_system_test_helper'
+
+RSpec.feature 'CE Direct Referrals', type: :system do
```
gigxz: Suggest more descriptive name since this is testing both direct and waitlist-based for AC hmis workflows

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
```diff
@@ -0,0 +1,411 @@
...
+    GrdaWarehouse::DataSource.hmis.delete_all
+
+    # Return enrollment form to normal. (See comment about form cleanup in rails_helper.rb)
+    HmisUtil::JsonForms.new.seed_record_form_definitions(roles: [:ENROLLMENT])
+
+    # Cleanup seeded referral step forms that were created in before(:all)
+    Hmis::Form::Definition.where(role: :CE_REFERRAL_STEP).delete_all
+    Hmis::Hud::CustomDataElementDefinition.delete_all
+    Hmis::Hud::CustomDataElement.delete_all
+  end
```
gigxz, line 41: Could use existing helper `delete_template_and_associated_data` to delete the workflow definition records:
```ruby
  after(:all) do
    # Clean up data source and workflow definition related records, since they were created in before(:all) and not in fixtures.
    # This helps avoid downstream issues in later tests.
    ['housing_workflow_v1', 'admin_assign_workflow'].each do |identifier|
      CeWorkflows::Shared::CeBuilderUtils.delete_template_and_associated_data(identifier)
    end
    Hmis::Ce::CustomReferralStatus.delete_all
    GrdaWarehouse::DataSource.hmis.find_by(hmis: 'localhost').delete

    # Cleanup seeded referral step forms that were created in before(:all)
    forms = Hmis::Form::Definition.where(role: :CE_REFERRAL_STEP)
    forms.each { |form| form.custom_data_element_definitions.delete_all }
    forms.delete_all

    # Return enrollment form to normal. (See comment about form cleanup in rails_helper.rb)
    HmisUtil::JsonForms.new.seed_record_form_definitions(roles: [:ENROLLMENT])
  end
```

I'd be surprised if `Hmis::Hud::CustomDataElement.delete_all` is needed, are we creating actual values in the before-all? I would expect it to only be configuration

spec/support/e2e_setup.rb
```diff
@@ -65,6 +65,25 @@ def mui_select(choice, from:)
     find('li[role=option]', text: choice).trigger(:click)
   end
 
+  # Given the label for a MUI select (dropdown) element, get the choices in the list
+  def get_mui_select_choices(select_label:)
```
gigxz: convention in this file is to prefix mui-specific  helpers with `mui_`. Maybe this can be `mui_select_option_list(from:)`?

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
```diff
@@ -0,0 +1,411 @@
...
+
+        # Household member was enrolled into the same unit
+        expect(target_enrollment.reload.household_members.count).to eq(2)
+        expect(target_enrollment.household_members.all.map(&:current_unit)).to eq([unit, unit])
+      end
+
+      # CE Staff completes the Confirm Success step
+      complete_ce_staff_confirm_success_step('Alice A', unit)
+    end
+  end
```
gigxz, line 257: Suggest adding a few additional cases here - I added these locally and tested:

```ruby
    context 'when a project does not receive referrals' do
      let!(:non_receiving_project) { create(:hmis_hud_project, data_source: ds1, ProjectType: 1, with_coc: true) }
      let!(:non_receiving_unit_group) { create(:hmis_unit_group, project: non_receiving_project, workflow_template: admin_assign_workflow_template) }
      let!(:non_receiving_unit) { create(:hmis_unit, project: non_receiving_project, unit_group: non_receiving_unit_group) }

      it 'does not appear in direct referral list' do
        # Navigate to source project and try to create a referral
        visit "/projects/#{source_project.id}/referrals"
        click_link 'Send Referral'
        mui_select('Alice A and 1 other', from: 'HoH Enrollment')

        # The project should not appear in the dropdown choices
        project_choices = get_mui_select_choices(select_label: 'Project')
        expect(project_choices).not_to include(match(/#{non_receiving_project.project_name}/))
      end
    end

    context 'when target project has a UnitGroup with no availability' do
      let!(:unavailable_unit_group) { create(:hmis_unit_group, project: target_project, workflow_template: admin_assign_workflow_template) }
      let!(:unit_without_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
      let!(:unit_with_closed_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
      let!(:unit_with_locked_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
      let!(:closed_opportunity) { create(:hmis_ce_opportunity, project: target_project, workflow_template: admin_assign_workflow_template, unit: unit_with_closed_opportunity, status: 'closed') }
      let!(:locked_opportunity) { create(:hmis_ce_opportunity, project: target_project, workflow_template: admin_assign_workflow_template, unit: unit_with_locked_opportunity, status: 'locked') }

      it 'cannot be referred to' do
        # Navigate to source project and try to create a referral
        visit "/projects/#{source_project.id}/referrals"
        click_link 'Send Referral'
        mui_select('Alice A and 1 other', from: 'HoH Enrollment')
        mui_select(target_project.project_name, from: 'Project')

        # The unit group should appear in choices but with "0 available" and be disabled
        unit_group_choices = get_mui_select_choices(select_label: 'Unit Group')
        expect(unit_group_choices).to include(match(/#{unavailable_unit_group.name}\s+0 available/))
        option = mui_find_select_option(unavailable_unit_group.name, from: 'Unit Group')
        expect(option['aria-disabled']).to eq('true')
      end
    end
```


e2e_setup change to support finding option element to check if its disabled:
```ruby
  def mui_select(choice, from:)
    option = mui_find_select_option(choice, from: from)
    option.trigger(:click)
  end

  def mui_find_select_option(choice, from:)
    label = find('label', text: from)
    scroll_to(label, align: :center)
    id = label['for']
    find("[id='#{id}']").click
    find('li[role=option]', text: choice)
  end
```

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
```diff
@@ -0,0 +1,411 @@
...
+
+        # Submit the Change Provider Outcome step to move the referral to denied pending status
+        complete_ce_step('Change Provider Outcome (Optional)') do
+          fill_in_step_with_notes(notes: 'Changing the result')
+          mui_radio_choose 'Inability to complete intake', from: 'Decline Reason'
+          expect(page).to have_content('The client\'s in-progress enrollment in this project will be deleted.')
+        end
+
+        expect(page).to have_content('Denial Review Available Today')
+        expect(page).to have_content('Denial Pending')
```
gigxz: This is OK but I think we might as well check that the referral status was updated on the record itself, like
`expect(referral.custom_status.name).to eq('Denial Pending')`

gigxz (approved):
Looks good! Just a few minor comments/questions and suggestion for a couple more cases

drivers/hmis/spec/system/hmis/ce/unit_management_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+    click_link 'Eligible Clients'
+    expect(page).to have_content('The eligible client list for this unit has not been generated yet.')
+
+    # Call the prioritization engine, then reload
+    Hmis::Ce::Match::Engine.call(score_pool)
+    score_pool.update!(candidates_generated_at: Time.current)
+    visit current_path
+    click_link 'Eligible Clients'
+
+    # Verify eligible clients are shown (ordered by priority score descending)
```
gigxz, line 130: note - This could be expanded to test hmis_table_configuration, but I think this is sufficient

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
martha: Thanks!

martha:
Thanks for all the feedback and additional test case suggestions! I think I've addressed everything, going to merge to 185.

gigxz:
great, nice work on this! and exciting to see it green!

PR #5843, opened 2025-10-01
Rescue misconfigured Superset

eanders:
## _Merging this PR_

## Description
* Fixes an issue where Superset is configured in the warehouse but doesn't actually exist and therefore you can't edit users.

## Type of change
* Bug fix

### DIFF
--- a/drivers/superset/app/models/superset.rb
+++ b/drivers/superset/app/models/superset.rb
@@ -38,7 +38,7 @@ module Superset
       roles = Superset::Api.new.roles['result']&.map { |role| role['name'] } || []
       roles.reject! { |role| ignored_roles.include?(role) }
       return roles.presence || default_roles if Superset::Api.new.available?
-    rescue Curl::Err::HostResolutionError => e
+    rescue Curl::Err::HostResolutionError, JSON::ParserError => e
       Rails.logger.error("Error fetching Superset roles: #{e.message}, using default roles")
     end
     # Fallback to the default roles if the API is not available


ttoomey (approved):
This works but it looks like SuperSet::Api#fetch_roles throws `JSON::ParserError` - suggest handling it there or just adding a guard:
```ruby
  private def fetch_roles
    path = '/security/roles/'
    # Superset has a non-standard pagination system, so we need to format the request like this:
    # ?q=(page:0,page_size:100)
    query = { q: '(page:0,page_size:100)' }
    response = get(path, query

    # maybe this (not tested)?
    return [] unless response.body.present? && response.status == 200

    JSON.parse(response.body)
  end
```

eanders:
> This works but it looks like SuperSet::Api#fetch_roles throws `JSON::ParserError` - suggest handling it there or just adding a guard:
> 
> ```ruby
>   private def fetch_roles
>     path = '/security/roles/'
>     # Superset has a non-standard pagination system, so we need to format the request like this:
>     # ?q=(page:0,page_size:100)
>     query = { q: '(page:0,page_size:100)' }
>     response = get(path, query
> 
>     # maybe this (not tested)?
>     return [] unless response.body.present? && response.status == 200
> 
>     JSON.parse(response.body)
>   end
> ```

Thanks for the suggestion.  I tried to track this down, but it turns out it's further up the chain in `login` where we start to hit this.  Without adding significant error checking code, I think we should go with what's here for now.
Ideally we don't have many mis-configured situations or maybe we can find a better mechanism to determine if the Superset API is actually available.

PR #5844, opened 2025-10-02
8339 Improve hmis sessions controller

ttoomey:
## _Merging this PR_

## Description

Improve HMIS sessions controller
*   Hardened authentication flow against timing attacks to prevent username enumeration.
*   Cleared password reset tokens upon successful login to enhance security.
*   Records Authtrail LoginActivity on failure
*   Updated login activity tracking to support multiple user models (`User` and `Hmis::User`).
*   Refactored session creation to allow for custom authentication failure handling.
*   Added comprehensive test coverage for new security enhancements using shared RSpec contexts.

User impact:
* HMIS authentication will take a least 2 seconds if using a username/pass

## Type of change
security hardening, audit logging, testing


drivers/hmis/app/controllers/hmis/sessions_controller.rb
```diff
@@ -13,8 +13,12 @@ class Hmis::SessionsController < Devise::SessionsController
   respond_to :json
 
   skip_before_action :verify_signed_out_user
+  prepend_before_action :begin_time_log, only: [:create]
   before_action :authenticate_with_2fa, only: [:create], if: -> { two_factor_enabled? }
 
+  # Minimum required login processing time for ALL login attempts (seconds)
+  MIN_REQ_LOGIN_TIME = 2
```
eanders: Why 2 and not 0.5 or something shorter?  Is that best practice, or some rough estimate of how long failed attempts currently take?

ttoomey: I just grabbed the code from what we're doing in the warehouse sessions controller, that was using 2 🤷

eanders (approved):
Just a couple questions, but no blockers I see.

drivers/hmis/app/controllers/hmis/sessions_controller.rb
eanders: Got it!

eanders:
Looks like the new tests aren't quite passing.

ttoomey:
> Looks like the new tests aren't quite passing.

Yeah, looks like warehouse failed logins in the spec take like 40 seconds for some reason; seems like it gets blocked on IO. Not sure what's the cause. Might need to just disable that test since it doesn't seem to be an issue in production.

drivers/hmis/app/controllers/hmis/sessions_controller.rb
```diff
@@ -100,8 +131,28 @@ def valid_backup_code_attempt?(user)
     render status: 401, json: { error: { type: type, message: I18n.t("devise.failure.#{type}") } }
   end
 
+  private def authentication_strategy
```
gigxz: Is this used by devise behind the scenes?  I don't see any references to it

drivers/hmis/app/controllers/hmis/sessions_controller.rb
```diff
@@ -100,8 +131,28 @@ def valid_backup_code_attempt?(user)
...
+
+    :password
+  end
+
+  private def handle_failed_authentication
+    # This will be caught by the warden middleware and handled by the CustomAuthFailure failure app, which will render a
+    # JSON response for API requests.
+    # The authtrail gem will automatically record the failed login activity
+    throw(:warden, scope: :hmis_user, message: :invalid)
+  end
```
gigxz, line 139: Just curious is this at all different from the error raised when the call to `user.authenticate!` fails? I know we tested that failed logins hit the CustomAuthFailure app with the correct scope–– I was under the impression that this warden error was thrown internally from authenticate-- but it's been a while.

drivers/hmis/app/controllers/hmis/sessions_controller.rb
gigxz: This seems like unnecessarily long to me..Testing in a deployed environment, I'm seeing the whole end-to-end login request taking in 200-400ms. If we don't have a clear reason to do 2 seconds, I would argue for something shorter like .5s.

gigxz (approved):
Looks good to me! I suggested decreasing the login time, otherwise just clarification questions.

I tested locally with failed logins to HMIS and confirmed login activities are created. Both with wrong password and wrong email, which log slightly differently but both as `hmis_user` scope and `invalid` failure reason.

drivers/hmis/app/controllers/hmis/sessions_controller.rb
ttoomey: Yes, I think you're right that `authenticate!` throws too. I'm not totally sure why it doesn't trigger loging but it may be that it throws before params are processed that would allow authtrail to find the user from the email.

drivers/hmis/app/controllers/hmis/sessions_controller.rb
ttoomey: this got left in by mistake from an earlier attempt to fix. Thanks for flagging!

drivers/hmis/app/controllers/hmis/sessions_controller.rb
ttoomey: Agree that 2 is probably excessive. I guess I'd say that to be effective it needs to mask timing impact when the system is under load. So we might not want to cut the timeout too close to the actual request time. How about we go for 1 second to provide some buffer?

gigxz: Fair enough, 1 sounds good to me

eanders: @dtgreiner do you remember why you went with 2 seconds?

dtgreiner: From what I remember there was some minimal testing to see how long we would need for masking the actual length here, but it almost definitely over-estimated to be on the safe side. I don't have any arguments against lowering it.

ttoomey:
I'll target this to 185 when it's cut

PR #5851, opened 2025-10-02
Support distinct workflows for direct referrals

martha:
## _Merging this PR_

## Description
- To `UnitGroup`, add new `direct_referral_workflow_template` alongside existing `workflow_template`.
- Remove unneeded `workflow_template_identifier` col from `Opportunity`
- Use the correct workflow template when creating a new referral. For direct referrals, default to `workflow_template` if a `direct_referral_workflow_template` is not explicitly set. See inline comment thread

## Type of change
New feature


drivers/hmis/app/models/hmis/unit_group.rb
```diff
@@ -61,6 +68,12 @@ def available_unit_count
       units.receiving_referrals.count
     end
 
+    def template_for_direct_referrals
```
martha, line 71: I waffled between a) always requiring `direct_referral_workflow_template` vs. b) defaulting to `workflow_template` when the former is not available. I think there are pros and cons of both.
a) would let us add validation that involves the project config. For example, if this unit group is in a project that accepts direct referrals, then `direct_referral_workflow_template` should be required, but if it isn't, it should be prohibited from being set. (Or vice versa, we could add a validation on the project config model.)
b) is backwards compatible, which is why I decided to go with b. This way current users won't experience any downtime for existing projects accepting direct referrals.

I think we could move towards a) in the future if we want, but it might be more churn than it's worth. Let me know if you disagree with the direction I've gone here.

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -4,30 +4,29 @@
...
     has_many :referrals, class_name: 'Hmis::Ce::Referral', dependent: :restrict_with_exception
     has_many :categorizations, class_name: 'Hmis::Ce::OpportunityCategorization', foreign_key: :opportunity_id, dependent: :destroy
     has_many :categories, through: :categorizations
     belongs_to :unit, -> { with_deleted }, class_name: 'Hmis::Unit', foreign_key: :unit_id
+    has_one :unit_group, through: :unit, class_name: 'Hmis::UnitGroup'
+    has_one :workflow_template, through: :unit_group, class_name: 'Hmis::WorkflowDefinition::Template'
+    has_one :direct_referral_workflow_template, through: :unit_group, class_name: 'Hmis::WorkflowDefinition::Template'
     has_one :active_referral, -> { active }, class_name: 'Hmis::Ce::Referral', foreign_key: :opportunity_id
     has_one :active_or_accepted_referral, -> { active_or_accepted }, class_name: 'Hmis::Ce::Referral', foreign_key: :opportunity_id
-    has_many :swimlanes, through: :workflow_template, class_name: 'Hmis::WorkflowDefinition::Swimlane'
```
martha, line 26: Removing this because it was unused

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -228,8 +228,7 @@ def ce_template
     def consistent_data_source
       msg = 'must be in same data source as opportunity'
 
-      # Opportunity takes the workflow template data source as the "source of truth", so do the same here
-      data_source = opportunity.workflow_template.data_source
+      data_source = opportunity.data_source
```
martha: `data_source` relation on `Opportunity` already existed; updating to use it

drivers/hmis/spec/jobs/hmis/ce/process_pools_job_spec.rb
```diff
@@ -97,6 +97,9 @@
           create(:hmis_ce_opportunity, candidate_pool: additional_pool)
           create(:hmis_ce_change_marker, trackable: additional_pool, current_version: 1, processed_version: 0)
         end
+
+        # Clear jobs enqueued during setup
+        clear_enqueued_jobs
```
martha, line 102: The `hmis_ce_opportunity` factory now (newly in this PR) creates a unit group, which enqueues the pool processing job, causing some of these tests to fail when they expect no jobs are enqueued yet. Adding `clear_enqueued_jobs` before the tests run is a quick fix.

drivers/hmis/spec/factories/hmis/ce/opportunities.rb
```diff
@@ -3,19 +3,34 @@
...
 
     sequence(:name) { |n| "Opportunity #{n}" }
     status { 'open' }
     project { association :hmis_hud_project, data_source: data_source }
-    workflow_template { association :hmis_workflow_definition_template, data_source: data_source }
-    unit { association :hmis_unit, project: project }
+
+    # Hooks to make the Opportunity factory return a valid opportunity *both* when creating (saving) a record *and* when just building it:
+
+    # In an after(:build) hook, create the associated unit and unit group. (All opportunities should have units and all units are now in groups)
```
martha, line 23: With a simpler approach that looks more like our existing pattern, something like
```ruby
transient do
  ...
  unit_group { association :hmis_unit_group, project: project, workflow_template: workflow_template }
end
...
unit { association :hmis_unit, project: project, unit_group: unit_group
```
I was running into save order issues, where the validation added in this PR on the Opportunity model would fire off an error because the unit wasn't persisted yet. This workaround seems successful, but let me know if you see a clearer more straightforward approach.

drivers/hmis/spec/models/hmis/ce/match/expression/client_field_map_spec.rb
martha, line 1: The change in this file (`target_project` -> `project` fixes an existing strangeness -- the `referrals` rspec factory accepts a `project` attribute, not `target_project`. I'm not sure why it wasn't causing errors before. Another and probably neater alternative would be to update the referral factory to `target_project` and replace all usages of `create(:hmis_ce_referral, project: ...)` with `create(:hmis_ce_referral, target_project: ...)`. The actual referral model has a `target_project`, not a `project`. I haven't done that here since it was a huge churn on top of an already sizable diff. Let me know what you think!

db/warehouse/migrate/20251002192540_remove_workflow_template_identifier_from_opportunities.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+class RemoveWorkflowTemplateIdentifierFromOpportunities < ActiveRecord::Migration[7.1]
+  def change
+    # safety_assured because we added this column to Hmis::Ce::Opportunity.ignored_columns.
+    safety_assured { remove_column :ce_opportunities, :workflow_template_identifier, :string }
```
ttoomey: I might be missing something but I don't think this is actually safe. If this migration runs while the old app is still running, it might try and select against this col. 

To be safe, we'd need a multi-step process: deploy the `ignored_columns` first; then deploy this migration afterwards as the application should no longer be accessing workflow_template_identifier.

Suggest staging this migration in release-186

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -40,9 +40,7 @@ def build_opportunity_for_unit(unit)
 
       unit_group = unit.unit_group
       raise 'Unit must be in a Unit Group to be marked available' unless unit_group
-
-      workflow_template = unit_group.workflow_template
-      raise 'Unit Group has no Workflow Template' unless workflow_template
+      raise 'Unit Group has no Workflow Template' unless unit_group.workflow_template || unit_group.direct_referral_workflow_template
```
ttoomey, line 43: So this means I can mark a unit as available if the unit_group has a direct_referral_workflow_template but no workflow_template. Does the UI prevent me from then trying to create a ce referral for that unit (CreateCeReferral)? Cause if not that might trigger an exception as workflow_template would be nil

drivers/hmis/app/models/hmis/workflow_definition/template.rb
```diff
@@ -90,5 +89,21 @@ def validate!
       validate
       raise ActiveRecord::RecordInvalid, self if errors.any?
     end
+
+    # Returns the form definition that should be used for initiating a direct referral using this workflow.
+    # Returns nil if the workflow is not valid for direct referrals. Caller is responsible for error handling (raise or AR validation error)
+    def direct_referral_form_definition
```
gigxz: This breaks the isolation of not having referral-specific code in the WorkflowDefinition module, I'd suggest a minor refactor to keep things separated. Something like:

```ruby
    # Returns the initial user tasks in the workflow
    def entry_user_tasks
      entrypoint_ids = nodes.entrypoints.pluck(:id)
      graph.walk(entrypoint_ids: entrypoint_ids, stop_when: lambda(&:user_task?))
    end
```

and then on UnitGroup model:
```ruby
    # Form definition to use for direct referrals to this Unit Group, if any (adjust to raise or return nil based on whether there 0 or 2+ entry user tasks)
    def direct_referral_form_definition
      template_for_direct_referrals&.entry_user_tasks&.sole&.form_definition
    end
```

That way the Template is responsible for finding the first user task, but the Unit Group ultimately holds the logic of "what forms should be used for direct referrals".

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -172,4 +173,11 @@ def unit_group_has_one_unit_type
 
     errors.add(:unit_type_id, "must be consistent with unit group's existing type")
   end
+
+  def validate_consistent_project
+    return if unit_group.nil? # TODO(#8157) remove
+    return if project == unit_group.project
+
+    errors.add(:project, "must be same as unit group's project")
+  end
```
gigxz, line 182: This seems like it could cause n+1s when we create units ([here](https://github.com/greenriver/hmis-warehouse/blob/cf54cb539c7c0ae825244eeac573067b56a7744a/drivers/hmis/app/graphql/mutations/create_units.rb#L43-L52)) so maybe need some preloads there?

drivers/hmis/app/models/hmis/unit_group.rb
gigxz: I agree with your decision making

drivers/hmis/app/models/hmis/workflow_definition/template.rb
```diff
@@ -14,7 +14,6 @@ class Template < GrdaWarehouseBase
     has_many :flows, class_name: 'Hmis::WorkflowDefinition::Flow', dependent: :destroy
     has_many :instances, class_name: 'Hmis::WorkflowExecution::Instance', dependent: :restrict_with_exception, foreign_key: 'template_id'
     has_many :swimlanes, class_name: 'Hmis::WorkflowDefinition::Swimlane', dependent: :restrict_with_exception, foreign_key: 'template_id'
-    has_many :ce_opportunities, class_name: 'Hmis::Ce::Opportunity', dependent: :restrict_with_exception, foreign_key: 'workflow_template_identifier', primary_key: 'identifier'
     belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
```
gigxz, line 20: Can we add a `has_many :unit_groups` here? Perhaps with restrict_with_exception or just nulling out the identifiers on the unit group

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -84,7 +84,10 @@ def self.delete_template_and_associated_data(template_identifier)
       puts "Deleting existing CE data associated with #{template_identifier}"
 
       templates = Hmis::WorkflowDefinition::Template.where(identifier: template_identifier)
-      opportunities = Hmis::Ce::Opportunity.where(workflow_template_identifier: template_identifier)
+      # Find opportunities through unit groups that use this template
+      unit_groups = Hmis::UnitGroup.where(workflow_template_identifier: template_identifier).
```
gigxz, line 88: Do the `workflow_template_identifier` and `direct_referral_workflow_template_identifier` columns get cleared out? I think they should, since the intent of the method is to fully delete the template. Relates to other comment https://github.com/greenriver/hmis-warehouse/pull/5851/files#r2433246908

drivers/hmis/spec/factories/hmis/ce/opportunities.rb
gigxz: Hmm, this does seem overly complicated. I can poke around at it and see what I find!

I wonder if the complexity is partially stemming from the fact that the Opportunity table has its own project_id reference. That's not really necessary, since the project can be determined through the unit. It might simplify things if we made that change.

ttoomey:
Looks good, just a question about the migration safety

drivers/hmis/spec/models/hmis/ce/match/expression/client_field_map_spec.rb
gigxz: Hm good find. That probably means we don't need to be passing project at all here, but maybe we do. The referral factory seems to build its own project based in the data source. I think it probably wasn't throwing an error because target_project is a relation on the referral model (through opportunity). Probably not worth the code churn imo!

drivers/hmis/app/models/hmis/ce/referral.rb
gigxz: I'm a bit confused by this/ what we're needing to validate. The opportunity doesn't have a direct reference to data source, but it has opportunity=>project=>datasource (or, opportunity=>unit=>project=>datasource).
If previously we were validating against the data source workflow template, then this seems like a meaningful change.

drivers/hmis/spec/models/hmis/ce/match/expression/client_field_map_spec.rb
martha: > That probably means we don't need to be passing project at all here

The test is testing what project types are included in the `open_referral_project_types`, so I think we do need to pass projects in order to test that

> I think it probably wasn't throwing an error because target_project is a relation on the referral model (through opportunity)

I agree, but what I meant by "I'm not sure why it wasn't causing errors before" is that it _is_ starting to cause errors now, and I haven't tracked down exactly what part of this PR's diff caused that to start. If you revert the change to this file and then run the rspec tests in this file, you see errors that stem from the `hmis_ce_referral` factory not successfully applying `project1`/`project2` to the referrals it creates, when project is passed using `target_project`. I tried systematically reverting a few suspect changes (like the diff in this PR in the referrals factory), but didn't figure it out and decided to move on due to time.

gigxz: Ah okay I see, sorry I wasn't looking closely enough! That does sound odd.

drivers/hmis/app/models/hmis/unit.rb
martha: I looked into this a bit, here's what I found:
- We do have an n+1 problem with the unit creation mutation. We are calling `.each(&:save)` instead of importing. It looks like this is because using `import`, while more performant, would fail to create papertrail versions. (causes this [test](https://github.com/greenriver/hmis-warehouse/blob/release-185/drivers/hmis/spec/requests/hmis/create_unit_spec.rb#L57) to fail.)
- It doesn't look like we can easily add preloads before validating the units, since that's an array of unpersisted units (not an active record relation that can preload).
- It doesn't look like the validation here causes additional n+1 problems thanks to postgres caching. When importing a bunch of units, the unit group and project are always the same, so those queries are cached. In other words, adding a new spec test like this to the spec file linked above:

```ruby
  context 'when creating many units' do
    let(:input) do
      {
        input: {
          input: {
            projectId: p1.id.to_s,
            count: 100,
            unitGroupId: unit_group.id,
            unitTypeId: unit_type.id,
          },
        },
      }
    end

    it 'queries the db a reasonable amount' do
      expect do
        response, result = post_graphql(input) { mutation }
      end.to make_database_queries(count: 50..60)
    end
  end
```
, it fails either way due to the n+1 problem described in my first bullet point, but running with vs. without the validation doesn't change the number of total db queries made.

I would propose leaving this as-is in this PR, since this diff hasn't worsened the problem, but I created a new ticket to address the n+1 problems on unit creation https://github.com/open-path/Green-River/issues/8413. It's in the icebox for now.

gigxz: Ok, Sounds good - thank you!

drivers/hmis/app/models/hmis/ce/referral.rb
martha: Thanks! Does this address your concern? https://github.com/greenriver/hmis-warehouse/pull/5851/commits/190f42c0208ead149b72363c7d150054a1580135

gigxz: Yeah I think so, since we keep the validation against the workflow template ds. thanks!

db/warehouse/migrate/20251002192540_remove_workflow_template_identifier_from_opportunities.rb
martha: Thanks @ttoomey I see, I didn't understand that previously.

> deploy the `ignored_columns` first

if I'm understanding correctly now, we can just remove the migration itself from this PR, but keep everything else as-is, merge this to release-185, then move the migration into the followup PR https://github.com/greenriver/hmis-warehouse/pull/5898, and target that to release-186. does that sound right?

ttoomey: yes, sounds right to me, thanks!

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
martha: thanks for the feedback, I can make a change to my frontend PR. to help address that

martha:
I gave the PR another once-over, updated the QA steps in the ticket, and ran through them quickly myself in dev. I think this is ready to be merged to release-186! I'll move the followup PRs back to draft to clarify that they should be merged in the following release branch (187).

PR #5855, opened 2025-10-03
Add state validation to state machine

gigxz:
## _Merging this PR_

## Description

Update state machine class to add class validation on state column.

**Testing**: I tested that creating/completing a referral workflow was happy, and marking units available/unavailable.

**Classes that use SimpleStateMachine**: `Hmis::Ce::Opportunity`, `Hmis::Ce:Referral`, `Hmis::WorkflowDefinition::Template`, `Hmis::WorkflowExecution::Step`

## Type of change
Bug fix / validation

### DIFF
--- a/app/models/concerns/simple_state_machine.rb
+++ b/app/models/concerns/simple_state_machine.rb
@@ -35,6 +35,7 @@ module SimpleStateMachine
       validates ssm_column, inclusion: {
         in: state_machine_states.map(&:to_s),
         message: "must be one of: #{state_machine_states.map(&:to_s).join(', ')}",
+        allow_nil: true,
       }
 
       # Define event methods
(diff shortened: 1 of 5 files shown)


app/models/concerns/simple_state_machine.rb
```diff
@@ -31,6 +31,13 @@ def state_machine_config(column: :state, &block)
         scope state_name, -> { where(ssm_column => state_name.to_s) }
       end
 
+      # Add validation for state column
+      validates ssm_column, inclusion: {
+        in: state_machine_states.map(&:to_s),
+        message: "must be one of: #{state_machine_states.map(&:to_s).join(', ')}",
+        allow_nil: true,
```
gigxz: In practice all the classes that use this have a non-nullable state column, but I figured the Concern itself shouldn't make that assumption

ttoomey: This is fine as is. It's academic since we don't have nullable status but it feels more intuitive to me to exclude nulls; a null status isn't really valid so it seems like a safe assumption to make 🤷

ttoomey (approved):
this is a great enhancement, thanks!

app/models/concerns/simple_state_machine.rb
gigxz: ok yeah that makes sense to me, it's not a valid state,  I removed the nil allowance

PR #5857, opened 2025-10-03
Mark clients dirty when legacy referrals change

martha:
## _Merging this PR_
- targets release-182 as a hotfix

## Description
Steps taken to verify locally:
- Create a waitlist that has an eligibility rule stating clients are only eligible if they don't have open referrals to specific project types.
- Create a legacy referral for a client who is on that waitlist.
- Confirm they fall off the waitlist. (after running `ProcessClientsJob`)
- Deny the referral
- Confirm they are back on the waitlist.

## Type of change
Bug fix

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
@@ -4,6 +4,8 @@
 # License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
 ###
 
+# frozen_string_literal: true
+
 module HmisExternalApis::AcHmis
   # A proposed fulfillment of a referral request
   class ReferralPosting < ::HmisExternalApis::HmisExternalApisBase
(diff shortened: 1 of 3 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
```diff
@@ -164,6 +166,23 @@ def track_status_changes
       end
     end
 
+    after_save :mark_household_members_dirty
+    def mark_household_members_dirty
+      return unless Hmis::Ce.configuration.enabled?
+
+      source_client_ids = referral.household_members.pluck(:client_id)
```
martha, line 173: Does this look correct? Based on my understanding of comments [here](https://github.com/greenriver/hmis-warehouse/blob/release-182/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral.rb), the referral household members relationship always exists, regardless of the origin of this referral (HMIS or external). (Unlike `enrollment_id` on `Referral`, or `household_id` on `ReferralPosting`.)

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
```diff
@@ -164,6 +166,23 @@ def track_status_changes
       end
     end
 
+    after_save :mark_household_members_dirty
```
martha, line 169: This will run sometimes unnecessarily, since my understanding is we don't strictly need to update CE client lists on all status changes, for example from assigned to accepted pending. I think that seems acceptable, but let me know if you want to get more specific with what status changes trigger this callback.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
gigxz: Yep that is correct!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/referral_posting.rb
gigxz: I think this is good! It matches the existing behavior of the MarkClientAsDirtyBehavior concern which  adds the hook after_save broadly

gigxz (approved):
looks good to me!

PR #5864, opened 2025-10-04
Asset Compliation

eanders:
## _Merging this PR_

## Description
* Removes CI for Asset Compilation (no longer needed since moving to bootstrap 5 with CSS properties)
* Adds asset recompilation to image build 
* Removes fetch of assets from deployment phase

## Type of change
* Code clean-up


config/application.rb
```diff
@@ -31,10 +31,12 @@ class Application < Rails::Application
     # encrypted credentials, it's not appropriate for an open-source project
     if File.exist?(Rails.root.join('config', 'secrets.yml'))
       config.secrets = config_for(:secrets) # loads from config/secrets.yml
-      config.secret_key_base = config.secrets[:secret_key_base]
+      if config.secrets.present?
```
eanders, line 34: Needed to get a deployment working.  I think, because there is a mis-configuration on the infrastructure side.

eanders:
Note for later, `7fe7936` successfully deployed

eanders:
3efb483 also successfully deployed to ECS
and EKS

.github/workflows/build_images.yml
```diff
@@ -95,6 +95,57 @@ jobs:
...
+        id: container-setup
+        uses: ./.github/workflows/container_setup
+
+      - name: Prepare app assets
+        env:
+          RAILS_ENV: production
+          SECRET_KEY_BASE: 0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdef
+        run: |
+          cp config/secrets.yml.sample config/secrets.yml
+          ruby -e "require 'yaml'; env = ENV.fetch('RAILS_ENV', 'production'); secrets = YAML.load_file('config/secrets.yml'); secrets[env] ||= {}; secrets[env]['secret_key_base'] ||= ENV.fetch('SECRET_KEY_BASE'); File.write('config/secrets.yml', secrets.to_yaml)"
```
yakloinsteak: Could this just be a script instead of inline?

config/application.rb
yakloinsteak: Usually, we get this value (secret key base)  from an environment variable and set it in the image build to a dummy value. What's the mis-configuration you're talking about? We should fix it, but I'm not clear what the issue is. I'd go further and suggest we don't need or want a secrets.yml file.

docker/app/entrypoint.sh
```diff
@@ -57,39 +57,6 @@ if [ "$CONTAINER_VARIANT" = "dj" ]; then
   fi
 fi
 
-echo 'Clobbering assets...'
```
yakloinsteak, line 60: This makes me feel so good.

.github/workflows/build_images.yml
```diff
@@ -95,6 +95,57 @@ jobs:
...
+          WAREHOUSE_DATABASE_ADAPTER: postgis
+          WAREHOUSE_DATABASE_HOST: postgres
+          WAREHOUSE_DATABASE_PASS: postgres
+          WAREHOUSE_DATABASE_USER: postgres
+          WAREHOUSE_DATABASE_DB: warehouse_db
+          HEALTH_DATABASE_DB: health_db
+          REPORTING_DATABASE_DB: reporting_db
+        run: |
+          echo "RAILS_ENV=$RAILS_ENV"
+          bundle exec rails assets:precompile
```
yakloinsteak: What's the purpose of building the assets here? I'd have expected it to be in the Dockerfile near the end of the production stage. I think I'm gathering that it's here and then the assets are copied into the build. That works, but I'd like to know more about why. Seems a bit non-standard to me.

yakloinsteak (commented):
Have some questions before I fully approve, but I'm excited this is going through soon!

yakloinsteak (approved):

PR #5871, opened 2025-10-06
PG_REPACK Supported Databases

dtgreiner:
## _Merging this PR_

## Description

Add a check to the database bloat methods so they will only target pg_repack on databases that should be supporting the pg_repack extension. This will prevent attempting to run repack on the Health database

## Type of change

Bug fix


app/models/dba/database_bloat.rb
```diff
@@ -120,9 +120,18 @@ def adjust_autovacuum_for(row)
     run(sql)
   end
 
+  def pg_repack_supported?
+    [ApplicationRecord, GrdaWarehouseBase, ReportingBase].include?(ar_base_class)
+  end
```
eanders, line 125: Stupid question, but could we interrogate the DB in some way to see if we have what we need?  That way we could repack the DBs that support it and roll it out automatically as we update databases.

eanders (approved):

app/models/dba/database_bloat.rb
dtgreiner: We can check if it's enabled on the database. The biggest reason I went this direction was that we want to catch instances where it should be enabled but isn't. Those situations should fail like the Health database currently is failing.

I can add some logic to still run pg_repack if the extension exists and is a known version for databases outside the supported databases while also allowing it to fail if a supported database doesn't have the extension or a supported version of the pg_repack extension

eanders: This is good as is, thanks

PR #5881, opened 2025-10-08
Raise error if attempting to over-request Units

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8359

Fix: Prevent over-requesting units with legacy ReferralPostings
* **Problem**: Providers could "over-request" units when legacy ReferralPostings exist, potentially assigning new referrals to units already slated for legacy referrals.
* **Solution**: Added validation to MarkUnitsAvailable mutation that prevents marking more units available than can be safely assigned. Calculates max available units as vacant_units - assigned_legacy_postings and rejects requests that exceed this limit.

This is a temporary fix for the transition period until legacy ReferralPostings are phased out. 

**How to test:** go to a project that has any assigned or accepted pendings. set it up for CE referrals. try to mark all units available such that it could max out the capacity, making it impossible for the postings to be accepted

## Type of change
Bug fix

  - [x] ensured this does not introduce N+1s **note** has some N+1 on number of unit types but thats minimal so I think not worth optimizing)

### DIFF
--- a/drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
+++ b/drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
@@ -87,7 +87,8 @@ module Mutations
           count
 
         # Calculate how many units can be marked available: vacant units - assigned postings
-        max_available_units = vacant_units_count - assigned_postings_count # TODO- dont allow to be negative, just say 0 if thats the case
+        max_available_units = vacant_units_count - assigned_postings_count
+        max_available_units = [max_available_units, 0].max # if already over-requested, max is 0
 
         # Skip if we're not trying to mark more units available than allowed
         next unless num_units_being_marked > max_available_units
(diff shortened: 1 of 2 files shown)


drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -57,5 +60,45 @@ def build_opportunity_for_unit(unit)
...
+      return unless project.external_referral_postings.active.exists? # early return if no active legacy ReferralPostings
+
+      # Count units being marked available by unit type
+      num_units_by_type = units.map(&:unit_type_id).tally
+
+      num_units_by_type.each do |unit_type_id, num_units_being_marked|
+        next if unit_type_id.nil? # Skip units without unit type
+
+        # Count total vacant units for this unit type in this project
+        vacant_units_count = project.units.
```
martha, line 78: for CE referrals we have future planned work to mark a unit as available on a future date. Do legacy referrals have any equivalent? I think no, just double checking.

martha (approved):
Everything looks great, the only question I could think of is in my inline comment.

Would you like to target this to release-184 since 184 will go to prod tomorrow anyway? I'm fine either way. (and I'm happy to deploy early in the day if helpful)

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
gigxz: No they don't. These are referrals that are already assigned to the project, so we basically need to "hold" a unit for them so that the units don't fill up before the legacy referral gets accepted

PR #5891, opened 2025-10-09
7547 system tests timeout

ttoomey:
## _Merging this PR_

## Description

Simplifies E2E test infrastructure by removing the external `browserless/chrome` Docker service and using Chromium directly within the app container instead. This improves test reliability and reduces setup complexity.

**Key changes:**
- Replace remote Chrome connection with local Chromium binary
- Update `cuprite` and `ferrum` to latest versions (unpinned)
- Add custom debug proxy to expose DevTools for inspection during development
- Restructure E2E testing documentation for better organization

## Type of change
Infra fix


eanders (approved):
Thanks @ttoomey, if this is consistent, that's amazing!

ttoomey:
yeah, looks like the old browserless images were the issue. I tried and failed to get us to v2 but ran out of time.

ttoomey:
@martha - does this seem okay with you? We do lose the live-debugging feature but screenshots are still captured.

martha:
Hm, I have to say that I do rely on the live debugger pretty heavily. I have been noticing that sometimes the screenshot capture incorrectly captures just a loading spinner, but using the live debugger helps me accurately debug the issue. 

Would it be possible to take different strategies in CI vs. local development?

ttoomey:
> Would it be possible to take different strategies in CI vs. local development?

For sure, I'll back out the changes to the dev env and keep the change for CI

docker/docker-compose.yml
```diff
@@ -75,6 +74,11 @@ services:
     env_file:
       - .env.local
       - .env.test
+    environment:
```
ttoomey, line 77: does anyone use this `spec` service?

drivers/hmis/E2E_TESTING_README.md
```diff
@@ -0,0 +1,130 @@
+# E2E Testing for HMIS
```
ttoomey, line 1: extracted from HMIS README

spec/support/e2e_debug_proxy.rb
```diff
@@ -0,0 +1,123 @@
...
+
+require 'socket'
+
+# Exposes Chromium's loopback-only DevTools socket on a host-accessible port so
+# Capybara + Cuprite can be inspected from outside the container. Chromium 136+
+# ignores the legacy remote-debugging-address flag, so we bridge traffic using a
+# lightweight TCP proxy instead of depending on additional system services.
+
+module E2eTests
+  module DebugProxy
```
ttoomey, line 11: this could be achieved with netcat but this inline proxy is simpler to manage as it is integrated into the test process; AI authored

ttoomey:
@martha  - I believe I've replicated the previous debugging experience although the steps to get into the debugger are a little different. See drivers/hmis/E2E_TESTING_README.md

Is this okay DX?

docker/docker-compose.yml
eanders: I use it all the time.

drivers/hmis/E2E_TESTING_README.md
```diff
@@ -0,0 +1,140 @@
...
+    ```
+
+    You can specify multiple fallback branches separated by colons, although this is mostly used in CI to guess at the right branch:
+    ```bash
+    BRANCH_NAME=my-feature-branch:main ./bin/run_hmis_system_tests.sh --dev
+    ```
+
+3. In a new terminal tab, enter the same Docker _container_ where you are running the frontend (opening a new shell session):
+    ```bash
+    docker exec -it $(docker ps -aqf "name=^hmis-warehouse-shell-run" | head -1) /bin/bash
```
martha, line 91: I believe this is a change from the instruction we previously had:
`docker exec -it $(docker ps -aqf "name=^hmis-warehouse_shell_run" | head -1) /bin/bash`

That previous instruction didn't work on my setup, which is why I added the "note if that doesn't work for you" right below. The version you have here works for me, but may not work for @gigxz if I'm remembering correctly?

martha (approved):
@ttoomey yep I think it looks great! Thank you!

drivers/hmis/E2E_TESTING_README.md
```diff
@@ -0,0 +1,140 @@
...
+
+2.  Run your test as usual. When the `debug` statement is hit, you'll see a message in the console with a URL.
+
+3.  In your host Chrome browser, open `chrome://inspect`. The paused pages should appear automatically under "Remote Target"—click "inspect" to attach Chrome DevTools to the live session.
+
+    Note: The proxy defaults to port 9222 (Chrome's default discovery port), which is `CHROME_DEBUGGING_PORT - 1`. If you need to use a different port (e.g., due to conflicts), set `CHROME_DEBUGGING_PROXY_PORT` explicitly and add that target manually in `chrome://inspect`.
+
+
+### Other Debugging Tips
+- Use `byebug`
```
martha: Just noting a change: I believe, using `byebug`/`pry`, we previously were able to pause and inspect both the browser state and the rails state at the same time. But now, the `debug` helper is necessary to setup the debug proxy, and in the `debug` paused state we don't get a rails console. Is that right? Maybe we can mention that here? That seems totally acceptable to me, I just want to confirm my understanding is correct and maybe call out the change in the readme.

ttoomey: Good catch. Looks like you can call `debug(binding)` to get the both pry and pause. Updating docs

drivers/hmis/E2E_TESTING_README.md
gigxz: This is the alias I use locally to attach to an existing container, not sure if the other one worked for me or not
```
dcr_attach: aliased to docker exec -it $(docker ps -aqf "name=^hmis-warehouse_shell_run" | head -1) /bin/bash
```

gigxz:
great work @ttoomey!! 🥳

PR #5899, opened 2025-10-14
Add hmis url to analytics.data_sources

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8401

Add the URL for OP HMIS data sources to the analytics view, so they can be used to populate backlinks to OP HMIS from reports.

## Type of change
Analytics view

### DIFF
--- a/db/views/analytics_data_sources_v02.sql
+++ b/db/views/analytics_data_sources_v02.sql
@@ -1,0 +1,6 @@
+SELECT
+  id, name, short_name, last_imported_at, source_type, visible_in_window, authoritative, authoritative_type, obey_consent, hmis
+FROM
+  data_sources
+WHERE
+  deleted_at IS NULL
--- a/db/views/analytics_data_sources_v02.sql
+++ b/db/views/analytics_data_sources_v02.sql
@@ -1,5 +1,5 @@
 SELECT
-  id, name, short_name, last_imported_at, source_type, visible_in_window, authoritative, authoritative_type, obey_consent, hmis
+  id, name, short_name, last_imported_at, source_type, visible_in_window, authoritative, authoritative_type, obey_consent, hmis as hmis_url
 FROM
   data_sources
 WHERE
--- a/db/warehouse/migrate/20251014152145_update_analytics_data_sources_to_version_2.rb
+++ b/db/warehouse/migrate/20251014152145_update_analytics_data_sources_to_version_2.rb
@@ -1,0 +1,13 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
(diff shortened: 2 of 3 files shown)


eanders (approved):
Looks great, thanks!

db/views/analytics_data_sources_v02.sql
```diff
@@ -0,0 +1,6 @@
+SELECT
+  id, name, short_name, last_imported_at, source_type, visible_in_window, authoritative, authoritative_type, obey_consent, hmis
```
ttoomey: * maybe we should alias `hmis` to `hmis_url` to be more descriptive?
* If the goal of the issue is to compose backlink URIs, do we need to expose all the fields on DS for that? For example, we might consider keeping fields like `visible_in_window` or `obey_consent` as internal unless they are needed for URI construction

ttoomey: This is fine;  the view is extremely simple. Not sure but it might be worth dropping in a quick unit test to protect against regressions. Or we can just rely on QA

Maybe something like this generated spec, not tested
```ruby
# spec/models/analytics/data_source_spec.rb
require 'rails_helper'

RSpec.describe Analytics::DataSource, type: :model do
  let!(:active_source) do
    create(
      :data_source,
      name: 'Active Source',
      hmis: 'https://active.hmis.com',
      deleted_at: nil
    )
  end

  let!(:deleted_source) do
    create(
      :data_source,
      name: 'Deleted Source',
      hmis: 'https://deleted.hmis.com',
      deleted_at: Time.current
    )
  end

  describe 'view content' do
    it 'includes columns from the underlying data_sources table' do
      record = described_class.find(active_source.id)

      expect(record.id).to eq(active_source.id)
      expect(record.name).to eq('Active Source')
      expect(record.hmis).to eq('https://active.hmis.com')
      # ... test other columns as needed ...
    end

    it 'only includes records that have not been soft-deleted' do
      record_ids = described_class.pluck(:id)

      expect(record_ids).to include(active_source.id)
      expect(record_ids).not_to include(deleted_source.id)
    end
  end
end

```

ttoomey (approved):

db/views/analytics_data_sources_v02.sql
gigxz: Agreed to aliasing to hmis_url, I'll do that!

I'm not sure about the usage of `visible_in_window` etc, these were already being included in the analytics view so it's possible they are already being used for something. The only new field I added was `hmis`.

gigxz: I'm unsure about the tests. That generated test is testing a model `Analytics::DataSource` that doesn't exit. As far as I can tell, we don't have any rspec tests for analytics views, so I don't just want to start us off with the wrong pattern. This is a simple change so I think I will just leave it as-is without a test

ttoomey (approved):
looks good, thanks!

PR #5904, opened 2025-10-15
Stop using redundant project_id association on Opportunity

gigxz:
## _Merging this PR_

## Description

This PR removes the direct `project_id` association from the `Opportunity` model. Opportunities now derive their project information through the `unit` association (unit → project) instead of maintaining a redundant direct link. The `project_id` column remains in the database but is made nullable and ignored by the model.

The associated Unit may be deleted, but the opportunity's association through unit to project still works, thanks to the existing `with_deleted` scope [here](https://github.com/greenriver/hmis-warehouse/blob/release-186/drivers/hmis/app/models/hmis/ce/opportunity.rb#L20)

This allows us to greatly simplify the Opportunity factory for testing, since we don't need to maintain the additional project reference. It also allows us to remove the project consistency validation check.

### QA steps
* Verify existing opportunities  (refer into them)
* Verify creating/removing opportunities
* Verify new opportunities (refer into them)
* Verify filtering opportunities (Client=>Available Units table filters)
* Verify project and admin "Available units" tables, with filtering
* Verify referral table filtering (filtering referrals by project/type on the admin referral page)

## Type of change
Code clean-up, database change


gigxz:
@martha this is where I got!
`rspec -f d -P "drivers/hmis/spec/models/hmis/ce/**/*_spec.rb"` is passing, but I didn't test anything else.
Feel free to merge this into your branch.

db/warehouse/migrate/20251015181853_make_opportunity_project_id_nullable.rb
```diff
@@ -0,0 +1,13 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class MakeOpportunityProjectIdNullable < ActiveRecord::Migration[7.1]
+  def change
+    change_column_null :ce_opportunities, :project_id, true
```
gigxz, line 11: On thinking this through @martha it may make sense to just open this migration and association changes as a separate PR against 185? I think if we don't change the factory yet, there shouldn't be too much churn, but I may be wrong about that. Just thinking it might be good to isolate this for review purposes, I'm slightly concerned there might be some side effects I'm not thinking of.
Just wanted to throw that idea out there, but your call - I'm not sure how much effort it would be.

martha: Agreed, yeah I was starting to have the same concern when I saw the graphql api fields that needed to be adjusted. I'll keep this unmerged and finish getting the base branch mke/8001-direct-v-waitlist ready.

gigxz: Sounds good, thanks

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -232,7 +233,7 @@ def consistent_data_source
 
       errors.add(:opportunity, msg) unless data_source == opportunity.data_source
       errors.add(:client, msg) unless data_source == client.data_source
-      errors.add(:target_enrollment, msg) if target_enrollment && data_source != target_enrollment.data_source
+      errors.add(:target_enrollment, msg) if target_enrollment && data_source != target_enrollment.data_source # need this?
```
martha: Target enrollment is a separate column keyed by enrollment ID (not by `**hmis_relation`), so I think yes, we should keep this, to make sure the target enrollment is always in the same data source as the referral?

db/health_structure.sql
```diff
@@ -1,8 +1,3 @@
--- \restrict d5jKJWDEudoxJPrGSz5tZm6Uzta7TBQwE87T7bBcMqO6SLKXLgGZD2agahItER2
```
martha, line 1: todo @martha - revert unneeded changes?

drivers/hmis/spec/requests/hmis/ce/project_ce_opportunities_spec.rb
```diff
@@ -98,7 +102,7 @@
             response, result = post_graphql(**variables) { query }
             expect(response.status).to eq(200), result.inspect
             expect(result.dig('data', 'project', 'ceOpportunities', 'nodesCount')).to eq(41)
-          end.to make_database_queries(count: 15..20)
+          end.to make_database_queries(count: 15..25)
```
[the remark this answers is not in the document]
gigxz: query went from 20 to 21, still reasonable

drivers/hmis/spec/models/hmis/ce/match_rule_spec.rb
```diff
@@ -8,8 +8,8 @@
   let!(:project1) { create :hmis_hud_project, organization: organization, project_type: 0 }
   let!(:project2) { create :hmis_hud_project, organization: organization, project_type: 5 }
   let!(:project3) { create :hmis_hud_project, data_source: organization.data_source }
-  let!(:unit_1a) { create(:hmis_unit_in_group, project: project1) }
-  let!(:unit_2a) { create(:hmis_unit_in_group, project: project2) }
+  let!(:unit_1a) { create(:hmis_unit, project: project1) }
+  let!(:unit_2a) { create(:hmis_unit, project: project2) }
   let(:opportunity_1a) { create(:hmis_ce_opportunity, workflow_template: template, project: project1, unit: unit_1a) }
   let(:opportunity_1b) { create(:hmis_ce_opportunity, workflow_template: template, project: project1) }
   let(:opportunity_2) { create(:hmis_ce_opportunity, workflow_template: template, project: project2) }
```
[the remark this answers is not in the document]
gigxz: legacy fixtures from when rules could be tied to opportunities, removing

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -190,7 +190,9 @@ def opportunity
     end
 
     def target_project_id
-      load_ar_association(object, :opportunity).project_id
+      opportunity = load_ar_association(object, :opportunity)
+      project = load_ar_association(opportunity, :project)
+      project.id
```
martha:
```suggestion
      load_ar_association(object, :target_project).id
```

drivers/hmis/app/models/hmis/ce/opportunity.rb
```diff
@@ -6,27 +6,28 @@
 
 module Hmis::Ce
   class Opportunity < GrdaWarehouseBase
-    self.ignored_columns += ['workflow_template_identifier'] # TODO - remove after migration 20251002192540 is run
+    self.ignored_columns += [
+      'workflow_template_identifier', # TODO - remove after migration 20251002192540 is run
```
martha:
```suggestion
```
I think this can now be removed. The comment refers to the wrong migration, which is my mistake, but the actual migration 20251016185810 was merged into release-187

drivers/hmis/app/models/hmis/hud/project.rb
```diff
@@ -57,7 +57,7 @@ class Hmis::Hud::Project < Hmis::Hud::Base
   has_many :hmis_services, through: :enrollments
   has_many :current_living_situations, through: :enrollments
   has_many :project_staff_assignment_configs, class_name: 'Hmis::ProjectStaffAssignmentConfig'
-  has_many :ce_opportunities, class_name: 'Hmis::Ce::Opportunity', foreign_key: :project_id, dependent: :destroy, inverse_of: :project
```
martha, line 60: Thinking through project destroy cascading behavior: It looks like the behavior used to be, when a project is destroyed, all of its opportunities (past and present) are destroyed.

The new behavior is: when a project is destroyed, its (active) units are destroyed. And when a unit is destroyed,
- its open opportunities are closed (not destroyed)
- its closed opportunities are unchanged
- if it has any locked opportunities, it raises

I think that sounds right, especially the last part where a locked opportunity prevents project deletion. 

But I'm curious, have you already investigated/tested project deletion to confirm it doesn't cause bugs with project-less opportunities? I haven't dug into this, but off the top of my head, a few places I would want to check:
- on the Admin Opportunities page - I think this might be OK since we only resolve open opportunities here, but that's implemented as a frontend filter, not a contractual guarantee
- when viewing a past closed Referral, such as from the admin page

This is probably low priority since project deletion is rare.

martha (approved):
The factory and test improvements look really great. Thank you for slogging through that! 🚋

drivers/hmis/app/models/hmis/hud/project.rb
gigxz: Good catch! I will test this to make sure that destruction is handled by the unit relationship, with that special logic that was implemented.

ttoomey (approved):
Looks great, thanks for cleaning up the specs!!

drivers/hmis/app/models/hmis/hud/project.rb
gigxz: I updated it to ensure that destroying the project destroys related referrals. Opportunities are not destroyed by default, they are closed as you mentinoed. I left that behavior rather than building special behavior to delete them. Deleting projects is rare enough I think it's ok https://github.com/greenriver/hmis-warehouse/pull/5904/changes/9b1016891db6d4b58bfab0b7eeb2965ef2d197c1

PR #5907, opened 2025-10-16
External forms patch for text input patterns

gigxz:
## _Merging this PR_

## Description
Fix 2 bugs in External Forms, possibly due to bootstraup upgrade and/or rails upgrade.

1. The old code was setting the `pattern` attribute without a value, [which leads to](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Attributes/pattern#constraint_validation) treating all input values as invalid. This fixes it by only setting the `pattern` attribute if an input_pattern is specified.
    * This bug isn't presenting on currently published forms (those appear to have the correct HTML structure), the bug only appears when you publish a new forms.
2. The old code was not evaluating validations for Date items that were not required. Dates should always be validated because there are constraints on values (using `pattern`).

Tested locally - I have a form set up at `https://hmis-warehouse.dev.test/hmis_external_api/external_forms/gig/pit.html`

Found when publishing a new form to staging for. https://github.com/open-path/Green-River/issues/8362. 

## Type of change
Bug fix


ttoomey (approved):
looks good, thanks for the fix!

PR #5915, opened 2025-10-17
Enroll household members when creating referral target enrollment

martha:
## _Merging this PR_

## Description
- In the Referral Enroller, when creating a referral's target enrollment, also enroll the source enrollment's household members.
- When deleting a referral's target enrollment, delete all the household members.
- Update spec and system tests

## Type of change
New feature

- [x] I have updated the documentation (or not applicable) - yes, updated "Coordinated Entry QA Workflows" doc


drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -26,47 +26,85 @@ def create_enrollment(message)
...
+      enrollments.delete(hoh_target_enrollment) # Remove from list so it's not saved again below
+      hoh_target_enrollment.assign_unit(unit: unit, start_date: entry_date, user: message.user)
+
+      # Save the HoH's (referred client's) enrollment and associate it with the referral.
+      hoh_target_enrollment.save_new_enrollment! # Saves as WIP or non-WIP, depending on auto-enter rules in the project
+      referral.update!(target_enrollment: hoh_target_enrollment)
+
+      # Reload the unit's latest opportunity, so that when assigning unit for the other hhm,
+      # the "active referral" is fresh and we know that they are members of the same household
+      unit.latest_opportunity&.reload
```
ttoomey: This is fine as is, just noting that the need to call `reload` is a bit of a code smell in `assign_unit`

Changing `assign_unit` it is out of scope but I imagine we could just pass referral in to clean it up 
```diff
-   def assign_unit(unit:, start_date:, user:)
+   def assign_unit(unit:, start_date:, user:, active_referral: nil)
+     ...
+     opportunity = active_referral ? active_referral.opportunity : unit.latest_opportunity
```

ttoomey (approved):
This looks good to me, nice work!

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: Thanks Theron! Hrm, yes I was wondering about this too so I'm grateful for the feedback.

I implemented your suggestion, https://github.com/greenriver/hmis-warehouse/pull/5915/commits/70b22cc7c104a0d88786ffe327b4d9ef21c54d10, with the added change of validating that the passed `active_referral`'s unit is the same as the unit trying to be assigned to. It still feels a little crunchy to me; I have the sense that ideally, the `assign_unit` function should take care of this internally, with the existence of an active opportunity/referral being something the unit "knows" accurately about itself, without the caller needing to worry about it. However, I can't think of a good way to do this without calling `reload`, which I agree is smelly.

Let me know if you think this looks okay, and thanks for helping me think it through!

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
```diff
@@ -26,47 +26,81 @@ def create_enrollment(message)
...
+
+      # Raise an exception if there are any entry date validation errors.
+      # Future improvement would be to return this as a validation error instead of raising
+      error_out(validation_errors.map(&:full_message).join(', ')) unless validation_errors.empty?
+
+      # Assign the HoH's (referred client's) enrollment to the unit and save.
+      # Do this before saving the other hh member enrollments, so that assign_unit validations pass.
+      hoh_target_enrollment = enrollments.find { |e| e.client == referral.client }
+      enrollments.delete(hoh_target_enrollment) # Remove from list so it's not saved again below
+      hoh_target_enrollment.assign_unit(unit: unit, start_date: entry_date, user: message.user)
```
ttoomey, line 71: probably want this here for consistency?

```suggestion
      hoh_target_enrollment.assign_unit(unit: unit, start_date: entry_date, user: message.user, active_referral: referral)
```

ttoomey (commented):
Looks good, sorry forgot to re-review. Just a small thing I noticed

drivers/hmis/app/models/hmis/ce/referral_enroller.rb
martha: Thanks! Opened https://github.com/greenriver/hmis-warehouse/pull/5931 as a follow-up

PR #5922, opened 2025-10-21
8156 enable enhanced ssn reconciliation

ttoomey:
## _Merging this PR_

### Description
Refactors the SSN selection logic into a dedicated service object to improve data quality and maintainability.

-   **`app/models/grda_warehouse/tasks/client_cleanup.rb`**: Updates the `choose_best_ssn` method to delegate to the new `GrdaWarehouse::SSNSelector` service, replaces the previous implementation.
-   **`lib/tasks/grda_warehouse.rake`**: Removes the obsolete `ssn:audit` Rake task

### Type of Change
Refactor


app/models/grda_warehouse/ssn_selector.rb
```diff
@@ -146,14 +146,5 @@ def coerce_dq(value)
 
       Integer(value, exception: false)
     end
-
-    def candidate_preferred?(existing, candidate)
```
ttoomey, line 150: removed unused code

ttoomey:
Marking as draft to avoid accidental merging; merge after sign-off

lib/tasks/grda_warehouse.rake
```diff
@@ -12,56 +12,6 @@ namespace :grda_warehouse do
   desc 'Setup a sample GRDA warehouse database'
   task setup: [:migrate, :seed_data_sources]
 
-  namespace :ssn do
```
ttoomey, line 15: removing this; only needed for investigation

eanders (approved):
Thanks @ttoomey !

gigxz:
Marked this as "ready for review" since the blocker has been resolved (more info on issue https://github.com/open-path/Green-River/issues/8156).

@ttoomey I made a couple changes in one test to get the legacy tests passing– I know the updated SSN functionality has more complete test coverage elsewhere.

app/models/grda_warehouse/tasks/client_cleanup.rb
```diff
@@ -466,28 +466,8 @@ def choose_best_pronouns dest_attr, source_clients
...
-
-        dest_attr[:SSN] = best[:SSN]
-        dest_attr[:SSNDataQuality] = best[:SSNDataQuality]
-      elsif dest_attr[:SSN].present?
-        dest_attr[:SSN] = nil
-        dest_attr[:SSNDataQuality] = 99
-      end
-      dest_attr
+    def choose_best_ssn(dest_attr, source_clients)
+      GrdaWarehouse::SSNSelector.call(dest_attr: dest_attr, source_clients: source_clients)
```
eanders, line 470: should this include `, use_oldest: use_oldest`?

eanders (approved):
Assuming we made an active decision not to allow toggling between newer and older SSNs (I think we should always be using the newer higher quality data) then this looks good, thanks!

app/models/grda_warehouse/tasks/client_cleanup.rb
gigxz: @eanders where would the use_oldest value come from? From what I can tell we don't currently have a site config around ssn order preference, we have one specifically for name preference

gigxz: > Assuming we made an active decision not to allow toggling between newer and older SSNs (I think we should always be using the newer higher quality data) then this looks good

Flagging that the default value on the SSNSelector class is `use_oldest: true` which seems to contradict Elliot's preference to use newer data. The way @ttoomey implemented prefers older data in a tie, which is what the old/current implementation does.

eanders: Good question.  I was mostly noting that that was in the comment that was removed, wanted to make sure that was an intentional change and not accidental.

eanders: I may be out of the loop on this, but I thought the decision was to prioritize new over old so corrections can be made.  If we always prefer older, we'll only get corrections if all records are updated.

ttoomey: > Flagging that the default value on the SSNSelector class is use_oldest: true 

That might have been a mistake or incomplete change; perhaps preparing for configurable behavior. My understanding is we were just waiting for sign off before changing to 'newest-first' globally. If some stakeholders wanted to keep the old behavior, we could make it configurable

gigxz: Thanks Theron. I don't believe we have gone through a process of getting stakeholder sign-off on changing from "use oldest" to "use newest", so I think it's correct that we aren't making that change here.

I opened https://github.com/open-path/Green-River/issues/8769 to revisit the logic and consider whether we need additional configuration (or to change the default behavior)

gigxz: @eanders are you good with me merging this PR?

eanders: I think so, just confirming, we don't think this is changing the existing prioritization algorithm, but we plan to adjust that later.

If that's the case, go for it!

gigxz: Exactly 👍

PR #5928, opened 2025-10-22
HMIS DQ Tool Client Section Visibility

dtgreiner:
## _Merging this PR_

## Description

The Client section in the HMIS DQ Tool Report differs from the rest of the sections because the Client model doesn't have an attached project. To get the clients grouped by project, we were querying the enrollments tied to the report. The issue here is the enrollments are tied to visibility filtering. This caused only the Clients section to be tied to user visibility filtering. 

The change sets the Clients to follow a similar pattern as the rest of the sections and not force visibility requirements when building the total numbers. Additionally, the link to the client details section is now tied to this visibility. This allows a user who has the ability to run the report for a project but does not have access to view clients to still do so while not being able to view the client data.

## Type of change

Bug Fix


dtgreiner:
@eanders this should be ready for a review

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
```diff
@@ -201,15 +201,6 @@ def self.detail_headers_for(slug, report, export:)
       headers.select { |k, _| k.in?(columns) }
     end
 
-    def download_value(key)
-      translator = self.class.detail_headers[key][:translator]
-      value = public_send(key)
-      return translator.call(value) if translator.present?
-      return value == true ? 'Yes' : 'No' if value.in?([true, false])
```
eanders, line 208: I think we lost TRUE/FALSE -> Yes/No, can we keep `download_value` or maybe this is defined elsewhere?  I see we're still calling it in drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/items.xlsx.axlsx

eanders (approved):
This looks good, but maybe we should add some simple tests to confirm access is working as expected and when a user only has some of the project data, they still get counts in the client section?

drivers/hmis_data_quality_tool/app/models/hmis_data_quality_tool/enrollment.rb
dtgreiner: This was overriding the method defined in dq_concern (included in this class). The two methods were identical, so I removed this one in favor of the base method

PR #5933, opened 2025-10-23
Support filter to exclude recently declined CE candidates

martha:
## _Merging this PR_

## Description
- Add a new filter to Opportunity Candidates query, `exclude_recently_declined`.
- Support that filter with a new query object, `Hmis::Ce::FilteredCandidatesQuery`

Frontend branch: https://github.com/greenriver/hmis-frontend/pull/1305

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
```diff
@@ -13,7 +13,12 @@ class HmisSchema::CeOpportunity < Types::BaseObject
     field :status, HmisSchema::Enums::CeOpportunityStatus, null: false
     field :expires_at, GraphQL::Types::ISO8601DateTime, null: true
     field :referral, Types::HmisSchema::CeReferral, null: true, description: 'Active or accepted referral'
-    field :candidates, Types::HmisSchema::CeCandidate.page_type, null: false
+    field :candidates, Types::HmisSchema::CeCandidate.page_type, null: false do
```
martha, line 16: I'm especially curious for feedback about this part, because it doesn't follow our usual graphql filter pattern, which is to create a filter class for the object being filtered -- `CeCandidateFilter` in this case. 

I had an early draft https://github.com/greenriver/hmis-warehouse/pull/5933/commits/2c1b89d1239a21bbbc66657c0ba85a3e03f40718 that followed that strategy. But I found that a custom filter on the _opportunity_ candidates query (here) ends up being necessary anyway. This enables a simple checkbox in the UI "exclude recent declines" rather than a dropdown of unit groups from which to exclude recent declines, which is the more generic filter option from the perspective of `CeCandidateFilter` (outside the context of an Opportunity).

In the end, candidates are never queried outside the context of an Opportunity anyway, so I just removed all that cruft in favor of the simple custom filter shown here. Please let me know if this looks all right.

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -0,0 +1,104 @@
...
+        where(form_identifier: form_identifiers).
+        group(:destination_client_id).
+        joins(:custom_assessment).
+        maximum(cas_t[:date_updated])
+    end
+
+    def determine_client_ids_to_exclude(most_recent_declines, most_recent_assessment_dates, source_to_dest_client_map)
+      # Exclude clients who have been declined, except those who have since been re-assessed
+      most_recent_declines.filter_map do |source_client_id, decline_date|
+        destination_client_id = source_to_dest_client_map[source_client_id]
```
ttoomey: maybe guard this being nil
```suggestion
        destination_client_id = source_to_dest_client_map[source_client_id]
        next unless destination_client_id
```

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -0,0 +1,104 @@
...
+        assessment_date = most_recent_assessment_dates[destination_client_id]
+
+        # Skip (don't exclude) client who has an assessment date more recent than their decline date
+        next nil if assessment_date.present? && assessment_date > decline_date
+
+        destination_client_id
```
ttoomey: This is fine as-is, maybe a little more readable to invert the condition:
```suggestion
        # Exclude client that have not been re-assessed since their decline. 
        destination_client_id if assessment_date.blank? || assessment_date <= decline_date
```

drivers/hmis/app/graphql/types/hmis_schema/ce_opportunity.rb
ttoomey: I think this a smart approach, using a query object instead of a more generic filter class is simplifying; nice work!

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -0,0 +1,104 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module Hmis::Ce
+  class FilteredCandidatesQuery
```
ttoomey, line 20: some class-level docs?
```suggestion
  # Query for Coordinated Entry opportunity candidates with an option to filter 
  # clients who have recently declined a referral.
  #
  # This filter identifies clients who have been declined from any opportunity within 
  # the same unit group as the target opportunity.
  #
  # An exception is made for clients who have been re-assessed since their most 
  # recent decline. Since a re-assessment may change their eligibility, they are
  # included in the candidate list regardless of the recent decline.
  #
  class FilteredCandidatesQuery
```

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -0,0 +1,104 @@
...
+    include ::Hmis::Concerns::HmisArelHelper
```
ttoomey: This is fine as-is, we use this mix-in a lot. However there's an alternative which uses composition instead.

This is better imo because it avoids  including all the helper methods into the classes nam
```ruby 
  def arel = Hmis::ArelHelper.instance

  # then 
  def foo
    arel.ug_t[:id].eq(@unit_group_id)
  end 
```

ttoomey (approved):
Looks great!

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -0,0 +1,111 @@
...
+    def resolve
+      return @base_candidate_scope unless @exclude_recently_declined
+
+      # { source_client_id => most_recent_decline_timestamp }
+      most_recent_declines = fetch_most_recent_declines
+      return @base_candidate_scope if most_recent_declines.blank?
+
+      # { source_client_id => dest_client_id }
+      source_to_dest_client_map = map_source_to_dest_clients(most_recent_declines.keys)
+      return @base_candidate_scope if source_to_dest_client_map.blank? # unexpected
```
gigxz: I think this would occur if no clients have been declined from the unit group, because then `most_recent_declines` would be empty -- so this doesn't seem unexpected, if I'm understanding correctly

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_candidates_spec.rb
```diff
@@ -0,0 +1,201 @@
...
+  let!(:form_definition) { create(:custom_assessment_with_custom_fields, role: :CUSTOM_ASSESSMENT, status: :published, version: 1) }
+  # custom_assessment_with_custom_fields factory generates cded "custom_question_1"
+  let!(:cded) { create :hmis_custom_data_element_definition, owner_type: 'Hmis::Hud::CustomAssessment', key: 'custom_question_1', form_definition: form_definition }
+  let!(:candidate_pool) { create :hmis_ce_match_candidate_pool, priority_expression: 'cde.custom_assessment.custom_question_1' }
+
+  let!(:unit_group) { create(:hmis_unit_group, project: p1, candidate_pool: candidate_pool) }
+  let!(:unit) { create(:hmis_unit_in_group, project: p1, unit_group: unit_group) }
+  let!(:opportunity) { create :hmis_ce_opportunity, project: p1, data_source: ds1, candidate_pool: candidate_pool, unit: unit }
+
+  describe 'ceOpportunity candidates query' do
```
gigxz, line 27: Take or leave it, but just pointing out: [`ce_opportunity_spec`](https://github.com/greenriver/hmis-warehouse/blob/791ddccf095ffae1dee8b9e3638cd24643248d22/drivers/hmis/spec/requests/hmis/ce/ce_opportunity_spec.rb#L183-L348) tests the candidates query as well - maybe we should move those specs over to this file so they're not arbitrarily spread across two files? or, name this one to be specifically testing the `excludeRecentlyDeclined` filter?

gigxz (approved):
Looks great!!

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
martha: I think we would exit from this method above on line 35 if there were no declines, so shouldn't arrive to this line. I can clarify the comment!

martha: Actually I think this *could* happen if there were a client, without a destination client, who had a declined _direct_ referral

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_candidates_spec.rb
martha: Good point! I had sort of avoided that file, because of the todo `# TODO: adjust this test to test the unit query, since unit will back the "opportunity page"`, which I didn't want to bring in-scope for this ticket... However, you are right, it's messy to have them spread across 2 files. I can try to make the request spec tests mirror the request structure on the frontend.

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -962,6 +962,10 @@ type CeOpportunity {
   unit: Unit
 }
 
+input CeOpportunityCandidatesFilterOptions {
+  excludeRecentlyDeclined: Boolean = false
```
gigxz: Image

Looking at the frontend here, I wonder if the filter text would be clearer as "**Exclude Declined Clients**"?

For me "Recently" implies that there is some fixed time window, which there isn't. I think it might be clearer to just communicate this as "declined clients are excluded", and then have the documented caveat that clients may reappear if they are reassessed. What do you think?

Edit: in the spirit of clear/plain text, maybe **Hide declined clients** or **Hide declined referrals**

martha: Yeah that makes sense! I renamed it.

PR #5936, opened 2025-10-24
Destroy instance, tasks when referral is destroyed

martha:
## _Merging this PR_

## Description
See ticket

## Type of change
Bug fix


ttoomey (approved):
Looks good to me, thanks!

In a follow up, we could consider if there any circumstances that a referral should prevent client record deletion, such as an in-progress or completed referral.

martha:
Sounds good! Looks like we return some [warnings](https://github.com/greenriver/hmis-warehouse/blob/da04cc26101c50dfc09aae4dbbde5d133c83f18c/drivers/hmis/app/graphql/mutations/delete_client.rb#L59-L74) based on problem enrollments, so we could do something similar with referrals. Let's think about that more if we get any feedback on this.

PR #5937, opened 2025-10-25
ADR and implementation plan for replacing rails drivers

eanders:
## _Merging this PR_

## Description
This is very much a "lets's discuss options" PR.  
This proposes:
1. Removing Rails Drivers
2. Keeping a concept of mini MVC packages to encapsulate functionality
3. Trying to make it more explicit how we use encapsulated functionality

## Type of change
* New Feature
* Code clean-up
* Dependency removal


ttoomey:
Thanks!! This is great. There's a lot to think about here.

Generally I think replacing the driver gems's shenanigans with standard ruby/rails mechanisms seems like the right way to go. IIRC there may be some load-order issues we may need to resolve.

## File Organization
An alternative to consider is what if we just merged everything into the `app` dir to achieve a more standard file system org. Then we don't need extra loader config.

Of course the top level of app/models (and controllers, views, etc.) becomes very "busy". But this would better reflect the reality that this is a single, large application; while we have many modules it is still a monolith with no real isolation. So form this perspective the "drivers" directory structure adds a layer of indirection that can be confusing.

Maybe this is a non-starter but just to illustrate what this might look like in `app/models` if we gradually migrated "drivers" under app:

```
app/models/
├── user.rb                     # Core model
├── role.rb                     # Core model
├── grda_warehouse/             # Core namespace
│
├── access_logs/                # <-- Formerly in the access_logs driver
├── adult_only_households_sub_pop # etc
├── adults_with_children_sub_pop/  
|
├── ... (and so on for all ~80 drivers)
│
└── concerns/                   # <-- all concerns and extensions
    ├── hmis/
        ├── hud/form_submittable.rb
     
```

## Feature flag (was `RailsDrivers.loaded`)

The draft plan suggests using `defined?` going forward. 

However we don't currently use load-time conditionals and I suggest we don't go there yet; I worry this would have a small upside (minor memory savings) and introduce new problems that are hard to identify (we have to worry about testing permutations of flags at load time).

My suggestion is to consider how we could better manage feature flags with a  a standard, centralized mechanism. For example:
```
    class Feature
      def self.enabled?(feature_name)
        # control with envs or some other mechanism
        env_var = env_feature_map.fetch(feature_name)
        ENV.fetch(env_var, 'false') == 'true'
      end
    end
```

## Extensions

It seems like a great first step to get rid of the "extensions".

Regardless of rails drivers, the current use of modules give me pause. We have 2 dozen different hud client extensions but all those are actually sharing the same namespace with no real modularity. The modules obscure the true size and complexity of the client class as well as potential naming collisions.

Using explicit modules at least makes this discoverable; which is a big win. But maybe we can identify some better patterns or refactoring.

ttoomey:
oh can we make this ADR 0007? I was going to use 0006 for the service_historty_services ADR

eanders:
@ttoomey, thanks for taking a look see below:
> An alternative to consider is what if we just merged everything into the app dir to achieve a more standard file system org. Then we don't need extra loader config.

We should review this with the team.  I've found having features compartmentalized much easier to work with.  The application is HUGE, and just navigating/finding the right controller, views, and models for a particular feature can be tricky sometimes.  That said, if I'm the only one who get's slowed down by a sprawling codebase, I shouldn't force everyone else into the non-standard structure.

> The draft plan suggests using defined? going forward.

I removed this from the ADR, guess I missed it in the plan.  I propose we just remove it and enforce that we always expect the models to exist.  I don't see a situation where we would need to enable/disable specific "drivers."

> Regardless of rails drivers, the current use of modules give me pause. We have 2 dozen different hud client extensions but all those are actually sharing the same namespace with no real modularity. The modules obscure the true size and complexity of the client class as well as potential naming collisions.

100% agree, this would be an interim step.  Once we've removed the drivers gem, we should revisit and streamline the use of concerns.  When we are no longer concerned with the ability to inject code and disable drivers, I would prefer these simply be added to the various classes or refactored into helper classes, etc.  Whatever makes the most sense and is maintainable.

> oh can we make this ADR 0007? I was going to use 0006 for the service_historty_services ADR

Yes, I saw the other one, before merging we should align numbering.

It sounds like this might be worth some effort to validate the approach.  I'd like to limit the code churn as much as possible while removing the gem, then in a second phase we can improve the ergonomics and maintainability.

eanders:
@ttoomey or @dtgreiner do either of you feel the need to actually add this ADR?  It would need a new number at least, but given that we've already removed the gem, maybe we can just close it?

ttoomey:
> do either of you feel the need to actually add this ADR? It would need a new number at least, but given that we've already removed the gem, maybe we can just close it?

Yeah, we have removed the gem but not the drivers structure. So we could skip the ADR for now as the change was relatively contained. Maybe we could back-fill an ADR at some point to document why the drivers structure was chosen but that feels low priority.

thanks!

eanders:
Closing now that drivers has been removed.

PR #5941, opened 2025-10-27
Return a validation error if referral step is not active

martha:
## _Merging this PR_

## Description

On both the Start Step and Submit Step mutations, raise a user-facing staleness error if the reason they can't successfully complete the operation is that someone else has already completed the step.

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1306

## Type of change
Bug fix


drivers/hmis/app/models/hmis/workflow_execution/engine.rb
```diff
@@ -22,7 +22,7 @@ def initialize(workflow_instance, message_handler:, assignment_handler:)
     end
 
     def active_steps
-      instance.steps.where(status: ['available', 'in_progress'])
+      instance.steps.open
```
martha, line 25: No functional change here; this scope already exists so we probably ought to use it

gigxz (approved):

martha:
I should've retargeted this to release-188. I'll redeploy release-187 to staging environments instead, so this can be QA'd.

PR #5942, opened 2025-10-28
Client Monitoring

eanders:
## _Merging this PR_

## Description
This is very much a "lets's discuss options" PR.  
This proposes:
1. Builds out a framework to collect and monitor changes over time for client, enrollment, or other data
2. Implements metric monitoring for days homeless and household size

## Type of change
* New Feature


app/models/task_queue.rb
```diff
@@ -50,4 +52,54 @@ def run!
     to_run.call
     update(completed_at: Time.current)
   end
+
+  # Register all one-time tasks that should be queued
```
eanders, line 56: This ended up being an unrelated change, originally I had a task for setting up the metrics and notifications, but that got moved to another location.  This feels like the right place for these though, so leaving them here.

dtgreiner (approved):

ttoomey:
sorry I didn't get a chance to review this before it merged; I thought it was still in the "let's discuss" phase. I'll review post merge and let you know if I have any feedback worth opening another PR for

eanders:
@ttoomey you are totally right!  Sorry, I saw the approval come through and merged it.  Happy to revert the commit!

ttoomey:
> you are totally right! Sorry, I saw the approval come through and merged it. Happy to revert the commit!

no worries! I'll open a new issue if I see something in review

db/warehouse/migrate/20251020142304_create_metric_calculation_runs.rb
```diff
@@ -0,0 +1,34 @@
...
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class CreateMetricCalculationRuns < ActiveRecord::Migration[7.1]
+  def change
+    create_table :metric_calculation_runs, comment: 'Log of daily metric calculation jobs' do |t|
```
ttoomey, line 11: seems like this (and other) migrations are not in `db/warehouse_structure.sql` - maybe should get regenerated?

docs/implementation-working-documents/alerting-implementation-status.md
```diff
@@ -101,7 +101,7 @@ All migrations are in `db/warehouse/migrate/`:
 ### Data Migration (Phase 2)
 
 - **TaskQueue Job:** Added to `config/application.rb` as `config.queued_tasks[:migrate_user_notification_preferences]`
-- Calls `GrdaWarehouse::AlertDefinition.seed_initial_definitions`
+- Calls `GrdaWarehouse::AlertDefinition.maintain!`
```
ttoomey, line 104: Do we really want a practice of keeping these plan artifacts? I'm assuming these are AI generated and don't have much value once the feature is built; if we want to retain them I suggest putting them into the issue or pull-request rather than keeping them in git.

docs/architecture/alerting.md
```diff
@@ -171,7 +171,7 @@ module GrdaWarehouse
     scope :by_category, ->(category) { where(category: category) }
     scope :system_alerts, -> { by_category('system') }
 
-    def self.seed_initial_definitions
+    def self.maintain!
```
ttoomey, line 174: This appears to be an AI generated implementation plan. I think docs/architecture should hold architectural documentation for the codebase (ADRs or maybe c4 diagrams); not sure this belongs here (probably better kept in github PR or issue itself)

docs/architecture/metric-tracking.md
```diff
@@ -0,0 +1,612 @@
+# Metric Tracking System - Architecture
```
ttoomey, line 1: Suggest this document would be better housed in docs/features rather than docs/architecture.

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
```diff
@@ -0,0 +1,105 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class CreateMetricSnapshots < ActiveRecord::Migration[7.1]
+  def up
+    safety_assured do
+      # Create parent table as partitioned table
```
ttoomey, line 12: I'm curious to see the size of this table in production. Better to start with partitions and not need the than the other way around.

db/warehouse/migrate/20251020141816_create_metric_definitions.rb
```diff
@@ -0,0 +1,30 @@
...
+      t.integer :calculation_window_days, comment: 'Lookback period in days (e.g., 1095 for 3 years)'
+      t.integer :count_change_threshold, comment: 'Only create new snapshot if value changes from initial_value by at least this amount'
+      t.decimal :percent_change_threshold, precision: 5, scale: 2, comment: 'Only create new snapshot if value changes by at least this percentage'
+      t.boolean :active, default: true, null: false, comment: 'Whether this metric is actively being calculated'
+
+      t.timestamps
+
+      t.index [:entity_type, :name], unique: true, name: 'index_metric_defs_on_entity_and_name'
+      t.index :active
+      t.index :category
```
ttoomey, line 27: these are low cardinality; I would be very surprised if these indexes are needed

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
```diff
@@ -0,0 +1,105 @@
...
+      execute <<-SQL
+        CREATE TABLE metric_snapshots (
+          id bigserial,
+          entity_type varchar NOT NULL,
```
ttoomey, line 16: potential optimization to move `entity_type` to `metric_definitions` but may not be worth the complexity; hard to know without a sense for how large the table will be in practice

db/warehouse/migrate/20251020142304_create_metric_calculation_runs.rb
```diff
@@ -0,0 +1,34 @@
...
+      t.integer :calculation_errors_count, default: 0, comment: 'Calculations that failed'
+
+      t.string :status, null: false, default: 'running', comment: 'running, completed, failed'
+      t.text :error_message, comment: 'Error details if status = failed'
+
+      t.timestamps
+
+      t.index [:entity_type, :calculation_date], unique: true, name: 'index_calculation_runs_unique'
+      t.index :calculation_date
+      t.index :status
```
ttoomey, line 31: low cardinality

app/views/notify_user/metric_threshold_crossed.text.haml
```diff
@@ -0,0 +1,28 @@
...
+      - change = data[:current_value] - data[:previous_value]
+      - sign = change > 0 ? '+' : ''
+      Change: #{sign}#{change}
+    - else
+      New Value: #{data[:current_value]}
+      (New baseline)
+
+\
+You are receiving this notification because you are subscribed to client metric alerts.
+To manage your alert subscriptions, contact your administrator.
```
ttoomey, line 28: maybe this is okay, and cheaper simpler. Generally would better to allow users to opt-out of receiving any automated email (supports 'unsubscribe' link in mail footer).

app/views/notify_user/metric_threshold_crossed.text.haml
```diff
@@ -0,0 +1,28 @@
...
+  \
+  #{metric_name}
+  = '-' * metric_name.length
+  #{snapshot_info[:total_count]} #{'client'.pluralize(snapshot_info[:total_count])} crossed the threshold for this metric.
+  - if snapshot_info[:truncated]
+    (Showing first 50)
+
+  - snapshot_info[:data].each do |data|
+    \
+    Client ##{data[:entity_id]}: #{client_url(id: data[:entity_id])}
```
ttoomey, line 15: We use ids, which is good. But I wonder about including client-level data in an email like this at all. Perhaps it would be safer to say "5 clients changed, click here to see details"- forcing the user to login to see more. Then we don't need to support two views and further minimizes risk of exposing client data

app/views/admin/metric_definitions/index.haml
```diff
@@ -0,0 +1,44 @@
...
+            %th Count Threshold
+            %th Percent Threshold
+            %th Active
+            %th Actions
+        %tbody
+          - @metric_definitions.group_by(&:category).each do |category, definitions|
+            %tr.table-secondary
+              %td{colspan: 8}
+                %strong= category.titleize
+            - definitions.each do |definition|
```
ttoomey, line 24: I think we'd like to pagination via `render_paginated_list` whenever possible. While this list is likely to be small, I'm not sure using a category grouping provides benefit sufficient to make an exception. It seems better to have consistent UX and implementation

app/models/grda_warehouse/monitoring/metric_snapshot.rb
```diff
@@ -0,0 +1,64 @@
...
+
+    validates :initial_observation_date, presence: true
+    validates :current_observation_date, presence: true
+    validates :initial_value, presence: true
+    validates :current_value, presence: true
+
+    validates :current_observation_date, uniqueness: {
+      scope: [:entity_type, :entity_id, :metric_definition_id],
+      message: 'Only one active snapshot per entity/metric',
+    }
```
ttoomey, line 25: Users never actually enter snapshots right? We are doing bulk inserts. So these just mirror db constraints but are unused(?). Suggest just removing them in that case

app/models/grda_warehouse/monitoring/metric_definition.rb
```diff
@@ -0,0 +1,191 @@
...
+
+    # Attributes in metric_definition_attributes that are not database columns
+    def self.non_database_attributes
+      [:alert_code]
+    end
+
+    # Initialize default metric definitions
+    # Called once via TaskQueue to populate the table
+    def self.maintain!
+      available_calculators.each do |calculator_class|
```
ttoomey, line 60: * I think with this use of find_or_create_by! we only sets attributes when the record is first created. If the metric_definition_attributes change, the definitions will become stale.
* probably want some kind of advisory lock to prevent simultaneous runs from colliding

app/models/grda_warehouse/alert_definition.rb
```diff
@@ -27,10 +27,15 @@ class AlertDefinition < GrdaWarehouseBase
     scope :by_category, ->(category) { where(category: category) }
     scope :system_alerts, -> { by_category('system') }
 
-    def self.seed_initial_definitions
+    # Attributes in initial_definitions that are not database columns
+    def self.non_database_attributes
+      [:visibility_check, :email_subject]
+    end
+
+    def self.maintain!
```
ttoomey, line 35: same issues as MetricDefinition.maintain!
* attributes can be come stale if the code changes
* no locking

app/models/grda_warehouse/monitoring/tasks/metric_snapshot_collector.rb
```diff
@@ -0,0 +1,380 @@
...
+      GrdaWarehouse::Monitoring::MetricDefinition.maintain!
+
+      run_record = create_run_record
+
+      metrics = load_metrics
+      entities = load_entities
+
+      Rails.logger.info "Collecting #{metrics.count} metrics for #{entities.count} #{@entity_type} records"
+
+      entities.in_groups_of(BATCH_SIZE, false) do |entity_batch|
```
ttoomey, line 53: `in_groups_of` is an array method right? So this loads the entirety of `entities`. Should we instead use find_in_batches?

app/jobs/collect_client_metrics_job.rb
```diff
@@ -0,0 +1,23 @@
...
+###
+
+# frozen_string_literal: true
+
+class CollectClientMetricsJob < BaseJob
+  queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
+  queue_with_priority 15
+
+  def perform(calculation_date = Date.current)
+    GrdaWarehouse::Monitoring::Tasks::MetricSnapshotCollector.run_daily_collection(
```
ttoomey, line 14: probably could use an enclosing advisory lock

app/jobs/notify_metric_threshold_crossings_job.rb
```diff
@@ -0,0 +1,58 @@
...
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+class NotifyMetricThresholdCrossingsJob < BaseJob
+  queue_as ENV.fetch('DJ_DEFAULT_QUEUE_NAME', :default)
+  queue_with_priority 10
+
+  def perform(calculation_date = Date.current)
```
ttoomey, line 13: needs an enclosing advisory lock

app/models/grda_warehouse/monitoring/metric_definition.rb
```diff
@@ -0,0 +1,191 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+module GrdaWarehouse::Monitoring
```
ttoomey, line 9: for this and other major models, encourage linking to the relevant docs using `@see`

```
 # @see docs/features/mertrics_and_alerting.md
```

app/jobs/notify_metric_threshold_crossings_job.rb
ttoomey: A concern is that there's no tracking which alerts went out and to whom successfully; that means: 
* not idempotent; job run more than once on same day will send duplicate emails
* no way to throttle messages
* can't easily see which alerts a user got

Looks like we do have proper tracking in `GrdaWarehouse::WarehouseReports::Project::DataQuality::Base#send_notifications`, can we do something similar here?
```ruby
create_table :metric_alert_notifications do |t|
  t.date :calculation_date, null: false
  t.references :metric_definition, null: false, foreign_key: { to_table: :metric_definitions }
  t.references :user, null: false, foreign_key: { to_table: :users }
  t.jsonb :entity_ids, default: []  # Store which entities were included
  t.datetime :sent_at, null: false
  t.timestamps
end
```

app/models/menu/menu.rb
```diff
@@ -455,6 +455,14 @@ def warehouse_configuration_menu
         title: 'Warehouse Alerts',
       ),
     )
+    menu.add_child(
+      Menu::Item.new(
+        user: user,
+        visible: ->(user) { user.can_edit_warehouse_alerts? }, # rubocop:disable Style/SymbolProc
+        path: admin_metric_definitions_path,
+        title: 'Warehouse Metrics',
```
ttoomey, line 463: We seem to have several names for this feature: "Warehouse Metrics", "Metric Tracking System", ""Client Metric Alert"

Can we arrive at a standard name for this system? Perhaps "Metric Threshold Monitoring"

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      # For cleanup queries
+      add_index :metric_snapshots,
+                [:metric_definition_id, :current_observation_date],
+                name: 'index_metric_snapshots_for_cleanup'
+
+      # For time series queries ordered by start date
+      add_index :metric_snapshots,
+                [:entity_type, :entity_id, :metric_definition_id, :initial_observation_date],
+                name: 'index_metric_snapshots_for_time_series'
```
ttoomey, line 62: this seems duplicative with `index_metric_snapshots_for_date_range`

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
```diff
@@ -0,0 +1,105 @@
...
+
+      # Foreign key
+      add_foreign_key :metric_snapshots, :metric_definitions
+
+      # Create partitions for last 3 years + next 10 years (13 years = 52 quarterly partitions)
+      # This avoids needing maintenance tasks to create future partitions
+      create_quarterly_partitions(3.years.ago, 10.years.from_now)
```
ttoomey, line 69: what maintains these beyond this migration? In the future the migration may not exist

app/models/grda_warehouse/monitoring/tasks/metric_snapshot_collector.rb
```diff
@@ -0,0 +1,380 @@
...
+      update_snapshots(snapshots_to_update)
+    end
+
+    def load_current_snapshots_for_batch(entities, metrics)
+      entity_ids = entities.map(&:id)
+      metric_ids = metrics.map(&:id)
+
+      # Find most recent snapshot for each entity/metric
+      GrdaWarehouse::Monitoring::MetricSnapshot.
+        where(entity_type: @entity_type, entity_id: entity_ids).
```
ttoomey, line 159: groups snapshots and takes values.first but there's no ordering. Postgres may return the older record first, so we keep extending the wrong baseline. Probably want an `order(id: :desc)` in this chan

app/jobs/notify_metric_threshold_crossings_job.rb
```diff
@@ -0,0 +1,58 @@
...
+    return if crossings_by_alert.empty?
+
+    # For each alert code, send notifications to subscribed users
+    crossings_by_alert.each do |alert_code, crossings_by_metric|
+      # Find the alert definition (warehouse database)
+      alert_definition = GrdaWarehouse::AlertDefinition.find_by(code: alert_code)
+      next unless alert_definition
+
+      # Get user IDs from contact_alert_subscriptions (warehouse database)
+      contact_ids = GrdaWarehouse::ContactAlertSubscription.
```
ttoomey, line 27: missing `.active` scope? probably don't want to email inactive subscriptions

app/models/grda_warehouse/monitoring/metric_calculators/min_household_size_calculator.rb
```diff
@@ -0,0 +1,56 @@
...
+  class MinHouseholdSizeCalculator < BaseCalculator
+    include ArelHelper
+
+    # Batch calculation for multiple clients
+    # Returns hash of { client_id => min_size }
+    def self.calculate_batch(entities, _calculation_date)
+      entity_ids = entities.map(&:id)
+
+      # First, get count of members for each [data_source_id, HouseholdID] combination
+      household_sizes = GrdaWarehouse::Hud::Enrollment.
```
ttoomey, line 19: Seems like this counts exited enrollments, that's probably incorrect. Also this is the min household size over all time? 

A optimization might be to pull household_sizes out to a temporary table as it's static for every calculation (and re-used by the max-household size). Maybe we're assuming it's very cheap to build, not sure that's true

app/models/grda_warehouse/monitoring/metric_definition.rb
```diff
@@ -0,0 +1,191 @@
...
+      attrs = calculator_class.constantize.metric_definition_attributes
+      attrs[:alert_code]
+    end
+
+    # Get threshold crossings grouped by alert code for a given calculation date
+    # Returns hash: { alert_code => { metric_display_name => { data: [...], total_count:, truncated: } } }
+    # Only includes metrics that have an alert_code defined
+    # Excludes initial observations (first snapshot for each entity/metric)
+    # Limits to 50 clients per metric to prevent overwhelming emails
+    def self.threshold_crossings_for_alerts(calculation_date, limit: 50)
```
ttoomey, line 125: I'm not sure but there might be performance issues here; an AI suggested using `LAG()` to solve this without pulling so many snapshot records down. I have low confidence in this solution but including it in case it's needed

```ruby
  def self.threshold_crossings_for_alerts(calculation_date, limit: 50)
      # We keep the heavy lifting in SQL so we only pull back the crossings we need.
      # For each metric we select snapshots up through the requested date, let Postgres
      # attach the previous snapshot's value to the current row (using WINDOW/LAG under
      # the hood), then filter down to snapshots that actually started on the target
      # date. That gives us the 'before' and 'after' values in one trip without
      # loading every historical snapshot into Ruby.
      results = {}
      connection = GrdaWarehouse::Monitoring::MetricSnapshot.connection
      limit_value = limit.to_i
      limit_value = 50 if limit_value <= 0
      quoted_date = connection.quote(calculation_date)

      active.find_each do |metric|
        next unless metric.alert_code

        lagged_relation = metric.metric_snapshots.
          where('metric_snapshots.initial_observation_date <= ?', calculation_date).
          select(
            'metric_snapshots.id',
            'metric_snapshots.entity_id',
            'metric_snapshots.initial_observation_date',
            'metric_snapshots.initial_value',
            "lag(metric_snapshots.current_value) OVER (PARTITION BY metric_snapshots.entity_id ORDER BY metric_snapshots.initial_observation_date, metric_snapshots.id) AS previous_value",
          )

        lagged_sql = lagged_relation.to_sql

        data_sql = <<~SQL
          SELECT entity_id,
                 initial_value AS current_value,
                 previous_value,
                 COUNT(*) OVER() AS total_count
          FROM (#{lagged_sql}) AS lagged_snapshots
          WHERE initial_observation_date = #{quoted_date}
            AND previous_value IS NOT NULL
          ORDER BY entity_id, id
          LIMIT #{limit_value}
        SQL

        rows = connection.exec_query(data_sql)
        next if rows.empty?

        total_count = rows.first['total_count'].to_i

        crossings = rows.map do |row|
          {
            entity_id: row['entity_id'],
            current_value: row['current_value'],
            previous_value: row['previous_value'],
          }
        end

        results[metric.alert_code] ||= {}
        results[metric.alert_code][metric.display_name] = {
          data: crossings,
          total_count: total_count,
          truncated: total_count > limit_value,
        }
      end

      results
    end
```

docs/implementation-working-documents/alerting-implementation-status.md
eanders: This is a good question.  I think I agree, once the project is complete, the artifact isn't particularly useful except to see some of the planning, but the overall goal should be captured in something like `docs/architecture/metric-tracking.md`.  I'll remove.

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
eanders: Let's keep that in our back pocket.  With the range based rows, I'm hoping these aren't huge, just large.  Given that will break standard polymorphic relationships, I think we should not implement it right now.

app/views/notify_user/metric_threshold_crossed.text.haml
eanders: We may want to allow opt-in/opt-out in the future.  It would be easy enough to have a section of the account edit screen to manage your subscriptions, but I think there are also use-cases where an administrator may want to control who is able to receive the notifications.  Everyone _can_ control if they receive the notifications as individual messages, digests, or only in-app.    Might be worth a cleanup of that note, but I'm not sure what would make it better.

app/views/notify_user/metric_threshold_crossed.text.haml
eanders: This is an interesting question.  I believe the request was to have them directly in the email, but I agree, from a product perspective, we'd rather force people into the app for details.  I'll update, we can revisit if that's not sufficient.

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
eanders: This should encode the partitions into the structure file, so for the next 10 years, it should be good.  If we need partitions beyond that we can make them in the future, but I'd guess we may have evolved this process by then.

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
eanders: Back of the napkin calculations indicated this would be worth while, though I believe those were before we moved to range-based storage.  It's possible this is overkill, but it didn't feel like it added too much complexity and is should scale and make future cleanup much simpler.

app/models/grda_warehouse/monitoring/metric_definition.rb
eanders: * find_or_create_by was intentional.  It means everyone has the same starting point, but if someone adjusts for their own installation it doesn't get overwritten when called again.
* Good call on the advisory lock, though I think this is pretty quick, it's just creating a handful of rows, so the likely hood of collision is relatively low.  That said, we now call it in the controller, so it is possible.

app/models/grda_warehouse/alert_definition.rb
eanders: * I think this is the same as `metric_definition.rb`, it should allow for drift (though I could definitely be wrong in this case)
* I'll add the lock, thanks!

app/models/grda_warehouse/monitoring/metric_definition.rb
eanders: Fixed it up with a cross lateral join. Coming soon.  Thanks!

db/warehouse/migrate/20251020142047_create_metric_snapshots.rb
ttoomey: We have the practice of occasionally collapsing our migrations. So if this migration is removed, I'm not sure how we will create these partitions.

ttoomey: On second thought, I guess it will be part of the schema? That should work. 10 years is a long time; I guess the failure case is it just overflows into the default table.

eanders: I'm counting on it being written to the structure.sql file.  And yes, there's a default, so if we forget or it takes us a while to get to creating more partitions, it won't fail in the meantime.  It's also easy enough to add 25 years instead of 10 if you think that's more appropriate.

ttoomey: should be fine as is. Thanks!

app/models/grda_warehouse/monitoring/metric_calculators/min_household_size_calculator.rb
eanders: I'm not 100% sure how effective this check will be, but the idea is that if someone is part of a multi-person household at some point, but then later their history no longer has them as part of any multi-person household, that something probably went wrong and should be investigated.  It intentionally includes exited enrollments (and ongoing) because it's looking to confirm history wasn't erased or significantly changed.

On an installation with 2,191,653 enrollments, this query takes roughly 5-10 seconds with no batching.

app/models/menu/menu.rb
eanders: Decided "Threshold Monitoring"

app/jobs/notify_metric_threshold_crossings_job.rb
eanders: Leaving as is for right now, we can use the DatabaseMailer artifacts to determine who received what, but acknowledging we need a more comprehensive message queuing system and history to prevent possible situations where the sender partially completes multiple times sending duplicates.

app/models/grda_warehouse/monitoring/metric_calculators/min_household_size_calculator.rb
ttoomey: Our HMIS support splitting/joining households (SplitHousehold mutation). So this is something we would expect to see

eanders: These aren't necessarily in place to only catch things we wouldn't expect, but to alert on situations that potentially warrant investigation.  Splitting/merging households is totally fine, and unless it changes the max or min household size it wouldn't trigger an alert.  Also, it may be that some installations just leave particular alerts disabled as they aren't relevant.

PR #5943, opened 2025-10-28
Mark form instances as inactive instead of deleting them

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8342

**Problem:**
There is currently no historical record of Form Instances. When they are deleted in the UI, they are hard-deleted. This makes auditing configuration history impossible.

**Proposed fix:**
* Instead of hard-deleting Instances, just mark them as `active: false`. This column already exists, and "inactive" rules are already ignored when checking for form applicability (I audited this, we do have test coverage too). Inactive rules are not currently shown in the UI, though technically the API has a filter for displaying them.
* Add Paper Trail to the form instance model. (Note: form instances are not actually _editable_ in the UI (Aside from marking inactive)).
Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1307

**Alternatives considered**:
- Adding "acts_as_paranoid", either on a new deleted_at column or onto the active column. This was a fair amount of churn. We already have the logic in place for ensuring only `active` rules are respected, so it felt simpler/safer to keep with that approach.

**How to test:**
- Add and remove form rules from the hmis admin tool

**Other changes:**
- Add a missing validation for service form instances
- Clean up a long-ago deprecated mutation

## Type of change
New feature


drivers/hmis/app/models/hmis/filter/form_instance_filter.rb
```diff
@@ -21,17 +23,15 @@ def filter_scope(scope)
...
       scope.where(definition_identifier: identifiers)
     end
   end
 
   def with_definition(scope)
     with_filter(scope, :definition) do
-      # TODO: filter active status / max version
-      identifiers = Hmis::Form::Definition.where(id: input.definition).pluck(:identifier)
-      scope.where(definition_identifier: identifiers)
+      identifier = Hmis::Form::Definition.find_by(id: input.definition)&.identifier
```
martha, line 33: If I'm understanding correctly, this is just cleanup because we already only accept 1 definition ID at a time on the [filter type](https://github.com/greenriver/hmis-warehouse/blob/ac10a63e7e0e8d4d88e10b35fe574ee17ce1a457/drivers/hmis/app/graphql/types/admin/form_rule.rb#L20), so filtering by multiple definitions was not only unused from the frontend, but not supported anyway. Does that sound right?

drivers/hmis/app/models/hmis/form/instance.rb
```diff
@@ -4,10 +4,42 @@
...
+#   which service(s) the form can collect
+#
+# Exclusive vs Inclusive Form Roles:
+# The application handles Form Instances differently based on whether the referenced Definition Role is
+# exclusive or inclusive. (That logic is defined in the frontend/graphql layer - see record_form_definition query,
+# assessment_form_definition query, etc).
+# - EXCLUSIVE roles (Project, Client, Enrollment): Only the single BEST matching form is displayed per context.
+#   The application picks one form using InstanceProjectMatch/InstanceEnrollmentMatch ranking logic.
+# - INCLUSIVE roles (Services, Custom Assessments): ALL matching forms are displayed.
+#   If any Form Instance enables a form for a project (by project, type, org, etc.), it appears.
```
martha, line 38: Thank you! 🤩 🙌

martha (approved):
Looks good to me!

ttoomey (approved):
Looks great!! Appreciate the documentation and additional tests

drivers/hmis/app/models/hmis/filter/form_instance_filter.rb
gigxz: Yep exactly

PR #5944, opened 2025-10-29
7846 add hoh only filter project group cohort

ttoomey:
## _Merging this PR_

### Description
Enhances cohort automation to support sub-population and head-of-household filtering, enabling users to create focused By-Name-Lists (e.g., "Veterans who are HoH") without manual client management.

- **Database**: Adds `automation_sub_population` and `automation_hoh_only` columns to support additional filter criteria beyond project groups
- **Model**: Refactors cohort maintenance to use the existing `FilterBase` infrastructure instead of direct enrollment queries, allowing filters to be combined (project group AND sub-population AND HoH status)
- **UI**: Adds filter controls to cohort edit form with human-readable descriptions of active automation criteria on the show page
- **Backward Compatibility**: Existing project-group-only cohorts continue working unchanged; new filters are optional additive constraints
- **Bug fix in project criteria**: if a project group is provided but has not projects, treat as excluding enrollments
- **Enhancements**: advisory lock around cohort maint for safety, add timestamp to track last automated refresh, adds ability to enqueue refresh (helps QA)

### Type of Change
New feature


app/models/grda_warehouse/cohort.rb
```diff
@@ -621,14 +656,37 @@ def maintain
...
-      incoming_client_ids = project_group.clients.
-        joins(:warehouse_client_source).
-        merge(GrdaWarehouse::Hud::Enrollment.open_on_date(Date.current)).
-        pluck(wc_t[:destination_id])
+      filter = build_automation_filter
+      # Note that we are using ServiceHistoryEnrollment here, but that model represents a denormalized
+      # view of enrollments that makes it easy to query for clients based on a variety of criteria
+      # and the resulting client_ids are for the `clients` table (GrdaWarehouse::Hud::Client)
+      # Use the :warehouse criteria set so project-scoped filters (like project groups)
+      # stay active; :client/:hud-only tags skip project filters and admit unrelated clients.
```
eanders: I don't understand this comment.

app/models/grda_warehouse/cohort.rb
```diff
@@ -621,14 +656,37 @@ def maintain
...
+        GrdaWarehouse::ServiceHistoryEnrollment.all,
+        tags: [:warehouse],
+        report_scope_source: nil,
+      ).pluck(:client_id).uniq
+
       to_remove = existing_client_ids - incoming_client_ids
       to_add = incoming_client_ids - existing_client_ids
-      remove_clients(to_remove, 'No longer enrolled in project group')
-      add_clients(to_add, 'Enrolled in project group')
+      remove_clients(to_remove, 'Client no longer matches automation criteria')
```
eanders: We should double check the string `'No longer enrolled in project group'` isn't used anywhere for reporting.

app/models/grda_warehouse/cohort.rb
```diff
@@ -73,7 +75,11 @@ class Cohort < GrdaWarehouseBase
     end
 
     scope :auto_maintained, -> do
-      where.not(project_group_id: nil)
+      table = arel_table
+      predicate = table[:project_group_id].not_eq(nil).
+        or(table[:automation_sub_population].not_eq(nil)).
+        or(table[:automation_hoh_only].eq(true))
+      where(predicate)
```
eanders: Given that we are using FilterBase below, could this take advantage of the `apply` method, or one of the other mechanisms the filter criteria provide to generate a scope?

If we can lean on the FilterBase, I think we can easily add more criteria in the future.

db/warehouse/migrate/20251027202455_add_automation_filters_to_cohorts.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+class AddAutomationFiltersToCohorts < ActiveRecord::Migration[7.1]
+  def change
+    # use names that won't conflict with Cohort.available_columns
+    add_column :cohorts, :automation_sub_population, :string
+    add_column :cohorts, :automation_hoh_only, :boolean, default: false
```
eanders, line 7: If we pivot to a `filter` json column, we could support additional filter options without needing to add columns to the table.  Even if for now we just set it to a has like: `{ automation_sub_population: :veteran, automation_hoh_only: false }` that feels like a good choice to me.   Am I missing something?

ttoomey: It seemed faster to get the form inputs and model validations working using distinct columns. We could store them as json but then it adds some complexity.

app/models/grda_warehouse/cohort.rb
ttoomey: Sorry, I'm not understanding. The `Cohort.auto_maintained` scope is just filtering the set of cohorts to those that have any automation filter configured. Where as `FilterBase.apply` is for filtering enrollment-like records. So I'm not sure that FilterBase is a good fit for the Cohort scope here

eanders (approved):
This is probably fine as is, but I left a few comments on thoughts for making this a bit more future-proof.
If you think those would add significantly to the build-time, feel free to go forward with what's here.

app/models/grda_warehouse/cohort.rb
ttoomey: we can probably lose the service history note; I'll simplify the tags comment, agree it is confusing

app/models/grda_warehouse/cohort.rb
ttoomey: Good thought, I don't see this string anywhere. I'll adjust the string to be similar

app/models/grda_warehouse/cohort.rb
eanders: Got it, that makes sense and seems fine. Thanks!

eanders (approved):

app/models/filters/criteria/filter_for_projects.rb
```diff
@@ -8,6 +8,7 @@ def applies?
   def apply(scope)
     scope = super(scope)
     project_ids = visible_project_ids
+    return scope.none if project_ids.blank? && input.project_group_ids.present?
```
ttoomey, line 11: this appears to be a bug, if a project group is provided but is empty, it is treated as thought no project group was provided.

eanders: Thanks!

eanders (approved):
Thanks @ttoomey !

ttoomey:
sorry @eanders - I made some more enhancements. Can you re-review?

app/controllers/cohorts_controller.rb
```diff
@@ -12,9 +12,9 @@ class CohortsController < ApplicationController
   include CohortClients
   include Search
   before_action :some_cohort_access!
-  before_action :require_can_configure_cohorts!, only: [:create, :destroy, :edit, :update, :new]
+  before_action :require_can_configure_cohorts!, only: [:create, :destroy, :edit, :update, :new, :maintain]
   before_action :require_can_access_cohort!, only: [:show]
-  before_action :set_cohort, only: [:edit, :update, :destroy, :show]
+  before_action :set_cohort, only: [:edit, :update, :destroy, :show, :maintain]
```
eanders, line 17: This is interesting, so accessing the cohort will queue a background job to ensure the cohort has the right set of people.

I _think_ that's ok.

eanders (approved):

app/controllers/cohorts_controller.rb
ttoomey: hmm, this should invoke `set_cohort` to initialize the `@cohort` variable before `maintain` is called. It shouldn't call `maintain` before `set_cohort` if that's what you mean

eanders: Right!  Sorry, read it in reverse.

PR #5950, opened 2025-10-31
8391-identify-duplicates-space-optimization

ttoomey:
## _Merging this PR_

### Description
Refactors the duplicate client matching SQL to improve performance and prevent database temp-table exhaustion.

**DOES NOT CHANGE QUERY IMPLEMENTATION**
We keep the old query in place here. This PR adds a new rake task to compare the output of the legacy and the refactored queries. This allows us the check equivalence before enabling the new query:
```bash
rails validate:duplicate_matching
```

A follow up will be needed to enable the functionality once this is deployed and tested (remove legacy functions, etc)

Details:
* Refactors the duplicate-matching SQL to eliminate the temp-table blowups caused by large duplicate clusters (e.g. thousands of clients sharing the same SSN/DOB). The new aggregation-based approach keeps those clusters in-memory arrays and avoids exhausting temp disk.
* Adds IdentifyDuplicatesQueryMatcher which centralizes the SSN/name/DOB queries for both :existing and :unprocessed matching paths
* Expands the IdentifyDuplicates specs with coverage that exercises the new matcher, ensuring regression protection for both query shapes and the IdentifyDuplicates integration flow.

### Type of Change
Bug fix


spec/models/grda_warehouse/data_source_spec.rb
```diff
@@ -155,7 +155,8 @@
       end
 
       it 'is not stalled when there was an import yesterday' do
-        imports.each.with_index { |import, i| import.update(completed_at: i.days.ago + 2.minutes) }
+        # 2-hour buffer so tests don't fail across DST boundaries
+        imports.each.with_index { |import, i| import.update(completed_at: i.days.ago + 2.hours) }
```
ttoomey, line 159: fails only on Nov 1st due to dst change

app/models/grda_warehouse/tasks/identify_duplicates_query_matcher.rb
```diff
@@ -0,0 +1,241 @@
...
+        # Default validate_ssn to true for SSN matches
+        validate_ssn = (field_alias == 'ssn') if validate_ssn.nil?
+
+        # Early return for unprocessed with no IDs
+        return [] if match_type == :unprocessed && unprocessed_ids.blank?
+
+        results = GrdaWarehouse::Hud::Client.connection.execute(to_sql(warehouse_id: warehouse_id))
+
+        # Filter SSNs if validation requested and SSN is in results
+        results = results.select { |r| ::HudHelper.util.valid_social?(r['ssn']) } if validate_ssn && include_value_in_results && field_alias == 'ssn'
```
eanders, line 133: isn't `&& field_alias == 'ssn'` duplicative?  Potentially someone could pass `validate_ssn: true` on some other field, but that seems unlikely.  That said, the check is not expensive.

app/models/grda_warehouse/tasks/identify_duplicates_query_matcher.rb
```diff
@@ -0,0 +1,241 @@
...
+        <<-SQL
+          FROM "Client" AS clients
+          INNER JOIN warehouse_clients ON clients.id = warehouse_clients.source_id
+          WHERE #{filters.join("\n            AND ")}
+        SQL
+      end
+
+      def unprocessed_sql
+        destination_ids_sql = destination_data_source_ids&.join(', ')
+        unprocessed_ids_sql = unprocessed_ids&.join(', ')
```
eanders, line 185: Are we concerned that these might contain hundreds of thousands of ids or more?  I think that's why we went to temp tables, but I might be mis-remembering.

eanders (approved):
Thanks @ttoomey ! 
I'm interested to see if this ends up being more performant or if it has issues when there are many unmatched clients.

app/models/grda_warehouse/tasks/identify_duplicates_query_matcher.rb
ttoomey: It looks like the old queries did the same thing; the `exact_[X]_for_unprocessed` methods interpolate `unprocessed_ids` into the query in the same way.

Agree that it may not be optimal; A temp table could be a good optimization but I don't see where that is implemented; could be missing it though.

PR #5958, opened 2025-11-03
Update vault_deploy to use assume and op-dev

gigxz:
## _Merging this PR_

## Description
Update vault_ssh and vault_deploy to be compatible with new dev setups:
1. Use `assume` instead of `aws-vault`
2. Expect profile to be named `op-dev` instead of `openpath` (this is a convention that I think has been adopted by the whole team, but please let me know if not)

**Why**
These scripts will be phased out once everything moves to EKS. The purpose is just to limit disruption to existing ECS deployment workflows. It also allows us to keep the check `Deployer.check_that_you_pushed_to_remote!` which wouldn't run if we ran the deploy script directly. 

**Notes**
**Confirm this is suitable for all op engineers before merging

I had to set `export GRANTED_ALIAS_CONFIGURED="true"` in my zshrc for the scripts to work, but that may just be me.

## Type of change
bug fix


dtgreiner:
Is it worth renaming these files from `vault_*` to `aws_*` or something similar? Really just semantics here, but we're moving away from using aws_vault, so maybe keep the file names showing as such too? either way, this looks good to me, thanks for doing it!

dtgreiner (approved):

eanders:
I had similar thoughts to @dtgreiner .  It seems odd to me that these are still named `vault_`, but given the timeline, see no real harm in it.  Maybe a comment in each file?  Something like:
```
# AWS vault has been superseded by granted.  This mechanism will be phased out in the coming months at which time we can remove this file completely.  Keeping the `vault_*` name for developer ergonomics during the transition.
```

eanders (approved):

gigxz:
good point, added elliot's comment suggestion!

martha (approved):
great, thanks!

> I had to set export GRANTED_ALIAS_CONFIGURED="true" in my zshrc for the scripts to work, but that may just be me.

me too

eanders:
@gigxz can we retarget this to 188 and merge?

PR #5963, opened 2025-11-04
Implied Consent - consent invalidation

dtgreiner:
## _Merging this PR_

## Description

Fix issue when using implied consent where invalidating_consent was clearing out housing_release_status. This should be going back to "Implied Consent" instead of nil in this scenario.

## Type of change

Bug Fix

### DIFF
--- a/app/models/grda_warehouse/hud/client.rb
+++ b/app/models/grda_warehouse/hud/client.rb
@@ -1068,13 +1068,19 @@ module GrdaWarehouse::Hud
     end
 
     def invalidate_consent!
+      hr_status = nil
+      # If using implied consent, we need to set the housing release status to the current consent type ("Implied Consent")
+      hr_status = GrdaWarehouse::Config.active_consent_class.new(client: self).current_consent_type if GrdaWarehouse::Config.implied_consent?
+
       update_columns(
         consent_form_id: nil,
-        housing_release_status: nil,
+        housing_release_status: hr_status,
         consent_form_signed_on: nil,
         consent_expires_on: nil,
         consented_coc_codes: [],
       )
+      # Remove any view caches for this client since permissions may have changed
+      clear_view_cache
     end
 
     def apply_housing_release_status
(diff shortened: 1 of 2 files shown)


app/models/grda_warehouse/hud/client.rb
```diff
@@ -1068,13 +1068,19 @@ def newest_consent_form
     end
 
     def invalidate_consent!
+      hr_status = nil
+      # If using implied consent, we need to set the housing release status to the current consent type ("Implied Consent")
+      hr_status = GrdaWarehouse::Config.active_consent_class.new(client: self).current_consent_type if GrdaWarehouse::Config.implied_consent?
```
eanders, line 1073: Are there some tests we can update so this doesn't happen again?  I thought we had some, but maybe not?

eanders (approved):
@dtgreiner this looks fine, but I'd like to see some regression tests so we don't hit this again.

app/models/grda_warehouse/hud/client.rb
dtgreiner: added some regression tests [here](https://github.com/greenriver/hmis-warehouse/pull/5963/commits/2f32fdefcca58ec7ab561eeb806659f590b00c64)

spec/models/grda_warehouse/hud/client_spec.rb
```diff
@@ -476,5 +476,65 @@ def whitespace_strict_search_expects(client)
...
+      end
+
+      context 'when using implied consent' do
+        before do
+          GrdaWarehouse::Config.delete_all
+          create(:config_va)
+          GrdaWarehouse::Config.invalidate_cache
+        end
+
+        it 'sets housing_release_status for implied consent when no consent file is present' do
```
eanders, line 516: I'm probably forgetting something, but shouldn't we just be treating `nil` as implied?  Rather than enforcing we get the status correct (though I see no reason not to try) if `nil` meant implied when consent was set to use implied consent, would that fix the issue?

eanders (approved):
Thanks @dtgreiner !

spec/models/grda_warehouse/hud/client_spec.rb
dtgreiner: I want to say I looked at that in a previous update. I don't fully remember. if there was a reason to not do it, maybe it was just the first fix was simpler to adjust. It's probably a good idea to take another look at it though, so I'll keep this in my queue to swing back around and take another look at why we aren't doing it that way.

PR #5968, opened 2025-11-05
Don't destroy unit occupancy ActiveRange on join/split

martha:
## _Merging this PR_

## Description
**Bug:** When splitting household, unit occupancy `ActiveRange` was getting deleted, instead of correctly updated with an end date.

**Root cause:** The `accepts_nested_attributes_for` has some behavior that we weren't accounting for. My understanding is that, with a `has_one` relationship, if there is an existing nested record and you _don't_ pass `id` in the nested attributes hash, then it creates a new record. (See inline comments)

**Proposed fixes:**
- Remove `accepts_nested_attributes_for` on `UnitOccupancy`, since we were not using it correctly, and I think this trickiness could be hard to remember in the future.
- Update usages to just set attributes on the nested record directly.
- Add `acts_as_paranoid` to `ActiveRange` (which already had a `deleted_at` column that wasn't being used)

## Type of change
Bug fix


drivers/hmis/app/graphql/mutations/join_household.rb
```diff
@@ -51,7 +51,8 @@ def resolve(receiving_household_id:, joining_enrollment_inputs:)
           enrollment.relationship_to_hoh = map_enrollment_id_to_relationship[enrollment.id.to_s]
 
           # Whether or not the receiving household has a unit assignment, clear the joining enrollment's current unit assignment
-          enrollment.active_unit_occupancy&.assign_attributes(occupancy_period_attributes: { end_date: Date.current })
+          # Update the occupancy period directly, instead of using enrollment.release_unit!, to avoid creating a nested transaction.
+          enrollment.active_unit_occupancy.occupancy_period.update!(end_date: Date.current, user: current_user) if enrollment.active_unit_occupancy.present?
```
martha: Question: Would it be better instead to call `enrollment.release_unit!`, even though it creates a nested transaction, because we actually do want to enqueue `track_availability` here? (unlike in the SplitHousehold mutation, which probably doesn't need to track availability, because the previously occupied unit is still occupied by the "remaining" household members who stayed behind after the split)

drivers/hmis/app/models/hmis/hud/processors/enrollment_processor.rb
```diff
@@ -81,7 +81,8 @@ def assign_metadata
       return if active_unit_occupancy&.unit_id == unit.id
 
       # If assigned to a different unit: unassign
-      enrollment.active_unit_occupancy&.assign_attributes(occupancy_period_attributes: { end_date: Date.current })
+      # Update the occupancy period directly, instead of using enrollment.release_unit!, to avoid creating a nested transaction.
```
martha: Question: Why does the form processor's `assign_unit` functionality silently release the previously assigned unit? Whereas, in other places when we call the _enrollment's_ `assign_unit` function, it raises if the enrollment is already assigned to another unit.

drivers/hmis/app/models/hmis/unit_occupancy.rb
```diff
@@ -29,7 +29,16 @@ class Hmis::UnitOccupancy < Hmis::HmisBase
   has_one :occupancy_period, class_name: 'Hmis::ActiveRange', as: :entity, dependent: :destroy
   # Service record that relates to this occupancy (likely a BedNight or BedNight-ish custom service)
   belongs_to :hmis_service, class_name: 'Hmis::Hud::HmisService', optional: true
-  accepts_nested_attributes_for :occupancy_period
+
+  # To discuss: accepts_nested_attributes_for has some trickiness that we weren't accounting for, so I'm proposing to remove it.
```
martha: Let's discuss this more. I think other alternatives would include (both untested):
- keep `accepts_nested_attributes_for` and add the `update_only` option
- keep `accepts_nested_attributes_for`, and pass an ID when updating an existing nested record

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -851,7 +851,7 @@ def process_record(record:, hud_values:, user:, definition:, save: true)
       hud_values = complete_hud_values.merge('currentUnit' => new_unit.id)
       process_record(record: e2, hud_values: hud_values, user: hmis_user, definition: definition)
       e2.reload
-      expect(old_uo.end_date).to be_present
+      expect(e2.unit_occupancies.first.end_date).to be_present
```
martha: The bug was getting masked in this test because `old_uo` was outdated. `old_uo.reload` also served to surface the bug.

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -506,14 +506,13 @@ def assign_unit(unit:, start_date:, user:, active_referral: nil)
...
     # include project id here since it may not be available during after_save hooks due to WIP
     self.unit_occupancy_changes = { project_id: unit.project_id, unit_type: unit.unit_type, user_id: user.id } if unit.unit_type
-    unit_occupancies.build(
-      unit: unit,
-      occupancy_period_attributes: {
-        start_date: start_date,
-        end_date: nil,
-        user: user,
-      },
+    occupancy = unit_occupancies.build(unit: unit)
```
martha: This change is necessitated by the change I'm proposing in `unit_occupancy.rb` -- see that thread for discussion/other alternatives.

martha:
I believe the test failure is unrelated as it's also failing on 188. https://github.com/greenriver/hmis-warehouse/pull/5960

eanders:
> I believe the test failure is unrelated as it's also failing on 188. #5960

Yes, sorry, that's fixed in 187, now pulled through to 188, if you update this branch it should go away.

drivers/hmis/app/graphql/mutations/join_household.rb
gigxz: I think this is fine, the unit syncing functionality is not needed anymore, so we can get rid of track_availability - I started a separated PR for fully removing unit sync integration https://github.com/greenriver/hmis-warehouse/pull/5974

drivers/hmis/app/models/hmis/hud/processors/enrollment_processor.rb
```diff
@@ -81,7 +81,8 @@ def assign_metadata
...
+      enrollment.active_unit_occupancy.occupancy_period.update!(end_date: Date.current, user: @processor.current_user) if enrollment.active_unit_occupancy.present?
```
ttoomey: This is fine as is. The code expects `active_unit_occupancy.occupancy_period` to always exist. But maybe we should make that expectation more explicit with a raise().

It might be a little cleaner if this could live in a method on enrollment

```ruby 
# Hmis::Hud::Enrollment
# callsites: join_household, split_household, enrollment_processor (could track_avail too but sounds like we're removing that)
def close_active_unit_occupancy!(end_date: Date.current, user:)
  raise "Missing occupancy period for enrollment #{enrollment.id}" unless active_unit_occupancy&.occupancy_period

  active_unit_occupancy.occupancy_period.update!(…)
end

```

drivers/hmis/app/models/hmis/unit_occupancy.rb
ttoomey: I think removing `accepts_nested_attributes` is a nice simplification

ttoomey (approved):

drivers/hmis/app/models/hmis/hud/processors/enrollment_processor.rb
gigxz: Changing from assign_attributes to update! means EnrollmentProcessor will persist changes immediately, which it didn't before. Will this cause any unintended side effects?
For example, we run the processors in validation-only scenarios (see SubmitAssessment mutation for example) . In those scenarios we run the processors and return early with validations; the assumption is processors don't persist changes so we don't Rollback. This pattern is not ideal, maybe it would be better if form processors saved everything and the caller rolled back when necessary. (But that's not how it currently works iirc)

drivers/hmis/app/models/hmis/hud/processors/enrollment_processor.rb
gigxz: The Enrollment processor "current_unit" value is used in forms that change an enrollment's assigned unit. If you are changing the assigned unit, the previous unit should be released, which is why this behavior is here.
I'll drop a link to https://github.com/open-path/Green-River/issues/8505 since its related and I just fleshed it out a bit more

drivers/hmis/app/models/hmis/hud/processors/enrollment_processor.rb
martha: Thanks! Hm, this is looking very similar to the existing `enrollment.release_unit!` helper, especially once Gig's changes to remove `track_availability` are merged:
https://github.com/greenriver/hmis-warehouse/pull/5974/files#diff-004a854c99341342b5969edfc1d169d54e53ad5bad3af748fc87af825fd0f11cR509-R515

I'll propose incorporating just that part of Gig's changelist into this PR, and then updating `release_unit!` to raise if an occupancy with no occupancy period is found. https://github.com/greenriver/hmis-warehouse/pull/5968/commits/e907b0f6bb596003986badd8fd04b3f9e4b18c64

martha: Thanks, that's helpful! I had to add some `autosave`s to get this wired up correctly, but I think it's working right. I adjusted an existing spec test to help check. https://github.com/greenriver/hmis-warehouse/pull/5968/commits/7d1da7201dcd3e920187a02e99022d2c6acb4faf

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -481,7 +481,7 @@ def track_unit_occupancy_changes
   def assign_unit(unit:, start_date:, user:, active_referral: nil)
     raise "Active referral unit is incorrect #{active_referral.opportunity.unit.id} != #{unit.id}" if active_referral.present? && active_referral.opportunity.unit != unit
 
-    current_occupancy = active_unit_occupancy.present? if active_unit_occupancy&.occupancy_period&.active?
+    current_occupancy = active_unit_occupancy if active_unit_occupancy&.occupancy_period&.active?
```
martha, line 484: I believe this was an existing bug, that we haven't exposed/run into yet. I added some tests below in `enrollment_spec.rb` to confirm the right behavior, let me know how it looks.

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
```diff
@@ -813,14 +813,21 @@ def process_record(record:, hud_values:, user:, definition:, save: true)
...
+      old_unit = create(:hmis_unit, project: p1)
+      e2.assign_unit(unit: old_unit, start_date: 1.week.ago, user: hmis_user)
+      e2.save!
+      e2.reload # reloading makes sure the enrollment's active_unit_occupancy is up-to-date
+      expect(e2.active_unit_occupancy.unit).to eq(old_unit)
+
+      # Use the form processor to reassign the enrollment to new_unit, but without saving
+      new_unit = create(:hmis_unit, project: p1)
+      hud_values = complete_hud_values.merge('currentUnit' => new_unit.id)
+      process_record(record: e2, hud_values: hud_values, user: hmis_user, definition: definition, save: false)
```
gigxz:
```suggestion
      expect do
         process_record(record: e2, hud_values: hud_values, user: hmis_user, definition: definition, save: false)
     end.not_to change(Hmis::UnitOccupancy, :count)
```

Maybe worth adding this to ensure no occupancy is added? To prevent a silent failure where new_unit and old_unit are somehow both assigned, but line 830 still passes because old unit is considered active

drivers/hmis/spec/models/hmis/form/form_processor_spec.rb
gigxz: Ah good find! I think it would be safer to compare with the record directly rather than `unit_occupancies.first` because unit_occupancies is not guaranteed to be sorted in a particular way, from what I understand.

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -470,4 +470,62 @@
       end
     end
   end
+
+  describe 'assign_unit' do
+    let!(:project) { create(:hmis_hud_project, data_source: ds1) }
+    let!(:enrollment) { create(:hmis_hud_enrollment, project: project, data_source: ds1) }
+    let!(:unit) { create(:hmis_unit, project: project) }
+    let(:hmis_user) { user.related_hmis_user(ds1) }
```
gigxz:
```suggestion
```
No need to redefine since its included in hmis_base_setup

drivers/hmis/spec/models/hmis/hud/enrollment_spec.rb
```diff
@@ -470,4 +470,62 @@
       end
     end
   end
+
+  describe 'assign_unit' do
```
gigxz, line 474: Thanks for adding these!

drivers/hmis/spec/requests/hmis/join_household_spec.rb
```diff
@@ -219,6 +219,10 @@ def perform_mutation(
           donor_child.reload
         end.to change(donor_hoh, :active_unit_occupancy).to(nil).
           and change(donor_child, :active_unit_occupancy).to(nil)
+
+        expect(donor_hoh.unit_occupancies.count).to eq(1) # Still has historical occupancy (regression #8379)
+        expect(donor_hoh.unit_occupancies.sole.occupancy_period).to be_present
+        expect(donor_hoh.unit_occupancies.sole.end_date).to be_present
       end
```
gigxz, line 228: It seems like both enrollments are meant to release their units (donor_hoh and child_hoh) so I'd suggest updating the test to test that. We can add the assertions to the existing "expect" block which I think is a little neater:

```suggestion
      let(:today) { Date.current }

      it 'removes the current unit from the joiners' do
        expect do
          perform_mutation
          [donor_hoh, donor_child, donor_hoh_occupancy, donor_child_occupancy].each(&:reload)
        end.to change(donor_hoh, :active_unit_occupancy).to(nil).
          and change(donor_child, :active_unit_occupancy).to(nil).
          # Unit is released from the joiners, and historical unit occupancy is still present (regression #8379)
          and change { donor_hoh_occupancy.occupancy_period.end_date }.from(nil).to(today).
          and change { donor_child_occupancy.occupancy_period.end_date }.from(nil).to(today)
      end
```

gigxz (approved):
Looks great!
Tested join/split with units locally as well as a general unit assignment behavior (added note to https://github.com/open-path/Green-River/issues/8379#issuecomment-3508338038 for QA)

Left a few notes just on tests.

drivers/hmis/spec/requests/hmis/join_household_spec.rb
martha: https://github.com/greenriver/hmis-warehouse/pull/5968/commits/d82db7351daa1d069aaf41f97196b05ea2087cf5

martha:
Thanks both for the feedback! I think I've addressed everything, I'll wait for the test suite and then merge.

PR #5976, opened 2025-11-06
8494 POC date picker improvement

ttoomey:
## _Merging this PR_

### Description
Adds a keyboard-friendly and accessible `date_picker2` input for improved date entry. Keeps backwards compatibility with existing date format while providing more robust user input handling and improved accessibility over the current `date_picker`

### Type of Change
New feature


app/inputs/date_picker2_input.rb
```diff
@@ -0,0 +1,213 @@
...
+  end
+
+  def hidden_input_html_options
+    @hidden_input_html_options ||= begin
+      options = input_html_options.deep_dup
+
+      options[:data] ||= {}
+      options[:data][:datepicker2_target] = 'hiddenInput'
+      options[:data].delete(:date_options)
+      options[:data].delete('date_options')
```
eanders, line 69: Do we need both of these? If so, why don't we need string and symbol for `:required`?

app/javascript/controllers/datepicker2_controller.js
eanders, line 1: This feels like a lot of code, which is probably why people use date picker libraries.  It looks like it can enforce min and max years, can we also enforce min and max dates?

ttoomey: It is some code although not that much more lines than the old picker. But would benefit for some pruning; I'm sure there's slop in there. This is a POC to give us something to react to; see if the UX is better. If it seems worth pursuing we can clean it up.

I also had a variant with three fields but the labels started go get awkward in our current form layout

ttoomey:
closing as we can't prioritize this right now

PR #5992, opened 2025-11-11
8518 missing ds in service history builder

ttoomey:
## _Merging this PR_

## Description

Fixes crash on missing data source during service history build. Adds automated cleanup for service history records associated with soft-deleted data sources and improves test coverage for service history enrollment rebuilds.

* Adds test coverage for `rebuild_service_history!` 
* Implements service object to purge `ServiceHistoryEnrollment` and `ServiceHistoryService` records for soft-deleted data sources

### Type of Change

New feature


app/models/grda_warehouse/tasks/service_history/purge_for_deleted_data_sources.rb
```diff
@@ -0,0 +1,103 @@
...
+    private
+
+    def find_deleted_data_sources
+      # Find data sources that:
+      # 1. Have been soft-deleted (deleted_at is present)
+      # 2. Were deleted before the retention date
+      GrdaWarehouse::DataSource.
+        with_deleted.
+        where.not(deleted_at: nil).
+        where('deleted_at < ?', @retain_at).
```
eanders: Can we stick with non-string interpolation for simple queries?
```suggestion
        where(deleted_at:  ..@retain_at).
```

app/models/grda_warehouse/tasks/service_history/purge_for_deleted_data_sources.rb
```diff
@@ -0,0 +1,103 @@
...
+      setup_notifier('Purge Service History for Deleted Data Sources')
+    end
+
+    def call
+      deleted_data_source_ids = find_deleted_data_sources
+
+      if deleted_data_source_ids.empty?
+        log 'No deleted data sources found with service history to purge'
+        return { enrollments_deleted: 0, services_deleted: 0 }
+      end
```
eanders: I'm not sure we need to log for this.  This is going to happen 99.999% of the time, so I think we can just return silently.

app/models/grda_warehouse/tasks/service_history/purge_for_deleted_data_sources.rb
```diff
@@ -0,0 +1,103 @@
...
+
+      count = 0
+      enrollment_scope.in_batches(of: 1000) do |batch|
+        enrollment_ids = batch.pluck(:id)
+        next if enrollment_ids.empty?
+
+        count += GrdaWarehouse::ServiceHistoryService.
+          where(service_history_enrollment_id: enrollment_ids).
+          delete_all
+      end
```
eanders, line 90: I'd be tempted to see how the DB does with a subquery here.  This might be fast, but it might be faster to simply:

```
enrollment_ids = GrdaWarehouse::ServiceHistoryEnrollment.
  where(data_source_id: data_source_ids).select(:id)
count = GrdaWarehouse::ServiceHistoryService.
    where(service_history_enrollment_id: enrollment_ids).
     delete_all
```

It will depend on how the query planner decides to do things.

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -73,6 +73,10 @@ def _perform
       end
 
       run_maintenance_task('Generate service history and related records') do
+        # Purge service history for deleted data sources before generating new records
+        GrdaWarehouse::Tasks::ServiceHistory::PurgeForDeletedDataSources.call
+        @notifier.ping('Purged service history for deleted data sources')
```
eanders: Since the service class also has the notifier setup, I'd be tempted to simply let it do the notifying, though I can see some reasons for also pinging here.

eanders (approved):

app/models/grda_warehouse/tasks/service_history/purge_for_deleted_data_sources.rb
ttoomey: I was a little nervous about one big delete on the SH services table since its has the possibility of being enormous; figured lots of small transactions are probably easier for the db to digest.

PR #5996, opened 2025-11-12
Remove MPER import project job

martha:
## _Merging this PR_

## Description
Remove no-longer-used custom importer

## Type of change
Code clean-up


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mper.rb
```diff
@@ -8,9 +8,6 @@
 
 ## MPER (to be removed)
 #
-# If the MPER credential is active, the ImportProjectsJob imports batch files from MPER nightly.
-# The files update the Project table and some related tables, including initial unit capacity.
-#
 # Note: this class may still be used for external ID lookups even if the MPER credential is no longer active.
```
martha, line 11: My understanding from the Note here is that this file should not be removed, let me know if that's right

gigxz: It's probably worth digging into this a bit more to see why we would need to keep it. Is there in-use code that relies on it still? Now that we removed unit syncing, this might not be needed anymore.

gigxz (approved):
Looks good, thanks!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mper.rb
martha: Hm, it was still used in a few jobs CreateReferralJob and CreateReferralRequestJob. Can those be removed now too? https://github.com/greenriver/hmis-warehouse/pull/5996/commits/cf21da2843c505b681621db49687f736c36fa05d

gigxz: I see. I want to double-check with client before removing the referral creation endpoint, so lets leave that as-is and leave the Mper class until we remove those Apis.

gigxz (changes requested):
add back referral creation endpoint

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mper.rb
martha: Sounds good, reverted

gigxz (approved):

martha:
Thanks! I'll retarget to 190 once available, since it sounds like staging deploys of 189 are already underway, and there is no rush to push this since it's just deleting unused code.

PR #5997, opened 2025-11-12
Add util for closing and reminting stale opportunities

martha:
## _Merging this PR_

## Description
- Add `Hmis::Ce::OpportunityRefresher` service class and associated rake task
- Refactor `build_opportunity_for_unit` into a helper on the unit class

How to test:
- Create a unit and mark it as accepting referrals.
- Add or modify a match rule that applies to that unit.
- Trigger reprocessing: `Hmis::Ce::Match::CandidatePoolBuilder.call(force_reprocessing: true)`, `Hmis::Ce::ProcessPoolsJob.perform_now`
- In the UI you can now see that this unit's "eligibility rules" are out of date
- Run the `ce_opportunity_refresh` rake task; the out-of-date unit should be updated.

## Type of change
New feature


drivers/hmis/app/models/hmis/ce/opportunity_refresher.rb
```diff
@@ -0,0 +1,58 @@
...
+    # @return [Hash] {
+    #   closed_count: Integer,
+    #   closed_opportunity_unit_ids: Array<Integer>,
+    #   created_count: Integer,
+    #   created_opportunity_ids: Array<Integer>,
+    #   skipped_count: Integer,
+    #   skipped_opportunity_ids: Array<Integer>
+    # }
+    def refresh_stale_opportunities(candidate_pool_ids: nil)
+      stale_scope = Hmis::Ce::Opportunity.stale.active
```
gigxz: Might we want to just stick to refreshing `open` opportunities, rather than including `locked` opportunities here and then filtering them out below on line 25? 

My understanding of the opportunity statuses is:
`open` = no active referral, "accepting referrals"
`locked` = has active referral
`closed` = closed and not accepting referrals

Maybe I'm missing some reasoning about wanting to have logging to identify the stale locked opportunities?

martha: Sure, that seems reasonable! And your understanding matches mine. 

I think I've run into this question before at some point -- whether to trust that `open` status always means no active referral while `locked` means there is a referral, or to actually check the `active_referral` association and confirm. Maybe this points to a code smell? But I think in this case you're right that we can just scope to `open` opportunities. 

Regarding the reason to log skipped opportunities: I was just doing this out of interest, since (I believe) if a referral is in progress but gets declined, that opportunity is now open again but is still stale. That might lead to some confusion with us (or eventually users) thinking "oh, I thought I refreshed all the opportunities, it didn't work?" However, probably a better way to track this would be to add some verbiage/warning alert on the Unit (opportunity) page noting that the requirements have gotten out of date. edit: something like this https://github.com/greenriver/hmis-frontend/pull/1320

drivers/hmis/app/models/hmis/ce/opportunity_refresher.rb
```diff
@@ -0,0 +1,50 @@
...
+    # @param candidate_pool_ids [Array<Integer>, nil] Optional array of candidate pool IDs to filter by.
+    #   If nil, processes ALL stale opportunities across the app.
+    # @return [Hash] {
+    #   closed_count: Integer,
+    #   closed_opportunity_unit_ids: Array<Integer>,
+    #   created_count: Integer,
+    #   created_opportunity_ids: Array<Integer>,
+    # }
+    def refresh_stale_opportunities(candidate_pool_ids: nil)
+      scope = Hmis::Ce::Opportunity.stale.active.open
```
gigxz:
```suggestion
      scope = Hmis::Ce::Opportunity.stale.open
```
Can remove redundant scope (`where.not(status: 'closed').where(status: 'open')`)

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -162,6 +162,26 @@ def to_pick_list_option
...
+    unit_desc = unit_type&.description
+    opportunity_name = "Unit #{id}#{unit_desc ? ' - ' : ''}#{unit_desc}"
+
+    rules = rule_resolver.rules_for_unit_group(unit_group)
+
+    Hmis::Ce::Opportunity.new(
+      unit: self,
+      project: project,
+      name: opportunity_name,
+      candidate_pool_id: unit_group.candidate_pool_id,
```
gigxz: Look at what happens when:
- unit group didn't have a candidate pool when opportunity was minted, and now it does? does the opportunity get marked stale? (i think in practice this happens like ProjectCeConfig updated => candidate pool assigned to unit group => existing opportunity should be stale or add the candidate pool?)

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -162,6 +162,26 @@ def to_pick_list_option
     { code: id, label: display_name }
   end
 
+  # Build an unsaved CE Opportunity for this unit
+  def build_ce_opportunity(rule_resolver: Hmis::Ce::Match::UnitGroupRuleResolver.new)
+    raise 'Unit already has an active opportunity' if latest_opportunity&.active?
+    raise 'Unit must be in a Unit Group to build CE opportunity' unless unit_group
+    raise 'Unit Group has no Workflow Template' unless unit_group.workflow_template || unit_group.direct_referral_workflow_template
```
gigxz: Maybe we could encapsulate this logic within the unit group, like `unit_group.any_workflow_template?`. That would make it easier to maintain in case we end up adding additional workflow template types later.

drivers/hmis/app/models/hmis/unit.rb
```diff
@@ -162,6 +162,26 @@ def to_pick_list_option
...
+
+    unit_desc = unit_type&.description
+    opportunity_name = "Unit #{id}#{unit_desc ? ' - ' : ''}#{unit_desc}"
+
+    rules = rule_resolver.rules_for_unit_group(unit_group)
+
+    Hmis::Ce::Opportunity.new(
+      unit: self,
```
gigxz: Since the name of this method mimics the built-in `build..` helpers, I think it would make sense to build this onto the unit directly since that is the typical expectation of `build..` methods:
```suggestion
    opportunities.build(
```

drivers/hmis/app/models/hmis/unit.rb
gigxz, line 185: I went on a bit of a rabbit hole trying to remember how this works, so it might be worth adding a comment stating that the UnitGroup may or may not have a candidate_pool_id. Unit groups in projects supporting waitlist-based referrals typically have an associated candidate pool (once processing has run), and unit groups in projects only supporting direct referrals typically do not have an associated candidate pool (but they may have one left over if they used to support waitlists).

I opened https://github.com/open-path/Green-River/issues/8555 to clarify that behavior a bit and see if we need to add any more constraints or routines for when those configs change.

drivers/hmis/app/models/hmis/ce/opportunity_refresher.rb
```diff
@@ -0,0 +1,50 @@
...
+          result = Hmis::Ce::Opportunity.import!(opportunities_to_create)
+          created_ids = result.ids
+        end
+
+        {
+          closed_count: closed_opportunity_unit_ids.length,
+          closed_opportunity_unit_ids: closed_opportunity_unit_ids,
+          created_count: created_ids.length,
+          created_opportunity_ids: created_ids,
+        }
```
gigxz, line 51: Is it ever expected that the number of opportunities closed and the number of opportunities created doesn't match? When would that come up?

If it's not expected, I think we should raise and rollback if that happens.

Again, I am not sure of the answer of my question, but if the numbers are expected to match maybe the logged object could be simplified:
```
        raise if closed_opportunity_unit_ids.uniq.size != created_ids.size
        {
          num_refreshed_units: closed_opportunity_unit_ids.size,
          refreshed_unit_ids: closed_opportunity_unit_ids,
          created_opportunity_ids: created_ids,
        }
```

drivers/hmis/lib/tasks/ce_opportunity_refresh.rake
```diff
@@ -0,0 +1,47 @@
...
+# Helper task to close stale CE opportunities and remint fresh ones
+# usage:
+# - rails driver:hmis:refresh_stale_opportunities
+# - passing optional candidate_pool_ids: rails driver:hmis:refresh_stale_opportunities[1,2,3]
+desc 'Close stale CE opportunities and create fresh ones with updated rules'
+task :refresh_stale_opportunities, [:pool_ids] => [:environment] do |_t, args|
+  raise 'CE is not enabled' unless Hmis::Ce.configuration.enabled?
+
+  pool_ids = args[:pool_ids]&.split(',')&.map(&:to_i)
+
```
gigxz, line 13: maybe add a check that the candidate pools exist? I could see accidentally passing different ids (like project ids or unit group ids) to this routine, which would have unexpected behavior.

edit: i suppose that check could go in OpportunityRefresher? my concern is really with the rake task since the arg isn't named

drivers/hmis/spec/models/hmis/hud/unit_spec.rb
```diff
@@ -178,4 +178,74 @@
...
+        rule_ids = opportunity.assignment_rules.map { |r| r['id'] }
+        expect(rule_ids).to contain_exactly(eligibility_rule.id, priority_rule.id)
+      end
+    end
+
+    context 'with unit group that has no rules' do
+      before do
+        Hmis::Ce::Match::CandidatePoolBuilder.call
+        unit_group.reload
+        expect(unit_group.candidate_pool).to be_nil # no rules = no candidate pool
```
gigxz: I think `no rules = no candidate pool` is not really true. Here's an example that I tested locally on this branch:
- add a match rule at the data source level
- go to project that only accepts direct referrals
- mark units as available. it will create opportunities that have assignment_rules but no candidate pool id.


Another example of this could be that the unit group was recently marked as supporting waitlists but the processing hasn't completed yet; so it has rules but is not yet associated to a candidate pool.

This is why I opened https://github.com/open-path/Green-River/issues/8555 because it feels like there is not clarity on what the expectations are. That said I think its probably fine (and good?) that opportunities get stored with assignment rules in all cases. It might make sense to just split this test out to test each case separately (unit group with no candidate pool; unit group with no match rules).

gigxz (approved):
Looks great! Just a few minor comments

drivers/hmis/app/models/hmis/unit.rb
martha: Great point and thanks for the ticket. I did add a comment along the lines of what you're suggesting, with a todo linking it to that ticket so we remember to update the comment if the behavior changes. https://github.com/greenriver/hmis-warehouse/pull/5997/commits/1d8f5914bb01612cff000b172169684a0e7e0eda

drivers/hmis/app/models/hmis/ce/opportunity_refresher.rb
martha: Oh, yep you are right it's unexpected! This extra bulky logging object was a relic of AI that I didn't critically cut down ✂️ . I'll make the change you're recommending.
- if an opportunity was recreated in between the close and create steps - no, this won't happen since it's inside of a transaction
- if there were somehow multiple stale opportunities per unit? - no, this shouldn't happen since there should only be one open opportunity per unit, which we validate on the opportunity

drivers/hmis/lib/tasks/ce_opportunity_refresh.rake
martha: Good plan, I added this (https://github.com/greenriver/hmis-warehouse/pull/5997/commits/1a48abb9d720d801b4eb11ff84985f57845c160d)

While testing that out, I also noticed that apparently passing a comma-separated list of args to a rake task is not as straightforward as I thought 🤦‍♀️ .  Updated the usage comment https://github.com/greenriver/hmis-warehouse/pull/5997/commits/3ed4d94a5542bff0133371c2cf743a0b5a56026c

drivers/hmis/spec/models/hmis/hud/unit_spec.rb
martha: I see, that's a good point this comment is inaccurate. I removed it and separated out the tests as you recommended.

I haven't added a test for the case where the unit group has a candidate pool, but no match rules. I couldn't think of a scenario for this other than when all rules were deleted and the candidate pool builder hasn't finished processing yet, which doesn't seem like it warrants its own test case (?) but lmk what you think! https://github.com/greenriver/hmis-warehouse/pull/5997/commits/3c3dd5f984766c47054ce09436ac92f50e97bf44

drivers/hmis/app/models/hmis/ce/opportunity_refresher.rb
gigxz: Good point, and displaying the stale flag seems like a good solution.

Agreed about opportunity status, we don't have validations ensuring that the status is in line with the `active_referral` presence. It looks like opportunity status is updated [from the engine](https://github.com/greenriver/hmis-warehouse/blob/30816020e297a52c3f2d64e900c2ee1625eb3104/drivers/hmis/app/models/hmis/ce/referral_message_handler.rb#L64-L87). Currently there isn't a way to delete a referral (I don't think? though depending on what is proposed in https://github.com/greenriver/hmis-warehouse/pull/6000 we may introduce it), but that is one scenario where this bug could appear (eg. Referral is deleted and opportunity is not closed). I can try to address that additional validation in the PR

gigxz (approved):

PR #5999, opened 2025-11-12
Resolve source CE referral on enrollment

martha:
## _Merging this PR_

## Description
- Repurpose the existing `can_view_summary?` referral policy for users who can view the target enrollment, even if they can't necessarily view details of the referral. (This policy was originally designed to display some limited data about the referral to a user who can send direct referrals from the source project, but can't view referrals in the target project.)
  - I thought this was ergonomically nice because the frontend can then resolve the `CeReferral` schema object and use its `access` field to determine whether to link to the referral from the enrollment.
  - I reviewed the list of summary fields on `CeReferral` that this would give the user access to, and they all seem okay to expose, especially since under this new use case, the user already can view enrollments in the target project.
  - Alternative, if we are concerned that this repurposing exposes too much data: We could just resolve the referral-related fields needed (`sourceCeReferralId`, `sourceCeReferralProjectName`, etc.) directly on the enrollment.
- Add `source_ce_referral` assoc on enrollment and resolve it on the graphql type

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1319

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -43,9 +43,10 @@ def self.authorized?(object, ctx)
     summary_field :custom_status, HmisSchema::CeCustomReferralStatus, null: true
     summary_field :client_id, ID, null: false
     summary_field :client_name, String, null: true, description: 'The name of the referred client. Always available to those who can view the full referral, even without full client record access.'
-    # Special case: Client is a "summary field" because it doesn't require referral visibility, but resolving it does require permission to view the client record.
     summary_field :created_at, GraphQL::Types::ISO8601DateTime, null: false
     summary_field :source_enrollment_id, ID, null: false
+    # source_project_name is resolved separately from source_enrollment as a summary field to minimize data exposure
+    summary_field :source_project_name, String, null: false
```
martha: I'm proposing to add `source_project_name` as a new summary-level field, even though it is also resolvable through the `sourceEnrollment` field. As we currently have it set up, `sourceEnrollment` is not a summary-level field, so it's only resolvable by those who can view the referral. I thought about elevating it to a summary field, but it contains a lot of data (like assessments and household members) that are not needed to solve the current requirement.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -43,9 +43,10 @@ def self.authorized?(object, ctx)
     summary_field :custom_status, HmisSchema::CeCustomReferralStatus, null: true
     summary_field :client_id, ID, null: false
     summary_field :client_name, String, null: true, description: 'The name of the referred client. Always available to those who can view the full referral, even without full client record access.'
-    # Special case: Client is a "summary field" because it doesn't require referral visibility, but resolving it does require permission to view the client record.
```
martha, line 46: Removing unrelated - I think this comment was outdated, since `client` is _not_ a summary-level field.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: I think this makes sense! I needs to be nullable though, because source_enrollment_id is not required on the referral table. (source_enrollment_id graphql field also should be nullable, which I incorporated into wip https://github.com/greenriver/hmis-warehouse/pull/6000 , but you can make the change here too).

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -221,6 +222,14 @@ def source_enrollment
       OpenStruct.new(enrollment: enrollment, definition_identifiers: [])
     end
 
+    def source_project_name
+      return unless object.source_enrollment_id
+
+      enrollment = load_ar_association(object, :source_enrollment)
+      project = load_ar_association(enrollment, :project)
```
gigxz, line 232: Until we resolve the dangling references issue, `enrollment` can be nil so we should handle that case here (https://github.com/greenriver/hmis-warehouse/pull/6001)

drivers/hmis/app/graphql/types/hmis_schema/enrollment.rb
```diff
@@ -267,6 +268,12 @@ def source_referral_posting
       load_ar_association(object, :source_postings).min_by(&:id)
     end
 
+    def source_ce_referral
+      # Skip checking permission here, because the current user can view the referral's target enrollment (this object),
+      # so we know they have at least can_view_summary? on the referral.
+      load_ar_association(object, :source_ce_referral)
+    end
```
gigxz, line 277: I think it would probably be worthwhile to make this policy check explicit; It shouldn't be a big cost because of the caching. That way if we change the logic of the can_view_summary method, we can track it back to this use-case, avoiding a scenario where can_view_summary became more restrictive but we still attempt to resolve it.

```suggestion
    def source_ce_referral
      source_referral = load_ar_association(object, :source_ce_referral)
      return unless source_referral # enrollment did not originate from a ce referral

      policy = current_user.policy_for(source_referral, policy_type: :ce_referral)
      source_referral if policy.can_view_summary?
    end
```

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -102,6 +102,9 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
 
   # All referrals where this enrollment is the source enrollment. NOT only 'direct' referrals
   has_many :outgoing_ce_referrals, class_name: 'Hmis::Ce::Referral', foreign_key: :source_enrollment_id
+  # The referral that created this enrollment (if any)
+  has_one :source_ce_referral, class_name: 'Hmis::Ce::Referral', foreign_key: :target_enrollment_id
```
gigxz, line 106: Thanks for adding this. This is a bit unrelated, but we have the same "dangling reference" problem as we do with the outgoing_ce_referrals association. If the target Enrollment is destroyed, the source enrollment will continue to have a `target_enrollment_id` that doesn't resolve to anything.

I will address this in https://github.com/greenriver/hmis-warehouse/pull/6000 since you're just adding the reference and not changing the default behavior here, it will need more tests etc. Just a note to self! I think we want `dependent: :nullify` In other words if the Target Enrollment is deleted, the referral should remain but have the `target_enrollment_id` nullified.

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -39,7 +39,13 @@ def can_view?
   def can_view_summary?
     return false unless Hmis::Ce.configuration.enabled?
```
gigxz, line 41: Can we add a `return true if can_view?`
It doesn't feel logical to have scenarios where the use can view the full referral, but they can't view the summary of it.

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -39,7 +39,13 @@ def can_view?
...
 
-    source_project_permissions.include?(:can_manage_outgoing_referrals)
+    # Users who can manage outgoing referrals from the source project
+    return true if source_project_permissions.include?(:can_manage_outgoing_referrals)
+
+    # Users who can view the target enrollment. Bakes in the assumption that the target enrollment is in the referral's project.
+    return true if referral.target_enrollment_id.present? && project_permissions.include?(:can_view_enrollment_details)
```
gigxz: If we had an Enrollment Policy, it would be nice if we could check against that here. Maybe we can add a TODO referencing https://github.com/open-path/Green-River/issues/8549 for that.

Full Enrollment visibility currently requires both can_view_enrollment_details and can_view_project so we should check that both of those are included in project_permissions to be comprehensive. (This is the thing that would be ideal to encapsulate in the policy!)

drivers/hmis/spec/models/hmis/auth_policies/ce_referral_policy_spec.rb
```diff
@@ -170,6 +170,59 @@
...
+    end
+
+    context 'when user can manage outgoing referrals on source project' do
+      it 'returns true' do
+        create_access_control(user, source_project, with_permission: [:can_manage_outgoing_referrals])
+        expect(policy.can_view_summary?).to be true
+      end
+    end
+
+    context 'when referral has a target enrollment' do
```
gigxz: Thanks for these tests! I think we are missing coverage for the case where the referral doesn't have a target enrollment. We should test that because I think we are wanting to limit summary visibility to only referrals that have resulted in an enrollment at the target project.

gigxz (commented):
Looks good!

drivers/hmis/app/models/hmis/hud/enrollment.rb
martha: I agree, that makes sense!

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
martha: Cool thanks for tracking down the open ticket! I'll add a todo and also add an extra check that the user can_view_project. https://github.com/greenriver/hmis-warehouse/pull/5999/commits/b855f955adaa93ffacf6db9eea4e69385210e072

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
martha: https://github.com/greenriver/hmis-warehouse/pull/5999/commits/d15b9ace2abdbd5b5139d02aadbb013b460a2709

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: https://github.com/greenriver/hmis-warehouse/pull/5999/commits/350ae2556625241d2b593a1343149bda01327de2

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: https://github.com/greenriver/hmis-warehouse/pull/5999/commits/3fa5e401f11c1111006e2046b2dfd10138409c50

drivers/hmis/spec/models/hmis/auth_policies/ce_referral_policy_spec.rb
martha: Oh yup makes sense, added test coverage for that https://github.com/greenriver/hmis-warehouse/pull/5999/commits/77b4f039632594be8f57bdd390e190396bb952f4

martha: I was beginning to type up the question, Do you think we need coverage for the case where the target enrollment ID points at an enrollment that isn't in the target project? (referring back to my comment [here](https://github.com/greenriver/hmis-warehouse/pull/5999/files#diff-5798e50327968067e4891358a08008c9b885660ecf90a2f3c8f12af372e224e3R47))

But, thinking about this more, that really sounds like it should be a validation on the referral model. I'm taking a stab at that, but wanted to split it out into a separate diff, in case it causes a bunch of tests to fail, so it doesn't block this work. https://github.com/greenriver/hmis-warehouse/pull/6007

gigxz (approved):
Looks great! Thanks!

PR #6000, opened 2025-11-12
Prevent orphaned CE referrals when deleting enrollments

gigxz:
## _Merging this PR_

## Description

Prevent orphaned CE referrals when deleting enrollments

### Problem
When an `Enrollment` with outgoing CE referrals was destroyed, the referrals were left with a `source_enrollment_id` pointing to a non-existent enrollment. This caused issues:
- **Direct referrals**: Lost their association to the source project, making them invisible on the source project's outgoing referrals list (they would "float" without context)
- **Waitlist referrals**: Could function without a source enrollment, but leaving orphaned references was inconsistent

### Solution
Implemented custom `before_destroy` hooks that handle referrals based on their origin and status:

1. **Active direct referrals prevent deletion**: Enrollments with active (initialized/in_progress) direct referrals cannot be deleted, preserving the source project context. This forces the user to deal with the situation by closing those referrals
2. **Completed direct referrals are destroyed**: When deletion is allowed, all direct referrals are destroyed since they require the source enrollment to identify the sending project
3. **Waitlist referrals are nullified**: Waitlist referrals have their `source_enrollment_id` set to `NULL` since they can exist without the source enrollment context

Also added validation requiring `source_enrollment_id` for direct referrals to prevent creating orphaned direct referrals.

### Alternatives Considered
Considered simpler approaches for the referral relation on Enrollment, I couldn't square any of these:

1. `dependent: :nullify` **Problem**: Works for waitlist referrals, but direct referrals become orphaned. The sending project loses visibility of the referral on their outgoing referrals list (though it remains visible on the receiving project's incoming referrals list).

2. `dependent: :restrict_with_exception` **Problem**: Prevents deletion of any enrollment that has ever been tied to a referral, even if the referral is completed. This is too restrictive and prevents necessary cleanup.

3. `dependent: :destroy` **Problem**: Might unintentionally delete legitimate referrals in waitlist-based scenarios where referrals should persist even if the source enrollment is removed.

## Type of change
Bug fix


drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -535,6 +538,26 @@ def build_synthetic_intake_assessment
     assessment
   end
 
+  # Prevents deletion of an enrollment that has active outgoing direct referrals.
+  private def check_active_outgoing_direct_referrals
+    return unless outgoing_ce_referrals.originated_from_direct_send.active.exists?
+
+    raise ActiveRecord::DeleteRestrictionError, 'Cannot delete enrollment because it has active outgoing referrals. Please complete or reject the referrals first.'
```
gigxz: This prevents the deletion, but the descriptive message doesn't make its way to the end-user (yet)

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -538,6 +541,26 @@ def build_synthetic_intake_assessment
...
 
+  # Prevents deletion of an enrollment that has active outgoing direct referrals.
+  private def check_active_outgoing_direct_referrals
+    return unless outgoing_ce_referrals.originated_from_direct_send.active.exists?
+
+    raise ActiveRecord::DeleteRestrictionError, 'active outgoing direct referrals'
+  end
+
+  # Nullifies source_enrollment_id on waitlist-based referrals when enrollment is deleted.
+  # Waitlist-based referrals can exist without as source enrollment since they're already resolved.
```
gigxz, line 552: needs ticket: support changing/adding source enrollment to referral

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -538,6 +541,26 @@ def build_synthetic_intake_assessment
...
+  private def nullify_outgoing_waitlist_referrals
+    outgoing_ce_referrals.originated_from_waitlist.each { |referral| referral.update!(source_enrollment_id: nil) }
+  end
+
+  # Destroys all direct send referrals when enrollment is deleted.
+  # Direct send referrals require the source enrollment to identify which project sent them.
+  # Without the source enrollment, they would "float" and not appear on the source project's outgoing referrals list.
```
gigxz, line 559: not as compelling, revisit this one.. its more about the target project no?

drivers/hmis/app/models/hmis/hud/enrollment.rb
```diff
@@ -112,6 +112,9 @@ class Hmis::Hud::Enrollment < Hmis::Hud::Base
 
   accepts_nested_attributes_for :move_in_addresses, allow_destroy: true
 
+  before_destroy :check_active_outgoing_direct_referrals
+  before_destroy :nullify_outgoing_waitlist_referrals
+  before_destroy :destroy_outgoing_direct_referrals
```
gigxz, line 117: To do - add handling for destroying **target** enrollment

PR #5953, opened 2025-10-31
Return field indicating aha failed reason

gigxz:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch
- use a merge-commit or rebase strategy for PRs targeting the stable branch

## Description
[//]: # (Summarize changes and include links related issue)
[//]: # (List any new dependencies or relevant ADRs)

Issue: https://github.com/open-path/Green-River/issues/8441
Frontend: https://github.com/greenriver/hmis-frontend/pull/1310

* Update the `FetchAhaScore` mutation to return an additional field "aha_failed_reason" indicating why AHA fetch failed
* Update the `FetchAhaScore` mutation to return the metadata fields (mci_quality_indicator/dw_client_id/generator) in cases when the AHA API returned -1 for a client
* Add some additional errors and tests

## Type of change
new feature

## Checklist before requesting review
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] If adding a new endpoint / exposing data in a new way, I have:
  - [x] ensured the API can't leak data from other data sources
  - [x] ensured this does not introduce N+1s
  - [x] ensured permissions and visibility checks are performed in the right places
- [x] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) 
- [x] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [ ] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #6003, opened 2025-11-13
Hotfix for AHA bug

gigxz:
## _Merging this PR_

## Description
Hotfix for bug introduced in https://github.com/greenriver/hmis-warehouse/pull/5953 and added test coverage for it.

## Type of change
bug fix

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
@@ -64,7 +64,7 @@ module HmisExternalApis::AcHmis
       # Raise the same error as if the client had no MCI Unique ID, so the mutation/frontend handle it in the same way.
       # Note: this is a hotfix for a bug introduced in release-188. IF we need to differentiate between the two cases,
       # we can add a new error class and adjust end-user messaging.
-      raise NoMciUniqueIdError if result.parsed_body&.dig('message')&.match?(/no client found/i)
+      raise NoMciUniqueIdError if client_not_found_response?(result)
 
       data = result.parsed_body&.dig('data')
       raise(Error, "AHA response missing `data` key. Response body: `#{result.parsed_body}`") unless data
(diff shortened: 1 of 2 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -115,18 +121,11 @@ def handle_error(result)
       # Check if HTTP status indicates success (200-299 range)
       return result if http_status_successful?(result.http_status)
 
-      # Handle specific case: 404 with "No client found" shouldn't raise, just return no ID
-      return result if client_not_found_response?(result)
```
gigxz, line 119: This code is not in use, the "no client found" response has a 200 response code

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -60,6 +60,12 @@ def fetch_score(client)
       result = conn.post('api/v1/clients/scores/search/', payload).
         then { |r| handle_error(r) }
 
+      # If the API returns "No client found" error, the external system was unable to find the MCI Unique ID.
+      # Raise the same error as if the client had no MCI Unique ID, so the mutation/frontend handle it in the same way.
+      # Note: this is a hotfix for a bug introduced in release-188. IF we need to differentiate between the two cases,
+      # we can add a new error class and adjust end-user messaging.
+      raise NoMciUniqueIdError if client_not_found_response?(result)
```
gigxz, line 67: In https://github.com/greenriver/hmis-warehouse/pull/5953 I inadvertently removed behavior that was causing "client not found" response to result in a `-1` score being successfully returned to the frontend. It was happening that way because the response also returns a `data` array that is empty. I updated it here to explicitly check for the message rather than just treating the empty array as -1. (The new code errors if the data array doesn't contain an AHA score).

martha (approved):
Thanks!

PR #6004, opened 2025-11-13
Add pre-merge mappings to merge audit

martha:
## _Merging this PR_

## Description
- Store `pre_merge_mappings` json blob on `ClientMergeAudit` object, storing clients' related records with their original foreign key.
  - This is largely cherry-picked from Gig's [wip](https://github.com/greenriver/hmis-warehouse/compare/release-189...gig/client-merge-undo) branch, with expanded comments and validations.
- Add more comprehensive test coverage for client merge
- Add documentation for client merge in `docs/features/`

## Type of change
new feature


drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -65,13 +66,54 @@ def perform(client_ids:, actor_id:)
...
+
+      # Validate mapping structure - each value should be a hash with attribute names
+      # mappings.each do |record_id, mapping_data|
+      #   unless mapping_data.is_a?(Hash)
+      #     raise ArgumentError, "Invalid mapping structure for #{key}[#{record_id}]: expected Hash, got #{mapping_data.class}"
+      #   end
+      #   if mapping_data.empty?
+      #     raise ArgumentError, "Invalid mapping structure for #{key}[#{record_id}]: mapping data cannot be empty"
+      #   end
+      # end
```
martha: commented out - unneeded?

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -65,13 +66,54 @@ def perform(client_ids:, actor_id:)
...
+          contact_points: {}, # [id => { PersonalID => value }]
+          custom_data_elements: {}, # [id => { owner_id => value }]
+          files: {}, # [id => { client_id => value }]
+          mci_ids: {}, # [id => { source_id => value }]
+          mci_unique_ids: {}, # [id => { source_id => value }]
+          scan_cards: {}, # [id => { client_id => value }]
+          client_locations: {}, # [id => { client_id => value }]
+          # Note: Enrollment-related records (assessments, services, disabilities, etc.)
+          # are tied to enrollments via EnrollmentID, so we don't need to store separate
+          # mappings for them. They will be updated when enrollments are transferred.
```
martha: doesn't it need to be initialized with all these keys, could be shortened? and move documentation comment up by build_and_update_merge_mappings?

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -158,6 +207,7 @@ def merge_custom_data_elements
 
         values = elements.sort_by(&:DateUpdated)
 
+        # Destroy duplicate CDEs (some may have been in the mappings, that's okay)
```
martha: explain comment more

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -190,17 +240,19 @@ def dedup_unpersisted(unpersisted_records)
 
     def update_client_id_foreign_keys
       candidates = [
-        ::GrdaWarehouse::ClientFile,
```
martha, line 193: why is `::GrdaWarehouse::ClientFile` being removed?

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -275,6 +343,9 @@ def delete_warehouse_clients
     end
 
     def update_personal_id_foreign_keys
+      # Enrollment-related records (assessments, services, disabilities, etc.) are tied to enrollments
+      # via EnrollmentID, so we only need to capture enrollment mappings. The enrollment-related records
+      # will be updated when enrollments are transferred during un-merge.
       candidates = [
```
martha, line 354: take consistent approach with `update_client_id_foreign_keys`?

```
candidates = [
  [Hmis::Hud::Enrollment, 'enrollments'],
  [Hmis::Hud::Assessment, nil]
]
```

drivers/hmis/app/models/hmis/client_merge_audit.rb
```diff
@@ -22,4 +22,57 @@ class Hmis::ClientMergeAudit < Hmis::HmisBase
   def self.apply_filters(input)
     Hmis::Filter::ClientMergeAuditFilter.new(input).filter_scope(self)
   end
+
+  # Get pre-merge mappings for a specific record type
+  # @param record_type [String] The key in pre_merge_mappings (e.g., 'enrollments', 'names')
+  # @return [Hash] Mapping of record_id => { attribute_name => original_value }
+  def mappings_for(record_type)
+    ((pre_merge_mappings || {})[record_type.to_s] || {}).transform_keys(&:to_i)
```
martha: this attr has been added with a default of `{}` in the migration, so probably don't need a similar fallback here in the model code?
```
(pre_merge_mappings[record_type.to_s] || {}).transform_keys(&:to_i)
```

drivers/hmis/app/models/hmis/client_merge_audit.rb
```diff
@@ -22,4 +22,57 @@ class Hmis::ClientMergeAudit < Hmis::HmisBase
...
+  def enrollment_ids
+    mappings_for('enrollments').keys
+  end
+
+  # Get the original PersonalID for a specific enrollment
+  # @param enrollment_id [Integer] The enrollment ID (primary key)
+  # @return [String, nil] The original PersonalID, or nil if not found
+  def original_personal_id_for_enrollment(enrollment_id)
+    mapping_data = mappings_for('enrollments')[enrollment_id]
+    mapping_data&.dig('PersonalID') || mapping_data&.dig('personal_id')
```
martha: both `PersonalID` and `personal_id` may not be needed?

drivers/hmis/app/models/hmis/client_merge_audit.rb
martha, line 1: perhaps the diff in this file can be simplified since the methods are unused, we should keep utils here that are most useful for doing manual forensics during a support request

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha, line 1: to do: store the current Warehouse Destination Client ID for each merged source client. This would be helpful because it's not easy to track down from WarehouseClient

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
martha, line 1: comprehensive test plan:
> This will include generally expanding test coverage of client merge which is currently incomplete.

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha, line 1: Highlights to hit for client merge documentation:
- we pick a client to retain, other clients are soft deleted
- related records are updated to point at the new client (as opposed to deleted and recreated)
- duplicate related records are not merged. (example: enrollments in same project - requires user cleanup)
- how client merge interacts with warehouse clients

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha: looks like this needs to be restored to get tests passing; what is difference btwn GrdaWarehouse::ClientFile and Hmis::File? need different keys in the pre-merge mappings object

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
martha: go through and make sure we have tests for each type, e.g. custom data elements edge cases

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha: doesn't need to be readme, can just be header comment? or docs/features

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha: records are same table

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha: we should have some validation of the structure; at least validate that each key stored in the pre_merge_mappings (enrollments, files, etc.) maps to the right attribute (PersonalId, client_id, etc.)

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -190,17 +240,19 @@ def dedup_unpersisted(unpersisted_records)
...
-        candidate.where(client_id: client_ids).update_all(client_id: client_to_retain.id)
+
+      candidates.each do |candidate, mapping_key|
+        candidate_scope = candidate.where(client_id: client_ids)
+        build_and_update_merge_mappings(key: mapping_key, scope: candidate_scope, attributes: 'client_id')
+        candidate_scope.update_all(client_id: client_to_retain.id)
       end
 
-      # Update ReferralHouseholdMembers in a way that respects uniquness constraint on (client_id, referral_id)
+      # Update ReferralHouseholdMembers in a way that respects uniqueness constraint on (client_id, referral_id)
```
martha: comment when something isn't added to mapping (in this case bc deprecated)

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
```diff
@@ -9,279 +9,594 @@
...
+        Hmis::MergeClientsJob.perform_now(client_ids: client_ids, actor_id: actor.id)
+        audit = Hmis::ClientMergeAudit.first
+        mappings = audit.mappings_for(key)
+        expect(mappings[record2.id]).to eq({ mapping_field => client2.send(client_field || mapping_field) })
+        # TODO- discuss, should we add this expectation?
+        # currently the behavior is not consistent. some related records are only stored in the mappings if they are updated,
+        # others are always stored even if they pointed at the retained client originally and aren't touched.
+        # - client names
+        # - custom data elements
+        # expect(mappings.keys).not_to include(record1.id)
```
martha: Do we need to expect that only records that _have_ been touched/changed are mentioned in `pre_merge_mappings`?

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
```diff
@@ -9,279 +9,594 @@
...
+        it 'retains one mci_unique_id' do
+          expect do
+            Hmis::MergeClientsJob.perform_now(client_ids: client_ids, actor_id: actor.id)
+          end.to change(HmisExternalApis::ExternalId.mci_unique_ids, :count).by(-1)
+          expect(client1.ac_hmis_mci_unique_id&.value).to eq(record1.value)
+        end
+
+        # TODO- discuss, do we need to add this expectation? I think no, because in this case the discarded unique ID is not updated.
+        # but I think it's fully deleted (not soft deleted) so this might be something to flag for unmerge
+        # it_behaves_like 'merge that saves mappings', 'mci_unique_ids', 'source_id', 'id'
```
martha: Do we need to make updates to mci unique ids so that they are fully recoverable when unmerging?

docs/features/hmis_client_merges.md
```diff
@@ -0,0 +1,96 @@
+# HMIS Client Merges
```
gigxz, line 1: The format we're following now is to place these in `docs/features` at the root of the repo. In https://github.com/greenriver/hmis-warehouse/pull/6014/files I proposed adding an "hmis" folder within the features directory. Maybe we can do the same for this? @ttoomey thoughts?

drivers/hmis/doc/readme_for_client_merges.md
```diff
@@ -0,0 +1,96 @@
...
+If both merged clients have enrollments in the same project that overlap,
+post-merge this would be bad data that needs to be cleaned up manually by the user.
+
+### Warehouse Clients and Client Merges
+
+The interaction between HMIS client merges and warehouse clients works as follows:
+
+1. When clients are merged in HMIS, the `WarehouseClient` linking records for the **source clients that are being deleted** are destroyed immediately as part of the merge.
+2. The destruction of `WarehouseClient` records may leave some warehouse destination clients without any remaining source clients
+3. Later, when the `IdentifyDuplicates` job runs, it identifies and cleans up "orphaned" warehouse clients that no longer have any source clients
```
gigxz:
```suggestion
3. Later, when the `ClientCleanup` job runs, it identifies and cleans up "orphaned" warehouse clients that no longer have any source clients
```

I think this is the responsibility of the client cleanup job https://github.com/greenriver/hmis-warehouse/blob/df6ae485f8347e22930122ffaf75b67c7dea9223/app/models/grda_warehouse/tasks/client_cleanup.rb#L24

IdentifyDuplicates may queue client cleanup, but it can also be run independently (and I think runs on a regular schedule - not totally sure)

drivers/hmis/doc/readme_for_client_merges.md
```diff
@@ -0,0 +1,96 @@
...
+
+However, note that **warehouse clients are not necessarily deleted** because:
```
gigxz:
```suggestion
However, note that **destination clients are not necessarily deleted** because:
```
To differentiate between `WarehouseClient`

drivers/hmis/spec/factories/hmis/files.rb
```diff
@@ -24,5 +24,10 @@
       file.client_file.attach(evaluator.blob) if evaluator.blob
       file.tag_list = evaluator.tags.map(&:id)
     end
+
+    trait :without_validations do
```
gigxz: Suggest naming skip_validate for consistency with the other factories that inherit from hmis_base_factory:
https://github.com/greenriver/hmis-warehouse/blob/df6ae485f8347e22930122ffaf75b67c7dea9223/drivers/hmis/spec/factories/hmis/hud/hmis_base_factory.rb#L17-L19

docs/features/hmis_client_merges.md
```diff
@@ -0,0 +1,96 @@
...
+- **Hmis::ClientMergeAudit:** `drivers/hmis/app/models/hmis/client_merge_audit.rb`
+  - Stores audit information for each merge operation, including:
+    - `pre_merge_state`: JSON storing attributes of clients before merge
+    - `pre_merge_mappings`: JSON storing original foreign key values for updated records
+- **Hmis::ClientMergeHistory:** `drivers/hmis/app/models/hmis/client_merge_history.rb`
+  - Links merge audits to the affected clients
+  - Each merge creates one `ClientMergeAudit`, and one or more `ClientMergeHistory` (one per deleted client).
+  - **Chain Handling**: When a previously-retained client is later merged into another client, all related `ClientMergeHistory` records are updated to point to the new retained client, preserving the complete merge chain.
+
+### Related Records involved
```
gigxz, line 36: This will be super helpful!

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
gigxz: > currently the behavior is not consistent. some related records are only stored in the mappings if they are updated, others are always stored even if they pointed at the retained client originally and aren't touched.

It would probably be clearer for our future selves if the behavior was consistent across related record types. Is there a good reason for the inconsistency right now?

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
```diff
@@ -9,279 +9,594 @@
...
+        and change(Hmis::ClientMergeHistory, :count).from(0).to(1)
+
       audit = Hmis::ClientMergeAudit.first
       expect(audit.client_merge_histories.count).to eq(1)
-      expect(Hmis::ClientMergeHistory.count).to eq(1)
-      expect(Hmis::ClientMergeHistory.first).to have_attributes(
+      # Stores pre-merge mappings
+      expect(audit.pre_merge_mappings).to be_present
+      expect(audit.pre_merge_mappings).to be_a(Hash)
+
```
gigxz, line 53: This might be a good place to verify `pre_merge_state` on the audit trail?

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
gigxz: Interesting, yeah it seems like we are missing something there. Good find! Maybe its as simple as adding soft-deletion support to the external_ids table. Would you mind capturing a ticket for the issue that MCI Unique ID history is not fully captured on merge? It seems like it could become a bigger question (which one do we keep?) and given that we are meeting tomorrow to rehash the MCI Unique ID integration in general, this should probably wait.

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
```diff
@@ -9,279 +9,594 @@
 require 'rails_helper'
```
gigxz, line 10: test coverage and structure rework looks great, thank you!

drivers/hmis/app/models/hmis/client_merge_audit.rb
```diff
@@ -22,4 +24,98 @@ class Hmis::ClientMergeAudit < Hmis::HmisBase
...
+    pre_merge_mappings[key] || []
+  end
+
+  # Check if a specific record was destroyed during merge
+  # @param record_type [String] The record type
+  # @param record_id [Integer] The record ID
+  # @return [Boolean] True if the record was destroyed
+  def record_destroyed?(record_type, record_id)
+    destroyed_record_ids(record_type).include?(record_id.to_s)
+  end
```
gigxz: I think these can be removed, we're not storing destroyed record ids like this right?

gigxz (commented):
Looks great!

docs/features/hmis_client_merges.md
gigxz: Nevermind, I updated https://github.com/greenriver/hmis-warehouse/pull/6014/files to just be consistent with the existing convention. Maybe we can just use the `hmis_` prefix.

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
martha: Sure! Created https://github.com/open-path/Green-River/issues/8590. I'll remove this comment, since if we solve this using paranoia, test coverage wouldn't use the 'merge that saves mappings' shared example.

drivers/hmis/spec/jobs/merge_clients_job_spec.rb
martha: Thanks, good point, nope I think it was just an implementation detail. https://github.com/greenriver/hmis-warehouse/pull/6004/commits/1201a1c6c4fd0c3fee2df8ae0428bce470d556fa

martha:
Thanks for the feedback! 🎉 I think this is ready to merge pending the test suite.

PR #6015, opened 2025-11-18
8568 spm perf

ttoomey:
## _Merging this PR_

### Description
Refactors the FY2026 System Performance Measures (SPM) report generator to address some performance issues, adds specs

-  **`SpmEnrollment`**: Reworks the enrollment loading process to use batches
- **`EpisodeBatch`**: Optimizes the service/bed night query
- **`HDXUpload`**: Replaces N+1 cell updates with a single `insert_all` bulk operation, reducing database queries required for the HDX export.
- **`HudFilterBase` & `ReportInstance`**: Adds memoization for frequently accessed data (user)
- **`docs/features/hud_spm_report.md`**: new documentation outlining the SPM architecture, data models, and calculation flow.
- **Specs**:  Add performance test
  - note, includes AI generated specs to check for regressions in all measures.

### Type of Change
Performance improvement


spec/rails_helper.rb
```diff
@@ -80,9 +80,19 @@
...
+    # Check if any examples are from HUD report drivers
+    hud_report_examples = example_file_paths.grep(/\/drivers\/(hud_path_report|hud_spm_report|hud_data_quality_report)\//).any?
+
+    # Check if all examples explicitly exclude fixpoints via metadata
+    # To exclude fixpoints, add `exclude_fixpoints: true` to your RSpec.describe block:
+    #   RSpec.describe MyClass, type: :model, exclude_fixpoints: true do
+    all_exclude_fixpoints = examples.all? { |e| e.metadata[:exclude_fixpoints] }
+
+    # Load fixpoints if we're running tests that may need them
+    if hud_report_examples && !all_exclude_fixpoints
```
ttoomey, line 95: this is for DX, allows iterating on a single SPM spec without paying the cost of loading fix points

drivers/hud_spm_report/app/models/hud_spm_report/adapters/service_history_enrollment_filter.rb
```diff
@@ -35,38 +35,9 @@ def enrollments
       scope = filter_for_cocs(scope)
       scope = @filter.apply_criteria(scope, tags: [:warehouse, :client])
 
-      GrdaWarehouse::Hud::Enrollment.where(id: scope.joins(:enrollment).select(e_t[:id])).select(*enrollment_columns)
-    end
-
-    # Limited columns to avoid pulling more data from the database than necessary
-    def enrollment_columns
```
ttoomey, line 42: pulled this into the SPM Enrollment; we want to tune this select list for when we are pre-calculating households vs all members (also generally better locality of behavior)

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/episode.rb
```diff
@@ -408,7 +408,7 @@ def compute_episode(enrollments, included_project_types:, excluded_project_types
     end
 
     private def filter
-      @filter ||= ::Filters::HudFilterBase.new(user_id: report.user.id).update(report.options)
+      @filter ||= ::Filters::HudFilterBase.new(user: report.user).update(report.options)
```
ttoomey, line 411: optional - passing in the user avoids having to re-query for the user and preserves any cached checks in the user itself. I don't think this matters much for performance because this is handled by rails query cache but does reduce the chatter when trying to read the query logs

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
```diff
@@ -34,7 +35,8 @@ class SpmEnrollment < HudReports::ReportClientBase
       # SPM only runs on residential projects,
       # residential projects do not receive service on their exit date,
       # exclude exit date
-      where(dates_overlaps_arel(range, a_t[:entry_date], a_t[:exit_date]), exit_date_included: false)
```
ttoomey, line 37: it looks like the paren was in the wrong place here, not sure why this didn't generate an exception.

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
```diff
@@ -261,7 +297,7 @@ def self.detail_headers
       (total_income(income_benefit) - earned_income(income_benefit)).clamp(0..)
     end
 
-    private_class_method def self.current_income_benefits(enrollment, end_date)
+    private_class_method def self.collect_current_income_benefits(enrollment, end_date)
```
ttoomey, line 293: renamed to avoid shadowing the active record association

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
```diff
@@ -280,7 +316,7 @@ def self.detail_headers
       end
     end
 
-    private_class_method def self.previous_income_benefits(enrollment, annual_date, end_date)
+    private_class_method def self.collect_previous_income_benefits(enrollment, annual_date, end_date)
```
ttoomey, line 312: renamed to avoid shadowing the active record association

app/models/filters/hud_filter_base.rb
```diff
@@ -16,6 +16,16 @@ class HudFilterBase < FilterBase
     # Provide a mechanism to limit relevant project types
     attribute :relevant_project_types, Array, default: []
 
+    def initialize(attributes = {})
+      # Extract user from attributes hash if present
+      user = attributes[:user] if attributes.is_a?(Hash) || attributes.respond_to?(:[])
```
eanders: I'd guess the following is sufficient:
```suggestion
      user = attributes[:user] if attributes.respond_to?(:[])
```

app/models/filters/hud_filter_base.rb
```diff
@@ -16,6 +16,16 @@ class HudFilterBase < FilterBase
...
+      attributes = attributes.except(:user) if user && attributes.respond_to?(:except)
+      super(attributes)
+      # Cache @user instance to avoid extra database lookups while still setting user_id for consistency
+      @user = user if user
+      self.user_id = user.id if user
```
eanders: Is there a reason we use `@user` but `self.user_id`?

docs/features/hud_spm_report.md
```diff
@@ -0,0 +1,35 @@
+## HUD System Performance Measures (SPM)
+
+### Introduction
+- For HUD’s official specification, see https://www.hudexchange.info/resource/4483/system-performance-measures-tools/.
+- Legacy generators for FY2020, FY2023, and FY2024 remain in the driver for backward compatibility, but FY2026 is the active implementation.
+- `HudSpmReport.current_generator` switches between FY2026 and FY2024 based on the configured default report version.
```
eanders: This should no longer switch back to FY2024.  Maybe:
```suggestion
- `HudSpmReport.current_generator` is generally fixed to point to the most-recent version, however during transition periods (usually September) it may point to the next fiscal year in some environments.
```

docs/features/hud_spm_report.md
```diff
@@ -0,0 +1,35 @@
...
+
+### Architecture
+- The SPM feature ships as a Rails driver under `drivers/hud_spm_report`. Controllers inherit from the shared HUD reports controller stack and mount under the `/hud_reports/spms` namespace.
+- The FY2026 generator exposes metadata (title, question list, filter class, upload capabilities) to the HUD reports framework. Each question maps to a dedicated measure class that encapsulates table preparation and summary calculation logic.
```
eanders:
```suggestion
- The active generator exposes metadata (title, question list, filter class, upload capabilities) to the HUD reports framework. Each question maps to a dedicated measure class that encapsulates table preparation and summary calculation logic.
```

docs/features/hud_spm_report.md
```diff
@@ -0,0 +1,35 @@
...
+- `Generator.questions` returns the ordered measure list (Measures 1–7 plus HDX upload). HUD report answers reference these classes via the question number.
```
eanders, line 11:
```suggestion
- `Generator.questions` returns the ordered measure list (Measures 1–7 plus HDX upload). HUD report answers reference these classes via the question number.
- Generally speaking, the code for SPMs should be written in such a way that it is self contained with year-specific logic in the driver.  There are a variety of shared concerns for standard functionality that has historically stayed consistent year-over-year which can be used (households, ages, incomes, etc. -- see `models/concerns/hud_reports`) however if the logic for the SPM differs from the historic standard, code should be copied into the driver and altered in-situ rather than applying overrides to the shared components.
```

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
```diff
@@ -132,29 +134,63 @@ def project_id
...
+        :DisablingCondition,
+        :DateOfEngagement,
+        :MoveInDate,
+        :DateCreated,
+        :DateUpdated,
+        :UserID,
+        :DateDeleted,
+        :ExportID,
+        :data_source_id,
+      )
```
eanders: We should verify this list of columns, I'm unclear how old that might be.  Perhaps worth putting it in a constant to elevate it a bit for re-review with each iteration.

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
```diff
@@ -330,20 +366,42 @@ def self.detail_headers
...
+
+      query.enrollment_batches(enrollment_scope) do |batch|
         batch.each do |enrollment|
-          result[enrollment.household_id] = HomelessnessInfo.new(
-            start_of_homelessness: enrollment.date_to_street_essh,
-            entry_date: enrollment.entry_date,
-            move_in_date: enrollment_own_move_in_date(enrollment),
-          )
+          key = [enrollment.data_source_id, enrollment.household_id]
+          # With ordering, first valid enrollment per household wins
```
eanders: This comment makes me nervous.  I don't see anything that prefers the HoH or adult data over a child's data, so if there's a move-in-date on a child record and for whatever reason it comes back first, it'll get picked.  Also, it seems possible that someone with a later `entry_date` might be chosen.  If we're going to rely on the sort, I think we should add `RelationshipToHoH` (prefer 1 over anything else) and `EntryDate` (prefer earliest) to the sort order.

drivers/hud_spm_report/spec/models/fy2024/active_record_preload_spec.rb
eanders, line 1: Can we remove this spec if we're no longer running FY24 SPMs?

drivers/hud_spm_report/spec/models/fy2026/episode_batch_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+      )
+      s.update_column(:data_source_id, other_data_source.id)
+
+      # Setup a report instance (this populates SpmEnrollment)
+      report = setup_report([project.id])
+
+      # Create a SpmEnrollment scope for just the batch enrollment
+      spm_enrollments = HudSpmReport::Fy2026::SpmEnrollment.where(
+        enrollment_id: enrollment_in_batch.id,
+      )
```
eanders: I'm probably missing something, but the comment says create, but this is just a query for existing SpmEnrollments

drivers/hud_spm_report/spec/models/fy2026/episode_batch_spec.rb
```diff
@@ -0,0 +1,149 @@
...
+      # Enrollment 1: In the batch
+      enrollment_1 = create_enrollment(
+        client: client,
+        project: project,
+        entry_date: '2023-01-01'.to_date,
+        exit_date: '2023-01-10'.to_date,
+      )
+      create_bed_night_service(enrollment: enrollment_1, date: '2023-01-05'.to_date)
+
+      # Enrollment 2: Same client, same project, but NOT in the batch (simulating filtered out enrollment)
```
eanders: Is this ghost client just to prove that the filter is project/data source aware, or is there some other reason we think it is relevant to have a client with identical enrollments in two data sources?  In the real world if that happened and both projects were included in the universe, it should only generate unique days for the client, so it shouldn't really matter that they have two enrollments.

drivers/hud_spm_report/spec/models/fy2026/measure_6_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative './shared_context'
+
+RSpec.describe HudSpmReport::Generators::Fy2026::MeasureSix, type: :model, exclude_fixpoints: true do
```
eanders, line 12: Do we actually calculate anything for Measure 6 yet?  If memory serves, historically this has been left blank, but I thought maybe it had spec now.  If not, this is great, if it does, I think we should see if we can get something other than zeros below.

drivers/hud_spm_report/spec/models/fy2026/spm_performance_spec.rb
eanders, line 1: I like this approach.  What do you think about running it as is but also with significantly more enrollments.  I'd like to know that just because we have more data doesn't mean we're increasing our number of queries.   Maybe that's a pipe dream and they actually will increase linearly with every thousand (or whatever that batch size is.)  Anyway, it feels like we should confirm if the number of clients/enrollments is larger than one or two batches, what happens to the number of queries.

spec/rails_helper.rb
eanders, line 95: I wonder if we should flip this and add a `load_fixpoints` option to those that actually use it.... I guess I had forgotten that we do this.  I'm not sure how much time loading the fixpoints adds, but it might be significant.

ttoomey: It's quite a bit. I think flipping it would be cleaner. But it looked like there were a lot more specs that would need this flag, felt it wasn't worth the churn right now

app/models/filters/hud_filter_base.rb
ttoomey: The underlying model uses `virtus` attrs for user_id but @user is memoized in an ivar. Virtus probably uses ivars too but I don't really know its mechanics; there's normalization happening and I didn't want to mess with it

spec/rails_helper.rb
eanders: Definitely fine as is.  Perhaps something we can flip in the future.

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
ttoomey: Thanks!

> I don't see anything that prefers the HoH or adult data over a child's data, so if there's a move-in-date on a child record and for whatever reason it comes back first, it'll get picked.

I think this is taken care of by the `heads_of_households` filter on  `enrollment_scope` on line 386. I'll call it out in a comment though.

> Also, it seems possible that someone with a later entry_date might be chosen.

Good catch; I guess we can use entry_date and move-in-date can be a tie breaker.

It looks like the old code didn't sort by EntryDate either (pretty sure the ordering is ignored when using `find_in_batches`). It also wasn't including data_source_id in the hash key, which could result in collisions

The old code:
```ruby
# renamed to `collect_household_info`
private_class_method def self.household(enrollments)
      result = {}

      scope = enrollments.heads_of_households.order(e_t[:household_id], e_t[:move_in_date].asc.nulls_last)
      scope.find_in_batches do |batch|
        batch.each do |enrollment|
          result[enrollment.household_id] = HomelessnessInfo.new(
            start_of_homelessness: enrollment.date_to_street_essh,
            entry_date: enrollment.entry_date,
            move_in_date: enrollment_own_move_in_date(enrollment),
          )
        end
      end
      result
    end
```

drivers/hud_spm_report/spec/models/fy2024/active_record_preload_spec.rb
ttoomey: We can - I was thinking we should do this in another issue:
https://github.com/open-path/Green-River/issues/8572

Note the issue body is an AI suggested implementation plan that may be nonsense. But I would like to develop an approach we can apply across all reports for retired years

drivers/hud_spm_report/spec/models/fy2026/episode_batch_spec.rb
ttoomey: Some of these specs are AI generated and could be tightened; the comment is not inaccurate but it shouldn't use the word 'create' as it's confusing

drivers/hud_spm_report/spec/models/fy2026/episode_batch_spec.rb
ttoomey: Good catch. Yeah it doesn't seem very useful- it's intended to check that the filter is correctly applied but enrollment 1 is passed in explicitly; probably not very useful as is

I'll tighten that up; we probably could use a check that service dates are filtered and count towards days homeless correctly

drivers/hud_spm_report/spec/models/fy2026/measure_6_spec.rb
ttoomey: no, if I understand the spec correctly it's still a placeholder and not required

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
eanders: Yeah, that looks like it might have issues.

ttoomey:
The HDX is generating a bunch of DQ reports. That also seems like it could be a performance issue

* adjusted the 2026 SPM to use the 2026 DQ report (was 2024)
* added unit tests for the relevant questions in the APR (DQ) that are used by the SPM to check for regressions due to perf optimization
* added some preloads to APR client generation (ran out of time to optimize APR further)

drivers/hud_spm_report/spec/models/fy2024/hdx_upload_spec.rb
```diff
@@ -194,15 +194,15 @@
     end
 
     # Test that all headers are properly set in the first row
-    it 'sets all column headers correctly' do
+    xit 'sets all column headers correctly' do
```
ttoomey, line 197: these should be removed in [8572](https://github.com/open-path/Green-River/issues/8572)

app/models/filters/hud_filter_base.rb
```diff
@@ -16,6 +16,16 @@ class HudFilterBase < FilterBase
...
 
+    def initialize(attributes = {})
+      # Extract user from attributes hash if present
+      user = attributes[:user] if attributes.respond_to?(:[])
+      attributes = attributes.except(:user) if user && attributes.respond_to?(:except)
+      super(attributes)
+      # Cache @user instance to avoid extra database lookups while still setting user_id for consistency
+      @user = user if user
+      self.user_id = user.id if user
+    end
```
eanders: Sorry I didn't catch this before, but should we just push this up to `FilterBase`?  I don't see any thing that would be troubling.  I can see this being useful for caching the user on any report.

drivers/hud_spm_report/app/models/hud_spm_report/fy2026/spm_enrollment.rb
eanders: Thank you!

eanders (approved):
Thanks @ttoomey, looks good to me!

app/models/filters/hud_filter_base.rb
ttoomey: good idea, I moved it

app/models/hud_reports/report_cell.rb
```diff
@@ -84,6 +84,24 @@ def add_universe_members(members)
       )
     end
 
+    # Add members to the universe of this cell
+    #
+    # @param members [Hash<Client | Integer, ReportClientBase] the members to be associated with this cell
+    def add_universe_members_from_client_ids(members)
+      # if the keys of members are integers, assume they are clients ids
```
eanders: I'm not seeing any check for integers here.  Did you intend to check the type of the keys?

drivers/hud_spm_report/spec/models/fy2026/spm_performance_spec.rb
ttoomey: I've expanded it to the DQ report as well and identified a few more n+1 queries. The difficulty with running it against large numbers of enrollments is that building the service history objects is quite slow; I've left the "large data set" tests disabled for the moment so they don't bog down CI. But they are a tool for additional performance work

I've adjusted the tests so that it should be easy to scale them up to test different scenarios. I ran the measures with around 1,000 enrollments and the number of queries didn't seem to grow too dramatically. However ensuring the data shape in our tests exercises all the test logic is difficult.

ttoomey:
Additional work to remove n+1 queries and add performance specs for DQ report

- **Service History**: Updates `RebuildEnrollmentsByBatchJob` to preload association,  adds progress reporting as development aid.
- **HUD Reports Shared**: Updates `Incomes` calculation to use in-memory filtering when associations are already loaded.
- **Report Engine**: Adds `add_universe_members_from_client_ids` to remove n+1 when loading enrollment.client association, Updates SPM Measures 1-7 to pass client IDs 
- **APR Report**: Adds project preloading to client enrollment scopes to reduce N+1 queries.
- **Performance Testing**: Implements dedicated performance specs for APR DQ reports

spec/shared_contexts/hud_report_performance_helpers.rb
```diff
@@ -0,0 +1,410 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+require_relative './hud_enrollment_builders'
+
+# Shared context for HUD report performance testing utilities
+# Provides bulk building methods optimized for creating large test datasets
+RSpec.shared_context 'HUD report performance helpers', shared_context: :metadata do
```
ttoomey, line 8: possible we could replace some of this with `spec/support/hmis_csv_fixture_factory.rb` once it is merged in

app/models/hud_reports/report_cell.rb
ttoomey: oh sorry I pushed but wasn't finished making changes. Should be corrected now

drivers/hud_spm_report/spec/models/fy2026/spm_performance_spec.rb
eanders: How big is the pagination?  I feel like we might need to test with > 10,000 enrollments to be sure, but good to know it's not dramatically different at 10x 100.

eanders (approved):
Looks good to me.  Thanks @ttoomey

eanders:
Oh, we should probably re-target 191.

drivers/hud_spm_report/spec/models/fy2026/spm_performance_spec.rb
ttoomey: @eanders 
I'd like to revisit this further but suggest we put a pin in this at this point; it has proven to be a time sink. Unless you object I suggest we get this out the door as-is and then plan to do another round of performance benchmarking and improvement. Sound okay?

eanders: yep, 🚢

ttoomey:
Additional changes
* consistent cache handling
* avoid crashing on soft-deleted clients
* avoid edge case issue in optimizing annual assessment queries in `HudReports::Incomes#annual_assessment`

eanders (approved):

PR #6016, opened 2025-11-18
8573-paginate-report-drilldowns

ttoomey:
## _Merging this PR_

### Description

Improves SPM drill-down performance by implementing server-side pagination, search, and asynchronous Excel exports.

Adds pagination, new `search` action for filtered results, and `apply_search_filter` method that handles both direct model searches and association-based searches (for Episode/Return models searching through SpmEnrollment). Preloads associations to avoid N+1 queries.

### Type of Change
Bug fix | Performance improvement


ttoomey:
@eanders curious what you think about this; source of the issue was
https://github.com/open-path/Green-River/issues/8509

This PR still needs some tightening but probably enough to react to

app/models/hud_reports/report_client_base.rb
```diff
@@ -22,6 +22,42 @@ def display_value(col, pii_policy:, include_content_tag: true, cell_val: nil, ca
...
+      columns = [
+        Arel::Nodes::NamedFunction.new('CAST', [table[:id].as('TEXT')]),
+        Arel::Nodes::NamedFunction.new('CAST', [table[:client_id].as('TEXT')]),
+      ]
+
+      ['first_name', 'last_name', 'personal_id'].each do |col|
+        columns << table[col] if column_names.include?(col)
+      end
+      columns
+    end
```
eanders, line 53: This means we can't search by project or other columns in the table right?  We're just looking at client name and identifiers?  
That might be ok, but it's been helpful in the past to be able to search by dates, project names, and other values.

eanders: Looks like this is overwritten for the specific cell details.  I think we should try and generate the list of columns from the same list that is used for the download/display.

drivers/hud_spm_report/app/controllers/hud_spm_report/cells_controller.rb
```diff
@@ -13,30 +13,101 @@ class CellsController < HudSpmReport::BaseController
...
+          report_id: @report.id,
+          measure_id: @question,
+          cell_id: @cell,
+          table: @table,
+        }.to_query,
+      )
+
+      HudSpmReport::CellDetailExportJob.perform_later(export_id: export.id)
+
+      flash[:notice] = 'Your cell-detail export is being generated. You will receive an email with a download link shortly.'
```
eanders, line 101: I think we should set this up in the same way we do other background downloads.  Those pop open a modal telling you that it's running and provide the download link when it's ready.  They also send an email.  Maybe that's what's happening here, but it wasn't clear to me.
I think starting by looking at the download link in `app/views/hud_reports/_show.haml` is probably the easiest place to start.  Also, maybe this already works this way?

drivers/hud_spm_report/app/views/hud_spm_report/cells/show.haml
```diff
@@ -7,55 +7,32 @@
...
-  = content_for :page_js do
-    :javascript
-      table = $('.datatable').DataTable({
-        paging: false,
-        "dom": 'lrtip'
-      });
-      $('#table_search').keyup(function(){
-        table.search($(this).val()).draw();
-      });
+.d-flex.align-items-end.mb-4
```
eanders, line 25: Not sure how important this is, but we lost the ability to sort the table.

Also, I think we're missing some small CSS so that the search button attaches to the search field.  Totally minor.

eanders (commented):
@ttoomey , I think generally we probably need to move to a server side search and pagination and background download.  We should discuss if we think losing sort is important enough to figure that out, and to see if we can provide search across other columns (or if we think that makes sense.)
Thanks for spiking this out.

app/models/hud_reports/report_client_base.rb
ttoomey: sure, we can enhance the searc cols. I guess my assumption is people can download to excel if they need more. We also lose the ability to sort by columns. But I think eliminating these unpaginated record sets is well worth it

drivers/hud_spm_report/app/controllers/hud_spm_report/cells_controller.rb
ttoomey: added the modal treatment

drivers/hud_spm_report/app/models/hud_spm_report/document_exports/cell_detail_export.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
+
+module HudSpmReport
+  module DocumentExports
+    class CellDetailExport < ::GrdaWarehouse::DocumentExport
+      def authorized?
+        report.user_id == user_id && user.can_view_any_reports?
```
ttoomey, line 7: double check me pls

ttoomey:
@eanders flagging for re-review. I think I've addressed your feedback. See QA instructions in issue

drivers/hud_spm_report/app/models/hud_spm_report/document_exports/cell_detail_export.rb
eanders: I'm not sure it makes sense to check against the report.user_id (also where is `_user_id` being set? Did you mean `user.id`?

If we compare to the person who ran the report, you'd never be able to see the details of a report someone else ran.  (I think that happens occasionally.)

ttoomey: `user_id` is a column on ::GrdaWarehouse::DocumentExport

We expect document exports to be shared - I guess that makes sense. I'll remove that condition

eanders: Ah, no sorry, I miss read the context.  The exports are definitely NOT shared.

drivers/hud_spm_report/app/models/hud_spm_report/cell_detail_export_builder.rb
```diff
@@ -0,0 +1,102 @@
...
+    end
+
+    def possible_generator_classes
+      {
+        fy2020: HudSpmReport::Generators::Fy2020::Generator,
+        fy2023: HudSpmReport::Generators::Fy2023::Generator,
+        fy2024: HudSpmReport::Generators::Fy2024::Generator,
+        fy2026: HudSpmReport::Generators::Fy2026::Generator,
+      }
+    end
```
eanders, line 51: If possible, I think we should delegate this to the controller `drivers/hud_spm_report/app/controllers/hud_spm_report/base_controller.rb` since we need to update it yearly.  Having it here as well increases the likelihood we'll miss it.

eanders:
There seems to be some duplication of pagination headers.

Screenshot 2025-12-12 at 3 45 06 PM
Screenshot 2025-12-12 at 3 45 19 PM

eanders (approved):
@ttoomey , just the two small fixes noted in prior comments.  This looks great.  Feel free to merge once those are fixed.

PR #6018, opened 2025-11-18
DQ Tool By Client Hotfix

dtgreiner:
## _Merging this PR_

## Description

Fixes issue in DQ Tool By Client page where the personalID value and link were not working correctly

## Type of change

Bug fix

### DIFF
--- a/drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/_by_client.haml
+++ b/drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/_by_client.haml
@@ -28,8 +28,7 @@
           -else
             %td
           %td
-            = link_to_if @report.can_see_client_details?(current_user), appropriate_client_path(client.destination_client_id) do
-              = client.personal_id
+            = link_to_if @report.can_see_client_details?(current_user), client.personal_id, appropriate_client_path(client.destination_client_id)
             - ds = client.client&.data_source
             - if ds&.hmis_link_available?
               = hmis_external_link(client.client, user: current_user, wrapper_class: 'mb-2 text-teeny')


drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/_by_client.haml
```diff
@@ -28,8 +28,7 @@
           -else
             %td
           %td
-            = link_to_if @report.can_see_client_details?(current_user), appropriate_client_path(client.destination_client_id) do
-              = client.personal_id
+            = link_to_if @report.can_see_client_details?(current_user), client.personal_id, appropriate_client_path(client.destination_client_id)
```
ttoomey, line 31: I don't have the context- just checking that the link text change is intentional here? It's not this right?

```suggestion
            = link_to_if @report.can_see_client_details?(current_user), client.destination_client_id, appropriate_client_path(client.destination_client_id)
```

ttoomey (approved):

drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/_by_client.haml
dtgreiner: This was originally using `client.personal_id` so I kept it as is, but I can do a little more digging to see if anything pops out.

eanders: My understanding is that the link should use `destination_client_id`, but the point of this column is to show the HMIS source `PersonalID`.  Do you know if we also expose the warehouse client ID on this page?  The desire to see the `PersonalID` stems from the need to track the issue back to the sending HMIS, and that's often the only ID that works in both places.

eanders (approved):

drivers/hmis_data_quality_tool/app/views/hmis_data_quality_tool/warehouse_reports/reports/_by_client.haml
dtgreiner: This is for `HmisDataQualityTool::Client` which as the personal_id value being built from the `source_client.PersonalID`. The column is also specified as the "Personal ID", so I think leaving it as showing the source client Personal ID and linking to the destination in the warehouse is reasonable. This will help identify where the DQ issue is coming from (which specific source client) while keeping the link working in the warehouse.

PR #6023, opened 2025-11-19
8570 idempotent HUD report runs

ttoomey:
## _Merging this PR_

### Description
* adjusts the SPM report generator to be idempotent and supports retries
* avoid incorrect data on retry for non-idempotent reports

#### Details
-   `GeneratorBase` adds a `supports_idempotent_retry?` class method (default `false`) to allow reports to opt-in to retryable behavior.
-  Report runner blocks attempted retries of non-idempotent reports runs 
-   Adjusts the SPM report generator to be idempotent and supports retries, skip questions that have already been completed successfully, deletes partially generated question and snapshot data
-   Adds a new `docs/features/hud-report-framework.md` to describe the reporting framework.

### Type of Change
New feature


app/jobs/reporting/hud/run_report_job.rb
```diff
@@ -62,8 +62,19 @@ def perform(class_name, report_id, email: true)
 
     protected def run_report(report, generator, email:)
       capture_failure(report) do
+        # Fail-fast if attempting to retry a report that doesn't support idempotent retry
+        # Check started_at to catch all retry scenarios (completed questions, partial runs, mid-question failures)
+        if report.started_at.present? && !generator.class.supports_idempotent_retry?
+          raise "Cannot retry #{generator.class.name}: this report does not support idempotent retry. " \
```
eanders: If we raise here, I think we'll lose the original error from Delayed Job.  Maybe we should just log and return?

eanders (approved):
@ttoomey this is great, thank you.  One small tweak in that I think we'd rather not raise if an existing report retries as I believe it'll replace any previous error which might be useful for debugging why it retried.  Better to simply fail silently and log that it failed.

app/jobs/reporting/hud/run_report_job.rb
ttoomey: Good catch! So I think if we return instead of raise, the job will be marked as "completed" but we should instead be marked as failed.

I've updated the guard so that when the runner retries a non-idempotent job, it still fails but preserves the original error message (if present).

Seem okay?

eanders (approved):

app/jobs/reporting/hud/run_report_job.rb
eanders: I think that's good.  I'm not super excited about the complexity, but I agree about the return marking it as completed being a bigger issue.

ttoomey:
I'll retarget this to 190 but push out as a hotfix on 189 where needed

PR #6030, opened 2025-11-21
8570 pristine snapshots 189

ttoomey:
Don't merge, this will go into 190 in https://github.com/greenriver/hmis-warehouse/pull/6023

Keep this against 189 for deployment prod as needed


ttoomey:
closing, we just deployed it in 190

PR #6035, opened 2025-11-24
Add permission to view outgoing referrals

martha:
## _Merging this PR_

## Description
- Add new permission to view outgoing referrals from a project
- Support the new permission, distinctly from the existing "can manage outgoing referrals" permission. We decided to keep "can manage"s existing functionality of granting summary-level access on these referrals, whereas "can view" can be used for detailed-level access. (See discussion in comments on the linked ticket)

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
```diff
@@ -63,6 +63,7 @@ def self.authorized?(object, ctx)
       field :can_view_source_enrollment_details, Boolean, null: false
       field :can_assign_referral_tasks, Boolean, null: false
       field :can_create_referral_note, Boolean, null: false, description: 'Whether or not the user can create a note on this referral at the top level, i.e., not tied to a specific task.'
+      field :can_view_referral_in_target_project, Boolean, null: false, description: 'Whether the user can view this referral in the target project context (separately from their ability to view the project itself)'
```
martha: Thread to discuss - This arose because in testing locally, I created a scenario where the user had:
- permission to view outgoing referrals in the source project
- permission to view the target project
- but _no_ permission to view referrals in the target project.

This is a niche scenario that's unlikely to arise. I added support for it only because our permission model technically supports it.

Another simpler path, I think, would be to _always_ link to the referral in the "floating" context when linking from the outgoing referrals page. I considered that, but didn't go with that approach, because I was hoping for consistent behavior with the Your Dashboard page, which links to the project context when you have the appropriate permissions. 

Lmk what you think about all this, I wasn't quite sure of the neatest way forward.

drivers/hmis/app/models/hmis/auth_policies/context_loaders/ce_referral_source_project_loader.rb
```diff
@@ -14,7 +14,7 @@ def initialize
 
     def get(referral_id)
       preload([referral_id]) unless @cache.key?(referral_id)
-      @cache[referral_id] || raise("No source project found for referral #{referral_id}")
+      @cache[referral_id]
```
martha, line 17: Bug fix, I think: Referrals aren't required to have a source enrollment/project (the `source_enrollment` col is nullable in the db), so I think this should just return nil instead of raising.

drivers/hmis/app/models/hmis/auth_policies/context_loaders/hmis_project_access_group_loader.rb
```diff
@@ -38,7 +38,7 @@ def preload(project_ids)
       @cache.merge!(results)
 
       # For projects that don't have any access groups, add `nil` to the cache, so that we don't check the db again
-      no_access_project_ids = project_ids - results.keys
+      no_access_project_ids = new_project_ids - results.keys
```
martha, line 41: Bug fix: only add `nil` to the cache for project IDs we are right now loading for the first time! Otherwise we end up overwriting projects with nil when they actually _do_ have previously loaded access groups.

The change on line 305-306 of `drivers/hmis/app/graphql/types/hmis_schema/project.rb` revealed this. I guess there aren't any other times when we load multiple projects some of which have already been loaded, so it was a quiet bug before.

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -30,6 +30,21 @@ def can_view?
...
 
+    # Referrals that the user can view because they have can_view_outgoing_referral_details in the source project.
+    # Note that can_view_outgoing_referral_details grants full referral details,
+    # whereas can_manage_outgoing_referrals only grants summary level permission.
+    return true if source_project_permissions.include?(:can_view_outgoing_referral_details)
+
+    can_view_floating?
+  end
+
+  def can_view_in_target_project?
```
martha: `can_view_in_target_project?` added to support the scenario described [above](https://github.com/greenriver/hmis-warehouse/pull/6035/files#r2557085833). We can revert this if we don't think supporting that is necessary.

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -30,6 +30,21 @@ def can_view?
...
+    return false unless project_permissions.include?(:can_view_project)
+
+    project_permissions.include?(:can_view_referrals) || can_view_floating?
+  end
+
+  def can_view_floating?
```
ttoomey: This method name might be a little hard to understand. Maybe `can_view_via_assigned_step?` or perhaps just mark it private and then the name doesn't matter as much

drivers/hmis/app/models/hmis/role.rb
```diff
@@ -263,8 +263,17 @@ def self.permissions_with_descriptions
         category: 'Project Access',
         sub_category: 'Referrals',
       },
+      can_view_outgoing_referral_details: {
+        description: 'Ability to view all outgoing referrals from the project',
+        requirements: [:can_view_project],
```
ttoomey: suggesting adding a comment to clarify that the `can_view_project` requirement applies to the source project here, given that it's ambiguous in this case without diving into HmisPermissionLoader

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
ttoomey: Thanks for the explanation! I think this makes sense. However I don't feel confident in my understanding of the front-end nuances here.

ttoomey (approved):
Looks great, thank you!

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -110,6 +110,7 @@ def self.authorized?(object, ctx)
       can :view_units
       can :manage_incoming_referrals
       can :manage_outgoing_referrals
+      can :view_outgoing_referral_details
```
gigxz, line 113: I think this is a place where we have the opportunity to resolve an access field that mirrors the policy. The policy has method `can_view_outgoing_referral_summaries?`, which smartly encapsulates the logic for whether the user can view outgoing referral summaries in the project. However, that logic is duplicated in the frontend code which checks `canManageOutgoingReferrals || canViewOutgoingReferralDetails` to determine whether to show the outgoing referrals page. So even though it's encapsulated by the policy _on the backend_, we'd still need to change the logic in two codebases if it changes in the future. (For example, say we were to add a permission to allow users to view outgoing referral summaries without the ability to send referrals).

Using our existing patterns, it could be done like this using the composite permission helper on BaseAccess:

```suggestion
      can :view_outgoing_referral_details
      composite_perm :can_view_outgoing_referral_summaries, permissions: [:manage_outgoing_referrals, :view_outgoing_referral_details], mode: :any
```

That still has the problem of the logic existing in two places, but at least they are on the backend so the API surface is stable. It would be better to rely on the policy, which we haven't really established a pattern for, yet. Perhaps we could implement it as above with a TODO for issue #8067?

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -42,7 +43,8 @@ class Referral < GrdaWarehouseBase
       # What makes a referral viewable by a user?
       # - If they have can_view_referrals at the target project, OR
       # - If they have can_view_own_referrals, AND are assigned a step in the referral, OR
-      # - If they have can_view_own_referrals, AND are assigned to a swimlane that has a completed step in the referral
+      # - If they have can_view_own_referrals, AND are assigned to a swimlane that has a completed step in the referral, OR
+      # - If they have can_view_outgoing_referrals at the *source* project
```
gigxz:
```suggestion
      # - If they have can_view_outgoing_referral_details at the *source* project
```

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
martha: Thanks, good point. I think I'll take both suggestions - rename and make private.

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
gigxz: I think the code you proposed is reasonable if we are prioritizing always displaying the referral _in the target project_ whenever possible. However, I'm not sure I agree with that behavior.

I think it would be preferable to stay in the current context when possible. For example, if I am looking at a "source" project and click into an outgoing Referral, it should open in the project I'm already looking at. IMO that should give a better user experience, particularly for a workflow where somebody is reviewing several outgoing referrals from their project – it wouldn't really make sense to jump around to a bunch of different project dashboards if it can be avoided. It also would give better performance since we don't need to fetch all the project details required by each source project dashboard. This behavior would also be consistent with the Admin Dashboard which opens the Referrals in the current context, for similar reasons.


So, looking at where we link to Referrals, this is what I would propose:

* Target Project > Incoming Referrals Table
  * open Referral in current context (current behavior)
* Admin > All Referrals Table
  * open Referral in current context (current behavior)
* Source Project > Outgoing Referrals Table
  * open Referral in current context (proposed)
* User Dashboard
  * ~~open Referral in target project if viewable~~ (propose to get rid of this to simplify? I'm not sure it adds much value. It is made more complex now by the fact that the user may now be able to view the referral "in" either the source project, _or_ the target project, _or_ neither.)
  * open Referral in current "floating" context


Sorry that was a bit ambiguous in the issue description, I should have clarified earlier! Let me know what you think.


The `ReferralPage` is already rendered in multiple contexts, but it would need some minor tweaking because it currently assumes that if its IS in a project context, that the project is the target of the referral.

drivers/hmis/app/models/hmis/auth_policies/context_loaders/hmis_project_access_group_loader.rb
gigxz: Good find!

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
gigxz: Holding off on reviewing this policy change based on my other UX comment 👀

gigxz (commented):
looking good!

drivers/hmis/app/graphql/types/hmis_schema/ce_referral.rb
martha: Thanks @gigxz awesome, thanks for helping to think this through, I agree with your reasoning.

One more case to consider:
- Client referrals table
  - current behavior: links to referral in target project if user can view target project, otherwise links to floating referral. (Matches current behavior of User Dashboard referral links)
  - proposed change: match User Dashboard, always link to floating referral?

gigxz: Sure I think linking to floating from the client dashboard is fine! Maybe it would be more consistent to show it within the client dashboard, but I don't think it makes much of a difference.

martha: Ok, thanks! I made this simplification in the frontend PR, and removed the unnecessarily complicated `can_view_referral_in_target_project` here. https://github.com/greenriver/hmis-warehouse/pull/6035/commits/3a5f0968bd39a183fba5fb0af375cbf1793f814b

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
martha: 👍 reverted!

drivers/hmis/app/graphql/types/hmis_schema/project.rb
martha: Thanks for the suggestion with `composite_perm`, that's helpful! Implemented here https://github.com/greenriver/hmis-warehouse/pull/6035/commits/267e52a4f8cf26fb449384b4f347d27e6f51f9b3

gigxz (approved):

PR #6038, opened 2025-11-25
Adjust pagination location when called from JS

eanders:
## _Merging this PR_

## Description
* Fixes the `running` JS file that replaces the paginated report list for HMIS CSV reports to adjust the path back to the report path instead pointing to the JS file, which didn't work

## Type of change
* Bug fix

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -469,16 +469,17 @@ module ApplicationHelper
   # @param [ActiveRecord::Relation] scope The scope used to fetch the paginated list
   # @param [String] item_name The singular name of the item being listed, used for messages.
   # @param [String] list_partial The path to the partial used to render the list items.
+  # @param [String, nil] url Optional URL to use for pagination links. If not provided, uses current request path.
   #
   # @return [String] HTML markup for the paginated list with controls.
   #
   # @example Usage:
   #   render_paginated_list(scope: users, item_name: 'user', list_partial: 'users/card')
   #
-  def render_paginated_list(scope:, item_name:, list_partial:)
+  def render_paginated_list(scope:, item_name:, list_partial:, url: nil)
     pagy_sym = scope.is_a?(Array) ? :pagy_array : :pagy
     pagy, list = controller.send(pagy_sym, scope)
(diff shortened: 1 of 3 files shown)


app/helpers/application_helper.rb
```diff
@@ -489,22 +490,35 @@ def render_paginated_list(scope:, item_name:, list_partial:)
...
 
+  def pagy_url_for(pagy, page, absolute: false, **vars)
+    return super unless pagy.vars[:custom_url].present?
+
+    # Supports setting the url from the calling context, which is needed for HMIS exports since the JS is in a different context
+    custom_url = pagy.vars[:custom_url]
+    params = {}
+    params[pagy.vars[:page_param]] = page unless page == 1
+    "#{custom_url}#{params.any? ? "?#{params.to_query}" : ''}"
+  end
```
ttoomey, line 544: I think we might lose some of the query params here. How about this:

```suggestion
  # Overrides Pagy's default URL generation when custom_url is set via pagy.vars[:custom_url].
  # Used when rendering paginated lists via AJAX where the current request path differs from
  # the desired pagination base URL.
  #
  # @param [Pagy] pagy Pagination object
  # @param [Integer] page Target page number
  # @param [Boolean] absolute Ignored when custom_url is present
  # @param [Hash] vars Additional vars (ignored when custom_url is present)
  #
  # @return [String] Pagination URL with merged query parameters
  #
  # Parameter precedence: Current request params override custom URL params, allowing users
  # to maintain filter state while paginating. Page parameter is always set to the target page
  # (or removed if page is 1).
  #
  def pagy_url_for(pagy, page, absolute: false, **vars)
    custom_url = pagy.vars[:custom_url]
    return super unless custom_url

    uri = URI.parse(custom_url)
    base_params = Rack::Utils.parse_query(uri.query)
    merged_params = base_params.merge(request.query_parameters)
    page_key = pagy.vars[:page_param].to_s

    if page == 1
      merged_params.delete(page_key)
    else
      merged_params[page_key] = page
    end

    uri.query = merged_params.to_query.presence
    uri.to_s
  end
```

ttoomey (commented):
Thanks! I think small change needed to `pagy_url_for`

ttoomey (approved):
Thanks!

PR #6039, opened 2025-11-25
SFTP CLI class with 

eanders:
## _Merging this PR_

## Description
* Adds some additional packages for SSH/SFTP access from the host container to remote servers
* Adds a SFTP CLI ruby class so we can shell out to the host for data transfers instead of using Net SFTP that doesn't support the most-recent security protocols.
* Does not yet use the new implementation

## Type of change
* New feature

### DIFF
--- a/app/models/sftp/cli.rb
+++ b/app/models/sftp/cli.rb
@@ -7,6 +7,8 @@
 # frozen_string_literal: true
 
 require 'English'
+require 'shellwords'
+require 'open3'
 
 # Wrapper around the system's sftp CLI client that provides a Net::SFTP-compatible API.
 # Each operation executes as a separate sftp connection using batch mode (-b -) with heredoc input.
@@ -36,14 +38,19 @@ module Sftp
       @username = username
       @options = options
       @password_file = nil
+      @batch_files = []
     end
 
-    # Removes temporary password file if one was created for authentication.
+    # Removes temporary password file and batch files if any were created.
     def cleanup
-      return unless @password_file
-
-      File.unlink(@password_file.path) if File.exist?(@password_file.path)
-      @password_file = nil
+      if @password_file
+        File.unlink(@password_file.path) if File.exist?(@password_file.path)
+        @password_file = nil
+      end
+      @batch_files.each do |batch_file|
+        File.unlink(batch_file.path) if File.exist?(batch_file.path)
+      end
+      @batch_files.clear
     end
 
     # Returns a proxy object for directory operations (e.g., dir.glob).
@@ -76,10 +83,9 @@ module Sftp
     # Executes an sftp command and raises StatusException on failure.
     # Captures both stdout and stderr, filtering error messages for clearer exceptions.
     def execute_sftp_command(command)
-      cmd = build_sftp_command(command)
-      output = `#{cmd} 2>&1`
-      status = $CHILD_STATUS.exitstatus
-      return if status.zero?
+      cmd_parts = build_sftp_command_parts(command)
+      output, status = Open3.capture2e(*cmd_parts)
+      return if status.success?
 
       error_msg = output.lines.grep(/error|failed|denied|refused/i).join("\n")
       error_msg = output if error_msg.empty?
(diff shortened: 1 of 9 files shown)


app/models/sftp/cli.rb
```diff
@@ -0,0 +1,294 @@
...
+      @password_file = Tempfile.new('sftp_password')
+      @password_file.write(@options[:password] || '')
+      @password_file.close
+      File.chmod(0o600, @password_file.path)
+    end
+
+    # Escapes special characters in file paths for safe use in shell commands.
+    def escape_path(path)
+      path.to_s.gsub(' ', '\\ ').gsub('(', '\\(').gsub(')', '\\)').gsub('[', '\\[').gsub(']', '\\]').gsub('*', '\\*').gsub('?', '\\?')
+    end
```
ttoomey, line 180: I think this should be
```suggestion
   def escape_path(path)
    Shellwords.escape(path.to_s)
  end
```

app/models/sftp/cli.rb
```diff
@@ -0,0 +1,294 @@
...
+      else
+        sftp_parts
+      end
+    end
+
+    # Builds SSH option flags as separate -o flag=value pairs for the sftp command.
+    def build_ssh_options_parts
+      opts = []
+      opts << '-o' << 'PasswordAuthentication=yes' if password_auth?
+      opts << '-o' << 'StrictHostKeyChecking=no' if @options[:append_all_supported_algorithms]
```
ttoomey: Do we want `StrictHostKeyChecking=no`? It's disabling a security feature. Maybe use a clearly named option to control it like `verify_host_key` rather than `append_all_supported_algorithms`

app/models/sftp/cli.rb
```diff
@@ -0,0 +1,294 @@
...
+
+          size, date_str, filename = match.captures
+          filename = filename.strip
+          next unless filename.match?(pattern_regex)
+
+          files << RemoteFile.new(
+            name: filename,
+            directory: directory,
+            size: size.to_i,
+            createtime: parse_date(date_str),
```
ttoomey: `date_str` is probably mtime, not createtime. Also if parsing fails, `parse_date` returns current time. That seems kind of loose.

ttoomey (commented):
Looks promising, thanks!

app/models/sftp/cli.rb
```diff
@@ -0,0 +1,266 @@
...
+    # Creates a temporary file with restricted permissions (0600) containing the password.
+    # This file is used by sshpass to avoid exposing the password in process lists or command history.
+    def ensure_password_file
+      return if @password_file && File.exist?(@password_file.path)
+
+      @password_file = Tempfile.new('sftp_password')
+      @password_file.write(@options[:password] || '')
+      @password_file.close
+      File.chmod(0o600, @password_file.path)
+    end
```
ttoomey, line 174: password written to Tempfile before permissions are set creates a brief window where the file may be readable

```suggestion
    def ensure_password_file
      return if @password_file && File.exist?(@password_file.path)

      @password_file = Tempfile.new('sftp_password', mode: 0o600)
      @password_file.write(@options[:password] || '')
      @password_file.close
    end
```

app/models/sftp/cli.rb
```diff
@@ -0,0 +1,266 @@
...
+
+      # Parses the output of 'ls -la' command, preserving the full line as longname.
+      def parse_ls_long_output(output)
+        output.each_line.filter_map do |line|
+          line = line.strip
+          next if line.empty? || line.start_with?('sftp>')
+
+          # Extract filename from the end of ls -l output
+          # Format: -rw-r--r-- 1 user group 1234 Jan 15 10:30 filename
+          match = line.match(/\s(\S+)$/)
```
ttoomey: looks like this might break if filenames contain spaces

app/models/sftp/cli.rb
```diff
@@ -0,0 +1,266 @@
...
+          next if filename.empty? || filename.start_with?('sftp>')
+
+          RemoteFile.new(name: filename, longname: filename) if filename.match?(pattern_regex)
+        end
+      end
+
+      # Converts a glob pattern (e.g., "*.csv") to a regex for matching filenames.
+      def glob_to_regex(pattern)
+        regex_str = pattern.gsub('.', '\.').gsub('*', '.*').gsub('?', '.')
+        Regexp.new("^#{regex_str}$", Regexp::IGNORECASE)
```
ttoomey: technically safest to escape regex special characters in pattern?

```suggestion
      escaped = Regexp.escape(pattern)
      regex_str = escaped.gsub('\*', '.*').gsub('\?', '.')
      Regexp.new("^#{regex_str}$", Regexp::IGNORECASE)
```

ttoomey (approved):
Looks good, left some suggestions. I think the password temp file perms should be fixed before merging

app/models/sftp/cli.rb
eanders: Thanks! `Tempfile.new` doesn't take `mode` but I've moved the chmod above the write, hopefully that's sufficient!

PR #6040, opened 2025-11-25
8410 nc imports

ttoomey:
## _Merging this PR_

### Description
Improves memory footprint of HMIS CSV custom augmentation imports

- **Custom Augmentation Importer**: Added dedicated `apply_augmentation_updates` method that batches updates to reduce memory size. Also iterates from source data rather than scanning entire warehouse tables
- **Importer Helpers**: Extracted `prepare_destination_for_update`, `mark_record_dirty`, `track_updated_client`, and `augmentation_upsert_columns` methods for cleaner code organization
- renamed and updated feature documentation
- Added test coverage for custom augmentation imports
- Adds factory helper to build CSV data sets

### Type of Change
Refactor


spec/support/hmis_csv_fixture_factory.rb
```diff
@@ -0,0 +1,376 @@
...
+#   factory.export_start_date = Date.new(2024, 1, 1)
+#   factory.export_end_date = Date.new(2024, 3, 31)
+#   factory.add_client(personal_id: 'client-1', first_name: 'Test', last_name: 'User')
+#   factory.add_enrollment(enrollment_id: 'enroll-1', personal_id: 'client-1', entry_date: Date.new(2024, 1, 15))
+#   factory.add_custom_enrollment_augmentation(enrollment_id: 'enroll-1', personal_id: 'client-1', sexual_orientation: 1)
+#   path = factory.create!
+#   # use path with import_hmis_csv_fixture
+#   factory.cleanup!
+#
+class HmisCsvFixtureFactory
```
ttoomey, line 28: Avoids needing to create a whole pile of static fixture CSV files in drivers/hmis_csv_importer/spec/fixtures/files/twenty_twenty_six/

docs/features/hmis-csv-importer.md
```diff
@@ -0,0 +1,48 @@
+# HMIS CSV Importer
```
ttoomey, line 1: renamed and adjusted content from `drivers/hmis_csv_importer/README.md`

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -899,6 +889,107 @@ def involved_project_ids
...
+        if batch.present?
+          Rails.logger.info "Processing final batch of #{batch.count} augmentation records in bulk for #{file_name}"
+          process_batch!(destination_class, batch, file_name, type: 'updated', upsert: true, columns: upsert_columns, update_only: true)
+        end
+
+        # If we tracked any dirty enrollments (from Exit augmentation), bulk update them now
+        if klass.hud_key == :ExitID && dirty_enrollment_ids.present?
+          GrdaWarehouse::Hud::Enrollment.where(data_source_id: data_source.id, EnrollmentID: dirty_enrollment_ids).
+            update_all(processed_as: nil)
+        end
```
eanders: Where are `dirty_enrollment_ids` coming from in this context?

eanders: Oh `prepare_destination_for_update` calls `mark_record_dirty` which will add to `dirty_enrollment_ids`

eanders: Dirty enrollments also get processed in `apply_updates` .  I _think_ we can remove it from here, but maybe not.  If we keep it, we should probably flush the `dirty_enrollment_ids` after clearing them, or they'll be cleared again in `apply_updates`

spec/support/hmis_csv_fixture_factory.rb
eanders, line 1: This is an interesting idea @ttoomey.  It might be a path forward for generating fake data that could be ingested nightly as well.

A few things give me pause:
1. This needs to be versioned (this looks like an FY26 model)
2. This includes a ton of duplication from things like `app/models/concerns/hmis_structure/enrollment.rb`  I'd like to see it re-worked a bit to lean on existing infrastructure that we need to update yearly.  I think given the structure files that indicate column names, types, lengths, etc. and the `HudHelper`  I bet we could keep a similar API but avoid the duplication and need to update each cycle.

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
ttoomey: good catch, I'll also DRY up the repeated code into `flush_dirty_enrollments_for_exit`

spec/support/hmis_csv_fixture_factory.rb
ttoomey: thanks for the feedback; I've invested some more time to clean it up and reduce duplication. Look okay?

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -829,20 +832,9 @@ def involved_project_ids
...
-            when :ExitID
-              # These are tracked separately so we can bulk update them at the end
-              track_dirty_enrollment(destination.EnrollmentID)
-            end
-            # ensure we always have the right data_source_id
-            destination.data_source_id = data_source.id
+            prepare_destination_for_update(klass, destination)
             batch << destination
-            if batch.count == INSERT_BATCH_SIZE && !custom_augmentation?(klass)
+            if batch.count == INSERT_BATCH_SIZE
```
eanders: I wonder if `prepare_destination_for_update` should just return the destination client to make it more clear what's going on, then this could be:
```suggestion
batch << prepare_destination_for_update(klass, destination)
            if batch.count == INSERT_BATCH_SIZE
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
```diff
@@ -899,6 +885,103 @@ def involved_project_ids
...
+        up_rps: ((records / bm.real).round(3) unless records.zero?),
+        up_cpu: "#{(bm.total * 100.0 / bm.real).round}%",
+      }
+      importer_log.summary[file_name].merge!(stats)
+      Rails.logger.debug do
+        "  Updated #{destination_class.table_name} #{hash_as_log_str({ updated: records }.merge(stats).merge(log_ids))}"
+      end
+    end
+
+    private def prepare_destination_for_update(klass, destination)
```
eanders, line 959: Can you add some method documentation here?  It's simple, but it has some side effects that are probably worth noting.

spec/support/hmis_csv_fixture_factory.rb
eanders: Thanks.  I think this is much better.

eanders (approved):

drivers/hmis_csv_importer/app/models/hmis_csv_importer/importer/importer.rb
ttoomey: good thought, thanks

eanders (approved):

PR #6046, opened 2025-11-26
Append IDs to projects and organizations

eanders:
## _Merging this PR_

## Description
Adds two new importer extensions:
1. Append Project ID to Project Names
2. Append Organization ID to Organization Names

These are both just duplicates of the prepend versions.

## Type of change
* New feature


drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/append_organization_id.rb
```diff
@@ -0,0 +1,50 @@
...
+
+# frozen_string_literal: true
+
+module HmisCsvImporter::HmisCsvCleanup
+  class AppendOrganizationId < Base
+    def cleanup!
+      organization_batch = []
+
+      organization_scope.find_each do |organization|
+        organization.OrganizationName = "#{organization.OrganizationName} (#{organization.OrganizationID})"
```
ttoomey, line 15: should we handle empty names?
```suggestion
        organization.OrganizationName = "#{organization.OrganizationName.presence || 'Unnamed Organization'} (#{organization.OrganizationID})"
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/append_project_id.rb
```diff
@@ -0,0 +1,50 @@
...
+        project_batch << project
+      end
+
+      project_source.import(
+        project_batch,
+        on_duplicate_key_update: {
+          conflict_target: conflict_target(project_source),
+          columns: [:ProjectName, :source_hash],
+        },
+      )
```
ttoomey: This is probably fine, for very large number of projects we'd probably want to batch upserts too

```suggestion
      project_scope.find_in_batches do |project_batch|
        project_batch.each do |project|
          project.ProjectName = "#{project.ProjectName} (#{project.ProjectID})"
          project.set_source_hash
        end
    
        project_source.import(
          project_batch,
          on_duplicate_key_update: {
            conflict_target: conflict_target(project_source),
            columns: [:ProjectName, :source_hash],
          },
        )
      end
```

ttoomey (approved):
Code looks fine. I don't have all the context here but generally I don't know that it's optimal to mutate data just to accommodate what is essentially a display concern. 

Perhaps this is just the shortest path. Maybe in the future we consider if that can be accomplished with runtime application logic:
```ruby
class Project 
  def project_name
    return "#{read_attribute :project_name} (#{project_id})" if feature_flag?(:append_project_id_to_name)
    read_attribute :project_name}
  end

  def self.sort_by_name(direction: )
     current_scope.order(project_name: direction, project_id: direction) 
  end

end
```

drivers/hmis_csv_importer/app/models/hmis_csv_importer/hmis_csv_cleanup/append_organization_id.rb
eanders: `OrganizationName`, and `ProjectName`, are required fields in the CSV spec, so I don't think we need to handle that at this time.

eanders:
Thanks @ttoomey, I think I agree, though I'm not sure if there's benefit to having it encoded in the field for future data exchange.  Given that we've been prepending for a long-while and this is just an alternate version to append, I think we should go with this as is for now and revisit import extensions more globally in the future.

> Code looks fine. I don't have all the context here but generally I don't know that it's optimal to mutate data just to accommodate what is essentially a display concern.
> 
> Perhaps this is just the shortest path. Maybe in the future we consider if that can be accomplished with runtime application logic:
> 
> ```ruby
> class Project 
>   def project_name
>     return "#{read_attribute :project_name} (#{project_id})" if feature_flag?(:append_project_id_to_name)
>     read_attribute :project_name}
>   end
> 
>   def self.sort_by_name(direction: )
>      current_scope.order(project_name: direction, project_id: direction) 
>   end
> 
> end
> ```

ttoomey (approved):

PR #6063, opened 2025-12-05
Hotfix - add sex field to client cleanup script

dtgreiner:
## _Merging this PR_

## Description

Adds the Sex field to the client cleanup script. 

The Sex field is prioritized by (in order)
1. Sex value preference (0 or 1 preferred over 8, 9, 99)
2. Record timestamp (DateUpdated), preferring most recent
3. Record ID (lower is better)

## Type of change

Bug Fix

### DIFF
--- a/app/models/grda_warehouse/sex_selector.rb
+++ b/app/models/grda_warehouse/sex_selector.rb
@@ -31,14 +31,20 @@ module GrdaWarehouse
   class SexSelector
     Candidate = Struct.new(:value, :tie_breakers, keyword_init: true)
 
-    def self.call(dest_attr:, source_clients:, use_oldest: false)
-      new(dest_attr: dest_attr, source_clients: source_clients, use_oldest: use_oldest).call
+    VALID_PRIORITIZATION_METHODS = [:newest_first].freeze
+
+    def self.call(dest_attr:, source_clients:, prioritization_method: :newest_first)
+      new(dest_attr: dest_attr, source_clients: source_clients, prioritization_method: prioritization_method).call
     end
 
-    def initialize(dest_attr:, source_clients:, use_oldest: false)
+    def initialize(dest_attr:, source_clients:, prioritization_method: :newest_first)
       @dest_attr = dest_attr.with_indifferent_access
       @source_clients = Array.wrap(source_clients).map { |client| normalize_source(client) }
-      @use_oldest = use_oldest
+      @prioritization_method = prioritization_method
+
+      return if VALID_PRIORITIZATION_METHODS.include?(@prioritization_method)
+
+      raise ArgumentError, "Invalid prioritization_method: #{@prioritization_method}. Valid methods: #{VALID_PRIORITIZATION_METHODS.join(', ')}"
     end
 
     def call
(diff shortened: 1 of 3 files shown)


app/models/grda_warehouse/sex_selector.rb
```diff
@@ -0,0 +1,143 @@
...
+  #   99 => Data not collected
+  #
+  # This class handles data normalization for various source data formats
+  # (e.g., Hashes, ActiveRecord objects). The goal is to provide a consistent
+  # and reliable way to determine the most trustworthy Sex value from potentially
+  # conflicting source data.
+  class SexSelector
+    Candidate = Struct.new(:value, :tie_breakers, keyword_init: true)
+
+    def self.call(dest_attr:, source_clients:, use_oldest: false)
```
eanders: I wonder if instead of `use_oldest` we should accept a `prioritization_method:`.  Something like `prioritization_method: :newest_first`.  That way we could check for known methods, and raise if we get an unknown one, but then just call the method to return the sort order.  It could give us better flexibility long-term like "oldest except use DQ to prioritize" or "newest with a value" or something like that.

Also totally fine to leave this as is for now, we can revisit later, but since it's our first actual use of this, it feels like it might be worth making a bit more flexible.  For now we can just write the `:newest_first` method and raise if anything else is passed.

eanders (approved):
Looks great, thanks @dtgreiner !

app/models/grda_warehouse/sex_selector.rb
dtgreiner: @eanders I've added updates to use a prioritization method instead of the boolean use_older. It looks like the PR auto requested a new Approval from you after the change too, but tagging you here in case it didnt send a notification.

app/models/grda_warehouse/sex_selector.rb
```diff
@@ -0,0 +1,158 @@
...
+    end
+
+    def initialize(dest_attr:, source_clients:, prioritization_method: :newest_first)
+      @dest_attr = dest_attr.with_indifferent_access
+      @source_clients = Array.wrap(source_clients).map { |client| normalize_source(client) }
+      @prioritization_method = prioritization_method
+
+      return if VALID_PRIORITIZATION_METHODS.include?(@prioritization_method)
+
+      raise ArgumentError, "Invalid prioritization_method: #{@prioritization_method}. Valid methods: #{VALID_PRIORITIZATION_METHODS.join(', ')}"
```
eanders, line 47: Might be easier to read:
```suggestion
      raise ArgumentError, "Invalid prioritization_method: #{@prioritization_method}. Valid methods: #{VALID_PRIORITIZATION_METHODS.join(', ')}" unless VALID_PRIORITIZATION_METHODS.include?(@prioritization_method)
```

app/models/grda_warehouse/sex_selector.rb
```diff
@@ -0,0 +1,158 @@
...
+
+      base
+    end
+
+    def priority_modifier(source)
+      case @prioritization_method
+      when :newest_first
+        newest_first_priority_modifier(source)
+      else
+        raise ArgumentError, "Invalid prioritization_method: #{@prioritization_method}. Valid methods: #{VALID_PRIORITIZATION_METHODS.join(', ')}"
```
eanders, line 106: Given that we validated this in `initialize` we probably don't need this, but no harm.

eanders (approved):
@dtgreiner Thanks for making that change.  I think we should go with this as is, just dropped two thoughts on the future.

PR #6065, opened 2025-12-08
8395 hopwa caper fix v3

ttoomey:
## _Merging this PR_

### Description
Add household-level income/insurance aggregation and household-level deduping to HOPWA CAPER FY2026 reporting.

- Add household income/insurance array columns to HOPWA CAPER enrollments to simplify query logic and provide visibility into HoH on drilldowns
- cast `percent_ami` to integer to match hud code
- Use latest head-of-household per client when building program and supportive services sheets to deduplicate households across projects.
- Improve tests and documentation 

### Type of Change
New feature


app/models/concerns/sql_helper.rb
```diff
@@ -58,16 +57,4 @@ module SqlHelper
     q_set = SqlHelper.quote_sql_array(set, type: type)
     "#{field} <@ #{q_set} AND #{field} != #{empty_q_set}"
   end
-
-  # SQL condition to check if a field (a PG array) has ANY overlap with a given set.
-  # This is an INCLUSIVE match: at least one element in `field` must be in `set`.
-  # For an EXCLUSIVE match (all elements in `set`), use `non_empty_array_subset_condition`.
-  # Uses PostgreSQL's && (overlap) operator.
-  # The `set` cannot be empty, as this would always return false in SQL.
-  module_function def array_overlap_condition(field:, set:, type:)
```
ttoomey, line 67: I missed that this is redundant with the `active_record_extended` gem's `overlap` util. Removing

db/warehouse/migrate/20251207120000_add_household_aggregates_to_hopwa_caper_enrollments.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+class AddHouseholdAggregatesToHopwaCaperEnrollments < ActiveRecord::Migration[7.0]
+  def change
+    safety_assured do
```
ttoomey, line 5: hopwa caper enrollments should be relatively small

docs/features/hopwa_caper.md
```diff
@@ -12,6 +12,9 @@ The report lives under the HUD Reports, use `HopwaCaper::BaseController` as an e
 - **Household eligibility tagging:** `update_hopwa_eligibility` assigns a single HOPWA-eligible member per household (preferring an HIV-positive head of household). This resolves cases where multiple members qualify.
 - **Universe links:** Both staging models inherit from `HudReports::ReportClientBase`, exposing `as_report_members` helpers that register each record as a `HudReports::UniverseMember`. This connects records to HUD report drill-downs and client detail tables.
 
+## Household Counting Methodology
+Household counts follow the HUD HMIS reporting glossary definition: "distinct count of personal IDs of all heads of households." When a household has multiple enrollments during the reporting period (e.g., enrolling at different projects), they are counted only once by deduplicating on the HoH.
```
eanders, line 16: Not sure if it's worth calling out, but this can be affected by bad data.  If we are counting PersonalID instead of HouseholdID, then if there are multiple (or no) clients marked as `RelationshipToHoH = 1` in a household, the count can be inflated or deflated.

drivers/hopwa_caper/spec/models/generators/fy2026/hopwa_caper_tbra_spec.rb
```diff
@@ -155,6 +191,39 @@
     end
   end
 
+  context 'with same household re-enrolling in different projects' do
```
eanders, line 194: This feels very weird to me.  If a client is the only client in the household, then this is correct, but if the household composition changes in different projects, you may want to count the two households in different categories.   Maybe the intent here is just to get the latest enrollment for the client that overlaps the reporting universe?  In that case this makes sense, but isn't exactly a household related item.

eanders (approved):
@ttoomey I'm not super familiar with the spec for this, so called out a couple things that struck me as slightly odd, but nothing to stop getting this merged and deployed.

docs/features/hopwa_caper.md
ttoomey: Good point; that's how most of our reports work right? We assume folks have corrected the data before reporting

eanders: Yes, as much as possible we prefer data quality issues be handled outside of the reporting logic.  It can sometimes make it difficult to track down reporting issues, but means the logic can often follow the specification more closely.

ttoomey:
> @ttoomey I'm not super familiar with the spec for this, so called out a couple things that struck me as slightly odd, but nothing to stop getting this merged and deployed.

Thanks; Unfortunately there isn't really a spec; so if things are odd we should discuss.

drivers/hopwa_caper/spec/models/generators/fy2026/hopwa_caper_tbra_spec.rb
ttoomey: discussed on huddle, we think this is okay

PR #6069, opened 2025-12-08
official-docker-pkgs

ttoomey:
## _Merging this PR_

## Description
After upgrading docker desktop I could no longer use the legacy ECS deploy script due to a version incompatibility error: "Error response from daemon: client version 1.41 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version"

Replacing the `docker.io` package with the official docker package seems to resolve the issue

Also removes neovim; probably not needed in our images anymore as we have regular vim  

## Type of change
Dependency update

### DIFF
--- a/docker/app/Dockerfile
+++ b/docker/app/Dockerfile
@@ -193,6 +193,11 @@ FROM base AS dev-build
 
 USER root
 
+RUN install -m 0755 -d /etc/apt/keyrings \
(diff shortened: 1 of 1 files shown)


eanders:
I think I hit this same issue and my work-around was `export DOCKER_API_VERSION=1.43`, but maybe mine was in the opposite direction.

yakloinsteak (approved):
:100:

eanders (approved):

PR #6079, opened 2025-12-10
Revert SPM perf fix from r191

ttoomey:
## _Merging this PR_

## Description
Reverts changes from a34c383 (8568) due to a performance regression. Reverts all files except:
* new unit tests for dq and spm (`drivers/hud_apr/spec/models/fy2026/dq, drivers/hud_spm_report/spec/models/fy2024`)
* feature documentation for the SPM (`docs/features/hud_spm_report.md`)
* support for `user` param in `app/models/filters/filter_base.rb`

essentially this is
```
git revert a34c383
git co release-191 <preserved specs & docs in 32bd352>
```

## Type of change
Bug fix


drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2026/hdx_upload.rb
```diff
@@ -345,15 +333,15 @@ def dq(section, table_name, cell_name)
       dq_filter.data_source_ids = []
       dq_filter.project_ids = project_ids
 
-      generator = HudApr::Generators::Dq::Fy2026::Generator
+      generator = HudApr::Generators::Dq::Fy2024::Generator
```
eanders: I could be wrong, but I don't think we want to reference FY2024 here.

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2026/measure_six.rb
eanders, line 1: I also think we should keep this, no functionality, but it is helpful information.

eanders (approved):

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2026/hdx_upload.rb
ttoomey: good catch

drivers/hud_spm_report/app/models/hud_spm_report/generators/fy2026/measure_six.rb
ttoomey: thanks, I'll keep it. My hope is to un-revert this in r192

eanders (approved):

PR #6081, opened 2025-12-12
Ensure there is always a primary name after merge

martha:
## _Merging this PR_

## Description
- Update `MergeClientsJob` to always choose a primary name.
  - I found the actual desired functionality here to be a bit of a rabbit hole. Previously, we were retaining the retained client's name as primary. But, I agreed with the ticket's note that we should choose the most recently updated name as primary. But, when reconstructing CustomClientName records for clients that have gotten out of sync, we don't really know which was most recently updated, so I kept using the retained client's name in that case. This might be getting overly complicated so I'm eager for feedback!
- Add test coverage.

## Type of change
Bug fix (that also entails functional change)


martha:
(Converting back to draft while cleaning up test failures)

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -164,20 +167,44 @@ def merge_and_find_primary_name
...
-      client_val = [client_to_retain.first_name, client_to_retain.middle_name, client_to_retain.last_name, client_to_retain.name_suffix]
-      name_scope.sort_by(&:id).each do |name|
-        custom_client_name_val = [name.first, name.middle, name.last, name.suffix]
-        # consider this name "primary" it matches the name on the client_to_retain's Client record
-        primary = (client_val == custom_client_name_val) && !primary_found
+      # Sort the names in order to choose one primary name. Desired behavior:
+      # - Names marked as primary (either because they were previously marked as primary pre-merge, or because the job marked it primary since the client didn't have a primary name pre-merge) are prioritized above other names
+      # - CustomClientName records that existed pre-merge are prioritized above name records created by this job. (DateUpdated on name records created by this job is set to now, which isn't meaningful)
+      # - Within name records newly created by this merge job, names on the client to retain are prioritized higher
+      # - Names that were updated more recently are prioritized higher (with tiebreak on ID)
```
gigxz: > I found the actual desired functionality here to be a bit of a rabbit hole. Previously, we were retaining the retained client's name as primary. But, I agreed with the ticket's note that we should choose the most recently updated name as primary. But, when reconstructing CustomClientName records for clients that have gotten out of sync, we don't really know which was most recently updated, so I kept using the retained client's name in that case. This might be getting overly complicated so I'm eager for feedback!


I think it would be a good idea to simplify this. This logic would be pretty hard to communicate to the customer in merge documentation, for example. The current (draft) user documentation for HMIS merge states that it uses the same logic as the Warehouse rollup when picking which attributes to "keep" on the retained client.

That's true for the most part, since we use `GrdaWarehouse::Tasks::ClientCleanup.new.choose_attributes_from_sources` to assign "best" attributes to the retained client. Now that I'm digging into this, I see that under the hood that method is in fact updating the retained client's name attributes based on [choose_best_name](https://github.com/greenriver/hmis-warehouse/blob/3989d35fef3e0aa16b1fbf6ce5926aafd31ff895/app/models/grda_warehouse/tasks/client_cleanup.rb#L421). 

I'd say that **that** is the name that we ultimately want to mark as primary (so we can keep the logic aligned with ClientCleanup), and any other names should be retained as non-primary names. I think that's what the old behavior was doing, but maybe it wasn't handling custom names correctly? I don't have it totally sorted out, but you can take it from here..

gigxz: Looking at the existing code: https://github.com/greenriver/hmis-warehouse/blob/af6c2ab8d1dd7251864390e7f48820c1e1b8b031/drivers/hmis/app/jobs/hmis/merge_clients_job.rb#L167-L180
The logic looks okay to me, knowing that client_to_retain has already been updated with the "chosen" name attributes.

martha: Ok. Makes sense. The ClientCleanup logic for choosing a "best" name does not look at CustomClientName records, so it doesn't pay attention to which names were previously marked "primary" before the merge.

To double check my understanding, in the following edge case:
- Client has name A on their client record, but they don't have name A in their CustomClientName records. 
- Instead they have a CustomClientName B that is marked "primary." 
- They are merging with another client who has name C on both their client record and their primary name record. 
- C was created later, so ClientCleanup is going to choose A's name as the "best" name. 

that means that post merge, we should ignore the fact that B was marked "primary," and choose A as the "primary" name for the merged client.

gigxz: Yeah I think that's correct

drivers/hmis/app/models/hmis/hud/client.rb
```diff
@@ -190,6 +190,10 @@ class << self
   def build_primary_custom_client_name
     return unless names.empty?
 
+    build_custom_client_name_from_client_record
+  end
+
+  def build_custom_client_name_from_client_record
     names.new(
```
gigxz, line 191: The method naming is confusing to me, `build_primary_custom_client_name` and `build_custom_client_name_from_client_record` effectively do the same thing, they build a primary CustomClientName based on the client record.

Since build_primary_custom_client_name is only called from one place in the codebase and it already checks presence of names first, maybe we can just remove line 191 instead of adding the second method?

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -139,18 +139,19 @@ def update_oldest_client_with_merged_attributes
...
+        name = client.build_custom_client_name_from_client_record
         name.CustomClientNameID = Hmis::Hud::Base.generate_uuid
-        name.primary = client.id == client_to_retain.id
         name
       end.compact
 
       # Dedup and save the new name records
-      dedup_unpersisted(unpersisted_name_records).map(&:save!)
+      deduped_names = dedup_unpersisted(unpersisted_name_records)
+      Hmis::Hud::CustomClientName.import!(deduped_names, validate: false, timestamps: true) if deduped_names.any?
```
gigxz: I think we don't want to record timestamps, since the client method sets the timestamps on the CustomClientName based on date created/updated of the client record. In most cases that should be more accurate, since if a Client record doesn't have any name records, we can only assume the name attributes were created/updated based on the timestamps on the Client record itself.

gigxz: Just to check my understanding here, this import now results in some clients having multiple `primary` names, but that's OK because we the loop on line 170 will ensure that exactly 1 is primary? Maybe worth adding a comment explaining that, if I'm understanding correctly

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
```diff
@@ -166,18 +167,23 @@ def merge_and_find_primary_name
...
-        custom_client_name_val = [name.first, name.middle, name.last, name.suffix]
-        # consider this name "primary" it matches the name on the client_to_retain's Client record
-        primary = (client_val == custom_client_name_val) && !primary_found
+        # consider this name "primary" if it matches the name on the client_to_retain's Client record,
+        # which was already set to the chosen "best" name by update_oldest_client_with_merged_attributes
+        primary = names_match?(client_to_retain, name) && !primary_found
 
         name.client = client_to_retain
         name.primary = primary ? true : false
         name.save!(validate: false)
```
gigxz, line 180: same as comment above, I think we should switch this to not update the timestamps on the name records

gigxz (commented):
looks good! a few comments below

drivers/hmis/app/jobs/hmis/merge_clients_job.rb
martha: I agreed about the above comment, but I have a question about this one. Throughout the rest of `merge_clients_job`, we don't manually prevent updating timestamps on related records. For example if the merge updates Services or Assessments, pointing them at a new PersonalID, those changed records have their DateUpdated timestamps updated. 

Maybe that is a bug? If so, I think it may be out of the scope of this change, but would be happy to create a separate ticket. For this PR, it seems preferable to keep the timestamp behavior for name record updates as it was before, which is consistent with other record updates made by this job.

gigxz: Ah okay that's a good point, it should be addressed more holistically if we want to change that.I agree leave it as-is then. I don't think we need a ticket it's not a priority. Thanks for pushing back! :)

gigxz (approved):

PR #6066, opened 2025-12-08
Add `receiving_opportunities` scope to CandidatePool

martha:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

## Description
See linked ticket for issue description.

Suggested fix:
- Add a new scope to Candidate Pool, `receiving_referrals`, as an alternative to the `active` scope
- Exclude candidate pools from this scope whose only active opportunities are locked.
  - Candidate pools that are still tied to unit groups will still be included, regardless of capacity.
- Use the new scope when returning client eligibility:
  - in the global waitlist
  - when using the "Calculate client eligibility" button
- Continue to use the old `active` scope in the processing jobs, so waitlists for stale locked opportunities don't fall out-of-date.

## Type of change
[//]: # (e.g., Bug fix, New feature, Documentation, Code clean-up, Dependency update)
Bug fix

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [x] If adding a new endpoint / exposing data in a new way, I have:
  - [x] ensured the API can't leak data from other data sources
  - [x] ensured this does not introduce N+1s
  - [x] ensured permissions and visibility checks are performed in the right places
- [x] Any major architectural changes are supported by an approved ADR (Architectural Decision Record) 
- [x] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [x] I have provided testing instructions in this PR or the related issue (or not applicable)

[//]: # NOTE: system tests may fail if there is no branch on the hmis-frontend that matches the Source  or Target branch of this PR. This is expected

PR #6083, opened 2025-12-12
Fix tests from merge conflict

gigxz:
## _Merging this PR_

## Description
The following PRs resulted in tests failing on release-192 due to factory change:
- https://github.com/greenriver/hmis-warehouse/pull/6066
- https://github.com/greenriver/hmis-warehouse/pull/5904

See test failures: https://github.com/greenriver/hmis-warehouse/pull/6074

## Type of change
Tests

### DIFF
--- a/drivers/hmis/spec/models/hmis/ce/candidate_pool_spec.rb
+++ b/drivers/hmis/spec/models/hmis/ce/candidate_pool_spec.rb
@@ -6,13 +6,17 @@ RSpec.describe Hmis::Ce::Match::CandidatePool do
   let!(:ds1) { create :hmis_data_source }
   let!(:p1) { create :hmis_hud_project, data_source: ds1, project_type: 13 }
   let!(:project_config) { create :hmis_project_ce_config, supports_waitlist_referrals: true, project: p1 }
+  # Initial setup: project with 1 unit group with 2 units (no opportunities/candidate pools yet)
+  let!(:ug) { create :hmis_unit_group, project: p1 }
+  let!(:unit1) { create :hmis_unit, project: p1, unit_group: ug }
+  let!(:unit2) { create :hmis_unit, project: p1, unit_group: ug }
 
   describe 'scopes' do
     let!(:pool) { create :hmis_ce_match_candidate_pool }
 
     describe '.active_for_maintenance' do
       context 'with open opportunities' do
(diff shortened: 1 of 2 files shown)


drivers/hmis/spec/models/hmis/ce/candidate_pool_spec.rb
```diff
@@ -6,29 +6,33 @@
   let!(:ds1) { create :hmis_data_source }
   let!(:p1) { create :hmis_hud_project, data_source: ds1, project_type: 13 }
   let!(:project_config) { create :hmis_project_ce_config, supports_waitlist_referrals: true, project: p1 }
+  # Initial setup: project with 1 unit group with 2 units (no opportunities/candidate pools yet)
```
gigxz, line 9: @martha I _think_ this update keeps the test logic in place, but it could use another pair of eyes.

The tests like `.active_for_maintenance with open opportunities` are testing against `opp` which references a candidate pool directly, and belongs to a unit group that does _not_ reference a candidate pool. I am pretty sure that matches the old behavior.

Tests that redefine `let!(:ug)` are the only ones that assign a candidate pool to the unit group.

martha: Looks good, I think your changes preserve the previous logic. I might propose an update to the the setup though, now that I am looking at this again. I think the the normal/default scenario we are testing for is "unit group associated with a non-stale candidate pool," so what if we make the initial setup reflect the normal scenario, and then the additional contexts modify from there, to be more like real life? I played around with this locally and pushed up a proposal, https://github.com/greenriver/hmis-warehouse/pull/6084/changes

Let me know what you think. I'm also happy to merge what you have as-is.

martha (approved):

drivers/hmis/spec/models/hmis/ce/candidate_pool_spec.rb
gigxz: I was thinking along the same lines, the current tests weren't as realistic. The update looks good to me!

gigxz: I'm going to merge this to fix builds on release-192, we can merge your test refactor separately

PR #6086, opened 2025-12-14
8649 HUD report checkpoints

ttoomey:
## _Merging this PR_

## Description

Adds runtime checkpoints for HUD report generation, supports accurate duration reporting even when report job is retried after interruption.

- **SPM Report Generation**: Adds checkpoint tracking for snapshot creation phase
- **Report Instance State Management**: `start_report` preserves `started_at` timestamp on retries to prevent duration inflation
- **HUD Report Authorization**: Break ground on policy class with `can_view_checkpoints?` method
- **Checkpoint UI**: Adds collapsible table on report detail page showing individual checkpoint names, status, runtime, and timestamps

### Type of Change

New feature


ttoomey:
### What Changed

Report duration now uses phase-level timing instead of simple start-to-complete, so retries and gaps don't inflate the displayed duration.

### Testing Focus

**Primary Testing** (Must verify):
- Completed reports show accurate duration in the status display
- Retried reports don't inflate duration (preserves original start time)

**Regression Check** (Quick verification):
- Report generation completes successfully
- Report status messages display correctly


### Test Steps

1. **Normal Report Completion**
   - Generate a new HUD report (any type, let's say SPM)
   - Wait for completion
   - Check the report status shows "Completed in [duration]"
   - Verify the duration matches the actual time taken

2. **Retried Report Duration**
   - No easy way to verify this by QA alone as we have no mechanism to stop / retry jobs

4. **Report Status Display**
   - Check that completed reports show: "Completed in [duration] [timestamp]"
   - Verify the format matches previous behavior (duration + timestamp)
   - Ensure older reports (before this feature was deployed) display the expected times

app/models/grda_warehouse/auth_policies/hud_report_policy.rb
```diff
@@ -0,0 +1,34 @@
+# frozen_string_literal: true
+
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+class GrdaWarehouse::AuthPolicies::HudReportPolicy < GrdaWarehouse::AuthPolicies::BasePolicy
```
ttoomey, line 9: breaking ground on a policy class for hud reports. Not sure if the permission set is gathered correctly

app/models/hud_reports/report_instance.rb
```diff
@@ -143,7 +159,28 @@ def start(question, tables)
     end
 
     def start_report
-      update!(state: 'Started', started_at: Time.current)
+      # Only set started_at on the very first run
+      self.started_at ||= Time.current
```
ttoomey, line 163: I think it's more useful to display when the report started running the first time

app/models/grda_warehouse/auth_policies/hud_report_policy.rb
```diff
@@ -0,0 +1,34 @@
...
+    ensure_arg_type!(arg, HudReports::ReportInstance)
+  end
+
+  memoize def resource_permissions
+    results = Set.new
+    resource.project_ids.each do |project_id|
+      results.merge(context.project_role_permissions(project_id))
+    end
+    results
+  end
```
eanders: What is this trying to check?
Generally speaking, until now at least, access to HUD reports is governed by `can_view_all_hud_reports` or `can_view_own_hud_reports`.  If you have either of these, you can see all HUD report categories, and either all reports, or those you ran.  Within a report, you have access to run them against projects you can report on, based on the `can_view_assigned_reports` permission.

eanders (commented):
@ttoomey I think this looks like a good start to better logging and display for HUD report runs.  I know you've already made the SPM have better resume logic, but it feels like calling these checkpoints could imply that any reports using this would have similar logic.  Totally fine for now that they don't, but it probably makes sense to queue up some work to ensure each checkpoint completion is aligned with a point at which the report can be resumed consistently.

app/models/grda_warehouse/auth_policies/hud_report_policy.rb
ttoomey: This is likely incorrect but the idea is to calculate the cumulative permissions the user holds for the projects included in the report instance. Happy to huddle on it

Looking at it again we should probably check the legacy role based permissions too. I don't think there is a notion of global permissions for ACL users at in terms of how it's supposed to be used, ACLs should be applied to some kind of resource right?

Also this code is doing a union of permissions but I think in this case it should be an intersection to require all projects to have the permission.

so probably:
```ruby
  memoize def resource_permissions
    # intersection, the set of all common permissions
    result = resource.project_ids.map do |project_id|
      context.project_role_permissions(project_id)
    end.reduce(:&)

    (result || Set.new).merge(context.legacy_permissions)
  end
```

ttoomey: Maybe this is more similar to our current usage

```
  # policy
  memoize def resource_permissions
    context.system_permissions
  end

  # acl context
  def system_permissions
    permissions_for_collection_ids(system_collection_ids(:data_sources))
  end
  # legacy context etc
```

ttoomey:
>  Totally fine for now that they don't, but it probably makes sense to queue up some work to ensure each checkpoint completion is aligned with a point at which the report can be resumed consistently.

@eanders seems reasonable; I adjusted track progress to skip completed sections. It's somewhat redundant with `completed_questions` and `snapshot_status` so maybe we can make another pass to simplify it

app/models/grda_warehouse/auth_policies/user_acl_context.rb
```diff
@@ -32,6 +32,10 @@ def initialize(user)
     permissions_for_collection_ids(collection_ids)
   end
 
+  def system_permissions
+    permissions_for_collection_ids(system_collection_ids(:data_sources))
+  end
```
eanders: Can you explain what's going on here?  Why are these called `system_permissions` but only referencing the system collection for data sources?  (Maybe I'm missing something.)  It feels like the intention is to somehow target permissions that don't need to reference a collection.

ttoomey: Sorry for the iteration regarding the hud report policy; I'm casting around for a "resource" to check here when I guess there isn't really one. I don't know what conventions people follow for this role but we shouldn't guess.

For now best we can do is to just wrap the methods on user. We can create a follow-up issue for how to deal with such global or system wide permissions

ttoomey:
@eanders - if the policy looks okay I'd like to get this to QA; although I can old off on merging to 192

eanders (approved):
Looks ok to me.  But yes, let's QA before merging, just to be safe.
QA should include all HUD reports I think.

PR #6092, opened 2025-12-15
bugfix

ttoomey:
## _Merging this PR_

## Description
Fixes a bug where a user cannot download an export of the drill down for a report created by another user.

## Type of change
Bug fix

### DIFF
--- a/drivers/hud_spm_report/app/models/hud_spm_report/document_exports/cell_detail_export.rb
+++ b/drivers/hud_spm_report/app/models/hud_spm_report/document_exports/cell_detail_export.rb
@@ -4,7 +4,7 @@ module HudSpmReport
   module DocumentExports
     class CellDetailExport < ::GrdaWarehouse::DocumentExport
       def authorized?
-        report.user_id == user_id && user.can_view_any_reports?
+        user.can_view_any_reports?
       end
 
       def perform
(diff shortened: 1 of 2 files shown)


drivers/hud_spm_report/app/models/hud_spm_report/document_exports/cell_detail_export.rb
```diff
@@ -4,7 +4,7 @@ module HudSpmReport
   module DocumentExports
     class CellDetailExport < ::GrdaWarehouse::DocumentExport
       def authorized?
-        report.user_id == user_id && user.can_view_any_reports?
```
ttoomey, line 7: This was too restrictive. User restriction seems to be handled `DocumentExportsController#export_scope`

eanders: You should not be able to download an export created by someone else, but it should create a new download for you, is that not working?

ttoomey: @eanders you can access a report created by someone else but, when you try and create an cell detail export, the authorization check runs and fails because the export user id doesn't match the user id. Happy to huddle if this is unclear

eanders (approved):

PR #6093, opened 2025-12-15
Persist cached MCI Unique ID on clearance

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8552

Clearance results now include an MCI Unique ID if one is available. Cache that mci_id=>mci_uniq_id mapping, and store associated mci unique id (if any) when client record is created/updated with a new MCI ID value.

This is a "best effort" attempt to associate newly cleared clients to existing MCI Unique IDs immediately when possible. If unable, the expectation is still that an MCI Unique ID should be assigned (or changed) with the next WarehouseChangesJob run.

## Type of change
New feature

- [x] I have run the code that is being changed under ideal conditions, and it doesn't fail -- sort of. I had to stub out a fair amount of code locally to run the code. Testing will be primarily on staging environment where API connection exists


drivers/hmis/app/graphql/mutations/ac_hmis/clear_mci.rb
```diff
@@ -24,13 +24,10 @@ def resolve(input:)
       return { errors: errors } if errors.any?
 
       mci = HmisExternalApis::AcHmis::Mci.new
-      response = mci.clearance(input.to_client)
+      response = mci.clearance(input.to_client, minimum_score: MATCH_THRESHOLD)
```
gigxz, line 27: Moved threshold filtering into MCI class, so we don't unnecessarily store mappings in the cache that we don't need. For example, clearance may return 50 results but only 5 of them are above the threshold to display to the user. This should also increase performance since we now only do ID/client lookups for the results that meet the threshold.

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mci.rb
```diff
@@ -76,6 +83,15 @@ def clearance(client)
       end
```
martha: maybe add `compact` here since the `next` on line 71 means the array may contain nils?

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mci.rb
```diff
@@ -76,6 +83,15 @@ def clearance(client)
...
     end
 
+    # Retrieve a cached MCI Unique ID for a given MCI ID. This was added as an optimization to try to associate
+    # newly cleared clients to existing MCI Unique IDs immediately when possible. (Rather than waiting for the next WarehouseChangesJob to run.)
```
martha, line 87: Maybe add one more sentence of comment here about why we decided on the cache approach (instead of the alternative discussed of threading it through the frontend/backend/api/client processor)

martha (approved):
🚀

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mci.rb
gigxz: good catch!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/mci.rb
gigxz: Skipping this, I think having the ability to click through to the PR and associated GitHub issue is probably OK. I'll update the issue to make it clearer why we decided on this approach

PR #6094, opened 2025-12-15
Accept unit type when creating a unit group

martha:
## _Merging this PR_

## Description
- Accept unit type when creating a unit group
- Stop accepting unit type when adding units to a unit group

## Type of change
UX clean-up


drivers/hmis/app/graphql/mutations/create_units.rb
```diff
@@ -55,15 +50,7 @@ def resolve(input:)
...
-      Hmis::Unit.transaction do
-        # If unit group doesn't have a unit type directly associated, add it
-        if unit_group.unit_type.nil?
-          unit_group.unit_type = unit_type
-          unit_group.save!
-        end
-
-        units.each(&:save!)
-      end
+      units.each(&:save!)
```
gigxz: I think we'd still want these saves to happen in a transaction

drivers/hmis/spec/support/ce_system_test_helper.rb
```diff
@@ -65,7 +65,7 @@
 
   let!(:sro_type) { create(:hmis_unit_type, description: 'SRO') }
   let!(:workflow_template) { create(:hmis_workflow_definition_template, data_source: ds1) }
-  let!(:unit_group) { create(:hmis_unit_group, project: target_project, name: 'SROs', workflow_template: workflow_template, candidate_pool: score_pool) }
+  let!(:unit_group) { create(:hmis_unit_group, project: target_project, name: 'SROs', workflow_template: workflow_template, candidate_pool: score_pool, unit_type: sro_type) }
```
gigxz, line 68: This is fine to add, but I think since unit type is expected to exist on all unit groups, it would make sense to adjust the default `hmis_unit_group` factory to have `unit_type { association :hmis_unit_type }` so it creates an associated unit type by default

drivers/hmis/app/graphql/schema.graphql
```diff
@@ -12911,6 +12911,7 @@ input UnitGroupInput {
   directReferralWorkflowTemplateIdentifier: String
   name: String
   projectId: ID
+  unitTypeId: ID
   workflowTemplateIdentifier: String
 }
```
gigxz, line 12917: While reviewing the frontend I noticed something with the backend to update, which is I think [this field](https://github.com/greenriver/hmis-warehouse/blob/66d5ecc8bc5e54ac646e440a0b7339ab726b92a7/drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb#L20) has an N+1, we can add a data loader method to resolve unit_type on the unit_group to resolve

drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb
```diff
@@ -229,10 +229,10 @@ def complete_ce_staff_confirm_success_step(client_name, unit)
...
 
     context 'when target project has a UnitGroup with no availability' do
-      let!(:unavailable_unit_group) { create(:hmis_unit_group, project: target_project, workflow_template: admin_assign_workflow_template) }
-      let!(:unit_without_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
-      let!(:unit_with_closed_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
-      let!(:unit_with_locked_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group) }
+      let!(:unavailable_unit_group) { create(:hmis_unit_group, project: target_project, workflow_template: admin_assign_workflow_template, unit_type: sro_type) }
+      let!(:unit_without_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group, unit_type: sro_type) }
+      let!(:unit_with_closed_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group, unit_type: sro_type) }
+      let!(:unit_with_locked_opportunity) { create(:hmis_unit, project: target_project, unit_group: unavailable_unit_group, unit_type: sro_type) }
```
gigxz, line 235: Hmm this is a bit of a bummer that we need to manage the redundant reference in tests. At least it doesn't seem like it caused a ton of churn, and if/when we decide to address the descoped piece of the ticket (removing the redundant reference) we may be able to clean this up

gigxz (approved):

drivers/hmis/spec/support/ce_system_test_helper.rb
martha: This makes sense but I think it's likely to result in a lot of test failures that require churn. Would it be okay to leave this as a todo for the descoped piece of the ticket? (Edit: I pushed this change, I'd like to wait to see how many tests fail before deciding whether to bring this in scope or not)

gigxz: sure that sounds reasonable!

martha: Link to failures: https://github.com/greenriver/hmis-warehouse/actions/runs/20281114793/job/58243066974?pr=6094

The failures arise because anytime we create a unit and a unit group in a spec file, if we specify unit type on the unit group, then all units in the group have to match that type. (related to your other comment below https://github.com/greenriver/hmis-warehouse/pull/6094#discussion_r2623611766)

I think/hope that if we address the descoped part of this ticket, removing the unit type column from the unit model, these tests will be easier to fix with less churn.

drivers/hmis/app/graphql/schema.graphql
martha: Thanks that's a good point! 

While I was trying to test this fix, I noticed that our existing test coverage doesn't match the query the frontend is actually making all that well; and there are also other n+1 problems on the `availability` and `capacity` fields.

I started going down this road and decided to split it out into a separate diff to make review easier, since you had already reviewed the code in this PR and it feels like these changes are expanding the scope. That's here, I'm still working on cleaning it up for review: https://github.com/greenriver/hmis-warehouse/pull/6098

gigxz: Ah okay I didn't realize there was more to it. THanks for splitting to another branch!

PR #6097, opened 2025-12-16
8663 support job cancellation

ttoomey:
## _Merging this PR_

### Description
Adds a job cancellation mechanism allowing administrators to cancel running delayed jobs 

### Functional Areas
- Added `cancel` action and UI controls to the delayed jobs dashboard.
- Added `handle_cancellation!` in `ApplicationJob` based on `cancellation_requested_at` timestamp.
- Added cancellation checks during report checkpoint progress tracking to abort execution if requested.
- Fixed up delayed job index page view
- Instrumented the HUD report runner and SPM enrollment filters with cancellation checkpoints to support mid-run stopping.
- interruptible jobs check for stopping workers and raise to exit and reschedule themselves
- Adds a dj plugins to track workers, enabling detection of stopping workers
- Corrects infinite loop in DJ processing in CI

### Type of Change
New feature


app/views/admin/delayed_jobs/_row.haml
```diff
@@ -0,0 +1,63 @@
...
+      %strong Failed:
+      = job.failed_at
+    - if job.cancellation_requested_at
+      %br
+      %span.icon-cancel-circled.text-warning
+      %strong Cancellation Requested:
+      = job.cancellation_requested_at
+  %td= job.queue
+  %td= job.priority
+  - if can_manage_config?
```
ttoomey, line 53: this is new, hiding the error for non-admins

app/jobs/application_job.rb
```diff
@@ -7,4 +7,19 @@
...
+
+  protected
+
+  def handle_cancellation!
+    return unless provider_job_id.present?
+
+    # We can't access the job record directly from the instance easily without
+    # looking it up.
+    job = Delayed::Job.find_by(id: provider_job_id)
+    job&.handle_cancellation!
```
eanders: I'm probably missing something, but isn't this a circular call?
I see it's also defined in the monkey patch, so maybe that's what get's called.  If so, it looks like it raises, which maybe would be caught by `discard_on JobCancelled`.  We don't have the discard gem, but is `discard_on` part of delayed job?  How is this actually stopping the worker?

Maybe some descriptive documentation would be helpful.

app/models/hud_reports/report_instance.rb
```diff
@@ -168,6 +168,8 @@ def start_report
     def track_progress(checkpoint_name)
       raise unless checkpoint_name.present?
 
+      related_job&.handle_cancellation!
```
eanders: Ok, so this is what will end up raising, which makes sense, that could kill the worker, but it looks like it's only implemented for HUD reports, but the option to cancel might show on all jobs.

app/jobs/application_job.rb
ttoomey: So this is in ActiveJob::Base; it delegates `handle_cancellation!` to the delayed job model which is where the actual cancellation logic and state live. Because these are different models, the call is not circular. 

The `discard_on` directive in active job is not related to the `discard` gem but actually part of the standard rails active job API. [See docs](https://guides.rubyonrails.org/active_job_basics.html#retrying-or-discarding-failed-jobs)

I'll add some documentation to help explain

eanders: Thanks!  
I think some docs will make this more clear, and nice to know about `discard_on`.

app/models/hud_reports/report_instance.rb
ttoomey: It shouldn't kill the worker, if I understand the code, it seems to rescue from the exception ([ref](https://github.com/rails/rails/blob/efe7f5241b751a70cccc86306fb659ee7c4094e2/activejob/lib/active_job/exceptions.rb#L109)).

However we can add the distinction between cancellable vs interruptible. With this change, all jobs can be cancelled before they start (due to `before_perform`) but only hud report jobs are interruptible. We could extend this to other jobs types to support cancellation

eanders: That's a good distinction (the cancel the job rather than kill the worker.)  
I think this is fine as is.  In the future we should extend it to:
1. All HUD reports
2. Importers (can't tell you how many times people have asked us to stop an upload they just uploaded.)
3. Any reports that take a long time to run

ttoomey:
changes since last review:

- **Job Management Dashboard**: Adds concurrency-safe actions and guards in the controller
- **HUD SPM Report**:  check for interruption during enrollment generation.
- **Job Metadata Extraction**: docs and support for varying payload formats (ActiveJob vs. native Delayed Job).
- **Admin Performance**: Optimization for batch loading only required users 
- ** Job Execution **: supports interruptible? execution

ttoomey:
@eanders 
considering additional enhancement to this branch; I feel like it would be good to also trap sigterm and then retry the job; this would allow long-running jobs to stop gracefully when the pod is evicted. I believe this would help the case where a job needs more memory, VPA tries to give it more, but the job keeps running in a degraded state rather than stopping on sigterm

spec/models/active_record_extended_merge_regression_spec.rb
```diff
@@ -14,11 +14,11 @@
   include_context 'report context'
 
   before(:all) do
-    setup(default_setup_path)
+    setup(default_setup_path) if ENV['SKIP_AREL_SQL_LITERAL_INITIALIZER'] == '1'
```
eanders: Why these changes?

ttoomey: Technically this saves some time time in CI since it runs setup but doesn't run the spec unless the env var is set

But I'll back it out; I was trying to fix a hang in CI that was actually caused by an infinite waiting for the job queue to be empty (fixed with `work_off_all_ready_jobs` helper

eanders (approved):

ttoomey:
Looks like still some lingering test failures; my plan is to merge deploy to QA before merging into the main release

ttoomey:
I had to iterate quite a bit to get specs to pass, a few things changed enough to warrant a quick re-review before merge.

spec/jobs/base_job_spec.rb
```diff
@@ -0,0 +1,103 @@
...
+    context 'when provider_job_id is present' do
+      let!(:dj_record) { Delayed::Job.create!(handler: 'dummy') }
+
+      before do
+        allow(job_instance).to receive(:provider_job_id).and_return(dj_record.id)
+      end
+
+      it 'returns the job record using provider_job_id' do
+        expect(job_instance.delayed_job).to eq(dj_record)
+      end
```
eanders, line 35: What is this testing?  It looks like we mock `provider_job_id` and tell it what to return, then check that the mock worked.  I'm probably missing something.

spec/support/delayed_job_helpers.rb
```diff
@@ -0,0 +1,39 @@
...
+    total_processed = 0
+    # Ensure any previous worker state is cleared from this thread
+    Thread.current[:delayed_job_worker] = nil
+
+    Delayed::Job.uncached do
+      worker = Delayed::Worker.new(queues: [])
+
+      until total_processed >= job_limit
+        # Handling potential clock skew in CI by forcing jobs to be ready
+        Delayed::Job.where(run_at: Time.current..).update_all(run_at: 1.minute.ago)
```
eanders: This seems like it could break tests that intentionally queue something in the future and expect it to not be run.  Not sure we have any of those, but if we're messing with time in the test, this could cause problems.

eanders (commented):
@ttoomey this looks good.  I dropped a few comments I think might need addressing before we merge.

spec/jobs/base_job_spec.rb
ttoomey: I think it's just to prove that the `delayed_job` method uses `provider_job_id` to find the record; I don't mind it I guess

spec/support/delayed_job_helpers.rb
ttoomey: Thanks, I removed it; it's still not totally clear to me what the issue was with this helper; I think the worker may have been trying to consume the wrong queue.

PR #6098, opened 2025-12-16
Resolve n+1s on unit group query and restructure tests

martha:
## _Merging this PR_

## Description
- Reduce n+1s on fields that are queried in batch from the frontend (see [GetProjectUnitGroups](https://github.com/greenriver/hmis-frontend/blob/61ebaaf27fdbf60823d78e19ca4a60807bc912e6/src/api/operations/units.queries.graphql#L22-L35))
- Restructure test coverage to mirror the fields that are actually being queried on the frontend. Separate `unit_groups_spec` into `project_unit_groups_spec` and `unit_group_spec`.

## Type of change
Performance improvement
Test update


drivers/hmis/app/graphql/types/hmis_schema/unit_group.rb
```diff
@@ -62,25 +62,44 @@ def units(**args)
 
     # Similar to `Project.unit_types`, this supports displaying availability by unit types. Don't resolve in batch.
     def unit_types
-      capacity = object.units.group(:unit_type_id).count
-      unoccupied = object.units.unoccupied_on.group(:unit_type_id).count
+      # Use preloaded units and active_unit_occupancies to avoid N+1 queries
```
martha, line 65: The diff in this file was proposed by AI to help reduce n+1s. I've reviewed the the line-by-line change and it makes sense to me, but zooming out, if/when we pull in the rest of the scope of #8157, will this field on UnitGroup even still exist?

On the other hand, I understand that is still an "if" and not necessarily a "when". So it does seem preferable to solve the n+1 issues, since this field is currently being queried in batch by the frontend. Let me know what you think.

martha: I've also certainly exceeded the 3h estimate by now, so it may not be worth pursuing this further.

gigxz: @martha this is feeling a bit risky in the amount of code changed, relative to the benefit. This field already acknowledged the n+1 ("Don't resolve in batch") so I'm wondering if it's really worth optimizing this? I think the assumption was that a project would only have a handful of unit groups, so it's an N+1 but the "n" is relatively small

gigxz: (and I'm not sure based on the diff if this logic had a proper regression test)

martha: Thanks @gigxz that makes sense to me, I'll close this PR without merging.

PR #6101, opened 2025-12-18
8635 HOPWA CAPER ATC sheet

ttoomey:
## _Merging this PR_

# Description
Adds Access to Care (ATC) reporting support to the FY2026 HOPWA CAPER.

- **Access to Care Sheet**: Implemented new reporting sheet to track housing stability outcomes, medical/income access, and supportive service intersections.
- **HOPWA CAPER Generator**: Added an enrichment step to the data pipeline to pull ATC outcomes from custom assessments using a historical lookback mechanism.
- **STRMU and Supportive Services Sheets**: Refactored household counting to be strictly service-based and deduplicated by HOH client ID to ensure consistency with expenditure reporting.
- **Site Configuration Admin**: Introduced database-backed application properties and an administrative UI to manage feature flags and dynamic field mappings for custom assessment data.
- **Reporting Framework**: Improved validation and error handling for question and column identifiers in the base question sheet builder.
- **Test Coverage**: Added specs for ATC outcome casting, service-based deduplication, and cross-sheet consistency.

### Type of Change
New feature


app/views/admin/app_config_properties/_action_menu.haml
```diff
@@ -0,0 +1,7 @@
+:ruby
+  items = [
+    { link_to: { path: edit_admin_app_config_property_path(property) }, icon: :pencil, label: 'Edit' },
+  ]
+  items << { link_to: { path: admin_app_config_property_path(property), method: :delete, data: { confirm: "Delete '#{property.key}'?" } }, icon: :cross, label: 'Delete'}
```
eanders: There's no conditional here, so maybe just include both in the array definition?
```suggestion
  items = [
    { link_to: { path: edit_admin_app_config_property_path(property) }, icon: :pencil, label: 'Edit' },
    { link_to: { path: admin_app_config_property_path(property), method: :delete, data: { confirm: "Delete '#{property.key}'?" } }, icon: :cross, label: 'Delete'},
  ]
```

app/views/admin/configs/index.haml
```diff
@@ -7,6 +7,7 @@
   .row
     .col-sm-3
       %nav.nav.flex-column
+        = link_to 'App Config Properties', admin_app_config_properties_path, class: 'nav-item nav-link'
```
eanders, line 10: This is probably fine, but I'd love to move all of these pages into the main menu now that we have nested menus.  No one can ever find Available File Tags or Public Files.

app/controllers/admin/app_config_properties_controller.rb
eanders, line 1: Should we add a simple feature doc for App Config Properties?

drivers/hopwa_caper/app/models/hopwa_caper/generators/fy2026/generator.rb
```diff
@@ -288,5 +294,91 @@ def custom_service_rows_for(enrollment_context, date_range:)
...
+      report.hopwa_caper_enrollments.in_batches(of: 100) do |batch|
+        # lookup map for the most recently submitted custom assessment
+        hud_enrollment_scope = Hmis::Hud::Enrollment.where(id: batch.map(&:enrollment_id))
+        assessments_by_enrollment = Hmis::Hud::CustomAssessment.
+          joins(:enrollment).merge(hud_enrollment_scope).
+          where(AssessmentDate: ..@report.end_date).
+          order(AssessmentDate: :desc, id: :desc).
+          preload(:enrollment).group_by { |r| r.enrollment.id }
+
+        batch.each do |spm_enrollment|
```
eanders: Why are we calling these `spm_enrollment` should it be `hopwa_caper_enrollment`?

eanders: or just `report_enrollment`

eanders (commented):
@ttoomey let me know if you want to discuss any of my questions.  I think the app config change looks like a significant improvement, but had some concern about variable names.

eanders (approved):

ttoomey:
changes since last review:

Distinguishes explicit "No Income/Insurance" responses from missing data and tightens household counting logic for STRMU service-based eligibility.

- **Admin Config UI**: Strip whitespace from keys and values on save; display compact timestamps in config/compliance/content page lists, add a better datetime format for tabular data
- **HOPWA CAPER Enrollment Staging**: Track income and insurance types with explicit sentinel markers (`NoIncomeSource`, `NoInsuranceSource`) to separate households that answered "No" from those with no assessment data
- **Income/Insurance Filters**: Add filter IDs for lookups/convenience accessor 
- **Access to Care Custom Data Elements**: improve custom assessment boolean values nil-handling to distinguish blank responses from Yes/No answers; skip nil values when searching for most recent valid response
- **Access to Care Sheet**: Deduplicate household counts within activity types; count heads of household consistently across all rows; use filter helpers for income/insurance queries
- **STRMU Sheet**: consistently apply requirement for financial assistance services received during reporting period rather than enrollment dates alone


### Type of Change
Bug fix

app/views/admin/compliance_requirements/_list.haml
```diff
@@ -19,5 +20,6 @@
           - else
             %span.text-muted Never
         %td= req.active? ? 'Active' : 'Inactive'
+        %td= l req.updated_at, format: :table_compact
```
ttoomey, line 23: small opportunistic change for consistency in recent admin UI

app/views/admin/content_pages/_list.haml
```diff
@@ -18,5 +18,5 @@
             = page.compliance_requirements.pluck(:name).join(', ')
           - else
             %span.text-muted None
-        %td= page.updated_at.to_fs(:short)
+        %td= l page.updated_at, format: :table_compact
```
ttoomey, line 21: small opportunistic change for consistency in recent admin UI

config/locales/en.yml
```diff
@@ -83,6 +83,9 @@ en:
     index:
       title: 'Projects'
       search_prompt: 'Search projects...'
+  time:
+    formats:
+      table_compact: "%m/%d/%y %I:%M%P"
```
ttoomey, line 88: this is more compact and has consistent width which seems better for tabular views

app/views/admin/compliance_requirements/_list.haml
eanders: This is probably ok to leave, though while `MM/DD/YYYY` is the US standard format for dates, historically this application has used `Jan 20, 2025`.  It's something we should probably revisit globally so we maintain consistency across the application for date and time formats.

config/locales/en.yml
eanders: See: https://github.com/greenriver/hmis-warehouse/pull/6101/changes#r2649157535

eanders (approved):

app/views/admin/compliance_requirements/_list.haml
ttoomey: Yes, it is a change. I suggest that we standardize on this format for dates & times in a tabular context as it optimizes for scannability, has a consistent width, and takes up less horizontal real-estate

app/models/hud_reports/question_sheet.rb
```diff
@@ -13,7 +13,7 @@ class QuestionSheet
 
     def initialize(report:, question:)
       @report = report
-      raise "invalid question id #{question.inspect}" unless question =~ /\AQ\d+/
+      raise ArgumentError, "invalid question id #{question.inspect}" unless question =~ /\AQ\d+/
```
ttoomey, line 16: added type to support unit tests

spec/models/hud_reports/question_sheet_spec.rb
```diff
@@ -0,0 +1,147 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe HudReports::QuestionSheet do
```
ttoomey, line 5: opportunistically adding a unit test for this core utility class

drivers/hopwa_caper/spec/models/generators/fy2026/generator_spec.rb
```diff
@@ -0,0 +1,223 @@
...
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+require 'rails_helper'
+require_relative 'hopwa_caper_shared_context'
+
+RSpec.describe HopwaCaper::Generators::Fy2026::Generator, type: :model do
```
ttoomey, line 12: add missing unit test for generator

ttoomey:
Changes since last review:
- **Base Reporting Sheet**: Adds a `services_with_hoh` helper to join services with Head of Household enrollments and refactors scoping to support client-based unduplication.
- **Access to Care Sheet and Supportive Services Sheet**: Shifts household counts from household IDs to distinct Head of Household client IDs.
- **STRMU Sheet**: Refactors service categorization (exclusive vs. multi-type) to group by client ID, ensuring clients with multiple enrollments are correctly deduplicated.
- **Test Coverage**: Adds validation for clients with multiple enrollments and varied service types to ensure accurate categorization and deduplication.
- **App config properties**: handles json values

app/models/hud_reports/question_sheet.rb
eanders: Aren't SPM questions actually things like `Measure 1`?

eanders: Oh, I see, this isn't a change to the check, just what is raised.  Looks ok.

eanders (approved):
Thanks @ttoomey !

app/models/hud_reports/report_client_base.rb
```diff
@@ -72,7 +72,7 @@ def fetch_cell_value(col)
     def format_complex_value(col, value, pii_policy, include_content_tag)
       if value.is_a?(Array)
         # For Arrays, calculate each array element's value using the column name for the array
-        value.map { |item| display_value(col, pii_policy: pii_policy, include_content_tag: include_content_tag, cell_val: item, calculate_cell: false) }
+        value = value.map { |item| display_value(col, pii_policy: pii_policy, include_content_tag: include_content_tag, cell_val: item, calculate_cell: false) }
```
ttoomey, line 75: This appears to be a broader bug, causes values containing arrays to skip display value

Note, the report client base seems like kind of a lot so I opted to build a presenter for the hopwa caper instead

drivers/hopwa_caper/app/models/hopwa_caper/drilldown_presenter.rb
```diff
@@ -0,0 +1,174 @@
+# frozen_string_literal: true
+
+module HopwaCaper
+  class DrilldownPresenter
```
ttoomey, line 4: extracts the drill-down logic from the hopwa caper enrollment; should be more maintainable

drivers/hopwa_caper/app/controllers/hopwa_caper/cells_controller.rb
```diff
@@ -18,10 +18,6 @@ def show
         preload(enrollment: :project).
         joins(hud_reports_universe_members: { report_cell: :report_instance }).
         merge(::HudReports::ReportCell.for_table(@table).for_cell(@cell))
-      @services = @report.hopwa_caper_services.
```
ttoomey, line 21: we aren't actually using services in a drilldown. Removed to simplify

ttoomey:
@eanders 
sorry for the additional review. Changes:
* fixed an issue in the base report client display value
* implement a presenter for the hopwa caper rather than the shared implementation which seems overly complex
* added support for "services summary" column to support data visibility into service records
* removed support for services from hopwa caper drilldowns (only enrollment are used)

app/models/hud_reports/report_client_base.rb
eanders: Agreed, this looks like a bug since it doesn't return or save the results.

drivers/hopwa_caper/app/models/hopwa_caper/drilldown_presenter.rb
eanders: This is probably fine, but it feels like much of this is generic, so in the future we should probably have a base DrilldownPresenter and then inherit from it and just override `enrollment_fields` and maybe have some mechanism to determine if a custom `display_value` is needed (like for `services_summary`).

eanders (approved):

drivers/hopwa_caper/app/models/hopwa_caper/drilldown_presenter.rb
ttoomey: Definitely, maybe the next time we need to mess with drilldowns on another report we can extract the common pieces to a base class

PR #6102, opened 2025-12-18
Follow up on Unit Group improvements

martha:
## _Merging this PR_
- should target release-194 when ready

## Description
- Remove the deprecated unit type argument on the Unit input. For backwards compatibility, this is the part that should target release-194
- Require unit type on the `CreateUnitGroup` mutation. 
  - Originally I thought this should be required on the input object, and therefore also needed to go into 194. However, matching our pattern from other input objects used by both a create and update mutation, (ServiceTypeInput, ProjectConfigInput, etc.), I've left it unrequired in the input and added validation errors in the create mutation if it's missing. So this change could have gone into 193 after all.

## Type of change
Code clean-up


drivers/hmis/app/graphql/types/hmis_schema/unit_group_input.rb
```diff
@@ -15,6 +15,6 @@ class HmisSchema::UnitGroupInput < BaseInputObject
     argument :ce_event_type, HmisSchema::Enums::Hud::EventType, required: false
     # argument :count, Integer, 'Number of units to create', required: false
     # argument :prefix, String, 'Prefix for unit names', required: false
-    argument :unit_type_id, ID, required: false # TODO(#8157) - require in next release for backwards compatibility
+    argument :unit_type_id, ID, required: false
```
gigxz, line 18: should this change to `required: true`?

gigxz (approved):

drivers/hmis/app/graphql/types/hmis_schema/unit_group_input.rb
martha: here I was matching our pattern from other input objects used by both a create and update mutation, (ServiceTypeInput, ProjectConfigInput, etc.), I've left it unrequired in the input and added validation errors in the create mutation if it's missing. Let me know if that doesn't sound right to you.

I think the alternative is to require it and make a frontend change so the frontend always passes it (which it doesn't currently), but the update mutation would still ignore it (or maybe raise if the user tried to change it?)

gigxz: Yes thats right, I forgot about that quirk. Thanks!

PR #6106, opened 2025-12-18
Expand usage of form definition policy

gigxz:
## _Merging this PR_

## Description
https://github.com/open-path/Green-River/issues/8375

Use existing Form Definition Policy in all applicable mutations. See comments for areas for discussion

## Type of change
Refactor


drivers/hmis/app/graphql/mutations/create_form_definition.rb
```diff
@@ -34,6 +30,8 @@ def resolve(input:)
         **attrs,
       )
 
+      access_denied! unless policy_for(definition, policy_type: :form_definition).can_create?
```
gigxz: Approaches considered for checking "create" capability:
1. pass unpersisted resource to form policy
2. pass nil resource to form policy (` policy_for(nil, policy_type: :form_definition).can_create?`)
3. move check to parent record (e.g. `policy_for(enrollment).can_create_note?`) (not applicable to form definition imo, no parent record)

I went with (1) for this case. I don't think that is necessarily the best approach in all cases (or here), for example if initializing the record would be an expensive operation.

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
```diff
@@ -7,28 +7,64 @@
...
+  # can_manage_forms permission grants access to edit certain form roles (SERVICE, CUSTOM_ASSESSMENT),
+  # while "super-admin" permission can_administrate_config grants access to edit all form roles.
+  def can_manage_form_by_role?
+    global_permissions.include?(:can_manage_forms) && manageable_form_role?
+  end
+
+  # Determines if the form definition is a non-super-admin form or a super-admin form.
+  def manageable_form_role?
+    form_definition.role.to_s.in?(Hmis::Form::Definition::NON_ADMIN_FORM_ROLES) || global_permissions.include?(:can_administrate_config)
+  end
```
gigxz: Come up with better naming? This is a rewrite of the logic that was previously in the shared UserPermissions concern.

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
```diff
@@ -7,28 +7,69 @@
...
+  def can_create_draft? = can_manage_form?
+
+  # Whether the user can edit a draft version of the form definition
+  def can_edit_draft? = can_manage_form?
+
+  # Whether the user can publish the form definition
+  def can_publish? = can_manage_form?
+
+  # Whether the user can duplicate the form definition
+  def can_duplicate?
```
gigxz, line 36: Is this going overboard? I think it's helpful to list out the various operations that can be performed on this record type.

drivers/hmis/app/graphql/mutations/create_form_definition.rb
martha: Another suggestion could be to pass the class,
`policy_for(Hmis::Form::Definition, policy_type: :form_definition).can_create?`

I was going to ask for feedback on establishing that pattern on the Client policy for global permissions like `can_merge_clients`. I'll link my draft although it's not ready for review yet, https://github.com/greenriver/hmis-warehouse/pull/6110/changes#diff-be251da99077ff6e7a11f0ef9e270f054abd8a339e1795cb082ce736e67bfef0R48-R56

drivers/hmis/app/graphql/mutations/delete_form_definition.rb
```diff
@@ -15,11 +15,10 @@ class DeleteFormDefinition < CleanBaseMutation
     def resolve(id:)
       definition = Hmis::Form::Definition.find_by(id: id)
       raise 'not found' unless definition
-
-      access_denied! unless current_user.can_manage_forms_for_role?(definition.role)
-
       raise 'can only delete draft forms' unless definition.draft?
```
martha, line 18: Just curious, what was the reason to move the access check below the draft check?

Also, it looks like we could potentially remove the draft check, since the policy now checks that.

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
martha: Maybe `non_admin_form_role` is clearer, and matches the existing `Hmis::Form::Definition::NON_ADMIN_FORM_ROLES`

martha (approved):

drivers/hmis/app/graphql/mutations/delete_form_definition.rb
gigxz: The spec tests specifically asserted on the draft message, so I moved it to keep those test assertions specific (asserting on draft issue vs general access failure).

drivers/hmis/app/graphql/mutations/create_form_definition.rb
ttoomey: This always a little fiddly. FYIW we are using the pass-a-class pattern in the policies already, for example:
```ruby
policy_for(Hmis::Ce::Referral, policy_type: :ce_referral).can_index?
```

drivers/hmis/app/models/hmis/auth_policies/form_definition_policy.rb
ttoomey: either seems fine; this is an internal protected method so the naming isn't super critical as long as it's well documented

ttoomey (approved):
Great work, thank you!

drivers/hmis/app/graphql/mutations/create_form_definition.rb
gigxz: Thank you both! Updated to use the pass-a-class pattern.

gigxz:
Test failure is unrelated; merging

PR #6107, opened 2025-12-18
Add cleanup script for client primary name

martha:
## _Merging this PR_
- ~~I'm proposing to target release-193 (rather than 194) since it's the same release branch as the functional bugfix (https://github.com/greenriver/hmis-warehouse/pull/6081). I ought to have included this cleanup script in the original PR.~~

## Description
- Add one-time rake cleanup task to find and fix clients who have associated name record(s), but no name is marked primary

## Type of change
One-time task


gigxz (approved):

martha:
thanks @gigxz for the improvements! I want to check before merging, do you agree with my earlier proposal to merge this to 193? Now that more time has passed, and 193 is already on production, I'm hesitating. Maybe we should retarget to 195 and deploy this on the normal deployment cycle. Edit: or split the difference and target to 194.

gigxz:
Yes I think we should retarget this to 195 and deploy as usual, it doesn't seem time-sensitive. thanks!

PR #6109, opened 2025-12-19
8677 attribute defaults

ttoomey:
## _Merging this PR_

### Description
Fix static defaults and accidental mutation of shared collection state in report filter attributes.

- Converted static date and class-method defaults to dynamic procs to ensure evaluation at instantiation time rather than class load time.
- Froze array literals used as attribute defaults across all filter classes to prevent accidental mutation of shared state between instances.
- **Cohort Column Definitions**: Wrapped localized titles in procs to ensure `Translation.translate` is called at instantiation time.

Also:
- skip cloud watch initializer in de/test; it seems to be causing performance issues for me locally

### Type of Change
Bug fix


app/models/filters/filter_base.rb
```diff
@@ -25,48 +25,48 @@ class FilterBase < ::ModelForm
...
-    attribute :default_start, Date, default: (Date.current - 1.year).beginning_of_year
-    attribute :default_end, Date, default: (Date.current - 1.year).end_of_year
+    attribute :veteran_statuses, Array, default: [].freeze
+    attribute :age_ranges, Array, default: [].freeze
+    attribute :genders, Array, default: [].freeze
+    attribute :races, Array, default: [].freeze
+    attribute :length_of_times, Array, default: [].freeze
+    attribute :destination_ids, Array, default: [].freeze
+    attribute :prior_living_situation_ids, Array, default: [].freeze
+    attribute :default_on, Date, default: ->(_, _) { Date.current }
```
ttoomey, line 38: this appears to be source of occasional test failures in spec/models/filters/filter_base_spec.rb

config/initializers/aws_cloudwatch.rb
```diff
@@ -1,6 +1,6 @@
 # Rails.logger.debug "Running initializer in #{__FILE__}"
 
-if ENV['EKS'] != 'true'
+if ENV['EKS'] != 'true' && (Rails.env.staging? || Rails.env.production?)
```
ttoomey: unrelated fix

ttoomey: I'm finding this elsewhere in the code I'll pull this out and fix this properly in another branch

eanders (approved):
I'm not up on why these are an issue, or if they are only an issue in staging, but I don't see anything wrong with the changes.  Thanks @ttoomey !

ttoomey:
> I'm not up on why these are an issue, or if they are only an issue in staging, but I don't see anything wrong with the changes. 

Basically the code misunderstands the context; variables assigned in a class context will leak across requests or jobs in unpredictable ways. If they values are truly static, such as an empty array, they should be frozen. Otherwise they should __always__ be evaluated at execution time. This type of thing can also lead to weird class loader issues too in development. Probably good to keep our eyes open for other cases where this might occur in the codebase as it's quite difficult to detect from behavior.  

Practically speaking this manifested in test failures that exercised the default values on filters (filter_base) because the default was assigned at load time and not at test run time.

However this could easily result in other bugs:
* default filter values for dates or translation are statically assigned when the server boots, not at execution time; so a user might get different filter behavior depending on when the server's container booted
* non-frozen defaults could easily be accidentally mutated by callers, leading to buggy and unexpected behavior over the lifetime of the process

PR #6112, opened 2025-12-19
Add Job Priority Constants

dtgreiner:
## _Merging this PR_

## Description

Replace Job Priority numbers with constants to better describe/define the priorities. 

Additionally, all jobs in the default queue were placed instead in the long-running queue.

## Type of change

Code clean-up


app/jobs/base_job.rb
```diff
@@ -10,6 +10,15 @@ class BaseJob < ApplicationJob
   include NotifierConfig
   include MaintenanceTaskInstrumentation
 
+  # Priority constants for job scheduling
+  # Lower numbers = higher priority (processed first)
```
eanders, line 14: Let's revisit these names, but I like the idea.

app/jobs/service_history/rebuild_enrollments_by_batch_job.rb
```diff
@@ -27,7 +27,8 @@ def perform
       end
     end
 
-    def enqueue(job, queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running))
+    def enqueue(job, _queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running))
+      job.priority = BaseJob::NO_RUSH_PRIORITY - 1
```
ttoomey: suggest removing the `queue` param entirely rather than renaming, assuming it is unused

app/models/grda_warehouse/hud/client.rb
```diff
@@ -2247,7 +2247,7 @@ def split(client_ids, hmis_receiver_id, health_receiver_id, current_user)
       end
 
       Rails.logger.info "Queueing cleanup for clients: #{to_clean.inspect}"
-      ClientCleanupJob.set(priority: 6).perform_later(to_clean.uniq)
+      ClientCleanupJob.set(priority: BaseJob::BACKGROUND_JOB_PRIORITY + 1).perform_later(to_clean.uniq)
```
ttoomey: Suggest we commit entirely to using names for priority and entirely eliminate offsets like this. Instead, define meaningful constants for every level

`CLIENT_CLEANUP_JOB_PRIORITY = 6`

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
```diff
@@ -26,9 +26,13 @@ class ExpireBaseJob < BaseJob
...
     queue_as ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running)
 
-    # low priority
+    # low priority - eventual consistency cleanup
     def self.default_priority
-      10
+      BaseJob::EVENTUAL_CONSISTENCY_PRIORITY
+    end
+
+    def enqueue(job, _queue: ENV.fetch('DJ_LONG_QUEUE_NAME', :long_running))
```
ttoomey: suggest removing `_queue` if unused

ttoomey (approved):
Looks good, thanks!
* suggest removing `_queue` param provided it is unused
* suggest replacing relative offsets  such as `NO_RUSH_PRIORITY - 1` with descriptive constants

app/models/grda_warehouse/hud/client.rb
eanders: This was my suggestion (the delta from a particular constant) as sometimes we need to ensure something runs at a higher or lower priority than something else, but should be roughly in the same band.  It may _only_ apply to the importer and enrollment SHE generation, not sure.  Maybe the name should be something like SLIGHTLY_HIGHER_THAN....

app/jobs/base_job.rb
ttoomey: How about the following (AI suggested). Maybe we could consider removing some of this granularity to make things simpler.

```ruby
  # Priority constants for job scheduling
  # Lower numbers = higher priority (processed first)
  
  # 1. User Interaction Tier
  UI_IMMEDIATE_PRIORITY        = -5   # User is actively waiting for results in the UI
  
  # 2. Critical System Tier
  HIGH_IMPORTANCE_PRIORITY     = 0    # Critical tasks needed to maintain system integrity
  
  # 3. Standard Operations Tier
  DEFAULT_BACKGROUND_PRIORITY  = 5    # Standard async tasks (default choice for most work)
  ROUTINE_CLEANUP_PRIORITY     = 6    # Non-urgent cleanup following standard operations
  
  # 4. Batch & Bulk Tier
  BULK_PROCESSING_PRIORITY     = 9    # High-priority batch work that should lead the bulk queue
  LARGE_BATCH_PRIORITY         = 10   # Standard bulk processing and large data exports
  
  # 5. Consistency & Cache Tier
  CACHE_REFRESH_PRIORITY       = 12   # Standard warming or rebuilding of data caches
  LOW_PRIORITY_CACHE_PRIORITY  = 13   # Non-urgent cache updates (e.g. external ID lookups)
  
  # 6. Maintenance Tier
  MAINTENANCE_PRIORITY         = 15   # Background housekeeping and eventual consistency
```

eanders: That's probably fine, and maybe it doesn't matter a ton, but if I ran into `BULK_PROCESSING_PRIORITY` and `LARGE_BATCH_PRIORITY` in the codebase I'd still need to lookup which was higher priority.  (Same for some of the others that are close together.  Maybe we should add the actual number to the end? `BULK_PROCESSING_PRIORITY_9` and `LARGE_BATCH_PRIORITY_10`.  That way we'd know the intention _and_ the actual outcome when using/finding it.

app/models/grda_warehouse/hud/client.rb
ttoomey: Fair. It reminds me of how people often deal with z-index in CSS (modal-z-index + 3) which quickly becomes hard to understand. But probably not worth worrying about.

I wonder if this just points to a bit of a gap around how we manage and schedule jobs; perhaps we are trying to do too much with delayed job. For example airflow might be better suited for ETL and nightly processing; sidekiq for UI updates and email; while DJ seems appropriate for reports. That would then greatly simplify the priorities

eanders: I would love to see us get more sophisticated.  I agree completely that Airflow would be a much better orchestrator for imports, maybe the nightly processes as well.  
And having something like sidekiq might be better for short running tasks, or whatever the standard rails 8 mechanism is.

drivers/hmis_csv_importer/app/jobs/hmis_csv_importer/cleanup/expire_base_job.rb
dtgreiner: this was removed in https://github.com/greenriver/hmis-warehouse/pull/6112/commits/067b137f0d495ae267c4a3ec9a47ea9d021ea844

app/jobs/service_history/rebuild_enrollments_by_batch_job.rb
dtgreiner: this was removed in https://github.com/greenriver/hmis-warehouse/pull/6112/commits/067b137f0d495ae267c4a3ec9a47ea9d021ea844

eanders (approved):
Thanks @dtgreiner I think this has a nice balance between being verbose and specific with the numbers.

PR #6115, opened 2025-12-22
8509 Household Context report optimization for 2026 SPM and APR

ttoomey:
## _Merging this PR_

### Description
Implements a pre-computed household context layer for FY2026 HUD reports to centralize business logic and improve generation performance.

### Functional Areas
- **Household Context Layer**: Introduced `HudReports::HouseholdContext` model and database table to cache pre-calculated household attributes (chronic status, move-in dates, household composition) per report run.
- **Business Logic Centralization**: Extracted core household rules (chronic inheritance, household typing, move-in date logic) into `HudReports::HouseholdLogic`, removing duplication between runtime concerns and the new pre-computation layer.
- **FY2026 Report Generators**: Refactored APR, CAPER, CE APR, and DQ generators to populate household contexts during the `prepare_report` phase and utilize `AprClientBuilder` for efficient snapshot creation.
- **SPM Optimization**: Configured SPM HDX upload to generate household contexts once and share them with subordinate Data Quality reports via the `source_report_id` mechanism.
- **Legacy Compatibility**: Updated `HudReports::Households` concern to delegate calculation logic to the new `HouseholdLogic` class, ensuring consistency across older report versions.
- **Maintenance**: Added a daily maintenance task to prune expired household context records.

### Type of Change
Refactor

**TBD:**
- Allow sharing households between SPM and DQ reports to avoid redundant builds.
- Periodically prune stale `HudReports::HouseholdContext` records.
- Expand builder pattern to other reports currently relying on the `HudReports::Households` module.

### Type of Change
Refactor

Review complete: 14 files analyzed (3 high, 6 medium, 5 low risk)
Findings: 0 critical, 4 warning, 3 suggestion


ttoomey:
design artifact

# Pre-computed Household Context for HUD Reports

This design shifts complex household-level calculations—chronic status inheritance, move-in date derivation, and population categorization—from runtime Ruby calculations into a pre-computed database layer.

## Design Goals

**Performance**: Compute complex household attributes once per report run rather than repeatedly during snapshot generation.

**Consistency**: Centralize business rules in a single source of truth, ensuring identical behavior across all report types and fiscal years.

**Maintainability**: Replace scattered runtime logic with a clear separation between computation (preparation phase) and consumption (snapshot generation phase).

## Architecture

### Core Components

- **`HudReports::HouseholdContext`**: Database table storing pre-computed household attributes per enrollment. Acts as a logic-specific cache layer between raw warehouse data and report snapshots.

- **`HudReports::HouseholdLogic`**: Single source of truth for all household business rules. Defines chronic status inheritance, move-in date propagation, household typing, and parenting youth identification.

- **`HudReports::HouseholdContextBuilder`**: Orchestrates the pre-computation process. Batches enrollments by household, applies `HouseholdLogic`, and bulk-imports results.

- **`AprClientBuilder`**: Maps pre-computed context fields and enrollment data into final report snapshot attributes.

### Two-Phase Execution

**Preparation Phase**: `HouseholdContextBuilder` analyzes all in-scope enrollments, applies business rules, and persists results to the context table.

**Snapshot Phase**: Report generators load contexts in batches and construct snapshot records using pre-computed values, eliminating expensive joins and repeated calculations.

## Data Model

The `hud_report_household_contexts` table caches:

- **Inherited attributes**: Chronic status, move-in date, household type
- **Head of Household snapshots**: Entry date, move-in date, CoC code, age, veteran status
- **Household aggregates**: Member count, age ranges, veteran flags, parenting youth indicators

This eliminates the need to traverse household relationships during snapshot generation.

## Cross-Report Optimization

For multi-report processes (e.g., SPM HDX Upload generating multiple DQ sub-reports):

1. **Shared Build**: Parent report computes contexts for all projects in scope
2. **Efficient Sharing**: Child reports copy only the subset of contexts they need via direct database operations
3. **Validation**: System enforces matching date ranges between source and target reports

This reduces redundant computation by 60-80% for complex report workflows.

## Backward Compatibility

Pre-FY2026 reports continue using runtime calculations via the `HudReports::Households` concern. This concern was refactored to delegate to `HouseholdLogic`, ensuring older fiscal years use identical business rules as the new optimized path.

FY2026+ generators bypass the concern entirely, reading from the pre-computed context table instead.

## Verification

If report values appear incorrect, inspect the `hud_report_household_contexts` table for the enrollment in question. The cached values should reflect the applied business rules. Test coverage exists at both the logic layer (`HouseholdLogic`) and orchestration layer (`HouseholdContextBuilder`).

app/jobs/importing/run_daily_imports_job.rb
```diff
@@ -177,6 +177,11 @@ def _perform
         end
       end
 
+      run_maintenance_task('Prune HUD report data') do
+        HudReports::HouseholdContext.prune!
```
eanders, line 181: Haven't looked yet, but want to confirm this doesn't break any of the view logic for a report.  Assuming this is ephemeral and can safely be deleted, computation results have been stored on the relevant report client/enrollment records.

ttoomey: Yes, the household contexts exist only to populate the apr client (or spm enrollment records soon). I tried to avoid them leaking out into the question layer; questions should be unchanged although it's possible something could leak through with the household module refactor; I'll do a re-review to check

app/models/hud_reports/household_context.rb
```diff
@@ -0,0 +1,53 @@
...
+  # - Scoped: Always tied to a specific `ReportInstance`
+  class HouseholdContext < GrdaWarehouseBase
+    self.table_name = 'hud_report_household_contexts'
+
+    belongs_to :report_instance, class_name: 'HudReports::ReportInstance'
+    belongs_to :service_history_enrollment, class_name: 'GrdaWarehouse::ServiceHistoryEnrollment'
+
+    def self.prune!
+      # Delete contexts for reports older than 2 weeks or reports that have been deleted
+      valid_report_ids = HudReports::ReportInstance.where('created_at >= ?', 2.weeks.ago).select(:id)
```
eanders: Maybe we want to keep letting simple queries be written as strings, but I'd like to try and keep things in standard active record syntax where possible.  I think this would generate the same query:

```suggestion
      valid_report_ids = HudReports::ReportInstance.where(created_at: ..2.weeks.ago).select(:id)
```

app/models/hud_reports/household_context_builder.rb
```diff
@@ -0,0 +1,281 @@
...
+
+    def import_batch_size
+      2000
+    end
+
+    def get_hh_id(she)
+      she.household_id || "#{she.enrollment_group_id}*HH"
+    end
+
+    def each_household_id_batch
```
eanders: This is probably fine, as it may cast a wider than intended net, but it is probably worth noting that `HouseholdID` is only guaranteed  unique within a data source, so there's a chance you'll load and generate household information that is never used.  Again, probably safe, but might be worth a note.

eanders: Or, maybe that's not true since you're starting from the SHE scoped to the included clients.

app/models/hud_reports/household_context_builder.rb
```diff
@@ -0,0 +1,281 @@
...
+
+        # Determine report_date for chronic calculation
+        report_date = @generator.filter&.on if @generator.respond_to?(:filter) && @generator.filter.present?
+
+        chronic_at_start = enrollment&.chronically_homeless_at_start
+        if report_date.blank? || report_date == enrollment&.EntryDate
+          pit_chronic_at_start = chronic_at_start
+        else
+          pit_chronic_at_start = enrollment&.chronically_homeless_at_start(date: report_date)
+        end
```
eanders: Is this because it is less expensive/we already need to use `chronically_homeless_at_start`?

```suggestion
        # If we happen to be looking for chronicity on the entry date, we don't need to recalculate it, save some processing and re-use this.
        if report_date.blank? || report_date == enrollment&.EntryDate
          pit_chronic_at_start = chronic_at_start
        else
          pit_chronic_at_start = enrollment&.chronically_homeless_at_start(date: report_date)
        end
```

app/models/hud_reports/household_context_builder.rb
```diff
@@ -0,0 +1,281 @@
...
+      else
+        scope.where(enrollment_group_id: synthetic_eg_ids, household_id: nil)
+      end
+
+      query.to_a.select { |m| m.enrollment.present? }
+    end
+
+    def build_contexts_for_household(hh_id, members)
+      # Pre-calculate common HH values
+      hh_member_hashes = members.map do |m|
```
eanders: `members` are SHE, we might want to rename that.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/apr_client_builder.rb
```diff
@@ -0,0 +1,466 @@
...
+      # Step 7. of the CE APR, but really all APR related should work this way.
+      # Filter the assessments/events, keeping only those where the CoC code assigned in step 6
+      # matches the CoC on which the report is being run.
+      @enrollment.enrollment_coc.in?(@report.coc_codes)
+    end
+
+    # rubocop:disable Metrics/AbcSize, Metrics/CyclomaticComplexity, Metrics/PerceivedComplexity
+    def map_standard_attributes
+      exit_date = @last_service_history_enrollment.last_date_in_program
+      exit_record = @last_service_history_enrollment.enrollment if exit_date.present? && exit_date <= @report.end_date
```
eanders: This doesn't look right to me.  It's calling an Enrollment `exit_record`.  Maybe `exited_enrollment` would be better.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/apr_client_builder.rb
```diff
@@ -0,0 +1,466 @@
...
+    include HudReports::LengthOfStays
+    include HudReports::Incomes
+
+    attr_reader :last_service_history_enrollment, :ctx, :source_client
+
+    private def a_t
+      @a_t ||= HudApr::Fy2020::AprClient.arel_table
+    end
+
+    def initialize(report:, client:, enrollments:, context_map:, hoh_enrollment_map:, needs_ce_assessments:, households:)
```
eanders, line 32: Can you add some comments about what each of these "map" things are?  I noticed later it calls `@ctx.age`, but it is unclear to me if that's the correct age calculation since I'm not sure what `@ctx` holds.

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/base.rb
```diff
@@ -65,369 +65,67 @@ def needs_ce_assessments?
...
+            context_map: contexts_by_she_id,
+            hoh_enrollment_map: hoh_enrollments_by_id,
+            needs_ce_assessments: needs_ce_assessments?,
+            households: batch_households,
+          )
+
+          next unless result[:success]
+
+          # Handle duplication check here (Orchestration concern)
+          if processed_source_clients.include?(result[:source_client_id])
```
eanders, line 125: What might cause this?

ttoomey:
Deployed to QA for testing with larger data set in EKS env. SPM generated in about 2 hours for ~9,000 client records with 141,000 enrollments. Completed without running out of memory

https://qa-warehouse.openpath.host/hud_reports/spms/==3UjLXGKfeiqAyRYGclzLoQ==

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/base.rb
ttoomey: Maybe too defensive, not expected during normal operation.  But could be:
* maybe due to splitting/merging destination client while report is active?
* we have a join product in client scope

Note that this is a carry-over from the legacy implementation, for example the 2023 APR generator has the same guard (drivers/hud_apr/app/models/hud_apr/generators/shared/fy2023/base.rb:142)

app/models/hud_reports/household_context_builder.rb
ttoomey: yes, it's a performance optimization, I'll add a comment

ttoomey:
this PR has been split into phases for easier review:

* https://github.com/greenriver/hmis-warehouse/pull/6303
* https://github.com/greenriver/hmis-warehouse/pull/6304
* https://github.com/greenriver/hmis-warehouse/pull/6305 (you are here)

PR #6126, opened 2025-12-30
Add HmisClientPolicy and use it in mutations

gigxz:
## _Merging this PR_

## Description
Add HMIS Client policy, and use it in mutations ~and for object-level authorization in GraphQL.~

## Type of change
New feature


drivers/hmis/app/models/hmis/auth_policies/hmis_client_policy.rb
```diff
@@ -0,0 +1,60 @@
...
+    raise ArgumentError, 'Must provide a client instance' unless resource.is_a?(Hmis::Hud::Client)
+
+    context.client_permissions(resource.id)
+  end
+
+  # Get global permissions for the user (across all projects they can access)
+  # These can be accessed when the policy is instantiated with a class
+  def global_permissions
+    context.global_permissions
+  end
```
gigxz: For discussion. The previously proposed implementation was:
```ruby
  memoize def client_permissions
    if resource.is_a?(Class)
      # Class-level policy (e.g., for merge operations)
      context.global_permissions
    else
      # Instance-level policy (for a specific client)
      context.client_permissions(resource.id)
    end
  end
```
I added this safety check because it seemed like it would be too easy to accidentally pass a class method when you really meant to pass an instance. This approach makes it so particular methods implicitly require an instance to be passed to the policy. This felt safer.

--

There are lots of other ways we could approach this, but they felt pretty clunky. For example an explicit check on the methods themselves:
```ruby
 def can_view_name?
    validate_instance_resource! # expected to be called on an instance-level policy
    client_permissions.include?(:can_view_client_name)
  end

 def can_view_any_names?
    validate_class_resource! # expected to be called on a class-level policy
    client_permissions.include?(:can_view_client_name)
  end
```

gigxz: Tagging @ttoomey your thoughts are welcome, I'll request review formally once tests are confirmed to pass.

ttoomey: Thanks, I think the tighter type checking is a great add. The way it's implemented here is low-impact.

But I wonder if we have a common need for global permission checks if it's worth building something a bit heavier. I had hopped we could avoid such global permissions by, say authorizing against a data-source or some other entity. Seems like we aren't there yet, maybe some day.

Perhaps this is over-engineering but here's what I'd suggest: https://github.com/greenriver/hmis-warehouse/pull/6131 (I asked AI to draft this up, it's fine to discard it if seems like the wrong direction). The idea is to avoid having to deal with type uncertainty around `policy.resource` at the cost of adding a small `factory` that is encapsulated in `user.policy_for()`

ttoomey (commented):
This is looking great, thanks! I had some ideas around the global policy question, see attached comment and PR

drivers/hmis/app/models/hmis/auth_policies/hmis_client_policy.rb
gigxz: Finally opened https://github.com/greenriver/hmis-warehouse/pull/6173, resolving this thread. I updated this branch to target it

drivers/hmis/app/graphql/mutations/delete_client.rb
```diff
@@ -20,28 +20,25 @@ def self.problem_enrollments_message(count)
 
     def resolve(id:, client_lock_version: nil, confirmed: false)
       client = Hmis::Hud::Client.viewable_by(current_user).find_by(id: id)
-      client.lock_version = client_lock_version if client_lock_version
-
-      # While this is redundant with the viewable_by() scope above, this check caches the authorization result so that
```
martha, line 25: @martha - might need to add this back for now

martha: yes, without this the DeleteClient mutation fails. I'm re-adding it and will remove it in the next PR. (draft link: https://github.com/greenriver/hmis-warehouse/pull/6193)

martha (approved):

martha:
@gigxz I extracted out the change to the client schema `authorized?` method. I'm still preparing a second PR, but in the meantime, I think this is ready to merge if you agree!

PR #6127, opened 2025-12-30
Expand usage of HmisProjectPolicy

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8327

## Type of change
Code clean up


drivers/hmis/app/graphql/mutations/delete_project.rb
```diff
@@ -14,12 +14,16 @@ class DeleteProject < BaseMutation
 
     def resolve(id:)
       record = Hmis::Hud::Project.viewable_by(current_user).find_by(id: id)
+      access_denied! unless record && policy_for(record, policy_type: :hmis_project).can_destroy?
 
-      # While this is redundant with the viewable_by() scope above, this check caches the authorization result so that
+      # While this is redundant with the checks above, this check caches the authorization result so that
       # the project object-level authorization check will succeed even after the project has been deleted
       access_denied! unless current_permission?(permission: :can_view_project, entity: record)
```
ttoomey: I hope that `current_permission` and similar methods can be deprecated in favor of policy checks; standardizing on one API for auth seems important.

I wonder if a deletion mutation really needs to return a stub record with just the id to purge the graphql cache?

Or, if we do need to return the full record, perhaps the resolver could use the policy rather than `current_permission`
```ruby
    # check for the most minimal permission needed to resolve this object
    def self.authorized?(object, ctx)
      # checks to prevent data source leakage, but it is a secondary guard;
      # the viewable_by scope is our primary defense against this.
      super && ctx[:current_user].policy_for(object, policy_type: :hmis_project).can_view?
    end
```

ttoomey (approved):
Nice!

drivers/hmis/app/graphql/mutations/delete_project.rb
gigxz: Thanks for the feedback @ttoomey! Mind taking another quick look at the changes here?

I glossed over this because of concerns with N+1, but I was able to resolve that by adding `preload_project_dependencies` to the `after_paginate` hook -- preloading all the project dependencies for the resolved page of projects. You can see that here: https://github.com/greenriver/hmis-warehouse/pull/6127/changes/c6d98d7b416edbfe20154d64066801fbbc27eae7


This also led to some test cleanup, I realized we didn't have adequate n+1 testing on the projects query, and we didn't have testing that mirrored the frontend, and it was split across a few files weirdly. Since these were some of the first tests we added. I did the cleanup to move them into get_projects_spec https://github.com/greenriver/hmis-warehouse/pull/6127/changes/da2e4c8ac181a714081c33a4d2688701efcabdd8. (And confirmed existing tests passed with my code changes before refactoring!)

gigxz: I realized this introduced another issue, https://github.com/greenriver/hmis-warehouse/pull/6127/changes#r2659257042, marking as draft for now

drivers/hmis/app/graphql/types/hmis_schema/project.rb
```diff
@@ -35,8 +35,7 @@ def self.configuration
     def self.authorized?(object, ctx)
       # current_permission_for_context? checks to prevent data source leakage, but it is a secondary guard;
       # the viewable_by scope is our primary defense against this.
-      permission = :can_view_project
-      super && GraphqlPermissionChecker.current_permission_for_context?(ctx, permission: permission, entity: object)
+      super && ctx[:current_user].policy_for(object, policy_type: :hmis_project).can_view?
```
gigxz, line 38: By removing this, we are losing the data source check: https://github.com/greenriver/hmis-warehouse/blob/a5cdf8a8da1aebc674d11181ee9321db1762bead/drivers/hmis/app/graphql/concerns/graphql_permission_checker.rb#L31-L40

We didn't have test coverage for this for project object-level authorization, but we did somewhere on client which is where I noticed this (other PR).

Check in on:
- should Policies check against a specific data source
  - look at: GraphqlApplciationHelper [policy_for](https://github.com/greenriver/hmis-warehouse/blob/a5cdf8a8da1aebc674d11181ee9321db1762bead/drivers/hmis/app/graphql/concerns/graphql_application_helper.rb#L29-L32) helper can determine the data source id and pass it like `current_user.policy_for(resource, policy_type: policy_type, data_source: data_source)`. That way we can still test policies independently of the `user.hmis_data_source_id` decoration by just passing it directly
  - UserContext is modified to accept data source. UserContext becomes responsible for validating data source, or at least for passing the data source along to the loaders for validation?
  - How do we want to handle cases where policy is invoked for a resource in the wrong data source? I think we want to raise an exception.

gigxz: Proof of concept here for one approach to validating data source https://github.com/greenriver/hmis-warehouse/pull/6134/changes

ttoomey: Thanks @gigxz  - That's a great find!  I think the PR you have should work well

Some questions:
* It looks like `current_permission_for_context?` just returns false for a DS mismatch so raising an error would be a change. Maybe we could just send it to sentry but not raise if we are concerned about breaking the UI?
* I think implementing the data source guard in the context layer rather than the loader is a bit tidier in terms of responsibilities 
* For a minimally breaking change, we could probably just return an empty permission set on mismatch from
* Also a loader could be used to cache the project/data_source relation (maybe worth it? I guess it's consistent with the other loaders)
* I guess we could pull hmis_data_source_id off the user without passing it in explicitly; idk

I made a quick AI-assisted pass at this, see:
https://github.com/greenriver/hmis-warehouse/pull/6135/files

gigxz: Thanks @ttoomey, this was super helpful. I like your approach. I updated my branch https://github.com/greenriver/hmis-warehouse/pull/6134 to incorporate much of that approach, and am waiting on the test suite.

gigxz: I'm going to go ahead and merge _this_ PR in, so I can rebase that other one and add a test case ensuring object-level auth on the Project type is catching cross-data source leaks.

PR #6129, opened 2025-12-31
2026 Test Kit Updates

dtgreiner:
## _Merging this PR_

## Description

#### Active Client method update to only use Method 2 for NBN
```
This report generally uses Active Client - Method 1 from the HMIS Reporting Glossary to determine which clients to include in the reporting universe for [project types] 0 and 2 through 14, except where noted on specific questions. Emergency Shelter – Night-by-Night ([project type] = 1) projects use Active Client – Method 2.
```

#### Update for Q19
Pull data from raw CSV income data instead of cleaned income data

#### Update for Q22
Account for PH vs non-PH in move in and exit clauses

#### Open up all CAPER organizations
Include CoC cleanup in test setup similar to how non-test environment imports work instead of using the raw CSV data. This cleanup fixed a large number of issues in the remaining CAPER organizations.

#### Fix relevant_project_types for TK organizations
A number of organizations had the incorrect relevant_project_types when running the internal consistency tup validations. 

#### Type of change

Bug Fix

### DIFF
--- a/drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/question_twenty_two.rb
+++ b/drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/question_twenty_two.rb
@@ -380,9 +380,38 @@ module HudApr::Generators::Shared::Fy2026
         group_scope = members.where(group.fetch(:cond))
         letter = col_letters.fetch(idx)
 
-        move_in_clause = a_t[:adjusted_move_in_date].not_eq(nil).and(move_in_col.not_eq(nil))
+        # Q22f: Only PH projects (PSH/RRH) - always uses [housing move-in date]
+        # Q22g: Both PH and non-PH projects - requires different logic:
+        #   - PH projects (3, 9, 13, and type 7 with Pay for Success) use [housing move-in date]
+        #   - All other project types use [project start date] as move-in date
+        ph_projects_condition = a_t[:project_type].in([3, 9, 13]).or(
+          a_t[:project_type].eq(7).and(a_t[:pay_for_success].eq(true)),
+        )
+        non_ph_projects_condition = ph_projects_condition.not
 
-        exit_scope = group_scope.where(a_t[:adjusted_move_in_date].eq(nil))
+        # PH projects: use adjusted_move_in_date (housing move-in date)
+        ph_move_in_condition = ph_projects_condition.
+          and(a_t[:adjusted_move_in_date].not_eq(nil))
+
+        # Non-PH projects: use first_date_in_program (project start date)
+        # Note: Only applies to Q22g since Q22f universe only contains PH projects
+        non_ph_move_in_condition = non_ph_projects_condition.
+          and(a_t[:first_date_in_program].not_eq(nil))
+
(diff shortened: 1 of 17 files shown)


drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/question_twenty_two.rb
```diff
@@ -380,9 +380,38 @@ def time_by_race_and_ethnicity_question(question:, move_in_col:, members:)
...
+        #   - PH projects (3, 9, 13, and type 7 with Pay for Success) use [housing move-in date]
+        #   - All other project types use [project start date] as move-in date
+        ph_projects_condition = a_t[:project_type].in([3, 9, 13]).or(
+          a_t[:project_type].eq(7).and(a_t[:pay_for_success].eq(true)),
+        )
+        non_ph_projects_condition = ph_projects_condition.not
+
+        # PH projects: use adjusted_move_in_date (housing move-in date)
+        ph_move_in_condition = ph_projects_condition.
+          and(a_t[:adjusted_move_in_date].not_eq(nil))
```
eanders: Should we just populate the `adjusted_move_in_date` appropriately on build instead of this complexity here, or do you think there's a reason to keep entry date out of the field for non-PH projects?

drivers/hud_apr/spec/models/fy2026/datalab_2_0_spec.rb
```diff
@@ -39,7 +39,7 @@ def run(generator, filter)
   end
 
   before(:all) do
-    setup
+    setup(cleanup_enrollment_cocs: true)
```
eanders, line 42: Do we actually want to do this?  I thought we were supposed to run the test kit against the source data without modification?

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/question_twenty_two.rb
dtgreiner: This was trying to keep the changes isolated to the question being addressed, but after looking at it more I think you're right and we can make the updates when building it out. I'll get an update together to move this logic

drivers/hud_apr/spec/models/fy2026/datalab_2_0_spec.rb
dtgreiner: I'm on the fence with this one. Using this logic makes it more similar to how the data is working in production, but I feel we've seen the opposite of this in the past when working with DQ sections in the test kit.

dtgreiner: It also makes it so we can open the CAPER organizations that we had commented out

drivers/hud_apr/app/models/hud_apr/generators/shared/fy2026/question_twenty_two.rb
dtgreiner: this has been updated [here](https://github.com/greenriver/hmis-warehouse/pull/6129/commits/235b52a65806035e4603dc18be3ca8d10083d248)

eanders (approved):

PR #6134, opened 2026-01-04
Scope HMIS auth policies to current data source

gigxz:
## _Merging this PR_

## Description
Implements a data source guard in the authorization policy to restrict project access to the user's active HMIS session.

- **Authorization Policy**: Adds a mandatory guard to project-level permissions that prevents access to records outside the user's current data source.
- **Authorization Context**: Introduces an efficient caching loader for project data source IDs to minimize database overhead during permission checks.

Based on https://github.com/greenriver/hmis-warehouse/pull/6135

## Type of change


drivers/hmis/app/models/hmis/auth_policies/user_context.rb
```diff
@@ -76,7 +79,7 @@ def clear_referral_assignment_cache!
   end
 
   memoize def project_access_group_loader
-    Hmis::AuthPolicies::ContextLoaders::HmisProjectAccessGroupLoader.new
+    Hmis::AuthPolicies::ContextLoaders::HmisProjectAccessGroupLoader.new(data_source_id: @data_source_id)
```
gigxz: With this approach, its not _super_ clear which loaders need to validate data sources and which dont. For example, the referral loaders don't need to explicitly check data source of the referral because that all goes through project. But that maybe can be clarified with comments

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
```diff
@@ -70,7 +78,23 @@ def clear_referral_assignment_cache!
...
+  def project_belongs_to_current_data_source?(project_id)
+    project_data_source_id = project_data_source_loader.get(project_id)
+    return true if project_data_source_id == user.hmis_data_source_id
+    return false if project_data_source_id.nil?
+
+    Sentry.capture_message(
+      "HMIS Data Source Mismatch: User #{user.id} (DS: #{user.hmis_data_source_id}) " \
+      "attempted to access Project #{project_id} (DS: #{project_data_source_id})",
+    )
+    false
```
martha: I'm not quite following the logic for needing the `project_data_source_id.nil?` check on line 84 -- I understand we want to return false if that happens, but why is there a separate check that returns before logging to sentry? Seems like it would be unexpected for a project not to be associated with a data source ID, and we would want to alert sentry in that case?

martha (approved):
Looks great, thank you Gig!

drivers/hmis/app/models/hmis/auth_policies/base_policy.rb
```diff
@@ -24,6 +24,9 @@ def initialize(context:, resource:)
   # convenience
   def user = context.user
 
+  # convenience
+  def data_source_id = context.data_source_id
+
```
ttoomey: unused?
```suggestion
```

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
ttoomey: the nil guard on `project_data_source_id` might be AI residue from a proof-of-concept I made and didn't review carefully

ttoomey (approved):
Looks good!

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
gigxz: Yes good point we would want to capture that, updated!

PR #6143, opened 2026-01-06
Bug fix on external forms household ID

gigxz:
## _Merging this PR_

## Description
Generated household ID is expected to be 28 chars (expectation is a ~20 lines below ), but it wasn't consistently 28 characters. Fixed in least intrusive way I could find, since this seems like a risky time to be making larger changes.
(AI suggested to use Uint8Array for generating the key but I wasn't sure about browser compatibility so opted not to)

## Type of change
bug fix

### DIFF
--- a/drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
+++ b/drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
@@ -148,8 +148,9 @@ window.addHouseholdTypeListener = function (individualOrFamilySelector) {
   // Generates a "household id" that is unlikely to have collisions among form submissions,
   // without using external library (uuid) or modern browser features (crypto)
   var generateHouseholdId = function () {
-    var current = (new Date()).getTime().toString()
-    var rand = Math.random().toString(16).substring(2)
+    // Keep the ID a constant length: "HH" + 13 (ms timestamp) + 13 (random hex-ish)
+    var current = (new Date()).getTime().toString().slice(-13).padStart(13, '0')
+    var rand = Math.random().toString(16).substring(2).padStart(13, '0').slice(-13)
     return String('HH' + current + rand).toUpperCase();
   }
 


drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
gigxz, line 163: we could alternatively remove the expectation that hh id is a particular length

eanders (approved):
This looks fine to me.  The HHID only needs to be unique within a data source, and 90% of the time only unique within a project is sufficient.  I'm also happy to expand the length requirement, CSV spec is for 32 characters, but that seems uncecessary.

drivers/hmis_external_apis/app/assets/javascripts/external_forms.js
```diff
@@ -148,8 +148,9 @@ window.addHouseholdTypeListener = function (individualOrFamilySelector) {
   // Generates a "household id" that is unlikely to have collisions among form submissions,
   // without using external library (uuid) or modern browser features (crypto)
   var generateHouseholdId = function () {
-    var current = (new Date()).getTime().toString()
-    var rand = Math.random().toString(16).substring(2)
+    // Keep the ID a constant length: "HH" + 13 (ms timestamp) + 13 (random hex-ish)
+    var current = (new Date()).getTime().toString().slice(-13).padStart(13, '0')
+    var rand = Math.random().toString(16).substring(2).padStart(13, '0').slice(-13)
```
ttoomey, line 153: I'm not sure that `padStart`[works in ie11](https://caniuse.com/?search=padstart). It's probably fine though; I suspect IE11 doesn't work with our save mechanics anyway (due to the AWS lamdba).

But if we wanted an es5 compatible version without `padStart`, AI suggests:
```js
var current = ("0000000000000" + (new Date()).getTime().toString()).slice(-13);
var rand = ("0000000000000" + Math.random().toString(16).substring(2)).slice(-13);
```

ttoomey (approved):
Thanks!

PR #6145, opened 2026-01-07
Support CE default contacts

martha:
## _Merging this PR_

## Description

Frontend PR: https://github.com/greenriver/hmis-frontend/pull/1351

- Add database table and rails model for default contacts
- Add graphql queries to support displaying default contacts by project/swimlane
- Add mutation for assigning default contacts 
- Update mutations that create referrals so that default contacts are automatically assigned as participants
- Create a new `CeAdminPolicy` (however, see inline thread for more discussion)
- Other misc changes to support UI updates:
  - Resolve new `active` field on User schema object
  - Allow resolving "eligible referral step assignment users" picklist without passing project. When project is not passed, return all users who can perform referral tasks in any project

## Type of change
New feature


drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -645,6 +645,43 @@ def ce_client(id:)
...
+          swimlane: swimlane,
+          contacts: assignments,
+        )
+      end
+    end
+
+    field :projects_with_ce_default_contacts, HmisSchema::Project.page_type, null: false, description: 'Projects with CE default contacts' do
+      argument :filters, HmisSchema::ProjectsWithCeDefaultContactsFilterOptions, required: false
+    end
+    def projects_with_ce_default_contacts(filters: nil)
```
martha: I lean towards implementing this as a separate query, rather than adding it onto the existing projects query, so that we can use custom filter options without adding them to the base project filters. Let me know how this looks.

db/warehouse/migrate/20260105194328_add_ce_default_swimlane_assignments_table.rb
martha, line 1: Overall, I'm interested in feedback about the terms “Default contacts” vs. “default swim lane assignments”. At first I tried using "default swim lane assignments" everywhere, thinking that "default contacts" is outdated CAS language, but then noticed that "default contacts" is what we decided to use in the UI, probably since it is simpler. I've taken the approach of using “default contacts” in the frontend and graphql layer, while the backend models/tables use “default swim lane assignments.” Let me know if you think this makes sense.

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
```diff
@@ -0,0 +1,32 @@
...
+# Global Coordinated Entry administration policy (not tied to a specific project)
+class Hmis::AuthPolicies::CeAdminPolicy < Hmis::AuthPolicies::BasePolicy
+  # Whether the user can manage global CE default contacts (data source-level)
+  def can_manage_contacts?
+    global_permissions.include?(:can_administrate_coordinated_entry)
+  end
+
+  protected
+
+  # todo - this (ai gen) isn't right, but I need more discussion/understanding of our future approach to "global" (data source level) permissions
```
martha: From the ticket,
> Consider implementing a global CE policy to back this? e.g. CeAdminPolicy.can_manage_contacts?
> For places where we expose the ability to view/manage ce contacts on a specific project, we should check that against the project policy

I'd like to discuss this more. A few things I'm getting tripped up on:
- We say "This is a global permission" in the permission description of the "can administrate coordinated entry" permission, so there is something a bit weird about checking the permission on a project policy. I think I agree/it makes sense, because even if we _say_ it's global, the permission model still supports applying it to some projects and not others. So maybe we should just update the permission description. But, then I think we would be ending up in a bit of a mixed state, where some things this permission governs are global (like being able to view the admin CE UIs) while others are at the project level. Maybe that's ok, but it is hard to explain concisely in a user-facing permission description!
- There is also something weird about calling it and using it as a "global" permission, when ideally what we would want is to authorize this at the data source level. But if I understand correctly, our permission model doesn't support that. In other words, in the permission UI you can create a collection that refers to a data source, and then grant someone this permission on that collection, but under the hood, it's really granting the permission for every project in the collection. Is that understanding correct?

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,89 @@
...
+          owner = Hmis::Hud::Project.viewable_by(current_user).find(input.project_id)
+          access_denied! unless policy_for(owner, policy_type: :hmis_project).can_manage_ce_default_contacts?
+
+          swimlanes = swimlanes.joins(:template).
+            merge(Hmis::WorkflowDefinition::Template.ce.published.viewable_by(current_user).used_in_projects([owner.id]))
+          users = users.can_perform_own_referral_tasks_for(owner).or(users.can_perform_any_referral_tasks_for(owner))
+        else
+          # If no project_id, it's a global assignment. Owner is the current user's HMIS data source
+          owner = GrdaWarehouse::DataSource.find(current_user.hmis_data_source_id)
+          access_denied! unless policy_for(GrdaWarehouse::DataSource, policy_type: :ce_admin).can_manage_contacts?
```
gigxz: Curious about the policy method names `can_manage_ce_default_contacts?` and `can_manage_contacts?` being different- what is the distinction there? If they are meant to authorize the same type of action I think they should be named the same

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,89 @@
...
+        user_ids = input.contacts.map(&:user_ids).flatten.uniq
+        users = Hmis::User.where(id: user_ids)
+
+        if input.project_id.present?
+          owner = Hmis::Hud::Project.viewable_by(current_user).find(input.project_id)
+          access_denied! unless policy_for(owner, policy_type: :hmis_project).can_manage_ce_default_contacts?
+
+          swimlanes = swimlanes.joins(:template).
+            merge(Hmis::WorkflowDefinition::Template.ce.published.viewable_by(current_user).used_in_projects([owner.id]))
+          users = users.can_perform_own_referral_tasks_for(owner).or(users.can_perform_any_referral_tasks_for(owner))
```
gigxz: In keeping with the initiative to move to policy-based authorization, I think it would be worth spending the time to implement these as policy checks. This operation would only be done for a handful of users, so I wouldn't be too worried about n+1s here (e.g. we're not assigning 100 users as default contacts to a swimlane)

```ruby
authorized_users = users.each do |user|
    user.hmis_data_source_id = current_user.hmis_data_source_id # ensure we validate each users access in *this* data source
    raise unless user.policy_for(project, policy_type: :hmis_project).can_perform_referral_tasks?
    # or
    raise unless user.policy_for(data_source, policy_type: :ce_admin).can_perform_referral_tasks?
end
```

gigxz: Maybe we need to use shared scopes _too_, I realize there is a need to be consistent with the picklist options list that the frontend is using 🤔

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -235,12 +235,25 @@ def self.eligible_staff_assignment_user_picklist(project)
...
+
+        # Return users who can perform referral tasks in this project
+        user_scope = user_scope.can_perform_any_referral_tasks_for(project).
+          or(user_scope.can_perform_own_referral_tasks_for(project))
+      else
+        # If project is not passed, return all users who have permission to perform referral tasks at any project.
+        # This is used by the global Default Contacts dropdowns.
+        # The list is too big, but we don't yet have a good way to know who is a CE admin.
+        user_scope = user_scope.can_perform_any_referral_tasks.
+          or(user_scope.can_perform_own_referral_tasks)
```
gigxz: Does this ensure that the users have permission to perform any/own referrals in the **current** data source? We wouldn't want to expose users who are only in other HMIS-es.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
```diff
@@ -235,12 +235,25 @@ def self.eligible_staff_assignment_user_picklist(project)
...
+        # Confirm the project has an applicable CE config
+        return [] if Hmis::ProjectCeConfig.detect_best_config_for_project(project).blank?
+
+        # Return users who can perform referral tasks in this project
+        user_scope = user_scope.can_perform_any_referral_tasks_for(project).
+          or(user_scope.can_perform_own_referral_tasks_for(project))
+      else
+        # If project is not passed, return all users who have permission to perform referral tasks at any project.
+        # This is used by the global Default Contacts dropdowns.
+        # The list is too big, but we don't yet have a good way to know who is a CE admin.
```
gigxz: I think it would be safe to say that users who only have "can perform **own** referral tasks" are not admins, and should not be assigned as default contacts at the data-source level. Maybe I'm forgetting something though. Do have an idea of a use-case there?

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -645,6 +645,43 @@ def ce_client(id:)
...
+        merge(Hmis::WorkflowDefinition::Template.ce.published.viewable_by(current_user)).
+        order(:name, :id).
+        distinct
+    end
+
+    field :global_ce_default_contacts, [HmisSchema::CeDefaultContactsBySwimlane], null: false, description: 'Global Coordinated Entry default contacts, grouped by swimlane'
+    def global_ce_default_contacts
+      access_denied! unless policy_for(GrdaWarehouse::DataSource, policy_type: :ce_admin).can_manage_contacts?
+
+      data_source = GrdaWarehouse::DataSource.find(current_user.hmis_data_source_id)
```
gigxz: Shouldn't we check "can manage contacts" for the specific data source, or does that happen internally? I might need to dig and see, I know I added data source checking to the project access group loader..

It just jumped out to me that it's not the other way around:
```ruby
      data_source = GrdaWarehouse::DataSource.find(current_user.hmis_data_source_id)
      access_denied! unless policy_for(data_source, policy_type: :ce_admin).can_manage_contacts?
```

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
```diff
@@ -0,0 +1,30 @@
...
+  # todo @martha - this (ai gen) isn't right, but I need more discussion/understanding of our approach to "global" (data source level) permissions
+  #  theron's point: >I had hoped we could avoid such global permissions by, say authorizing against a data-source or some other entity.
+  def validate_resource!(arg)
+    return if arg.is_a?(Class)
+
+    raise ArgumentError, "CeAdminPolicy expects a Class, got #{arg.class}"
+  end
+
+  def global_permissions
+    # todo @martha - use context.global_permissions once gig/8328-client-policy is merged
```
gigxz: Sorry that one has been sitting! It got big once I started adding the type-checking - I will revisit and maybe open it as 2 smaller PRs

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
gigxz: Yes you're right, something is missing here. We've been trying to create less future work in https://github.com/open-path/Green-River/issues/6612 by trying to incorporate data-source checks as much as possible, but the policies checking "global permissions" don't account for it yet.

I think the behavior _should_ be: `policy_for(data_source, :admin).can_manage_forms?` returns true if the user has that permission for ANY entity in the data source.

When the permission description says "this is a global permission" my read of that is essentially: "_you can grant this permission to any entity-- project, org, ds-- but the permission will be granted for the whole data source that the entity belongs to_". Which admittedly, could be very confusing for administrators. We could've approached it more strictly and say you must only apply those perms to a data source entity (eg creating an ACL with certain Roles<>Collections is not allowed, or adjusting a role in a certain way is not allowed, it becomes tricky..).

martha: Thanks @gigxz this is helpful! I agree with the desired api for admin policies: `policy_for(data_source, :admin).can_manage_forms?` In this case, `policy_for(data_source, :ce_admin).can_manage_default_contacts?` I'm working on that change. Meanwhile I would like to discuss this part more:
> returns true if the user has that permission for ANY entity in the data source.

because while poking around this, I realized I was wrong about something I wrote in my earlier comment:
> in the permission UI you can create a collection that refers to a data source, and then grant someone this permission on that collection, but under the hood, it's really granting the permission for every project in the collection.

This was a misunderstanding on my part. We do retain the info of which "level" the permission was granted at, through the `Hmis::GroupViewableEntity` model. 

So, for global permission logic, a possible alternative to "allow if the user has the permission granted for any entity in the data source" would be: allow if the user has the permission granted _at the level of the data source_, aka, the user has a Role granting them that permission on an AccessControl whose AccessGroup refers to a *data source* `Hmis::GroupViewableEntity`.

Does that make any sense? What do you think about that approach? (it sounds more aligned with the concept of global permissions to me, but I think it might also cause permission churn because it's less similar to how global permissions are implemented currently)

martha: That might look something like this: https://github.com/greenriver/hmis-warehouse/pull/6145/commits/78610cf040135dc8c4992802afcb4c8a7b470dcc

I haven't tested it enough yet, but figured I'd push what I have since it might be a helpful tool for discussion.

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
martha: Sounds good! My thinking was just that the CE admin policy is already specific to CE, whereas the method on the Project policy needs a bit more context. But I agree, it's less confusing if they match. I will update.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: Yes thanks for the catch, good point!

On the same topic as [the thread above](https://github.com/greenriver/hmis-warehouse/pull/6145#discussion_r2700015996),

> The list is too big, but we don't yet have a good way to know who is a CE admin.

Now that I have a better understanding of `GroupViewableEntities`, I'm wondering if maybe we *do* have a way to know who is a CE admin: someone who has "can perform any referral task" at the level of this data source, as opposed to having that permission on any entity in the data source. Is that list potentially more accurate?

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
martha: notes from convo
- we don't currently check against data source entity, would be hesitant to introduce this
- there might be setups that use project groups instead. e.g. "everything in the DS minus a handful of projects"
- ACL UI also doesn't indicate anything about global permissions needing to be on a DS

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -107,6 +107,23 @@ def load_effective_permissions
     end
   end
 
+  scope :can_perform_referral_tasks_in_project, ->(project) do
```
martha: This scope is ok as-is because the project has to be viewable to the current user; add comments

drivers/hmis/app/models/hmis/user.rb
```diff
@@ -107,6 +107,23 @@ def load_effective_permissions
...
+    can_perform_any_referral_tasks_for(project).or(can_perform_own_referral_tasks_for(project))
+  end
+
+  scope :can_be_global_ce_default_contact, ->(data_source_id) do
```
martha: Update this scope to check the permission at any entity in the datasource, not require it globally at the level of the data source

drivers/hmis/app/models/hmis/user.rb
martha: Don't expose these scopes on the user, just put them in the picklist. Use the authorization policy to actually authorize the action

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
martha: No worries! We're no longer using `global_permissions` here so, no rush on that PR.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Yep you're right, updated!

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: for posterity: We discussed and agreed to keep the previous behavior: this list contains everyone who "can perform any referral tasks" on any entity in the DS, so it's _not_ required to be granted at the level of the whole data source.) 

Reasoning: For example, an installation might have a project group representing "all projects in the DS minus a small handful of projects." A user with this permission on that group _should_ be considered a possible global default contact.

drivers/hmis/app/graphql/types/forms/pick_list_option.rb
martha: This should be correct now with the updated code. I also added some comments to help clarify and rspec to confirm.

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
martha: Decision from our call: Skip using the shared scopes here and just rely on the policies to authorize

martha:
Thanks @gigxz for your helpful comments! I've addressed the permissions issues we've been discussing, so I think this and the frontend PR are ready for continued review (pending the test suite).

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,103 @@
...
+  module Ce
+    class AssignCeDefaultContacts < Mutations::CleanBaseMutation
+      argument :input, Types::HmisSchema::CeDefaultContactsInput, required: true
+
+      field :default_contacts, [Types::HmisSchema::CeDefaultContact], null: false
+
+      def resolve(input:)
+        # Upfront, load all swimlanes and users from the input to confirm they are valid
+        swimlane_ids = input.contacts.map(&:swimlane_id).uniq
+        swimlanes = Hmis::WorkflowDefinition::Swimlane.where(id: swimlane_ids).joins(:template).merge(Hmis::WorkflowDefinition::Template.ce.published.viewable_by(current_user))
```
gigxz: I guess rubocop is okay with this, so I suppose I should be, but I think using newlines is nicer than wrapping such a long statement

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,103 @@
...
+          access_denied! unless policy_for(project, policy_type: :hmis_project).can_manage_ce_default_contacts?
+
+          # Further refine the swimlane scope to only swimlanes that are used in this project
+          swimlanes = swimlanes.merge(Hmis::WorkflowDefinition::Template.used_in_projects([project.id]))
+
+          # Authorize each user with the project policy. Causes n+1 but that's acceptable because the list of users is expected to be small
+          authorized_users = users.filter do |user|
+            user.hmis_data_source_id = current_user.hmis_data_source_id
+            user.policy_for(project, policy_type: :hmis_project).can_perform_referral_tasks?
+          end
```
gigxz: I think this would also be fine as:
```
          # Authorize each user with the project policy. Causes n+1 but that's acceptable because the list of users is expected to be small
          users.each do |user|
            user.hmis_data_source_id = current_user.hmis_data_source_id
            raise "User #{user.id} not authorized to be default contact for Project##{project.id}" unless user.policy_for(project, policy_type: :hmis_project).can_perform_referral_tasks?
          end
```
Since this is an unexpected error, I'm not sure we need to raise with all the user ids as opposed to just raising the first time we see it

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,103 @@
...
+            user.hmis_data_source_id = current_user.hmis_data_source_id
+            user.policy_for(data_source, policy_type: :ce_admin).can_perform_referral_tasks?
+          end
+
+          owner = data_source
+        end
+
+        raise "Swimlane(s) not found: #{swimlane_ids.join(', ')}" unless swimlanes.size == swimlane_ids.size
+        raise "User(s) not found or unauthorized: #{user_ids.join(', ')}" unless authorized_users.size == user_ids.size
+
```
gigxz, line 31: The logic in this mutation looks sound to me! I think my only feedback would be that it's long for a mutation. If I'm reading correctly, lines 17-53 are the validation logic, and lines 55-100 are the mutation logic. Maybe consider refactoring the validations out to a private method (Eg. `validate_input(input)` so it's easier to identify the mutation logic. Take it or leave it, it may be more of a refactoring rabbit hole than I have in mind.

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,103 @@
...
+        Hmis::Ce::DefaultSwimlaneAssignment.import!(to_create) unless to_create.empty?
+
+        # Determine which existing assignments to remove
+        assignments_to_remove = existing_assignments.reject { |key, _| desired_keys.include?(key) }.values
+
+        # Bulk delete removed assignments
+        if assignments_to_remove.any?
+          Hmis::Ce::DefaultSwimlaneAssignment.
+            where(id: assignments_to_remove.map(&:id)).
+            destroy_all
```
gigxz: I think we want  each:destroy! so it raises if any destroy fails

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,103 @@
...
+
+        raise "Swimlane(s) not found: #{swimlane_ids.join(', ')}" unless swimlanes.size == swimlane_ids.size
+        raise "User(s) not found or unauthorized: #{user_ids.join(', ')}" unless authorized_users.size == user_ids.size
+
+        # Load all existing assignments for this owner and swimlanes
+        existing_assignments = Hmis::Ce::DefaultSwimlaneAssignment.
+          where(owner: owner, swimlane_id: swimlane_ids).
+          index_by { |a| [a.user_id, a.swimlane_id] }
+
+        # Determine which assignments should exist, based on the input
```
gigxz: Is the assumption that the `input` contains ALL default contact assignments for the owner (And all swimlanes?)? or could you use it in a partial way to just add new assignments? I think the expectation is that it must be all, which makes sense, but could be clarified with code comments.

drivers/hmis/app/graphql/types/hmis_schema/ce_default_contact.rb
```diff
@@ -0,0 +1,54 @@
...
+    # underlying object is a Hmis::Ce::DefaultSwimlaneAssignment
+
+    field :id, ID, null: false
+    field :user, Application::User, null: false
+    field :swimlane, HmisSchema::CeSwimlane, null: false
+
+    # Polymorphic owner - only one will be non-null. If all are null, this is a global default (owned by data source)
+    field :project, HmisSchema::Project, null: true
+    field :organization, HmisSchema::Organization, null: true
+    field :unit_group, HmisSchema::UnitGroup, null: true
```
gigxz: I'm wondering if it really makes sense to resolve the whole objects here as opposed to just their IDs/Names, for example:
```
    field :project_id, ID, null: true
    field :project_name, String, null: true
    field :organization_id, ID, null: true
    field :organization_name, String, null: false
```

I'm thinking of 2 things here:
1. if we _only_ need the name/id I think it is preferable to only resolve those, to limit the api surface. (eg make it impossible for a malicious user to query `ceDefaultContact { project { enrollments { client { ... } } }`. Of course object-level authorization should prevent unauthorized access, but imo it's just better to keep client data off a type that doesn't need it.
2. Can we guarantee that the user will have access to the owner record if they can access this type? I'm unsure of that. For example, say we expose the list of default contacts on the Project page, including "inherited" default contacts. I think in practice yes the user would have access to resolve organization if they can resolve project, so maybe it's a non-issue _now._ But I have a vague concern about expecting the user to be authorized to resolve the owner object more generally.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
```diff
@@ -645,6 +645,43 @@ def ce_client(id:)
...
+        includes(:user, swimlane: :template).
+        joins(:swimlane).
+        order(Hmis::WorkflowDefinition::Swimlane.arel_table[:id])
+
+      all_assignments.group_by(&:swimlane).map do |swimlane, assignments|
+        OpenStruct.new(
+          swimlane: swimlane,
+          contacts: assignments,
+        )
+      end
```
gigxz: I'm seeing this code repeated in the Project type `ce_default_contacts` field. I'd suggest pulling it out into a concern so the implementations don't drift (See "HasCeOpportunities", etc). That way it's also much simpler when we get to adding contacts at other levels (e.g. Organization).

Then this could be simplified to:

```suggestion
      resolve_ce_default_contacts(data_source.ce_default_swimlane_assignments)
```

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
gigxz: This feels a bit off to me. I know we previously talked about concerns with the top-level query bloat (https://github.com/open-path/Green-River/issues/7994), and we didn't resolve a specific approach, but this seems to explicitly go against our pattern of re-using existing queries for multiple use-cases. What was the concern about adding a user (or,  "default-ce-contact-user") filter to the regular project query?

drivers/hmis/app/models/hmis/filter/projects_with_ce_default_contacts_filter.rb
```diff
@@ -0,0 +1,74 @@
...
+
+      # Create Arel join node for assignments
+      assignment_join = a_t.create_join(a_t, a_t.create_on(assignment_join_condition), Arel::Nodes::InnerJoin)
+
+      scope.
+        joins(:organization, :data_source).
+        joins(assignment_join).
+        where(a_t[:id].in(assignment_ids)).
+        distinct
+    end
```
gigxz: Woah. I'm trying to wrap my head around this use-case and if there's a simpler way to achieve it. We used to have some code like this for form instance applicability, and the addition of the InstanceProjectMatch class https://github.com/greenriver/hmis-warehouse/blob/1fa33bbc4add0f5c7ba9b40a1f5c119cb23e3165/drivers/hmis/app/models/hmis/form/instance.rb#L172-L175 really helped encapsulate that applicability logic so it wasn't repeated in lots of places. (With a tradeoff for performance, which hasn't been a problem). I worry about re-adding a similar pattern and the maintenance complexity it comes with. I don't think we are in a super rush to add this user filtering capability, so maybe it makes sense to pull out into it's own issue?

martha: Thanks @gigxz yeah I agree, this complexity was definitely giving me pause as well.

I think, given that I'm over estimate on this ticket already, I like the idea of splitting this out into a separate issue. Created https://github.com/open-path/Green-River/issues/8753.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Sounds good, I can revert.

I was just thinking it feels messy to add such a specific filter to such a general query, but sounds like that's not a concern

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
martha: Thanks! I ended up refactoring a little more, let me know how you think it looks.

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
martha: Yes, that's the assumption and I've updated the comments.

I guess strictly speaking, if a whole swimlane wasn't included in the input, the mutation wouldn't delete existing assignments in that swimlane. So in that sense it would be possible to pass partial input. But that's not how the frontend uses it.

drivers/hmis/app/graphql/types/hmis_schema/ce_default_contact.rb
martha: Thanks, that makes sense. Right now the frontend really only needs to know if the contact is global or not.

drivers/hmis/app/graphql/types/hmis_schema/query_type.rb
martha: Thanks, sounds good!

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
gigxz: Got it, thanks!

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
gigxz: I think thats a nice improvement for readability, thanks!

drivers/hmis/app/graphql/mutations/ce/assign_ce_default_contacts.rb
```diff
@@ -0,0 +1,111 @@
...
+            owner: owner,
+          )
+        end
+        Hmis::Ce::DefaultSwimlaneAssignment.import!(to_create)
+
+        # Delete assignments that aren't included in the input
+        to_remove = existing_assignments.reject { |key, _| desired_keys.include?(key) }.values
+        Hmis::Ce::DefaultSwimlaneAssignment.
+          where(id: to_remove.map(&:id)).
+          each(&:destroy!)
```
gigxz: I though I had a comment about this but can't find it now - is there a reason we're not doing import+destroy inside a transaction?

martha: No we should! I'll fix

martha:
@gigxz, thanks for all your help and feedback on this. I think I've addressed everything here and on the frontend PR, and did a self-review once-over, so hopefully it's ready to merge this week. Let me know if you'd like to take another look since a few of the changes were pretty substantial. Thank you!

drivers/hmis/app/graphql/types/hmis_schema/has_ce_default_contacts.rb
```diff
@@ -0,0 +1,48 @@
...
+          default_field_options = { type: [HmisSchema::CeDefaultContactsBySwimlane], null: false, description: description }
+          field_options = default_field_options.merge(override_options)
+          field(name, **field_options) do
+            instance_eval(&block) if block_given?
+          end
+        end
+      end
+
+      included do
+        def resolve_ce_default_contacts(scope = Hmis::Ce::DefaultSwimlaneAssignment, **args)
```
gigxz: Our pattern in most other concerns is to default to the share association name on the object. That feels a bit safer than defaulting to all records:
```suggestion
        def resolve_ce_default_contacts(scope = object.ce_default_swimlane_assignments, **args)
```

drivers/hmis/app/graphql/types/hmis_schema/has_ce_default_contacts.rb
```diff
@@ -0,0 +1,48 @@
...
+        return Hmis::Ce::DefaultSwimlaneAssignment.none unless Hmis::Ce.configuration.enabled?
+
+        scope = scope.includes(:user, swimlane: :template).
+          joins(:swimlane).
+          order(Hmis::WorkflowDefinition::Swimlane.arel_table[:id])
+
+        scope.group_by(&:swimlane).map do |swimlane, assignments|
+          OpenStruct.new(
+            swimlane: swimlane,
+            contacts: assignments,
```
gigxz: Suggest sorting the assignments in a deterministic sort order


```suggestion
            contacts: assignments.sort_by(&:id),
```

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -174,6 +174,34 @@ def active?
...
+
+      # Collect all possible owners
+      owners = [target_project, unit_group, organization, data_source]
+
+      # Get all default assignments for these owners and this referral's swimlanes
+      default_assignments = Hmis::Ce::DefaultSwimlaneAssignment.
+        where(owner: owners).
+        where(swimlane_id: swimlanes.pluck(:id)).
+        includes(:user, :swimlane, :owner)
+
```
gigxz: [Take-or-leave comment, I think its OK as is.] This logic feels like it should go on the DefaultSwimlaneAssignment model. For example when debugging we may want to see "who is default assigned to this unit group/referral" and we would have to reconstruct this logic manually since its hidden inside the creation method.

For example:
```ruby
    # in DefaultSwimlaneAssignment class

    # included suffix '_including_inherited' for clarity, but it should be consistent with for_project
    scope :for_unit_group_including_inherited, ->(unit_group) do
      project = unit_group.project
      owners = [unit_group, project, project.organization, project.data_source].compact
      where(owner: owners)
    end


    # here:
    def create_default_participants!
      default_swimlane_assignments = Hmis::Ce::DefaultSwimlaneAssignment.for_unit_group_including_inherited(unit.unit_group).
        where(swimlane_id: swimlanes.pluck(:id)).
        includes(:user, :swimlane)

      default_swimlane_assignments.each do |assignment|
        participants.find_or_create_by!(user: assignment.user, swimlane: assignment.swimlane)
      end
    end
```

drivers/hmis/app/models/hmis/ce/default_swimlane_assignment.rb
```diff
@@ -0,0 +1,31 @@
...
+
+    # Fetch assignments for a project, including inherited assignments from org and data source
+    scope :for_project, ->(project) do
+      swimlane_scope = Hmis::WorkflowDefinition::Swimlane.
+        joins(:template).
+        merge(Hmis::WorkflowDefinition::Template.ce.published.used_in_projects([project.id]))
+
+      owners = [project, project.organization, project.data_source].compact
+      joins(:swimlane).merge(swimlane_scope).where(owner: owners)
+    end
```
gigxz, line 25: It seems like there is a lot going on here with the added swimlane filtering, that's not obvious by the naming of the scope. Maybe it can be split out into separate scopes?

```ruby
project_swimlanes = Hmis::WorkflowDefinition::Swimlane.used_in_project(project) # and: ce_swimlanes that is resolved on the GQL Project type could also use this? currently there is logic in the graphql field
project_default_assignments = Hmis::Ce::DefaultSwimlaneAssignment.for_project_including_inherited(project)

project_assignments_for_project_swimlanes = my_default_assignments.joins(:swimlane).merge(project_swimlanes)
```

If I understand correctly, the use-case is to filter _out_ default assignments from higher levels that are irrelevant to the project. For example if there is a default data source assignment for swimlane XYZ, but swimlane XYZ is not used in project X. Technically that assignment would still be "inherited" by the project, but we don't care about showing it in the list of default contacts for the project (I think thats the only place this scope is used) because it's not relevant. Is that right?

martha: Sure! Yes, your description of the use-case is correct.

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: true
+
+# Global Coordinated Entry administration policy (not tied to a specific project)
+class Hmis::AuthPolicies::CeAdminPolicy < Hmis::AuthPolicies::BasePolicy
+  # Whether the user can manage CE default contacts in their data source.
+  def can_manage_ce_default_contacts?
```
martha: - Default Contacts policy
- Or, put this on the Global referral policy (see similar example of Form Instance which lives on form policy) <-- do this option for now. if we end up with a lot of default-contact-related checks it might make sense to move (add comments)
- in the future people might be able to manage default contacts at their own project

drivers/hmis/app/models/hmis/auth_policies/ce_admin_policy.rb
```diff
@@ -0,0 +1,29 @@
...
+    # This is a global permission, so the policy returns true if the user has this permission at any entity in the data source.
+    data_source_permissions.include?(:can_administrate_coordinated_entry)
+  end
+
+  # Whether the user can perform referral tasks at any entity in the data source.
+  def can_perform_referral_tasks?
```
martha: moves to global Referral policy

gigxz:
I squash-merged https://github.com/greenriver/hmis-warehouse/pull/6173 into release-198, so hopefully it should be straightforward to resolve the conflicts. I'll hold off on re-reviewing until those are fixed and re requested. thanks!

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -104,6 +104,16 @@ def can_index?
       true
     end
 
+    def can_perform_referral_tasks?
+      context.global_permissions.include?(:can_perform_any_referral_tasks)
```
gigxz: `global_permissions` is made available on the BasePolicy, so this can be simply `global_permissions.include?(:can_perform_any_referral_tasks)`

drivers/hmis/app/models/hmis/auth_policies/hmis_project_policy.rb
```diff
@@ -36,6 +36,16 @@ def can_view_outgoing_referral_summaries?
       project_permissions.include?(:can_view_outgoing_referral_details) || project_permissions.include?(:can_manage_outgoing_referrals)
     end
 
+    def can_manage_ce_default_contacts?
+      project_permissions.include?(:can_administrate_coordinated_entry)
```
gigxz: `can_administrate_coordinated_entry` is considered a "global" role (marked as such in the role definition), which I interpret to mean "if you have this anywhere you have this everywhere (in the data source)". It's not a distinction that we actually apply in practice, there is nothing stopping us from checking it at the project level like you did here. But it is a little confusing to me.

Was the idea here to future-proof the policy API to account for a future world where we have a project-level permission (like "can manage CE contacts at project")? I can see the argument for that, but I'm not sure it's worth the potential confusion that may come up from having "can_administrate_coordinated_entry" applied at the project level. Let me know what you think and if I was interpreting that correctly!

martha: Thanks, good point, it is confusing. This was my interpretation of the ticket's suggestion,
> For places where we expose the ability to view/manage ce contacts on a specific project, we should check that against the project policy.

I think your summary is correct -- we usually interpret global roles as you described -- but I also think there is something weird about the setup where a user has "can administrate coordinated entry" at Project A, and that grants them the ability to manage default contacts at Project B. So I guess I am on the fence.

I'm happy to go with your suggestion here. Just to make sure I understand, do you think we should remove this project policy method entirely, or keep it and have it check against the user's global (not project) permissions?

gigxz: Maybe keep it on this project instance policy with a comment like:

```ruby
def can_manage_ce_default_contacts?
      # Currently CE contact management is authorized using the global permission "can_administrate_coordinated_entry" which grants access to manage default contacts at all projects
      # In the future if/when we allow project-lvel admins to manage their own contacts, this may be replaced with a check for a new non-admin permission like `project_permissions.include?(:can_manage_ce_default_contacts)`
      global_permissions.include?(:can_administrate_coordinated_entry)
```

gigxz (approved):
Thanks for your work on this!

PR #6149, opened 2026-01-08
Simple Report CSV Archival

dtgreiner:
## _Merging this PR_

## Description

Add the ability to archive Simple Report instances using CSV files stored in ActiveStorage. This feature will create the CSV files on report build and after a defined grace period (currently set to 60 days), purge the database data. This flags the reprot as "Archived" and on report load, a the user can reload the report data into the database through the UI. Once done, this resets the 60 day grace period before the database data is again purged.

## Type of change

New feature


dtgreiner:
Screenshot 2026-01-08 at 10 59 54 AM
Screenshot 2026-01-08 at 11 00 07 AM

app/models/concerns/report_archival.rb
```diff
@@ -0,0 +1,146 @@
...
+    end
+  end
+
+  def incomplete_archival?
+    return false unless archival_metadata.present? && archival_metadata['archived_at'].present?
+
+    !archival_complete?
+  end
+
+  def missing_archival_files?
```
eanders: Is this method called anywhere?

app/models/concerns/report_archival.rb
```diff
@@ -0,0 +1,146 @@
...
+  # ============================================================================
+
+  def archive_to_csv!
+    service = Reports::ArchiveReportService.new(self)
+
+    # Reload report to ensure associations are fresh after bulk imports
+    reload
+
+    success = service.archive!
+    return if success
```
eanders: Maybe:
```suggestion
    return true if success
```

app/models/concerns/report_archival.rb
```diff
@@ -0,0 +1,146 @@
...
+
+    Rails.logger.error("Failed to archive report ##{id} (#{self.class.name}). Errors: #{service.errors.inspect}")
```
eanders:
```suggestion
    Rails.logger.error("Failed to archive report ##{id} (#{self.class.name}). Errors: #{service.errors.inspect}")
    false
```

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+      return false unless eligible?
+
+      # Skip if already archived to avoid re-attaching files (which triggers PurgeJob)
+      if report.archived? && report.archival_complete?
+        Rails.logger.info("ArchiveReportService: Report ##{report.id} is already archived, skipping")
+        return true
+      end
+
+      config = report.archival_csv_config
+      return false if config.empty?
```
eanders: This check _should_ be unnecessary as it was also completed as part of `eligible?` which is called on line 36.  Maybe we should make `config` a method that memoizes the result of `report.archival_csv_config` that way you wouldn't have to load it multiple times.  Although, if is just a hash so maybe that doesn't matter.

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+
+        generate_and_attach_csv(attachment_name, csv_config)
+        update_file_status(attachment_name, attached: true, attached_at: Time.current.iso8601)
+      rescue StandardError => e
+        @errors << { attachment: attachment_name, error: e.message }
+        update_file_status(attachment_name, attached: false, error: e.message)
+      end
+
+      # Mark as complete if all files attached
+      if @errors.empty?
```
eanders: maybe:
```suggestion
      return false unless @errors.empty?
```

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+          @errors << { attachment: attachment_name, error: e.message }
+          update_file_status(attachment_name, attached: false, error: e.message)
+        end
+      end
+
+      # Mark as complete if all files attached
+      update_archival_metadata(complete: true, completed_at: Time.current.iso8601) if missing_files.empty?
+
+      @errors.empty?
+    end
```
eanders: When would we call `complete_archival` instead of `archive!`?  It seems like we should only get here if something fails.

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+      # Get data from association
+      association_name = csv_config[:association]
+      raise "No association specified for #{attachment_name}" unless association_name
+
+      # Ensure report is reloaded to get fresh associations after bulk imports
+      report.reload unless report.new_record?
+      association = report.send(association_name)
+      # Force reload if it's a relation
+      association = association.reload if association.is_a?(ActiveRecord::Relation)
+      records = association
```
eanders: This seems like it's trying to hard.  We should be able to just:
```suggestion
      records = report.send(association_name).association
```

Was that not working for some reason?

Looks like we might even be able to add a `to_a` on the end and then save directly to `records_to_csv`  In what case would the association not be an association at this point?

Ok, one more thought, I'm not sure we want to call `to_a` on it, we probably want `find_in_batches` and then to write out the CSV in batches.  If we call `.to_a` on a HUGE data set (like an SPM for a full state) we'll probably run out of memory.

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+      end
+
+      # Attach to report
+      attachment = report.send(attachment_name)
+      attachment.attach(
+        io: StringIO.new(csv_content),
+        filename: filename,
+        content_type: 'text/csv',
+      )
+    end
```
eanders: I _think_ all of this should be setup to stream directly to the attachment on S3 or a tmp file that can then be moved.  Keeping it all in memory twice (once for the array, once for the CSV) seems like a really bad idea.

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+            column_names.map { |col| record.send(col) }
+          elsif record.respond_to?(:[])
+            column_names.map { |col| record[col] }
+          else
+            column_names.map { |col| record.send(col) }
+          end
+          csv << values
+        end
+      end
+    end
```
eanders: Given that this is new code, and we plan to send it a hash of associations, can we just assume we have an association of `ActiveRecord` based models here?  Again, I think we actually want to stream this, but the concept is the same.

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+
+    def update_archival_metadata(updates)
+      current_metadata = report.archival_metadata || {}
+      report.update_column(:archival_metadata, current_metadata.merge(updates.with_indifferent_access))
```
eanders: Is there a reason we're not calling:
```suggestion
      report.update(archival_metadata: current_metadata.merge(updates.with_indifferent_access))
```

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,249 @@
...
+      if result[:success]
+        Rails.logger.info("Purged database data for report #{report.class.name} ##{report.id} after archival")
+      else
+        # Log errors but don't fail archival if purge fails
+        Rails.logger.error("Failed to purge database data for report #{report.class.name} ##{report.id}: #{result[:errors].join(', ')}")
+        @errors << { purge: result[:errors] }
+      end
+    rescue StandardError => e
+      # Log errors but don't fail archival if purge fails
+      Rails.logger.error("Error purging database data for report #{report.class.name} ##{report.id}: #{e.message}")
```
eanders: Maybe we should send this to sentry?

app/services/reports/constants.rb
eanders, line 1: This file should be called `reports.rb` or should include a `class Constants` I think.  Either way, I'd probably just remove it and put the constant in one of the other files and reference it as necessary.

Alternatively, you could use AppConfigProperties and set a default 60 days config.  Then we could override when someone needs a longer (or shorter) grace period.

app/services/reports/purge_archived_report_data_service.rb
```diff
@@ -0,0 +1,282 @@
...
+        Rails.logger.error("#{error_msg}\n#{e.backtrace.first(5).join("\n")}")
+
+        {
+          success: false,
+          errors: @errors,
+        }
+      end
+    end
+
+    def can_purge?
```
eanders: Is this being used?

app/models/concerns/report_archival.rb
dtgreiner: nope, I'm removing it

app/services/reports/archive_report_service.rb
dtgreiner: cleaning this up to add the memoized config method and remove the extra check

app/services/reports/archive_report_service.rb
dtgreiner: Removing `complete_archival` method. I missed stripping this one out

app/services/reports/archive_report_service.rb
dtgreiner: When we were reading the data from the CSV for the view, there was a case where I was looking at storing data beyond just hte association. This is leftover from that. Now it should just be a 1:1 relationship with the table data so these extra checks shouldn't be needed. I'm cleaning this up and including batching

lib/tasks/reports/migrate_to_csv.rake
eanders, line 1: I think we can probably delete this.  We should, instead, have a scheduled task, or method on some class, that gets called periodically, that introspects on the reports that CAN be archived, and moves some batch if any need archiving each day.  Not sure on batch size, it'll probably depend on how long it takes, but maybe we could do 100 per run.

drivers/performance_measurement/app/controllers/performance_measurement/warehouse_reports/reports_controller.rb
```diff
@@ -126,6 +126,20 @@ def provider_comparisons
       end
     end
 
+    def reload_from_csv
+      require_can_publish_reports!
```
eanders, line 130: Is this the right permission?  We're not actually publishing, so I think we can just check if you can view this report. We'll need the check that looks at who you are and if you can see _this_ report rather than if you can see the report type in general.

docs/features/report_archival/report-csv-archival-user-guide.md
eanders, line 1: This file could use some cleanup, I think there's just too much in there, and most of it isn't really relevant or necessary to understand how the thing works.  I don't think anything is inherently wrong, but let's tighten it up when the code has settled to just cover what an engineer would need to use the feature.

docs/features/report_archival/report-csv-archival-migration-guide.md
```diff
@@ -0,0 +1,306 @@
...
+  complete
+end
+```
+
+**Note:** 
+- The `archive_to_csv!` method is automatically available when you include the `ReportArchival` concern - no need to implement it yourself.
+- Archival eligibility is automatically determined by checking if the report includes the `ReportArchival` concern and has a non-empty `archival_csv_config`. No additional configuration is needed.
+- Archival happens automatically after successful report generation. Database data remains intact during the grace period and can be purged after the grace period expires.
+
+### Step 5: Add Reload Button to Views (Optional)
```
eanders: I'd say this isn't optional, if we archive the data, we should add the restore button.  Probably worth making a centralized partial you can just `render button_partial(report_path)` or similar so we can easily add it.

docs/features/report_archival/report-csv-archival-migration-guide.md
```diff
@@ -0,0 +1,306 @@
...
+    flash[:error] = "Failed to reload report data: #{result[:errors].join(', ')}"
+  end
+
+  redirect_to your_report_path(@report)
+end
+```
+
+**Note:** The reload button will only show when `@report.purged?` returns `true`. When not purged, the normal report content is displayed.
+
+### Step 6: Add Tests
```
eanders: I think everything here down can be removed.  The top is the important part.

config/schedule.rb
```diff
@@ -186,6 +186,12 @@
     at: import_cleanup_time,
     interruptable: false,
   },
+  {
+    task: 'reports:csv:purge_eligible',
```
eanders: Maybe we do want to keep the rake task.  As we discussed, I think it should actually not archive on run, but should archive and then purge when eligible.  That'll prevent duplicate data being saved for any period of time and it streamlines the migration for new reports.  Just make it so the report _can_ be archived, then this job should be able to find it and archive in batches.

app/views/warehouse_reports/_reload_archived_report.haml
eanders, line 1: Maybe this is the partial I was thinking of!

app/services/reports/reload_report_from_csv_service.rb
eanders, line 1: I wonder if archiving a SQL insert statement would be safer.  That might be just as fast and could preserve types and such.  I worry if we have to manipulate the data too much, it may not match what was exported.  That said, it's probably fine since it's just for reference.

app/services/reports/constants.rb
dtgreiner: Moving this to use `AppConfigProperty`

app/services/reports/purge_archived_report_data_service.rb
dtgreiner: nope, removing

dtgreiner:
@eanders  I've addressed most of the code issues you've brought up. I did not make meaningful changes to the .md nor the spec files yet. I have some more work to clean up here but wanted to get it checked in before the weekend.

app/models/concerns/report_archival.rb
```diff
@@ -0,0 +1,143 @@
...
+
+    purge_eligible_at_str = archival_metadata['purge_eligible_at']
+    return false unless purge_eligible_at_str
+
+    purge_eligible_at = Time.parse(purge_eligible_at_str)
+    purge_eligible_at <= Time.current
+  end
+
+  def archival_status
+    has_csv = archival_metadata.present? && archival_metadata['archived_at'].present?
```
eanders: Given that `archival_metadata.present? && archival_metadata['archived_at'].present?` is called 3 times, maybe a private `csv_exists?` method is worthwhile?
```
def csv_exists?
  archival_metadata.present? && archival_metadata['archived_at'].present?
end
```

app/services/reports/archive_report_service.rb
```diff
@@ -0,0 +1,179 @@
...
+        return true
+      end
+
+      # Use existing expected_files from metadata if present, otherwise calculate from config
+      existing_metadata = report.archival_metadata || {}
+      expected_files = existing_metadata['expected_files'] || config.keys.map(&:to_s)
+      expected_file_count = expected_files.size
+
+      # Ensure report is reloaded once before processing attachments
+      report.reload unless report.new_record?
```
eanders: Is there a scenario `report` would be a `new_record?` now?
I'm pretty sure we should only ever get here if the report has been completed for X days, which would mean it should never be new.

Also, not sure why we'd need to reload it.

If there is a reason, let's document it

lib/report_csv_reader.rb
```diff
@@ -0,0 +1,106 @@
...
+      attachment.first
+    else
+      attachment
+    end
+
+    return false unless file.present?
+
+    # Download and parse CSV
+    csv_content = file.download
+    @csv_data = CSV.parse(csv_content, headers: true, header_converters: :symbol)
```
eanders: Is this going to load the entire CSV memory?  I think we probably want to open this as some sort of IO stream so we don't end up loading millions of rows into RAM.

Above I believe you call `all.count`. We should see if this reader class could just memoize the count of rows so we don't have to read the whole thing.  I think the [docs say you can do it this way](https://ruby-doc.org/stdlib-3.0.0/libdoc/csv/rdoc/CSV.html):


```
File.open(path) do |file|
  CSV.table(file)
end # => #<CSV::Table mode:col_or_row row_count:3>
``` 

or maybe just:

```
CSV.table(path) # => #<CSV::Table mode:col_or_row row_count:3>
```

I'm unclear if the `row_count` is available without reading the entire file.

app/services/reports/reload_report_from_csv_service.rb
```diff
@@ -0,0 +1,208 @@
...
+      config_to_reload = config.select { |attachment_name, _| expected_files.include?(attachment_name.to_s) }
+      return { success: false, errors: ['No matching archival configuration found for expected files'] } if config_to_reload.empty?
+
+      reload_associations(config_to_reload)
+    end
+
+    def reload_association(attachment_name, csv_config)
+      # Read CSV data
+      reader = ReportCsvReader.new(report, attachment_name)
+      csv_data = reader.all
```
eanders: Ok, I think this is another place we need to stream rather than do a full load.  if `reader.csv_row_count` or similar can be added, then I think we should be able to stream read the file and batch insert every 5,000 rows below.

app/services/reports/purge_archived_report_data_service.rb
```diff
@@ -0,0 +1,251 @@
...
+
+    def csv_row_count(attachment_name)
+      # Ensure attachment_name is a symbol
+      attachment_name = attachment_name.to_sym if attachment_name.is_a?(String)
+      attachment = report.send(attachment_name)
+      return 0 unless attachment.attached?
+
+      # Use ReportCsvReader to count rows
+      reader = ReportCsvReader.new(report, attachment_name)
+      reader.all.count
```
eanders: Another spot where we should just get the count from the `CSV Table` instead of loading all the data.

drivers/homeless_summary_report/app/models/homeless_summary_report/report.rb
```diff
@@ -36,7 +41,7 @@ class Report < SimpleReports::ReportInstance
     end
 
     scope :ordered, -> do
-      order(updated_at: :desc)
+      order(completed_at: :desc)
```
eanders: I think we'll want `nulls first` or to at least confirm that's true, so incomplete reports don't drop off the page.

drivers/performance_measurement/app/models/performance_measurement/report.rb
```diff
@@ -45,15 +52,15 @@ class Report < SimpleReports::ReportInstance
     end
 
     scope :ordered, -> do
-      order(updated_at: :desc)
+      order(created_at: :desc)
```
eanders: This differs from drivers/homeless_summary_report/app/models/homeless_summary_report/report.rb, which used `completed_at`.  Is that intentional?

lib/tasks/reports/migrate_to_csv.rake
```diff
@@ -0,0 +1,121 @@
...
+        where.not(completed_at: nil).
+        where(Arel.sql("archival_metadata->>'purged_at' IS NULL")).
+        where(
+          Arel.sql(
+            "((archival_metadata->>'purge_eligible_at' IS NOT NULL AND (archival_metadata->>'purge_eligible_at')::timestamp <= ?) OR " \
+            "(archival_metadata->>'purge_eligible_at' IS NULL AND completed_at + INTERVAL '#{grace_period_days} days' <= ?))",
+            now, now
+          ),
+        ).
+        order(completed_at: :asc)
```
eanders: Should some of this be moved to a scope on `SimpleReports::ReportInstance`?

drivers/performance_measurement/app/models/performance_measurement/report.rb
dtgreiner: I thought I had switched both of them. 

When rehydrating a report, the `updated_at` timestamp was being updated causing the old report to bubble to the top of the page. It felt like `created_at` would fit better here and at a glance it looks like we don't have a standard for how to order the history for reports. Some are using `updated_at` and others are using `created_at`. Unless you feel strongly about about the best way to order these, I might include updating report ordering for all of the `updated_at` reports to switch to `created_at`.

lib/tasks/reports/migrate_to_csv.rake
dtgreiner: that makes sense to me, I'll get an update pushed

drivers/performance_measurement/app/models/performance_measurement/report.rb
eanders: I think that's reasonable (using `created_at` globally.)  
The benefit to using `updated_at` is that if one takes longer to run it ends up a the top.  What would you think about just preventing updating the `updated_at` when re-hydrating?  If that's at all hard though, let's just standardize on `created_at`.

dtgreiner: We could probably bypass the updated_at by using `report.update_columns(...)` instead of `report.upddate(...)`. I'll give that a look and see if it's going to be a lot of effort.

app/models/concerns/report_archival.rb
eanders: @dtgreiner was this a no go?

app/services/reports/archive_report_service.rb
eanders: @dtgreiner maybe I'm reviewing the wrong code, but why do we need this?

app/models/concerns/report_archival.rb
dtgreiner: It looks like a handful of these comments either didn't come through when I was looking at it earlier or were missed. Running through them now.

dtgreiner: This will be included in the next commit

app/services/reports/archive_report_service.rb
dtgreiner: These appear to be unneeded and will be cleared out in the next commit

lib/report_csv_reader.rb
dtgreiner: doing some cleanup on this file for memory efficiency. It looks like the `CSV.table(file)` does still load the whole CSV into memory to get the size, but we should be able to stream/memoize as suggested when accessing the file

dtgreiner:
@eanders this should be ready now with the recent comments addressed. I'm planning on bumping this to release-196 too after we have things ready to go.

eanders (commented):
@dtgreiner this looks great to me.  Thank you!
I say retarget 196 and let's do some QA!

eanders (approved):

PR #6152, opened 2026-01-09
CI improvements

ttoomey:
## _Merging this PR_

## Description
Consolidate Rails CI workflows into a single pipeline with smart routing and support for targeted test execution.

Also rebalances our buckets so the test suite completes more quickly.

- **CI Infrastructure**: Merged `rails_system_tests.yml` and `warehouse_system_tests.yml` into `rails_tests.yml` to reduce maintenance overhead and unify the testing pipeline.
- **Matrix Routing**: Promoted CI matrix generation logic to a standalone Ruby script (`bin/ci_matrix_router.rb`) and added unit test coverage in `spec/bin/ci_matrix_router_spec.rb`.
- **Targeted Test Execution**: Implemented "Focused Runs" allowing developers to execute specific specs via `workflow_dispatch` inputs or `ci-focus:` tags in commit messages.
- **Smart Routing**: Added logic to automatically route focused tests to the appropriate environment (Unit, HMIS System, or Warehouse System) and skip irrelevant jobs.

## Type of Change
Refactor / Code clean-up


.github/workflows/rails_tests.yml
```diff
@@ -4,77 +4,142 @@ name: Rails Tests
...
 on:
   push:
+    branches:
+      - stable
+      - "release-*"
+  pull_request:
     branches:
       - "*"
       - "**/*"
+  workflow_dispatch:
```
ttoomey, line 14: ai suggested this, I don't think it actually works unless it's merged to the default branch

eanders: What is this supposed to do?

ttoomey:
Seems to be mostly working as intended. I'll rebase this to r195

eanders (approved):
@ttoomey I think this looks wonderful.  I like the idea of having one **Rails Tests** check and having system tests run as sub-sections.  The ability to filter to a single test or set of tests also seems worth-while.  Thanks!

lib/tasks/db.rake
```diff
@@ -69,7 +69,7 @@ def load_structure_sql(connection_class, filename)
   env = {}
   password = config[:password]
   env['PGPASSWORD'] = password if password && !password.empty?
-  command = ['psql', '-v', 'ON_ERROR_STOP=1']
+  command = ['psql', '-v', 'ON_ERROR_STOP=1', '--quiet']
```
ttoomey, line 72: reduces log verbosity, prunes 40k lines of log output

.github/workflows/rails_tests.yml
ttoomey: In theory, when this is merged to `stable` it will allow us [to re-run targeted specs](workflow_dispatch) (without having to push). Individual specs could be run using the GH interface. Not sure if it will work but thought it might be a convenience if it does.

ttoomey:
re-request review
* rebalanced tests to get CI to complete in ~1 hour
* easier to rebalance with [ci-profile]
* improved docs
* safer path handling in matrix_router
* undo unnecessary changes to env vars

eanders (approved):

PR #6154, opened 2026-01-12
Add created_by_id to CE opportunity

martha:
## _Merging this PR_

## Description
- This PR introduces a new "created_by_id" column on the CE opportunity table, as pre-work for the linked ticket.
- While working on this, I noticed that we are **not** currently creating PaperTrail versions for opportunities when they are created, due to the [`import!`](https://github.com/greenriver/hmis-warehouse/pull/6154/changes#diff-0c06452b4f6696e54b16f8d6dfb53616e90612673b0d563cd47f5ca739749c70R32). Therefore, I haven't made a one-time data backfill script to populate the `created_by_id` from PaperTrail's `whodunnit`, since I don't think we have that information.

**Followup question**: Should we change that import statement to a one-at-a-time save, so that PaperTrail versions are always created?
- arguments for: it would let us view the exact object as it was at the time of creation, which we can't currently do; it is more consistent with our other usages of PaperTrail
- arguments against: the mutation would take a performance hit; and now we are storing the created_by_id on the table anyway

## Type of change
New feature


db/warehouse/migrate/20260112144747_add_created_by_to_ce_opportunity.rb
```diff
@@ -0,0 +1,20 @@
...
+class AddCreatedByToCeOpportunity < ActiveRecord::Migration[7.2]
+  def change
+    safety_assured do
+      # created_by_id refers to the users table, but no foreign key constraint
+      # because users are in a different database boundary (app db vs warehouse db)
+      add_reference :ce_opportunities, :created_by, null: true, index: false
+    end
+  end
+end
+
```
gigxz, line 18: The diff on `db/warehouse_structure.sql` is too big to render - can you redo the migration locally to get just the relevant changes? Thanks!

gigxz (commented):
Looks good! The structure file just needs to be cleaned up.
I'm also unable to see the test failures, I think github may just be having trouble.

martha:
discussed - we should change the `import!` to each save

drivers/hmis/app/graphql/mutations/ce/mark_units_available.rb
```diff
@@ -28,7 +28,7 @@ def resolve(unit_ids:)
       validate_unit_availability_against_referral_postings(units, project)
 
       rule_resolver = Hmis::Ce::Match::UnitGroupRuleResolver.new
-      opportunities = units.map { |unit| unit.build_ce_opportunity(rule_resolver: rule_resolver) }
+      opportunities = units.map { |unit| unit.build_ce_opportunity(rule_resolver: rule_resolver, created_by: current_user) }
       Hmis::Ce::Opportunity.import!(opportunities)
```
gigxz:
```
      opportunities.each(&:save!)
```

PR #6159, opened 2026-01-13
8663 job re-queue improvement

ttoomey:
## _Merging this PR_

### Description
Standardizes job retry logic in `BaseJob`, DRY up `RunReportJob` and `GenericReportJob` to use the shared `requeue_at` method when detecting concurrent report executions. Use supports_idempotent_retry? to simplify max_attempts logic

No behavior changes are intended

### Type of Change
Refactor


app/jobs/warehouse_reports/generic_report_job.rb
```diff
@@ -18,50 +18,45 @@ class GenericReportJob < BaseJob
...
-        report_completed = false
-        klass = allowed_reports[report_class]
-        if klass
-          report = klass.find_by(id: report_id)
-          # Occassionally people delete the report before it actually runs
-          return unless report.present?
+      # Attempt to acquire an advisory lock to ensure only one instance of this report class runs at a time.
+      # If the lock is already held by another worker, the job will be postponed.
+      lock_acquired = false
+      execution_result = ApplicationRecord.with_advisory_lock(advisory_lock_name(report_class), timeout_seconds: 0) do
```
eanders: This moves the advisory lock to a different database, do we care?  Previously it was accessing the warehouse database.

spec/jobs/base_job_spec.rb
```diff
@@ -59,21 +59,40 @@ def perform
...
         new_job = Delayed::Job.last
+        expect(new_job.id).not_to eq(original_id)
         expect(new_job.run_at.to_i).to eq(timestamp.to_i)
         expect(new_job.attempts).to eq(0)
         expect(new_job.locked_at).to be_nil
+        expect(new_job.locked_by).to be_nil
+        expect(new_job.failed_at).to be_nil
+        expect(new_job.last_error).to be_nil
+
+        # Verify original record remains unchanged
```
eanders: Doesn't the original job usually get cleaned up by DJ?  I believe there should be no record of the original job after the new one is queued and the original worker finishes the job.

app/jobs/warehouse_reports/generic_report_job.rb
ttoomey: I think we mostly don't care; but on second thought I guess if we deployed it while a long-running job was running, the check might not work. For that edge case I'll change it to the warehouse base class

spec/jobs/base_job_spec.rb
```diff
@@ -100,4 +133,34 @@ def perform
       end
     end
   end
+
+  describe 'retry behavior and initial attempts' do
+    let(:max_attempts) { Delayed::Worker.max_attempts || 25 }
```
eanders: It seems to have added the magic `|| 25` in a handful of places.  I'd pull that out.

eanders (approved):
I think this looks good if we remove the `|| 25`

PR #6160, opened 2026-01-13
Add Decline Reason for CE Referrals

martha:
## _Merging this PR_

## Description
- Add a new `ce_decline_reason` table and `Hmis::Ce::ReferralDeclineReason` model
- Add `ReferralMessageHandler` functionality for setting and clearing decline reason
  - ~Pass `all_submitted_values` with referral message, which wasn't previously accessible from within the message handler~
- Update the PH workflow template builder to create new draft versions of the PH direct referral templates
  - Previously this builder was destructive on rerun. Now, it's intended to be idempotent on rerun (until publish)
  - See the changes in `README_FOR_PH_CE_WORKFLOWS.md` for details, and please let me know if you like this approach! I could think of a couple other good paths and am not attached to this one.
- Update the shared `CeBuilderUtils` to add functions like create draft, publish, unpublish
- Fix misc non-manifesting bugs with direct referrals
  - Don't show a unit group in the dropdown for sending a direct referral unless it has a direct referral form definition. Otherwise, the frontend will get a 500 when it can't load the form
  - Unit Group's `workflow_template` and `direct_referral_workflow_template` relations should scope to only published templates. Otherwise, when a draft template exists, the frontend sees an error when trying to set the workflow template due to [this validation](https://github.com/greenriver/hmis-warehouse/blob/mke/8624-decline-reason/drivers/hmis/app/models/hmis/unit_group.rb#L123)

## Type of change

- [x] I have provided testing instructions in this PR or the related issue (or not applicable)- see comment on linked issue


drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -126,6 +130,22 @@ def set_custom_referral_status(status_key:)
       referral.update!(custom_status: status)
     end
 
+    def set_referral_decline_reason(message) # rubocop:disable Naming/AccessorMethodName
+      decline_reason_field = message.params['decline_reason_field'] || 'decline_reason'
```
martha: I was thinking this could get more flexible in the future, if for example we need to set the decline reason based on combinations of different values submitted on different steps (like provider decline + admin approve decline step). We might do something like: make the message `trigger_config` accept another argument, not `decline_reason_field` but just the decline reason itself, and then use conditional gateways based on response values to set the specific decline reason for that combination of responses.

I think what I have here is sufficient to meet the needs we have now, so I would propose to kick the can on adding that complexity, but let me know what you think.

drivers/hmis/lib/ce_workflows/ph/workflow_builder.rb
```diff
@@ -51,59 +60,83 @@ def build_outreach_referral_workflow
...
       # Step 1: Send referral
-      send_referral_task = Hmis::WorkflowDefinition::UserTask.create!(
+      send_referral_task = Hmis::WorkflowDefinition::UserTask.find_or_initialize_by(
         name: outgoing_step_name,
         form_definition_identifier: outgoing_step_form_identifier,
         template: template,
-        swimlane: provider_swimlane, # Swimlane is irrelevant since this is just for direct referrals and is completed by the sending project
       )
+      # Swimlane is mostly irrelevant since this is just for direct referrals and is completed by the sending project.
+      # But, it's less confusing to use a different swimlane to avoid confusion in the UI, especially with default contacts.
```
martha, line 104: This is an unrelated change, but I figured I'd bring it in scope while we are updating the templates.

This is related to the Default Contacts work, because I noticed it's confusing when the outgoing task is listed as one of the _provider_ tasks in the helper text. 

Solving it with an additional swimlane isn't perfect either, because admins will then see a Referrer column with a bunch of "Missing" values, which doesn't make sense for an outgoing referral task. We might consider adding a flag on swimlane to indicate swimlanes that shouldn't be included in default contacts? Anyway, we can discuss that more on the other ticket/PR, I just wanted to give the backstory here to explain the change.

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -177,30 +267,42 @@ def self.create_gateway(template, name, gateway_type: 'exclusive')
       )
     end
 
+    def self.find_or_create_gateway(template, name, gateway_type: 'exclusive')
```
martha, line 263: Somewhat messily, I've updated the other util methods (like task creation, event creation, etc.) to internally use `find_or_create`, but this one I've needed to create a separate method to avoid failures in AC-specific system tests. This is somewhat messy but I was trying to avoid the need to change any code/workflows in AC-world, so maybe it is an acceptable tradeoff.

A possible cleanup would be to rename the rest of these methods to make it clear that they do `find_or_create` internally, let me know if that change sounds worthwhile to you.

drivers/hmis/spec/models/hmis/ce/referral_workflow_spec.rb
```diff
@@ -200,7 +200,7 @@
     end
 
     let(:enrollment_task) do
-      create(:hmis_workflow_definition_user_task, template: template, name: 'income check task')
+      create(:hmis_workflow_definition_user_task, template: template, name: 'enrollment task')
```
martha: Unrelated change, this just seemed like a typo

drivers/hmis/spec/models/hmis/ce/referral_workflow_spec.rb
```diff
@@ -253,6 +253,7 @@
       end
 
       it 'the first step can not be rolled back' do
+        instance.steps.reset
```
martha: This fixes a test failure that was introduced by the addition of a call to `all_submitted_values` in `process_triggers`. Here is what I think is happening, but my understanding is not perfect.
- Before the addition of `all_submitted_values`: `instance.steps` wasn't loaded in Rails cache, so when it was called here in the test on line 257-258, we were loading fresh data straight from the database.
- Now though, `all_submitted_values` calls `instance.steps.reset`, loading step data into the Rails cache during the `complete_step!` procedure. Later, here in the test, when we try calling `instance.steps` on lines 257-258, we're getting outdated cached steps unless we reload, which is the fix I've introduced on line 256.

For this specific test change, I think this is an ok fix. In the real app we aren't calling `may_undo_complete_step` anywhere; we are not undoing steps or exposing that functionality to our users.

More broadly, I've tried to think through whether the addition of `all_submitted_values` might have caused any _other_ hidden issues due to the same caching change. (For example, I checked: when completing a step from the UI, does the graphql api still correctly return that the step you just completed is completed now? Yes, everything looked normal to me here.) Overall, I think I feel comfortable merging this change, once the test suite passes, but let me know if you can think of anything I might be missing.

drivers/hmis/app/models/hmis/ce/referral.rb
```diff
@@ -259,5 +261,11 @@ def consistent_project
 
       errors.add(:target_enrollment, 'must be in same project as referral') unless target_enrollment.project == target_project
     end
+
+    def no_decline_reason_when_accepted
+      return unless accepted? && decline_reason.present?
+
+      errors.add(:decline_reason, 'cannot be set on an accepted referral')
```
gigxz: I could see this being easy to accidentally misconfigure, if you forget to `clear_referral_decline_reason` in all the right places. What do you think about making this more forgiving, and simply having a presave hook that clears out decline_reason on the referral if the referral is accepted?
(Thinking about the fact that, as written, this kind of misconfiguration would prevent users from completing workflows. Is it worth it?)

drivers/hmis/app/models/hmis/ce/referral_decline_reason.rb
```diff
@@ -0,0 +1,23 @@
...
+
+module Hmis::Ce
+  class ReferralDeclineReason < GrdaWarehouseBase
+    acts_as_paranoid
+    has_paper_trail
+
+    belongs_to :data_source, class_name: 'GrdaWarehouse::DataSource'
+
+    validates :key, presence: true, uniqueness: { scope: :data_source_id }
+    validates :name, presence: true
```
gigxz: Typically we don't add rails validations for things that are already validated by the database. Is there some reason we need this?

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
gigxz: Interesting!

What you're describing sounds more like how we use the CE Event result message, where you have script task for every possible result, with a separate gateway connecting each one with a `condition` based on a form response (see provider_rejects_ce_event_task in Ac::WorkflowBuilder).

That's honestly what I was _thinking_ we would do for decline reasons, but I agree your approach here is much simpler from a workflow setup perspective. It seems to have an interesting dependency from the form to the decline reason though, and some limitations on what you can do in the form.


Am I understanding correctly that the task form would need to be set up like this?
```json
{
   "type": "CHOICE",
   "text": "select a decline reason",
   "pick_list_options": [
       {
            "code":  "unable_to_locate",  <=== MUST match key in Referral Decline Reasons table
            "label": "Unable to locate client",
      },
      ...
```

gigxz (commented):
review in progress 🚶

drivers/hmis/app/models/hmis/ce/referral_decline_reason.rb
martha: No, I'll remove! :)

drivers/hmis/app/models/hmis/ce/referral.rb
martha: Yes, I waffled about this  so thanks for the feedback! I agree with your suggestion, I'll make that change

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
martha: > What you're describing sounds more like how we use the CE Event result message, where you have script task for every possible result, with a separate gateway connecting each one with a condition based on a form response (see provider_rejects_ce_event_task in Ac::WorkflowBuilder).

yes exactly!

> Am I understanding correctly that the task form would need to be set up like this?

Yes that's right, it does create a dependency from the form to the decline reasons. This PR proposes a [helper](https://github.com/greenriver/hmis-warehouse/pull/6160/changes#diff-4e5cff9ef8168205408c423389c379000a89e158c665e0953e53a096cbbb75ccR81-R91) `ensure_decline_reasons_from_form_item` to set those up, but definitely one compelling argument _against_ this approach is that after setup it could be hard to ensure that they stay in sync.

I think the simplicity tradeoff might be worth it, for now. But if you are not a fan of this approach I'm happy to go down the other avenue

gigxz: thoughts for convo:


- workflow configuration is simple (not so many new gateways/script tasks /etc)
- introduces dependency on specific link_id
  - not an entirely new pattern, we have MOVE_IN_DATE_LINK_ID and COC_CODE_LINK_ID operating in much the same way. Also there is a way provided to specify the link id optionally.
- PickListOption codes must match codes in decline reason table; without validation on form def
  - I think we may want to change the approach to explicitly define decline reasons (like we do with custom statuses), rather then populate them based on whatever values are found in the forms. This would help prevent against accidentally having a bunch of duplicate-ish reasons that just have slightly different wording. Customers should be thinking about defining/managing their set of decline reasons upfront, since they are key to reporting. (And for reporting we likely need to support rolling them up into groups; but I don't think we need to worry about that here). Workflow builder could have logic for _validating_ them rather than populating them
- new pattern: start passing `all_submitted_values` to message handlers. I think this is ok?

gigxz: notes:
- ce event gateway appraoch would be hard to maintain
- pull out link id into const just expect it to be DECLINE_REASON_LINK_ID
- see if we can use `message.step.submitted_values` and only allow updating decline reason directly after being collected. all_submitted_values might be hard to reason about and had a caching issue

drivers/hmis/spec/models/hmis/ce/referral_message_handler_spec.rb
```diff
@@ -62,4 +62,65 @@
...
+      client_acceptance = engine.active_steps.sole
+      engine.start_step!(client_acceptance, user: hmis_user)
+      engine.complete_step!(client_acceptance, user: hmis_user, submitted_values: { 'client_accepted' => '1' })
+    end
+
+    it 'updates the referral decline reason from submitted values' do
+      expect do
+        provider_acceptance = engine.active_steps.sole
+        engine.start_step!(provider_acceptance, user: hmis_user)
+        provider_acceptance.form_definition = ce_step_definition # this is set in the mutation, not the engine complete_step!
```
martha, line 99: This follows the pattern from https://github.com/greenriver/hmis-warehouse/blob/5413cfd3028e066a827638e88556532da3ee62d9/drivers/hmis/spec/models/hmis/ce/referral_enroller_spec.rb#L312-L313

but maybe the necessity for this in the unit tests is an indication that we _should_ move the setting of the step's form_definition into the engine? I have to think this through a bit more from a separation-of-concerns perspective.

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -57,4 +58,27 @@ def validate_nodes_reachable(record)
...
+
+      unless form.link_id_item_hash[link_id].present? && form.link_id_item_hash[link_id]['pick_list_options'].present?
+        record.errors.add(:base, "Decline reason form '#{form.identifier}' must collect decline reason on a choice item with link_id '#{link_id}'")
+        next nil
+      end
+
+      form.link_id_item_hash[link_id]['pick_list_options'].map { |option| option['code'] }
+    end.compact.flatten.to_set
+
+    decline_reasons = Hmis::Ce::ReferralDeclineReason.where(key: decline_reason_keys)
```
gigxz:
```suggestion
    decline_reasons = Hmis::Ce::ReferralDeclineReason.where(key: decline_reason_keys, data_source_id: record.data_source_id)
```

drivers/hmis/app/models/hmis/ce/referral_message_handler.rb
```diff
@@ -126,6 +131,26 @@ def set_custom_referral_status(status_key:)
...
     end
 
+    def set_referral_decline_reason(message) # rubocop:disable Naming/AccessorMethodName
+      form_definition = message.step.form_definition
+
+      raise "Trying to set decline reason for referral #{referral.id}, step #{message.step.id}, but form definition is nil. This probably indicates a mistake in the workflow configuration." unless form_definition.present?
+      raise "Trying to set decline reason for referral #{referral.id}, step #{message.step.id}, but form definition '#{form_definition.identifier}' doesn't collect it. This probably indicates a mistake in the workflow configuration. The form must collect decline reason on an item with link_id '#{DECLINE_REASON_LINK_ID}'" unless form_definition.link_id_item_hash[DECLINE_REASON_LINK_ID].present?
+
+      decline_reason_key = message.step.submitted_values&.fetch(DECLINE_REASON_LINK_ID, nil)
+      return unless decline_reason_key.present?
```
gigxz, line 142:
```suggestion
      # Skip if no decline reason submitted. This can happen when the form only conditionally collects decline reason, or if selecting the decline reason is optional.
      return unless decline_reason_key.present?
```

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -57,4 +58,27 @@ def validate_nodes_reachable(record)
...
+
+  def validate_decline_reasons(record)
+    decline_reason_nodes = record.nodes.filter do |node|
+      node.trigger_config&.any? { |trigger| trigger['message'] == 'set_referral_decline_reason' } || false
+    end
+
+    link_id = Hmis::Ce::ReferralMessageHandler::DECLINE_REASON_LINK_ID
+
+    decline_reason_keys = decline_reason_nodes.map do |node|
+      form = node.form_definition
```
gigxz: Checking `node.form_definition` on a non-UserTask will fail, since they don't have that association. Suggest adding a descriptive message in that case:

```suggestion
      unless node.user_task?
        record.errors.add(:base, "Node '#{node.name}' is not a user task, so it cannot set a decline reason.")
        next
      end
      form = node.form_definition
```

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -57,4 +58,27 @@ def validate_nodes_reachable(record)
     unreachable = record.graph.unreachable_nodes
     record.errors.add(:base, "The following nodes are unreachable: #{unreachable.map(&:name).join(', ')}") if unreachable.any?
   end
+
+  def validate_decline_reasons(record)
```
gigxz: Suggest adding a header comment
```suggestion
  # Validates that nodes with decline reason triggers are properly configured:
  # - Nodes with 'set_referral_decline_reason' triggers must be user tasks
  # - The form on each such node must have a choice item with the decline reason link_id
  # - All decline reason codes collected by the forms must exist in the database as ReferralDeclineReasons
  def validate_decline_reasons(record)
```

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
```diff
@@ -17,6 +17,7 @@ def validate(record)
     validate_end(record)
     validate_tasks_and_gateways(record)
     validate_nodes_reachable(record)
+    validate_decline_reasons(record)
```
gigxz: Just noticed we are blurring the boundary here and adding CE-specific validation into the workflow definition module. I'm not seeing an obvious place to move this to. One thing we could do is conditionally validate based on the template type
```suggestion
    validate_decline_reasons(record) if record.template_type == 'ce_referral'
```

drivers/hmis/lib/ce_workflows/ph/README_FOR_PH_CE_WORKFLOWS.md
```diff
@@ -4,30 +4,24 @@
...
 
-These workflows expect client-specific forms to be available.
+### Usage
+These workflows are generated and updated using the `CeWorkflows::Ph::WorkflowBuilder` utility class. 
 
-The forms can be loaded with `CLIENT=client rails driver:hmis:seed_definitions` or `HmisUtil::JsonForms.new(env_key: 'client').seed_record_form_definitions(roles: [:CE_REFERRAL_STEP])`
+These workflows expect client-specific forms to be available. The forms can be loaded with `CLIENT=client rails driver:hmis:seed_definitions` or `HmisUtil::JsonForms.new(env_key: 'client').seed_record_form_definitions(roles: [:CE_REFERRAL_STEP])`
 
-### Direct Referral Workflow
+#### Updates
```
gigxz, line 19: Can you add a note here about the relationship to the task FormDefinitions? Those are "managed in version control" and unversioned, so they are basically independent of the workflow changes. I just think that may be confusing in the future so is worth calling out.

drivers/hmis/lib/ce_workflows/ph/workflow_builder.rb
```diff
@@ -51,59 +71,75 @@ def build_outreach_referral_workflow
...
+        version: 1,
       )
     end
 
-    def build_direct_referral_workflow(identifier:, name:, outgoing_step_form_identifier:, outgoing_step_name:)
-      CeWorkflows::Shared::CeBuilderUtils.delete_template_and_associated_data(identifier) unless @unsafe_run_in_production
-
-      template = CeWorkflows::Shared::CeBuilderUtils.create_template(identifier, name, @data_source)
+    def build_direct_referral_workflow(identifier:, name:, outgoing_step_form_identifier:, outgoing_step_name:, version: 0)
+      template = CeWorkflows::Shared::CeBuilderUtils.create_draft_template(
```
gigxz: maybe we can rename this method to `find_or_create_draft_template`? This is a bit confusing because without looking into the method you would think it's always creating a new template

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -132,41 +162,90 @@ def self.create_template(identifier, name, data_source)
       )
     end
 
+    def self.create_draft_template(identifier:, name:, data_source:, version: 0)
```
gigxz: I think this could use a brief code comment to explain how its used / why version is hard-coded

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -132,41 +162,90 @@ def self.create_template(identifier, name, data_source)
...
+
+      Hmis::WorkflowDefinition::Template.transaction do
+        template.update!(status: 'draft')
+        template_to_republish&.update!(status: 'published')
+      end
+    end
+
     def self.create_start_event(template)
-      Hmis::WorkflowDefinition::StartEvent.create!(
+      event = Hmis::WorkflowDefinition::StartEvent.find_or_initialize_by(
```
gigxz: can rename to find_or_create start event?

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -132,41 +162,90 @@ def self.create_template(identifier, name, data_source)
...
+        {
+          event: 'start_workflow',
+          message: 'start_referral',
+        },
+      ]
+      event.save!
+      event
     end
 
     def self.create_accept_event(template, name: 'Referral Accepted', update_ce_event: false)
```
gigxz: find or create accept event?

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
```diff
@@ -177,30 +256,42 @@ def self.create_gateway(template, name, gateway_type: 'exclusive')
...
 
+    def self.find_or_create_gateway(template, name, gateway_type: 'exclusive')
+      Hmis::WorkflowDefinition::Gateway.find_or_create_by!(
+        template: template,
+        gateway_type: gateway_type,
+        name: "#{gateway_type.capitalize} Gateway: #{name}",
+      )
+    end
+
     def self.create_decline_event(template, name: 'Referral Declined', ce_event_result: nil)
```
gigxz: find or create declien event

gigxz (commented):
Looks good! in-progress review, next thing I will do is run the publish cycle locally to test that!

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
martha: Thanks, good point, version doesn't need to be hard-coded here.

drivers/hmis/lib/ce_workflows/shared/ce_builder_utils.rb
martha: thanks yes, I'll do these renames per your feedback. (mentioned this in an [earlier comment thread](https://github.com/greenriver/hmis-warehouse/pull/6160#discussion_r2688211471) but the reason I originally skipped them is that I was trying to avoid changes that would create a diff in AC-world)

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
martha: Thanks, good point. I'm still looking into this one. I agree there isn't a great place to move it but I do think we could try to separate out the concerns a bit more with a separate `CeWorkflowTemplateValidator`.

martha: Ok, [here's](https://github.com/greenriver/hmis-warehouse/pull/6160/commits/5da8c9cb6abc93d88ea36ae8c9e8eb3d0ba7f681) my shot at that.

And here's another (untested) suggestion from claude about using a "registry" pattern to more fully separate the concerns. Maybe worthwhile to maintain clear boundaries between the modules. What do you think?

Option 2: Registry Pattern
Create a validator registry that allows different modules to register their validators:
New file: `drivers/hmis/app/models/hmis/workflow_definition/validators/registry.rb`

```ruby
module Hmis::WorkflowDefinition::Validators
  class Registry
    @validators = {}

    class << self
      def register(template_type, validator_class)
        @validators[template_type] = validator_class
      end

      def validator_for(template_type)
        @validators[template_type]
      end
    end
  end
end
```
Then in an initializer or the CE validator:
```ruby
# Register the CE validator
Hmis::WorkflowDefinition::Validators::Registry.register(
  'ce_referral',
  Hmis::Ce::Validators::ReferralWorkflowTemplateValidator
)
```
Update the main validator:
```ruby
def run_type_specific_validations(record)
  validator_class = Hmis::WorkflowDefinition::Validators::Registry.validator_for(record.template_type)
  validator_class&.new&.validate(record)
end
```

drivers/hmis/lib/ce_workflows/ph/workflow_builder.rb
gigxz: Okay I think this makes sense to me, thanks for explaining!

drivers/hmis/spec/models/hmis/ce/referral_message_handler_spec.rb
gigxz: This feels out of scope to address in this PR

drivers/hmis/app/models/hmis/workflow_definition/validators/workflow_template_validator.rb
gigxz: I think the change you proposed on this branch clear and lightweight, I say we stick with that. thanks!

gigxz (approved):
Nice work!

PR #6166, opened 2026-01-14
Pass additional fields when requesting AHA score

martha:
## _Merging this PR_

## Description
- add lookup catalyst and reason to the graphql schema as optional arguments for the `FetchAhaScore` mutation
- when they are provided, pass them to the external API when requesting AHA score

Related frontend PR: https://github.com/greenriver/hmis-frontend/pull/1354

## Type of change
New feature

### DIFF
--- a/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
+++ b/drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
@@ -133,10 +133,23 @@ module HmisExternalApis::AcHmis
         payload[:dw_client_id__dw_client_id__includes] = mci_uniq_ids.first
       end
 
-      payload[:lookup_catalyst] = lookup_catalyst if lookup_catalyst && LOOKUP_CATALYST_ALLOWED_VALUES.include?(lookup_catalyst)
+      # For lookup catalyst and lookup reasons, don't send unknown values to the external API,
+      # but log to Sentry if we receive unexpected values so we don't silently skip them.
+      if lookup_catalyst.present?
+        if LOOKUP_CATALYST_ALLOWED_VALUES.include?(lookup_catalyst)
+          payload[:lookup_catalyst] = lookup_catalyst
+        else
+          Sentry.capture_message("AHA received unexpected lookup catalyst: #{lookup_catalyst}")
+        end
+      end
 
-      valid_reasons = lookup_reason&.filter { |r| LOOKUP_REASON_ALLOWED_VALUES.include?(r) }
-      payload[:lookup_reason] = valid_reasons if valid_reasons&.any?
+      if lookup_reason.present?
+        valid_reasons, unknown_reasons = lookup_reason.partition { |r| LOOKUP_REASON_ALLOWED_VALUES.include?(r) }
+
+        payload[:lookup_reason] = valid_reasons if valid_reasons.any?
+
+        Sentry.capture_message("AHA received unexpected lookup reason(s): #{unknown_reasons.join(', ')}") if unknown_reasons.any?
+      end
 
       payload
     end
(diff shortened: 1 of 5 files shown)


drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -33,6 +33,22 @@
 ###
 module HmisExternalApis::AcHmis
   class Aha
+    # Lookup catalyst and reason are optional values; if present on the form, they should be submitted when we make the fetch call.
+    # Before submitting, validate against the list of allowed values.
+    # (Validate here, instead of in the graphql schema, so we have flexibility to ignore instead of raise when receiving invalid values)
```
martha, line 38: I'd like a second opinion on this. At first I planned to validate using the graphql schema, and I still have a local branch where I started taking that approach. But I didn't love it for 2 reasons:
- Flexibility, as mentioned in the comment. This form can be managed by users in the form builder. With graphql validation, if a user added another option to the picklist, the fetch call would break
- Not mentioned in the comment, but also a factor: The form definition currently stores the human-readable strings as the picklist codes. Translating those human-readable strings into HmisSchema::Enums, and then translating them back again to pass them to the AHA API, seemed like more hassle than it was worth. And I think we want to avoid making changes to the form definition.

Let me know what you think about this!

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -42,20 +58,16 @@ class Aha
...
     class NoMciUniqueIdError < HmisErrors::ApiError; end
 
-    def fetch_score(client)
+    def fetch_score(client, lookup_catalyst: nil, lookup_reason: nil)
       # Collect MCI unique IDs for this client and all source clients with the same destination client
       clients = [client, *client.destination_client&.source_clients&.to_a]
       mci_uniq_ids = clients.compact.uniq.filter_map do |c|
         c.ac_hmis_mci_unique_id&.value
-      end.uniq
+      end.uniq.sort
```
martha, line 66: This change is to hopefully resolve what I think is a transient error that I saw crop up on an unrelated PR:

https://github.com/greenriver/hmis-warehouse/actions/runs/21050122670/job/60534252108?pr=6160

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
```diff
@@ -112,6 +124,23 @@ def conn
...
+
+      if mci_uniq_ids.size > 1
+        payload[:dw_client_id__dw_client_id__overlap] = mci_uniq_ids.join(',')
+      else
+        payload[:dw_client_id__dw_client_id__includes] = mci_uniq_ids.first
+      end
+
+      payload[:lookup_catalyst] = lookup_catalyst if lookup_catalyst && LOOKUP_CATALYST_ALLOWED_VALUES.include?(lookup_catalyst)
+
+      valid_reasons = lookup_reason&.filter { |r| LOOKUP_REASON_ALLOWED_VALUES.include?(r) }
```
gigxz: I think we should send to sentry if we have unexpected values here. Otherwise if new catalysts/reasons get added, they will just be skipped.

gigxz (approved):

drivers/hmis_external_apis/app/models/hmis_external_apis/ac_hmis/aha.rb
gigxz: I think this is a fine approach. The receiving API doesn't validate the inputs, so I almost would say we don't need to validate the values at all, but, it does feel unsafe to pass along an unknown string to an external API (though we could sanitize etc..). I think this is good 👍

PR #6167, opened 2026-01-15
Pure Ruby Uploader for Health Care files

eanders:
## _Merging this PR_

## Description
* Regression tests for health care file uploads
* Replaces CarrierWave for health files with a local version of file validation
* Does not change file storage or process significantly

## Type of change
* Code clean-up


app/models/health/premium_payment.rb
```diff
@@ -100,40 +122,92 @@ def summary
...
-          sender_code = rmr[1][:value][:raw] rescue nil
-          premium_payment = rmr[3][:value][:raw] rescue nil
-          premium_billed = rmr[4][:value][:raw] rescue nil
-          date_time_qualifier = dtm[0][:value][:raw] rescue nil
-          date = dtm[1][:value][:raw] rescue nil
-          time = dtm[2][:value][:raw] rescue nil
-          coverage_period = dtm[5][:value][:raw] rescue nil
-          adjustment_amount = adx[0][:value][:raw] rescue nil
-          adjustment_reason = adx[1][:value][:raw] rescue nil
+          ent = patient.detect { |h| h.keys.include? :ENT }&.values&.flatten&.map(&:values)&.flatten
```
eanders, line 140: This was all just rubocop cleaning things

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,122 @@
...
+      end
+    end
+
+    private
+
+    def valid_magic_bytes?(file_content, file_extension)
+      case file_extension.downcase
+      when '.csv'
+        # CSV files should contain commas and newlines
+        file_content.include?(',') && file_content.include?("\n")
```
ttoomey: this check seems a bit loose. AI suggested actually attempting to parse:
```ruby
        # 1. Must be printable text (no binary null bytes)
        # 2. Should not contain HTML tags
        # 3. Must contain at least one comma and one newline
        # 4. Try parsing the first line as CSV
        return false if file_content.include?("\x00")
        return false if file_content.include?('<html>') || file_content.include?('<!DOCTYPE')
        return false unless file_content.include?(',') && file_content.include?("\n")

        begin
          first_line = file_content.lines.first
          return false if first_line.blank?

          CSV.parse_line(first_line)
          true
        rescue CSV::MalformedCSVError
          false
        end
```

app/models/health/health_file.rb
```diff
@@ -62,23 +90,83 @@ def set_calculated!(user_id, client_id)
...
+        Rails.logger.error("Failed to process file_cache: #{e.message}")
       end
     end
+
+    # Set file_cache for form resubmission support
+    def set_file_cache
+      return unless content.present?
+
+      self.file_cache = {
+        content: Base64.encode64(content),
```
ttoomey, line 166: So this is to solve the UX problem where a submitted form should persist the file data inline, to allow the user to correct validation issues?

* The file content may be PHI, are there any concerns this html page could get cached or exposed? Maybe this could be tighter
* The file hidden field be quite large with base64 encoding

What if we just cleared the field and forced the user to attach it again to the form if validation fails?

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,122 @@
...
+      # Check file size first
+      if file_content.bytesize > 25.megabytes
+        return {
+          valid: false,
+          error: "File is too large (#{(file_content.bytesize / 1.megabyte).round(1)} MB). Maximum size is 25 MB.",
+        }
+      end
+
+      # Write content to temporary file for Marcel to analyze
+      require 'tempfile'
```
ttoomey: probably move outside of class definition (also would expect this to already be required, a require may not be needed)

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,122 @@
...
+
+module FileContentValidator
+  extend ActiveSupport::Concern
+
+  class_methods do
+    def validate_file_content(file_content, claimed_content_type, allowed_types, file_extension)
+      return { valid: false, error: 'No file content provided' } if file_content.blank?
+
+      # Check file size first
+      if file_content.bytesize > 25.megabytes
```
ttoomey, line 17: do users ever hit this limit? Perhaps we should pull this out to an app config property

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,122 @@
...
+        temp_file.binmode
+        temp_file.write(file_content)
+        temp_file.rewind
+
+        # Use Marcel to detect MIME type from the actual file content (without filename)
+        detected_type = Marcel::MimeType.for(temp_file.path)
+
+        # Marcel correctly detects XLSX files as application/zip since they are ZIP archives
+        # We should accept this and normalize it to the expected type
+        detected_type = 'application/vnd.openxmlformats-officedocument.spreadsheetml.sheet' if detected_type == 'application/zip' && file_extension.downcase == '.xlsx'
```
ttoomey: what if we give marcel the name with extension?
```suggestion
        # Use Marcel to detect MIME type from the actual file content and filename hint
        detected_type = Marcel::MimeType.for(temp_file, name: "upload#{file_extension}")
```

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,122 @@
...
+
+        # If Marcel can't detect the type from content, it returns application/octet-stream
+        # In that case, we need to validate the magic bytes manually for security
+        if detected_type == 'application/octet-stream'
```
ttoomey: do we want plain csvs in here too?
```suggestion
        if detected_type == 'application/octet-stream' || detected_type == 'text/plain'
```

ttoomey (commented):
looking good; initial review feedback

app/controllers/warehouse_reports/health/premium_payments_controller.rb
```diff
@@ -34,8 +34,12 @@ def create
         content: premium_params[:content].read,
         original_filename: premium_params[:content].original_filename,
       )
-      Health::ConvertPaymentPremiumFileJob.perform_later
-      respond_with(@file, location: warehouse_reports_health_premium_payments_path)
+      if @file.persisted?
+        Health::ConvertPaymentPremiumFileJob.perform_later
+        redirect_to warehouse_reports_health_premium_payments_path, notice: 'Premium Payment File (820) was successfully created.'
+      else
+        redirect_to warehouse_reports_health_premium_payments_path, alert: @file.errors.full_messages.join(', ')
```
ttoomey, line 41: Is this correct? Often we render `new` from create for a failed form to allow the user to correct the issue but maybe the workflow is such that a redir makes sense

app/models/concerns/file_content_validator.rb
eanders: Thanks.  This is probably worth some iteration.
I'll see if tests pass with the above suggestion.

app/models/health/health_file.rb
eanders: Basically, yes.  This is to provide the same level of functionality we get with CarrierWave where if you upload a file and the form has an error, it doesn't lose the file, which can cause a different error when you resubmit and don't notice you have to re-upload.  
I don't know why this would be an issue in terms of PHI, it would only show for the user who recently uploaded the same file.  I'll double check we're not caching the file anywhere else.  I agree, base64 encoding isn't ideal for large files, happy to entertain a different approach.

app/models/concerns/file_content_validator.rb
eanders: Yes, people sometimes upload many hundred MB PDFs (like those scanned by some scanners.)  Basically this is just a sanity check that someone isn't uploading something huge.  The forms generally allow larger files for HMIS CSV zip files and such, but these should never be that large.

app/models/concerns/file_content_validator.rb
eanders: I'll give that a try, thanks!

app/models/concerns/file_content_validator.rb
eanders: I need to double check, but in at one place where we said we accepted Excel and CSV, the code didn't know how to handle CSVs...

eanders:
@ttoomey I took another stab at tightening the file type validations, it wasn't quite as simple as hinting Marcel since that caused it to incorrectly identify the file base on the extension.  Let me know if you think it needs additional work!

app/models/health/health_file.rb
ttoomey: Reference to [PR](https://github.com/greenriver/hmis-warehouse/pull/6174) with example. However this doesn't work as s3 doesn't meet our requirements.

For a cleared file, could we just use an HTML required tag to block the submission? (might take some adjustment to the containing `form` to remove `novalidate` ( simple form adds it)
`<input type="file"  id="user-file" name="uploaded_file"  required>`

Otherwise I guess bas64 encoding is the best we can do

app/models/concerns/file_content_validator.rb
```diff
@@ -0,0 +1,127 @@
...
+      case file_extension.downcase
+      when '.csv'
+        # CSV has no magic bytes; validate by parsing a small text sample.
+        sample = file_content.byteslice(0, 256.kilobytes)
+        # CSV files should not contain null bytes
+        return false if sample.include?("\x00")
+
+        text = sample.sub(/\A\xEF\xBB\xBF/, '').encode('UTF-8', invalid: :replace, undef: :replace, replace: '')
+        # CSV files should contain commas and newlines
+        rows = CSV.new(StringIO.new(text), col_sep: ',', liberal_parsing: true).take(5)
```
ttoomey: Don't crash on invalid csv?
```suggestion
        begin
          rows = CSV.new(StringIO.new(text), col_sep: ',', liberal_parsing: true).take(5)
        rescue CSV::MalformedCSVError
          return false
        end
```

app/controllers/warehouse_reports/health/enrollments_disenrollments_controller.rb
```diff
@@ -32,10 +32,20 @@ def report
...
       file = params.dig(:report, :file)
       if file.present?
-        @enrollment_reasons = Health::EnrollmentReasons.create(
-          file: file,
+        @enrollment_reasons = Health::EnrollmentReasons.new(
+          file: file.original_filename,
           content: file.read,
+          content_type: file.content_type,
         )
+
```
ttoomey: rewind after read on ln 37? (If file is read again during validation)
```suggestion
          file.rewind
```

ttoomey (approved):
Left some additional suggestions. Looks good!

app/controllers/warehouse_reports/health/premium_payments_controller.rb
eanders: There is no `new` for this report.

PR #6169, opened 2026-01-15
Don't report to sentry if referral is missing source project

gigxz:
## _Merging this PR_

## Description
Issue: https://github.com/open-path/Green-River/issues/8736

## Type of change
Bug fix


drivers/hmis/app/models/hmis/auth_policies/user_context.rb
```diff
@@ -33,6 +33,7 @@ def initialize(user)
 
   # Project-specific permissions
   def project_permissions(project_id)
+    return EMPTY_SET if project_id.blank?
```
gigxz, line 36: This isn't needed to fix the immediate issue, but figured it would help to prevent this from occurring in other cases where we might be passing nil to this method.

martha (approved):
Thanks!

PR #6131, opened 2025-12-31
improve type validation on global auth policies

ttoomey:
## _Merging this PR_
- use the squash-merge strategy for PRs targeting a release-X branch

# Description
Implements a factory-based authorization policy system to strictly separate global and instance-level permission checks.

- **Auth System**: Adds a `for_resource` factory that dispatches to `Global` or `Instance` policy subclasses, ensuring type-safe access to resource-specific logic.
- **User Integration**: Updates the main policy entry point in `Hmis::User` to support the new factory-based instantiation.

### Type of Change
Refactor

## Checklist before requesting review
[//]: # (Remove any items that are not applicable)
- [x] I have performed a self-review of my code
- [ ] I have run the code that is being changed under ideal conditions, and it doesn't fail
- [ ] I have updated the documentation (or not applicable)
- [x] I have added spec tests (or not applicable)
- [ ] I have provided testing instructions in this PR or the related issue (or not applicable)

PR #6173, opened 2026-01-16
Refactor HMIS auth policies to split out Global vs Instance

gigxz:
## _Merging this PR_

## Description

Implements a factory-based authorization policy system to strictly separate global and instance-level permission checks.

- **Auth System**: Adds a `for_resource` factory that dispatches to `Global` or `Instance` policy subclasses, ensuring type-safe access to resource-specific logic.
- **User Integration**: Updates the main policy entry point in `Hmis::User` to support the new factory-based instantiation.

(Heavily based on https://github.com/greenriver/hmis-warehouse/pull/6131. Split out from client policy changes to make reviewing easier)

## Type of change
Refactor


gigxz:
There was one test failure https://github.com/greenriver/hmis-warehouse/actions/runs/21361171012/job/61480695166 that seems unrelated, seeing if its resolved by rerun with latest branch

drivers/hmis/app/models/hmis/auth_policies/base_policy.rb
```diff
@@ -24,16 +28,23 @@ def initialize(context:, resource:)
   # convenience
   def user = context.user
 
-  # sanity check, is the resource what we expect
-  def validate_resource!(_arg) = raise 'override this in child class'
+  # Get global permissions for the user (across all projects)
```
martha: If we keep this, maybe we should clarify in the comment:
```suggestion
  # Get global permissions for the user (across all projects/data sources)
```
(however, see my other comment below about `global_permissions` in general)

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -13,93 +13,99 @@
...
+      project_id = context.referral_source_project_id(referral.id)
+      context.project_permissions(project_id)
+    end
+
+    def validate_resource!(arg) = ensure_arg_type!(arg, Hmis::Ce::Referral)
   end
 
-  protected
+  class Global < Hmis::AuthPolicies::BasePolicy
+    def can_index?
```
martha, line 97: Should this be scoped to the user's data source, instead of all permissions the user has globally? See my comment below about `data_source_permissions`

(Separately, I think we should avoid `potential_permissions` because it does not respect permission requirements. If we do decide to keep this scoped broadly, then it should probably use `global_permissions` instead of `potential_permissions`)

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
```diff
@@ -13,93 +13,99 @@
...
+      return false unless (context.potential_permissions & [:can_view_referrals, :can_view_own_referrals]).any?
+      return false unless (context.potential_permissions & [:can_perform_any_referral_tasks, :can_perform_own_referral_tasks]).any?
 
-  # *target* project
-  def project_permissions
-    project_id = context.referral_project_id(referral.id)
-    context.project_permissions(project_id)
-  end
+      true
+    end
```
martha, line 105: This isn't feedback on this PR, but a comment to try to work through how this integrates with current open work on Default Contacts:

Maybe the data-source-level `can_perform_referral_tasks?` perm, currently being introduced in the Default Contacts PR to the "CE Admin" policy, actually belongs here? https://github.com/greenriver/hmis-warehouse/pull/6145/changes#diff-971e879f3da70fdd192f9a5fbc4309f66c2d621a42a7e8fb9361a486d1ca6d6eR17-R20

I'm wondering because I was trying to detangle when we would use a Global permission like this, vs. an Admin permission like we introduced over there. It struck me that this permission is not really "admin" specific. Otoh, we are using it to get at an "admin" concept of "who can be a global default contact?" So, I'm not sure. What do you think?

drivers/hmis/app/models/hmis/auth_policies/staff_assignment_policy.rb
```diff
@@ -6,15 +6,37 @@
 
 # frozen_string_literal: true
 
-class Hmis::AuthPolicies::StaffAssignmentPolicy < Hmis::AuthPolicies::BasePolicy
-  memoize def can_index?
-    return false unless Hmis::ProjectStaffAssignmentConfig.exists?
+class Hmis::AuthPolicies::StaffAssignmentPolicy < Hmis::AuthPolicies::ResourcePolicy
+  module Shared
```
martha: Can you explain or add comments about why `can_index?` should be available in both the Instance and Global policies here? It looks like this is specific to `StaffAssignmentPolicy`; the Referral policy also has a `can_index`, but doesn't use a `Shared` module.

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
```diff
@@ -31,6 +31,11 @@ def initialize(user)
     user.roles.flat_map(&:granted_permissions).to_set.freeze
   end
 
+  memoize def global_permissions
```
martha, line 35: I think I'm the one who introduced `global_permissions`, in a previous version of the ClientPolicy PR, but I'm now thinking it should be removed and any usages of it should be replaced with something more like `data_source_permissions` (introduced in default contacts PR) instead. https://github.com/greenriver/hmis-warehouse/pull/6145/changes#diff-f39d276dcd943c446875dba332d8852d2d83f2f12dc79263e8a96d1dcc3da6c6R43-R55

drivers/hmis/app/models/hmis/auth_policies/staff_assignment_policy.rb
gigxz: Good catch, there is no need for this. That was my bad not reviewing generated code closely enough. I updated to remove the Instance policy since it wasn't used.

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
gigxz: Thanks for pointing this out! I updated it to use `global_permissions`, and updated `global_permissions` to only return permissions for the current data source (Which is considered "global" in the scope of an HMIS request). Commit incoming..

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
gigxz: I'm not totally following, maybe we can jump on a huddle?

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
gigxz: Alright I pushed up the change. Here is the proposed implementation of global_permissions which is pulled from your branch "data_source_permissions" method. I opted to call this "global" because these permissions are considered global within the scope of this UserContext. Definitely open to other suggestions though. We could call it "potential_permissions" for example (that was unused now, so I removed it). In the future, I think we may also need something named data_source_permissions (or similar) to provide the set of permissions that the user has _at the data source level_ specifically (E.g. "can edit data source" ).


In my comment I am trying to be explicit that these are permissions that the user has for **some** entity(ies) in the data source, NOT for **any** entity(ies) in the data source. I feel like the latter is more likely to be misinterpreted to mean that you can do this thing for any entity. I found this hard to communicate!

```ruby
  # Set of permissions that the user has for some entity in the current data source
  # Examples:
  # - User has can_view_project for a Project in this data source => global_permissions includes can_view_project
  # - User has can_view_project for a Project in another data source => global_permissions does not include can_view_project
  memoize def global_permissions
    data_source = GrdaWarehouse::DataSource.find(@data_source_id)

    # All access groups that grant any permission on any entity in the data source
    access_group_ids = ::Hmis::GroupViewableEntity.
      includes_any_entity_in_data_source(data_source).
      pluck(:collection_id)
    permission_loader.for_access_group_ids(access_group_ids)
  end
```

drivers/hmis/app/models/hmis/auth_policies/user_context.rb
gigxz: `global_permissions` has been updated to reflect the set of potential permissions that the user has in the current data source

drivers/hmis/app/models/hmis/auth_policies/resource_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+# class Hmis::AuthPolicies::SomethingPolicy < Hmis::AuthPolicies::ResourcePolicy
+#   class Instance < Hmis::AuthPolicies::BasePolicy; end
+#   class Global < Hmis::AuthPolicies::BasePolicy; end
+# end
+class Hmis::AuthPolicies::ResourcePolicy
+  def self.for_resource(context:, resource:)
+    policy_class = resource.is_a?(Class) ? const_get(:Global) : const_get(:Instance)
+    policy_class.new(context: context, resource: resource)
+  end
+end
```
ttoomey, line 35: This is very concise and is fine as-is. But maybe worth adjusting to 1) allow subclasses to be more flexibly configured 2) provide a more useful error if a subclass is not defined (several polices don't define `Global`)

```suggestion
class Hmis::AuthPolicies::ResourcePolicy
  def self.for_resource(context:, resource:)
    policy_class = resource.is_a?(Class) ? global_policy_class : instance_policy_class
    policy_class.new(context: context, resource: resource)
  end

  def self.global_policy_class
    const_get(:Global, false)
  rescue NameError
    raise NotImplementedError, "#{name} needs a 'Global' nested class"
  end

  def self.instance_policy_class
    const_get(:Instance, false)
  rescue NameError
    raise NotImplementedError, "#{name} needs an 'Instance' nested class"
  end
end
```

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
```diff
@@ -7,27 +7,29 @@
 # frozen_string_literal: true
 
 # Determines a user's permissions for CE Opportunities
-class Hmis::AuthPolicies::CeOpportunityPolicy < Hmis::AuthPolicies::BasePolicy
-  # Accepts client parameter because referral authorization requires validating
-  # the relationship between three entities: user, opportunity, and client.
-  def can_create_referral?(client:)
-    return false unless Hmis::Ce.configuration.enabled?
+class Hmis::AuthPolicies::CeOpportunityPolicy < Hmis::AuthPolicies::ResourcePolicy
+  class Instance < Hmis::AuthPolicies::BasePolicy
```
ttoomey, line 11: This is fine and it's arguably conceptually cleaner to wrap everything in resource policy but it is boilerplate for classes that don't have a need both `Instance` and `Global` perms. Alternative would be to leave these "single-type" classes unchanged and just add this to `BasePolicy`
```ruby
# In BasePolicy
def self.for_resource(...) = new(...)
```

ttoomey (approved):
Looks great, thank you!!

drivers/hmis/app/models/hmis/auth_policies/ce_referral_policy.rb
martha: Great! I think you communicated it well ✨

martha (approved):

drivers/hmis/app/models/hmis/auth_policies/ce_opportunity_policy.rb
gigxz: Interesting idea. I think I'm going to leave it as-is because for several of these, they don't have a global need _yet_ but they likely will later, or vice versa. I think I'll leave them as-is so it's less churn later on to add to it.

drivers/hmis/app/models/hmis/auth_policies/resource_policy.rb
gigxz: Good idea, updated

PR #6179, opened 2026-01-22
Handle New RRH Assessment Name

dtgreiner:
## _Merging this PR_

## Description

Add logic for handling an updated name for the RRH Transfer 2024 Assessment. This will now come through using the transfer name instead of the assessment default where applicable.

## Type of change
Bug Fix

### DIFF
--- a/app/models/grda_warehouse/hud/assessment_question.rb
+++ b/app/models/grda_warehouse/hud/assessment_question.rb
@@ -90,6 +90,7 @@ module GrdaWarehouse::Hud
       [
         'RRH-PSH Transfer',
         'RRH-PSH Transfer 2024',
+        'RRH Transfer 2024',
       ]
     end
 
@@ -108,6 +109,7 @@ module GrdaWarehouse::Hud
       return 'Pathways 2021' if pathways_2021?
       return 'RRH-PSH Transfer' if transfer_2021?
       return 'RRH-PSH Transfer 2024' if transfer_2024?
+      return 'RRH Transfer 2024' if transfer_2024_v2?
       return 'Family Pathways 2024' if family_pathways_2024?
 
       # unknown from assessment questions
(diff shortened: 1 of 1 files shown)


app/models/grda_warehouse/hud/assessment_question.rb
```diff
@@ -108,6 +109,7 @@ def assessment_name
       return 'Pathways 2021' if pathways_2021?
       return 'RRH-PSH Transfer' if transfer_2021?
       return 'RRH-PSH Transfer 2024' if transfer_2024?
```
eanders: Can you double check we don't use `transfer_2024?` or check `RRH-PSH Transfer 2024` elsewhere in the codebase?  If we don't, this looks good to me, if we do, I'd be tempted to simply make the check for which one it is to be `lookup&.response_text.in?(['RRH-PSH Transfer 2024', 'RRH Transfer 2024'])`.

eanders (approved):

app/models/grda_warehouse/hud/assessment_question.rb
dtgreiner: This is the only place these are used (outside of one test). 

I debated on combining them off the bat but if we follow the current pattern, we would have all matching assessments show as either 'RRH Transfer 2024' (or 'RRH-PSH Transfer 2024' if we chose that direction). It's probably not a big deal to have them all show as the updated name, but this way keeps the legacy name showing where applicable. 

We could also adjust it to be slightly off-pattern and pass the name we receive if it matches one of the two. (e.g. `lookup.response_text if transfer_2024?` and have `transfer_2024?` match on `lookup&.response_text.in?(['RRH-PSH Transfer 2024', 'RRH Transfer 2024'])` )

eanders: Let's go with what you wrote, thanks!  One small tweak, can those `transfer_2024?` and similar methods be made private?

dtgreiner: Sounds like a plan to me! I'll get it pushed up

PR #6181, opened 2026-01-23
Add search term filter and scope to CE Candidates

martha:
## _Merging this PR_

## Description
- Add `matching_search_term` scope to CE Candidates
- Add `search_term` filter to FilteredCandidatesQuery
- Add spec

Related frontend PR: https://github.com/greenriver/hmis-frontend/pull/1360

## Type of change
New feature

### DIFF
--- a/drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
+++ b/drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
@@ -25,11 +25,10 @@ module Hmis::Ce
       @unit_group_id = opportunity.unit_group.id
       @exclude_declined_clients = exclude_declined_clients
       @search_term = search_term
-      @base_candidate_scope = Hmis::Ce::Match::Candidate.for_opportunity(opportunity).prioritized
     end
 
     def resolve
-      scope = @base_candidate_scope
+      scope = Hmis::Ce::Match::Candidate.for_opportunity(@opportunity).prioritized
       scope = scope.matching_search_term(@search_term) if @search_term.present?
       return scope unless @exclude_declined_clients
 
--- a/drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
+++ b/drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
@@ -20,19 +20,22 @@ module Hmis::Ce
   class FilteredCandidatesQuery
     def arel = Hmis::ArelHelper.instance
 
-    def initialize(opportunity:, exclude_declined_clients: false)
+    def initialize(opportunity:, exclude_declined_clients: false, search_term: nil)
       @opportunity = opportunity
       @unit_group_id = opportunity.unit_group.id
       @exclude_declined_clients = exclude_declined_clients
+      @search_term = search_term
       @base_candidate_scope = Hmis::Ce::Match::Candidate.for_opportunity(opportunity).prioritized
     end
 
     def resolve
-      return @base_candidate_scope unless @exclude_declined_clients
+      scope = @base_candidate_scope
+      scope = scope.matching_search_term(@search_term) if @search_term.present?
+      return scope unless @exclude_declined_clients
 
       # { source_client_id => most_recent_decline_timestamp }
       most_recent_declines = fetch_most_recent_declines
-      return @base_candidate_scope if most_recent_declines.blank?
+      return scope if most_recent_declines.blank?
 
       # { source_client_id => dest_client_id }
       source_to_dest_client_map = map_source_to_dest_clients(most_recent_declines.keys)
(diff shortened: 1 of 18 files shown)


drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -24,5 +24,23 @@ class Candidate < GrdaWarehouseBase
...
+
+    # Free-text search for Candidate
+    scope :matching_search_term, ->(search_term) do
+      search_term = search_term.strip
+
+      # If it's a possible PK, check if it's a Candidate primary key
+      if possibly_pk?(search_term)
+        matching_candidates = where(id: search_term.to_i)
+        return matching_candidates if matching_candidates.exists?
+      end
```
martha: To discuss: I wasn't sure how useful it is to search by _candidate_ ID (as opposed to source client ID)?

But, candidate ID is what the user sees if they don't have permission to view client names:
Image

so in a way maybe it does make sense for that to be the ID that this table is searchable by. I'm open to feedback on this though.

gigxz: I don't think it's necessary since it's not something we expose to the user. (Aside from this case when they can't see names, like you pointed out, but even then – the candidate id has no real meaning, its not a long-lived identifier for the client. I think its more likely we would want to change that view to show the client ID if we ever have real users viewing waitlists without names)

drivers/hmis/app/models/hmis/ce/match/candidate.rb
```diff
@@ -24,5 +24,23 @@ class Candidate < GrdaWarehouseBase
...
+
+      # Search by client through client_proxy -> destination_client -> hmis_source_clients
+      candidate_ids = joins(client_proxy: { destination_client: :hmis_source_clients }).
+        merge(Hmis::Hud::Client.matching_search_term(search_term)).
+        pluck(:id).uniq
+
+      where(id: candidate_ids)
```
gigxz: I think this can be simplified by using the existing scope on client proxy (which backs the search feature on the admin global eligible clients list). We can avoid some joins and the pluck that way:
```ruby
  scope :matching_search_term, ->(search_term) do
      # Search by client through client_proxy using the ClientProxy scope (searches against destination client)
      joins(:client_proxy).
        merge(Hmis::Ce::ClientProxy.matching_search_term(search_term))
  end
```

drivers/hmis/app/models/hmis/ce/filtered_candidates_query.rb
```diff
@@ -20,27 +20,30 @@ module Hmis::Ce
...
       @opportunity = opportunity
       @unit_group_id = opportunity.unit_group.id
       @exclude_declined_clients = exclude_declined_clients
+      @search_term = search_term
       @base_candidate_scope = Hmis::Ce::Match::Candidate.for_opportunity(opportunity).prioritized
     end
 
     def resolve
-      return @base_candidate_scope unless @exclude_declined_clients
+      scope = @base_candidate_scope
```
gigxz: Does @base_candidate_scope need to be an instance variable? I'm not seeing it used anywhere except here. 

This seems like it could simply be:

```
scope = Hmis::Ce::Match::Candidate.for_opportunity(@opportunity).prioritized
```

martha: Thanks, yes that's much neater

drivers/hmis/app/models/hmis/ce/match/candidate.rb
martha: thanks, nice! This also gets us search by destination and source client ID for free, looks like.

gigxz: Yep it does, so did your code fwiw since it also used the underlying `text_searcher` method thats in a shared concern

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_candidates_spec.rb
```diff
@@ -286,5 +286,128 @@
...
+
+      context 'when searching by source client ID' do
+        let(:variables) do
+          {
+            opportunityId: opportunity.id,
+            filters: { searchTerm: client2.id.to_s },
+          }
+        end
+
+        # todo @martha - will this wokr?
```
gigxz:
```suggestion
```

gigxz (approved):

drivers/hmis/spec/requests/hmis/ce/ce_opportunity_candidates_spec.rb
martha: Ha thank you

PR #6184, opened 2026-01-26
Add analytics views for Units/Unit Occupancy

martha:
## _Merging this PR_

## Description
Adds the analytics views for Units and Unit Occupancy as proposed in the spreadsheet from the linked ticket.

## Type of change
New feature


db/views/analytics_hmis_units_v01.sql
```diff
@@ -0,0 +1,9 @@
+SELECT id,
+  name,
```
martha: Unit name in production is currently always nil, because we don't expose a way to create names when creating units.

Is this field still worth exporting? Should we copy some of [this logic](https://github.com/greenriver/hmis-warehouse/blob/e143ca11033011a749899c2a2552910d15831f9e/drivers/hmis/app/models/hmis/unit.rb#L152-L157) for unit "display name"? I leaned away from doing that because I didn't want to duplicate rails logic in sql views, and the actual information encapsulated in that display name -- unit type and unit ID -- are already available in this view. But let me know what you think.

gigxz: I'd lean towards removing it from the view so it doesn't end up causing confusion / end up in a user-facing dataset. It's something we could add later if/when we start supporting meaningful naming of units (a la https://github.com/open-path/Green-River/issues/8490), and if/when that's needed in reporting

gigxz (approved):

PR #6188, opened 2026-01-26
Add system test for CE Referrals page tabs

martha:
## _Merging this PR_

## Description
Adds a system test for which tabs are shown on the Referrals page. See related frontend logic: https://github.com/greenriver/hmis-frontend/blob/f650ef20ba9551cccb15f6e3c940464e715dfc93/src/modules/ce/components/project/ProjectReferralsPage.tsx#L29-L61

## Type of change
New test

### DIFF
--- a/drivers/hmis/spec/system/hmis/ce/project_referrals_page_spec.rb
+++ b/drivers/hmis/spec/system/hmis/ce/project_referrals_page_spec.rb
@@ -1,0 +1,144 @@
+###
+# Copyright 2016 - 2025 Green River Data Analysis, LLC
+#
+# License detail: https://github.com/greenriver/hmis-warehouse/blob/production/LICENSE.md
+###
+
+# frozen_string_literal: false
+
+require 'rails_helper'
+
+# Tests what tabs are available on the CE referrals page, depending on a combo of user permissions + project config.
+# In the future, coverage could be expanded to test actual referral fields shown.
+RSpec.feature 'CE Project Referrals Page', type: :system do
+  # Use find_or_create_by since some tests create the data source in a before(:all) block for performance reasons.
+  let!(:ds1) { GrdaWarehouse::DataSource.find_or_create_by!(hmis: 'localhost', name: 'HMIS', short_name: 'HMIS', authoritative: true) }
+  let!(:project) { create(:hmis_hud_project, data_source: ds1, ProjectType: 3, with_coc: true) }
+
+  let!(:admin) { create(:hmis_user, data_source: ds1, first_name: 'Alexandra', last_name: 'Admin') }
+  let!(:access_control) { create_access_control(admin, ds1) }
+
+  before(:each) do
+    allow_any_instance_of(Hmis::Ce::Configuration).to receive(:enabled?).and_return(true)
+    sign_in(admin)
+  end
+
+  def expect_tabs(tabs_names)
+    tabs = page.all('[aria-label="Project CE Tabs"] a')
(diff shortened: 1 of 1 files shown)


drivers/hmis/spec/system/hmis/ce/project_referrals_page_spec.rb
```diff
@@ -0,0 +1,144 @@
...
+
+# frozen_string_literal: false
+
+require 'rails_helper'
+
+# Tests what tabs are available on the CE referrals page, depending on a combo of user permissions + project config.
+# In the future, coverage could be expanded to test actual referral fields shown.
+RSpec.feature 'CE Project Referrals Page', type: :system do
+  # Use find_or_create_by since some tests create the data source in a before(:all) block for performance reasons.
+  let!(:ds1) { GrdaWarehouse::DataSource.find_or_create_by!(hmis: 'localhost', name: 'HMIS', short_name: 'HMIS', authoritative: true) }
```
gigxz: Isn't that what the new factory was added for? I see another system test using `let!(:ds1) { create :hmis_primary_data_source, hmis: 'localhost' }`

gigxz (approved):
looks good!

drivers/hmis/spec/system/hmis/ce/project_referrals_page_spec.rb
gigxz: I'm not really following the before-all and why thats an issue, can you point me to where we do that? I'm curious because it seems like some system tests are just fine creating a new data source, and this one isn't - which feels like a future potential for flakiness if tests are dependent on the order they are run.

martha: Oh hm, good question, I'm not sure -- I just copied this from `https://github.com/greenriver/hmis-warehouse/blob/77d50f1df5873a666f2baf2bc74538ce388632f7/drivers/hmis/spec/support/ce_system_test_helper.rb#L14-L15`. Maybe that is outdated. I will try using the factory!

martha: Ah ok, I did a little digging and can explain the `before(:all)` issue, but tl;dr it doesn't impact this test, which can and should just use the new factory as you suggested.

[This system test](https://github.com/greenriver/hmis-warehouse/blob/77d50f1df5873a666f2baf2bc74538ce388632f7/drivers/hmis/spec/system/hmis/ac_hmis/ac_ce_referral_workflows_spec.rb#L15-L41) creates `ds1` in a `before(:all)` group because we reference it when building the workflows, which we want to do only once for the whole test file (not once per test). But we can't reference it inside a `before(:all)` if it's created as a fixture. That test file cleans up after itself, so it should not impact other tests or cause flaky problems with test order. 

However, that test file also uses the [CE System Test Helper](https://github.com/greenriver/hmis-warehouse/blob/77d50f1df5873a666f2baf2bc74538ce388632f7/drivers/hmis/spec/support/ce_system_test_helper.rb#L14-L15) to set up some fixtures, which is _also_ used by another CE system test. So, the CE system test helper uses `find_or_create_by!` when creating the `ds1` in order to be compatible with both. 

That is messy and should probably be cleaned up, but it's unrelated to the current PR. I made the change you suggested, so I'll merge this once tests pass. Thank you!

gigxz: I see, thanks!

